PR #2088, opened 2019-11-13
Automatic run ending in Keras and TensorFlow.Keras autologging

juntai-zheng:

Previously, runs created by `mlflow.tensorflow.autolog()` (for `tf.keras models`) or `mlflow.keras.autolog()` were never automatically ended by `autolog()`. The behavior is now defined as follows:

- `Keras`: If there is no run active when `keras.model.fit()` or `keras.model.fit_generator()` is called, then MLflow will create a run at the start of training and end it at the end of training.
- `tf.keras`: If there is no run active when TensorFlow logs to TensorBoard, or when `tf.keras.model.fit()` is called, then MLflow will create a run and end it at the end of training. This currently does not support `tf.keras.fit_generator()` or `tf.estimator.train`, both of which will have support added in future PRs.

In addition, tests for both TensorFlow and Keras integration were refactored.

## How is this patch tested?

Unit tests that test if automatically started runs are ended appropriately and if manually created runs are persisted appropriately.

- [] Yes. Give a description of this change to be included in the release notes for MLflow users.

Will write a description after `tf.estimator.train` and `tf.keras.fit_generator()` support are added in.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python


tests/keras_autolog/test_keras_autolog.py
```diff
@@ -53,7 +82,32 @@ def generator():
...
+        assert mlflow.active_run()
+        assert mlflow.active_run().info.run_id == run.info.run_id
+
+
+@pytest.fixture
+def keras_random_data_run(random_train_data, fit_variant, random_one_hot_labels, manual_run):
+    if manual_run:
+        mlflow.start_run()
+
+    mlflow.keras.autolog()
```
smurching: Maybe wrap this in a `try-finally` block so that we always end the run? e.g.

```suggestion
try:
    mlflow.keras.autolog()
    data = random_train_data
    labels = random_one_hot_labels
    model = create_model()
    if fit_variant == 'fit_generator':
        def generator():
            while True:
                yield data, labels
        model.fit_generator(generator(), epochs=10, steps_per_epoch=1)
    else:
         model.fit(data, labels, epochs=10)
finally:
    mlflow.end_run()
```

tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -31,33 +31,78 @@ def random_train_data():
...
+    n, n_class = (1000, 10)
+    classes = np.random.randint(0, n_class, n)
+    labels = np.zeros((n, n_class))
+    labels[np.arange(n), classes] = 1
+    return labels
+
+
+@pytest.fixture(params=[True, False])
+def manual_run(request):
+    return request.param
```
smurching: Actually, I wonder if we could actually make this fixture start & end the run so that we don't need to do so in each of our tests, e.g.:


```suggestion
    if request.param:
        mlflow.start_run()
    yield
    mlflow.end_run()
```

The `yield` statement will pause execution of the fixture till after the test completes, and then `mlflow.end_run()` will be called after the test completes

smurching (commented):
One small suggestion, otherwise LGTM!

PR #2094, opened 2019-11-14
TensorFlow autolog automatic run ending

juntai-zheng:

This PR allows autolog to automatically end runs it creates when using `tf.estimator.export_saved_model()`, `tf.estimator.export_savedmodel()`, and `tf.estimator.train()`. In addition, the parameters `steps` and `max_steps` are automatically logged when calling `tf.estimator.train()`.

## How is this patch tested?

Unit tests.

Runs automatically created by `mlflow.tensorflow.autolog()` and `mlflow.keras.autolog()` (#2088) are now automatically ended after training and/or exporting your model. See the docs for more details.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python


mlflow/tensorflow.py
```diff
@@ -657,26 +657,67 @@ def autolog(every_n_iter=100):
                       "1.12 <= v <= 2.0.0 are supported.")
         return
 
+    @gorilla.patch(tensorflow.estimator.Estimator)
+    def train(self, *args, **kwargs):
+        global _AUTO_END_RUN
+        if not mlflow.active_run():
+            mlflow.start_run()
```
smurching: As per comments on https://github.com/mlflow/mlflow/pull/2095, we should wrap this in a `try_mlflow_log` as well

mlflow/tensorflow.py
```diff
@@ -657,26 +657,67 @@ def autolog(every_n_iter=100):
...
+                try_mlflow_log(mlflow.log_param, 'max_steps', args[3])
+        if 'steps' in kwargs:
+            try_mlflow_log(mlflow.log_param, 'steps', kwargs['steps'])
+        if 'max_steps' in kwargs:
+            try_mlflow_log(mlflow.log_param, 'max_steps', kwargs['max_steps'])
+
+        result = original(self, *args, **kwargs)
+
+        if _AUTO_END_RUN:
+            mlflow.end_run()
```
smurching:
```suggestion
            try_mlflow_log(mlflow.end_run())
```

mlflow/tensorflow.py
```diff
@@ -657,26 +657,67 @@ def autolog(every_n_iter=100):
                       "1.12 <= v <= 2.0.0 are supported.")
         return
 
+    @gorilla.patch(tensorflow.estimator.Estimator)
+    def train(self, *args, **kwargs):
```
smurching: Let's also update the documentation of `autolog` to note that both `train` and `export_saved_model` will start/end runs, and that to have everything autologged under a single run the user should wrap their training logic in a `with mlflow.start_run()`?

mlflow/tensorflow.py
```diff
@@ -657,26 +657,67 @@ def autolog(every_n_iter=100):
...
+            _AUTO_END_RUN = True
         original = gorilla.get_original_attribute(tensorflow.estimator.Estimator,
                                                   'export_saved_model')
         serialized = original(self, *args, **kwargs)
         try_mlflow_log(log_model, tf_saved_model_dir=serialized.decode('utf-8'),
                        tf_meta_graph_tags=[tag_constants.SERVING],
                        tf_signature_def_key='predict',
                        artifact_path='model')
+        if _AUTO_END_RUN:
+            mlflow.end_run()
```
smurching:
```suggestion
            try_mlflow_log(mlflow.end_run())
```

mlflow/tensorflow.py
```diff
@@ -657,26 +657,67 @@ def autolog(every_n_iter=100):
...
+            _AUTO_END_RUN = True
         original = gorilla.get_original_attribute(tensorflow.estimator.Estimator,
                                                   'export_savedmodel')
         serialized = original(self, *args, **kwargs)
         try_mlflow_log(log_model, tf_saved_model_dir=serialized.decode('utf-8'),
                        tf_meta_graph_tags=[tag_constants.SERVING],
                        tf_signature_def_key='predict',
                        artifact_path='model')
+        if _AUTO_END_RUN:
+            mlflow.end_run()
```
smurching:
```suggestion
            try_mlflow_log(mlflow.end_run())
```

mlflow/tensorflow.py
smurching: We should also update https://github.com/mlflow/mlflow/blob/master/examples/tensorflow/tf1/train_predict.py and https://github.com/mlflow/mlflow/blob/master/examples/tensorflow/tf2/train_predict_2.py to follow this practice

tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -182,24 +182,26 @@ def input_fn(features, labels, training=True, batch_size=256):
...
 def test_tf_estimator_autolog_logs_metrics(tf_estimator_random_data_run):
     assert 'loss' in tf_estimator_random_data_run.data.metrics
     metrics = client.get_metric_history(tf_estimator_random_data_run.info.run_id, 'loss')
     assert all((x.step-1) % 100 == 0 for x in metrics)
 
 
 @pytest.mark.large
+@pytest.mark.parametrize('export', [True])
 def test_tf_estimator_autolog_model_can_load_from_artifact(tf_estimator_random_data_run):
     artifacts = client.list_artifacts(tf_estimator_random_data_run.info.run_id)
```
smurching: Hm how does this test work? As it stands, it only pass if we have `manual_run=True` (because otherwise the run we log metrics etc & the run we log the saved model under will be different?)

smurching (commented):
BTW, here's another way we could implement the auto-run-ending for TF in a way that could allow us to log to the same run even if the user does `export_saved_model`: we keep track of the last run ID created by autolog, i.e. we have a global `_AUTOLOG_RUN_ID` initialized to `None`, set it to the ID of the run started in `train` if there's no active run, and then resume that run in `export_saved_model` by calling `mlflow.start_run(run_id=_AUTOLOG_RUN_ID)` instead of just `mlflow.start_run()`. Could be worth doing that actually, I don't think the code change is too large, and then we'd continue to have the nice behavior of things "just working" if you add `mlflow.tensorflow.autolog`, without needing to recommend wrapping your code in a `mlflow.start_run`

tests/tensorflow_autolog/test_tensorflow2_autolog.py
juntai-zheng: It works because the fixture `tf_estimator_random_data_run` would return the most recent run, where in both cases of `manual_run=True` and `manual_run=False` would return the run where the model is logged.

mlflow/tensorflow.py
juntai-zheng: Sure, I'll do the doc updates in this PR and update the examples in a follow-up.

mlflow/tensorflow.py
```diff
@@ -55,6 +55,7 @@
 
 # For tracking if the run was started by autologging.
 _AUTO_END_RUN = False
```
smurching: Could we just remove this global & check whether `_AUTOLOG_RUN_ID` is `None` in our logic below?

tests/tensorflow_autolog/test_tensorflow_autolog.py
```diff
@@ -167,68 +167,101 @@ def test_tf_core_autolog_logs_scalars(tf_core_random_tensors):
...
-        classifier = tf.estimator.DNNClassifier(
-            feature_columns=my_feature_columns,
-            # Two hidden layers of 10 nodes each.
-            hidden_units=[30, 10],
-            # The model must choose between 3 classes.
-            n_classes=3,
-            model_dir=dir)
+@pytest.mark.large
+@pytest.mark.parametrize('export', [True, False])
+def test_tf_keras_autolog_persists_manually_created_run(export):
```
smurching: BTW, I'd use pytest's [`tmpdir` fixture](http://doc.pytest.org/en/latest/tmpdir.html) here, which handles automatically creating & deleting a tempdir for you

mlflow/tensorflow.py
juntai-zheng: Yup, now that you mention it I see now I can refactor it.

tests/tensorflow_autolog/test_tensorflow_autolog.py
juntai-zheng: Totally forgot about that feature, will put it in. Thanks for the catch!

docs/source/tracking.rst
```diff
@@ -217,20 +217,30 @@ log statements. See example usages with `Keras <https://github.com/mlflow/mlflow
...
+| TensorFlow Core  | All ``tf.summary.scalar`` calls                        | --                                                       | --            | --                                                                                                                                               |
++------------------+--------------------------------------------------------+----------------------------------------------------------+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------+
 
 Note that autologging for ``tf.keras`` is handled by :py:func:`mlflow.tensorflow.autolog`, not :py:func:`mlflow.keras.autolog`. 
 
+If no active run exists when `autolog()` captures data, MLflow will automatically create a run to log information to.
+Once training ends via calls to `tf.estimator.train()`, `tf.keras.fit()`, `tf.keras.fit_generator()`, `keras.fit()` or `keras.fit_generator()`,
+or once `tf.estimator` models are exported via `tf.estimator.export_saved_model()` or `tf.estimator.export_savedmodel()`, MLflow will automatically end that run.
```
smurching:
```suggestion
or once `tf.estimator` models are exported via `tf.estimator.export_saved_model()`, MLflow will automatically end that run.
```

(`tf.estimator.export_savedmodel` was deprecated in the last TF 1.x release, so I don't think we need to mention it here)

docs/source/tracking.rst
```diff
@@ -217,20 +217,30 @@ log statements. See example usages with `Keras <https://github.com/mlflow/mlflow
...
+
+**Note**: When a run is automatically created by `mlflow.tensorflow.autolog`, a call to `tf.estimator.train()` will end the current run but will not log the trained model to that run.
```
smurching: Maybe we can reword this more concisely as:

```suggestion
**Note**: When using the `tf.Estimator()` API, `mlflow.tensorflow.autolog` will record training metrics & params passed to `train`. Autologging also logs the models saved by `tf.estimator.export_saved_model()` to MLflow.
```

mlflow/tensorflow.py
```diff
@@ -584,8 +584,8 @@ def _log_event(event):
     """
     if not mlflow.active_run():
         try_mlflow_log(mlflow.start_run)
```
smurching: Should we pull this logic for checking if a run is active & creating one if not into a helper like `_maybe_create_active_run`? Would make it easier to standardize

mlflow/tensorflow.py
```diff
@@ -657,34 +657,82 @@ def autolog(every_n_iter=100):
...
+            if len(args) >= 4:
+                try_mlflow_log(mlflow.log_param, 'max_steps', args[3])
+        if 'steps' in kwargs:
+            try_mlflow_log(mlflow.log_param, 'steps', kwargs['steps'])
+        if 'max_steps' in kwargs:
+            try_mlflow_log(mlflow.log_param, 'max_steps', kwargs['max_steps'])
+
+        result = original(self, *args, **kwargs)
+
+        if mlflow.active_run().info.run_id == _AUTOLOG_RUN_ID:
```
smurching:
```suggestion
        if mlflow.active_run() is not None and mlflow.active_run().info.run_id == _AUTOLOG_RUN_ID
```

Just to be safe. Also, we could pull this into a helper as well (e.g. `_maybe_end_active_run`) to reduce duplication

mlflow/tensorflow.py
```diff
@@ -657,34 +657,82 @@ def autolog(every_n_iter=100):
...
+                auto_end = True
+
         original = gorilla.get_original_attribute(tensorflow.estimator.Estimator,
                                                   'export_savedmodel')
         serialized = original(self, *args, **kwargs)
         try_mlflow_log(log_model, tf_saved_model_dir=serialized.decode('utf-8'),
                        tf_meta_graph_tags=[tag_constants.SERVING],
                        tf_signature_def_key='predict',
                        artifact_path='model')
+        if mlflow.active_run().info.run_id == _AUTOLOG_RUN_ID or auto_end:
```
smurching:
```suggestion
        if (mlflow.active_run() is not None and mlflow.active_run().info.run_id == _AUTOLOG_RUN_ID) or auto_end:
```

mlflow/tensorflow.py
smurching: Actually, you could wrap both the run-creation & run-ending logic in a `contextmanager`, e.g.

```
    @contextmanager
    def _manage_active_run():
        if not mlflow.active_run():
        yield mlflow.active_run()
        if mlflow.active_run() is not None and mlflow.active_run().info.run_id == _AUTOLOG_RUN_ID:
            try_mlflow_log(mlflow.end_run)        
```

and then use it via `with _manage_active_run(): ... # do stuff in with block`

https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager has more info on this pattern, we also use it in https://github.com/mlflow/mlflow/blob/690c7e933f4c1007f477fa2b240631a9f8f515cf/mlflow/store/db/utils.py#L68

juntai-zheng: Looks good to me, I just changed the function a little to the following:

```
    @contextmanager
    def _manage_active_run():
        if not mlflow.active_run():
            global _AUTOLOG_RUN_ID
            if mlflow.active_run() is not None: # defensive check in case `mlflow.start_run` fails
                _AUTOLOG_RUN_ID = mlflow.active_run().info.run_id
        yield mlflow.active_run()
        if mlflow.active_run() is not None and mlflow.active_run().info.run_id == _AUTOLOG_RUN_ID:
            try_mlflow_log(mlflow.end_run)  
```
This is so `_AUTOLOG_RUN_ID` wouldn't be set unless the run is created by the first `if` statement.

smurching (approved):
LGTM thanks for the updates @juntai-zheng!

docs/source/tracking.rst
```diff
@@ -217,20 +217,30 @@ log statements. See example usages with `Keras <https://github.com/mlflow/mlflow
...
++------------------+--------------------------------------------------------+----------------------------------------------------------+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------+
 
 Note that autologging for ``tf.keras`` is handled by :py:func:`mlflow.tensorflow.autolog`, not :py:func:`mlflow.keras.autolog`. 
 
+If no active run exists when `autolog()` captures data, MLflow will automatically create a run to log information to.
+Once training ends via calls to `tf.estimator.train()`, `tf.keras.fit()`, `tf.keras.fit_generator()`, `keras.fit()` or `keras.fit_generator()`,
+or once `tf.estimator` models are exported via `tf.estimator.export_saved_model()`, MLflow will automatically end that run.
+
+**Note**: When using the `tf.Estimator()` API, `mlflow.tensorflow.autolog` will record training metrics & params passed to `train`. Autologging also logs the models saved by `tf.estimator.export_saved_model()` to MLflow.
+To track a model in the same run, you will need to call `tf.estimator.export_saved_model()`, which will start the most recent `autolog`-created run
```
smurching:
```suggestion
```

docs/source/tracking.rst
```diff
@@ -217,20 +217,30 @@ log statements. See example usages with `Keras <https://github.com/mlflow/mlflow
...
+and log the model there. The run will be automatically ended afterwards.
```
smurching:
```suggestion
```

PR #1757, opened 2019-08-18
Added fit_generator hooks for autolog

charnger:
## What changes are proposed in this pull request?
 
When using keras fit_generator instead it fit, the callback hooks are currently not installed. This patch fixes it.
 
## How is this patch tested?
 
Only by running it on my data. I wasn't able to run the test_keras_autolog.py without errors on master with reasonable effort yet.
 
### Is this a user-facing change? 

- [ ] No. You can skip the rest of this section.
- [X] Yes. Give a description of this change to be included in the release notes for MLflow users.

Keras autolog hooks now work when using model.fit_generator, not just model.fit.
 
(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)
 
### What component(s) does this PR affect?
 
- [ ] UI
- [ ] CLI 
- [ ] API 
- [ ] REST-API 
- [ ] Examples 
- [ ] Docs
- [ ] Tracking
- [ ] Projects 
- [ ] Artifacts 
- [ ] Models 
- [ ] Scoring 
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### How should the PR be classified in the release notes? Choose one:
 
- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [X] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [ ] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #2095, opened 2019-11-14
Added fit_generator hooks for TensorFlow.Keras autolog

juntai-zheng:

As a mirror to #1757, this PR adds `fit_generator()` hooks for `tf.keras` autologging.

## How is this patch tested?

Unit tests.

TensorFlow autologging with `tf.keras` now supports `fit_generator()`, in addition to `fit()`.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python


smurching (approved):
LGTM!

mlflow/tensorflow.py
```diff
@@ -705,6 +705,36 @@ def fit(self, *args, **kwargs):
...
             mlflow.end_run()
         _AUTO_END_RUN = False
+
+        return result
+
+    @gorilla.patch(tensorflow.keras.Model)
+    def fit_generator(self, *args, **kwargs):
+        global _AUTO_END_RUN
+        if not mlflow.active_run():
+            mlflow.start_run()
```
smurching: Actually I realized - `mlflow.start_run` can make an API request, so we should wrap it in a `try_mlflow_log` it along with `mlflow.end_run`, and only attempt end the run if starting the run succeeds. We should also do the same for Keras autologging in a follow-up, sorry I missed this the first time around!

smurching: Hm thinking about it, there are a few things that could happen:

1) `start_run()` fails due to server unavailability but at least some successive logging calls succeed (could happen e.g. if the server is being updated during the `start_run`)
2) `start_run()` & all successive logging calls fail so autologging never starts a run, but `end_run` succeeds (pretty unlikely given that we typically make several calls between `start_run` and `end_run`)
3) `start_run()` succeeds, but `end_run` fails so we never terminate the run (could happen if the server becomes unavailable by the time the run finishes)

I think 1) & 3) could happen (they're hopefully unlikely) but 2) will almost never happen, so maybe we don't need to do:
>  only attempt to end the run (only set _AUTO_END_RUN = True) if starting the run succeeds

and can simply just wrap the `start_run` and `end_run` in a `try_mlflow_log`

smurching (approved):
LGTM!

PR #2097, opened 2019-11-14
Added optional prometheus exporter

t-henri:

A new parameter to the `mlflow server` cli allowing to activate a prometheus exporter which expose metrics on the /metrics endpoint.
The following metrics are exported for the search_run, create_run, log_param, log_metric and set_tag endpoints:
- Total number of calls
- Buckets with the response time of each call.

## How is this patch tested?

Locally, by trying to launch the server with or without the argument `--activate-prometheus`, and by checking that metrics are indeed available on /metrics endpoint.

A prometheus exporter exposing metrics on /metrics can be activated using --activate-parameter argument of the mlfllow server cli.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -234,8 +234,10 @@ def _validate_static_prefix(ctx, param, value):  # pylint: disable=unused-argume
               help="Additional command line options forwarded to gunicorn processes.")
 @click.option("--waitress-opts", default=None,
               help="Additional command line options for waitress-serve.")
+@click.option("--activate-prometheus", default=False, is_flag=True,
+              help="Activate prometheus exporter to expose metrics on /metrics endpoint.")
 def server(backend_store_uri, default_artifact_root, host, port,
-           workers, static_prefix, gunicorn_opts, waitress_opts):
+           workers, static_prefix, gunicorn_opts, waitress_opts, activate_prometheus):
     """
     Run the MLflow tracking server.
 
@@ -268,7 +270,7 @@ def server(backend_store_uri, default_artifact_root, host, port,
 
     try:
         _run_server(backend_store_uri, default_artifact_root, host, port,
-                    static_prefix, workers, gunicorn_opts, waitress_opts)
+                    static_prefix, workers, gunicorn_opts, waitress_opts, activate_prometheus)
     except ShellCommandException:
         eprint("Running the mlflow server failed. Please see the logs above for details.")
         sys.exit(1)
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -234,10 +234,12 @@ def _validate_static_prefix(ctx, param, value):  # pylint: disable=unused-argume
               help="Additional command line options forwarded to gunicorn processes.")
 @click.option("--waitress-opts", default=None,
               help="Additional command line options for waitress-serve.")
-@click.option("--activate-prometheus", default=False, is_flag=True,
-              help="Activate prometheus exporter to expose metrics on /metrics endpoint.")
+@click.option("--expose-prometheus", default=False, is_flag=True,
+              help="Activate prometheus exporter to expose metrics on /metrics endpoint."
+                   "You need to set the environment variable 'prometheus_multiproc_dir' "
+                   "to point to a directory where the metrics will be stored.")
 def server(backend_store_uri, default_artifact_root, host, port,
-           workers, static_prefix, gunicorn_opts, waitress_opts, activate_prometheus):
+           workers, static_prefix, gunicorn_opts, waitress_opts, expose_prometheus):
     """
     Run the MLflow tracking server.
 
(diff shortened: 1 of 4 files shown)


mlflow/cli.py
```diff
@@ -234,8 +234,10 @@ def _validate_static_prefix(ctx, param, value):  # pylint: disable=unused-argume
               help="Additional command line options forwarded to gunicorn processes.")
 @click.option("--waitress-opts", default=None,
               help="Additional command line options for waitress-serve.")
+@click.option("--activate-prometheus", default=False, is_flag=True,
+              help="Activate prometheus exporter to expose metrics on /metrics endpoint.")
```
mateiz: Might be better to call this "enable-prometheus" or "expose-prometheus" since it's a more common way to name flags. We don't really run Prometheus ourselves.

t-henri: I agree, it is clearer with your recommandation.

mlflow/cli.py
```diff
@@ -234,8 +234,12 @@ def _validate_static_prefix(ctx, param, value):  # pylint: disable=unused-argume
               help="Additional command line options forwarded to gunicorn processes.")
 @click.option("--waitress-opts", default=None,
               help="Additional command line options for waitress-serve.")
+@click.option("--expose-prometheus", default=False, is_flag=True,
```
dbczumar: Is it possible to change this from a flag to a string argument that defines the target directory for prometheus metrics and sets the appropriate  environment variable via the `env` argument to `exec_cmd`? It would also be helpful if the command creates this directory in case it doesn't exist. These changes would make the feature easier to use.

mlflow/server/prometheus_exporter.py
```diff
@@ -0,0 +1,23 @@
...
+                                  labels={'status': lambda r: r.status_code,
+                                          'path': lambda: change_path_for_metric(request.path)})
+    for func_name, func in endpoint.items():
+        if func_name in ["_search_runs", "_log_metric", "_log_param", "_set_tag", "_create_run"]:
+            app.view_functions[func_name] = histogram(func)
+
+    return app
+
+
+def change_path_for_metric(path):
```
dbczumar: Can we add docs to this method explaining what it's doing?

dbczumar (changes requested):
@t-henri This is a neat feature! Left a couple comments. I tried it out by running `prometheus_multiproc_dir="/tmp/mydir" mlflow server --expose-prometheus` and unfortunately am not sure how to access the prometheus-exported metrics. I tried navigating to `http://<mlflow_server_url>/metrics` in the browser and didn't see anything. I also tried curling `http://<mlflow_server_url>/metrics` and didn't receive any metric-related output. It's worth noting that histogram files were created in `/tmp/mydir`. Can you explain what I'm doing wrong and/or update the `mlflow server` CLI documentation to explain how to access these metrics?

mlflow/server/prometheus_exporter.py
t-henri: Done

mlflow/cli.py
t-henri: Agreed, it is much easier to use this way. I made the modification.

t-henri:
Hello @dbczumar, I am not sure why you couldn't access the metrics, you seem to be doing it right. 

The only thing I can think of is that when you launch your server with the following command line:
`mlflow server --backend-store-uri _MLFLOW_SERVER_FILE_STORE --default-artifact-root _MLFLOW_SERVER_ARTIFACT_ROOT --host 0.0.0.0 --port 5000 --expose-prometheus /tmp/test`
When you go on your server, the url is `http://localhost:5000/#/` and to access the metrics you should go to `http://localhost:5000/metrics` (notice that we dropped the #).

mlflow/server/__init__.py
```diff
@@ -22,6 +23,11 @@
 for http_path, handler, methods in handlers.get_endpoints():
     app.add_url_rule(http_path, handler.__name__, handler, methods=methods)
 
+if os.getenv(PROMETHEUS_EXPORTER_ENV_VAR):
+    from mlflow.server.prometheus_exporter import activate_prometheus_exporter
+    os.makedirs(os.getenv(PROMETHEUS_EXPORTER_ENV_VAR), exist_ok=True)
```
dbczumar: @t-henri Looks like the `exist_ok` flag is unsupported in Python 2. Can we do something like this instead?:
```
prometheus_metrics_path = os.getenv(PROMETHEUS_EXPORTER_ENV_VAR)
if not os.path.exists(prometheus_metrics_path):
    os.makedirs(prometheus_metrics_path)
```

dbczumar (commented):
@t-henri I left one comment about compatibility for Python 2.7. After that's addressed, we should be good to go!

mlflow/server/__init__.py
t-henri: Ha yes sorry my bad I forgot the python 2 support, I just updated the PR.

dbczumar (approved):
LGTM! Thanks @t-henri!

PR #2119, opened 2019-11-20
Adding TensorFlow.Keras and Keras parameter logging

juntai-zheng:

`Keras` and `TensorFlow.Keras` autologging now logs parameters that are passed into `fit()`/`fit_generator()`.

Data, validation data, callbacks, verbosity, and generators are not logged.

Parameters not explicitly passed by users (parameters that use default values) in `keras.Model.fit_generator()` will not be logged due to `inspect.getargspec()` limitations.

Docs are updated to reflect the changes.

## How is this patch tested?

Added checks to existing tests checking for automatic logging.

`Keras` and `TensorFlow.Keras` autologging now logs parameters that are passed into `fit()`/`fit_generator()`.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [X] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python


mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
         fn(*args, **kwargs)
     except Exception as e:  # pylint: disable=broad-except
         warnings.warn("Logging to MLflow failed: " + str(e), stacklevel=2)
+
+
+def param_logger(fn, args=[], kwargs=None, unlogged=[]):
```
smurching:
```suggestion
def log_fn_args_as_params(fn, args=[], kwargs=None, unlogged=[]):
```

Just for clarify of name

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+    """
+    Log parameters explicitly passed to a function.
+    :param fn: function whose parameters are to be logged
+    :param args: arguments explicitly passed into fn
+    :param kwargs: kwargs explicitly passed into fn
+    :param unlogged: parameters not to be logged
+    :return: None
+    """
+    # Names of all parameters for the function
+    all_param_names = inspect.getargspec(fn)[0]
```
smurching: We could unpack this btw as `all_param_names, _, _, all_default_values = inspect.getargspec(fn)`

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+
+    # Default values of all parameters with default values. Has length of n, and corresponds
+    # to values of last n elements in all_param_names
+    all_default_values = inspect.getargspec(fn)[3]
+
+    # Checking if default values are present for logging. Known bug that getargspec will return an
```
smurching: Ah interesting is this actually a known bug / if so do we have a link to the Python bug report?

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+
+        default_params = zip(kwargs_and_default_names, all_default_values)
+
+        # Filtering out the parameters that have been passed in by the user as a kwarg.
+        default_params_to_be_logged = []
+        for param in default_params:
+            if param[0] not in kwargs:
+                default_params_to_be_logged += [param]
+
+        for param in default_params_to_be_logged:
```
smurching:
```suggestion
        for param_key, param_val in default_params_to_be_logged:
            if param_key not in unlogged:
                try_mlflow_log(mlflow.log_param, param_key, param_val)        
```

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+    # Names of all parameters for the function
+    all_param_names = inspect.getargspec(fn)[0]
+
+    # Default values of all parameters with default values. Has length of n, and corresponds
+    # to values of last n elements in all_param_names
+    all_default_values = inspect.getargspec(fn)[3]
+
+    # Checking if default values are present for logging. Known bug that getargspec will return an
+    # empty argspec for certain functions, despite the functions having an argspec.
+    if all_default_values:
```
smurching: Maybe we should check `len(all_default_values) > 0` for clarity? Although it's interesting that the truth value of an empty list is False

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+        # Removing the first len(args) elements from all_param_names - these are values already
```
smurching:
```suggestion
        # Removing the first len(args) elements from all_param_names - these are names of arguments
        # passed as positional arguments by the user and don't need to be logged with default values.
```

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+        # passed in explicitly by the user and don't need to be logged with default values.
```
smurching:
```suggestion
```

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+        kwargs_and_default_names = all_param_names[len(args):]
+
+        # If there are more parameter names than default values left, we know that the parameters
+        # not covered by the default values are passed in as kwargs (assuming all non-default
```
smurching:
```suggestion
        # not covered by the default values were passed by the user as kwargs (assuming all non-default
```

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+        # parameters are passed to the function)
+        if len(kwargs_and_default_names) > len(all_default_values):
+            kwargs_and_default_names = kwargs_and_default_names[len(kwargs_and_default_names)
+                                                                - len(all_default_values):]
+        # Otherwise, if there are more default values than parameter names, we know that some of the
+        # parameters with default values have been entered by the user in args
+        elif len(kwargs_and_default_names) < len(all_default_values):
```
smurching: How do we know that we can simply exclude the last `kwargs_and_default_names` default values (i.e. that these are the values that correspond to user-specified params)?

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+            all_default_values = all_default_values[len(all_default_values)
+                                                    - len(kwargs_and_default_names):]
+
+        default_params = zip(kwargs_and_default_names, all_default_values)
+
+        # Filtering out the parameters that have been passed in by the user as a kwarg.
+        default_params_to_be_logged = []
```
smurching: You can probably implement this via set subtraction in Python instead of looping through the list, e.g. compute a set of default-param-names & subtract out `set(kwargs.keys())`, then look the remaining params up in `default-params`

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+        if len(kwargs_and_default_names) > len(all_default_values):
+            kwargs_and_default_names = kwargs_and_default_names[len(kwargs_and_default_names)
+                                                                - len(all_default_values):]
+        # Otherwise, if there are more default values than parameter names, we know that some of the
+        # parameters with default values have been entered by the user in args
+        elif len(kwargs_and_default_names) < len(all_default_values):
+            all_default_values = all_default_values[len(all_default_values)
+                                                    - len(kwargs_and_default_names):]
+
+        default_params = zip(kwargs_and_default_names, all_default_values)
```
smurching: Could make this a dict to simplify lookups 
```suggestion
        default_params = dict(zip(kwargs_and_default_names, all_default_values))
```

mlflow/utils/autologging_utils.py
```diff
@@ -9,3 +11,62 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+        for param in default_params:
+            if param[0] not in kwargs:
+                default_params_to_be_logged += [param]
+
+        for param in default_params_to_be_logged:
+            if param[0] not in unlogged:
+                try_mlflow_log(mlflow.log_param, param[0], param[1])
+
+    # List of tuples of parameter names and args that are passed by the user
+    params_list = zip(inspect.getargspec(fn)[0][:len(args)], args)
```
smurching:
```suggestion
    params_dict = dict([(key, value) for key, value in zip(all_param_names[:len(args)], args)) if key not in unlogged])
    try_mlflow_log(mlflow.log_params, params_dict)
    
```

smurching (commented):
Looks pretty good! Mainly had comments on the new argument-capturing utility - could we add some tests for it under `tests/utils/test_autologging_utils.py`? Maybe we can make the `param_logger` function actually just return a dictionary of the params to log (i.e. a dictionary from  default_param_name to default_param_value) & rename it to something like `get_unspecified_default_args` (& then test that the utility returns the appropriate default args for a couple of dummy functions)

mlflow/utils/autologging_utils.py
juntai-zheng: In these cases we are excluding the first elements of the longer list, not the last. Also, the length of the elements excluded is the difference between `all_default_values` and `kwargs_and_default_values`. In the `elif` case (which I'm assuming you're talking about), we can exclude the first default values since parameters with default arguments must come after parameters without default arguments. At this point of the code, we've already removed the parameter names which have been covered by arguments explicitly passed in by the user in `kwargs_and_default_values`, and if we don't have a 1:1 match of parameter names to values then the first `len(all_default_values)-len(kwargs_and_default_values)` elements of `all_default_values` are known to be covered by user-passed params.

juntai-zheng:
Currently, `param_logger` takes care of logging user-passed `args`, `kwargs`, and default parameters - this leads to the least amount of code duplication between the different autologging modules and I think is clean for those specific purposes. What do you think about just making the logic for making a default argument dictionary a different util function (like `get_unspecified_default_args`), and having `param_logger` call that function? It would take a little more work in terms of writing tests, but would make the functions more convenient as utilities.

mlflow/utils/autologging_utils.py
juntai-zheng: Actually, there might be a way around this. `keras` has its own `getargspec()` [seen here](https://github.com/keras-team/keras/blob/master/keras/utils/generic_utils.py#L246), so I'll try and see if we can get around not logging `fit_generator()` params.

juntai-zheng: After testing, even the `keras` specific `getargspec()` returns the same output as `inspect.getargspec()`. Turns out it wasn't a bug that was logged anywhere, but this function is deprecated in Python 3 but the only solution for backwards compatibility with Python 2.

tests/utils/test_autologging_utils.py
```diff
@@ -0,0 +1,107 @@
...
+                         ['arg1', 'default1', 'default2'], [1, 2], {'default2': 2}),
+                         (['arg1', 'default1'], {},
+                         ['arg1', 'default1', 'default2'], [1, 2], {'default2': 2}),
+                         (['arg1'], {},
+                         ['arg1', 'default1', 'default2'], [1, 2], {'default1': 1, 'default2': 2}),
+                         ([], {'arg1': 42},
+                         ['arg1', 'default1', 'default2'], [1, 2], {'default1': 1, 'default2': 2})]
+
+
+@pytest.mark.large
```
smurching: We could probably keep these tests small right? Are they slow to run?

juntai-zheng: We could reduce them, but all the tests together only take about 1 second to run entirely.

mlflow/utils/autologging_utils.py
```diff
@@ -13,7 +13,41 @@ def try_mlflow_log(fn, *args, **kwargs):
...
+    :param user_args: list of arguments passed in by the user
+    :param user_kwargs: dictionary of kwargs passed in by the user
+    :param all_param_names: names of all of the parameters of the function
+    :param all_default_values: values of all default parameters
+    :return: a dictionary mapping arguments not specified by the user -> default value
+    """
+    num_default_args = len(all_default_values)
+
+    # all_default_values correspond to the last len(all_default_values) elements of the arguments
+    default_param_names = all_param_names[-num_default_args:]
```
smurching: BTW, this logic seems to assume that `num_default_args > 0` (if it's 0 then `all_param_names[0:]` returns the whole list), probably better to compute `num_args_without_default_value` earlier & write this as `all_param_names[num_args_without_default_value:]`

tests/utils/test_autologging_utils.py
```diff
@@ -0,0 +1,107 @@
...
+                                            all_default_values, expected):
+
+    default_dict = get_unspecified_default_args(user_args, user_kwargs,
+                                                all_param_names, all_default_values)
+
+    assert default_dict == expected
+
+
+@pytest.fixture
+def start_run():
```
smurching: We should use the `tracking_uri_mock` fixture as an argument to `start_run` to make sure we don't actually log to the local filesystem

smurching (commented):
Small comments, LGTM once they're addressed

mlflow/utils/autologging_utils.py
juntai-zheng: Ah good catch, will fix

smurching (approved):
LGTM thanks for the updates!

PR #2120, opened 2019-11-20
Fixing TensorFlow estimator autolog test names

juntai-zheng:

A few tests for autologging with `tf.estimator` were incorrectly named, causing some of the tests in the suite not to be run since the names were the same as other tests in the suite.

Also renamed some `keras` tests to make them more specific.

## How is this patch tested?

Manually

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/tests/keras_autolog/test_keras_autolog.py
+++ b/tests/keras_autolog/test_keras_autolog.py
@@ -46,7 +46,7 @@ def create_model():
 
 @pytest.mark.large
 @pytest.mark.parametrize('fit_variant', ['fit', 'fit_generator'])
-def test_autolog_ends_auto_created_run(random_train_data, random_one_hot_labels, fit_variant):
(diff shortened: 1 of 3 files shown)


smurching (approved):
LGTM. For the future, is there a way we could get pytest to warn us / fail when this happens?

juntai-zheng:
> LGTM. For the future, is there a way we could get pytest to warn us / fail when this happens?

The most related thing I could find was [this](https://stackoverflow.com/questions/46387285/pytest-3-2-2-duplicate-test), which says that we can't run multiple tests with the same name. Couldn't find anything about detecting multiple tests with the same name. I guess we just have to be more conscious about it in the future.

PR #2124, opened 2019-11-20
Add support for XGBoost flavor

harupy:

Add support for XGBoost flavor 

## How is this patch tested?

Unit tests

Model API supports XGBoost.

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/xgboost.py
+++ b/mlflow/xgboost.py
@@ -45,7 +45,7 @@ def save_model(xgb_model, path, conda_env=None, mlflow_model=Model()):
     :param xgb_model: XGBoost model to be saved.
     :param path: Local path where the model is to be saved.
     :param conda_env: Either a dictionary representation of a Conda environment or the path to a
-                      Conda environment yaml file. If provided, this decribes the environment
+                      Conda environment yaml file. If provided, this describes the environment
                       this model should be run in. At minimum, it should specify the dependencies
                       contained in :func:`get_default_conda_env()`. If ``None``, the default
                       :func:`get_default_conda_env()` environment is added to the model.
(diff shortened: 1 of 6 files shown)


mlflow/xgboost.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    :return: The default Conda environment for MLflow Models produced by calls to
+             :func:`save_model()` and :func:`log_model()`.
+    """
+    import xgboost as xgb
+
+    return _mlflow_conda_env(
+        additional_conda_deps=None,
+        additional_pip_deps=[
+            "xgboost=={}".format(xgb.__version__),
```
smurching: Ah interesting, surprising that xgboost isn't available on the default conda channels :P (verified this by running `conda install xgboost` on my Mac, which failed)

mlflow/xgboost.py
```diff
@@ -0,0 +1,174 @@
...
+                            ]
+                        }
+
+    :param mlflow_model: :py:mod:`mlflow.models.Model` this flavor is being added to.
+    """
+    import xgboost as xgb
+
+    path = os.path.abspath(path)
+    if os.path.exists(path):
+        raise Exception("Path '{}' already exists".format(path))
```
smurching:
```suggestion
        raise MlflowException("Path '{}' already exists".format(path), DIRECTORY_NOT_EMPTY)
```

The suggestion is to raise the more specific MlflowException so users can catch that if they want (note that they can still catch a broader `Exception`)

smurching (commented):
The implementation looks pretty good (barring small comments) :)

Would you mind adding tests for `log_model` and `save_model`? You can see  `tests/keras/test_keras_model_export.py` for an example to work from

mlflow/xgboost.py
harupy: Did you try `conda install -c conda-forge xgboost`?
I always use this command to install xgboost on conda.

https://anaconda.org/conda-forge/xgboost

mlflow/xgboost.py
harupy: will fix this

harupy:
@smurching Thanks for the review! I'll add tests based on keras flavor.

smurching (approved):
LGTM thanks for the updates @harupy!

harupy:
@smurching Thanks for merging! I'll start working on LightGBM flavor.

PR #2136, opened 2019-11-24
Add support for LightGBM flavor

harupy:

Add support for lightgbm flavor.

## How is this patch tested?

Unit tests

Model API now supports LightGBM.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -173,4 +173,4 @@ class _LGBModelWrapper:
 
     def predict(self, dataframe):
         import lightgbm as lgb
-        return self.lgb_model.predict(lgb.Dataset(dataframe))
+        return self.lgb_model.predict(dataframe)
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -172,5 +172,4 @@ class _LGBModelWrapper:
         self.lgb_model = lgb_model
 
     def predict(self, dataframe):
-        import lightgbm as lgb
         return self.lgb_model.predict(dataframe)
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -44,7 +44,7 @@ def save_model(lgb_model, path, conda_env=None, mlflow_model=Model()):
     """
     Save a LightGBM model to a path on the local file system.
 
-    :param lgb_model: LightGBM model to be saved.
+    :param lgb_model: LightGBM model (an instance of ``lightgbm.Booster``) to be saved.
     :param path: Local path where the model is to be saved.
     :param conda_env: Either a dictionary representation of a Conda environment or the path to a
                       Conda environment yaml file. If provided, this describes the environment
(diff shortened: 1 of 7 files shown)


harupy:
The implementation is almost the same as XGBoost flavor.

harupy:
ping @smurching

mlflow/lightgbm.py
```diff
@@ -0,0 +1,175 @@
...
+            "lightgbm=={}".format(lgb.__version__),
+        ],
+        additional_conda_channels=None)
+
+
+def save_model(lgb_model, path, conda_env=None, mlflow_model=Model()):
+    """
+    Save a LightGBM model to a path on the local file system.
+
+    :param lgb_model: LightGBM model to be saved.
```
smurching:
```suggestion
    :param lgb_model: LightGBM model (an instance of lightgbm.Booster) to be saved.
```

Assuming my understanding is correct that the `save_model`, `log_model`, `load_model` APIs only work for `lightgbm.Booster` models, and not models that implement the [scikit-learn API](https://lightgbm.readthedocs.io/en/latest/Python-API.html#scikit-learn-api) - it'd also be good to add this clarification to the docstrings of `log_model` and `load_model`

smurching (commented):
One small comment but otherwise looks good, thanks @harupy!

mlflow/lightgbm.py
harupy: @smurching Thanks for the comment. I'll add this clarification to XGBoost flavor as well.

mlflow/lightgbm.py
```diff
@@ -158,8 +158,7 @@ def load_model(model_uri):
                       `Referencing Artifacts <https://www.mlflow.org/docs/latest/tracking.html#
                       artifact-locations>`_.
 
-    :return: An `lightgbm.Booster model object
```
smurching: Ah, I think the old doc here was actually fine, sorry for the confusion - seems nice to link to the lightgbm docs, but this is also fine as is.

smurching (approved):
LGTM, looks like one of the TensorFlow tests is being flaky so retriggered it :) thanks @harupy!

harupy:
@smurching Thanks!

mlflow/lightgbm.py
harupy: @smurching Do you think we should warn users that LightGBM and XGBoost flavor don't yet support models that implement the scikit-learn API?

smurching: Yeah I think that'd definitely be helpful, I'd be happy to review a docstring update that adds such a warning, for example:

```suggestion
    :param lgb_model: LightGBM model (instance of `lightgbm.Booster <https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.Booster.html#lightgbm.Booster>`_) to be saved. Note that saving LightGBM models that implement the `scikit-learn API <https://lightgbm.readthedocs.io/en/latest/Python-API.html#scikit-learn-api>`_  is not supported.
```

harupy: @smurching Thanks. I'll open a PR to add this warning.

harupy: @smurching I opened a PR to update docstring https://github.com/mlflow/mlflow/pull/2170

PR #2137, opened 2019-11-25
Strip "dev" suffix from PySpark versions

dbczumar:

Dev versions of PySpark are used by default on certain platforms (e.g., Databricks). Dev versions of PySpark are not available on Anaconda; as a result, Conda environments generated for SparkML models saved with MLflow on Databricks contain an uninstallable PySpark version dependency. When a model with such an invalid dependency is served, the serving process fails.

## How is this patch tested?

Unit tests

## Release Notes

"dev" versions of PySpark are now replaced with stable versions when saving SparkML models with MLflow.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [X] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/mlflow/spark.py
+++ b/mlflow/spark.py
@@ -51,7 +51,12 @@ _logger = logging.getLogger(__name__)
 def get_default_conda_env():
     """
     :return: The default Conda environment for MLflow Models produced by calls to
-             :func:`save_model()` and :func:`log_model()`.
+             :func:`save_model()` and :func:`log_model()`. This Conda environment
+             contains the current version of PySpark that is installed on the caller's
+             system. ``dev`` versions of PySpark are replaced with stable versions in
+             the resulting Conda environment (e.g., if you are running PySpark version
+             ``2.4.5.dev0``, invoking this method produces a Conda environment with a
+             dependency on PySpark version ``2.4.5``).
     """
     import pyspark
     # Strip the suffix from `dev` versions of PySpark, which are not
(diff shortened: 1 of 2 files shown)


mlflow/spark.py
```diff
@@ -50,13 +51,21 @@
...
+             contains the current version of PySpark that is installed on the caller's
+             system. ``dev`` versions of PySpark are replaced with stable versions in
+             the resulting Conda environment (e.g., if you are running PySpark version
+             ``2.4.5.dev0``, invoking this method produces a Conda environment with a
+             dependency on PySpark version ``2.4.5``).
     """
     import pyspark
+    # Strip the suffix from `dev` versions of PySpark, which are not
+    # available for installation from Anaconda or PyPI
+    pyspark_version = re.sub(r"dev.*", "", pyspark.__version__).rstrip(".")
```
mateiz: Why not just sub the expression `"\.dev.*" with the empty string? It seems weird to both sub and rstrip.

dbczumar: I'd like to handle the case where `dev` isn't prefixed with a dot, which seems like something that could reasonably occur. I'm more than happy to move this into the regular expression.

mateiz:
LGTM!

PR #2149, opened 2019-11-27
Make kube-context config optional (default in-cluster)

zanitete:

Make kube-context optional and default to the in-cluster configuration. This allows to run MLflow from within a Kubernetes cluster leveraging the role of the service account running it.

## How is this patch tested?

Unit tested and manually tested by running MLflow in a pod.

MLflow projects can be scheduled from within a Kubernetes cluster.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [x] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### DIFF
--- a/docs/source/projects.rst
+++ b/docs/source/projects.rst
@@ -474,7 +474,7 @@ You can run your MLflow Project on Kubernetes by following these steps:
    - ``kube-context``
      The `Kubernetes context
      <https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#context>`_
-     where MLflow will run the job.
+     where MLflow will run the job. If not provided, MLflow will assume it is running in a Kubernetes cluster.
    - ``repository-uri``
      The URI of the docker repository where the Project execution Docker image will be uploaded
      (pushed). Your Kubernetes cluster must have access to this repository in order to run your
--- a/docs/source/projects.rst
+++ b/docs/source/projects.rst
@@ -474,7 +474,9 @@ You can run your MLflow Project on Kubernetes by following these steps:
    - ``kube-context``
      The `Kubernetes context
      <https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#context>`_
-     where MLflow will run the job. If not provided, MLflow will assume it is running in a Kubernetes cluster.
+     where MLflow will run the job. If not provided, MLflow will use the current context.
+     If no context is available, MLFlow will assume it is running in a Kubernetes cluster
+     and it will use the Kubernetes service account running the current pod ('in-cluster' configuration).
    - ``repository-uri``
      The URI of the docker repository where the Project execution Docker image will be uploaded
      (pushed). Your Kubernetes cluster must have access to this repository in order to run your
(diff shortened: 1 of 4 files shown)


bgroenks96:
Was literally  just about to make this change myself! Thanks!

Could we get this merged soon? It's a simple but very helpful (and necessary) change!

docs/source/projects.rst
```diff
@@ -446,7 +446,7 @@ You can run your MLflow Project on Kubernetes by following these steps:
    - ``kube-context``
      The `Kubernetes context
      <https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#context>`_
-     where MLflow will run the job.
+     where MLflow will run the job. If not provided, MLflow will assume it is running in a Kubernetes cluster.
```
dbczumar: In the case where `kube-context` is unspecified, can you clarify that the in-cluster configuration will be selected and that this will use the Kubernetes service account configured in the current pod?

mlflow/projects/__init__.py
```diff
@@ -724,7 +724,7 @@ def _parse_kubernetes_config(backend_config):
         raise ExecutionException("Could not find 'kube-job-template-path': {}".format(
             kube_job_template))
     if 'kube-context' not in backend_config.keys():
-        raise ExecutionException("Could not find kube-context in backend_config.")
+        _logger.info("Could not find kube-context in backend_config. Using in-cluster config.")
```
dbczumar: Nit: Can we make this a warning instead?

dbczumar (changes requested):
@zanitete Thanks for the contribution - this is super useful. I left a couple nit comments; this should be good to merge after they're addressed!

mlflow/projects/__init__.py
ReDucTor: For the case where this is inside a pod already having this as a warning could be the wrong signal, as I would expect after this change it would be expected behavior.

imho, even outside of kubernetes itself I would expect it to pick the current context when not provided, as someone who switches contexts to switch clusters having to also modify some file is less then ideal

dbczumar: @ReDucTor That's a great point. @zanitete is it possible to use the current context if none is specified? Will this also work for in-cluster contexts?

zanitete: The in-cluster configuration does not rely on a config file, so there's no "current context" in that scenario.

We could try first to use the current context (that covers the cases when `KUBECONFIG` is set or `~/.kube/config` exists) and fallback to the in-cluster configuration as last option. Would do you think?

dbczumar: @zanitete That sounds great!

mlflow/projects/__init__.py
```diff
@@ -724,7 +724,8 @@ def _parse_kubernetes_config(backend_config):
         raise ExecutionException("Could not find 'kube-job-template-path': {}".format(
             kube_job_template))
     if 'kube-context' not in backend_config.keys():
-        raise ExecutionException("Could not find kube-context in backend_config.")
+        _logger.debug("Could not find kube-context in backend_config."
```
zanitete: @dbczumar @ReDucTor, since the configuration is valid even without 'kube-context', I made it a debug log instead

dbczumar (approved):
LGTM! Thanks @zanitete!

PR #2161, opened 2019-12-04
Travis Nightly Build

juntai-zheng:

Configures `travis.yml` to accept nightly cron jobs. Currently, the tests run are the end-to-end `examples` suite. This should reduce the bulkiness of tests run when other PRs are being tested.

Nightly tests will trigger with regular builds if Travis detects files related to the nightly tests are changed. This change is reflected in `CONTRIBUTING.rst`.

## How is this patch tested?

Through Travis.
This build shows that the nightly tests are triggered when affected files are changed: https://travis-ci.org/mlflow/mlflow/builds/620914892

This build shows that the nightly tests are triggered when the job is a CRON job: Will be tested after this is merged and cron jobs are enabled.

This build shows that nightly tests are not triggered when the job is a PR and none of the nightly build files are changed: https://travis-ci.org/mlflow/mlflow/builds/621259245

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,7 @@ matrix:
         - npm run lint
         - npm test -- --coverage
     - stage: large
+      if: type != cron
     - language: python
       python: 2.7
       if: type != cron
@@ -107,6 +108,8 @@ matrix:
       script:
         - cd docs
         - make rsthtml SPHINXOPTS="-W --keep-going" # Interpret Sphinx warnings as errors via the `-W` flag
+    # Nightly builds run the tests/examples end-to-end test suite.
+    # Does not include Windows since Projects/Models are mostly not compatible with Windows as of now.
     - stage: nightly
     - language: python
       python: 2.7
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -7,9 +7,27 @@ services:
...
   include:
+    # Nightly builds run the tests/examples end-to-end test suite.
+    # Does not include Windows since Projects/Models are mostly not compatible with Windows as of now.
+    - stage: Nightly
+    - language: python
+      python: 2.7
+      install:
+        # CHANGED_FILES detects whether or not files related to the nightly build are changed.
+        # True is so that the command returns a 0 exit code whether or not CHANGED_FILES is empty.
+        - CHANGED_FILES=$(git diff --name-only master..HEAD | grep "tests/examples\|examples") || true;
```
smurching: Question, was there a reason not to use `TRAVIS_COMMIT_RANGE`? Sometimes we merge PRs into branches other than master (e.g. release branches like `branch-1.4`), so it might be safest to not assume that we should diff `master` & `HEAD` & use `TRAVIS_COMMIT_RANGE` instead.

smurching (commented):
Only one comment really, otherwise LGTM!

.travis.yml
juntai-zheng: Makes sense, though will have to use `${TRAVIS_COMMIT_RANGE/.../..}` instead according to [this Travis issue](https://github.com/travis-ci/travis-ci/issues/4596).

juntai-zheng: > There's also this issue posted by someone in the issue:
> Note: there's a further underlying problem here that crops up for long-lived feature branches not rebased onto master. Travis uses a shallow clone with a default depth of 50. If the shallow clone doesn't include the originating commit where the feature branch diverged, then `$TRAVIS_COMMIT_RANGE` will always include the full repository ancestry of the feature branch going back to the first commit. Both double dot and triple dot commit ranges are predicated on having the full repository ancestry.

Do you think this would be an issue regarding our repository?

juntai-zheng: After offline discussion, we decided that using `master..HEAD` was ok for now since PRs for release branches are rare and it wouldn't hurt to run nightly tests for them.

PR #2176, opened 2019-12-06
Add host option to `mlflow ui` CLI command

### LINKED ISSUE #1481: Can not connect to remote machine in local network
### System information
- **OS Platform and Distribution `Linux Ubuntu 16.04.5`**:
- **MLflow installed from `binary`**: 
- **MLflow version `1.0.0`**:
- **Python version `3.6.6`**: 
`mlflow server --host 0.0.0.0`
`mlflow server --host <local_ip_addr>`

### Describe the problem
I am trying to connect remote machine on my local network. But it seems the  `mlflow server` did not listens all network interfaces. I tried to start server with local ip address of machine but it did not work. I am sure I can access to the machine because I have different services running on it. The problem maybe caused by gunicorn.

gandroz:
Add host url as an option to ui command

### DIFF
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -165,7 +165,9 @@ def _validate_server_args(gunicorn_opts=None, workers=None, waitress_opts=None):
                    "Default: " + DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH)
 @click.option("--port", "-p", default=5000,
               help="The port to listen on (default: 5000).")
-def ui(backend_store_uri, default_artifact_root, port):
+@click.option("--host", "-h", default="127.0.0.1",
+              help="The host URL")
+def ui(backend_store_uri, default_artifact_root, port, host):
     """
     Launch the MLflow tracking UI for local viewing of run results. To launch a production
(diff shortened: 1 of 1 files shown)


mlflow/cli.py
```diff
@@ -165,7 +165,9 @@ def _validate_server_args(gunicorn_opts=None, workers=None, waitress_opts=None):
                    "Default: " + DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH)
 @click.option("--port", "-p", default=5000,
               help="The port to listen on (default: 5000).")
-def ui(backend_store_uri, default_artifact_root, port):
+@click.option("--host", "-h", default="127.0.0.1",
```
mparkhe: Rename `host` to `host-ip-address` to make it clear. Also fix the help to reflect this change.

gandroz: I let the parameter named `host` to be consistent with the `server` command. I also updaded the decorators to the generic `utils.cli_args` host and port ones.

mparkhe: ah yes! And thanks for using `@cli_args` decorators

PR #2181, opened 2019-12-09
Add Java implementation of Spark datasource autologging

smurching:

Adds Java implementation of Spark datasource autologging as described in [this design doc](https://docs.google.com/document/d/11nhwZtj-rps0stxuIioFBM9lkvIh_ua45cAFy_PqdHU/edit) under a new `mlflow-spark` maven package. The current implementation notifies subscribers with datasource path, version, and format information. A follow-up PR will add logic to register a subscriber from Python when `mlflow.spark.autolog()` is called.

At a high level we add the following classes under the `org.mlflow.spark.autologging` package (open to renaming them):
* `MlflowAutologEventPublisher` - provides APIs for:
    * Registering subscribers (e.g. Python callbacks) who will receive notifications on datasource reads
    * Initialization: creating a SparkListener that will watch for datasource reads as well as a GC thread to clean up unresponsive subscribers (e.g. if the Python process associated with a Python callback has died)
    * Stopping the SparkListener & GC thread
    * Notifying subscribers about a datasource read (this API is called by the SparkListener)
 * `MlflowAutologEventSubscriber`: subscriber interface for receiving info on datasource reads
 * `SparkDataSourceListener`: SparkListener that watches for datasource reads
 * `DatasourceAttributeExtractor`: helper object exposing methods for extracting Spark datasource attributes from a Spark logical plan.

## How is this patch tested?

Added tests under `SparkAutologgingIntegrationSuite.scala`. Also built Javadocs & verified we don't document any new public APIs.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [x] Java
- [ ] Python


mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -0,0 +1,130 @@
...
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  private[autologging] var sparkQueryListener: SparkDataSourceListener = _
+  private val ex = new ScheduledThreadPoolExecutor(1)
+  private[autologging] var subscribers: mutable.LinkedHashMap[String, MlflowAutologEventSubscriber] =
+    mutable.LinkedHashMap[String, MlflowAutologEventSubscriber]()
+  private var scheduledTask: ScheduledFuture[_] = _
+
+  def spark: SparkSession = {
+    SparkSession.builder.getOrCreate()
```
tomasatdatabricks: Should we call get or create here or just try to access the existing session and fail if it does not exist? In other places I have opted for the latter so that we do not change the state in an unexpected way.

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -0,0 +1,130 @@
...
+ * Trait implementing a publisher interface for publishing events on Spark datasource reads to
+ * a set of listeners. See the design doc:
+ * https://docs.google.com/document/d/11nhwZtj-rps0stxuIioFBM9lkvIh_ua45cAFy_PqdHU/edit for more
+ * details.
+ */
+private[autologging] trait MlflowAutologEventPublisherImpl {
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  private[autologging] var sparkQueryListener: SparkDataSourceListener = _
+  private val ex = new ScheduledThreadPoolExecutor(1)
```
tomasatdatabricks: Nit: rename ex -> executor?

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -0,0 +1,130 @@
...
+      }
+      subscribers = mutable.LinkedHashMap()
+    }
+  }
+
+  def register(subscriber: MlflowAutologEventSubscriber): String = synchronized {
+    if (sparkQueryListener == null) {
+      throw new RuntimeException("Please call init() before attempting to register a subscriber")
+    }
+    this.synchronized {
```
tomasatdatabricks: what is the second synchronized for?

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -0,0 +1,130 @@
...
+      val uuid = java.util.UUID.randomUUID().toString
+      subscribers.put(uuid, subscriber)
+      uuid
+    }
+  }
+
+  /** Unregister subscribers broken e.g. due to detaching of the associated Python REPL */
+  private[autologging] def unregisterBrokenSubscribers(): Unit = synchronized {
+    val brokenUuids = subscribers.flatMap { case (uuid, listener) =>
```
tomasatdatabricks: Do wee need to do the pings under the lock?

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private[autologging] def publishEvent(
+      replIdOpt: Option[String],
+      sparkTableInfo: SparkTableInfo): Unit = synchronized {
+    sparkTableInfo match {
+      case SparkTableInfo(path, version, format) =>
+        for ((uuid, listener) <- subscribers) {
+          try {
+            if (replIdOpt.isEmpty || listener.replId == replIdOpt.get) {
```
tomasatdatabricks: when is replIdOpt empty?

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  def register(subscriber: MlflowAutologEventSubscriber): String = synchronized {
+    if (sparkQueryListener == null) {
+      throw new RuntimeException("Please call init() before attempting to register a subscriber")
+    }
+    this.synchronized {
+      val uuid = java.util.UUID.randomUUID().toString
+      subscribers.put(uuid, subscriber)
```
tomasatdatabricks: why do we need uuid? Seems like it is not really used for anything other than addressing broken subcribers?

Can we a) use replId  or b) use sequential id instead?

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -0,0 +1,130 @@
...
+      sparkTableInfo: SparkTableInfo): Unit = synchronized {
+    sparkTableInfo match {
+      case SparkTableInfo(path, version, format) =>
+        for ((uuid, listener) <- subscribers) {
+          try {
+            if (replIdOpt.isEmpty || listener.replId == replIdOpt.get) {
+              listener.notify(path, version.getOrElse("unknown"), format.getOrElse("unknown"))
+            }
+          } catch {
+            case e: Py4JException =>
```
tomasatdatabricks: should we catch non fatal to be safe?

mlflow/java/spark/src/test/scala/org/mlflow/spark/autologging/SparkAutologgingIntegrationSuite.scala
```diff
@@ -0,0 +1,246 @@
...
+    MlflowAutologEventPublisher.register(subscriber)
+    leftDf.join(rightDf).collect()
+    // Sleep a second to let the SparkListener trigger read
+    Thread.sleep(1000)
+    verify(subscriber, times(2)).notify(any(), any(), any())
+    verify(subscriber, times(1)).notify(getFileUri(leftPath), "unknown", leftFormat)
+    verify(subscriber, times(1)).notify(getFileUri(rightPath), "unknown", rightFormat)
+  }
+
+  test("MlflowAutologEventPublisher publishes to subscribers in order of registration") {
```
tomasatdatabricks: is this a requirement?

tomasatdatabricks (commented):
Looks good! I letf some comments / questions.

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
smurching: oops, good call it's unnecessary

mlflow/java/spark/pom.xml
```diff
@@ -43,12 +43,6 @@
     </dependency>
 
     <!-- Test -->
-    <dependency>
```
smurching: Removing this as we don't test against Delta here, as Delta Lake doesn't yet work with Spark 3.0 (see https://stackoverflow.com/questions/59116065/spark-3-0-not-able-to-save-a-df-as-delta-table,)

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
smurching: Good idea, makes sense to use replId. Will do that & make sure it's documented that replId is expected to be unique

mlflow/java/spark/src/test/scala/org/mlflow/spark/autologging/SparkAutologgingIntegrationSuite.scala
smurching: It's not, just added it to help ensure correctness of the [next test](https://github.com/mlflow/mlflow/pull/2181/files#diff-2bc6bc37f0b21089d80596747c27a92cR192) ("MlflowAutologEventPublisher can publish to working subscribers even when others are broken"), which assumes that subscribers are notified in the order they're registered. Can do this more simply by exposing & calling a helper for iterating over subscribers in `MlflowAutologEventPublisher` & mocking it out in the `"MlflowAutologEventPublisher can publish to working subscribers even when others are broken"` test to yield a broken subscriber before a working one, or I could add a comment explaining the purpose of this test / that the behavior it guarantees isn't fundamental to the API (fine with doing either)

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
smurching: Good question it's actually currently always empty, as we don't know which REPL triggered the datasource-read event that is being published in `publishEvent`, and so we publish to all subscribers.

In REPL-ID aware environments (e.g. Databricks), we will be able to pull the REPL ID associated with the query in our SparkListener, so this logic was added in anticipation of that use case - will remove it for now & reintroduce it with a follow-up PR that adds a REPL-ID aware SparkListener

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
smurching: Hm I don't think we do, I'll confirm this w/ you tomorrow.

In particular we need to:
1. Get a snapshot iterator `it` over the elements of `subscribers` at some point in time
2. Iterate over `it` & ping each subscriber, get a list of bad repl IDs
3. Remove each bad repl ID from our map

Steps 2 & 3 can definitely be done outside of a lock (replIds are unique, so we won't ever accidentally remove a working REPL in step 3). Step 1 we can achieve by making a copy of `subscribers` (e.g. via `subscribers.toSeq`). The [Scala API docs](https://www.scala-lang.org/api/2.12.0/scala/collection/mutable/LinkedHashMap.html#toSeq:Seq[(K,V)]) don't guarantee that `toSeq` is threadsafe, so maybe we should synchronize this part at least.

smurching: That being said, we are assuming here that there's only one thread ever running `unregisterBrokenSubscribers`, so should at least add a comment stating that assumption when we create the GC thread/in this method docstring

tomasatdatabricks (approved):
lgtm!

PR #2188, opened 2019-12-10
Update contributor dependencies

andychow-db:

Updated contributor dependencies for MLflow. 

1. In `dev-requirements.txt`, bumped sphinx-click from 2.3.0 to 2.3.1. Otherwise, `pip install -r dev-requirements.txt` fails complete due to transient dependency error.
2. In `dev-requirements.txt`, added `gitpython` to ensure that the dependency matches the `setup.py` gitpython. Otherwise, pypi-publisher introduces an older, incompatible gitpython version.
3. Update `generate-protos.sh` to also use Protobuf 3.6.1 also. Otherwise, some users of Protobuf 3.6 cannot run the script `generate-protos`
4. Update `test-generate-protos.sh` with a success message to let contributors know that the test was successful. 

## How is this patch tested?

1. TravisCI ran the integration tests successfully. 

2. Validated protoc@3.6.0 and protoc@3.6.1 both output the same models by running both using `test-generate-protoc.sh`. Both versions passed the test. 

3. Deleted my conda environment and reran the instructions. 

## Release Notes

1. In `travis/large-requirements.txt`, it requires fbprohet but it does not work via pip install right now. To bypass this issue, I installed it via conda using the command `conda install -c conda-forge fbprophet`.

## Open questions (This to do before committing)

1. Are there additional tests to manually run to validate the change is well behaved?

## Closed questions (This to do before committing)
1. Is there a backwards incompatible difference in the Protobuf format between 3.6.0 and 3.6.1?
    I ran `test-generate-protoc.sh` with both protoc versions. Both editions successfully passed. Therefore, they output the same python and java files. 

- [X ] No. You can skip the rest of this section.

### DIFF
--- a/generate-protos.sh
+++ b/generate-protos.sh
@@ -3,7 +3,7 @@
 set -ex
 PROTOC_VERSION="$(protoc --version)"
 if [ "$PROTOC_VERSION" != "libprotoc 3.6.0" || "$PROTOC_VERSION" != "libprotoc 3.6.1" ]; then
-	echo "Must have libprotoc version 3.6.1."
+	echo "Required libprotoc versions to be 3.6.0 or 3.6.1 (preferred)."
 	echo "We found: $PROTOC_VERSION"
 	exit 1
 fi
--- a/generate-protos.sh
+++ b/generate-protos.sh
@@ -2,8 +2,8 @@
 
 set -ex
 PROTOC_VERSION="$(protoc --version)"
-if [ "$PROTOC_VERSION" != "libprotoc 3.6.0" ]; then
-	echo "Must have libprotoc version 3.6.0."
+if [ "$PROTOC_VERSION" != "libprotoc 3.6.0" || "$PROTOC_VERSION" != "libprotoc 3.6.1" ]; then
+	echo "Must have libprotoc version 3.6.1."
 	echo "We found: $PROTOC_VERSION"
 	exit 1
 fi
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,7 +2,7 @@
 sphinx==2.2.1
 sphinx_rtd_theme==0.4.3
 sphinx-autobuild==0.7.1
-sphinx-click==2.3.0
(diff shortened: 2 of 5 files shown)


generate-protos.sh
```diff
@@ -2,8 +2,8 @@
 
 set -ex
 PROTOC_VERSION="$(protoc --version)"
-if [ "$PROTOC_VERSION" != "libprotoc 3.6.0" ]; then
-	echo "Must have libprotoc version 3.6.0."
+if [ "$PROTOC_VERSION" != "libprotoc 3.6.0" || "$PROTOC_VERSION" != "libprotoc 3.6.1" ]; then
```
mparkhe: do we need this `||` condition or should we just check for `3.6.1`?
also do we need to pin `protoc` to `3.6.1` in `dev-requirements.txt` file?

andychow-db: If we remove the `||`, then we would require contributors to update from `3.6.0` to 3.6.1`. I'd prefer not to ask people to update to `3.6.1` because it's not necessary.

The python package does not have `protoc` and I had to install it separately using `brew install protobuf@3.6`. Secondly, in `setup.py`, we already specify `protobuf>=3.6.0` which means that new contributors will be running `3.11.1`.  Therefore, I don't think we should add it to `dev-requireents.txt`

Furthermore, looking at the protobuf [releases versions(https://github.com/protocolbuffers/protobuf/tags?after=v3.9.0), we only need to handle 3.6.1 and 3.6.0 right now. The latest protobuf version is 3.11.x and it does not past the test script.

mparkhe: I see. Thanks for clarifying. In that case, can rewrite the error message to something like--

```
Required libprotoc versions to be 3.6.0 or 3.6.1 (preferred).
```

Or something like that?

andychow-db: Of course, that makes a lot of sense.

mparkhe (approved):
Looks cool! Thanks

PR #2197, opened 2019-12-12
Add note about active run instantiation side effect in fluent APIs

andychow-db:
# What changes are proposed in this pull request?

In some APIs such as `get_artifact_uri`, the function starts an active
run if no there is no active run. This should be considered a
side-effect because these it is not apparent that these functions would
instantiate the active run.

As a short-term solution, I have updated the documentation in
functions that have this behavior.

## How is this patch tested?

Standard automated tests

## Release Notes

- Only updated documentation. There should be no code change as part of this PR.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [X ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

- [ X] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

### DIFF
--- a/mlflow/models/__init__.py
+++ b/mlflow/models/__init__.py
@@ -64,6 +64,8 @@ class Model(object):
         """
         Log model using supplied flavor module.
 
+        Note: If no run is active, it will instantiate a run to obtain a run_id.
+
         :param artifact_path: Run relative path identifying the model.
         :param flavor: Flavor module to save the model with. The module must have
                        the ``save_model`` function that will persist the model as a valid
--- a/mlflow/models/__init__.py
+++ b/mlflow/models/__init__.py
@@ -62,9 +62,7 @@ class Model(object):
     @classmethod
     def log(cls, artifact_path, flavor, registered_model_name=None, **kwargs):
         """
-        Log model using supplied flavor module.
-
-        Note: If no run is active, this method will create a new active run.
+        Log model using supplied flavor module. If no run is active, this method will create a new active run.
 
         :param artifact_path: Run relative path identifying the model.
         :param flavor: Flavor module to save the model with. The module must have
--- a/mlflow/models/__init__.py
+++ b/mlflow/models/__init__.py
@@ -64,7 +64,7 @@ class Model(object):
         """
         Log model using supplied flavor module.
 
-        Note: If no run is active, it will instantiate a run to obtain a run_id.
+        Note: If no run is active, this method will create a new active run.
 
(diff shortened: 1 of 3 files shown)


mlflow/models/__init__.py
```diff
@@ -64,6 +64,8 @@ def log(cls, artifact_path, flavor, registered_model_name=None, **kwargs):
         """
         Log model using supplied flavor module.
 
+        Note: If no run is active, it will instantiate a run to obtain a run_id.
```
smurching:
```suggestion
        Note: If no run is active, this method will create a new active run.
```

Just a small wording suggestion, lmk what you think (want to make it clear that we set the "active" run)

mlflow/tracking/fluent.py
```diff
@@ -206,6 +206,8 @@ def log_param(key, value):
     """
     Log a parameter under the current run, creating a run if necessary.
 
+    **Note:** If no run is active, it will instantiate a new run.
```
smurching: Thanks for adding these - I see the docs for most of these functions already mention that we create a run, maybe we could update that wording to reflect that we create an "active" run instead of repeating that a run is created?

e.g. for this method we could say:

> Log a parameter under the current active run, creating a run if necessary.

mlflow/tracking/fluent.py
```diff
@@ -371,6 +389,8 @@ def get_artifact_uri(artifact_path=None):
     run will be returned; calls to ``log_artifact`` and ``log_artifacts`` write
     artifact(s) to subdirectories of the artifact root URI.
 
+    **Note:** If no run is active, it will instantiate a new run.
```
smurching: +1 thanks for this fix. It looks like we never mentioned that this method was side-effecting at all, which is pretty scary :P, hopefully this helps reduce some of the surprise encountered when calling it.

smurching (commented):
Mostly looks good! Just small comments

mlflow/models/__init__.py
andychow-db: I agree, I think the suggestion is clearer.

mlflow/tracking/fluent.py
andychow-db: Done.

smurching (approved):
Thanks @andychow-db this LGTM!

PR #2200, opened 2019-12-13
Fix mlflow_list_artifacts() for the general case (R package)

lorenzwalthert:

Make `mlflow_list_artifacts()` work for the general case. More concretely, for cases where: 
- there is at least one directory in the artifacts directory that is not the first when sorting all names alphabetically and
- at least two other files in the artifact directory.

**Details**

As per the [documentation for the REST API to list artifacts](https://mlflow.org/docs/latest/rest-api.html#list-artifacts) and the specific [file info](https://mlflow.org/docs/latest/rest-api.html#mlflowfileinfo), the file size is unset for directories in the REST response, which is not accounted for in `mlflow_list_artifacts()` in the R API, specifically not in https://github.com/mlflow/mlflow/blob/master/mlflow/R/mlflow/R/tracking-experiments.R#L43-L46. For the reader's convenience, I c/p the code linked above:

```r
files_list %>%
    purrr::transpose() %>%
    purrr::map(unlist) %>%
    tibble::as_tibble()
```

As the file size is not set for directories, `as_tibble()` won't work in general because the column `file_size` will be shorter than the other columns if there was at least one directory and the above will fail like this:

```
Error: Tibble columns must have consistent lengths, only values of length one are recycled:
* Length 2: Column `file_size`
* Length 4: Columns `path`, `is_dir`
```

The problem has not occurred so far in the tests (tests/testthat/test/tracking-runs) because the first file (alphabetically) was a directory without file information and `purrr::transpose()` only uses the name of the first element that should be transposed, so the file_size was subsequentially omitted. Also, the problem does not occur if there is only one directory because as the above error message says: "values of length one are recycled".

## How is this patch tested?

Adapted unit tests and locally in project.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [x] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/R/mlflow/R/tracking-runs.R
+++ b/mlflow/R/mlflow/R/tracking-runs.R
@@ -355,6 +355,12 @@ mlflow_list_artifacts <- function(path = NULL, run_id = NULL, client = NULL) {
   message(glue::glue("Root URI: {uri}", uri = response$root_uri))
 
   files_list <- if (!is.null(response$files)) response$files else list()
+  files_list <- purrr::map(files_list, function(file_info) {
+    if (is.null(file_info$file_size)) {
+      file_info$file_size <- NA
+    }
+    file_info
+  })
   files_list %>%
     purrr::transpose() %>%
     purrr::map(unlist) %>%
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
Can't pass the R built until travis is [fixed for R](https://github.com/travis-ci/travis-build/pull/1830).

tomasatdatabricks (approved):
lgtm, thanks @lorenzwalthert!

PR #2204, opened 2019-12-13
Add TypeError check to call_endpoint validation

andychow-db:

When an HTTP request is executed, it tries to parse an error response as
JSON. This check only handles ValueError. `json.loads` can throw both
ValueError and TypeError. Therefore, I have added addition logic and
tests to safeguard against the behavior

## How is this patch tested?

I wrote several new unit tests to validate the JSON.loads behavior. 

- [ ] UI
- [x] CLI
- [ ] API
- [x] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### DIFF
--- a/mlflow/utils/rest_utils.py
+++ b/mlflow/utils/rest_utils.py
@@ -84,7 +84,7 @@ def _can_parse_as_json(string):
     try:
         json.loads(string)
         return True
-    except (ValueError, TypeError):
+    except Exception:  # pylint: disable=broad-except
         return False
 
 
--- a/mlflow/utils/rest_utils.py
+++ b/mlflow/utils/rest_utils.py
@@ -84,7 +84,7 @@ def _can_parse_as_json(string):
     try:
         json.loads(string)
         return True
-    except ValueError:
+    except (ValueError, TypeError):
         return False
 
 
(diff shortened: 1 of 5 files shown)


mlflow/utils/rest_utils.py
```diff
@@ -84,7 +84,7 @@ def _can_parse_as_json(string):
     try:
         json.loads(string)
         return True
-    except ValueError:
+    except (ValueError, TypeError):
```
dbczumar: @andychow-db  Are these the only exception types guaranteed to be thrown by `json.loads`? It appears that the method documentation for `json.loads` refers to `JSONDecodeError` also being thrown: https://docs.python.org/3/library/json.html#json.JSONDecodeError. Is this a subclass of `TypeError`? If not, can we simply catch `Exception` (it's probably not the best practice, but I'd rather not miss a class of exception here)?

andychow-db: JSONDecodeError is a subclass of ValueError [ref](https://docs.python.org/3/library/json.html#json.JSONDecodeError) and already handled with the existing code.

smurching: To confirm, the TypeError is thrown when the argument passed to `json.loads` is `None` right?

smurching: Agree with @dbczumar that it might be best to just catch a general `Exception` here, since the exception-throwing behavior of `json.loads` [isn't well-documented](https://docs.python.org/3/library/json.html#json.loads). It seems more likely that we'd miss a class of exception inadvertently (by not catching `Exception`) than that we'd accidentally swallow an important exception (if we were to catch `Exception`)

smurching (commented):
Agree with @dbczumar's comment around just catching broad exceptions when validating that an object can be parsed as JSON, otherwise LGTM!

mlflow/utils/rest_utils.py
andychow-db: Agreed, I've replaced the explicit error types with the general Exception type.

smurching (approved):
LGTM

PR #2209, opened 2019-12-16
Added backendConfig tag for mlflow project

t-henri:

Added a tag mlflow.project.backendConfig to store this information in the DB. Retrieve this tag in the UI to add the backend-config option in the run command.

## How is this patch tested?

Locally, checking that the backend config appears in the command in the UI

Added backend config in run command line in the ui.

- [X] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/projects/__init__.py
+++ b/mlflow/projects/__init__.py
@@ -131,8 +131,10 @@ def _run(uri, experiment_id, entry_point="main", version=None, parameters=None,
     if backend == "databricks":
         tracking.MlflowClient().set_tag(active_run.info.run_id, MLFLOW_PROJECT_BACKEND,
                                         "databricks")
-        tracking.MlflowClient().set_tag(active_run.info.run_id, MLFLOW_PROJECT_BACKEND_CONFIG,
-                                        backend_config)
+        if (backend_config and type(backend_config) != dict
+                and os.path.splitext(backend_config)[-1] == ".json"):
+            tracking.MlflowClient().set_tag(active_run.info.run_id, MLFLOW_PROJECT_BACKEND_CONFIG,
+                                            backend_config)
         from mlflow.projects.databricks import run_databricks
         return run_databricks(
             remote_run=active_run,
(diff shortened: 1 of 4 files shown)


dbczumar:
@t-henri Thanks for this contribution. I'm a little hesitant to merge immediately because displaying the config information in the UI seems to have the potential to make sensitive info (credentials, etc) visible; I'd like to explore the ramifications for the Databricks config in particular. I'll discuss this with the team and get back to you!

t-henri:
Hello @dbczumar, from my understanding backend-config is a path to a file so it shouldn't display sensitive informations in the UI.

t-henri:
Hello @dbczumar, as we discussed yesterday I added a condition to log the backendConfig. I used the same condition used in the `run` function that checks if we need to read the file.

fhoering:
@dbczumar 
I also think the config should be logged to be able to replay the project easily. It is some important information with which specification the project worked at that time.
Shouldn't the sensitive information be rather stored in some environment variables and not in the backend_config file ?

For the YARN backend it contains information like cross cluster access, memory/vcore useage for the container. Something that is pretty much the same for each run inside the project (but still project specific). 
That's why I implemented to pickup a default backend_config.json from the project sources. Such that the user doesn't have to specify this all the time.
(https://github.com/criteo/mlflow-yarn/blob/master/mlflow_yarn/yarn_backend.py#L170)

WDYT ?

dbczumar:
> @dbczumar
> I also think the config should be logged to be able to replay the project easily. It is some important information with which specification the project worked at that time.
> Shouldn't the sensitive information be rather stored in some environment variables and not in the backend_config file ?
> 
> For the YARN backend it contains information like cross cluster access, memory/vcore useage for the container. Something that is pretty much the same for each run inside the project (but still project specific).
> That's why I implemented to pickup a default backend_config.json from the project sources. Such that the user doesn't have to specify this all the time.
> (https://github.com/criteo/mlflow-yarn/blob/master/mlflow_yarn/yarn_backend.py#L170)
> 
> WDYT ?

@fhoering @t-henri The use case makes sense to me, but I'm reluctant to start logging this tag by default, since users often do have sensitive content in their backend configs. Instead, we can add an flag / option to `mlflow run` (and `mlflow.projects.run()`) specifying whether or not to log the backend config with default value `False`?

fhoering:
OK. I suggest to put it in the ML project yaml file such that the user doesn't have to specify this all the time.

Ideally there won't be any more parameters added to mlflow.run. I think it should be easy for a user run a project. 
'mlflow run entrypoint backend custom_params'. Done. 

Every more parameter will make the user think about what to do. logging the backend config could be set up once by one person for the whole project and then the team can experiment with 'mllfow run' all the time easily.

PR #2210, opened 2019-12-16
Correctly logging model summary with mlflow.keras.autolog()

mobiuscreek:

* Removes following line that raises the exception below. The exception is related to ```mlflow.set_tag()```  (correctly) not accepting strings with length more than 5000.   

https://github.com/mlflow/mlflow/blob/9bf984f403910aa416edf8f492956b3c21e63a7e/mlflow/keras.py#L405

> 2019/12/15 06:03:33 ERROR mlflow.utils.rest_utils: API request to http://localhost:5000/api/2.0/mlflow/runs/set-tag failed with code 500 != 200, retrying up to 0 more times. API response body: {"error_code": "INTERNAL_ERROR", "message": "Tag value 'Model: \"model_1\"\n__________________________________________________________________________________________________\nLayer (type)                    Output Shape         Param #     Connected to                     \n===================================' had length 10390, which exceeded length limit of 5000"}

* Fixes logging the summary of a keras model instead of creating a temp directory and then deleting it.

## How is this patch tested?

* The code has been tested locally and writes a ``` model_summary.txt``` file inside the ```artifacts/``` directory of an mlflow experiment.

* The code is **not** PEP8 compliant.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [x] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


mlflow/keras.py
```diff
@@ -385,33 +383,32 @@ class __MLflowKerasCallback(keras.callbacks.Callback):
...
                 epsilon = self.model.optimizer.epsilon if \
-                    type(self.model.optimizer.epsilon) is float \
+                    isinstance(self.model.optimizer.epsilon, float) \
                     else keras.backend.eval(self.model.optimizer.epsilon)
                 try_mlflow_log(mlflow.log_param, 'epsilon', epsilon)
 
             sum_list = []
             self.model.summary(print_fn=sum_list.append)
             summary = '\n'.join(sum_list)
-            try_mlflow_log(mlflow.set_tag, 'model_summary', summary)
```
juntai-zheng: We'd actually like to keep the `model_summary` as a tag for now. Instead of removing the tag completely to avoid the 5000 character limit, why don't we instead make a check for the length of the tag? If less than or equal to 5000 characters, we can log `model_summary` as a tag. If more than 5000 characters, we can instead log a tag that says `"Model summary too long to set as a tag. See the artifacts for full summary."` or something to that effect.

Additionally, would you be able to add a test that checks for this behavior? The test's location would be in `tests/keras_autolog/test_keras_autolog.py`.

mlflow/keras.py
```diff
@@ -385,33 +383,32 @@ class __MLflowKerasCallback(keras.callbacks.Callback):
...
-
-            tempdir = tempfile.mkdtemp()
```
juntai-zheng: The reason we wrote the code like this was because we don't want to write random files to the user's local directory. We temporarily create a `tempdir` to write `model_summary.txt` to, log the file as an artifact (which copies it over to the artifact uri), and then tear down the `tempdir` we created. Your implementation, while correct in end result, would leave `model_summary.txt` as a file in the local directory (which we want to avoid). I would recommend reverting this section.

juntai-zheng (changes requested):
Thanks for your work on this PR! I left a few comments - let me know if you need any help/guidance on what to do.

mlflow/keras.py
mobiuscreek: Thank you for the valuable comments! I fixed this behavior of `keras.py`, according to your suggestions [here](https://github.com/mlflow/mlflow/commit/e0fd234c02ecd8f3bfbf1d1960cf3159033f665f#diff-08cdd7de7cb6f904bb83e904b3e8fbf7R411-R424)

It now checks the length of the model summary. It sets it as a tag if the model's character length is less than 5000. If the model is larger, it sets as tag a message that it informs the user that it can't be set as a tag. The model is instead logged as an artifact in the artifacts' directory in the `model_summary.txt` file.


A test was also written based on the changes from @metazool in this [section](https://github.com/mlflow/mlflow/pull/2210/commits/d054b7c4e4b7cfcac4d62a32f0b28e2cab19b9c0#diff-eae3d2e34d3ebeb578b7c98a295e7727R136-R156)

A second large model had to be used to test this behavior and it's defined [here](https://github.com/mlflow/mlflow/pull/2210/commits/d054b7c4e4b7cfcac4d62a32f0b28e2cab19b9c0#diff-eae3d2e34d3ebeb578b7c98a295e7727R50)

mlflow/keras.py
mobiuscreek: As mentioned in the previous comment this behavior has been changed and nothing is saved in the user's directory. It is only logged as an artifact if the model is larger than 5000 characters.

mobiuscreek:
Another thing is that I couldn't run the tests that had custom markers in my local machine. The tests get deselected because the markers are not defined. Shouldn't there be a `pytest.ini` file with the markers? I tested with `pytest` version `5.3.2` and I get a `PytestUnknownMarkWarning` for `pytest.mark.large`.

metazool:
How are we doing with this, tests run green, it would be wonderful to have the work available for production use

juntai-zheng:
@metazool Thanks for the ping, taking a look now! @mobiuscreek Sorry about the delay - this repository uses `pytest==3.2.1`, so if you use that version the tests should be good to run.

mlflow/keras.py
```diff
@@ -385,33 +385,43 @@ class __MLflowKerasCallback(keras.callbacks.Callback):
...
+                               'Model summary too long to be set as a tag. '
+                               'Please see the artifacts for the model summary.')
+                tempdir = tempfile.mkdtemp()
+                try:
+                    summary_file = os.path.join(tempdir, "model_summary.txt")
+                    with open(summary_file, 'w') as f:
+                        f.write(summary)
+                    try_mlflow_log(mlflow.log_artifact, local_path=summary_file)
+                finally:
+                    shutil.rmtree(tempdir)
```
juntai-zheng: Sorry if I was unclear before, but we should be logging the model summary as an artifact regardless of whether or not the tag is longer than 5000 characters. You can move this out of the else block. 

Just to be clear, the only thing that should be changing here is what is being set as a tag. My previous comment about not writing "random files to the user's local directory" was referring to cleaning up the `tempdir` we create to order to write the artifact. Let me know if this makes sense!

tests/keras_autolog/test_keras_autolog.py
```diff
@@ -129,15 +150,15 @@ def test_keras_autolog_logs_expected_data(keras_random_data_run):
     assert data.params['optimizer_name'] == 'Adam'
     assert 'epsilon' in data.params
     assert data.params['epsilon'] == '1e-07'
-    assert 'model_summary' in data.tags
-    assert 'Total params: 6,922' in data.tags['model_summary']
-    artifacts = client.list_artifacts(keras_random_data_run.info.run_id)
-    artifacts = map(lambda x: x.path, artifacts)
-    assert 'model_summary.txt' in artifacts
```
juntai-zheng: Since we want to log the model summary as an artifact regardless of text length, we should add this line back in.

tests/keras_autolog/test_keras_autolog.py
```diff
@@ -129,15 +150,15 @@ def test_keras_autolog_logs_expected_data(keras_random_data_run):
     assert data.params['optimizer_name'] == 'Adam'
     assert 'epsilon' in data.params
     assert data.params['epsilon'] == '1e-07'
-    assert 'model_summary' in data.tags
```
juntai-zheng: I would still recommend keeping this line since it's a defensive check that the tag actually exists in `data.tags`.

juntai-zheng (commented):
Great changes! Just a few more small changes and I think we can get this in. Let me know if you have any more questions.

mobiuscreek:
Hi, I hope all the outstanding issues are resolved now.

juntai-zheng:
@mobiuscreek Thanks for the changes! Could you resolve the conflict merges in `test_keras_autolog.py`?

mobiuscreek:
😭  needs more fixing

juntai-zheng (changes requested):
Hi @mobiuscreek and @metazool, thanks for the changes! We could probably move the deprecation fix for `getargspec` into a separate PR so we don't block this PR. Also as a quick note, the tests related to your changes after fixing your merge conflicts [look like they passed](https://travis-ci.org/mlflow/mlflow/builds/650795956?utm_source=github_status&utm_medium=notification). The tests that failed look unrelated (flaky tests that are [being worked on](https://github.com/mlflow/mlflow/issues/2344)) - if you retrigger the tests without the deprecation fixes the tests should pass.

mobiuscreek:
Sorry for the very late reply. I reverted the changes made by the PR request from @metazool. Like you mention, this is a separate issue and can be a different PR without any of the changes made here.

metazool:
Aye, my pull request to you was speculative and I'd underestimated the impact of the changes, suggest that you in turn close this one and start a PR afresh, only including the original changes for the large model and parameterised tests (as long as they are all running!). Happy to talk it over @mobiuscreek

juntai-zheng:
Hi @mobiuscreek and @metazool, we've had some feedback that some people would rather remove the model summary from being logged in the tags entirely since the model summary is already logged as an artifact and it's difficult to work with the summary in the tags. This would remove the need for this PR and we would instead just remove the line logging the summary as a tag. 

What do y'all think of this? Did you have a specific use case from the model summary being in the tags, or would the model summary being just in the artifacts be ok for your workflow?

metazool:
This sounds completely fine to me and probably in line with @mobiuscreek's original intentions too

While i'm here, we've been trying out this PR from @arocketman on an old spacy transfer learning project and (from the client side at least) it is working very well 
https://github.com/mlflow/mlflow/pull/2242

mobiuscreek:
The original intention was to remove model logging from the tags but since this is addressed elsewhere there's no need for the changes in this PR. Thank you both 👍

juntai-zheng:
@mobiuscreek @metazool Thanks for the input! Sorry about going through this in a roundabout way - if you want to edit the PR/open a new PR to remove the model summary tag logging that would be awesome. If y'all also wanted to open a PR for changing `getargspec` to `getfullargspec` (especially since MLflow is dropping Python 2 support soon), I'd be very willing to help getting that in.

@metazool I'll poke the relevant people about the spaCy PR!

Issue #2084, opened 2019-11-12
[FR] New component for table of runs


rom1504:
## Describe the proposal

I'm proposing a new component for the table of runs using https://github.com/tannerlinsley/react-table .
Benefits : 
* easy to sort (click on the right column)
* easy to filter (write your filter on top of the column)
* proper pagination

### Motivation
The use case for us at Criteo is to be able to compare many runs (for example we do some hyperparameter tunings in different teams) on several metrics / parameters / tags.

### Proposed Changes
On top of these already merged changes : 
* filtering in sql as implemented by @jcuquemelle in https://github.com/mlflow/mlflow/pull/2030
* sorting and limiting in sql as implemented by @jcuquemelle in https://github.com/mlflow/mlflow/pull/2059

I'm proposing : 
* a whitelisting option in search_runs to retrieve only some runs, tags, metrics : I will propose a PR for this soon
=> these 3 changes make the backend fast enough (<500ms in our case)

and 2 small additional changes : 
* return the total number of runs in search_runs : this is required to have a proper pagination in the ui : I will soon propose a PR for this
* a new endpoint that retrieve the list of tags, metrics, params for an experiment : required to be able to select what columns to show in UI : I will soon propose a PR for this

All these backend changes can then be used in a new ui based on react-table.

It looks like this : 

![image](https://user-images.githubusercontent.com/2346494/68680214-98198f80-0561-11ea-8da1-99917dd2b5f8.png)

I will also submit a PR for this change (the code is mostly ready and working, but there's a few things to finish)

(cc @Zangr )

This should handle both https://github.com/mlflow/mlflow/issues/1839 and https://github.com/mlflow/mlflow/issues/1666

mateiz:
This seems great. Can it show nested runs and let the user expand those too (runs with a parent run ID)?

Zangr:
@rom1504 thank you for sketching the feature request! Glad to see continuous interest on improving the runs table. It's a good timing because I also plan to work on runs table next week to bring back the columns view. Is there a particular reason to choose ReactTable vs the current ReactVirtualized? If we were to bring in a new table component, I'd consider those ones with more out-of-the-box features.

cc @mparkhe for API changes
cc @gioa for design

rom1504:
Hello,
Here is the first PR, it's for whitelisting columns, I put it on our fork first https://github.com/criteo-forks/mlflow/pull/38 ; you can already have a look, I'll put it in this repo next.
[...]

PR #2212, opened 2019-12-16
Whitelisting columns

rom1504:

Introduce new whitelisting feature for tags, params and metrics in search_runs

This feature makes it possible to keep only some of the tags,
params and metrics in search_runs.

This is useful particularly in case of retrieving many runs but only few columns.

This commit is mostly split in 2 parts :
1. Adding tags_whitelist, metrics_whitelist, params_whitelist to the API as a non-breaking change : they default to None which mean keep all fields
2. Actually implement it in stores and in particular the sqlalchemy store

The second part is a bit subtle because the subquery loading strategy used in mlflow sqlalchemy store does not support filtering.
As a consequence I reimplemented a "disjoint_load" function that
is a simple implementation of subquery loading, adding the filtering functionality. I added many comments in it to explain how it works.
An additional change is removing the unneeded distinct in search_runs to avoid from_self from
propagating the column from ordering to the main fields of the main select. (it's not required to put fields from order by in select if the distinct keyword is not used h
ttps://dba.stackexchange.com/a/34970)

This change makes it possible to decrease the search_run time from 30s to 2s for an experiment with many runs (7000) and keeping only a few columns.

## How is this patch tested?

* adapted the existing tests
* added tests in sql alchemy store tests and file store tests to test the whitelisting

Add tags_whitelist params_whitelist metrics_whitelist optional options in search_runs.

- [ ] UI
- [ ] CLI
- [x] API
- [x] REST-API
- [x] Examples
- [x] Docs
- [x] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [x] R
- [x] Java
- [x] Python

### DIFF
--- a/mlflow/protos/service.proto
+++ b/mlflow/protos/service.proto
@@ -934,10 +934,22 @@ message SearchRuns {
 
   optional string page_token = 7;
 
+  // These metrics, params and tags whitelist limit the fields returned by the search runs query
+  // A None whitelist will select all the fields
+  optional WhiteList metrics_whitelist = 8;
+  optional WhiteList params_whitelist = 9;
+  optional WhiteList tags_whitelist = 10;
+
   message Response {
     // Runs that match the search criteria.
     repeated Run runs = 1;
     optional string next_page_token = 2;
+  }
+
+  // this whitelist message exists to be able to have an optional whitelist :
+  // no whitelist = all fields are available
+  message WhiteList  {
+    repeated string fields = 1;
   }
 }
 
--- a/mlflow/store/tracking/sqlalchemy_store.py
+++ b/mlflow/store/tracking/sqlalchemy_store.py
@@ -5,6 +5,8 @@ import math
 import posixpath
 import sqlalchemy
 import sqlalchemy.sql.expression as sql
+from itertools import groupby
+from sqlalchemy.orm import attributes
 
 from mlflow.entities.lifecycle_stage import LifecycleStage
 from mlflow.store.tracking import SEARCH_MAX_RESULTS_THRESHOLD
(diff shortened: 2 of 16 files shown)


rom1504:
Would be great if you could review this @dbczumar @mparkhe

mlflow/protos/service.proto
```diff
@@ -934,11 +934,23 @@ message SearchRuns {
 
   optional string page_token = 7;
 
+  // These metrics, params and tags whitelist limit the fields returned by the search runs query
```
mparkhe: not sure if you saw the [RFC for pagination](https://docs.google.com/document/d/1Ygti9oyMbi4DJNbaM0D49nNIKb8BLXwGYQ7WmrOKJ3s/edit?usp=sharing) we talked about in #2084 

The proposed mechanism is to use a single `column` field in `SearchRuns` proto. This would be similar to `order_by` field which is a comma separated list of columns for result ordering. 

Using `columns` field -- User would specify expected columns like `metrics.accuracy, metrics.mse, params.layers, tags.project`. Backend will parse this using the same logic uses for column names `filter` and `order_by` and return only these columns from the runs returned.

It would make sense to also return columns referred in `filter` and `order_by` query to handle pagination correctly.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -607,16 +656,30 @@ def compute_next_token(current_size):
...
+                .filter(SqlRun.experiment_id.in_(experiment_ids),
+                        SqlRun.lifecycle_stage.in_(stages),
+                        *_get_attributes_filtering_clauses(parsed_filters)) \
                 .order_by(*parsed_orderby) \
-                .offset(offset).limit(max_results).all()
-
-            runs = [run.to_mlflow_entity() for run in queried_runs]
+                .offset(offset).limit(max_results)
+
+            all_runs = queried_runs.all()
```
mparkhe: will this query database for all runs that match the filter without max limit? So for instance if there are 1M runs in the backend database, that match the filter query, are they all being pulled into memory here?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -607,16 +656,30 @@ def compute_next_token(current_size):
...
+            all_runs = disjoint_load(queried_runs, all_runs, SqlRun.tags,
+                                     SqlTag.key.in_(tags_whitelist)
+                                     if tags_whitelist is not None else None)
+            all_runs = disjoint_load(queried_runs, all_runs, SqlRun.latest_metrics,
+                                     SqlLatestMetric.key.in_(metrics_whitelist)
+                                     if metrics_whitelist is not None
+                                     else None)
+            all_runs = disjoint_load(queried_runs, all_runs, SqlRun.params,
+                                     SqlParam.key.in_(params_whitelist)
+                                     if params_whitelist is not None else None)
```
mparkhe: Can you explain how this logic works? Is this creating a SqlAlchemy query with all these criteria or is this going back to the database repeatedly and pulling more data?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -38,6 +40,53 @@
 sqlalchemy.orm.configure_mappers()
 
 
+# similar to https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#subquery-eager-loading
+# but with (optional) filtering
+def disjoint_load(query, parents, attr, whitelist=None):
```
mparkhe: This is super complex operation. Can you please explain what is happening here? It would help to add some examples in the function header to explain what each of these arguments are doing? 
It would be beneficial to understand what is happening in this method, inputs, outputs, and what is being processed.  For instance it looks like `query` is a SQL Alchemy query but it is unclear what `parents` means here! Also `attr` from the caller looks like a SQL table column but is not very clear from the function description.

mparkhe (changes requested):
Requesting changes. Some notes:

1.  This proposes changes to `SearchRuns` API that are different than the proposed RFC for pagination. We want to keep the API definition as generic as possible which will be used by multiple backend and frontends.

2. In addition to columns requested (referred to as whitelisted in this pull request), it makes sense to also return columns referred in `filter` and `order_by` fields so that
- it is clear to the user what is being filtered and returned (eg `filter = "metrics.accuracy > 0.95"` it would make sense to return `metrics.accuracy` column though not specifically requested). The same logic applies to `order_by`.
-  Not 100% convinced that pagination would work correctly if columns from `order_by` list are not fetched from backend and returned with token.

3. Finally, some of the changes and new functions added to `SqlAlchemyStore` could use some clarifying documentation and explanation. Would help to know what these functions (eg: `disjoint_load`) are doing and how it helps optimize runtime. You have added some explanation to description of the PR but would helpful to add that to function header in the code with additional examples on what is being done. Additionally, in order to review this code it would be helpful to understand more details on why this is done through subqueries rather than JOIN. What are the design decisions and tradeoffs here? Would be beneficial to document that.

mparkhe:
The use case for whitelisting on a set of metrics, params, tags is not super clear. In fact, this raises more questions like "would a user be able to search based on non-white listed params, tags ... etc". If some params, tags, metrics are not important, why are they logged. More over this property is sticky on runs so would affect different users.

Finally for search .. etc, needs whitelisting this may be achieved by projecting based on specific columns and can be achieved in clients. In future, we may choose to additional fields like `string columns` that only that API. Based on inputs, we will add this to `SearchRuns` API. Please feel free to comment in the [RFC](https://docs.google.com/document/d/1Ygti9oyMbi4DJNbaM0D49nNIKb8BLXwGYQ7WmrOKJ3s/edit?usp=sharing) mentioned above.

I am closing out this pull request since this is too specific a solution and not easily generalizable.

Issue #45, opened 2018-06-14
Metrics chart wrongly connects points on X axis


zegerhoogeboom:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: Yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 18.04
- **MLflow installed from (source or binary)**: source
- **MLflow version (run ``python -c "from mlflow import version; print(version.VERSION)"``)**: 0.1.0
- **Python version**: 3.6.5
- **npm version (if running the dev UI)**: 6.0.1
- **Exact command to reproduce**: NA

### Describe the problem
In the UI at the metrics graph view, it looks like this for me:
![image](https://user-images.githubusercontent.com/5371096/41408703-dc1d732e-6fd3-11e8-968f-f18192b908db.png)

The data points are logged at the end of an epoch using this Keras callback:
```
class LossHistory(Callback):
    def on_epoch_end(self, epoch, logs=None):
        log_metric("loss", logs.get('loss'))
        # in reality more metrics here
```
`logs.get('loss')` just returns a number. The connections of the dots in the interface should be connecting X=0 to X=1 to X=2 and so on. Why they are currently connected in this strange way, I have no clue.

mateiz:
Is there any chance you could share your MLflow run folder that caused this (`mlruns/0/<runID>`) or the file for the metric inside there? Also, which browser are you using?

zegerhoogeboom:
Terribly sorry for the late reply. `cat mlruns/0/2924aa6b449944bc849d4381e0fa3570/metrics/loss`:

```
0 4.54306590325
0 3.50579357207
0 3.63842143944
0 3.63074730358
0 3.62895112526
0 3.60019535707
0 3.34938469377
0 3.69437173072
0 3.4442458695
0 5.05011896209
```

`cat mlruns/0/2924aa6b449944bc849d4381e0fa3570/meta.yml`:
```
artifact_uri: /app/mlruns/0/2924aa6b449944bc849d4381e0fa3570/artifacts
end_time: null
entry_point_name: !!python/unicode ''
experiment_id: !!python/long '0'
name: Run 5
run_uuid: 2924aa6b449944bc849d4381e0fa3570
source_name: !!python/unicode '/home/{my source code location}.py'
source_type: 4
source_version: !!python/unicode '94de7c2a7562525db2ead17fe8d020a7ca1cf32a'
start_time: !!python/long '1529310274917'
[...]

PR #734, opened 2018-11-29
[UI] Optionally configure the JS to disable the experiment list and header

andrewmchen:
Add webpack env defines to allow us to conditionally disable the experiment list and header.

PR #2214, opened 2019-12-17
Inject private git webserver in mlflow (#45)

amrqt:

Allow usage of a private version control system when displaying experiment source and version links

We want to package the app without having to rebuild the UI with custom VCS settings so these settings are passed at runtime to the Python backend through environment variables and served to the UI.

In `mlflow/server/js/src/utils/Utils.js`:
* Added cases to `Utils.getGitRepoUrl()` and `Utils.getGitCommitUrl()` to generate urls to a private vcs. 
* Added `getPrivateVcsRegex` and `getPrivateVcsUrl` which are called when Utils.js is loaded to set the values we need for regex matching and url generation.

## How is this patch tested?

Added tests for the new JS functions, and manual testing.
 
Displays links pointing to a private version control system for run source repo and version.

- [X] UI
- [ ] CLI
- [ ] API
- [x] REST-API
- [ ] Examples
- [x] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/docs/source/rest-api.rst
+++ b/docs/source/rest-api.rst
@@ -2187,3 +2187,87 @@ View type for ListExperiments query.
 +--------------+------------------------------------------+
 | ALL          | Get all experiments.                     |
 +--------------+------------------------------------------+
+
+
+===========================
+
+
+
+.. _mlflowMlflowServicePrivateVcs:
+
+Private VCS
+===========
+
+Utilities to integrate a private version control system to MLFlow UI.
+
+The UI links the experiment source name to the desired VCS location
+by matching the sourceNmae to a regex and uses template URLs to render links to the project repository or commits.
+The regex and URL templates are read at runtime by the Python backend from environment variables:
+
++-------------------------------+-------------------------------------------+
+|     Name                      |                Description                |
++===============================+===========================================+
+| MLFLOW_PRIVATE_VCS_REGEX      | Regex used to match experiment sourceName |
++-------------------------------+-------------------------------------------+
+| MLFLOW_PRIVATE_VCS_REPO_URL   | URL template to private VCS repo          |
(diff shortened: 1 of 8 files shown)


smurching:
Hi @amrqt, sorry for my ignorance on this topic - would you mind clarifying which challenge(s) you're currently facing with  exposing private VCS URLs in the MLflow UI? I'd expect such URLs to render properly in the UI, and only those with access to be able to actually view the code in the repositories

amrqt:
Hi @smurching,
The current implementation (in js/src/utils/Utils.js) only recognizes and generates links for Github, Gitlab and Bitbucket repositories. We wanted to be able to display repository and version links to our own vcs which is a self-hosted Gerrit.
In addition we wanted to avoid setting the required regex and vcs host name directly in the UI js files to keep the distributable generic and configurable at runtime through environment variables, thus the need for the UI to fetch these parameters from the backend. We tried to make this as generic as possible so that anyone with another VCS solution can pass a regex and host name for their own setup.
I hope this answer your questions, I'm available if you have more and also open to suggestions for a better implementation

smurching:
@amrqt sorry for the delay, that use case makes sense & definitely agree that it'd be good to make the VCS-URL-parsing configurable vs adding more hardcoded regexes.

I wonder if we could make the list of recognized regexes configurable via an environment variable read at JS build time - that way we could avoid adding a new MLflow REST API to satisfy this particular UI-only use case. See this PR for example https://github.com/mlflow/mlflow/pull/734/files, let me know what you think!

aarondav:
@amrqt Did you get a chance to look at @smurching 's feedback?  What do you think?

PR #2216, opened 2019-12-17
Fix mlflow.keras.save_model with DBFS

andychow-db:

Fixed a bug where `mlflow.keras.save_model` failed against "/dbfs/..." paths due to the lack of support for random writes when using DBFS FUSE.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -37,6 +37,8 @@ _CUSTOM_OBJECTS_SAVE_PATH = "custom_objects.cloudpickle"
 _KERAS_MODULE_SPEC_PATH = "keras_module.txt"
 # File name to which keras model is saved
 _MODEL_SAVE_PATH = "model.h5"
+# Conda env subpath when saving/loading model
+_CONDA_ENV_SUBPATH = "conda.yaml"
 
 
 def get_default_conda_env(include_cloudpickle=False, keras_module=None):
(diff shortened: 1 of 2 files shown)


tomasatdatabricks (approved):
lgtm! thx @andychow-db

tomasatdatabricks:
Hey @andychow-db is this ready for review / merge? I have already approved it, do you think it misses something? I would merge it otherwise.

mlflow/keras.py
```diff
@@ -150,33 +152,58 @@ def _is_tf_keras(model):
     elif type(keras_module) == str:
         keras_module = importlib.import_module(keras_module)
 
+    # check if path exists
```
smurching: One small note on comments, we tend to prefer not adding them in favor of writing code that's easy to follow, reserving their use for cases where it's hard to tell exactly what's going on (e.g. here from the variable names & logic we can see that we're checking whether the path already exists). That's super fuzzy I know but hopefully it makes some sense - the idea is that comments can go out of date, so it's always better if the code is simple enough to follow on its own, but for more complex cases comments are definitely helpful

mlflow/keras.py
```diff
@@ -150,33 +152,58 @@ def _is_tf_keras(model):
...
+    # save keras module spec to path/data/keras_module.txt
     with open(os.path.join(data_path, _KERAS_MODULE_SPEC_PATH), "w") as f:
         f.write(keras_module.__name__)
+
+    # save keras model to path/data/model.h5
     model_subpath = os.path.join(data_subpath, _MODEL_SAVE_PATH)
-    keras_model.save(os.path.join(path, model_subpath), **kwargs)
+    model_path = os.path.join(path, model_subpath)
+    if path.startswith('/dbfs/'):
+        # The Databricks Filesystem uses a FUSE implementation that does not support
```
smurching: This is a great example of a good comment, it's not obvious why we're checking for "/dbfs/" paths without it

smurching (approved):
LGTM! Just left a note on comments for the future

mlflow/keras.py
andychow-db: Sounds good. I'll be sure to avoid superfluous comments in the future.

PR #2030, opened 2019-10-31
Push filter function from Python to SQL

jcuquemelle:
* Mutualize test data generation
* extract filtering string validity checks
* Check filtering on latest_metrics name of metrics changed to use an underscore, easier to use as filter keys
* Add sqlAlchemy filtering
* Don't join subqueries for SqlRun attributes filtering directly specify them as filters on the main query
* Mutualize switch on operators using a dict of operators
* Document the fact that key_name is a valid searchable attribute
* Add a field name conversion method from mlflow entity to SQLRun

## What changes are proposed in this pull request?

(Please fill in changes proposed in this fix)

## How is this patch tested?

(Details)

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

Further improve performance of search against SQL-backed stores by pushing filter logic down to SQL from Python

### What component(s) does this PR affect?

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [ ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [ ] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #2217, opened 2019-12-17
Add like operator support

t-henri:

Added support for LIKE/ILIKE operator in filter string when filtering runs.

## How is this patch tested?

With unit tests and locally in the UI.

Added support for LIKE/ILIKE operator in filter string when filtering runs.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

@dbczumar I merged the PR in our fork, let's continue the discussion here.

### DIFF
--- a/mlflow/store/tracking/sqlalchemy_store.py
+++ b/mlflow/store/tracking/sqlalchemy_store.py
@@ -652,14 +652,16 @@ def _get_attributes_filtering_clauses(parsed):
         key_type = sql_statement.get('type')
         key_name = sql_statement.get('key')
         value = sql_statement.get('value')
-        comparator = sql_statement.get('comparator')
+        comparator = sql_statement.get('comparator').lower()
         if SearchUtils.is_attribute(key_type, comparator):
-            # validity of the comparator is checked in SearchUtils.parse_search_filter()
-            op = SearchUtils.filter_ops.get(comparator)
-            if op:
-                # key_name is guaranteed to be a valid searchable attribute of entities.RunInfo
-                # by the call to parse_search_filter
-                attribute_name = SqlRun.get_attribute_name(key_name)
+            # key_name is guaranteed to be a valid searchable attribute of entities.RunInfo
+            # by the call to parse_search_filter
+            attribute_name = SqlRun.get_attribute_name(key_name)
+            if comparator in ['like', 'ilike']:
+                op = SearchUtils.get_sql_filter_ops(getattr(SqlRun, attribute_name), comparator)
+                clauses.append(op(value))
+            elif comparator in SearchUtils.filter_ops:
+                op = SearchUtils.filter_ops.get(comparator)
                 clauses.append(op(getattr(SqlRun, attribute_name), value))
     return clauses
 
@@ -668,7 +670,7 @@ def _to_sqlalchemy_filtering_statement(sql_statement, session):
     key_type = sql_statement.get('type')
     key_name = sql_statement.get('key')
     value = sql_statement.get('value')
-    comparator = sql_statement.get('comparator')
+    comparator = sql_statement.get('comparator').lower()
 
     if SearchUtils.is_metric(key_type, comparator):
         entity = SqlLatestMetric
@@ -683,9 +685,16 @@ def _to_sqlalchemy_filtering_statement(sql_statement, session):
         raise MlflowException("Invalid search expression type '%s'" % key_type,
                               error_code=INVALID_PARAMETER_VALUE)
 
-    # validity of the comparator is checked in SearchUtils.parse_search_filter()
-    op = SearchUtils.filter_ops.get(comparator)
-    if op:
+    if comparator in ['like', 'ilike']:
+        op = SearchUtils.get_sql_filter_ops(entity.value, comparator)
+        return (
+            session
+            .query(entity)
+            .filter(entity.key == key_name, op(value))
+            .subquery()
+        )
+    elif comparator in SearchUtils.filter_ops:
+        op = SearchUtils.filter_ops.get(comparator)
         return (
             session
             .query(entity)
--- a/mlflow/store/tracking/sqlalchemy_store.py
+++ b/mlflow/store/tracking/sqlalchemy_store.py
@@ -652,17 +652,17 @@ def _get_attributes_filtering_clauses(parsed):
         key_type = sql_statement.get('type')
         key_name = sql_statement.get('key')
         value = sql_statement.get('value')
-        comparator = sql_statement.get('comparator').lower()
+        comparator = sql_statement.get('comparator').upper()
         if SearchUtils.is_attribute(key_type, comparator):
             # key_name is guaranteed to be a valid searchable attribute of entities.RunInfo
             # by the call to parse_search_filter
-            attribute_name = SqlRun.get_attribute_name(key_name)
-            if comparator in ['like', 'ilike']:
-                op = SearchUtils.get_sql_filter_ops(getattr(SqlRun, attribute_name), comparator)
+            attribute = getattr(SqlRun, SqlRun.get_attribute_name(key_name))
+            if comparator in SearchUtils.CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS:
+                op = SearchUtils.get_sql_filter_ops(attribute, comparator)
                 clauses.append(op(value))
             elif comparator in SearchUtils.filter_ops:
                 op = SearchUtils.filter_ops.get(comparator)
-                clauses.append(op(getattr(SqlRun, attribute_name), value))
+                clauses.append(op(attribute, value))
     return clauses
 
 
@@ -670,7 +670,7 @@ def _to_sqlalchemy_filtering_statement(sql_statement, session):
     key_type = sql_statement.get('type')
     key_name = sql_statement.get('key')
     value = sql_statement.get('value')
-    comparator = sql_statement.get('comparator').lower()
+    comparator = sql_statement.get('comparator').upper()
 
     if SearchUtils.is_metric(key_type, comparator):
         entity = SqlLatestMetric
@@ -685,7 +685,7 @@ def _to_sqlalchemy_filtering_statement(sql_statement, session):
         raise MlflowException("Invalid search expression type '%s'" % key_type,
                               error_code=INVALID_PARAMETER_VALUE)
 
-    if comparator in ['like', 'ilike']:
+    if comparator in SearchUtils.CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS:
         op = SearchUtils.get_sql_filter_ops(entity.value, comparator)
         return (
             session
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -1,9 +1,10 @@
 import base64
 import json
 import operator
+import re
 
 import sqlparse
-from sqlparse.sql import Identifier, Token, Comparison, Statement
+from sqlparse.sql import Identifier, Token, Comparison, Statement, TokenList
 from sqlparse.tokens import Token as TokenType
 
 from mlflow.entities import RunInfo
@@ -15,9 +16,9 @@ import math
 
 class SearchUtils(object):
     VALID_METRIC_COMPARATORS = set(['>', '>=', '!=', '=', '<', '<='])
-    VALID_PARAM_COMPARATORS = set(['!=', '='])
-    VALID_TAG_COMPARATORS = set(['!=', '='])
-    VALID_STRING_ATTRIBUTE_COMPARATORS = set(['!=', '='])
+    VALID_PARAM_COMPARATORS = set(['!=', '=', 'LIKE', 'like', 'ILIKE', 'ilike'])
+    VALID_TAG_COMPARATORS = set(['!=', '=', 'LIKE', 'like', 'ILIKE', 'ilike'])
+    VALID_STRING_ATTRIBUTE_COMPARATORS = set(['!=', '=', 'LIKE', 'like', 'ILIKE', 'ilike'])
     VALID_SEARCH_ATTRIBUTE_KEYS = set(RunInfo.get_searchable_attributes())
     VALID_ORDER_BY_ATTRIBUTE_KEYS = set(RunInfo.get_orderable_attributes())
     _METRIC_IDENTIFIER = "metric"
@@ -44,7 +45,17 @@ class SearchUtils(object):
         '!=': operator.ne,
         '<=': operator.le,
         '<': operator.lt,
+        'like': re.match,
+        'ilike': re.match
     }
+
+    @classmethod
+    def get_sql_filter_ops(cls, column, operator):
+        sql_filter_ops = {
+            'like': column.like,
+            'ilike': column.ilike
+        }
+        return sql_filter_ops[operator]
 
     @classmethod
     def _trim_ends(cls, string_value):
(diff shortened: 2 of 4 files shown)


mlflow/utils/search_utils.py
```diff
@@ -15,9 +16,9 @@
 
 class SearchUtils(object):
     VALID_METRIC_COMPARATORS = set(['>', '>=', '!=', '=', '<', '<='])
-    VALID_PARAM_COMPARATORS = set(['!=', '='])
-    VALID_TAG_COMPARATORS = set(['!=', '='])
-    VALID_STRING_ATTRIBUTE_COMPARATORS = set(['!=', '='])
+    VALID_PARAM_COMPARATORS = set(['!=', '=', 'LIKE', 'like', 'ILIKE', 'ilike'])
```
mparkhe: as long as operators are converted to **UPPER** case, only `LIKE` and `ILIKE` should be required.
Prefer **UPPER** case over lower case to distinguish SQL tokens from python code.

mlflow/utils/search_utils.py
```diff
@@ -187,7 +198,21 @@ def _invalid_statement_token(cls, token):
...
         invalids = list(filter(cls._invalid_statement_token, statement.tokens))
-        if len(invalids) > 0:
+        if len(invalids) % 3 == 0 and len(invalids) != 0:
+            groupby_invalids = [invalids[i:i+3] for i in range(0, len(invalids), 3)]
+            invalid_clauses = ""
+            for invalid in groupby_invalids:
+                if invalid[1].value.lower() in ['like', 'ilike']:
+                    like_comparison = Comparison(TokenList(invalid))
+                    statement.tokens = [x for x in statement.tokens if type(x) == Comparison]
+                    statement.tokens.append(like_comparison)
```
mparkhe: Can parsing and validating `LIKE` and `ILIKE` tokens be pushed down to `_invalid_statement_token`. Let's keep the logic of on invalids as purely invalid.

mlflow/utils/search_utils.py
```diff
@@ -187,7 +198,21 @@ def _invalid_statement_token(cls, token):
     def _process_statement(cls, statement):
         # check validity
         invalids = list(filter(cls._invalid_statement_token, statement.tokens))
-        if len(invalids) > 0:
+        if len(invalids) % 3 == 0 and len(invalids) != 0:
+            groupby_invalids = [invalids[i:i+3] for i in range(0, len(invalids), 3)]
+            invalid_clauses = ""
+            for invalid in groupby_invalids:
+                if invalid[1].value.lower() in ['like', 'ilike']:
```
mparkhe: 1. Define as constants (defined at the top of the file) rather than hard coding here
2. Prefer **UPPER** case for SQL keywords to distinguish from python code.

mlflow/utils/search_utils.py
```diff
@@ -272,7 +297,20 @@ def _does_run_match_clause(cls, run, sed):
                                   error_code=INVALID_PARAMETER_VALUE)
         if lhs is None:
             return False
-        if comparator in cls.filter_ops.keys():
+
+        if comparator in ['like', 'ilike']:
```
mparkhe: use constants rather than hardcoding tokens here

mlflow/utils/search_utils.py
```diff
@@ -272,7 +297,20 @@ def _does_run_match_clause(cls, run, sed):
...
+            # Change value from sql syntax to regex syntax
+            if comparator == 'ilike':
+                value = value.lower()
+                lhs = lhs.lower()
+            if not value.startswith('%'):
+                value = '^' + value
+            if not value.endswith('%'):
+                value = value + '$'
```
mparkhe: Is this required for SQLAlchemy? As if query is specified as `tags.algorithm LIKE '%st'` SQL code would only return values ending in **st**. Is this `$` enforcement required?

mlflow/utils/search_utils.py
```diff
@@ -272,7 +297,20 @@ def _does_run_match_clause(cls, run, sed):
...
+            value = value.replace('_', '.').replace('%', '.*')
```
mparkhe: Not clear on the logic here--can you help understand this translation?
I tried sqlite and `%` should work.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -652,14 +652,16 @@ def _get_attributes_filtering_clauses(parsed):
         key_type = sql_statement.get('type')
         key_name = sql_statement.get('key')
         value = sql_statement.get('value')
-        comparator = sql_statement.get('comparator')
+        comparator = sql_statement.get('comparator').lower()
```
mparkhe: prefer `.upper()`

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -652,14 +652,16 @@ def _get_attributes_filtering_clauses(parsed):
...
-                # by the call to parse_search_filter
-                attribute_name = SqlRun.get_attribute_name(key_name)
+            # key_name is guaranteed to be a valid searchable attribute of entities.RunInfo
+            # by the call to parse_search_filter
+            attribute_name = SqlRun.get_attribute_name(key_name)
+            if comparator in ['like', 'ilike']:
+                op = SearchUtils.get_sql_filter_ops(getattr(SqlRun, attribute_name), comparator)
+                clauses.append(op(value))
+            elif comparator in SearchUtils.filter_ops:
+                op = SearchUtils.filter_ops.get(comparator)
```
mparkhe: Can these two cases share `getattr(SqlRun, attribute_name)`. 
Also add a constant like `CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS` inside `SearchUtils` as `set(['LIKE', 'ILIKE'])` and use that everywhere.

mparkhe (changes requested):
Thank you for taking up this feature! Some requested changes. Details inlined...

mlflow/utils/search_utils.py
t-henri: I can do that, but it would be even better to wait for this pr from @dbczumar to be merged and released: https://github.com/andialbrecht/sqlparse/pull/525
What do you think ? Do I make the change while waiting for the pr to be merged ?

mlflow/utils/search_utils.py
t-henri: In fact this part of the code is only used by file_store, so the goal here is to convert the sql syntax to a regex syntax so '^' and '$' are used to say that we want to match the beginning or the end of the string.

mlflow/utils/search_utils.py
t-henri: Same that the previous comment, it is only used by file store, so % (match any characters) is remplaced by .* in a regex.

mlflow/utils/search_utils.py
t-henri: Done

mlflow/utils/search_utils.py
t-henri: Done

mlflow/utils/search_utils.py
t-henri: Done

mlflow/store/tracking/sqlalchemy_store.py
t-henri: Done

mlflow/store/tracking/sqlalchemy_store.py
t-henri: Done

mlflow/utils/search_utils.py
mparkhe: I wonder if store specific syntax parsing and conversion belongs in those specific store. This is the generic API level code parsing and the objective is to be SQL compliant.

Can you move the `^` and `$` handling to `FileStore`?

t-henri: I agree, this should be in the file store the same way the functions to filter sql are in the sql alchemy store. (since this pr: https://github.com/mlflow/mlflow/pull/2030)

But if we follow this logic, `search_utils.filter` and `search_utils._does_run_match_clause` should be entirely in file store since it is only used by this store and the logic is specific for searching in a file. Should I move these two functions to the file store ?

mlflow/utils/search_utils.py
aarondav: I posted an issue asking when they plan to release @dbczumar's change: https://github.com/andialbrecht/sqlparse/issues/533

mlflow/utils/search_utils.py
aarondav: The functions you mention used to be used by both file store and sql store, before we pushed down the SQL implementation to the database. Now you are right, those should exist only in the file store. I think it is reasonable to move them in this PR.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -652,23 +652,25 @@ def _get_attributes_filtering_clauses(parsed):
...
-            op = SearchUtils.filter_ops.get(comparator)
-            if op:
-                # key_name is guaranteed to be a valid searchable attribute of entities.RunInfo
-                # by the call to parse_search_filter
-                attribute_name = SqlRun.get_attribute_name(key_name)
-                clauses.append(op(getattr(SqlRun, attribute_name), value))
+            # key_name is guaranteed to be a valid searchable attribute of entities.RunInfo
+            # by the call to parse_search_filter
+            attribute = getattr(SqlRun, SqlRun.get_attribute_name(key_name))
+            if comparator in SearchUtils.CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS:
```
aarondav: Can we add documentation on LIKE and ILIKE, for [this page](https://mlflow.org/docs/latest/search-syntax.html#id9)? We should also either clarify that LIKE may also be case-insensitive depending on the circumstance (e.g., sqlite is case-insensitive) or somehow make it always case-sensitive (seems preferable if possible).

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -939,6 +939,12 @@ def test_search_params(self):
         filter_string = "params.p_b = 'ABC'"
         six.assertCountEqual(self, [r2], self._search(experiment_id, filter_string))
 
+        filter_string = "params.generic_2 LIKE '%other%'"
+        six.assertCountEqual(self, [r2], self._search(experiment_id, filter_string))
```
aarondav: Can we have a negative case for `LIKE '%Other%'`?  This should return no results (or the same results if sqlite, with a comment).

mlflow/utils/search_utils.py
```diff
@@ -15,9 +16,10 @@
 
 class SearchUtils(object):
     VALID_METRIC_COMPARATORS = set(['>', '>=', '!=', '=', '<', '<='])
-    VALID_PARAM_COMPARATORS = set(['!=', '='])
-    VALID_TAG_COMPARATORS = set(['!=', '='])
-    VALID_STRING_ATTRIBUTE_COMPARATORS = set(['!=', '='])
+    VALID_PARAM_COMPARATORS = set(['!=', '=', 'LIKE', 'ILIKE'])
+    VALID_TAG_COMPARATORS = set(['!=', '=', 'LIKE', 'ILIKE'])
+    VALID_STRING_ATTRIBUTE_COMPARATORS = set(['!=', '=', 'LIKE', 'ILIKE'])
+    CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS = set(['LIKE', 'ILIKE'])
```
aarondav: I don't understand the meaning of `CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS`. Are there non-case-insensitive operators? Could we not use the same strategy for. all `VALID_STRING_ATTRIBUTE_COMPARATORS`?

mlflow/utils/search_utils.py
```diff
@@ -187,7 +199,21 @@ def _invalid_statement_token(cls, token):
     def _process_statement(cls, statement):
         # check validity
         invalids = list(filter(cls._invalid_statement_token, statement.tokens))
-        if len(invalids) > 0:
+        if len(invalids) % 3 == 0 and len(invalids) != 0:
```
aarondav: Could you add a comment on what this is doing and why? I'm not sure why this PR requires this addition, but it seems complicated.

mlflow/utils/search_utils.py
```diff
@@ -44,8 +46,18 @@ class SearchUtils(object):
         '!=': operator.ne,
         '<=': operator.le,
         '<': operator.lt,
+        'LIKE': re.match,
```
aarondav: This doesn't seem to be the same kind of thing as the other operators, what's going on?

mlflow/utils/search_utils.py
```diff
@@ -44,8 +46,18 @@ class SearchUtils(object):
...
+        'ILIKE': re.match
     }
 
+    @classmethod
+    def get_sql_filter_ops(cls, column, operator):
+        sql_filter_ops = {
+            'LIKE': column.like,
```
aarondav: What is `column.like`?

t-henri: column.like is the way we perform the like operation with sqlalchemy. (column being a column of a sqlalchemy object that we pass as parameter here)

mlflow/utils/search_utils.py
t-henri: I'm not sure to understand your comment, here the operator corresponding to the LIKE operator is a regexp for the file store, we couldn't use a basic operator like equal or le.

mlflow/utils/search_utils.py
t-henri: This is a part of the code we will remove once @dbczumar PR will be released in sqlparse, before that it didn't recognize the LIKE keyword.

tests/store/tracking/test_sqlalchemy_store.py
t-henri: Actually the sqlalchemy store tests run on sqllite, so this would return the same results

mlflow/store/tracking/sqlalchemy_store.py
t-henri: Sure, I'll add documentation when I update my PR after a new release of sqlparse (the pr is kind of blocked waiting for that)

aarondav: The author of that repo said there should be a new release by the end of the month. Since this is not super urgent, I guess we can wait for that.

dbczumar:
@aarondav @t-henri Now that `sqlparse` has been updated, is it possible to revive this PR and move towards merge?

t-henri:
Hello @dbczumar thanks for the information, I just pushed a commit that removes the part that parsed the invalid statement to get the like operator.

mparkhe (commented):
This looks great. Thank you for the fixed. I would recommend adding negative tests for SQL `like`

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -980,6 +986,20 @@ def test_search_tags(self):
                                                       filter_string="tags.p_a = 'abc'"))
         six.assertCountEqual(self, [r2], self._search(experiment_id,
                                                       filter_string="tags.p_b = 'ABC'"))
+        six.assertCountEqual(self, [r2],
+                             self._search(experiment_id,
+                                          filter_string="tags.generic_2 LIKE '%other%'"))
```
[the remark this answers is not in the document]
mparkhe: create #2708 to handle this and more tests

mparkhe (approved):

PR #2219, opened 2019-12-18
EarlyStopping Callback support for Keras Autologging

juntai-zheng:

If `EarlyStopping` is used with Keras autologging and the `restore_best_weights` parameter of the call back is set to be `True`, then MLflow will log the metrics of the model restored by `EarlyStopping` as an extra step/epoch (past where training ended). This allows for metrics such as `loss` and `acc` to be more easily compared with other models, as searches/comparisons will now compare the metrics of the restored model rather than the metrics of the model at the end of training.

This PR also adds minor changes to `tests/keras_autolog/test_keras_autolog.py` by adding `tracking_uri_mock` and a deterministic seed. 

## How is this patch tested?

Tests are added to `tests/keras_autolog/test_keras_autolog.py`.

Keras autologging now supports the EarlyStopping callback. If `EarlyStopping.restore_best_weights==True`, then the metrics of the restored model will be logged as an extra step.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [X] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python


mlflow/keras.py
```diff
@@ -434,21 +440,31 @@ def fit(self, *args, **kwargs):
...
         else:
             kwargs['callbacks'] = [__MLflowKerasCallback()]
 
-        result = original(self, *args, **kwargs)
+        history = original(self, *args, **kwargs)
+
+        if early_stop_callback and early_stop_callback.restore_best_weights:
+            stopped_epoch_metrics = {key: history.history[key][early_stop_callback.stopped_epoch]
+                                     for key in history.history.keys()}
+            last_epoch = len(history.history['loss'])
```
smurching: Let's be defensive here in case 'loss' is absent
```suggestion
            loss_history = history.history.get('loss')
            if loss_history is None:
                continue
            last_epoch = len(loss_history)
            ...
```

Or could we just use one of the metrics that we know for sure exists (e.g. pick a key from `history.history.keys()`)?

smurching (commented):
Mostly looks good! Just a small comment

mlflow/keras.py
juntai-zheng: Good catch, meant to change it forgot to go back. Will pick a key that exists.

mlflow/keras.py
```diff
@@ -498,8 +501,11 @@ def fit_generator(self, *args, **kwargs):
         if early_stop_callback and early_stop_callback.restore_best_weights:
             stopped_epoch_metrics = {key: history.history[key][early_stop_callback.stopped_epoch]
                                      for key in history.history.keys()}
-            last_epoch = len(history.history['loss'])
-            try_mlflow_log(mlflow.log_metrics, stopped_epoch_metrics, step=last_epoch+1)
+            # Checking that a metric history exists
+            metric_key = next(iter(history.history))
```
smurching: I think this actually throws instead of returning `None`:

```
(mlflow-python3) ➜  mlflow git:(python-mlflow-spark-autologging) python
Python 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> empty_lst = []
>>> next(iter(empty_lst))
StopIteration
```

juntai-zheng: Well that's unfortunate. I'll add `None` as a default argument for `next` to fix it.

mlflow/keras.py
```diff
@@ -421,6 +421,12 @@ def on_epoch_end(self, epoch, logs=None):
         def on_train_end(self, logs=None):
             try_mlflow_log(log_model, self.model, artifact_path='model')
 
+    def early_stop_check(callbacks):
+        for callback in callbacks:
+            if isinstance(callback, keras.callbacks.callbacks.Callback):
```
srowen: This is probably an ignorant question, but is this definitely an early stopping callback? it seems to just check if it's a keras callback, which could be many things.

juntai-zheng: Not ignorant - I copy pasted the wrong class (the parent) by accident. Thanks for the catch!

tests/keras_autolog/test_keras_autolog.py
```diff
@@ -149,8 +152,80 @@ def test_keras_autolog_logs_default_params(keras_random_data_run):
...
+def keras_random_data_run_with_early_stop(all_zeros_train_data, fit_variant,
+                                          all_zeros_one_hot_labels, manual_run,
+                                          restore_best_weights):
+    mlflow.keras.autolog()
+
+    data = all_zeros_train_data
+    labels = all_zeros_one_hot_labels
+
+    model = create_model()
+    callback = keras.callbacks.callbacks.EarlyStopping(monitor='loss', patience=5,
```
smurching: Thanks @srowen for the catch on the callback type check! @juntai-zheng let's add a preventative "regression" test against that bug by adding a test that runs with another unrelated callback, e.g. a `ProgbarLogger` callback (see how to construct one [from the Keras docs](https://keras.io/callbacks/)). We should be able to verify that the test fails without the fix & passes with the fix.

mlflow/keras.py
```diff
@@ -461,21 +467,36 @@ def fit(self, *args, **kwargs):
...
             kwargs['callbacks'] = [__MLflowKerasCallback()]
 
-        result = original(self, *args, **kwargs)
+        history = original(self, *args, **kwargs)
+
+        if early_stop_callback:
+            try_mlflow_log(mlflow.log_metric, 'stopped_epoch', early_stop_callback.stopped_epoch)
+            if early_stop_callback.restore_best_weights:
+                stopped_metrics = {key: history.history[key][early_stop_callback.stopped_epoch]
+                                   for key in history.history.keys()}
```
harupy: When the user set `patience` to > 0, the stopped epoch does not correspond to the best epoch.

https://github.com/keras-team/keras/blob/16bd239e34d853f3bf7ae99f14e0324f924ca371/keras/callbacks/callbacks.py#L742

srowen: Good point, this has to be stopped_epoch - patience, effectively, right?

harupy: We need to subtract `max(1, patience)`.

harupy: We also need to consider the case where early stopping didn't occur.

juntai-zheng: Thanks for the catch! Will put the fix in for logging the correct epoch. In the case of early stopping not occurring and `restore_best_weights` is set to be true, the best weights aren't actually restored (see [this issue](https://github.com/keras-team/keras/issues/12511)). Therefore, I don't think we need to find the best epoch when early stopping doesn't happen, though I do need to add a check for not logging an extra step if that is the case.

mlflow/keras.py
```diff
@@ -461,21 +467,43 @@ def fit(self, *args, **kwargs):
...
+            if stopped_epoch != 0 and early_stop_callback.restore_best_weights:
+                best_epoch = stopped_epoch - max(1, patience)
+                try_mlflow_log(mlflow.log_metric, 'best_epoch', best_epoch)
+                best_metrics = {key: history.history[key][best_epoch]
+                                for key in history.history.keys()}
+                # Checking that a metric history exists
+                metric_key = next(iter(history.history), None)
+                if metric_key is not None:
+                    last_epoch = len(history.history[metric_key])
+                    try_mlflow_log(mlflow.log_metrics, best_metrics, step=last_epoch)
```
harupy: I think these lines should be put in a function so that we can reuse it on fit_generator.

juntai-zheng: Good idea! Will do that.

docs/source/tracking.rst
```diff
@@ -260,6 +260,23 @@ If a run exists when ``autolog()`` captures data, MLflow will log to that run an
   - This feature is experimental - the API and format of the logged data are subject to change.
 
 
+EarlyStopping Integration with Keras Automatic Logging
```
smurching: Thanks for the detail here, IMO we can put this content into the docstring of `mlflow.keras.autolog()` instead of in this top-level tracking doc as it's pretty specific to Keras autologging & the goal of this doc page is to summarize tracking features at a high-level.

mlflow/keras.py
```diff
@@ -449,11 +449,33 @@ def on_train_end(self, logs=None):
...
+            es_callback = keras.callbacks.callbacks.EarlyStopping
         for callback in callbacks:
-            if isinstance(callback, keras.callbacks.callbacks.EarlyStopping):
+            if isinstance(callback, es_callback):
                 return callback
         return None
 
+    def log_early_stop(callback, history):
+        if callback:
+            stopped_epoch = callback.stopped_epoch
```
smurching: Since `stopped_epoch` & `patience` aren't publicly-documented fields, we might want to try-catch this block for forwards-compatibility, e.g. define a helper like

```
def _get_callback_attrs(callback):
    try:
        stopped_epoch = callback.stopped_epoch
        patience = callback.patience
        restore_best_weights = callback.restore_best_weights
        return stopped_epoch, patience, restore_best_weights
    except Exception:
        return None
```

and call it here:

```
callback_attrs = _get_epoch_and_patience(callback)
if callback_attrs is None:
  return
epoch, patience, best_weights = callback_attrs
# Proceed as before, using the callback attributes
```

smurching (commented):
Two comments but LGTM after they're addressed!

tests/keras_autolog/test_keras_autolog.py
```diff
@@ -149,8 +152,148 @@ def test_keras_autolog_logs_default_params(keras_random_data_run):
...
+    assert 'loss' in history.history
+    num_of_epochs = len(history.history['loss'])
+    client = mlflow.tracking.MlflowClient()
+    metric_history = client.get_metric_history(run.info.run_id, 'loss')
+    # Check the test epoch numbers are correct
+    assert num_of_epochs == max(1, callback.patience) + 1
+    # Check that MLflow has logged the metrics of the "best" model
+    assert len(metric_history) == num_of_epochs + 1
+    # Check that MLflow has logged the correct data
+    assert history.history['loss'][callback.stopped_epoch] == metric_history[-1].value
```
harupy: Is this line correct? If early stopping is triggered, metrics at the stopped epoch is NOT the best metrics.

juntai-zheng: Yes, this is incorrect. The test itself wasn't robust - I went ahead and fixed this/refactored the tests so that it would work correctly. Thanks for the catch!

mlflow/keras.py
```diff
@@ -458,93 +478,87 @@ def early_stop_check(callbacks):
                 return callback
         return None
 
-    def log_early_stop(callback, history):
+    def _log_early_stop_callback_params(callback):
         if callback:
-            stopped_epoch = callback.stopped_epoch
-            patience = callback.patience
+            try:
+                params_to_be_logged = {'monitor': callback.monitor,
```
smurching:
```suggestion
                earlystopping_params = {'monitor': callback.monitor,
```

mlflow/keras.py
```diff
@@ -458,93 +478,87 @@ def early_stop_check(callbacks):
...
+                                       'min_delta': callback.min_delta,
+                                       'patience': callback.patience,
+                                       'baseline': callback.baseline,
+                                       'restore_best_weights': callback.restore_best_weights}
+                log_fn_args_as_params(type(callback), [], params_to_be_logged, ['verbose', 'mode'])
```
smurching: Could we just use 
```suggestion
                try_mlflow_log(log_params, earlystopping_params)
```

smurching (approved):
LGTM

PR #2221, opened 2019-12-18
Fix artifact URI constructions in FileStore and SqlAlchemyStore

dbczumar:

Currently, the FileStore and SqlAlchemyStore construct experiment and run artifact root URIs by appending the experiment ID or run artifact path to the store's artifact root URI via the `posixpath.join` function. Unfortunately, this behavior is not aware of URI structures such as queries and fragments. As a result, for an artifact root of `db://root:password@localhost:port/dbname?driver=odbc`, experiment URIs take the form `db://root:password@localhost:port/dbname?driver=odbc/{experiment_id}`, and run artifact URIs take the form `db://root:password@localhost:port/dbname?driver=odbc/{experiment_id}/{run_id}/artifacts`. We see that the experiment and run information is being appended to the *query* component of the URI, rather than the path.

This PR fixes the behavior to correctly append experiment and run details to the *path* component of artifact root URIs. For the case above, the stores will now construct experiment and run artifact URIs as follows:
```
db://root:password@localhost:port/dbname/{experiment_id}?driver=odbc
db://root:password@localhost:port/dbname/{experiment_id}/{run_id}/artifacts?driver=odbc
```

## How is this patch tested?

Thorough unit tests, extensive manual testing

**Manual test plan**
This PR was manually tested by verifying the correct behavior of the MLflow artifact and model . tracking APIs against artifact stores with the following URIs:
- The current working directory (local filesystem)
- A specific local filesystem directory ("/tmp/directory@")
- An Amazon S3 bucket location of the form ("s3://bucket_name/artifact/root/path")
- A Microsoft SQL Server database running in a Docker container with the following URI: `mssql+pyodbc://root:password@localhost:1433/dbname?driver=ODBC+Driver+17+for+SQL+Server`, after merging these changes with a copy of the pending DB artifact storage plugin branch (https://github.com/dbczumar/mlflow/tree/dbartifacts_uri_update).

For each case, I ran the following code (replacing the artifact location with the specified URI):
```
import mlflow
import mlflow.pyfunc
from mlflow.tracking.artifact_utils import _download_artifact_from_uri
from mlflow.tracking.client import MlflowClient
from mlflow.pyfunc import PythonModel

client = MlflowClient()

exp = client.create_experiment("s3exp", artifact_location="s3://artifact-testing-for-mlflow-bucket/subpath/root")


mlflow.set_experiment("s3exp")

with mlflow.start_run():

    class Mod(PythonModel):
        
        def predict(self, ctx, inp):
            return 7

    mlflow.pyfunc.log_model(
        python_model=Mod(),
        artifact_path="mymodel")

    model_uri = "runs:/{run_id}/mymodel".format(run_id=mlflow.active_run().info.run_id)

print(mlflow.pyfunc.load_model(model_uri))

with mlflow.start_run():
    local_path = "/tmp/myartifact.txt"
    with open(local_path, "w") as f:
        f.write("MY FILE")

    mlflow.log_artifact(local_path, "artifact/sublocation")
    artifact_uri = "runs:/{run_id}/artifact/sublocation/myartifact.txt".format(run_id=mlflow.active_run().info.run_id)
    downloaded_location = _download_artifact_from_uri(artifact_uri)
    with open(downloaded_location, "r") as f:
        print(f.read())
```

I then verified that artifacts were written to the appropriate storage locations (e.g., S3, local filesystem, SQL server) and that the script produced correct output of the form:
```
<mlflow.pyfunc.model._PythonModelPyfuncWrapper object at 0x1100051d0>
MY FILE
```

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [X] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/mlflow/utils/uri.py
+++ b/mlflow/utils/uri.py
@@ -67,22 +67,29 @@ def get_uri_scheme(uri_or_path):
         return scheme
 
 
-def append_to_uri_path(uri, path):
+def append_to_uri_path(uri, *paths):
     """
-    Appends the specified posixpath `path` to the path component of the specified `uri`.
+    Appends the specified POSIX path `paths` to the path component of the specified `uri`.
 
-    :param uri: The input URI, represented as a string.    
-    :param path: The posixpath to append to the specified `uri`'s path component.
+    :param uri: The input URI, represented as a string.
+    :param path: The POSIX paths to append to the specified `uri`'s path component.
     :return: A new URI with a path component consisting of the specified `path` appended to
              the path component of the specified `uri`.
 
-    >>> uri = "s3://root/base/path?param=value"
-    >>> uri = append_to_uri_path(uri, "some/subpath")
-    >>> assert uri == "s3://root/base/path/some/subpath?param=value" 
+    >>> uri1 = "s3://root/base/path?param=value"
+    >>> uri1 = append_to_uri_path(uri, "some/subpath", "/anotherpath")
+    >>> assert uri1 == "s3://root/base/path/some/subpath/anotherpath?param=value"
+    >>> uri2 = "a/posix/path"
+    >>> uri2 = append_to_uri_path(uri, "/some", "subpath")
+    >>> assert uri2 == "a/posixpath/some/subpath"
     """
+    path = ""
+    for subpath in paths:
+        path = _join_posixpaths_and_append_absolute_suffixes(path, subpath)
+
     parsed_uri = urllib.parse.urlparse(uri)
     if len(parsed_uri.scheme) == 0:
-        # If the input URI does not define a scheme, we assume that it is a posixpath
+        # If the input URI does not define a scheme, we assume that it is a POSIX path
         # and join it with the specified input path
         return _join_posixpaths_and_append_absolute_suffixes(uri, path)
 
@@ -90,7 +97,7 @@ def append_to_uri_path(uri, path):
     if not parsed_uri.path.startswith("/"):
         # For certain URI schemes (e.g., "file:"), urllib's unparse routine does
         # not preserve the relative URI path component properly. In certain cases,
-        # urlunparse converts relative paths to absolute paths. We introduce this logic 
+        # urlunparse converts relative paths to absolute paths. We introduce this logic
         # to circumvent urlunparse's erroneous conversion
         prefix = parsed_uri.scheme + ":"
         parsed_uri = parsed_uri._replace(scheme="")
@@ -102,7 +109,7 @@ def append_to_uri_path(uri, path):
 
 def _join_posixpaths_and_append_absolute_suffixes(prefix_path, suffix_path):
     """
-    Joins the posixpath `prefix_path` with the posixpath `suffix_path`. Unlike posixpath.join(),
+    Joins the POSIX path `prefix_path` with the POSIX path `suffix_path`. Unlike posixpath.join(),
     if `suffix_path` is an absolute path, it is appended to prefix_path.
 
     >>> result1 = _join_posixpaths_and_append_absolute_suffixes("relpath1", "relpath2")
@@ -119,6 +126,6 @@ def _join_posixpaths_and_append_absolute_suffixes(prefix_path, suffix_path):
 
     # If the specified prefix path is non-empty, we must relativize the suffix path by removing 
     # the leading slash, if present. Otherwise, posixpath.join() would omit the prefix from the
-    # joined path 
+    # joined path
     suffix_path = suffix_path.lstrip(posixpath.sep)
     return posixpath.join(prefix_path, suffix_path)
--- a/mlflow/utils/uri.py
+++ b/mlflow/utils/uri.py
@@ -1,3 +1,4 @@
+import posixpath
 from six.moves import urllib
 
 from mlflow.exceptions import MlflowException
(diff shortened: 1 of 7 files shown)


mlflow/utils/uri.py
```diff
@@ -64,3 +65,67 @@ def get_uri_scheme(uri_or_path):
...
+def append_to_uri_path(uri, *paths):
+    """
+    Appends the specified POSIX `paths` to the path component of the specified `uri`.
+
+    :param uri: The input URI, represented as a string.
+    :param paths: The POSIX paths to append to the specified `uri`'s path component.
+    :return: A new URI with a path component consisting of the specified `paths` appended to
+             the path component of the specified `uri`.
+
+    >>> uri1 = "s3://root/base/path?param=value"
```
sueann: `uri = ` ?

mlflow/utils/uri.py
```diff
@@ -64,3 +65,67 @@ def get_uri_scheme(uri_or_path):
...
+    >>> uri1 = append_to_uri_path(uri, "some/subpath", "/anotherpath")
+    >>> assert uri1 == "s3://root/base/path/some/subpath/anotherpath?param=value"
+    >>> uri2 = "a/posix/path"
```
sueann: `uri = ` ?

mlflow/utils/uri.py
```diff
@@ -64,3 +65,67 @@ def get_uri_scheme(uri_or_path):
...
+    if `suffix_path` is an absolute path, it is appended to prefix_path.
+
+    >>> result1 = _join_posixpaths_and_append_absolute_suffixes("relpath1", "relpath2")
+    >>> assert result1 == "relpath1/relpath2"
+    >>> result2 = _join_posixpaths_and_append_absolute_suffixes("relpath", "/absolutepath")
+    >>> assert result2 == "relpath/absolutepath"
+    >>> result3 = _join_posixpaths_and_append_absolute_suffixes("/absolutepath", "relpath")
+    >>> assert result3 == "/absolutepath/relpath"
+    >>> result4 = _join_posixpaths_and_append_absolute_suffixes("/absolutepath1", "/absolutepath2")
+    >>> assert result4 == "/absolutepath1/absolutepath2"
```
sueann: we should have unit tests for these - can't just trust the docs :-) it is covered by the million test cases here though so if we need to get this in i'm ok with it as is.

tests/utils/test_uri.py
```diff
@@ -47,3 +48,146 @@ def test_uri_types():
...
+        ("path", "", "path/"),
+        ("path", "subpath", "path/subpath"),
+        ("path/", "subpath", "path/subpath"),
+        ("path/", "/subpath", "path/subpath"),
+        ("path", "/subpath", "path/subpath"),
+        ("/path", "/subpath", "/path/subpath"),
+        ("//path", "/subpath", "//path/subpath"),
+        ("///path", "/subpath", "///path/subpath"),
+        ("/path", "/subpath/subdir", "/path/subpath/subdir"),
+        ("file:path", "", "file:path/"),
```
sueann: it's interesting that for 1-level deep paths we return a trailing "/" and for deeper paths we don't. is that intentional?

sueann: and more importantly, are there any discrepancies in downstream behaviors this difference can cause?

sueann (commented):
Looks mostly good to me. A few things:

- Left some inline questions / comments - only 1 blocking question.
- Would you mind adding some details of the manual test in the PR description (especially where/if it tests what unit tests do not cover)?
- Not sure if the test failure is flaky but it looks somewhat real.
- I agree the test cases could be a bit more modularized and de-duped, but the coverage is good so I'm ok merging with the tests as is.

mlflow/utils/uri.py
dbczumar: Good point! Included these here to demonstrate how the function works. The existing unit tests should cover these cases.

dbczumar:
@sueann Thanks for reviewing! I've updated the PR description to include my manual test plan (already executed). Regarding the failing test, the errant case appears to be:
```

**Comment by dbczumar** [comment]
tests/tensorflow/test_tensorflow_model_export.py::test_iris_data_model_can_be_loaded_and_evaluated_as_pyfunc FAILED
```

This is a known flaky test that will hopefully be fixed by @jdlesage in #2223.

tests/utils/test_uri.py
dbczumar: This appears to be caused by the behavior of `posixpath.join()`. I tried the following cases out in Python2 and Python 3:
```
>>> import posixpath
>>> posixpath.join("base", "")
'base/'
>>> posixpath.join("base/", "")
'base/'
>>> posixpath.join("/base", "",)
'/base/'
```

Given that we were previously leveraging `posixpath.join()` for the FileStore and SqlAlchemyStore and that multiple artifact stores rely on it for URI path component construction, this behavior seems consistent with pre-existing behavior and is (hopefully) safe.

mlflow/utils/uri.py
dbczumar: Good catch! Will file a followup PR to fix this, since we're close to branch cut.

mlflow/utils/uri.py
dbczumar: Good catch! Will file a followup PR to fix this, since we're close to branch cut.

sueann (approved):
lgtm - thanks!

PR #2238, opened 2019-12-21
Add autologging functionality for XGBoost flavor

harupy:

Add autologging functionality that logs the following  for XGBoost flavor

- booster parameters
- metrics on each iteration
- feature importance (logged as artifacts in json)
- trained model (logged as an artifact)
- best_score, best_iteration, and best_ntree_limit (if early stopping is set)

Sample code

```python
import xgboost as xgb
from mlflow.xgboost import autolog

from sklearn import datasets

autolog()

X, y = datasets.load_iris(return_X_y=True)

# train model
params = {
    'objective': 'multi:softprob',
    'num_class': 3,
    'eval_metric': 'mlogloss',
    'seed': 42,
}
model = xgb.train(params, dtrain, evals=[(dtest, 'test')], early_stopping_rounds=10)
```

and the output

Dec-27-2019 14-21-38

## How is this patch tested?

Unit tests

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [x] Examples
- [x] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### DIFF
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -75,7 +75,7 @@ def log_fn_args_as_params(fn, args, kwargs, unlogged=[], unpacked=[]):  # pylint
                      in zip(all_param_names[:len(args)], args)
                      if param_name not in unlogged)
 
-    # Unpack dict parameters if unpack is not empty.
+    # Unpack dict parameters
     for param_name in unpacked:
         popped = args_dict.pop(param_name)
         args_dict = {**args_dict, **popped}
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -46,13 +46,14 @@ def get_unspecified_default_args(user_args, user_kwargs, all_param_names, all_de
             if name not in user_specified_arg_names}
 
 
-def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W0102
+def log_fn_args_as_params(fn, args, kwargs, unlogged=[], unpacked=[]):  # pylint: disable=W0102
     """
     Log parameters explicitly passed to a function.
     :param fn: function whose parameters are to be logged
     :param args: arguments explicitly passed into fn
     :param kwargs: kwargs explicitly passed into fn
     :param unlogged: parameters not to be logged
+    :param unpacked: dict parameters to be unpacked
     :return: None
     """
     # all_default_values has length n, corresponding to values of the
@@ -73,10 +74,21 @@ def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W01
     args_dict = dict((param_name, param_val) for param_name, param_val
                      in zip(all_param_names[:len(args)], args)
                      if param_name not in unlogged)
+
+    # Unpack dict parameters if unpack is not empty.
+    for param_name in unpacked:
+        popped = args_dict.pop(param_name)
+        args_dict = {**args_dict, **popped}
+
     if len(args_dict.keys()) > 0:
         try_mlflow_log(mlflow.log_params, args_dict)
 
     # Logging the kwargs passed by the user
     for param_name in kwargs:
-        if param_name not in unlogged:
+        if param_name in unlogged:
+            continue
+
+        if param_name in unpacked:
+            try_mlflow_log(mlflow.log_params, kwargs[param_name])
+        else:
             try_mlflow_log(mlflow.log_param, param_name, kwargs[param_name])
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -46,14 +46,13 @@ def get_unspecified_default_args(user_args, user_kwargs, all_param_names, all_de
             if name not in user_specified_arg_names}
 
 
-def log_fn_args_as_params(fn, args, kwargs, unlogged=[], unpacked=[]):  # pylint: disable=W0102
+def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W0102
     """
     Log parameters explicitly passed to a function.
     :param fn: function whose parameters are to be logged
     :param args: arguments explicitly passed into fn
     :param kwargs: kwargs explicitly passed into fn
     :param unlogged: parameters not to be logged
-    :param unpacked: dict parameters to be unpacked
     :return: None
     """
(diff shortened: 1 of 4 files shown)


mlflow/utils/autologging_utils.py
```diff
@@ -46,13 +46,14 @@ def get_unspecified_default_args(user_args, user_kwargs, all_param_names, all_de
...
 
-def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W0102
+def log_fn_args_as_params(fn, args, kwargs, unlogged=[], unpacked=[]):  # pylint: disable=W0102
     """
     Log parameters explicitly passed to a function.
     :param fn: function whose parameters are to be logged
     :param args: arguments explicitly passed into fn
     :param kwargs: kwargs explicitly passed into fn
     :param unlogged: parameters not to be logged
+    :param unpacked: dict parameters to be unpacked
```
smurching: Maybe clarify this comment, seems like the idea of this parameter is to allow special handling of function arguments that correspond to dictionaries of params (e.g. `{"learning_rate": 0.01"}`)? Maybe we should just exclude such params via `unlogged` & log them separately for now, since AFAICT the XGBoost autologging functionality is the only one with this use case at the moment?

mlflow/xgboost.py
```diff
@@ -184,3 +192,89 @@ def __init__(self, xgb_model):
...
+
+    """
+    import xgboost
+
+    @gorilla.patch(xgboost)
+    def train(*args, **kwargs):
+
+        def mlflow_callback(env):
+            """
+            Callback for auto-logging metrics on each iteration.
```
smurching: How quick are the XGBoost iterations? We may end up wanting to log these metrics asynchronously if there are many iterations so that the API request overhead doesn't slow down training, as in [TensorFlow autologging](https://github.com/mlflow/mlflow/blob/master/mlflow/tensorflow.py#L53). However we could add the asynchronous logging in a follow-up & start with something synchronous

mlflow/xgboost.py
```diff
@@ -184,3 +192,89 @@ def __init__(self, xgb_model):
...
+        # even if early stopping didn't occur.
+        if early_stopping:
+            attrs = ['best_score', 'best_iteration', 'best_ntree_limit']
+            try_mlflow_log(mlflow.log_metrics, {attr: getattr(model, attr) for attr in attrs})
+
+        # logging feature importance as artifacts.
+        for imp_type in importance_types:
+            imp = model.get_score(importance_type=imp_type)
+            filename = 'feature_importance_{}.json'.format(imp_type)
+            filepath = os.path.join(tempfile.mkdtemp(), filename)
```
smurching: We should make sure to remove this tempdir once we log our artifact

mlflow/xgboost.py
```diff
@@ -184,3 +192,89 @@ def __init__(self, xgb_model):
...
+            kwargs['callbacks'] = [mlflow_callback]
+
+        model = original(*args, **kwargs)
+
+        # checking if the 'early_stopping_rounds' argument of train() is present.
+        early_stopping_index = all_arg_names.index('early_stopping_rounds')
+        early_stopping = (num_pos_args >= early_stopping_index + 1 or
+                          'early_stopping_rounds' in kwargs)
+
+        # if 'early_stopping_rounds' is present, the output model has
```
smurching: With other autologging integrations, we handled early stopping by logging an extra metric with the max step + 1 at the end of training, rather than logging separate `best_<metric-name>` metrics, e.g. see https://github.com/mlflow/mlflow/pull/2219 for Keras autologging. Ideally the integrations would behave similarly, but I'm curious if you found the `best_<metric-name>` format to be preferable?

mlflow/utils/autologging_utils.py
harupy: Agree with excluding dict params via `unlogged` and logging them separately. I'll revert this change.

mlflow/xgboost.py
harupy: will fix this

mlflow/xgboost.py
harupy: The iterations speed depends on the # of records and features. I'll take care of the asynchronous logging in a follow-up PR if necessary.

mlflow/xgboost.py
harupy: Either way would be fine for me. I'll go with logging an extra metric for other autologging integrations.

harupy:
I fixed the code to log metrics after the training finishes.

mlflow/xgboost.py
```diff
@@ -184,3 +193,102 @@ def __init__(self, xgb_model):
     def predict(self, dataframe):
         import xgboost as xgb
         return self.xgb_model.predict(xgb.DMatrix(dataframe))
+
+
+@experimental
+def autolog(importance_types=['weight']):  # pylint: disable=W0102
+    """
+    Enables autologging from XGBoost to MLflow. Logs the following.
```
smurching: Would be good to mention that this `autolog` API only supports the XGBoost Booster API

smurching (commented):
The code changes look good! 

Trying this out locally now - my only other comments are that we should mention XGBoost autologging in the main tracking docs (https://github.com/mlflow/mlflow/blob/master/docs/source/tracking.rst). In particular, if you're able to restructure the many different autologging sections in `tracking.rst` so that we have a single-top level section for "Automatic Logging" and create sub-sections for "TensorFlow", "Gluon", "Xgboost" that'd be awesome. 

LGTM after that's addressed. Also, after the next release, we should also update the XGBoost example under `examples/xgboost` to use the new autologging functionality :)

smurching:
Oh also, just tried this locally - I see (as the docs specify) that no metrics are logged unless `evals` is specified - is it possible to extract & log the value of the objective function (e.g. the model loss) if the user doesn't pass `evals`? No worries if not was just curious. If it is possible to autolog model loss/accuracy, we can add such functionality in a follow-up PR.

harupy:
@smurching 
XGBoost doesn't seem to provide methods or attributes to get the model loss.

harupy:
@smurching Thanks for the comments :) I think it's ready for another review.

smurching (approved):
Pushed some small doc edits, LGTM! Will merge once tests pass

harupy:
Thanks!

PR #2241, opened 2019-12-26
Add hdfs support for path parameters

t-henri:

In a MLproject file, when a parameter has a path type and is an hdfs uri, the artifact will be downloaded from hdfs.

## How is this patch tested?

With unit test and by creating an MLflow project that tries to download a file from hdfs.

Added support of hdfs for path parameters in MLflow project.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [X] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/data.py
+++ b/mlflow/data.py
@@ -6,13 +6,18 @@ import re
 from six.moves import urllib
 
 from mlflow.utils import process
+from mlflow.store.artifact.hdfs_artifact_repo import HdfsArtifactRepository
 
 DBFS_PREFIX = "dbfs:/"
 S3_PREFIX = "s3://"
 GS_PREFIX = "gs://"
+VIEWFS_PREFIX = "viewfs://"
+HDFS_PREFIX = "hdfs://"
 DBFS_REGEX = re.compile("^%s" % re.escape(DBFS_PREFIX))
 S3_REGEX = re.compile("^%s" % re.escape(S3_PREFIX))
 GS_REGEX = re.compile("^%s" % re.escape(GS_PREFIX))
+VIEWFS_REGEX = re.compile("^%s" % re.escape(VIEWFS_PREFIX))
+HDFS_REGEX = re.compile("^%s" % re.escape(HDFS_PREFIX))
 
 
 class DownloadException(Exception):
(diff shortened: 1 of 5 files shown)


mlflow/data.py
```diff
@@ -27,33 +33,30 @@ def _fetch_dbfs(uri, local_path):
...
-    if parsed.scheme != "s3":
-        raise Exception("Not an S3 URI: %s" % uri)
-    path = parsed.path
-    if path.startswith('/'):
-        path = path[1:]
-    return parsed.netloc, path
+def _fetch_hdfs(uri, local_path):
+    print("=== Downloading HDFS file %s to local path %s ===" % (uri, os.path.abspath(local_path)))
+    parse_simple_uri(uri, ["hdfs", "viewfs"])
+    host, port, path = _resolve_connection_params(uri)
```
mparkhe: This uses private functions from `hdfs_artifact_repo`. Can we rewrite this use `HdfsArtifactRepository` directly?

mlflow/data.py
```diff
@@ -27,33 +33,30 @@ def _fetch_dbfs(uri, local_path):
...
+    with hdfs_system(host=host, port=port) as hdfs:
+        local_path = os.path.join(local_path, os.path.normpath(uri.split('/')[-1]))
```
mparkhe: How does this handle URI paths like `"hdfs:/this/is/a/test/.."`? In this case `local_path` will have a trailing `/..` appended to it. Would help to add some unit tests for this

tests/data/test_data.py
```diff
@@ -31,14 +31,18 @@ def test_is_uri():
     assert is_uri("gs://some/gs/path")
     assert is_uri("dbfs:/some/dbfs/path")
     assert is_uri("file://some/local/path")
+    assert is_uri("hdfs://some/hdfs/path")
+    assert is_uri("viewfs://some/viewfs/path")
     assert not is_uri("/tmp/some/local/path")
```
mparkhe: Reminder: Add unit tests for error cases

mparkhe (changes requested):
Some requested changes. Comments inlined...

mlflow/data.py
t-henri: I made a change to use HdfsArtifactRepository directly, but I had to use a private method of the class that I implemented: _download_file.

I'm not sure it is what you meant with your comment, but I couldn't use the already existing method download_artifacts since it always create a temporary directory where the file will be downloaded, so the file wouldn't be downloaded in the output_path provided in data.py.

What I could do instead of this is that the method download_uri in data.py returns a new output path that we could get in _project_spec.py, so I could call download_artifacts and return the new path.

mlflow/data.py
t-henri: Good catch, I made the change and added unit tests

tests/data/test_data.py
t-henri: I'm not sure what test cases you want me to add here, could you give me more indications please ?

t-henri:
Hello @mparkhe, thanks for your review. I adressed some of the comments and answered some others.

mlflow/store/artifact/hdfs_artifact_repo.py
```diff
@@ -145,7 +144,12 @@ def download_artifacts(self, artifact_path, dst_path=None):
             return local_dir
 
     def _download_file(self, remote_file_path, local_path):
-        raise MlflowException('This is not implemented. Should never be called.')
+        _, _, path = _resolve_connection_params(remote_file_path)
+        if path.endswith('/'):
+            path = path[:-1]
+        with hdfs_system(host=self.host, port=self.port) as hdfs:
+            local_path = os.path.join(local_path, os.path.normpath(path.split('/')[-1]))
+            _download_hdfs_file(hdfs, path, local_path)
```
mparkhe: downloading a file is already included in `download_artifacts` function. Is this doing any special handling other than handling the trailing `/`. Do you need to do that?

Maybe we can revert this functionality and make `download_artifacts` work for all cases.

mlflow/data.py
```diff
@@ -27,33 +32,28 @@ def _fetch_dbfs(uri, local_path):
...
-        raise Exception("Not an S3 URI: %s" % uri)
-    path = parsed.path
-    if path.startswith('/'):
-        path = path[1:]
-    return parsed.netloc, path
+def _fetch_hdfs(uri, local_path):
+    print("=== Downloading HDFS file %s to local path %s ===" % (uri, os.path.abspath(local_path)))
+    parse_simple_uri(uri, ["hdfs", "viewfs"])
+    hdfs_repo = HdfsArtifactRepository(uri)
+    hdfs_repo._download_file(uri, local_path)
```
mparkhe: Can you use

```
  hdfs_repo.download_artifacts(uri, local_path)
```

If you need to special handling for the trailing `/` can you do it here in the before calling `download_artifacts`?

tests/store/artifact/test_hdfs_artifact_repo.py
```diff
@@ -191,3 +191,18 @@ def test_download_artifacts():
                             os.path.join(tmp_dir.path(), artifact_path))
         with open(os.path.join(tmp_dir.path(), artifact_path), "rb") as fd:
             assert expected_data == fd.read()
+
+
+@mock.patch('pyarrow.hdfs.HadoopFileSystem', spec=HadoopFileSystem)
+def test_download_file(hdfs_system_mock):
+    expected_data = b"hello"
+    artifact_paths = ["hdfs://host/test.txt", "hdfs://host/test.txt/"]
```
mparkhe: Both the elements in the list are identical. Did you intend this to be --

```
artifact_paths = ["hdfs://host/test.txt", "viewfs://host/test.txt/"]
```

tests/store/artifact/test_hdfs_artifact_repo.py
```diff
@@ -191,3 +191,18 @@ def test_download_artifacts():
...
+    hdfs_system_mock.return_value.open = mock_open(read_data=expected_data)
+
+    for artifact_path in artifact_paths:
+        hdfs_repo = HdfsArtifactRepository(artifact_path)
+        with TempDir() as tmp_dir:
+            hdfs_repo._download_file(artifact_path,
```
mparkhe: use `download_artifacts` instead of `_download_file`.

in fact, we are missing a test that `_download_file` raises an error

Issue #500, opened 2018-09-16
Feature request: Support for SpaCy models


suzil:
It would be nice if there was a `mlflow.spacy` module for SpaCy models. [Here is an example usage of training a SpaCy text classification model.](https://spacy.io/usage/training#textcat)

The two relevant functions for creating the module would be `nlp = spacy.load(model_path)` and `nlp.to_disk(output_path)`. To make a prediction, you would run something like `nlp("Here is some example text to classify").cats['CATEGORY']`.

I think this should be easy to implement? The part that I'm not so sure about is how to write the prediction part.

juntai-zheng:
SpaCy support has been added in! Closing this issue now.

PR #2242, opened 2019-12-26
Add support for spacy flavor

arocketman:

References #500 , allows to use spacy as a flavor and log its models using mlflow. 

## How is this patch tested?

Taking inspirations from other flavours, several unit tests have been created in test_spacy_model_export.py, moreover an integration test has been done to ensure the correct work and integration with mlflow ui.

image

A full blown example is also present in the examples directory.

Simple usage:
```python
    mlflow.spacy.log_model(spacy_model=nlp, artifact_path='model')
    model_uri = "runs:/{run_id}/{artifact_path}".format(
        run_id=mlflow.active_run().info.run_id,
        artifact_path='model')

    # Load model
    nlp2 = mlflow.spacy.load_model(model_uri=model_uri)
```

* Created spacy.py module to save/load the model using SpaCy
* Created unit tests test_spacy_model_export
* Created an example on usage in /examples

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [x] Examples
- [X] Docs
- [X] Tracking
- [ ] Projects
- [X] Artifacts
- [X] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


arocketman:
Build is currently failing for python 2.7 while it passes for all the other steps. It's something related to the actual tests and not the module itself. I will look into in the next days.

travis/run-large-python-tests.sh
```diff
@@ -21,7 +21,7 @@ pytest tests --large --ignore=tests/examples --ignore=tests/h2o --ignore=tests/k
   --ignore=tests/spark --ignore=tests/tensorflow --ignore=tests/azureml --ignore=tests/onnx \
   --ignore=tests/keras_autolog --ignore=tests/tensorflow_autolog --ignore=tests/gluon \
   --ignore=tests/gluon_autolog --ignore=tests/xgboost --ignore=tests/lightgbm \
-  --ignore tests/spark_autologging --ignore=tests/models
+  --ignore=tests/spacy --ignore=tests/spark_autologging --ignore=tests/models
```
arocketman: It seems to me like an "=" is missing here in the master's branch. Adding it made the Travis build pass. @smurching

travis/run-large-python-tests.sh
```diff
@@ -21,7 +21,7 @@ pytest tests --large --ignore=tests/examples --ignore=tests/h2o --ignore=tests/k
   --ignore=tests/spark --ignore=tests/tensorflow --ignore=tests/azureml --ignore=tests/onnx \
   --ignore=tests/keras_autolog --ignore=tests/tensorflow_autolog --ignore=tests/gluon \
   --ignore=tests/gluon_autolog --ignore=tests/xgboost --ignore=tests/lightgbm \
-  --ignore tests/spark_autologging --ignore=tests/models
```
arocketman: It seems to me like an "=" is missing here in the master's branch. Adding it made the Travis build pass. @smurching

dbczumar:
@arocketman Thank you for your contribution and so sorry for the delay! We'll find a reviewer for this asap (either @smurching or another MLflow reviewer).

mlflow/spacy.py
```diff
@@ -0,0 +1,189 @@
...
+    return spacy.load(path)
+
+
+class _SpacyModelWrapper:
+    def __init__(self, spacy_model):
+        self.spacy_model = spacy_model
+
+    def predict(self, dataframe):
+        """
+        Only works for predicting using text categorizer.
```
smurching: I'm not too familiar with Spacy, but it seems it'd be worth documenting this limitation in the `save_model` & `log_model` docstrings? Seems like a good idea to keep the functionality here minimal in the initial PR & expand upon it later, but do you have an idea what it'd take to extend it to supporting other types of Spacy pipeline components in the future?

mlflow/spacy.py
```diff
@@ -0,0 +1,189 @@
...
+        Not suitable for other pipeline components (e.g: parser)
+        :param dataframe: dataframe containing texts to be categorized
+        :return: dataframe with predictions
+        """
+        return pd.DataFrame({
+            'predictions': dataframe.ix[:, 0].apply(lambda text: self.spacy_model(text).cats)
```
smurching: Should we check the dimensions of the input here & throw an `MlflowException` if it's invalid? Would be nice to make our assumptions about the input shape here (that it contains a single column) clear to the end user.

tests/spacy/test_spacy_model_export.py
```diff
@@ -0,0 +1,220 @@
...
+@pytest.mark.large
+def test_model_log(spacy_model_with_data):
+    spacy_model = spacy_model_with_data.model
+    old_uri = mlflow.get_tracking_uri()
+    # should_start_run tests whether or not calling log_model() automatically starts a run.
+    for should_start_run in [False, True]:
+        with TempDir(chdr=True, remove_on_exit=True):
+            try:
+                artifact_path = "model"
+                mlflow.set_tracking_uri("test")
```
smurching: You should be able to use the `tracking_uri_mock` fixture from tests.projects.utils to avoid having to explicitly mock out the tracking URI here, see e.g. https://github.com/mlflow/mlflow/blob/360c0cada0174b71500a83ded72260b6b6942d90/tests/sklearn/test_sklearn_model_export.py#L159

smurching (commented):
Left some initial clarifying questions/comments, thanks @arocketman for the PR!

examples/spacy/conda.yaml
```diff
@@ -0,0 +1,10 @@
+name: spacy-example
+channels:
+  - defaults
+  - anaconda
+dependencies:
+  - python==3.6
```
harupy: `python=3.6`?

harupy: It seems both `=` and `==` work.

harupy: `conda env export` uses `=` for conda dependencies and `==` for pip dependencies.

```
$ conda env export
name: mlflow-dev-env
channels:
  - conda-forge
  - defaults
dependencies:
  - _py-xgboost-mutex=2.0=cpu_0
  - blas=1.0=mkl
  - ca-certificates=2019.11.28=hecc5488_0
  - catboost=0.20.1=py36_0
  ...
  - pip:
    - absl-py==0.8.1
    - adal==1.2.2
    - alabaster==0.7.12
    - alembic==1.2.1
    ...
```

tests/spacy/test_spacy_model_export.py
arocketman: Thanks for the the tip. I have implemented it in the latest commit.

mlflow/spacy.py
arocketman: Added check before prediction and some shape info in the docstrings for the dataframe parameter.

mlflow/spacy.py
arocketman: It depends on how "flexible" we want to keep the predict pyfunc method. Right now I have used it only as the "Text classifier/categorizer", however when the Language/nlp object is called (see https://spacy.io/usage/spacy-101#annotations) a Doc is created. Then you can use these Doc attributes to get the relevant information.

I would say that the text categorizer predict is the one that fits more the pyfunc predict. However the loaded model can be called and generate the Doc object at anytime. I would say **the Saved and Loaded model has no limitations. Only the predict pyfunc api works for the text categorizer.**

If you wish I can add it also to save_model and log_model, however I worry that might be misleading.

examples/spacy/conda.yaml
arocketman: Would you like me to switch it to '=' ? I was taking inspiration from the examples/sklearn_logistic_regression where also == is used. I can change it if you guys prefer it that way.

tests/spacy/test_spacy_model_export.py
arocketman: For some reason travis CI is giving me the following error: 

tests/spacy/test_spacy_model_export.py (91,42): [W0613 unused-argument] Unused argument 'tracking_uri_mock'

even tho I explicitly disabled the warning for pylint:

https://github.com/mlflow/mlflow/blob/dcc0b0930daf4661d68a89761fc619dfa5cd3a30/tests/spacy/test_spacy_model_export.py#L21

Any ideas?

examples/spacy/conda.yaml
harupy: Sorry for the confusion. `==` is fine.

.travis.yml
```diff
@@ -101,7 +101,7 @@ matrix:
         - pip install -r travis/small-requirements.txt
         - pip install -e .
       script:
-        - pytest --verbose --ignore=tests/h2o --ignore=tests/keras --ignore=tests/pytorch --ignore=tests/pyfunc --ignore=tests/sagemaker --ignore=tests/sklearn  --ignore=tests/spark --ignore=tests/tensorflow --ignore=tests/keras_autolog --ignore=tests/tensorflow_autolog --ignore tests/azureml --ignore tests/onnx --ignore tests/projects --ignore=tests/xgboost --ignore=tests/lightgbm --ignore=tests/spark_autologging tests
+        - pytest --verbose --ignore=tests/h2o --ignore=tests/keras --ignore=tests/pytorch --ignore=tests/pyfunc --ignore=tests/sagemaker --ignore=tests/sklearn  --ignore=tests/spark --ignore=tests/tensorflow --ignore=tests/keras_autolog --ignore=tests/tensorflow_autolog --ignore tests/azureml --ignore tests/onnx --ignore tests/projects --ignore=tests/xgboost --ignore=tests/lightgbm --ignore=tests/spark_autologging tests --ignore=tests/spacy tests
```
harupy: There are two `tests` in this line. One of them should be removed as follows, right?

```diff
- ... --ignore=tests/spark_autologging tests --ignore=tests/spacy tests
+ ... --ignore=tests/spark_autologging --ignore=tests/spacy tests
```

arocketman: Yes, good catch. Pushed the fix. Thanks

arocketman:
I implemented few changes suggested here. Let me know how those looks. Thanks

mlflow/spacy.py
```diff
@@ -0,0 +1,194 @@
+"""
+The ``mlflow.spacy`` module provides an API for logging and loading SpaCy models.
```
harupy: nit: `SpaCy` -> `spaCy`?

examples/spacy/train.py
```diff
@@ -0,0 +1,65 @@
...
+    # add labels
+    for _, annotations in TRAIN_DATA:
+        for ent in annotations.get("entities"):
+            ner.add_label(ent[2])
+
+    metrics = {
+        'n_iter':100,
+        'drop': 0.5
+    }
+    mlflow.log_params(metrics)
```
harupy: `metrics` -> `params`?

examples/spacy/train.py
```diff
@@ -0,0 +1,65 @@
...
+                losses=losses,
+            )
+        print("Losses", losses)
+        mlflow.log_metrics(losses)
+
+    # load it again (just as an example) and log metrics
+    mlflow.spacy.log_model(spacy_model=nlp, artifact_path='model')
+    model_uri = "runs:/{run_id}/{artifact_path}".format(
+        run_id=mlflow.active_run().info.run_id,
+        artifact_path='model')
```
harupy: Is the comment here correct?

examples/spacy/train.py
arocketman: Done, thanks

mlflow/spacy.py
arocketman: Done!

examples/spacy/train.py
arocketman: It was intended for some of the lines below (starting from 52), but I see how that could be a little misleading. Changed them a little in the latest commit.
Thanks

mlflow/spacy.py
harupy: The code still contains `spacy`, `Spacy`, and `SpaCy`.  They should be changed to `spaCy`.

arocketman: Thanks, it seems like these casing things are haunting me :) . Hopefully got them all this time.

mlflow/spacy.py
```diff
@@ -0,0 +1,194 @@
+"""
+The ``mlflow.spacy`` module provides an API for logging and loading spaCy models.
+This module exports spacy models with the following flavors:
+
+Spacy (native) format
```
harupy: spaCy

mlflow/spacy.py
```diff
@@ -0,0 +1,194 @@
...
+        additional_conda_deps=None,
+        additional_pip_deps=[
+            "spacy=={}".format(spacy.__version__),
+        ],
+        additional_conda_channels=None)
+
+
+def save_model(spacy_model, path, conda_env=None, mlflow_model=Model()):
+    """
+    Save a spacy model to a path on the local file system.
```
harupy: spaCy

mlflow/spacy.py
```diff
@@ -0,0 +1,194 @@
...
+    Load PyFunc implementation. Called by ``pyfunc.load_pyfunc``.
+
+    :param path: Local filesystem path to the MLflow Model with the ``spacy`` flavor.
+    """
+    return _SpacyModelWrapper(_load_model(path))
+
+
+def load_model(model_uri):
+    """
+    Load a SpaCy model from a local file (if ``run_id`` is ``None``) or a run.
```
harupy: spaCy

arocketman: Done

mlflow/spacy.py
arocketman: Done

mlflow/spacy.py
arocketman: This one I changed just the one on line 5 and kept the "mlflow.spacy" as is as it references the module itself. If that makes sense to you

harupy: Yes, `mlflow.spacy` should be kept as is.

arocketman:
Is anything more needed from my side for this PR? @harupy @smurching

smurching:
@arocketman sorry for the delay, haven't quite been able to make time to familiarize myself with the spaCy APIs, but can do so & take another pass. @harupy are you more familiar with spaCy by any chance / would you be able to take an initial pass?

mlflow/spacy.py
smurching: Got it, so if I understand correctly:
* Users can save & load back spaCy models without any limitations (makes sense, since `mlflow.spacy.load_model` simply loads back the original `spaCy` model object)
* However, in order to use the saved `spaCy` model with tools dependent on the python-function (pyfunc) flavor, the saved model must be a text categorizer (one thing I'd appreciate help understanding is if most common spaCy use cases only involve a text categorizer, or if this limitation would affect many spaCy use cases). 

My intuition here is that it would be strange to allow saving/logging spaCy models that can't be used with pyfunc predict.

smurching (commented):
Some more comments/questions that I'm hoping will help me better understand spaCy to help wrap up this PR - thanks @arocketman for your patience!

mlflow/spacy.py
arocketman: Yes, I think your sum up is correct. I think the text categorizer is not the only use case of spacy, as a matter of fact the different components of the pipeline are used. However, If I understood correctly the way the pyfunc predict method works here, it usually appends a column with the "prediction" at the end of the input dataframe, and I have found that among the different spaCy pipeline components the text categorizer would fit the description.

If we want, we can try to work around the different components and find a format for what spaCy is returning when the nlp object is called.
The table here: https://spacy.io/usage/spacy-101#pipelines explains fairly well how a different pipeline component is mapped to the object it creates.

My idea was to start from the text categorizer and build up from there as new use cases arise for the pyfunc.

arocketman:
Hi @smurching and @harupy , I had to fix some pylinting issues and an import (tracking_uri_mock) as it was recently moved to conftest.py. 

As this PR has been open for a while (Dec 2019) and the master keeps drifting from my fork, I would like to understand with you if we want to proceed with merging this. 
Otherwise I'll close it as it is getting stale.

Thanks!

mlflow/spacy.py
smurching: @arocketman got it, makes sense that it'd be nice to incrementally add support for additional pipeline components.  Could we check via `model.pipelines` that the last component of the persisted model is an instance of `spacy.pipeline.pipes.TextCategorizer`, and only generate the pyfunc flavor if that's the case? 

If the final pipeline component isn't a TextCategorizer, we can log a warning like:

```
WARNING: Generating only the spacy flavor for the provided spacy model. This means the model can be loaded back via `mlflow.spacy.load_model`, but cannot be loaded back using pyfunc APIs like `mlflow.pyfunc.load_model` or via the `mlflow models` CLI commands. MLflow will only generate the pyfunc flavor for spacy models whose final pipeline component is an instance of spacy.pipeline.pipes.TextCategorizer.
```

I think such a change would result in nicer behavior - we'll only generate a pyfunc flavor for Spacy models that can be served as pyfuncs (users won't hit issues where they attempt to run `mlflow models serve /path/to/saved/spacy/model` and the model errors out at inference time because it's not a text classifier). We can then later add support for serving other types of spacy models as pyfuncs later on.

smurching (commented):
@arocketman apologies again for the delay & thanks for the `spacy` API doc links. Just had one more suggestion around only generating the `pyfunc` flavor for `spacy` models that we can serve as `pyfunc`s (we should also add a test verifying that the `pyfunc` flavor is not generated if the user calls `mlflow.spacy.log_model` on a pipeline that does not end in a `TextCategorizer`, e.g. something like:

```
from mlflow.models import Model
spacy_model = spacy.blank("en") # create and log anything other than a TextCategorizer
mlflow.spacy.log_model(spacy_model, "/tmp/my-path")
loaded_model = Model.load("/tmp/my-path")
assert set(loaded_model.flavors.keys()) == set("spacy")
```

Would also be nice to update https://github.com/mlflow/mlflow/blob/master/docs/source/tutorials-and-examples/index.rst to reference your new spaCy - thanks for the continued work on this, LGTM otherwise!

arocketman:
hi @smurching , with the latest commit I have added two tests, one is exactly the one you requested, the other one is related to the presence of the pyfunc flavor in case the text categorizer is present.
I also updated the tutorials-and-examples.

Let me know if this is all good.

smurching:
Hi @arocketman, thanks for the updates the changes look good! I meant to push some small tweaks to log a warning in the case where a pyfunc model could not be saved & improve test runtime by reusing the fitted spacy model across the whole test module. Somehow in doing so I managed to make this PR an empty diff & cause github to close it.

I've opened https://github.com/mlflow/mlflow/pull/2242 to reintroduce the code from this PR into your master branch, apologies for the hassle / not sure how that happened - thanks again for your contribution!!

smurching:
In particular, I expect that once https://github.com/arocketman/mlflow/pull/1 is merged, the appropriate diff for this PR will show up again, and we can merge - thanks!

arocketman:
Hi again @smurching , I merged your PR into my master branch. I can see it here but the PR still is closed. Can you check if you can re-open it? Otherwise, I can try to re-do the PR and reference this one from there.

Thanks!

smurching (approved):
LGTM

smurching:
Awesome, tests passed - merged, thanks for all the hard work on this @arocketman :)!

arocketman:
Great! Thanks and have a good one :)

examples/spacy/train.py
```diff
@@ -0,0 +1,65 @@
...
+        for batch in batches:
+            texts, annotations = zip(*batch)
+            nlp.update(
+                texts,  # batch of texts
+                annotations,  # batch of annotations
+                drop=params['drop'],  # dropout - make it harder to memorise data
+                losses=losses,
+            )
+        print("Losses", losses)
+        mlflow.log_metrics(losses)
```
harupy, line 50:
```
mlflow.log_metrics(losses, step=itn)
```
Add `step` here?

harupy: Sorry for the late reply after this PR is merged.

smurching: @harupy no worries! That change makes sense, happy to review a follow-up PR that makes the change

louisguitton:
@smurching I see you're releasing quite often lately (every 2 to 4 weeks) (and congrats for that 🎉 ).
Do you know if that PR would be part of 1.7.3 ?

examples/spacy/train.py
arocketman: Hi, sure thing I'll make a follow-up PR with this small change ASAP. Thanks!

PR #2246, opened 2019-12-27
Solve race condition for directory creation

avaucher:

This PR solves a race condition when creating the directory `mlruns` may cause Python programs running simultaneously to fail.

Concretely, the following error came up several times when a job scheduler launched several runs at simultaneously:
```
FileExistsError: [Errno 17] File exists: './mlruns'
```
Upon analysis, when trying to create the `mlruns` directory, [here](https://github.com/mlflow/mlflow/blob/e6764eb63dd66e8a809dcd422d547d091b74549a/mlflow/utils/file_utils.py#L102), 
```python
        if not exists(target):
            os.makedirs(target)
```
it looks like the filesystem was fast to check for the existence of the directory but slow to create it, and therefore all except one simultaneous runs failed on `os.makedirs()`.

To avoid this race condition, and be consistent the EAFP coding style (Easier to Ask for Forgiveness than Permission), this PR does not check for existence of the directory but inspects the exception, if any.

See https://stackoverflow.com/a/14364249 for a detailed discussion on StackOverflow.

## How is this patch tested?

-

### DIFF
--- a/mlflow/utils/file_utils.py
+++ b/mlflow/utils/file_utils.py
@@ -1,4 +1,5 @@
 import codecs
+import errno
 import gzip
 import os
 import posixpath
@@ -99,11 +100,11 @@ def mkdir(root, name=None):  # noqa
     """
     target = os.path.join(root, name) if name is not None else root
     try:
-        if not exists(target):
-            os.makedirs(target)
-            return target
+        os.makedirs(target)
     except OSError as e:
-        raise e
+        if e.errno != errno.EEXIST:
+            raise e
+    return target
 
 
 def make_containing_dirs(path):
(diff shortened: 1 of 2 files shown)


mlflow/utils/file_utils.py
```diff
@@ -99,11 +100,11 @@ def mkdir(root, name=None):  # noqa
     """
     target = os.path.join(root, name) if name is not None else root
     try:
-        if not exists(target):
-            os.makedirs(target)
-            return target
+        os.makedirs(target)
     except OSError as e:
-        raise e
+        if e.errno != errno.EEXIST:
```
mparkhe: This is a good change! Can we add one more check here that `target`, if it exists is a directory? If not, it should throw...
Also please could you add a unit test that throws if `target` is a file

avaucher: Updated!

mparkhe (approved):
Thanks! This looks good.

PR #2251, opened 2019-12-30
New runs table column view

Zangr:

This PR implements a new runs table column view based on `ag-grid` with a series of features as listed below:
- Nested runs
- Serverside sorting
- Load more runs
- Column custom selection with a checkable tree dropdown
- Column resizing
- Column reordering
- Column pinning
- Column highlighting
- Column toggling
- Row selection batching with `Shift + Click`
- Cell `-` placeholder
- Table loading overlay
- Wire up the old compact view with the column selection tree dropdown

new-multi-column-view

## How is this patch tested?
Manual, Unit tests.

This PR implements a new runs table column view based on `ag-grid` with a series of features 

- [x] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
+++ b/mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
@@ -26,7 +26,6 @@ export class RunsTableColumnSelectionDropdown extends React.Component {
   };
 
   handleCheck = (checkedKeys, allKeys) => {
-    console.log('checkedKeys', checkedKeys);
     const { onCheck } = this.props;
     if (onCheck) {
       onCheck(getCategorizedUncheckedKeys(checkedKeys, allKeys));
--- a/mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
+++ b/mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
@@ -1,0 +1,153 @@
+import React from 'react';
+import { Button, Dropdown, Icon, Menu } from 'antd';
+import { SearchTree } from '../common/components/SearchTree';
+import { ColumnTypes } from '../common/utils';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+import ExperimentViewUtil from './ExperimentViewUtil';
+
+export class RunsTableColumnSelectionDropdown extends React.Component {
+  static propTypes = {
+    paramKeyList: PropTypes.array,
+    metricKeyList: PropTypes.array,
+    visibleTagKeyList: PropTypes.array,
+    onCheck: PropTypes.func,
+    categorizedUncheckedKeys: PropTypes.object.isRequired,
+  };
+
+  static defaultProps = {
+    paramKeyList: [],
+    metricKeyList: [],
+    visibleTagKeyList: [],
+  };
+
+  state = {
+    menuVisible: false,
+  };
+
+  handleCheck = (checkedKeys, allKeys) => {
+    console.log('checkedKeys', checkedKeys);
+    const { onCheck } = this.props;
+    if (onCheck) {
+      onCheck(getCategorizedUncheckedKeys(checkedKeys, allKeys));
+    }
+  };
+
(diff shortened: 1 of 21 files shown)


smurching (commented):
Thanks for this PR! As per offline discussion / review - the UX & implementation largely looks good. The wider-sort area for columns seems to be working, but causing columns to jump around gif. Maybe we can defer the wider-sort area to a follow-up PR? LGTM otherwise

Zangr:
> The wider-sort area for columns seems to be working, but causing columns to jump around


The "jump around" is actually expected for here because sorting is server-side based and the availability of metrics/params/tags before and after sort varies.

mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
```diff
@@ -0,0 +1,152 @@
...
+    metricKeyList: [],
+    visibleTagKeyList: [],
+  };
+
+  state = {
+    menuVisible: false,
+  };
+
+  handleCheck = (checkedKeys, allKeys) => {
+    console.log('checkedKeys', checkedKeys);
```
harupy: This line should be removed.

harupy:
The metrics checkbox can't be unchecked. Is this a bug or am I missing something?
bug

mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
```diff
@@ -0,0 +1,152 @@
...
+function getCategorizedUncheckedKeys(checkedKeys, allKeys) {
+  const uncheckedKeys = _.difference(allKeys, checkedKeys);
+  const result = {
+    [ColumnTypes.ATTRIBUTES]: [],
+    [ColumnTypes.PARAMS]: [],
+    [ColumnTypes.METRICS]: [],
+    [ColumnTypes.TAGS]: [],
+  };
+  uncheckedKeys.forEach((key) => {
+    const [columnType, rawKey] = key.split('-');
```
harupy: This line doesn't split `key` properly when a metric (or parameter) name contains `-`.

harupy: Tested the code below and it worked.

```diff
+    const [columnType, rawKey] = key.split('-');
-    const [columnType, rawKey] = key.split(/-(.+)/);
```

Ref. : https://stackoverflow.com/questions/4607745/split-string-only-on-first-instance-of-specified-character

Zangr: Fixed

mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
Zangr: Removed

Zangr (commented):
Addressed comments, disabled column drag/drop for now since we have not implemented persistence of column ordering. Also, added run status display for the new view.

mlflow/server/js/src/components/RunsTableColumnSelectionDropdown.js
harupy: Maybe we should let the linter detect `console.log`.

Zangr: I think it's OK to leave it open for now. It's used a lot during development phase, we probably don't want running a lint fix to auto remove them.

harupy: Agreed.

smurching (approved):
LGTM thanks @Zangr! Let's get this in & iterate on it in follow-ups if needed, thanks for the huge UX improvement :)

Issue #2254, opened 2020-01-02
mlflow models build-docker does not work on windows


jahas:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md)
for information on what types of issues we address. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

Please fill in this template and do not delete it unless you are sure your issue is outside its scope.

### System information
- Custom code and direct CLI usage as well
- Windows
- Installation from binary and from source as well
- Latest (1.4.0)
- 3.7.3
- **npm version, if running the dev UI**:
- mlflow models build-docker -m {model_path} -n {image_tag}

### Describe the problem
When you try to build docker image, then the code of mlflow generates the Dockerfile file by itself and puts path to the model into it. 
Because the command is run on Windows, the code uses windows specific path separator "\" instead of posix specific "/" 

### Code to reproduce issue
Provide a reproducible test case that is the bare minimum necessary to generate the problem.

### Other info / logs
Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks,
please include the full traceback. Large logs and files should be attached.

jahas:
This is the change that should be applied to *mlflow/pyfunc/backend.py* file : 
```
diff --git a/mlflow/pyfunc/backend.py b/mlflow/pyfunc/backend.py
index a7def318..c0d8d011 100644
--- a/mlflow/pyfunc/backend.py
+++ b/mlflow/pyfunc/backend.py
@@ -2,7 +2,7 @@ import logging
 import os

 import subprocess
+import posixpath
 from mlflow.models import FlavorBackend
 from mlflow.models.docker_utils import _build_image, DISABLE_ENV_CREATION
 from mlflow.pyfunc import ENV
@@ -125,7 +125,7 @@ class PyFuncBackend(FlavorBackend):
                 ENV {disable_env}="true"
                 """.format(
                 disable_env=DISABLE_ENV_CREATION,
-                model_dir=os.path.join("model_dir", os.path.basename(model_path)),
+                model_dir=str(posixpath.join("model_dir", os.path.basename(model_path))),
                 install_mlflow=repr(install_mlflow)
             )
```

jahas:
Pull request which solves this issue: https://github.com/mlflow/mlflow/pull/2257

juntai-zheng:
[...]

PR #2257, opened 2020-01-03
Fixed model docker image building on docker

jahas:

https://github.com/mlflow/mlflow/issues/2254

Describe the problem
When you try to build docker image, then the code of mlflow generates the Dockerfile file by itself and puts path to the model into it.
Because the command is run on Windows, the code uses windows specific path separator "" instead of posix specific "/"

Code to reproduce issue
Provide a reproducible test case that is the bare minimum necessary to generate the problem.

Other info / logs
Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks,
please include the full traceback. Large logs and files should be attached.

## How is this patch tested?

Run the code on Linux and on Windows. 

I did not managed to run successfuly all unit tests because of API changes in dependant libraries

- [ ] UI
- [ X] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ X] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

- [ X] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes

### DIFF
--- a/mlflow/pyfunc/backend.py
+++ b/mlflow/pyfunc/backend.py
@@ -2,7 +2,7 @@ import logging
 import os
 
 import subprocess
-
+import posixpath
 from mlflow.models import FlavorBackend
 from mlflow.models.docker_utils import _build_image, DISABLE_ENV_CREATION
 from mlflow.pyfunc import ENV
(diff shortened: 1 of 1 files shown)


tomasatdatabricks (approved):
lgtm, thanks @jahas!

PR #2272, opened 2020-01-08
Add autodetection of job environments to R client

dbczumar:

Adds autodetection of Databricks job environments to the MLflow R client.

## How is this patch tested?

These MLflow client changes were tested manually on Databricks against an updated Spark image that defines subroutines for fetching job information from the SparkR driver.

**Unit tests are pending**

## Release Notes

The MLflow R client now detects and records Databricks Job information when the client is used within a Databricks Job.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [X] R
- [ ] Java
- [ ] Python


mlflow/R/mlflow/R/databricks-utils.R
```diff
@@ -98,27 +98,62 @@ get_databricks_config <- function(profile) {
...
-      tags[[MLFLOW_TAGS$MLFLOW_SOURCE_TYPE]] <- MLFLOW_SOURCE_TYPE$NOTEBOOK
-      list(
+      return(list(
         client = client,
-        tags = tags,
+        tags = build_context_tags_from_databricks_notebook_info(notebook_info),
         experiment_id = experiment_id %||% notebook_info$id,
         ...
-      )
+      ))
```
tomasatdatabricks: Would it be cleaner to do this as if(notebook) {...} elseif (job) {...} else {nextMethod()} ?

tomasatdatabricks (commented):
Looks good, I had just a minor comment / suggestion.

mlflow/R/mlflow/R/databricks-utils.R
dbczumar: This would look cleaner, but I think we'd have to fetch the job info prior to the if/else-if/else block to clean this up, resulting in extra function call within Databricks notebooks. Let me know if you see a way around this / think that's a better idea than the current solution.

tomasatdatabricks: Oh I see. I think that's ok then.

tomasatdatabricks (approved):
lgtm

PR #2275, opened 2020-01-08
Add auto logging functionality for LightGBM flavor.

harupy:

Add auto logging functionality for LightGBM flavor that logs the following:

- booster parameters
- metrics on each iteration
- best and stopped iteration (if early stopping activated)
- metrics at the best iteration  (if early stopping activated)
- feature importance (logged as an artifact in json format)
- trained model (logged as an artifact)

### Sample Code

```python
import lightgbm as lgb
from mlflow.lightgbm import autolog

from sklearn import datasets

autolog()

X, y = datasets.load_iris(return_X_y=True)
train_set = lgb.Dataset(X_train)
valid_set = lgb.Dataset(X_test)

params = {
    'objective': 'multiclass',
    'num_class': 3,
    'metric': 'multi_logloss',
    'random_state': 42,
}
model = lgb.train(params, train_set, early_stopping_rounds=10,
                  valid_sets=[train_set, valid_set], valid_names=['train', 'valid'])

```

lgb-autolog-demo

## How is this patch tested?

Unit tests

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [x] Examples
- [x] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### DIFF
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -20,6 +20,11 @@ from __future__ import absolute_import
 
 import os
 import yaml
+import json
+import tempfile
+import shutil
+import inspect
+import gorilla
 
 import mlflow
 from mlflow import pyfunc
@@ -28,6 +33,9 @@ from mlflow.tracking.artifact_utils import _download_artifact_from_uri
 from mlflow.utils.environment import _mlflow_conda_env
 from mlflow.utils.model_utils import _get_flavor_configuration
 from mlflow.exceptions import MlflowException
+from mlflow.utils.annotations import experimental
+from mlflow.utils.autologging_utils import try_mlflow_log, log_fn_args_as_params
+
 
 FLAVOR_NAME = "lightgbm"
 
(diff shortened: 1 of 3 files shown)


mlflow/lightgbm.py
```diff
@@ -182,3 +192,111 @@ def __init__(self, lgb_model):
 
     def predict(self, dataframe):
         return self.lgb_model.predict(dataframe)
+
+
+@experimental
+def autolog(importance_types=['split']):  # pylint: disable=W0102
```
smurching: Hm is it expensive to compute the feature importances? Maybe we could just log both split & gain feature importance for now, & simplify the API by removing this parameter.

mlflow/lightgbm.py
```diff
@@ -182,3 +192,111 @@ def __init__(self, lgb_model):
...
+
+        # If early_stopping_rounds is present, logging metrics at the best iteration
+        # as extra metrics with the max step + 1.
+        early_stopping_index = all_arg_names.index('early_stopping_rounds')
+        early_stopping = (num_pos_args >= early_stopping_index + 1 or
+                          'early_stopping_rounds' in kwargs)
+        if early_stopping:
+            extra_step = len(eval_results)
+            try_mlflow_log(mlflow.log_metric, 'stopped_iteration', len(eval_results))
+            try_mlflow_log(mlflow.log_metric, 'best_iteration', model.best_iteration)
```
smurching: Just to confirm, it sounds ([from the docs](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.train.html#lightgbm-train)) that `best_iteration` is set as long as the user configures their Booster to use early stopping, even if we end up training for the full number of training epochs & do not stop early (?)

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -0,0 +1,270 @@
...
+
+    for param in unlogged_params:
+        assert param not in params
+
+
+@pytest.mark.large
+def test_lgb_autolog_logs_specified_params(bst_params, train_set):
+    mlflow.lightgbm.autolog()
+    expected_params = {
+        'num_boost_round': 10,
```
smurching: How long do these tests take to run btw? Could we speed them up by training for fewer epochs, or is the duration negligible (<5 seconds or so)? Mainly just wondering if we can save some CI time

smurching (commented):
Just some small comments but largely looks good, thanks @harupy!

mlflow/lightgbm.py
harupy: No, it just iterates over nodes of each tree and counts the split feature (for split) or sums the split gain (for gain). I agree with logging both. Will fix this.

How feature importance is calculated: 
https://github.com/microsoft/LightGBM/blob/edb9149c13fe5a901898db525c22b441f0db6e49/src/boosting/gbdt_model_text.cpp#L553-L580

mlflow/lightgbm.py
harupy: Yes, it doesn't matter whether or not the training early stopped.

https://lightgbm.readthedocs.io/en/latest/_modules/lightgbm/callback.html#early_stopping

mlflow/lightgbm.py
harupy: Wrote a simple test code to confirm this.

```python
# code
model = lgb.train(params,
                  train_set,
                  num_boost_round=2,
                  early_stopping_rounds=10,  # setting larger value than num_boost_round
                  valid_sets=[train_set, valid_set],
                  valid_names=['train', 'valid'])
print(model.best_iteration)
print(model.best_score)
```

output

```
[1]     train's multi_logloss: 0        valid's multi_logloss: 0
Training until validation scores don't improve for 10 rounds
[LightGBM] [Warning] Stopped training because there are no more leaves that meet the split requirements
[2]     train's multi_logloss: 0        valid's multi_logloss: 0
Did not meet early stopping. Best iteration is:
[1]     train's multi_logloss: 0        valid's multi_logloss: 0
1
defaultdict(<class 'collections.OrderedDict'>, {'train': OrderedDict([('multi_logloss', 0.0)]), 'valid': OrderedDict([('multi_logloss', 0.0)])})
```

tests/lightgbm/test_lightgbm_autolog.py
harupy: I measured how it takes to run the training part (`lgb.train(...)`). It took `0.667` seconds on my macbook pro (16GB RAM).

harupy: However, some tests use the default `num_boost_round` value (100) which I think is too large. I will fix these tests to use fewer epochs.

mlflow/lightgbm.py
harupy: Wrote a simple test code to confirm this.

```python
# code
model = lgb.train(params,
                  train_set,
                  num_boost_round=2,
                  early_stopping_rounds=10,  # setting larger value than num_boost_round
                  valid_sets=[train_set, valid_set],
                  valid_names=['train', 'valid'])
print(model.best_iteration)
print(model.best_score)
```

output

```
[1]     train's multi_logloss: 0        valid's multi_logloss: 0
Training until validation scores don't improve for 10 rounds
[LightGBM] [Warning] Stopped training because there are no more leaves that meet the split requirements
[2]     train's multi_logloss: 0        valid's multi_logloss: 0
Did not meet early stopping. Best iteration is:
[1]     train's multi_logloss: 0        valid's multi_logloss: 0
1
defaultdict(<class 'collections.OrderedDict'>, {'train': OrderedDict([('multi_logloss', 0.0)]), 'valid': OrderedDict([('multi_logloss', 0.0)])})
```

harupy: Added a comment on this behavior for clarification.

harupy:
I'm wondering if we should plot feature importance. It's logged as an JSON file now, but it'd be more helpful if it's plotted as a bar chart like the attached figure.

image

smurching (approved):
LGTM thanks for the updates! Agreed on the bar chart point, happy to review a follow-up that plots the feature importance in bar-chart form :)

harupy:
@smurching Thanks! I'll open a follow-up PR.

PR #2277, opened 2020-01-08
Remove autologging support for tf.keras.fit_generator

smurching:

``tf.keras.Model.fit_generator`` has been deprecated in TF 2.1.0, as per the [release notes](https://github.com/tensorflow/tensorflow/releases/tag/v2.1.0):

> Note that Model.fit_generator, Model.evaluate_generator, and Model.predict_generator are deprecated endpoints. They are subsumed by Model.fit, Model.evaluate, and Model.predict which now support generators and Sequences.

In particular, the tf.keras implementation of `fit_generator` now simply calls `fit`, which breaks our autologging support.

This PR circumvents the issue by removing autologging support for `fit_generator` entirely, which seems reasonable as:
* TF autologging is experimental
* `fit_generator` is not as commonly-used an API as `fit`
* User code that uses `fit_generator` with autologging will still run normally (just not autolog anything)
* Users of `fit_generator` with autologging have a workaround (replace `fit_generator` with `fit`).

## How is this patch tested?
Existing tests

Removes autologging support for `tf.keras.Model.fit_generator`

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/tensorflow.py
+++ b/mlflow/tensorflow.py
@@ -758,32 +758,6 @@ def autolog(every_n_iter=100):
             _flush_queue()
             _log_artifacts_with_warning(local_dir=log_dir, artifact_path='tensorboard_logs')
             shutil.rmtree(log_dir)
-
-            return result
-
-    @gorilla.patch(tensorflow.keras.Model)
-    def fit_generator(self, *args, **kwargs):
-        with _manage_active_run():
-            original = gorilla.get_original_attribute(tensorflow.keras.Model, 'fit_generator')
-
-            unlogged_params = ['self', 'generator', 'callbacks', 'validation_data', 'verbose']
-
-            log_fn_args_as_params(original, args, kwargs, unlogged_params)
-
-            # Checking if the 'callback' argument of fit() is set
-            if len(args) >= 5:
-                tmp_list = list(args)
-                tmp_list[4], log_dir = _setup_callbacks(tmp_list[4])
-                args = tuple(tmp_list)
-            elif 'callbacks' in kwargs:
-                kwargs['callbacks'], log_dir = _setup_callbacks(kwargs['callbacks'])
-            else:
-                kwargs['callbacks'], log_dir = _setup_callbacks([])
-            result = original(self, *args, **kwargs)
-            _flush_queue()
-            _log_artifacts_with_warning(local_dir=log_dir, artifact_path='tensorboard_logs')
-            shutil.rmtree(log_dir)
-
             return result
 
     @gorilla.patch(EventFileWriter)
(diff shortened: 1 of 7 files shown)


tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -175,15 +148,13 @@ def test_tf_keras_autolog_model_can_load_from_artifact(tf_keras_random_data_run,
 
 def create_tf_estimator_model(dir, export):
     CSV_COLUMN_NAMES = ['SepalLength', 'SepalWidth', 'PetalLength', 'PetalWidth', 'Species']
-    SPECIES = ['Setosa', 'Versicolor', 'Virginica']
 
     train = pd.read_csv(os.path.join(os.path.dirname(__file__), "iris_training.csv"),
                         names=CSV_COLUMN_NAMES, header=0)
     test = pd.read_csv(os.path.join(os.path.dirname(__file__), "iris_test.csv"),
```
juntai-zheng: We could probably remove this like you did in the TF1 file.

mlflow/tensorflow.py
```diff
@@ -758,32 +758,6 @@ def fit(self, *args, **kwargs):
             _flush_queue()
             _log_artifacts_with_warning(local_dir=log_dir, artifact_path='tensorboard_logs')
             shutil.rmtree(log_dir)
-
-            return result
-
-    @gorilla.patch(tensorflow.keras.Model)
-    def fit_generator(self, *args, **kwargs):
```
juntai-zheng: If `fit_generator()` is only deprecated in TF 2.1.0, do we need to remove it from TF 1.X? Are we trying to keep both versions consistent with each other?

smurching: Good question, we only have a single autologging patch for `tf.keras.fit_generator`, so removing `fit_generator` support here removes it for both 1.x & 2.x. It seems 1.x users should also be able to pass a generator to `fit`, as per the TF 1.14 API docs: https://www.tensorflow.org/versions/r1.14/api_docs/python/tf/keras/Model#fit

juntai-zheng: Ah I had thought that if `fit_generator` did something different from `fit`, we could've patched the method for TF 1.X only by having the version check before the patch. But it turns out that `fit` [just calls](https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/python/keras/engine/training.py#L682) `fit_generator()`. LGTM!

PR #2293, opened 2020-01-11
Add feature importance plot to LightGBM auto logging

harupy:

Follow-up for #2275. Add feature importance plot to LightGBM auto logging.

## How is this patch tested?

Unit tests

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [ ] Examples
- [x] Docs
- [x] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### DIFF
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -237,10 +237,10 @@ def autolog(normalize=False, max_num_features=None):
             Plot feature importance.
             """
 
+            # pylint: disable=invalid-unary-operand-type
             if max_num_features is None:
                 indices = np.argsort(importance)
             else:
-                # pylint: disable=invalid-unary-operand-type
                 indices = np.argsort(importance)[-max_num_features:]
 
             features = np.array(features)[indices]
@@ -263,6 +263,7 @@ def autolog(normalize=False, max_num_features=None):
 
             tmpdir = tempfile.mkdtemp()
             try:
+                # pylint: disable=undefined-loop-variable
                 filepath = os.path.join(tmpdir, 'feature_importance_{}.png'.format(imp_type))
                 fig.savefig(filepath)
                 try_mlflow_log(mlflow.log_artifact, filepath)
@@ -335,7 +336,6 @@ def autolog(normalize=False, max_num_features=None):
             imp = {ft: imp for ft, imp in zip(features, importance.tolist())}
             tmpdir = tempfile.mkdtemp()
             try:
-                # pylint: disable=undefined-loop-variable
                 filepath = os.path.join(tmpdir, 'feature_importance_{}.json'.format(imp_type))
                 with open(filepath, 'w') as f:
                     json.dump(imp, f, indent=2)
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -323,7 +323,7 @@ def autolog():
             importance = model.feature_importance(importance_type=imp_type)
             try:
                 log_feature_importance_plot(features, importance, imp_type)
-            except Exception as e:
+            except Exception:  # pylint: disable=broad-except
                 _logger.exception('Failed to log feature importance plot. LightGBM autologging'
                                   'will ignore the failure and continue. Exception: ')
             log_feature_importance_plot(features, importance, imp_type)
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -195,7 +195,7 @@ class _LGBModelWrapper:
 
 
 @experimental
-def autolog(normalize=False, max_num_features=None):
+def autolog():
     """
     Enables automatic logging from LightGBM to MLflow. Logs the following.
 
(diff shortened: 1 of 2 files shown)


mlflow/lightgbm.py
```diff
@@ -225,6 +232,42 @@ def callback(env):
...
+            fig.tight_layout()
+
+            tmpdir = tempfile.mkdtemp()
+            try:
+                filepath = os.path.join(tmpdir, 'feature_importance_{}.png'.format(imp_type))
+                fig.savefig(filepath)
+                try_mlflow_log(mlflow.log_artifact, filepath)
+            finally:
+                shutil.rmtree(tmpdir)
+
```
harupy: Maybe this functions should be moved to [`mlflow/utils`](https://github.com/mlflow/mlflow/tree/master/mlflow/utils) so that we can reuse it on `xgboost.autolog`.

harupy:
LightGBM has a function named [`plot_importance`](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.plot_importance.html#lightgbm-plot-importance), but this function:
- doesn't provide an option to normalize feature importance.
- doesn't adjust the figure height when the number of features to plot is large.

mlflow/lightgbm.py
```diff
@@ -195,19 +195,26 @@ def predict(self, dataframe):
...
     Enables automatic logging from LightGBM to MLflow. Logs the following.
 
     - parameters specified in `lightgbm.train`_.
     - metrics on each iteration (if ``valid_sets`` specified).
     - metrics at the best iteration (if ``early_stopping_rounds`` specified).
-    - feature importance (both "split" and "gain").
+    - feature importance (both "split" and "gain") as JSON files and plots.
     - trained model.
 
+    :param normalize: If ``True``, divide each value in feature importance
```
smurching: One question around these parameters (I'm curious since I'm not a lightgbm user myself) - how commonly do you imagine users would want to configure these? If it's rare, or if the defaults would work reasonably well, we could just remove these parameters for now & wait for user feedback (& you might even be able to provide this feedback yourself if you've been using lightgbm) that suggests we should include them, as it's good to default to keeping APIs simple where possible.

harupy: Agreed, will remove them for now.

mlflow/lightgbm.py
```diff
@@ -225,6 +227,38 @@ def callback(env):
                 eval_results.append(res)
             return callback
 
+        def plot_importance(features, importance, importance_type):
```
smurching:
```suggestion
        def log_feature_importance_plot(features, importance, importance_type):
```

Just to make the name a bit clearer

mlflow/lightgbm.py
```diff
@@ -284,12 +318,13 @@ def callback(env):
         for imp_type in ['split', 'gain']:
             features = model.feature_name()
             importance = model.feature_importance(importance_type=imp_type)
+            plot_importance(features, importance, imp_type)
```
smurching: Looking at Travis tests, it seems generating the feature importance plot via matplotlib can fail if the user's environment isn't properly configured (see [this SO post](https://stackoverflow.com/questions/37604289/tkinter-tclerror-no-display-name-and-no-display-environment-variable)). Given that, we should probably try-catch this plotting logic. We can initialize a logger object at the top of our module (see analogous [mlflow.tensorflow code](https://github.com/mlflow/mlflow/blob/8447c4201ae7306d69b692f5e634c7eb59cf1462/mlflow/tensorflow.py#L47) for example) & then use it to log any plotting failures that occur:

```suggestion
      try:
            log_feature_importance_plot(features, importance, imp_type)
      except Exception as e:
            _logger.exception("Failed to log feature importance plot. LightGBM autologging will ignore the failure and continue. Exception: ")
```

smurching: We likely also need to configure matplotlib to not use the Tkinter backend in our Travis tests, as described in the above SO post.

mlflow/lightgbm.py
```diff
@@ -284,12 +321,18 @@ def callback(env):
         for imp_type in ['split', 'gain']:
             features = model.feature_name()
             importance = model.feature_importance(importance_type=imp_type)
+            try:
+                log_feature_importance_plot(features, importance, imp_type)
+            except Exception:  # pylint: disable=broad-except
+                _logger.exception('Failed to log feature importance plot. LightGBM autologging'
```
smurching:
```suggestion
                _logger.exception('Failed to log feature importance plot. LightGBM autologging '
```

Micronit: just making sure there are spaces between words in the error message

smurching (approved):
Tiny nit, LGTM after that's addressed

mlflow/lightgbm.py
harupy: Oh, thanks for the catch!

harupy: Fixed it.

smurching (approved):
LGTM

PR #2295, opened 2020-01-12
Add run links popover to metrics plot

### LINKED ISSUE #2279: [FR][UI] Enable clicking through to individual runs from metric plots
## Describe the proposal

Currently, once you view a metric/plot comparing many runs in the MLflow UI, you cannot click through to individual runs that appear in the plot (e.g. the run with the best final accuracy etc). We should fix this, starting with metric comparison plots defined in [MetricPlotView.js](https://github.com/mlflow/mlflow/blob/master/mlflow/server/js/src/components/MetricsPlotView.js) - the exact fix may be nontrivial as it's not clear that Plotly supports adding links to legends/tooltips, although [this doc page](https://help.plot.ly/adding-HTML-and-links-to-charts/)) suggests that it's possible.

harupy:

Resolves #2279 

## How is this patch tested?

(Details)

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


harupy:
It seems that the plot area gets initialized when `setState` triggers a re-render.

demo

juntai-zheng:
Hey @harupy, @smurching and I took a look at what was causing the issue. We've isolated it to the `metrics` prop changing in the `MetricsPlotView`, causing it to refresh whenever there is a click and the popover is created. This also happens when you're not zoomed in, though it's not noticeable because it doesn't change the graph. You can see how we found the bug in this screenshot: 
image

We used the hook from [here](https://stackoverflow.com/questions/41004631/trace-why-a-react-component-is-re-rendering/51082563#51082563) to log which prop was changed.

Do you think you could take a closer look as to why `metrics` is changing? If that doesn't change then the zoom in problem might be solved. Thanks!

harupy:
@juntai-zheng Thanks for the investigation. The actual cause is when toggling the popover,  the state `popoverVisible` changes. This change triggers re-rendering and the plot gets zoomed out.

juntai-zheng:
@harupy ah sorry, forgot to mention that we had actually removed `popoverVisible` as another part of our investigation since we saw that it was causing re-rendering and didn't seem needed in `MetricsPlotView`. When we did, `metrics` was changing as well (causing the re-rendering).

harupy:
@juntai-zheng Can you share the branch you're working on?

juntai-zheng:
@harupy Here it is: https://github.com/juntai-zheng/mlflow/tree/popover

Just checking, the links created by the popover link me back to the MLflow home page. Is linking to the appropriate run a WIP?

harupy:
@juntai-zheng 

>  Is linking to the appropriate run a WIP?
Yes

harupy:
@juntai-zheng 

I checked your branch. In JavaScript, `===` compares identity for objects, not equality.

```javascript
> const a = {};
> const b = {};
> const c = a;

> a === b
false

>>> a === c
true
``` 

So the comparison below doesn't work when `val` is an object or a list.

```javascript
prevProps[key] !== val
```

harupy:
One way to freeze the plot when toggling the popover is to use `shouldComponentUpdate` in `MetricsPlotView` and prevent the re-render if the current props equals to the previous ones.

juntai-zheng:
Hi @harupy, you were correct in the hook being incorrect and `metrics` changing. We did some more digging and found out that because the state of the popover was contained in the state of `MetricsPlotPanel`, the changes to the state of the popover would also change the state of `MetricsPlotPanel`. Since `MetricsPlotPanel` is a parent to the popover and `MetricsPlotView`, whenever `MetricsPlotPanel` would re-render, so would `MetricsPlotView`. To avoid this, we made a [ref](https://reactjs.org/docs/refs-and-the-dom.html) to the popover in `MetricsPlotPanel` and only changed the state of `RunLinksPopover`, which subsequently only re-renders the popover and left the plot alone. 

We made rough changes to my branch [https://github.com/juntai-zheng/mlflow/tree/popover](https://github.com/juntai-zheng/mlflow/tree/popover), so you can check it out from there.

harupy:
@juntai-zheng Awesome. I'll reflect the change on this branch.

harupy:
Now it looks like:

demo1

juntai-zheng:
@harupy the popover looks great! Looks like a couple test are failing because of extra props being passed with our modifications. Do you think you could apply a fix for the tests?

harupy:
@juntai-zheng I will fix the tests!

@gioa 
Any thoughts on the popover design?

harupy:
Maybe we should insert a message somewhere that describes how to display the popover.

gioa:
@harupy How is the popover triggered? It seems a user can click on anywhere to trigger it. Would that interfere with the zoom actions? 

The popover looks great! One tiny detail is people may not know they can click on the run details, adding a blue fa-external-link icon on the right or some hover style may help its discoverability.

harupy:
@gioa The popover is triggered when a user performs a single click on the plot. It doesn't interfere with the zoom actions.

harupy:
@gioa 
I added the icon to the run links. Now it looks like below.

icon

harupy:
When the icons are placed on the right:

Screen Shot 2020-02-04 at 19 51 24

gioa:
Thank you for adding the icon, it looks good! Could we also support the "ESC" key to close the popover so users don't need to move to the close icon to dismiss it? Thank you!

harupy:
The popover doesn't seem to support the keypress event.

- https://ant.design/components/popover/#API
- https://ant.design/components/tooltip/#API

Zangr:
@harupy one way to do it is to bind keydown event to document. I created an [example here](https://stackblitz.com/edit/react-an4qj7?file=index.js).

harupy:
@Zangr Thanks! I'll add these life-cycle methods you implemented in the example.

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -126,8 +130,38 @@ export class MetricsPlotPanel extends React.Component {
...
+        })).reverse();
+
+        // Instead of setting state on MetricPlotPanel (current component),
+        // should call a method on the Popover directly via the ref.
+        popover.setState({
+          visible: !visible || !clickedSamePoint,
+          x: clientX,
+          y: clientY,
+          runItems,
+        });
```
Zangr: I'd recommend creating a method in RunLinksPopover with 4 parameters to avoid exposing child components state directly to the parent component.

mlflow/server/js/src/components/MetricsPlotView.js
```diff
@@ -107,12 +111,13 @@ export class MetricsPlotView extends React.Component {
...
 
   render() {
+    const { onClick, onRelayout } = this.props;
     const plotProps =
       this.props.chartType === CHART_TYPE_BAR
         ? this.getPlotPropsForBarChart()
         : this.getPlotPropsForLineChart();
     return (
-      <div className='metrics-plot-view-container'>
+      <div className="metrics-plot-view-container">
```
Zangr: single-quote in all places. Linter: [tinyurl.com/mlflowjs](https://tinyurl.com/mlflowjs)

Zangr (commented):
Some drive-by comments, mostly nits.

mlflow/server/js/src/components/RunLinksPopover.js
```diff
@@ -0,0 +1,94 @@
...
+            top: y,
+            position: 'absolute',
+          }}
+        />
+      </Popover>
+    );
+  }
+}
+
+export default RunLinksPopover;
```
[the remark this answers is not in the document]
harupy: Why?

mlflow/server/js/src/components/MetricsPlotPanel.js
harupy: Do you mean to create a method that looks like this?

```javascript
class RunLinksPopover extends React.Component {
  ...

  method = (visible, x, y, runItems) => this.setState(...);

  ...
}
```

mlflow/server/js/src/components/MetricsPlotView.js
harupy: Will fix this.

mlflow/server/js/src/components/RunLinksPopover.js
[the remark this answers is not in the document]
Zangr: Related articles: 
https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/
https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

mlflow/server/js/src/components/MetricsPlotPanel.js
Zangr: Yes.

mlflow/server/js/src/components/RunLinksPopover.js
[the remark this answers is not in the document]
harupy: I agree. Will fix this.

juntai-zheng:
@harupy Looks like there's a couple of conflicts from #2408. Do you think you could resolve them? It'd be great to get this PR in.

@Zangr How does the PR look?

Zangr (approved):
LGTM

harupy:
Should we enable this feature only when a user compares multiple runs?

harupy:
Added the following updates.
- Display the popover on the left to prevent it from overlapping the tooltips on the plot.
- Display metric values after run names.

latest

harupy:
Maybe we should truncate run uuid in the run name (for example displaying only the first 8 characters)  because the plot width is reduced when comparing multiple runs.

Screen Shot 2020-02-23 at 17 48 44

Zangr:
@harupy , the changes look merge ready, would you like to resolve the conflicts?

harupy:
@Zangr Fixed the conflict!

harupy:
> Maybe we should truncate run uuid in the run name (for example displaying only the first 8 characters) because the plot width is reduced when comparing multiple runs.
> 
> Screen Shot 2020-02-23 at 17 48 44

@Zangr What do you think about this?

Zangr:
> Maybe we should truncate run uuid in the run name (for example displaying only the first 8 characters) because the plot width is reduced when comparing multiple runs.

Agree, I think we can show a short version and show the full id in the tooltip. cc @gioa

harupy:
Yes, that makes sense. I'll open a PR.

harupy:
@Zangr @smurching 
I found a bug. The popover shows up after relayout by double click. I'll open a PR to fix this bug.

popover-bug

vandit2209:
Hey, guys I dont see anyway to reach to compare view and get hoverable plots as you have, any guidlines?

PR #2297, opened 2020-01-12
Add get latest and download model version interface to java client

andychow-db:
Add Java interfaces defininga the following functions

    1. getLatestModelVersionPerStage
    2. getModelVersionDownloadUri
    3. downloadModel

## How is this patch tested?

(Details)

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [x] Java
- [ ] Python

### DIFF
--- a/mlflow/java/client/pom.xml
+++ b/mlflow/java/client/pom.xml
@@ -50,11 +50,6 @@
       <groupId>org.ini4j</groupId>
       <artifactId>ini4j</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>jsr305</artifactId>
-      <version>3.0.2</version>
-    </dependency>
 
     <!-- Test dependencies -->
     <dependency>
--- a/mlflow/java/client/pom.xml
+++ b/mlflow/java/client/pom.xml
@@ -50,6 +50,11 @@
       <groupId>org.ini4j</groupId>
       <artifactId>ini4j</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.2</version>
+    </dependency>
 
     <!-- Test dependencies -->
     <dependency>
--- a/mlflow/java/client/pom.xml
+++ b/mlflow/java/client/pom.xml
@@ -85,7 +85,7 @@
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
           <sourcepath>${project.basedir}/src/main/java</sourcepath>
-          <excludePackageNames>com.databricks.api.proto.databricks:org.mlflow.scalapb_interface:org.mlflow.tracking.samples:org.mlflow.artifacts</excludePackageNames>
+          <excludePackageNames>com.databricks.api.proto.databricks:org.mlflow.scalapb_interface:org.mlflow.tracking:org.mlflow.artifacts:org.mlflow.api</excludePackageNames>
           <groups>
             <group>
               <title>Tracking API</title>
--- a/mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
+++ b/mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
@@ -681,25 +681,29 @@ public class MlflowClient {
   }
 
   /**
-   * Return the latest model version for each stage..
+   * Return the latest model version for each stage requested.
    * The current stages are: [None, Staging, Production, Archived].
    *
    *    <pre>
-   *        ModelVersionDetailed details = getLatestVersions("model", "Staging");
+   *        List<ModelVersionDetailed> detailsList  = getLatestVersions("model",
+   *                                                              Lists.newArrayList<>("Staging"));
    *
-   *        System.out.println("Model Name: " + details.getModelVersion()
-   *                                                   .getRegisteredModel()
-   *                                                   .getName());
-   *        System.out.println("Model Version: " + details.getModelVersion().getVersion());
-   *        System.out.println("Current Stage: " + details.getCurrentStage());
+   *        for (ModelVersionDetailed details : detailsList) {
+   *            System.out.println("Model Name: " + details.getModelVersion()
+   *                                                       .getRegisteredModel()
+   *                                                       .getName());
+   *            System.out.println("Model Version: " + details.getModelVersion().getVersion());
+   *            System.out.println("Current Stage: " + details.getCurrentStage());
+   *        }
    *    </pre>
    *
    * @param modelName The name of the model
-   * @param stage The name of the stage
+   * @param stages A list of stages
    * @return The latest model version
    *         {@link org.mlflow.api.proto.ModelRegistry.ModelVersionDetailed}
    */
-  public ModelVersionDetailed getLatestVersions(@Nonnull String modelName, @Nonnull String stage) {
+  public List<ModelVersionDetailed> getLatestVersions(@Nonnull String modelName,
+                                                      @Nonnull Iterable<String> stages) {
     throw new UnsupportedOperationException("getLatestVersion is not supported");
   }
 
--- a/mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
+++ b/mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
@@ -22,6 +22,7 @@ import org.mlflow.tracking.creds.DatabricksDynamicHostCredsProvider;
 import org.mlflow.tracking.creds.HostCredsProviderChain;
 import org.mlflow.tracking.creds.MlflowHostCredsProvider;
 
+import javax.annotation.Nonnull;
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
--- a/mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
+++ b/mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
@@ -676,7 +676,7 @@ public class MlflowClient {
    * @param modelName The name of the model
    * @return A collection of {@link org.mlflow.api.proto.ModelRegistry.ModelVersionDetailed}
    */
-  public List<ModelVersionDetailed> getLatestVersions(@Nonnull String modelName) {
+  public List<ModelVersionDetailed> getLatestVersions(String modelName) {
     throw new UnsupportedOperationException("getLatestVersion is not supported");
   }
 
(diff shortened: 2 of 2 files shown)


mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,81 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *
+   *    <pre>
+   *        String modelUri = getModelVersionDownloadUri("model", 0);
+   *    </pre>
+   *
+   * @param modelName The name of the model
+   * @param modelVersion The version number of the model
+   * @return The specified model version's URI.
+   */
+  public String getModelVersionDownloadUri(String modelName, long modelVersion) {
```
sueann: `modelVersion` -> `version` for consistency and also because we've been using `version` to refer to the number and `modelVersion` to refer to objects in the rest of MLflow.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,81 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *            System.out.println("Model Name: " + details.getModelVersion().getRegisteredModel().getName());
+   *            System.out.println("Model Version: " + details.getModelVersion().getVersion());
+   *            System.out.println("Current Stage: " + details.getCurrentStage());
+   *        }
+   *    </pre>
+   *
+   * @param modelName The name of the model
+   * @return A collection of {@link org.mlflow.api.proto.ModelRegistry.ModelVersionDetailed}
+   */
+  public List<ModelVersionDetailed> getLatestModelVersionPerStage(String modelName) {
```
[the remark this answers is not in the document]
sueann: And `modelName` -> `name` - this one is a bit debatable for clarity, but I think not too bad (at least within the context of model registry there is no other name)

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
sueann: And `modelName` -> `name`

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,81 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *
+   *    <pre>
+   *        File modelVersionFile = downloadModelVersion("model", 0);
+   *    </pre>
+   *
+   * @param modelName The name of the model
+   * @param modelVersion The version number of the model
+   * @return A the local file or directory ({@ java.io.File}) containing model artifacts
+   */
+  public File downloadModelVersion(String modelName, long modelVersion) {
```
sueann: we could use the `models://` URI with one `downloadModelVersion` method instead of two: the `models:/` URI can be either `models:/name/version` as well as `models:/name/stage` (the latter to get the latest version in that stage). i think the implementation will be simple for this as well - using `ArtifactRepository`. even for the currently proposed 2 methods, most likely you'd end up using `ArtifactRepository` with `models:/` URIs constructed within the methods.

sueann (commented):
thanks! left some comments - we can discuss more tomorrow.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,81 @@ private ArtifactRepository getArtifactRepository(String runId) {
     URI baseArtifactUri = URI.create(getRun(runId).getInfo().getArtifactUri());
     return artifactRepositoryFactory.getArtifactRepository(baseArtifactUri, runId);
   }
+
+
+  // ********************
+  // * Model Repository *
```
mparkhe: Model Registry

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,81 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+  // ********************
+
+  /**
+   * Return the latest model version for each stage (devo, staging, prod).
```
mparkhe: Currently supported stages are : `None`, `Staging`, `Production`, and `Archived`

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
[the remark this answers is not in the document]
mparkhe: +1 let's make stages as optional argument (default: returns model versions for all stages)
I guess to be consistent with rest of the Java APIs (and be Java-esque) we need two APIs

```
public List<ModelVersionDetailed> getLatestModelVersion(String modelName, List<String> stages) { ... }
public List<ModelVersionDetailed> getLatestModelVersion(String modelName) { 
  // calls the other with defaults
}
```

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,81 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *    <pre>
+   *        File modelVersionFile = downloadLatestModelVersion("model");
+   *    </pre>
+   *
+   * (i.e., the contents of the local directory are now available).
+   *
+   * @param modelName The name of the model
+   * @return A the local file or directory ({@ java.io.File}) containing model artifacts
+   */
+  public File downloadLatestModelVersion(String modelName) {
```
mparkhe: optional `stage` as argument. default value `"Production"` .

Similar comment as above about 2 flavors-

```
public File downloadLatestModelVersion(String name, String stage) { ... }
public File downloadLatestModelVersion(String name) {
  downloadLatestModelVersion(name, "Production")
}
```

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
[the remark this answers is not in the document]
andychow-db: Sounds good. I'll update the name to `getLatestVersions` and add a second function that includes the `stage` parameter. 

`modelName` to `name` is not consistent with the other APIs such as `experimentName` in `createExperiment`. Would you still like me to change it to `name`?

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Done.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Done

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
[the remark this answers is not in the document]
sueann: argh - i guess there we deviated from python. Let's keep it as `modelName` - or `registeredModelName` to be more precise. Mayyybe it'll be less confusing than `modelName` which could be a model not in model registry.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: I've updated the docs with those stages. Would it be better to link to a doc about staging instead?

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Done

andychow-db: In discussion, getting rid of this API.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,120 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *        File modelVersionFile = downloadLatestModelVersion("model", "Staging");
+   *    </pre>
+   *
+   * (i.e., the contents of the local directory are now available).
+   *
+   * @param modelName The name of the model
+   * @param stage The name of the stage
+   * @return A the local file or directory ({@ java.io.File}) containing model artifacts
+   */
+  public File downloadLatestModelVersion(String modelName, String stage) {
```
andychow-db: Add `@NonNull` to params

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,120 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
   }
+
+
+  // ********************
+  // * Model Registry *
+  // ********************
+
+  /**
+   * Return the latest model version for each stage.
+   * Currently only the following stages are supported: [None, Staging, Production, Archived].
```
andychow-db: 1. Update to just call getStages API
2. Add new API getStages

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: The agreement is remove the documentation about the specific stages and ask customers to call the getStages API

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: I've taken a look at the getModelVersionStages API and for discovering that stages are available, it encounters a chicken-egg problem. You need to know the available stage names but you need to provide a model version to get all the available stage names. The default is to use the latest model version which won't be deployed to any stage.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Done

mlflow/java/client/pom.xml
```diff
@@ -50,6 +50,11 @@
       <groupId>org.ini4j</groupId>
       <artifactId>ini4j</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.2</version>
```
andychow-db: Supports the `@Nonnull` annotation.

PR #2301, opened 2020-01-13
Add EarlyStopping integration to TensorFlow.Keras autologging

juntai-zheng:

Mirrors work on #2219 by adding EarlyStopping callback integration to TensorFlow.Keras (both 1.X and 2.X). Does not include `fit_generator()` support.

## How is this patch tested?

Unit tests written in `tests/tensorflow_autolog/test_tensorflow_autolog.py` and `tests/tensorflow_autolog/test_tensorflow2_autolog.py`

Same as #2219 , but this is for TensorFlow.Keras.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/mlflow/tensorflow.py
+++ b/mlflow/tensorflow.py
@@ -746,12 +746,12 @@ def autolog(every_n_iter=100):
     def _log_early_stop_callback_params(callback):
         if callback:
             try:
-                params_to_be_logged = {'monitor': callback.monitor,
-                                       'min_delta': callback.min_delta,
-                                       'patience': callback.patience,
-                                       'baseline': callback.baseline,
-                                       'restore_best_weights': callback.restore_best_weights}
-                log_fn_args_as_params(type(callback), [], params_to_be_logged, ['verbose', 'mode'])
+                earlystopping_params = {'monitor': callback.monitor,
+                                        'min_delta': callback.min_delta,
+                                        'patience': callback.patience,
+                                        'baseline': callback.baseline,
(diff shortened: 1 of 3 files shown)


smurching (commented):
Mostly looks good. We should decide what to do regarding the tradeoff of:
1. Log both loss & epoch_loss, accuracy & epoch_acc metrics, including in the non-early stopping case (the current behavior - IMO this might confuse users). We could make this better by only logging the extra `loss` & `acc` metrics in the early-stopping case
2. Remove Tensorboard callback from Keras model to avoid generating `epoch_loss`, `epoch_acc` metrics. This has the downside of us no longer generating Tensorboard logs as an artifact, & if the user does include a TensorBoard callback we'll still log `epoch_loss` etc.
3. Provide special handling when logging metrics of the restored model after early stopping - log loss & accuracy with the `epoch_` prefix, & don't log an extra step of other metrics

Out of these options I think I like 3 the most, if it's not too complicated - let me know what you think.

smurching (approved):
LGTM

juntai-zheng:
We decided it was best to stick with option number 3 for simplicity. Note that the callback logs `epoch_loss` and `epoch_acc` for TF 1.X, but `loss` and `acc` for TF 2.X.

PR #2306, opened 2020-01-14
Travis job not erroring fix

juntai-zheng:

Python 3 tests have been failing due to an issue with PyTorch, but Travis hasn't been catching them because the return value of the script that runs the stage is dependent only on the last `if` statement. This fix ~~separates the lines of the script (using dashes) so that if one command of the script fails, Travis will recognize the error.~~ adds external wrapper scripts to take care of multi-command issues (WIP, see this issue [here](https://github.com/travis-ci/dpl/issues/673)).

You can see an example of the tests failing but job passing [here](https://travis-ci.org/mlflow/mlflow/jobs/636624763#L1787). This fix will make the stage fail for our builds after this fix is implemented, until we fix the Pillow import problem (As of 1/15 this is now fixed). ~~This means the build for this PR will fail because of the broken tests, but this is the expected behavior.~~

You can see why the script is faulty in the test PR #2305 .

You can see an example of the script detecting a faulty test in [this job](https://travis-ci.org/mlflow/mlflow/jobs/637696148).

## How is this patch tested?

Through manual verification on Travis

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


smurching (approved):
LGTM

smurching:
From offline discussion - tests still passed after this fix, so something's still off. It turns out Travis will only fail the build if the last command fails (or something along those lines?), so @juntai-zheng is working on pulling out build stage logic into bash scripts so that we can run them via a single command

smurching (approved):
LGTM

PR #2308, opened 2020-01-15
[ML-9117] Java API - Model Registry API Download Models Implementation

andychow-db:

This change introduces the implementation for using Java to discover and download registered models from the model registry.

## How is this patch tested?

1. Unit tests were implemented for the new MlflowClient APIs.
2. Manually validated that the download APIs work against DBFS by hitting dogfood. 

Introduces the ability for the Java client to list and download models from the model registry. The new functions are part of the `MlflowClient` class. Note that they are experimental and have been documented as such in the JavaDocs. 

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [x] Java
- [ ] Python


mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *            System.out.println("Current Stage: " + details.getCurrentStage());
+   *        }
+   *    </pre>
+   *
+   * @param modelName The name of the model
+   * @return A collection of {@link org.mlflow.api.proto.ModelRegistry.ModelVersionDetailed}
+   */
+  public List<ModelVersionDetailed> getLatestVersions(@Nonnull String modelName) {
+    String json = sendPost("registered-models/get-latest-versions",
+            mapper.makeGetLatestVersion(modelName));
```
sueann: nit: weird indent?

mlflow/entities/model_registry/model_version_stages.py
```diff
@@ -9,7 +9,7 @@
 STAGE_DELETED_INTERNAL = "Deleted_Internal"
 
 ALL_STAGES = [STAGE_NONE, STAGE_STAGING, STAGE_PRODUCTION, STAGE_ARCHIVED]
-DEFAULT_STAGES_FOR_GET_LATEST_VERSIONS = [STAGE_STAGING, STAGE_PRODUCTION]
+DEFAULT_STAGES_FOR_GET_LATEST_VERSIONS = [STAGE_NONE, STAGE_STAGING, STAGE_PRODUCTION]
```
sueann: @mparkhe and I discussed this and `NONE` was excluded intentionally. Should we change the behavior?

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+    GetLatestVersions.Response response =  mapper.toGetLatestVersionsResponse(json);
```
sueann: nit: extra space before `mapper`

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *        System.out.println("Model Version: " + details.getModelVersion().getVersion());
+   *        System.out.println("Current Stage: " + details.getCurrentStage());
+   *    </pre>
+   *
+   * @param modelName The name of the model
+   * @param stage The name of the stage
+   * @return The latest model version
+   *         {@link org.mlflow.api.proto.ModelRegistry.ModelVersionDetailed}
+   */
+  public ModelVersionDetailed getLatestVersions(@Nonnull String modelName, @Nonnull String stage) {
```
sueann: i think we want plural `stages` (as in the [python version](https://www.mlflow.org/docs/latest/python_api/mlflow.tracking.html#mlflow.tracking.MlflowClient.get_latest_versions))

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+    String json = sendPost("registered-models/get-latest-versions",
+            mapper.makeGetLatestVersion(modelName, stage));
+
+    GetLatestVersions.Response response =  mapper.toGetLatestVersionsResponse(json);
+
+    if (response.getModelVersionsDetailedCount() < 1) {
+      throw new MlflowClientException(String.format("The model %s does not exist", modelName));
```
sueann: in which case we may not want to throw here

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
   }
+
+
+  // ********************
+  // * Model Registry *
+  // ********************
+
+  /**
+   * Return the latest model version for each stage.
+   * The current stages are: [None, Staging, Production, Archived].
```
sueann: "The currently valid stages are" ?

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+  public List<ModelVersionDetailed> getLatestVersions(@Nonnull String modelName) {
+    String json = sendPost("registered-models/get-latest-versions",
+            mapper.makeGetLatestVersion(modelName));
+    GetLatestVersions.Response response =  mapper.toGetLatestVersionsResponse(json);
+    return response.getModelVersionsDetailedList();
+  }
+
+  /**
+   * Return the latest model version for each stage..
+   * The current stages are: [None, Staging, Production, Archived].
```
sueann: "The currently valid stages are" ?

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+  /**
+   * Return the model URI containing for the given model version. The model URI can be used
+   * to download the model version artifacts.
+   *
+   * For example, the model URI
+   *
+   *    <pre>
+   *        String modelUri = getModelVersionDownloadUri("model", 0);
+   *    </pre>
+   *
```
sueann: finish the sentence?

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +638,137 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *    <pre>
+   *        File modelVersionFile = downloadModelVersion("model", 0);
+   *    </pre>
+   *
+   * @param modelName The name of the model
+   * @param version The version number of the model
+   * @return A the local file or directory ({@ java.io.File}) containing model artifacts
+   */
+  public File downloadModelVersion(@Nonnull String modelName, long version) {
```
sueann: were you able to test this against the databricks model registry (which uses dbfs paths for the model download uri)?

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowProtobufMapper.java
```diff
@@ -118,6 +121,68 @@ String makeRestoreRun(String runId) {
...
+  String makeGetLatestVersion(String modelName, String stage) {
+    RegisteredModel model = RegisteredModel.newBuilder()
+            .setName(modelName)
+            .build();
+
+    GetLatestVersions.Builder builder = GetLatestVersions.newBuilder();
+    builder.setRegisteredModel(model);
+
+    if (stage != null) {
+      builder.addStages(stage);
```
sueann: surprised this takes in a single stage

mlflow/java/client/src/test/java/org/mlflow/tracking/MlflowClientTest.java
```diff
@@ -11,11 +11,14 @@
 import java.util.Set;
 import java.util.HashSet;
 import java.util.Stack;
+import java.util.UUID;
```
sueann: clean up imports

mlflow/java/client/src/test/java/org/mlflow/tracking/ModelRegistryMlflowClientTest.java
```diff
@@ -0,0 +1,117 @@
...
+        FileUtils.writeStringToFile(tempFile, content, StandardCharsets.UTF_8);
+
+        client.sendPost("registered-models/create",
+                mapper.makeCreateModel(modelName));
+
+        client.sendPost("model-versions/create",
+                mapper.makeCreateModelVersion(modelName, runId, tempFile.getAbsolutePath()));
+
+
+        client.sendPost("model-versions/get-details",
```
sueann: was this for debugging?

mlflow/java/client/src/test/java/org/mlflow/tracking/ModelRegistryMlflowClientTest.java
```diff
@@ -0,0 +1,117 @@
...
+    private String modelName;
+    private File tempFile;
+
+    private static final String content = "Hello, Worldz!";
+
+    @BeforeSuite
+    public void beforeAll() throws IOException {
+        client = Mockito.spy(testClientProvider.initializeClientAndSqlLiteBasedServer());
+        modelName = "Model-" + UUID.randomUUID().toString();
+
```
sueann: nit: too many newlines

mlflow/java/client/src/main/java/org/mlflow/utils/CliCommand.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Runs MLflow CLI commands
+ *
+ * TODO(andy): CliBasedArtifactRepository was the progenitor to this function. It contains
```
sueann: yeah i think we should discuss this -- the python `ArtifactRepository` is somewhat run-based but not explicitly (it has an artifact_root but it does not explicitly say that is based on a run). I think the java `CliArtifactRepository` has `runId` for convenience, not because it's meant to be run-based. it'll take some work to re-work `CliArtifactRepository` (which seems more ideal than having to duplicate the functionality here) so we should discuss when you get back.

sueann (commented):
some comments inline - looks generally good to me but we have a few things to figure out.

mlflow/entities/model_registry/model_version_stages.py
andychow-db: If y'all had that discussion then I'll change it back. 

The docs should be updated to reflect this behavior. 
```
Latest version models for each requests stage. Only return models with current READY status. If no stages provided, returns the latest version for each stage, including "None".
```

https://www.mlflow.org/docs/latest/rest-api.html#get-latest-modelversions

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Done

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Okay, I will update the interface also to reflect this change.

andychow-db: Done. I've updated it in the interface PR.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Agreed. I've removed these two checks.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Good point. Not yet. This was only tested against the local mlflow server so far. We should add an integration test that can ensure the DBFS implementation works.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: I've removed the sentence.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowProtobufMapper.java
andychow-db: I've updated the interface to take in a set of stages now.

mlflow/java/client/src/main/java/org/mlflow/utils/CliCommand.java
andychow-db: Agreed. I'm happy to discuss this and if we can come up with a way to use the ArtifactRepository, I'd like to do it.

---
Below is some of my thoughts on it.

I originally intended to implement it using the `CliBasedArtifactRepository`. The problem I ran into is that the `runId` is part of the class's constructor. 

In order to refactor `CliBasedArtifactRepository` I would need to have a way to construct a `CliBasedArtifactRepository` without the runId, we have several options.

1. Pass in nulls into the existing constructor. I don't recommend this because it means that that while `downloadArtifactUri` may work, the other functions will only generate an error. 
2. Change the base `ArtifactRepository` interface to include `runId` as a parameter which would be a backwards breaking change. I don't recommend this because it will break anybody who is using the methods.
3. Extend `CliBasedArtifactRepository` and throw `UnsupportedOperationException`. I could but it turns everything into a runtime exception and will surprise users if they try to use it.
4. Create a new interface and implementing class. I don't recommend this either because it's not the `ArtifactRepository` interface at this point.

mlflow/java/client/src/test/java/org/mlflow/tracking/ModelRegistryMlflowClientTest.java
andychow-db: Removed

mlflow/java/client/src/test/java/org/mlflow/tracking/ModelRegistryMlflowClientTest.java
andychow-db: Removed

mlflow/java/client/src/test/java/org/mlflow/tracking/MlflowClientTest.java
andychow-db: Done

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: I have manually test this against the dogfood stage right now. I will create a follow-up task to implement an integration test in the databricks universe

mlflow/java/client/pom.xml
```diff
@@ -80,7 +85,7 @@
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
           <sourcepath>${project.basedir}/src/main/java</sourcepath>
-          <excludePackageNames>com.databricks.api.proto.databricks:org.mlflow.scalapb_interface:org.mlflow.tracking.samples:org.mlflow.artifacts</excludePackageNames>
+          <excludePackageNames>com.databricks.api.proto.databricks:org.mlflow.scalapb_interface:org.mlflow.tracking:org.mlflow.artifacts:org.mlflow.api</excludePackageNames>
```
sueann: what does this change do? exclude these packages from some lint rules?

mlflow/java/client/src/main/java/org/mlflow/artifacts/CliBasedArtifactRepository.java
```diff
@@ -140,6 +140,22 @@ public File downloadArtifacts() {
     return listArtifacts(null);
   }
 
+  /**
+   * Only available in the CliBasedArtifactRepository. Downloads an artifact to the local
```
sueann: Add experimental warning here too? Since the class and its methods do show up public docs (e.g. https://mlflow.org/docs/0.5.2/java_api/org/mlflow/artifacts/CliBasedArtifactRepository.html)

mlflow/java/client/src/main/java/org/mlflow/artifacts/CliBasedArtifactRepository.java
```diff
@@ -140,6 +140,22 @@ public File downloadArtifacts() {
...
+   * filesystem when provided with an artifact uri. This method should not be used directly
+   * by the user. Please use {@link org.mlflow.tracking.MlflowClient}
+   *
+   * @param artifactUri Artifact uri
+   * @return Directory/file of the artifact
+   */
+  public File downloadArtifactUri(String artifactUri) {
```
sueann: `downloadArtifactFromUri` might be more clear (since we are not downloading the URI itself).

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -1,18 +1,21 @@
 package org.mlflow.tracking;
 
+import com.google.common.collect.Lists;
 import org.apache.http.client.utils.URIBuilder;
-
+import org.mlflow.api.proto.ModelRegistry.*;
 import org.mlflow.api.proto.Service.*;
+import org.mlflow.artifacts.CliBasedArtifactRepository;
+import org.mlflow.tracking.creds.*;
```
sueann: nit: move imports to the alphabetically correct position

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -636,4 +643,149 @@ private ArtifactRepository getArtifactRepository(String runId) {
...
+   *    <pre>
+   *        File modelVersionFile = downloadModelVersion("model", 0);
+   *    </pre>
+   *
+   * @param modelName The name of the model
+   * @param version The version number of the model
+   * @return A local file or directory ({@ java.io.File}) containing model artifacts
+   */
+  public File downloadModelVersion(@Nonnull String modelName, long version) {
+    String downloadUri = getModelVersionDownloadUri(modelName,
```
sueann: nit: oneline

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowHttpCaller.java
```diff
@@ -128,6 +129,22 @@ String post(String path, String json) {
     }
   }
 
+  String patch(String path, String json) {
+    logger.debug("Sending PATCH " + path + ": " + json);
+    HttpPatch request = new HttpPatch();
+    fillRequestSettings(request, path);
```
sueann: there's a lot of copypasta between get, post and patch (especially the latter two) - we could consolidate, but probably ok as is.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowProtobufMapper.java
```diff
@@ -1,11 +1,14 @@
 package org.mlflow.tracking;
 
 import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
 import com.google.protobuf.MessageOrBuilder;
 import com.google.protobuf.util.JsonFormat;
 
 import java.lang.Iterable;
 
+import org.mlflow.api.proto.ModelRegistry;
```
sueann: is this import needed - for docstring?

mlflow/java/client/src/test/java/org/mlflow/tracking/MlflowClientTest.java
```diff
@@ -1,31 +1,49 @@
...
-import java.util.Arrays;
-import java.util.Set;
-import java.util.HashSet;
-import java.util.Stack;
-import java.util.Vector;
-import java.util.LinkedList;
-
 import com.google.common.collect.Lists;
 import org.apache.commons.io.FileUtils;
+import org.mlflow.api.proto.Service.CreateRun;
```
sueann: i assume intellij did whatever it wants with the imports? usually we recommend the following ordering of imports:
1. java.* and javax.*
2. scala.* (when scala)
3. Third-party libraries (org.*, com.*, etc)
4. Project classes (com.mlflow.* if you are working on Spark)

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowProtobufMapper.java
```diff
@@ -1,11 +1,14 @@
 package org.mlflow.tracking;
 
 import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
```
sueann: unused?

sueann (approved):
lgtm - just mostly nits inline. feel free to (squash and) merge once they're addressed. thanks!

mlflow/java/client/src/main/java/org/mlflow/artifacts/CliBasedArtifactRepository.java
andychow-db: Done

mlflow/java/client/src/main/java/org/mlflow/artifacts/CliBasedArtifactRepository.java
andychow-db: Done

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Done

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowProtobufMapper.java
andychow-db: No, I've removed it.

mlflow/java/client/src/test/java/org/mlflow/tracking/MlflowClientTest.java
andychow-db: Done

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
andychow-db: Sorry, I don't understand. Do you mean like this?

```
public File downloadModelVersion(...)  {

    String downaloadUri = getModelVersionDownloadFromUri(...);

}
```

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowHttpCaller.java
andychow-db: I've created a new private function that encapsulates the common logic here.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowProtobufMapper.java
andychow-db: Removed.

mlflow/java/client/pom.xml
andychow-db: The linter fails the build because the the protobuf code does not conform to the requirements. I added that line to prevent it from failing.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
sueann: ya that’s all i mean - it looks like the whole function call would fit in one line.

mlflow/java/client/src/main/java/org/mlflow/api/proto/ModelRegistry.java
```diff
@@ -883,6 +883,7 @@ public RegisteredModel parsePartialFrom(
     /**
      * <pre>
      * User that created this ``registered_model``
+     * NOTE: this field is not currently returned.
```
andychow-db: I had to regenerate the protobuf files for some reason.

andychow-db: I found the commit. It appears the protobuf was not regenerated
https://github.com/mlflow/mlflow/commit/88135f51a1192f654d28d4fb6736b3dff076ae1a

PR #2313, opened 2020-01-16
Adds an example of querying for the best model

AveshCSingh:

Note that this does not add documentation of finding the best model across multiple experiments. I feel this would be redundant, as we now have examples of:
(1) Find the best run
(2) Search for runs with particular parameters between experiments

From these examples, the user will infer how to find the best run across multiple experiments.

## How is this patch tested?

Docs were built and viewed in the browser.

- [Y] Yes. Give a description of this change to be included in the release notes for MLflow users.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [Y] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

- [Y] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section

### DIFF
--- a/docs/source/search-syntax.rst
+++ b/docs/source/search-syntax.rst
@@ -165,14 +165,14 @@ For example, if you'd like to identify the best `active` run from experiment ID 
 
 
 To get all active runs from experiments IDs 3, 4, and 17 that used a CNN model
-with 10 layers and had an accuracy of 94.5% or higher, use:
+with 10 layers and had a prediction accuracy of 94.5% or higher, use:
 
 .. code-block:: py
 
   from mlflow.tracking.client import MlflowClient
   from mlflow.entities import ViewType
 
-  query = "params.model = 'CNN' and params.layers = '10' and metrics.accuracy >= 0.945"
+  query = "params.model = 'CNN' and params.layers = '10' and metrics.`prediction accuracy` >= 0.945"
   runs = MlflowClient().search_runs(experiment_ids=["3", "4", "17"], filter_string=query, run_view_type=ViewType.ACTIVE_ONLY)
 
 To search all known experiments for any MLflow runs created using the Inception model architecture:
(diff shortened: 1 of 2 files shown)


docs/source/search-syntax.rst
```diff
@@ -148,7 +148,17 @@ search programmatically. You can specify the list of columns to order by
 optional ``DESC`` or ``ASC`` value; the default is ``ASC``. The default ordering is to sort by 
 ``start_time DESC``, then ``run_id``.
 
-For example, to get all `active` runs from experiments IDs 3, 4, and 17 that used a CNN model
+For example, to extract the best `active` run from experiment ID 0 by MAE, use:
```
smurching:
```suggestion
For example, say you want to identify the best model for production deployment. You can use search to find the best `finished` run from experiment ID 0 by accuracy:
```

(was thinking it might actually be good to pass `ViewType.FINISHED` here & search across finished runs, because the likely use case for this is to find a model to deploy etc, so you may only want to include models from finished training runs)

docs/source/search-syntax.rst
```diff
@@ -148,7 +148,17 @@ search programmatically. You can specify the list of columns to order by
...
+
+.. code-block:: py
+
+  from mlflow.tracking.client import MlflowClient
+  from mlflow.entities import ViewType
+
+  run = MlflowClient().search_runs("0", "", ViewType.ACTIVE_ONLY, max_results=1, order_by=["metrics.mae DESC"])[0]
```
smurching:
```suggestion
  best_run = MlflowClient().search_runs(experiment_ids="0", filter_string="", run_view_type=ViewType.FINISHED, max_results=1, order_by=["metrics.mae DESC"])[0]
```

Just updating the example to pass FINISHED as the view type plus a random suggestion to use keyword arguments for clarity

I prefer using the fluent API here (which returns run info as a pandas DF) as

smurching (commented):
Mostly looks good! Just small comments

smurching (approved):
LGTM, let's merge once tests pass

Issue #2274, opened 2020-01-08
[FR] Make deployment tool integration pluggable


hhsecond:
## Describe the proposal

Making integration with different deployment tools (like we have sagemaker) pluggable. Similar to the plugin system for Tracking & Artifact repository, a plugin system based on python's entrypoint groups which is then picked up by mlflow is one way of doing it

There will be three (or more in the future) allowed serving methods - `deploy`, `delete` & `run-local` (shamelessly copied from sagemaker). User can specify which method to call and which plugin to use. MLflow look for the plugin and then searches whether the plugin has that function defined or not. If defined, it calls the function with given parameters

#### From CLI
```mlflow redisai deploy ...```
We could also group it under `serving` subcommand but that makes the whole command ugly. It could look like ```mlflow serving redisai deploy```

#### From program
```mlflow.serving.deploy(plugin='redisai', *args, **kwargs)```

Links:
* PR: #2327 
* [API usage examples](https://gist.github.com/smurching/f00e3b771b2956da7281c427b5f5a681)
* [Design doc](https://docs.google.com/document/d/1KsyYO82OtHPEV4Uq8_ukDyZdBfgt7EuCtE2zFSQVea4/edit?usp=drive_web&ouid=109660911842186433457)

smurching:
Thanks @hhsecond for filing this, it'd be great to make MLflow model deployment more pluggable. 

It'd be helpful to fully list out the options & tradeoffs (pros/cons) for the CLI & programmatic API.

For example, another option to consider is exposing pluggable deployment functionality under the `mlflow models` CLI (e.g. `mlflow models deploy ...` and `mlflow.models.deploy(...)`), for ease of discoverability given that we already have `mlflow models serve` and `mlflow models predict` CLI commands. We could also consider exposing deploy functionality through `mlflow models serve` (currently deploys a model for local serving but could be extended to deploy elsewhere), although I think users might be confused by the name.

We should also explicitly write out answers to the following
* How will developers write/specify plugins (likely via a similar mechanism to [existing tracking/artifact plugins](https://github.com/smurching/mlflow/blob/contributor-guide-plugins/docs/source/plugins.rst), i.e. using the [Python entry-point mechanism](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins))? How will plugin names like “redisai” be mapped to deployment logic (Python entry-points should make this easy)?
[...]

PR #2327, opened 2020-01-21
Introduce pluggable API for deploying models to custom serving tools

hhsecond:
This PR implements the pluggable model deployment API proposed in https://github.com/mlflow/mlflow/issues/2274.

## How is this patch tested?
Unit tests for model deployment Python APIs and CLI.

Introduces a pluggable ``mlflow.deployments`` Python API and CLI for deploying MLflow models to custom serving tools.

- [ ] UI
- [x] CLI
- [x] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [x] Serving
- [ ] R
- [ ] Java
- [x] Python


smurching (commented):
@hhsecond thank you! Leaving initial comments

mlflow/deployments/interface.py
```diff
@@ -0,0 +1,38 @@
+from mlflow.deployments.plugin_manager import DeploymentPlugins
+
+listType = list
+plugin_store = DeploymentPlugins()
+
+
+# TODO: standardise exceptions
```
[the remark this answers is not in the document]
hhsecond: I see both standard python exceptions and `MlflowException` used in the source code. What's the rationale behind when to use `MlflowException`?

mlflow/deployments/interface.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def create(target, model_uri, flavor=None, *args, **kwargs):
+    deployment = plugin_store[target].create(model_uri, flavor, *args, **kwargs)
```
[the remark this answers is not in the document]
hhsecond: Make sense. The `KeyError` is raised from `plugin_store` if the `target` is not found. I'll change that to a `MlflowException` with the required error message. Would that work?

Although, raising the `KeyError` from `plugin_store` and re-raising that as an `MlflowException` might seem pythonic here, that'll make us write the `try` catch` block for each interface functions

mlflow/deployments/base_plugin.py
```diff
@@ -0,0 +1,31 @@
+import sys
+import abc
+
+
+if sys.version_info >= (3, 4):
```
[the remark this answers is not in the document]
hhsecond: Should we wait for that PR to merge? Or maybe I can take this up in a consecutive PR?

mlflow/deployments/base_plugin.py
```diff
@@ -0,0 +1,31 @@
...
+    ABC = abc.ABC
+else:
+    ABC = abc.ABCMeta('ABC', (), {})
+
+
+class BasePlugin(ABC):
```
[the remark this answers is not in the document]
hhsecond: Sure, documentation & test cases are on the way

mlflow/deployments/base_plugin.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @abc.abstractmethod
+    def create(self, *args, **kwargs):
```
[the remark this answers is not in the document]
hhsecond: Your assumption is right here. I'll add the mandatory arguments here as well and document it

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,77 @@
...
+deployment_target = click.option("--target", "-t", required=True, help="Deployment target")
+deployment_id = click.option("--id", "_deployment_id", required=True,
+                             help="Deployment ID for the deployment that needs to be deleted")
+context_settings = dict(allow_extra_args=True, ignore_unknown_options=True,)
+
+
+@click.group("deployments")
+def commands():
+    """
+    Deploy MLflow models using deployment plugin. It requires user to install the plugin
```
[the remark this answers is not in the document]
hhsecond: Make sense

hhsecond:
@smurching 
Thanks a lot for the early review. I'll update the PR along with documentation and test cases. So it is quite apparent that we'd need a follow-up PR(s) for those tasks we have mentioned above.
- Making one default deployment plugin and remove the existing deployment tools from the core of Mlflow
- Clean up after dropping python2 support

Another candidate that could go in is the clean up of plugin manager class. We have the same functionality named as `StoreRegistry` at `mlflow/tracking/registry.py`. We essentially need only one plugin manager class. This could also serve for any other future plugin system we implement.

hhsecond:
Also, about the documentation. Do you think we'd need a separate section for **MLFlow Deployments** and then point to the plugin section for anything related to plugin or being under the **MLFlow Plugins** is good enough?

mlflow/deployments/base_plugin.py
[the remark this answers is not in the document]
hhsecond: Also, It's better to notify you that the metaclass setup is broken in several places in MLFlow source code. It's because there are classes that use python 2.x style metaclass declaration (using `__metaclass__`) which is ignored in python 3.x. So when you migrate those metaclasses to python3x, there is a chance that child classes are going to break

mlflow/deployments/interface.py
```diff
@@ -0,0 +1,41 @@
...
+    plugin_store[target].delete(deployment_id, **kwargs)
+
+
+def update(target, deployment_id, model_uri=None, rollback=False, **kwargs):
+    if all((rollback, model_uri)):
+        raise RuntimeError("``update`` has got both ``model_uri`` and ``rollback``")
+    plugin_store[target].update(deployment_id, rollback, model_uri, **kwargs)
+
+
+# TODO: It's a good practise to avoid using ``list`` here.
```
hhsecond: Personally, I'd hate to use a reserved keyword here but I could come up with a better name than `list` here. Appreciate any help

mlflow/deployments/base_plugin.py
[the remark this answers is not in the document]
smurching: Thanks for the heads up, will take a look at the metaclass issue - IMO it's fine to take this up later unless #2534 merges first

mlflow/deployments/interface.py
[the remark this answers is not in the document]
smurching: There's not a huge difference, IMO they're mainly useful for allowing users to specifically catch errors thrown by MLflow vs other parts of their program. Other libraries (e.g. the popular requests library) [follow a similar pattern](https://2.python-requests.org/en/master/_modules/requests/exceptions/)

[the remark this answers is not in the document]
smurching: And makes sense that we should mention in the API method docstrings that plugins should throw `MlflowException`s on failures

mlflow/deployments/__init__.py
```diff
@@ -2,7 +2,7 @@
 from mlflow.deployments import interface
 
 
-# TODO: It's a good practise to avoid using ``list`` here. Looking for a better alternative
+# TODO: It's a good practise to avoid using ``list`` here
```
smurching: Thanks for bringing this up, IMO we could go with `create_deployment`, `delete_deployment`, `update_deployment`, `list_deployments`, `describe_deployment` here. This matches what we do e.g. in the MlflowClient class / fluent API (`mlflow.list_experiments`, `MlflowClient.list_experiments`, etc). While some OSS libraries do seem to actually define methods named `list` (see e.g. the [docker Python client](https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.list)), I agree that it'd be a bad idea (adding `from mlflow.deployments import *` would suddenly break all usages of lists in your program).

Alternatively, maybe there's a great synonym of  `list` that we can use instead, but I can't quite think of one.

mlflow/deployments/__init__.py
```diff
@@ -0,0 +1,24 @@
+from mlflow.deployments.base_plugin import BasePlugin
+from mlflow.deployments import interface
+
+
+# TODO: It's a good practise to avoid using ``list`` here
+
+__all__ = ['BasePlugin', 'create', 'delete', 'update', 'list', 'describe']
+
+
+def __dir__():
```
smurching: Hm, what's the purpose of defining `__dir__` and `__getattr__` here? Could we perform the lazy loading in interface.py itself?

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,82 @@
...
+@deployment_id
+@deployment_target
+def delete_cli(target, _deployment_id, **kwargs):
+    interface.delete(target, _deployment_id, **kwargs)
+    click.echo("Deployment {} is deleted".format(_deployment_id))
+
+
+@commands.command("update", context_settings=context_settings)
+@parse_custom_arguments
+@click.option("--rollback", help="Should the deployment be rolled back. Rollback is an "
```
smurching: Is there any disadvantage to just making `--rollback` a custom argument individual plugins can implement, rather than a standard argument that some plugins accept, but that other plugins handle by ignoring it or raising an exception?

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,82 @@
...
+@commands.command("create", context_settings=context_settings)
+@parse_custom_arguments
+@click.option("--flavor", "-f", help="Which flavor to be deployed. This will be auto "
+                                     "inferred if it's not given")
+@deployment_target
+@cli_args.MODEL_URI
+def create_cli(model_uri, target, flavor, **kwargs):
+    deployment = interface.create(target, model_uri, flavor, **kwargs)
+    # TODO: Support async here and everywhere requires
+    click.echo("\n{} deployment {} is created".format(deployment.flavor, deployment.id))
```
smurching:
```suggestion
    click.echo("\nCreated deployment of model at {} using flavor {}. Deployment ID (can be used to describe, update, and delete the model): {} is created".format(deployment.flavor, deployment.id))
```

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@commands.command("delete", context_settings=context_settings)
+@parse_custom_arguments
+@deployment_id
+@deployment_target
+def delete_cli(target, _deployment_id, **kwargs):
+    interface.delete(target, _deployment_id, **kwargs)
+    click.echo("Deployment {} is deleted".format(_deployment_id))
```
smurching:
```suggestion
    click.echo("Deleted deployment {} from target {}".format(_deployment_id, target))
```

mlflow/deployments/base_plugin.py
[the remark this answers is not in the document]
smurching: Sounds good! Would also be good to mention that the APIs / CLIs are experimental using the `mlflow.utils.experimental` annotation, see the [AzureML CLI](https://github.com/mlflow/mlflow/blob/branch-1.7/mlflow/azureml/cli.py#L47) for an example of marking a CLI as experimental, and the [model registry methods](https://github.com/mlflow/mlflow/blob/branch-1.5/mlflow/tracking/client.py)  for an example of marking an API experimental

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,82 @@
...
+              help="URI to the model. A local path, a 'runs:/' URI, or a"
+              " remote storage URI (e.g., an 's3://' URI). For more information"
+              " about supported remote URIs for model artifacts, see"
+              " https://mlflow.org/docs/latest/tracking.html"
+              "#artifact-stores")  # optional model_uri
+@deployment_id
+@deployment_target
+def update_cli(target, _deployment_id, rollback, model_uri, **kwargs):
+    interface.update(target, _deployment_id, rollback=rollback, model_uri=model_uri, **kwargs)
+    click.echo("Deployment {} is updated".format(_deployment_id))
```
smurching:
```suggestion
    click.echo("Updated deployment {} of target {}".format(_deployment_id, target))
```

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@commands.command("list", context_settings=context_settings)
+@parse_custom_arguments
+@deployment_target
+def list_cli(target, **kwargs):
+    ids = interface.list(target, **kwargs)
+    click.echo("List of all deployments:\n{}".format(ids))
```
smurching:
```suggestion
    click.echo("Deployment IDs for target {}:\n{}".format(target, ids))
```

mlflow/deployments/cli.py
hhsecond: The API will be still confusing. For instance, we don't need the `model_uri`. So `model_uri` in the case of `rollback` and hence need to keep it as an optional parameter to the `update` command. This makes `--help` of `update` says, `model_uri` is optional (even for a non-rollback workflow) but the plugin throws an error if `model_uri` is not given.
I mean, the confusion here is not a deal-breaker, we can live with that. But we can live with the `--rollback` option being there too. Anyway, I am good with both, LMK what do you think

mlflow/deployments/__init__.py
hhsecond: I don't think we can. Since those functions are defined in interface.py, the name scope of that file gets precedence over the `__getattr__`, so the function call won't reach `__getattr__`.
Even if we find a way to do that, we'll end up doing the plugin registration inside each interface function. We probably don't want that. 
Why do you prefer not doing this in `__init__.py`?

mlflow/deployments/__init__.py
hhsecond: `from mlflow.deployments import *` is exactly what I am afraid of. I can't find a good synonym too. Unless, no one has a better solution, I'll go with `list` itself then and probably prefer this usage all over the docs
```python
from mlflow import deployments
deployments.list()
```

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,103 @@
+import click
+from mlflow.utils import cli_args
+from mlflow.deployments import interface
+from mlflow.deployments.utils import parse_custom_arguments
+
+
+deployment_target = click.option("--target", "-t", required=True, help="Deployment target")
+deployment_id = click.option("--id", "_deployment_id", required=True,
+                             help="Deployment ID for the deployment that needs to be deleted")
```
smurching:
```suggestion
                             help="ID of the deployment to be deleted")
```

We should probably define this option for each command vs globally, this help string makes sense for the `delete` command but not for `update`, where "ID of the deployment to be updated" would make more sense

mlflow/deployments/plugin_manager.py
```diff
@@ -0,0 +1,105 @@
...
+import sys
+import abc
+
+import entrypoints
+from mlflow.deployments.base_plugin import BasePlugin
+from mlflow.exceptions import MlflowException
+from mlflow.protos.databricks_pb2 import RESOURCE_DOES_NOT_EXIST, INTERNAL_ERROR
+
+
+if sys.version_info >= (3, 4):
```
smurching: Should be able to remove this check now that we've dropped Python 2 support in #2534 :)

mlflow/deployments/cli.py
smurching: Ah gotcha. What about making the API of `update` mirror the update of `create`, but with optional arguments? 
The idea would be that a deployment is defined by the tuple of arguments passed to create (`model_uri`, `flavor`, `**kwargs`) and that users should therefore be able to independently update one or more members of this tuple (although IMO `model_uri` should be required in order to specify a new `flavor` via `update`, it doesn't seem to add much value to be able to update the flavor independently of the `model_uri` and would make plugin logic more complex).

In that scenario, users could specify `model_uri` and `flavor` arguments, as well as custom `kwargs` defined by the plugin. If the user specifies none of these (or specifies `flavor` without `model_uri`), it makes sense for the `update` command to throw - otherwise, the update command should just update the desired fields.

This mirrors the pattern followed elsewhere in MLflow, e.g. in `update_registered_model`: https://github.com/mlflow/mlflow/blob/e0e19c6c19734b1847c417b3d59a799d9e007336/mlflow/tracking/client.py#L375

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,103 @@
...
+context_settings = dict(allow_extra_args=True, ignore_unknown_options=True,)
+
+
+@click.group("deployments")
+def commands():
+    """
+    Deploy MLflow models.  Downstream functions calls the plugin registered for the
```
smurching:
```suggestion
    [experimental] Deploy MLflow models to custom targets.

    Downstream functions calls the plugin registered for the...
```

The idea is to make the `deployments` commands stand out as different (and experimental) when users run `mlflow --help`, which currently yields the following (note that the `models` & `deployments` CLIs sound pretty similar):

```
Usage: mlflow [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  artifacts    Upload, list, and download artifacts from an MLflow artifact...
  azureml      Serve models on Azure ML.
  db           Commands for managing an MLflow tracking database.
  deployments  Deploy MLflow models.
  experiments  Manage experiments.
  gc           ..
  models       Deploy MLflow models locally.
  run          Run an MLflow project from the given URI.
  runs         Manage runs.
  sagemaker    Serve models on SageMaker.
  server       Run the MLflow tracking server.
  ui           Launch the MLflow tracking UI for local viewing of run...
```

but would ideally yield something like:

```
Usage: mlflow [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  artifacts    Upload, list, and download artifacts from an MLflow artifact...
  azureml      Serve models on Azure ML.
  db           Commands for managing an MLflow tracking database.
  deployments  [experimental] Deploy MLflow models to custom targets.
  experiments  Manage experiments.
  gc           ..
  models       Deploy MLflow models locally.
  run          Run an MLflow project from the given URI.
  runs         Manage runs.
  sagemaker    Serve models on SageMaker.
  server       Run the MLflow tracking server.
  ui           Launch the MLflow tracking UI for local viewing of run...
```

mlflow/deployments/cli.py
smurching: IMO probably the best approach for supporting rollbacks (between adding it later as a separate command or keeping it as an argument to `update`) is just to leave them out entirely now, and think about how to support rollback functionality later on, especially as it's not super clear what the requirements around rollback functionality are for now (would be better to keep the interface simple IMO & collect user feedback on this).

If we do add rollback support later, it can be an API/CLI command separate from `update` that some plugins support, and for legacy plugins/plugins that don't implement `rollback`, users can always achieve a rollback by maintaining deployment history themselves & invoking `update` with appropriate arguments.

mlflow/deployments/plugin_manager.py
hhsecond: Awesome, What is the minimum version we are supporting? 3.5?

mlflow/deployments/cli.py
hhsecond: We are going back to our initial approach here. Here are my thoughts
1. About rollback, I am OK to defer the implementation to get user feedback
2. About the signature of update; if that's the case, why would we even need the update command. We can have a `mode` argument (like in sagemaker's current implementation) and then we can minimize the surface

smurching: Gotcha, yeah for 2) it still seems worthwhile to have an `update` command separate from `create`.

The difference between `update` here and in the sagemaker use case (where the CLI allows for `mlflow sagemaker deploy -m [replace/add/create]`  is that we want to be flexible to updating individual attributes of the deployment (e.g. for a model deployment on k8s we may want to specifically increase the number of model serving replicas). 

Moreover, the signatures of `create` and `update` are different - in the Sagemaker case, there's a common "app name" that defines a deployment & is shared across `create` and `update`, whereas in our case we allow the deployment ID to be defined by the deployment target, and don't know it at creation time

mlflow/deployments/plugin_manager.py
smurching: yep!

mlflow/deployments/__init__.py
smurching: Got it, mainly just didn't quite understand what this code was doing / it adds some complexity. Would you mind adding a clarifying comment to the code explaining your reasoning above?

mlflow/deployments/cli.py
hhsecond: Awesome, Thanks

mlflow/deployments/__init__.py
hhsecond: Will do

mlflow/deployments/__init__.py
smurching: Sorry, should have been clearer - I think it'd actually be better to name them `create_deployment`, `delete_deployment`, `update_deployment`, `list_deployments`, so usage would look like:

```
from mlflow.deployments import *
list_deployments()
```

or

```
from mlflow import deployments
deployments.list_deployments()
```

It makes the method names more verbose, but avoids the conflict with `list` and would also us to be consistent if we eventually promoted deployment functionality to the top-level MLflow fluent API, e.g. via APIs like `mlflow.create_deployment`, `mlflow.update_deployment`, etc.

hhsecond: Got it. However, I preferred `xxx` over `xxx_deployment` since that would be the replica of the CLI commands. Plus another major concern I have is about the human language aesthetic. For instance:
`list_deployments` is accurate because it can list more than one deployment but `create_deployments` is wrong because we are not creating more than one deployment (Note the plurality of the command)

smurching: I think we'd handle the pluralization by just using the appropriate pluralization for the use case (e.g. we'd define a singular `create_deployment` but a pluralized `list_deployments`)

mlflow/deployments/plugin_manager.py
hhsecond: I just figured pandas officially supports only 3.6+ and mlflow has a dependency on pandas. So shouldn't the minimal supported version must be 3.6?

mlflow/deployments/__init__.py
hhsecond: Restating our conversation in slack:
> Really?. Not a big fan of that because I think that makes users life miserable. But, that's probably a personal choice. I'll stick with your suggestion

mlflow/deployments/plugin_manager.py
smurching: Ah, yeah new releases of pandas only support 3.6 and above, people can still use MLflow with older versions of pandas on Python 3.5.2 (which will [reach EOL in September this year](https://devguide.python.org/#status-of-python-branches))

hhsecond: Alright, make sense. But setup.py will always tries to install the lates pandas now which will raise an error for those who has 3.5

gkorland:
@smurching any chance you got to review it?

smurching:
@hhsecond @gkorland thanks for the hard work on this PR / for chiming in (& especially for all the doc updates). I gave the plugin interface a try (trying to run through the user workflow of actually using a deployment plugin) & had the following comments on the API - I'm happy to help make some of the necessary changes to help get this PR in, here are the comments/let me know what you think:

1. We should accept custom target-specific arguments for deployment creation & update via a repeated `-C key=value (--config)` argument in the CLI & Python API, similar to the `-P key=value` CLI argument to `mlflow run`. IMO this makes it clearer how to pass custom target-specific arguments to the CLI - there will be a documented CLI argument for said target-specific args with a clear input format. Otherwise, some confusing/unexpected behavior can arise e.g. when trying to pass flags to the API:

```
➜  mlflow git:(deploy-plugin)  mlflow deployments create -t faketarget  -m models:/somemodel/uri --arg0 argval --arg1 anotherval --canipassaflag --whathappenswhenido
In create method, got kwargs: {'arg0': 'argval', 'arg1': 'anotherval', 'canipassaflag': '--whathappenswhenido'}
```
I started writing code for this change in [this commit](https://github.com/mlflow/mlflow/pull/2327/commits/b8b7fe304b973a000857e0423b6d4e1efbf8e913). 
2. Don't accept custom target-specific arguments in the `delete_deployment`, `describe_deployment`, & `list_deployment` APIs - I don't think there's a legitimate use case for target-specific args to these APIs yet
3. Rename `describe_deployment` -> `get_deployment` for parallel naming structure with other MLflow APIs (e.g. `mlflow.get_run`, `mlflow.get_experiment`)

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,154 @@
...
+target_name = click.option("--target-name", "t_name", required=True,
+                          help="Deployment target name.")
+deployment_name = click.option("--name", "name", required=True,
+                               help="Name of the deployment")
+parse_custom_arguments = click.option("--config", "-C", metavar="NAME=VALUE", multiple=True,
+                                      help="Extra target-specific config for the model deployment,"
+                                           " of the form -C name=value. See documentation/help for "
+                                           "your deployment target for a list of supported config"
+                                           " options.")
+context_settings = dict(allow_extra_args=True, ignore_unknown_options=True,)
```
smurching, line 37: I think we can remove this, as we don't need to allow custom CLI arguments to any of the commands anymore, right?

mlflow/deployments/interface.py
```diff
@@ -0,0 +1,55 @@
...
+def get_deploy_client(target_uri):
+    """
+    It calls the `get_deploy_client` method from the plugin and return an instance of a subclass of
+    :py:class:`mlflow.deployments.BaseDeploymentClient` that can be used to deploy models to the
+    specified target
+
+    :param: target_uri: URI of target to deploy to. Run ``mlflow deployments --help`` via the CLI
+                        for more information on supported deployment targets
+    """
+    # TODO: Maybe we should keep the separator as ":/" instead of "://"?
```
smurching, line 19: Agreed, `:/` seems like the right syntax for a URI, thanks for catching this

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,154 @@
...
+@target_name
+def target_help(t_name):
+    """
+    Specific help command for deployment plugins. This will call the `target_help` function from
+    the plugin to display the help string specific for each plugin
+    """
+    click.echo(interface.target_help(t_name))
+
+
+@commands.command("run-local", context_settings=context_settings)
```
smurching, line 151: Should update this to have the same arguments as the `create_deployment` CLI command

docs/source/models.rst
```diff
@@ -828,3 +828,44 @@ argument. The following values are supported:
...
+Also, see:
+
+.. code-block:: bash
+
+    mlflow deployments --help
+    mlflow deployments create --help
+    mlflow deployments delete --help
+    mlflow deployments update --help
+    mlflow deployments list --help
+    mlflow deployments get --help
```
smurching, line 871: Update to include `help` and `run-local` commands?

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,154 @@
...
+def _user_args_to_dict(user_list):
+    # Similar function in mlflow.cli is throwing exception on import
+    user_dict = {}
+    for s in user_list:
+        try:
+            name, value = s.split('=')
+        except ValueError:
+            # not enough values to unpack
+            raise click.BadOptionUsage("config", "Config options must be a pair and should be"
+                                                 "provided as ``--C key=value`` or "
```
smurching, line 15:
```suggestion
                                                 "provided as ``-C key=value`` or "
```

mlflow/deployments/interface.py
smurching: Oops, ^I was wrong, `://`  is actually the right syntax: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Definition

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,154 @@
...
+                                                 "``--config key=value``")
+        if name in user_dict:
+            raise click.ClickException("Repeated parameter: '{}'".format(name))
+        user_dict[name] = value
+    return user_dict
+
+
+target_uri = click.option("--target-uri", "t_uri", required=True,
+                          help="Deployment target URI. Check the documentation/help for each "
```
smurching, line 24: It'd be cool to include the name of all currently-installed plugins in the help string as a bulleted list, e.g something like:

```
help="""
Deployment target URI, e.g. 'sagemaker'.  Run `mlflow deployments help -t <target-name>`
for more details on the supported URI format and config options for a given target.

Currently-supported targets (see other supported deployment targets and instructions for installing support in  https://mlflow.org/docs/latest/plugins.html#community-plugins):
{supported_targets}
""".format(supported_targets="\n".join(["* %s" % target for target in list_of_targets_from_plugin_registry)]))
```

tests/deployments/test_deployments.py
```diff
@@ -0,0 +1,71 @@
...
+    ret2 = client.create_deployment(f_deployment_id, f_model_uri)
+    assert ret2['flavor'] is None
+
+
+def test_delete_success(patched_plugin_store):  # pylint: disable=W0613
+    client = deployments.get_deploy_client(f_target)
+    assert client.delete_deployment(f_deployment_id) is None
+
+
+def test_update_success(patched_plugin_store):  # pylint: disable=W0613
```
smurching, line 30: Can we also test the error case of trying to update with only `flavor` and no `model_uri`, and verify that an exception is thrown?

tests/deployments/test_cli.py
```diff
@@ -0,0 +1,69 @@
...
+
+def test_target_help(patched_plugin_store):
+    runner = CliRunner()
+    res = runner.invoke(cli.target_help, ["--target-name", f_target])
+    assert "Target help is called".format(f_target) in res.stdout
+
+
+def test_run_local(patched_plugin_store):
+    runner = CliRunner()
+    res = runner.invoke(cli.run_local, [])
```
smurching, line 69: Maybe also assert on the output of `run_local` here, as we do for the `help` command?

mlflow/deployments/interface.py
```diff
@@ -0,0 +1,55 @@
+from mlflow.deployments.plugin_manager import DeploymentPlugins
+from mlflow.utils import experimental
+
+
+listType = list
+plugin_store = DeploymentPlugins(auto_register=True)
+
+
+@experimental
```
smurching, line 9: Given that we've gotten broader community agreement on the APIs, we can remove the experimental tags from these methods :)

tests/deployments/test_deployments.py
smurching: Ah, I guess we don't actually enforce that that `model_uri` is specified when `flavor` is specified in `BaseDeploymentClient`, seems ok not to test then.

tests/deployments/test_cli.py
```diff
@@ -0,0 +1,69 @@
...
+from mlflow.deployments import cli
+
+
+f_model_uri = 'fake_model_uri'
+f_name = 'fake_deployment_id'
+f_flavor = 'fake_flavor'
+f_target = 'fake_target'
+
+
+def test_create(patched_plugin_store):  # pylint: disable=W0613
```
smurching, line 11: Could we add a test that uses the shorthand form of the flavor, model URI, and target URI arguments (e.g. `-f flavor -m model_uri -t target`)?

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,154 @@
...
+    for key, val in desc.items():
+        click.echo("{}: {}".format(key, val))
+    click.echo('\n')
+
+
+@commands.command("help", context_settings=context_settings)
+@target_name
+def target_help(t_name):
+    """
+    Specific help command for deployment plugins. This will call the `target_help` function from
```
smurching, line 145:
```suggestion
    Print help information for a deployment target, e.g. the supported target URI format and supported config
    options for creating deployments.
```

mlflow/deployments/cli.py
```diff
@@ -0,0 +1,154 @@
+import click
```
smurching, line 1: We should also add the `predict` function to the CLI + tests, but I think that can come in a follow-up to help expedite merging this PR first :)

tests/deployments/conftest.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+import entrypoints
+from . import dummy_plugin
+
+
+f_target = 'fake_target'
+
+
+def custom_entrypoint_group_all(name):
```
smurching, line 9: I think it might simplify the tests + docs if we actually just included this as a real plugin under https://github.com/mlflow/mlflow/tree/master/tests/resources/mlflow-test-plugin/mlflow_test_plugin & registered it as an entrypoint in https://github.com/mlflow/mlflow/blob/master/tests/resources/mlflow-test-plugin/setup.py. Then we wouldn't need to mock out entrypoint logic here & could reference the fake deployment plugin in the main plugin docs

mlflow/deployments/base.py
```diff
@@ -0,0 +1,168 @@
...
+      "sagemaker://<aws-cli-profile-name>", where aws-cli-profile-name is the name of an AWS
+      CLI profile https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html)
+    * Any other target-specific details.
+    """
+    raise NotImplementedError("This function should be implemented in the deployment plugin. It is"
+                              "kept here only for documentation purpose and shouldn't be used in"
+                              "your application")
+
+
+def get_deploy_client(target_uri):
```
smurching, line 46: ``get_deploy_client`` actually should not be defined by plugins - instead, we should look up & construct the appropriate subclass of ``BaseDeploymentClient`` based on the provided target URI (e.g. if the scheme of the URI is `redisai`, delegate to the Sagemaker plugin, etc). See the pattern used in the [ArtifactRepositoryRegistry](https://github.com/mlflow/mlflow/blob/1eefa145d62c64cc253b2ba3108bbcbbb852eabc/mlflow/store/artifact/artifact_repository_registry.py#L52-L71) for example

mlflow/deployments/interface.py
hhsecond: Alright, let's go with `://` (although we have `:/` for models and runs)

mlflow/deployments/cli.py
hhsecond: Ok, should we change the command to `deploy-local` or `create-local` then?

mlflow/deployments/cli.py
hhsecond: Currently, supported or currently installed in the user's machine? Because we don't have an automated way of getting currently supported plugins unless we hardcode it there

mlflow/deployments/cli.py
smurching: > Also our model uri follows `:/` as the separator and we are doing `://` as the separator here

This is a good point, using :/ makes more sense to retain parity with other "URI"s in MLflow. I also realize I misread [this URI definition](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Definition) - in general, something like "file:/" or "file:///" is a valid URI, but "file://" in fact is not.

> Ok, should we change the command to deploy-local or create-local then?

`deploy-local` seems reasonable, although I actually slightly prefer `run-local` as it avoids repeating the word "deploy" in the CLI command (e.g. compare `mlflow deployments run-local -t redisai ...` to `mlflow deployments deploy-local -t redisai ...`. I prefer both of these to `create-local` as it suggests that there should also be equivalent CRUD APIs for managing local deployments (e.g. `list-local`, `delete-local`, etc commands)

smurching: > Also, we don't really need "target URI" for run local, we just need the name of target plugin. Same is applicable in target_help. May be another parameter called --target-name. I don't really like target name on the first look but then advertising --target as target URI wouldn't do a good job here as well

On this point, I think for target_help it'd be good to still accept any valid target URI & show the appropriate help (e.g. if someone tells you to deploy models to `redisai:/prod`, it's nice to be able to use target help on `redisai:/prod` see the help printed vs an error message).

For the `run_local` method, it's a bit trickier, as `mlflow deployments run-local -t redisai:/prod ...` would actually deploy locally vs to your remote target - I think this is also ok though, it's nice to be able to reuse the same target URI across commands/have parallel structure across the APIs, if the user does specify a URI with a path (e.g. `redisai:/prod`) vs just a target name (like `redisai`), we can print a warning to let them know that the model is being deployed locally - no need to add the warning in this PR, we can add it in a follow-up.

mlflow/deployments/base.py
```diff
@@ -1,23 +1,28 @@
...
+1. Client class subclassed from :py:class:`BaseDeploymentClient`
+2. :py:func:`run_local` for test run the deployment on local system
+3. :py:func:`target_help` for displaying the help message about the plugin
 """
 
 import abc
 
 
-def run_local(model_uri, flavor=None, config=None):
+def run_local(target, name, model_uri, flavor=None, config=None):
```
smurching, line 14:
```suggestion
def run_local(target, model_uri, flavor=None, config=None):
```

We don't use name for local deployments / allow you to manage deployments by name for local deployments, so should remove it from the API

mlflow/deployments/base.py
```diff
@@ -1,23 +1,28 @@
...
     """
-    Deploys the specified model locally, for testing. This method should be defined within the module
-    specified by the plugin author.
+    Deploys the specified model locally, for testing. This function should be defined
+    within the module specified by the plugin author.
 
+    :param target: Which target to use. This information is used to call the appropriate plugin
+    :param name:  Unique name to use for deployment. If another deployment exists with the same
+                     name, create_deployment will raise a
+                     `:py:class:mlflow.exceptions.MlflowException`
```
smurching, line 22:
```suggestion
```

mlflow/deployments/cli.py
hhsecond: Awesome. I'll go with `:/` then.
About naming, I prefer `run-local` too, but it doesn't logically make sense because you are not really `running` anything, right? You are deploying it to a local instance and then "run" it later

mlflow/deployments/base.py
hhsecond: can't user have multiple local deployments? If we don't allow names, how'd we point to a particular deployment? For instance, in RedisAI, every deployment is a `MODELSET` command that sets the model to a redis key. While the plugin developer can create the name/key implicitly, I think it still makes sense to have that option enabled to handle multiple local deployments. What's the rationale behind not having `name` option?

mlflow/deployments/cli.py
hhsecond: `target_uri` suggestion is sensible. I'll do that

mlflow/deployments/base.py
smurching: Got it, the redisai use case makes sense - for most deployment targets, deploying locally involves just running a server on some port, so users can reference locally-deployed models via their URLs (e.g. I could have one model deployed locally at http://localhost:5000 and another at http://localhost:5001).

I think it's fine to keep it in - was previously thinking all targets would ignore the `name` when deploying locally, in which case it wouldn't be a particularly useful field to keep.

hhsecond: Awesome. I think it could even be useful to craft the URI like http://localhost:5000/name with the given `name` (we do something like this for a custom made REST API for one of our clients)

mlflow/deployments/interface.py
```diff
@@ -1,47 +1,55 @@
...
+
+    .. Note::
+        The plugin should only have one child class of
+        :py:class:`mlflow.deployments.BaseDeploymentClient` else it throws.
 
     :param: target_uri: URI of target to deploy to. Run ``mlflow deployments --help`` via the CLI
                         for more information on supported deployment targets
     """
-    # TODO: Maybe we should keep the separator as ":/" instead of "://"?
     target = target_uri.split("://")[0]
```
smurching: As commented above, can we use logic like that in the [RunsArtifactRepository](https://github.com/mlflow/mlflow/blob/07472fe6fe3b162ff1afbd75969c3b0b35851ee0/mlflow/store/artifact/runs_artifact_repo.py#L36-L50) to verify that the target URI is a valid URI before trying to index into the plugin store?

mlflow/deployments/cli.py
smurching: Yeah good question - for most deployment targets (sagemaker, kube, azureml, gcp, etc) the user will be running a local server that hosts their model behind a REST API - maybe guess in the redisai case we can make the same argument (that we run a local redisai server for the user if none exists & host the model on it)?

mlflow/deployments/interface.py
hhsecond: Sounds good

mlflow/deployments/cli.py
hhsecond: Ok, that sounds like a good explanation. Thank you

mlflow/deployments/cli.py
```diff
@@ -12,29 +12,34 @@ def _user_args_to_dict(user_list):
...
+                              help="Deployment target URI. Check the documentation/help for each "
+                                   "plugin to understand the target URI format the plugins "
+                                   "expect. Run "
+                                   "`mlflow deployments help --target-name <target-name>` for "
+                                   "more details on the supported URI format and config options "
+                                   "for a given target."
+                                   "Support is currently installed for the following targets (see"
+                                   " other deployment targets and installation instructions in "
+                                   "https://mlflow.org/docs/latest/plugins.html#community-plugins"
+                                   "): {}".format(registered))
```
smurching, line 35: The formatting of the helpstring for the `--target` argument currently looks a bit odd (looks like `click` ignores the newline characters), maybe we could make it more readable via something like: 

```suggestion
registered = [target for target in interface.plugin_store.registry]

target_details = click.option("--target", "-t", required=True,
                              help="""
                                   Deployment target URI. Check the documentation/help for each "
                                   plugin to understand the target URI format the plugins 
                                   expect. Run 
                                   `mlflow deployments help --target-name <target-name>` for 
                                   more details on the supported URI format and config options 
                                   for a given target. 
                                   Support is currently installed for the following targets: 
                                   {}. 
                                   
                                   See other deployment targets and installation instructions in 
                                   https://mlflow.org/docs/latest/plugins.html#community-plugins
                                   """.format(registered))
```

smurching (approved):
LGTM thanks @hhsecond for all the hard work on this over the past few months!

AveshCSingh:
+1! Thank you for implementing this, @hhsecond!

It looks like the links in the [Deployment to Custom Targets](https://github.com/mlflow/mlflow/blob/master/docs/source/models.rst#deployment-to-custom-targets) section are invalid. Would you mind sending a follow-up to fix them?

AveshCSingh:
(I think you just need to change .html to .rst)

hhsecond:
@AveshCSingh I am not an rst expert but I think (I could be wrong here) changing to `.rst` would break the documentation (at mlflow.org/docs). Also, this is a standard used by tools like rst2html and has been used across the current mlflow documentation. May be I can use `:ref:`? @smurching

AveshCSingh:
Ah, you are right. This is already the behavior of our existing links, and it should work fine on the published documentation. Please ignore my comment.

mlflow/deployments/base.py
```diff
@@ -0,0 +1,188 @@
...
+   and deployment config
+"""
+
+import abc
+
+from mlflow.utils import experimental
+
+
+@experimental
+def run_local(target, name, model_uri, flavor=None, config=None):  # pylint: disable=W0613
```
Ark-kun, line 18: How should the plugin use the `target` argument?

PR #2352, opened 2020-01-28
Fix killed status constraint upgrade

### LINKED ISSUE #2330: [SETUP-BUG] Table schema in latest_schema.sql gives errors on Microsoft SQL Server
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md)
for information on what types of issues we address.

Please fill in this template and do not delete it unless you are sure your issue is outside its scope.

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:  Ubuntu 18.4 / Microsoft Windows 10 
- **MLflow installed from (source or binary)**:	source
- **MLflow version (run ``mlflow --version``)**:	1.5.0
- **Python version**:					        Python 3.7.3
- **Database version**:				                Microsoft SQL Azure (RTM) - 12.0.2000.8		

Run the schema from tests/resources/db/latest_schema.sql on Microsoft SQL database.
For example:

•	Start Microsoft SQL Server Management Studio
•	Connect to the database where mlflow tables should be created
•	Start SQL editor (New Query)
•	Cut and paste code from latest_schema.sql (or just the code snippet below) into the SQL editor
•	Execute the SQL

CREATE TABLE metrics_test (
        key VARCHAR(250) NOT NULL,
        value FLOAT NOT NULL,
        timestamp BIGINT NOT NULL,
        run_uuid VARCHAR(32) NOT NULL,
        step BIGINT DEFAULT '0' NOT NULL,
        is_nan BOOLEAN DEFAULT '0' NOT NULL,
        CONSTRAINT metric_pk PRIMARY KEY (key, value, timestamp, run_uuid, step, is_nan),
        FOREIGN KEY(run_uuid) REFERENCES runs (run_uuid)
)

Errors encountered are displayed below. 

Issue 1:  Column names:

key VARCHAR(250) NOT NULL,
Msg 156, Level 15, State 1, Line 11
Incorrect syntax near the keyword 'key'.

Fix applied:	Use square brackets for keywords, for example:

                        [key] VARCHAR(250) NOT NULL,

Issue 2:  Boolean data type:

is_nan BOOLEAN DEFAULT '0' NOT NULL,
Msg 2715, Level 16, State 6, Line 90
Column, parameter, or variable #5: Cannot find data type BOOLEAN.

Fix applied:	Use data type BIT instead of BOOLEAN, for example:

                        is_nan BIT DEFAULT '0' NOT NULL,

Abstract

The table schema in tests/resources/db/latest_schema.sql gives errors when Microsoft SQL database is used as the backend store.
What database types are supported by mlflow? Is Microsoft SQL Server database supported?
Can we amend the mlflow database schema so it works with Microsoft SQL database?

Additional information

I have successfully installed mlflow version 0.91 in May 2019 but had to amend the table schema first as I was getting syntax errors. I am now trying to install the latest version and still have issues.

One of the issues I encountered last year was with 'lifecycle_stage' column being duplicated in two tables, experiments and runs. I can see this has now been corrected in version 1.0. Thank you.

I am still encountering syntax errors when the table column use SQL reserved words i.e. ‘key’.
In addition, Microsoft SQL Server does not support the BOOLEAN type.
Is Microsoft SQL database supported as the backend store in mlflow?

dbczumar:

This PR fixes #2330 by patching the migration that adds the `KILLED` run state to the collection of constraints imposed on the `runs` table in the `SqlAlchemyStore` for MLflow tracking.

## How is this patch tested?

Manual testing of the fresh database creation and database migration flow from 1.4 -> `master` against MySQL versions 8.0.15 and 8.0.18, SQL Server, sqlite3, and Postgres.

## Release Notes

Fixed a migration bug in the SqlAlchemyStore that prevented users from running MLflow Tracking on SQL Server.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/mlflow/store/db_migrations/versions/cfd24bdc0731_update_run_status_constraint_with_killed.py
+++ b/mlflow/store/db_migrations/versions/cfd24bdc0731_update_run_status_constraint_with_killed.py
@@ -6,9 +6,13 @@ Create Date: 2019-10-11 15:55:10.853449
 
 """
 from alembic import op, context
-from mlflow.entities import RunStatus
+from mlflow.entities import RunStatus, ViewType
+from mlflow.entities.lifecycle_stage import LifecycleStage
 from sqlalchemy import CheckConstraint, Enum
 from sqlalchemy.engine.reflection import Inspector
+
+from mlflow.store.tracking.dbmodels.models import SqlRun, SourceTypes
+
 
 # revision identifiers, used by Alembic.
 revision = 'cfd24bdc0731'
@@ -16,7 +20,7 @@ down_revision = '2b4d017a5e9b'
 branch_labels = None
 depends_on = None
 
-new_run_status = [
+new_run_statuses = [
     RunStatus.to_string(RunStatus.SCHEDULED),
     RunStatus.to_string(RunStatus.FAILED),
     RunStatus.to_string(RunStatus.FINISHED),
@@ -24,7 +28,7 @@ new_run_status = [
     RunStatus.to_string(RunStatus.KILLED)
 ]
 
-previous_run_status = [
+previous_run_statuses = [
     RunStatus.to_string(RunStatus.SCHEDULED),
     RunStatus.to_string(RunStatus.FAILED),
     RunStatus.to_string(RunStatus.FINISHED),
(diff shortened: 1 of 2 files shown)


mlflow/store/db_migrations/versions/33f7226d03b2_fix_killed_status_constraint.py
```diff
@@ -61,7 +61,6 @@ def upgrade():
         # database-backend-dependent enum (see https://docs.sqlalchemy.org/en/13/core/
         # type_basics.html#sqlalchemy.types.Enum.params.native_enum)
         batch_op.alter_column("status",
-                              existing_type=String,
```
smurching: As per alembic docs, this seems fine to do:

> existing_type – Optional; a TypeEngine type object to specify the previous type. This is required for all MySQL column alter operations that don’t otherwise specify a new type, as well as for when nullability is being changed on a SQL Server column. It is also used if the type is a so-called SQLlchemy “schema” type which may define a constraint (i.e. Boolean, Enum), so that the constraint can be dropped.

mlflow/store/db_migrations/versions/cfd24bdc0731_update_run_status_constraint_with_killed.py
```diff
@@ -5,70 +5,51 @@
...
-                                  type_=Enum(*new_run_status, create_constraint=True,
-                                             constraint_name="status", native_enum=False))
+    with op.batch_alter_table("runs", table_args=check_constraint_table_args) as batch_op:
+        # Transform the "status" column to an `Enum` and define a new check constraint. Specify
+        # `native_enum=False` to create a check constraint rather than a
+        # database-backend-dependent enum (see https://docs.sqlalchemy.org/en/13/core/
+        # type_basics.html#sqlalchemy.types.Enum.params.native_enum)
+        batch_op.alter_column("status",
+                              type_=Enum(*new_run_statuses, create_constraint=True,
+                                         constraint_name="status", native_enum=False))
```
smurching: Somehow the constraint name here doesn't seem to be honored - as a fresh user, on MySQL I get a constraint named "runs_chk_1":

```
  CONSTRAINT `runs_chk_1` CHECK ((`status` in (_utf8mb4'SCHEDULED',_utf8mb4'FAILED',_utf8mb4'FINISHED',_utf8mb4'RUNNING',_utf8mb4'KILLED'))),
```

& @AveshCSingh gets a different constraint name on postgres. As far as I can tell, this is what seems to be happening:
* For a MySQL user who previously upgraded to 1.5.0, we (in the old migration logic) created a new constraint w/ an unexpected name (e.g. `runs_chk_1`) & the old `status` constraint remains around. Upgrading to 1.6.0 then drops the `status` constraint, leaving behind `runs_chk_1`)
* For a new user who launches the server, we end up doing the same thing - creating a new constraint here with name like `runs_chk_1`, and then dropping the old status constraint in the following migration.

Also, my status column weirdly is still a VARCHAR on a fresh MySQL DB, maybe this is expected?

```
  `status` varchar(9) DEFAULT NULL,
```

smurching: @AveshCSingh investigated, looks like `constraint_name` isn't a valid argument to the [constructor of Enum](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Enum.__init__) columns in alembic, which is likely why it's not honored. We think this is ok though - if we need to update RunStatus in the future, we probably don't need to directly reference the constraint name/drop it, but instead can just alter the column with a new set of allowed enum values.

smurching: I guess the main actionable thing here then is just to remove the `constraint_name` arg as it's confusing.

smurching (commented):
One suggestion in https://github.com/mlflow/mlflow/pull/2352/files#r372134874, otherwise LGTM! Avesh & I QA'd on fresh & existing (1.5.0) MySQL & Postgres DBs

mlflow/store/db_migrations/versions/cfd24bdc0731_update_run_status_constraint_with_killed.py
dbczumar: @smurching @AveshCSingh Thanks for catching the `constraint_name` issue. I've removed this no-op parameter from the `Enum` constructor!

smurching (approved):
LGTM

AveshCSingh (approved):

PR #2358, opened 2020-01-29
Fix protobuf generation

jcuquemelle:

proto generation has been silently broken, as the test doesn't fail if protobufs have not been changed

at the end of the python small test, the test-generate-proto fails without breaking Travis

```
./test-generate-protos.sh
+++protoc --version
++PROTOC_VERSION='libprotoc 3.6.0'
++'[' 'libprotoc 3.6.0' '!=' 'libprotoc 3.6.0'
generate-protos.sh: line 5: [: missing `]'
++'libprotoc 3.6.0' '!=' 'libprotoc 3.6.1' ']'
generate-protos.sh: line 5: libprotoc 3.6.0: command not found
```

## How is this patch tested?

proto generation works in Travis test


dbczumar (approved):
LGTM! Thanks @jcuquemelle

PR #2369, opened 2020-01-30
Add log model endpoint.

tomasatdatabricks:
Introduce a new api to keep track of logged models for a run. The model infos are kept in a " _mlflow.models_ tag as a list in json format.

## How is this patch tested?
Unit test for sql alchemy store, file store and rest store.

## Release Notes
Introduce a new api to keep track of logged models in the tracking store. The actual models are still logged via artifact repository, however, the new endpoint allows MLflow clients to export some basic model info to the tracking store. Tracking store is expected to maintain a list of model infos (contents of MLmodel files)  for models logged from within the given run for each run. The data is kept in a "mlflow.models" tag and is stored as json. 

This API is experimental and can change in the future.

- [ ] UI
- [ ] CLI
- [x] API
- [x] REST-API
- [ ] Examples
- [ ] Docs
- [x] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python


mlflow/models/__init__.py
```diff
@@ -42,9 +48,15 @@ def add_flavor(self, name, **params):
         self.flavors[name] = params
         return self
 
+    def to_dict(self):
+        return self.__dict__
```
smurching: Was looking at the [Python docs](https://docs.python.org/3/library/stdtypes.html#object.__dict__) for this, will/does this also include attributes of a `Model` instance that are functions etc?

tomasatdatabricks: Good question, it does not include methods, but it does include whatever you assign to the objects even if it was a function. Any weird objects such as lambdas may fail to serialize to json, however, we use the same mechanism to export Model to yaml format, so such Model would not be loggable anyways.

mlflow/models/__init__.py
```diff
@@ -81,6 +93,17 @@ def log(cls, artifact_path, flavor, registered_model_name=None, **kwargs):
             mlflow_model = cls(artifact_path=artifact_path, run_id=run_id)
             flavor.save_model(path=local_path, mlflow_model=mlflow_model, **kwargs)
             mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
+            try:
+                mlflow.tracking.fluent._record_logged_model(mlflow_model)
+            except MlflowException as e:
+                if e.error_code == databricks_pb2.ErrorCode.Name(
+                        databricks_pb2.ENDPOINT_NOT_FOUND):
+                    _logger.warning(
+                        "The current store does not support log model api. "
```
smurching:
```suggestion
                        "MLflow clients 1.7.0 and above attempt to log models to an artifact store and record model "
                        "metadata to the tracking store. The model has been logged to the artifact store under %s, "
                        "but the current tracking store does not support recording model metadata, so model metadata "
                         "will not be logged to the tracking store. If logging to a server via REST, consider "
                         "upgrading the server version to MLflow 1.7.0 or above." % mlflow.get_artifact_uri()) 
```

I think the wording in the last sentence is debatable, but would be good to give users an actionable warning message so they know to upgrade their server if that's the issue (I'm assuming that's the most common case)

mlflow/protos/service.proto
```diff
@@ -471,6 +471,24 @@ service MlflowService {
       rpc_doc_title: "Log Batch",
     };
   }
+
+  // .. note::
+  //     Experimental: This API may change or be removed in a future release without warning.
+  rpc logModel (LogModel) returns (LogModel.Response) {
+    option (rpc) = {
```
smurching: Why do we have both the preview & non-preview paths here? Happy to discuss offline

mlflow/store/tracking/abstract_store.py
```diff
@@ -271,3 +271,20 @@ def log_batch(self, run_id, metrics, params, tags):
         :return: None.
         """
         pass
+
+    @abstractmethod
+    def record_logged_model(self, run_id, mlflow_model):
+        """
+        Record logged model information with tracking store. The list of logged model infos is
+        maintained in a mlflow.models tag in json format.
```
smurching:
```suggestion
        maintained in a mlflow.models tag in JSON format.
```

Is it worth specifying the initial schema of logged model info?

mlflow/store/tracking/abstract_store.py
```diff
@@ -271,3 +271,20 @@ def log_batch(self, run_id, metrics, params, tags):
         :return: None.
         """
         pass
+
+    @abstractmethod
+    def record_logged_model(self, run_id, mlflow_model):
```
smurching: Can we add the @mlflow.utils.experimental annotation here?

mlflow/store/tracking/abstract_store.py
```diff
@@ -271,3 +271,20 @@ def log_batch(self, run_id, metrics, params, tags):
...
+        """
+        Record logged model information with tracking store. The list of logged model infos is
+        maintained in a mlflow.models tag in json format.
+
+        Note: The actual models are logged as artifacts via artifact repository.
+
+        :param run_id: String id for the run
+        :param mlflow_model: Model object to be recorded.
+
+        NB: This api is experimental and may change in the future.
```
smurching:
```suggestion
        NB: This API is experimental and may change in the future. The default implementation is a no-op.
```

mlflow/tracking/_tracking_service/client.py
```diff
@@ -230,6 +231,11 @@ def log_batch(self, run_id, metrics=(), params=(), tags=()):
             _validate_tag_name(tag.key)
         self.store.log_batch(run_id=run_id, metrics=metrics, params=params, tags=tags)
 
+    def record_logged_model(self, run_id, mlflow_model):
+        if not isinstance(mlflow_model, Model):
+            raise Exception("wrong model")
```
smurching: Could we have a more informative message here?

mlflow/tracking/client.py
```diff
@@ -259,6 +259,9 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
         """
         self._tracking_client.log_artifacts(run_id, local_dir, artifact_path)
 
+    def record_logged_model(self, run_id, mlflow_model):
```
smurching: Let's add a docstring/annotate this as experimental?

smurching (commented):
Looks pretty good, seems like we should add tests for the behavior of the log model endpoint in the FileStore/SQLAlchemyStore/RestStore?

mlflow/models/__init__.py
tomasatdatabricks: Yeah makes sense, thanks

mlflow/protos/service.proto
tomasatdatabricks: Just because we have it for all the other endpoints. I did not want this to behave differently from the others (e.g. I am not sure if R client does not use preview still?) We should remove them all at once I think.

mlflow/tracking/_tracking_service/client.py
tomasatdatabricks: Oh yeah, I meant to fix this one, thx for the catch

mlflow/tracking/client.py
tomasatdatabricks: Yeah I think it's safer to make it private.

mlflow/store/tracking/abstract_store.py
tomasatdatabricks: yeah, good point.

tomasatdatabricks:
Thanks for the review @smurching! What do you mean by adding tests? There are tests for file, sql and rest store? Do you think they do not cover everything?

mlflow/server/handlers.py
```diff
@@ -443,6 +450,28 @@ def _log_batch():
...
 
 
+@catch_mlflow_exception
+def _log_model():
+    request_message = _get_request_message(LogModel())
+    try:
+        _logger.error(request_message.model_json)
+        model = json.loads(request_message.model_json)
+    except JSONDecodeError:
+        raise MlflowException("Model info is not a valid json.", error_code=INVALID_PARAMETER_VALUE)
```
smurching:
```suggestion
        raise MlflowException("Model info '%s' is not a valid json." % request_message.model_json, error_code=INVALID_PARAMETER_VALUE)
```

Include the bad JSON in the error message for ease of debugging?

mlflow/tracking/fluent.py
```diff
@@ -323,6 +323,11 @@ def log_artifacts(local_dir, artifact_path=None):
     MlflowClient().log_artifacts(run_id, local_dir, artifact_path)
 
 
+def _record_logged_model(mlflow_model):
```
smurching: I wonder if we should remove this, it's a private method that's not used anywhere

mlflow/models/__init__.py
```diff
@@ -81,6 +101,20 @@ def log(cls, artifact_path, flavor, registered_model_name=None, **kwargs):
             mlflow_model = cls(artifact_path=artifact_path, run_id=run_id)
             flavor.save_model(path=local_path, mlflow_model=mlflow_model, **kwargs)
             mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
+            try:
+                mlflow.tracking.fluent._record_logged_model(mlflow_model)
+            except MlflowException as e:
+                if e.error_code == databricks_pb2.ErrorCode.Name(
```
smurching: Hm unfortunately the old MLflow 1.6.0 flask server doesn't return an `ENDPOINT_NOT_FOUND`, but rather an `INTERNAL_ERROR`, so we might have to just swallow `INTERNAL_ERROR`s as well. I tested by attempting to log a model using the client from this PR against the 1.6.0 server (adding a print statement here to log the error code, which you can see in the output below):

```
>>> import mlflow; import mlflow.pyfunc; mlflow.set_tracking_uri("http://localhost:5000"); mlflow.pyfunc.log_model("abc", loader_module="mlflow.pyfunc")
/Users/sid.murching/miniconda3/envs/mlflow-python3/lib/python3.6/site-packages/pyspark/cloudpickle.py:47: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
@SID got error code INTERNAL_ERROR
  File "/Users/sid.murching/code/mlflow/mlflow/pyfunc/__init__.py", line 689, in log_model
    registered_model_name=registered_model_name)
  File "/Users/sid.murching/code/mlflow/mlflow/models/__init__.py", line 117, in log
    raise e
  File "/Users/sid.murching/code/mlflow/mlflow/models/__init__.py", line 105, in log
    mlflow.tracking.fluent._record_logged_model(mlflow_model)
  File "/Users/sid.murching/code/mlflow/mlflow/tracking/fluent.py", line 328, in _record_logged_model
    MlflowClient()._record_logged_model(run_id, mlflow_model)
  File "/Users/sid.murching/code/mlflow/mlflow/tracking/client.py", line 269, in _record_logged_model
    self._tracking_client._record_logged_model(run_id, mlflow_model)
  File "/Users/sid.murching/code/mlflow/mlflow/tracking/_tracking_service/client.py", line 238, in _record_logged_model
    self.store.record_logged_model(run_id, mlflow_model)
  File "/Users/sid.murching/code/mlflow/mlflow/store/tracking/rest_store.py", line 246, in record_logged_model
    self._call_endpoint(LogModel, req_body)
  File "/Users/sid.murching/code/mlflow/mlflow/store/tracking/rest_store.py", line 32, in _call_endpoint
  File "/Users/sid.murching/code/mlflow/mlflow/utils/rest_utils.py", line 137, in call_endpoint
  File "/Users/sid.murching/code/mlflow/mlflow/utils/rest_utils.py", line 107, in verify_rest_response
    raise MlflowException("%s. Response body: '%s'" % (base_msg, response.text))
mlflow.exceptions.MlflowException: API request to endpoint /api/2.0/mlflow/runs/log-model failed with error code 404 != 200. Response body: '<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.</p>
'
```

smurching (commented):
Oops, regarding the tests - I missed that we pulled shared tests into the new `AbstractStoreTest` class, that's cool :)

Just had two more comments, LGTM once they're addressed!

mlflow/tracking/fluent.py
tomasatdatabricks: It's used in Model.log(...)

mlflow/server/handlers.py
tomasatdatabricks: Makes sense.

mlflow/models/__init__.py
tomasatdatabricks: Oh, that's unfortunate, we'll have swallow up all exceptions then. Thanks for checking!

mlflow/tracking/fluent.py
smurching: Ah shoot searched for `mlflow._record_logged_model` instead of `mlflow.tracking.fluent._record_logged_model`. Makes sense, thanks!

smurching (approved):
LGTM thanks for the updates!

PR #2370, opened 2020-01-31
Update test requirement

AveshCSingh:

Adds a `tf2onnx` to our test requirements. It seems this was previously added implicitly as a dependency of another package we require.

## How is this patch tested?

Manually verified that `pytest tests/onnx/test_onnx_model_export.py::test_model_save_load_multiple_inputs` succeeds.


smurching (approved):
LGTM

PR #2374, opened 2020-01-31
[WIP] Selective test execution with testmon

smurching:

Prototype of a number of different improvements to MLflow's CI build:
* Selective test execution using pytest-testmon (only run tests corresponding to code that's changed)
* Cache pip/conda dependencies

Testmon seems in general to work, and it seems we could go with a model of running testmon on PR builds & full tests on master builds (so that if testmon misses anything we catch it shortly after merging a PR). There are a couple of tricky issues/things to investigate further:
* If you leak environment variables (or other global state) that other tests depend on, you can influence the outcome of successive tests (e.g. break them) but testmon won't necessarily execute them. However, such breakages would be caught on merges to master (when we'd run the full tests), and hopefully would be rare.
* Testmon doesn't pick up on changes in Python dependencies (e.g. if we upgrade some test deps tests may not reexecute)
    * Would we need to put deps in test files themselves or something to make testmon pick this up? 
    * Or just condition usage of testmon on test dep files not changing
* Testmon requires Python 3 (can’t use for Python 2 builds)
* You won't be able to rerun flaky tests by retriggering a build once testmon is enabled (testmon will just see that there was no code change & won't reexecute the test)
* Should verify that testmon works alright for selective execution of code that runs subprocesses (e.g. mlflow.projects.run). [Testmon docs](https://testmon.org/determining-affected-tests.html) suggest that it'd do alright - in particular, they state that testmon has the same limitations as coverage.py, which [seems capable of measuring subprocesses](https://coverage.readthedocs.io/en/coverage-5.0.3/subprocess.html#measuring-sub-processes)

## How is this patch tested?

Existing tests

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


aarondav:
Can't wait!

aarondav:
Closing for now to reduce backlog since it's not mergable.

Issue #2284, opened 2020-01-09
[FR] Update mlflow to use new Azure APIs


csteegz:
### Describe the proposal
MLFlow is currently using out-of-date APIs for deploying models to Azure. These introduce added complexity and an extra step of building an image when not required. We should update the Azure implementation to leverage up-to-date Model.package() and Model.deploy() APIs.

### Motivation
Keep MLFlow's Azure implementation up to date and simplify model deployment.
 
### Proposed Changes
Modify existing [Azure ML Module](https://github.com/mlflow/mlflow/blob/master/mlflow/azureml/__init__.py#L30) to use new Model.deploy() API as opposed to webservice.deploy_from_image.

Introduce a new `deploy` API similar to the [SageMaker Module](https://github.com/mlflow/mlflow/blob/master/mlflow/sagemaker/__init__.py#L138). This will use the new [Azure ML API](https://docs.microsoft.com/python/api/azureml-core/azureml.core.model(class)?view=azure-ml-py#deploy-workspace--name--models--inference-config-none--deployment-config-none--deployment-target-none--overwrite-false-) to deploy the model.

akshaya-a:
@pogil we're happy to pick up the work here - just wanted to get some clarity on what exactly it'd take to remove the experimental tag from the deployment commands

dbczumar:
Now that #2375 is merged, I'm going to go ahead and close this! Thanks @csteegz !

PR #2375, opened 2020-01-31
Add new API for Azure Deployment

csteegz:

Implements #2284 by adding new API to mlflow.

## How is this patch tested?

I have manually tested deployment to Azure Machine Learning. 

Introduce new APIs for deploying to Azure Machine Learning.

- [ ] UI
- [ ] CLI
- [X] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [X] Serving
- [ ] R
- [ ] Java
- [ ] Python


akshaya-a:
@pogil  @tomasatdatabricks  @dbczumar

akshaya-a:
Please look at the tests here: https://github.com/mlflow/mlflow/blob/master/tests/azureml/test_image_creation.py

Will want to add something similar for the new API

csteegz:
Oh wow - we have better tests here then in the Azure ML SDK. I'll get working on them.

mlflow/azureml/__init__.py
```diff
@@ -297,7 +455,7 @@ def _load_pyfunc_conf(model_path):
                 message=("The specified model does not contain the `python_function` flavor. This "
                          " flavor is required for model deployment required for model deployment."),
                 error_code=INVALID_PARAMETER_VALUE)
-    return model.flavors[pyfunc.FLAVOR_NAME]
+    return model.flavors[pyfunc.FLAVOR_NAME], model.run_id
```
dbczumar: This change breaks the existing `build_image` implementation. It's also possible that the `run_id` property doesn't exist for the given `Model` object (see https://github.com/mlflow/mlflow/blob/3eaaa046ce3501a48f390b4d778e34e4168aadac/mlflow/models/__init__.py#L41-L43), in which case this method would throw

mlflow/azureml/__init__.py
```diff
@@ -206,8 +207,163 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+    # still be accessible for import from Python 2. Therefore, we will only import from the SDK
+    # upon method invocation.
+    # pylint: disable=import-error
+    from azureml.core.model import Model as AzureModel, InferenceConfig
+    from azureml.core import Environment as AzureEnvironment
+    from azureml.core import VERSION as AZUREML_VERSION
+
+    absolute_model_path = _download_artifact_from_uri(model_uri)
+
+    model_pyfunc_conf, run_id = _load_pyfunc_conf(model_path=absolute_model_path)
```
dbczumar: As discussed in `_load_pyfunc_conf`, a run ID is not always guaranteed to exist in the MLmodel configuration. For example, a user may choose to invoke `mlflow.pyfunc.save_model` to produce a model independently of an MLflow run. This seems to impact the `model_name` and `service_name` generated below. Are there other naming schemes that you could adopt without relying on run ID?

dbczumar: We should also make sure to test the case where models without run IDs are deployed via this method.

mlflow/azureml/__init__.py
```diff
@@ -206,8 +207,163 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+            raise MlflowException("Cannot deploy pyfunc model without an environment to Azure ML.")
+
+        if mlflow_home is not None:
+            wheel = _create_mlflow_wheel(mlflow_home, tmp.path("dist"))
+            whl_url = AzureEnvironment.add_private_pip_wheel(workspace=workspace, file_path = wheel, exist_ok=True)
+            environment.python.conda_dependencies.add_pip_package(whl_url)
+        else:
+            environment.python.conda_dependencies.add_pip_package("mlflow~={}".format(mlflow_version))
+
+        environment.python.conda_dependencies.add_pip_package("azureml-defaults~={}".format(AZUREML_VERSION))
```
dbczumar: Can we add a comment explaining why installing `azureml-defaults` is necessary? The description from https://pypi.org/project/azureml-defaults/ is pretty vague:

```
The azureml-defaults is a metapackage that is used internally by Azure Machine Learning.
```

mlflow/azureml/__init__.py
```diff
@@ -311,6 +469,32 @@ def _get_mlflow_azure_resource_name():
...
+    Create the wheel of MLFlow by using setup.py bdist_wheel in the outdir.
+
+    :param model_path: The absolute path to the mlflow.
+    :param out_dir: The absolute path to the outdir.
+    :return: The absolute path to the wheel.
+    """
+    out_path = Path(out_dir).resolve()
+    process = subprocess.run([sys.executable, "setup.py", "bdist_wheel", "-d", out_path], cwd=mlflow_dir, check=True)
+    files = list(out_path.glob("./*.whl"))
+    if len(files) != 1:
```
dbczumar: Nit: What happens if `len(files) > 1`? The error message doesn't seem to apply in this case (we found the wheel, but there were multiple such wheels...)

mlflow/azureml/__init__.py
```diff
@@ -206,8 +207,163 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
             image.wait_for_creation(show_output=True)
         return image, registered_model
 
+@experimental
+def deploy(model_uri, workspace, deployment_config = None, service_name=None, model_name=None,
+                mlflow_home=None, synchronous=True):
+    """
+    Register an MLflow model with Azure ML and build an Azure ML ContainerImage for deployment.
```
dbczumar: @csteegz This documentation needs to be updated; it is currently identical to `build_image`.

mlflow/azureml/__init__.py
```diff
@@ -311,6 +469,32 @@ def _get_mlflow_azure_resource_name():
...
+
+def _create_mlflow_wheel(mlflow_dir, out_dir):
+    """
+    Create the wheel of MLFlow by using setup.py bdist_wheel in the outdir.
+
+    :param model_path: The absolute path to the mlflow.
+    :param out_dir: The absolute path to the outdir.
+    :return: The absolute path to the wheel.
+    """
+    out_path = Path(out_dir).resolve()
```
dbczumar: I'm pretty sure that `out_dir` has to exist before `Path(out_dir).resolve()` will work (at least on Python 2); we're currently passing `out_dir = TempDir().path("dist")`, but the `dist` subdir does not yet exist.

The following small example fails in Python 2:
```
 from mlflow.utils.file_utils import TempDir
 with TempDir() as t:
     p = t.path("dist")
     from pathlib import Path
     Path(p).resolve()

OSError: [Errno 2] No such file or directory: '/private/var/folders/rq/g5vkbkv93r12yb1t374lfj0r0000gn/T/tmpNHXXk7/dist'
```

mlflow/azureml/__init__.py
```diff
@@ -206,8 +207,163 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+        _create_execution_script(output_path=execution_script_path, azure_model=registered_model)
+
+        environment = None
+        if pyfunc.ENV in model_pyfunc_conf:
+            environment = AzureEnvironment.from_conda_specification(_get_mlflow_azure_name_from_run(run_id), os.path.join(tmp_model_path, model_pyfunc_conf[pyfunc.ENV]))
+        else:
+            raise MlflowException("Cannot deploy pyfunc model without an environment to Azure ML.")
+
+        if mlflow_home is not None:
+            wheel = _create_mlflow_wheel(mlflow_home, tmp.path("dist"))
```
dbczumar: Nit: Can we emit a log message indicating that we are about to build an MLflow wheel before beginning the build process?

mlflow/azureml/__init__.py
```diff
@@ -206,8 +207,163 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+        # image creation, so we create the execution script as a temporary file in the current
+        # working directory.
+        execution_script_path = tmp.path("execution_script.py")
+        _create_execution_script(output_path=execution_script_path, azure_model=registered_model)
+
+        environment = None
+        if pyfunc.ENV in model_pyfunc_conf:
+            environment = AzureEnvironment.from_conda_specification(_get_mlflow_azure_name_from_run(run_id), os.path.join(tmp_model_path, model_pyfunc_conf[pyfunc.ENV]))
+        else:
+            raise MlflowException("Cannot deploy pyfunc model without an environment to Azure ML.")
```
dbczumar: @csteegz Why does this restriction exist? We don't currently perform validation for the presence of a conda environment in the `build_image` routine; in fact, the `conda_file` argument of `ContainerImage.image_configuration()` (which is called/constructed during `mlflow.azureml.build_image`) has a default value of `None`, which seems to imply that it's optional: https://docs.microsoft.com/en-us/python/api/azureml-core/azureml.core.image.containerimage?view=azure-ml-py#image-configuration-execution-script--runtime--conda-file-none--docker-file-none--schema-file-none--dependencies-none--enable-gpu-none--tags-none--properties-none--description-none--base-image-none--base-image-registry-none--cuda-version-none-

Can you provide more context here? Did we miss something in `build_image`, or is there a requirements difference between the Azure resources created by `build_image` and `deploy` regarding the `conda.yaml` file?

dbczumar (changes requested):
@csteegz Thanks for submitting this PR to keep our AzureML deployment functionality up-to-date with best practices! I left a few comments/questions.

The other significant outstanding item is test coverage; let me know if you run into any trouble writing / executing tests for this functionality!

mlflow/azureml/__init__.py
jpe316: the containerimage class is deprecated in favor of environments

mlflow/azureml/__init__.py
jpe316: It installs the flask serve stack into the image - poorly named, we're working on it

mlflow/azureml/__init__.py
akshaya-a: I think there is better defaulting we can do here that @dbczumar's question leads to - it's just a rejiggering of concepts but ContainerImage  implied a certain default Environment that we should add here as well

mlflow/azureml/__init__.py
csteegz: I'm going with a Guid if run ID isn't present but perhaps a timestamp would be more useful? not really sure.

mlflow/azureml/__init__.py
csteegz: Added making the directory.

semirelateldly, the  whole package doesn't support python2 per 289, although I guess I'm not entirely clear what that message means.

mlflow/azureml/__init__.py
akshaya-a: @csteegz  is this thread actually resolved? can we do the equivalent of python/python-slim like the old APIs used to do here while perhaps emitting a warning?

csteegz: Uh - What I switched to do was use a default environment with azureml-defaults and MLFlow added, which should be the equivalent of python-slim. Not sure why you weren't able to see that change (https://github.com/mlflow/mlflow/pull/2375/commits/af7d320f22296b317040577e190a4388b1209a83#diff-6e19c0d7aba3a117d4d11e9b592639a5R346)

dbczumar:
@csteegz Sounds good! I'll take another pass within the next 24 hours!

mlflow/azureml/__init__.py
```diff
@@ -312,18 +503,65 @@ def _get_mlflow_azure_resource_name():
...
+    return files[0]
+
+
+def parse_deploy_config(json_dict_input):
+    from azureml.core.webservice import AciWebservice, AksWebservice
+    computes = {"aks": AksWebservice, "aci": AciWebservice}
+    try:
+        compute_target_type = json_dict_input["computeType"].lower()
+        del json_dict_input["computeType"]
+        return computes[compute_target_type].deploy_configuration(**compute_target_type)
```
trangevi, line 546: Sorry, I don't understand what the `**compute_target_type` doing here?

mlflow/azureml/cli.py
```diff
@@ -68,3 +68,48 @@ def build_image(model_uri, workspace_name, subscription_id, image_name, model_na
     mlflow.azureml.build_image(
             model_uri=model_uri, workspace=workspace, image_name=image_name, model_name=model_name,
             mlflow_home=mlflow_home, description=description, tags=tags, synchronous=True)
+
+
+@commands.command("deploy")
+@cli_args.MODEL_URI
+@click.option("--workspace-name", "-w", required=True,
+              help="The name of the Azure Workspace in which to build the image.")
```
trangevi, line 76: in which to deploy the webservice?

mlflow/azureml/cli.py
```diff
@@ -68,3 +68,48 @@ def build_image(model_uri, workspace_name, subscription_id, image_name, model_na
...
+@click.option("--subscription-id", "-s", default=None,
+              help=("The subscription id associated with the Azure Workspace in which to build"
+                    " the service"))
+@click.option("--service-name", "-s", default=None,
+              help=("The name to assign the Azure Container Image that is created. If unspecified,"
+                    " a unique service name will be generated."))
```
trangevi, line 82: May be good to provide information on how it will be generated?

mlflow/azureml/cli.py
```diff
@@ -68,3 +68,48 @@ def build_image(model_uri, workspace_name, subscription_id, image_name, model_na
...
+@click.option("--model-name", "-n", default=None,
+              help=("The name to assign the Azure Model that is created. If unspecified,"
+                    " a unique image name will be generated."))
```
trangevi, line 85: Model name instead of image name. Same comment about how it will be generated.

mlflow/azureml/cli.py
```diff
@@ -68,3 +68,48 @@ def build_image(model_uri, workspace_name, subscription_id, image_name, model_na
...
+@cli_args.MLFLOW_HOME
+@click.option("--deployment-config", "-d", default=None,
+              help=("A path to a JSON formatted deployment config to for the"
```
trangevi, line 88: looks like a type with the extra "to"

mlflow/azureml/cli.py
```diff
@@ -68,3 +68,48 @@ def build_image(model_uri, workspace_name, subscription_id, image_name, model_na
...
+                    " Azure ML Webservice that is created."))
+@experimental
+def deploy(model_uri, workspace_name, subscription_id, service_name, model_name,
+           mlflow_home, deployment_config):
+    """
```
trangevi, line 93: Looks like this docstring needs to be updated

mlflow/azureml/cli.py
```diff
@@ -68,3 +68,48 @@ def build_image(model_uri, workspace_name, subscription_id, image_name, model_na
...
+    documentation: https://www.mlflow.org/docs/latest/models.html#azureml-deployment.
+    """
+    # The Azure ML SDK is only compatible with Python 3. However, this CLI should still be
+    # accessible for inspection rom Python 2. Therefore, we will only import from the SDK
+    # upon command invocation.
+    # pylint: disable=import-error
+    from azureml.core import Workspace
+
+    workspace = Workspace.get(name=workspace_name, subscription_id=subscription_id)
+    if deployment_config is not None:
```
trangevi, line 109: I would make this just `if deployment config:`. Considering it's CLI handling it would be easy to end up with an empty string, which will cause this to fail

mlflow/azureml/cli.py
```diff
@@ -68,3 +68,48 @@ def build_image(model_uri, workspace_name, subscription_id, image_name, model_na
...
+        deployment_dict = json.loads(deployment_config)
```
trangevi, line 110: Based on the help message, deployment_config is a path to a file, shouldn't this be `json.load(deployment_config)`?

travis/large-requirements.txt
```diff
@@ -28,7 +28,7 @@ typing==3.6.6
 pysftp==0.2.9
 keras==2.3.1
 attrdict==2.0.0
-azureml-sdk==1.0.17; python_version >= "3.0"
+azureml-sdk==1.2.0; python_version >= "3.0"
```
trangevi, line 31: Would there be value in making this some kind of ~= dependency?

trangevi (changes requested):

mlflow/azureml/__init__.py
```diff
@@ -207,7 +210,181 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+                message=("Azure ML can only deploy models trained in Python 3 and above. See"
+                         " the following MLflow GitHub issue for a thorough explanation of this"
+                         " limitation and a workaround to enable support for deploying models"
+                         " trained in Python 2: https://github.com/mlflow/mlflow/issues/668"),
+                error_code=INVALID_PARAMETER_VALUE)
+
+    tags = _build_tags(model_uri=model_uri, model_python_version=model_python_version)
+
+    if service_name is None:
+        service_name = _get_mlflow_azure_name_from_run(run_id)
```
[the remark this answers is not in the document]
csteegz: I'm not gonna change this - I think that calling the function each time is clearer and adds very little overhead.

mlflow/azureml/__init__.py
```diff
@@ -290,14 +469,26 @@ def _load_pyfunc_conf(model_path):
...
+    return name
+
+
+def _load_pyfunc_conf_with_model(model_path):
+    """
+    Loads the `python_function` flavor configuration for the specified model or throws an exception
+    if the model does not contain the `python_function` flavor.
+
+    :param model_path: The absolute path to the model.
+    :return: The model's `python_function` flavor configuration and the model.
```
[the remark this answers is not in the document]
csteegz: that doesn't appear to be convention for this project - not gonna do it.

mlflow/azureml/__init__.py
csteegz: ah, that's a mistake, should be **json_dict_input

travis/large-requirements.txt
csteegz: No. AzureML doesn't follow semver (because we're bad), so we need to pin to an exact version.

mlflow/azureml/__init__.py
```diff
@@ -207,7 +210,181 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+    if service_name is None:
+        service_name = _get_mlflow_azure_name_from_run(run_id)
+    if model_name is None:
+        model_name = _get_mlflow_azure_name_from_run(run_id)
+
+    with TempDir(chdr=True) as tmp:
+        model_directory_path = tmp.path("model")
+        tmp_model_path = os.path.join(
+            model_directory_path,
+            _copy_file_or_tree(src=absolute_model_path, dst=model_directory_path))
```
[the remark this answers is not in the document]
csteegz: for whatever reason this doesn't play nice with relative paths. When I tried it, multiple tests started to fail.

mlflow/azureml/__init__.py
```diff
@@ -207,7 +210,181 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+    if model_python_version is not None and\
+            StrictVersion(model_python_version) < StrictVersion("3.0.0"):
+        raise MlflowException(
+                message=("Azure ML can only deploy models trained in Python 3 and above. See"
+                         " the following MLflow GitHub issue for a thorough explanation of this"
+                         " limitation and a workaround to enable support for deploying models"
+                         " trained in Python 2: https://github.com/mlflow/mlflow/issues/668"),
+                error_code=INVALID_PARAMETER_VALUE)
+
+    tags = _build_tags(model_uri=model_uri, model_python_version=model_python_version)
```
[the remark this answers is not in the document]
csteegz: Tags for the service are specified in the deployment config now.

trangevi (approved):

trangevi:
From an AML perspective I think this looks good.

dbczumar:
@trangevi @csteegz I will have this reviewed from the MLflow side by EOD tomorrow! Sorry for the delay.

mlflow/azureml/__init__.py
```diff
@@ -207,7 +210,180 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+    :param workspace: The AzureML workspace in which to deploy the service. This is a
+                      `azureml.core.Workspace` object.
+    :param deployment_config: The configuration for the Azure web service. This configuration
+                              allows you to specify the resources the webservice will use and
+                              the compute cluster it will be deployed in. If unspecified, the web
+                              service will be deployed into a Azure Container Instance. This is a
+                              `azureml.core.DeploymentConfig` object. For more information, see
+                              `<https://docs.microsoft.com/python/api/azureml-core/
+                              azureml.core.webservice.aks.aksservicedeploymentconfiguration>` and
+                              `<https://docs.microsoft.com/en-us/python/api/azureml-core/azureml.core.webservice.aci.aciservicedeploymentconfiguration>
```
dbczumar, line 251: These documentation links don't seem to be rendering properly:

Screen Shot 2020-04-10 at 5 55 24 PM

Here's a fix:
```suggestion
                              `<https://docs.microsoft.com/python/api/azureml-core/
                              azureml.core.webservice.aks.aksservicedeploymentconfiguration>`_ and
                              `<https://docs.microsoft.com/en-us/python/api/azureml-core/azureml
                              .core.webservice.aci.aciservicedeploymentconfiguration>`_
```

mlflow/azureml/__init__.py
```diff
@@ -207,7 +210,180 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+            _logger.info("Bulding temporary MLFlow wheel in %s", path)
+            wheel = _create_mlflow_wheel(mlflow_home, path)
+            whl_url = AzureEnvironment.add_private_pip_wheel(
+                                       workspace=workspace,
+                                       file_path=wheel,
+                                       exist_ok=True)
+            environment.python.conda_dependencies.add_pip_package(whl_url)
+        else:
+            environment.python.conda_dependencies.add_pip_package(
+                "mlflow~={}".format(mlflow_version))
```
dbczumar, line 362: @csteegz I think we should depend on exactly the user's current version of MLflow for deployment, as there may be properties of a particular MLflow release that a user's model code depends on (especially if we break APIs in 2.0). Can we change this to `==` instead of `~=`?

dbczumar (changes requested):
@csteegz I left a few more comments - I think this is very close! Thanks for all the iteration.

We should also update the MLflow Models documentation for Azure ML to reflect the new `deploy()` command and provide an updated example workflow, since this is the preferred approach. Is this something you'd be willing to do, either as part of this PR or as a follow-up? Here's the current docs page for reference: file:///Users/czumar/mlflow/docs/build/html/models.html#deploy-a-python-function-model-on-microsoft-azure-ml.

mlflow/azureml/__init__.py
```diff
@@ -207,7 +210,180 @@ def build_image(model_uri, workspace, image_name=None, model_name=None,
...
+    >>> workspace_name = "<Name of your Azure ML workspace>"
+    >>> subscription_id = "<Your Azure subscription ID>"
+    >>> resource_group = "<Name of the Azure resource group in which to create Azure ML resources>"
+    >>> location = "<Name of the Azure location (region) in which to create Azure ML resources>"
+    >>> azure_workspace = Workspace.create(name=workspace_name,
+    >>>                                    subscription_id=subscription_id,
+    >>>                                    resource_group=resource_group,
+    >>>                                    location=location,
+    >>>                                    create_resource_group=True,
+    >>>                                    exist_okay=True)
```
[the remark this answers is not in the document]
csteegz: @akshaya-a  was this a change or a mistake before?

dbczumar:
@csteegz It would be awesome to get this merged in advance of the next MLflow release (we're intending to cut the release branch tomorrow evening). Is it possible to address remaining comments by then?

mlflow/azureml/__init__.py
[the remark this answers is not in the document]
csteegz: Code looks to be correct, have updated documentation.

akshaya-a:
@dbczumar  to your previous question, we'd definitely be willing to update the doc you mentioned - I suggest in a followup to make it nonblocking to your planned release

csteegz:
@akshaya-a / @dbczumar it looks like the way of documenting everything changed recently? I'll need to take another pass tomorrow - what time would you need to merge it by?

dbczumar:
@csteegz Sorry for the inconvenience - we may some minor tweaks to the way that example code snippets are rendered in the MLflow docs (i.e., we removed instances of `>>>` and are now using RST blocks instead).

Regarding timelines, we're hoping to cut the branch by the end of the day (~7pm PST).

csteegz:
I'm not feeling well so likely not finishing this up today.

akshaya-a:
@csteegz what's pending at this point? I can take a crack at it

dbczumar:
@akshaya-a I think we're just missing a `tags` parameter for the new deployment API, as well as a unit test for tags (https://github.com/mlflow/mlflow/pull/2375#discussion_r406989500, https://github.com/mlflow/mlflow/pull/2375#discussion_r406990861). Can you give this a shot? If not, I'll try to add these later today.

csteegz:
Also I haven't tested since doing the PR revisions, so another test
deployment before merging would be good.

mlflow/azureml/__init__.py
```diff
@@ -290,14 +468,26 @@ def _load_pyfunc_conf(model_path):
     :param model_path: The absolute path to the model.
     :return: The model's `python_function` flavor configuration.
     """
+    (name, _) = _load_pyfunc_conf_with_model(model_path)
```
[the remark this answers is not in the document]
akshaya-a: addressed

travis/large-requirements.txt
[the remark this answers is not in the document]
akshaya-a: Addressed, though I dropped the python version qualifier as MLflow dropped py2 as well, and i slimmed the dep to azureml-core instead of azureml-sdk which is a metapackage for more things than we need here

dbczumar (approved):
@csteegz @akshaya-a I'm going to LGTM this PR and get it merged. I'll follow up by reviewing / merging https://github.com/mlflow/mlflow/pull/2711 (unfortunately, I can't merge https://github.com/mlflow/mlflow/pull/2711 directly into this PR branch due to a lack of permissions). I've done some manual testing of deployment on https://github.com/mlflow/mlflow/pull/2711; it appears to be working properly.

PR #2376, opened 2020-01-31
FileStore: Fix parameter reading logic to support param values with newlines

### LINKED ISSUE #2235: log_param failes logging string containing \n
### System information
mlflow, version 1.4.0
Python 3.7.4

### Describe the problem
mlflow throws and exception: "Exception: Unexpected data for param <key>. Param recorded more than once" when trying to log a multiline string with log_param. 

Ok, this is a special case and I don't see many use-cases, but I stumbled this today. That said, I don't think the Exception description is not very clear (took me a while do understand what actually was wrong - I only tried to log the param once...).

### Code to reproduce issue
Provide a reproducible test case that is the bare minimum necessary to generate the problem.
```
import mlflow

# create a multiline string
a_two_line_str = "a\nb"

with mlflow.start_run() as mlrun:

    # this works
    clean_str = a_two_line_str.replace("\n|\r", "")
    mlflow.log_param("two_liner", clean_str)
    print("no exception yet")

    # this fails
    print(a_two_line_str)
    mlflow.log_param("two_liner", a_two_line_str)
    # >> Exception: Unexpected data for param 'two_liner'. Param recorded more than once

```

### Other info / logs
The reason for the exception is that `len(param_data)>1`, since `len("a\nb")>1`.

Propose either 1) a fix to strip for \n and \r, or 2) give a more describing Exception string, e.g:

#### 1) Strip \n and \r from param data strings
mlflow.tracking._tracking_service.client.TrackingServiceClient.log_param:
```
def log_param(self, run_id, key, value):
    """
    Log a parameter against the run ID. Value is converted to a string.
    """
    _validate_param_name(key)
    param = Param(key, str(value).replace("\n|\r", "")) #change: the '.replace(..)
    self.store.log_param(run_id, param)
```

#### 2) Change Exception description
mlflow.store.tracking.file_store.FileStore._get_param_from_file:
```
@staticmethod
def _get_param_from_file(parent_path, param_name):
    _validate_param_name(param_name)
    param_data = read_file_lines(parent_path, param_name)
    if len(param_data) > 1:
        raise Exception("Unexpected data for param '%s'. Len of param data is bigger than 1"
                        % param_name)
    # The only cause for param_data's length to be zero is the param's
    # value is an empty string
    value = '' if len(param_data) == 0 else str(param_data[0].strip())
    return Param(param_name, value)
```

dbczumar:

The FileStore currently reads parameter files line-by-line and throws an exception if multiple lines are encountered for the same parameter. As a result, parameters whose values contain newlines are not supported. However, the FileStore does not *write* parameter values line-by-line, nor does it support storing multiple parameter values in the same file. Accordingly, there's no obvious reason to read the file line-by-line or include this validation.

This PR corrects & simplifies the FileStore logic used for reading parameter values, introducing support for parameter values containing newlines. It fixes #2235.

## How is this patch tested?

Unit tests

## Release Notes

* Fixes a bug in the FileStore that prevented users from reading parameters whose values contain newlines (`\n`).

* The FileStore now correctly enforces the immutability of parameter values.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/mlflow/store/tracking/file_store.py
+++ b/mlflow/store/tracking/file_store.py
@@ -637,8 +637,29 @@ class FileStore(AbstractStore):
 
     def _log_run_param(self, run_info, param):
         param_path = self._get_param_path(run_info.experiment_id, run_info.run_id, param.key)
+        writeable_param_value = self._writeable_value(param.value)
+        if os.path.exists(param_path):
+            self._validate_new_param_value(
+                param_path=param_path, param_key=param.key,
+                run_id=run_info.run_id, new_value=writeable_param_value)
         make_containing_dirs(param_path)
-        write_to(param_path, self._writeable_value(param.value))
+        write_to(param_path, writeable_param_value)
+
+    def _validate_new_param_value(self, param_path, param_key, run_id, new_value):
+        """
+        When logging a parameter with a key that already exists, this function is used to
+        enforce immutability by verifying that the specified parameter value matches the existing
+        value.
+        :raises: py:class:`mlflow.exceptions.MlflowException` if the specified new parameter value
+                 does not match the existing parameter value.
+        """
+        with open(param_path, "r") as param_file:
+            current_value = param_file.read()
+        if current_value != new_value:
+            raise MlflowException(
+                "Changing param value is not allowed. Param with key='{}' was already"
+                " logged with value='{}' for run ID='{}'. Attempted logging new value"
+                " '{}'.".format(param_key, current_value, run_id, new_value))
 
     def set_experiment_tag(self, experiment_id, tag):
         """
--- a/mlflow/store/tracking/file_store.py
+++ b/mlflow/store/tracking/file_store.py
@@ -657,9 +657,10 @@ class FileStore(AbstractStore):
             current_value = param_file.read()
         if current_value != new_value:
             raise MlflowException(
-                "Changing param value is not allowed. Param with key='{}' was already"
+                "Changing param values is not allowed. Param with key='{}' was already"
                 " logged with value='{}' for run ID='{}'. Attempted logging new value"
-                " '{}'.".format(param_key, current_value, run_id, new_value))
+                " '{}'.".format(param_key, current_value, run_id, new_value),
+                databricks_pb2.INVALID_PARAMETER_VALUE)
 
     def set_experiment_tag(self, experiment_id, tag):
         """
--- a/mlflow/store/tracking/file_store.py
+++ b/mlflow/store/tracking/file_store.py
@@ -515,13 +515,7 @@ class FileStore(AbstractStore):
     @staticmethod
     def _get_param_from_file(parent_path, param_name):
         _validate_param_name(param_name)
-        param_data = read_file_lines(parent_path, param_name)
-        if len(param_data) > 1:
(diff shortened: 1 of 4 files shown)


mlflow/store/tracking/file_store.py
```diff
@@ -515,13 +515,7 @@ def _get_metric_history(self, run_info, metric_key):
...
     def _get_param_from_file(parent_path, param_name):
         _validate_param_name(param_name)
-        param_data = read_file_lines(parent_path, param_name)
-        if len(param_data) > 1:
-            raise Exception("Unexpected data for param '%s'. Param recorded more than once"
-                            % param_name)
-        # The only cause for param_data's length to be zero is the param's
-        # value is an empty string
-        value = '' if len(param_data) == 0 else str(param_data[0].strip())
+        value = read_file(parent_path, param_name)
```
AveshCSingh: Quick sanity-check: What's the behavior when the FileStore writes the same parameter twice with different values? 

eg
```
with mlflow.start_run() as mlrun:
    mlflow.log_param("foo", "bar")
    mlflow.log_param("foo", "baz")
```

will the resulting file for "foo" be appended to or overwritten? The former would be problematic since read_file will return both values.

It might be good to add a unit test for this scenario since we no longer raise an obvious error ("Param recorded more than once") in that case.

AveshCSingh (commented):
Nice fix! I have one question inline. Otherwise, this looks good.

mlflow/store/tracking/file_store.py
juntai-zheng: I believe it should be the latter case since the function used (correct me if I'm wrong) to write the param truncates the param value when writing:
```
def write_to(filename, data):
    with codecs.open(filename, mode="w", encoding=ENCODING) as handle:
        handle.write(data)
```

I think writing a test for this case would be great though!

dbczumar: @AveshCSingh @juntai-zheng Great catches! Parameter values are meant to be immutable. Accordingly, I've introduced validation logic to ensure that multiple writes for the same parameter key with different values are rejected.

tests/store/tracking/test_file_store.py
```diff
@@ -595,14 +597,36 @@ def test_weird_param_names(self):
...
+
+    def test_log_param_enforces_value_immutability(self):
+        param_name = "new param"
+        fs = FileStore(self.test_root)
+        run_id = self.exp_data[FileStore.DEFAULT_EXPERIMENT_ID]["runs"][0]
+        fs.log_param(run_id, Param(param_name, "value1"))
+        # Duplicate calls to `log_param` with the same key and value should succeed
+        fs.log_param(run_id, Param(param_name, "value1"))
+        with pytest.raises(MlflowException) as exc:
+            fs.log_param(run_id, Param(param_name, "value2"))
```
AveshCSingh: nice!

AveshCSingh (approved):
Awesome! LGTM.

juntai-zheng (approved):
LGTM

PR #2378, opened 2020-02-01
Update XGBoost and LightGBM examples

harupy:

Update the XGboost and LightGBM examples to use automatic logging.

## How is this patch tested?

`pytest tests/examples/test_examples.py`

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


examples/lightgbm/train.py
```diff
@@ -29,31 +29,32 @@ def main():
...
-
-    # log parameters, metrics, and model
-    with mlflow.start_run() as run:
-        mlflow.log_params(params)
+    train_set = lgb.Dataset(X_train, label=y_train)
+
+    # enable auto logging
+    mlflow.lightgbm.autolog()
+
+    with mlflow.start_run():
```
smurching: If we use autologging, we shouldn't need a `mlflow.start_run` anymore, right?

examples/xgboost/train.py
```diff
@@ -32,29 +32,30 @@ def main():
...
-    loss = log_loss(y_test, y_proba)
-    acc = accuracy_score(y_test, y_pred)
-
-    # log parameters, metrics, and model
-    with mlflow.start_run() as run:
-        mlflow.log_params(params)
+    # enable auto logging
+    mlflow.xgboost.autolog()
+
+    with mlflow.start_run():
```
smurching: Same here, wondering if we can remove the `mlflow.start_run`

smurching (commented):
Just small comments, thanks @harupy!

examples/lightgbm/train.py
harupy: If we didn't use `mlflow.start_run` here, `log_metrics` after `lgb.train` would create a new run and record metrics in it.

harupy:
`test_examples.py` in python 2.7 fails due to the following error:

https://travis-ci.org/mlflow/mlflow/jobs/645884094#L2468

```
_tkinter.TclError: no display name and no $DISPLAY environment variable
```

We can avoid this error by setting Matplotlib to use non-interactive backend with:
```python
import matplotlib
matplotlib.use('Agg')
```

However, I'm wondering if it's ok to add this code to the example scripts because users might think this code is required to use auto logging.

examples/lightgbm/train.py
```diff
@@ -4,18 +4,21 @@
...
 
 def parse_args():
     parser = argparse.ArgumentParser(description='LightGBM example')
     parser.add_argument('--colsample-bytree', type=float, default=1.0,
                         help='subsample ratio of columns when constructing each tree (default: 1.0)')
     parser.add_argument('--subsample', type=float, default=1.0,
                         help='subsample ratio of the training instances (default: 1.0)')
-    parser.add_argument('--conda-env',  type=str, default=None,
+    parser.add_argument('--conda-env', type=str, default=None,
                         help='the path to a conda environment yaml file (default: None)')
```
smurching: Let's remove this argument now that it's no longer used? Same in the xgboost example

smurching (approved):
Sorry for the delay @harupy, had one last observation, otherwise LGTM

harupy:
@smurching 
Fixed the `fbprophet` issue and all the tests passed.

harupy:
@smurching Thank you always!

PR #2386, opened 2020-02-04
Close figure after auto logging

harupy:

The current implementation of `autolog` for the XGBoost and LightGBM flavors creates new figures when `xxx.train` is called. If a user calls it multiple times, the following warning is raised. This PR adds `plt.close(fig)`  to close logged figures and remove the warning.

```
RuntimeWarning: More than 20 figures have been opened.
Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are
retained until explicitly closed and may consume too much memory.
(To control this warning, see the rcParam `figure.max_open_warning`).
```

Code to reproduce the warning:

```python
import numpy as np
import lightgbm as lgb

import mlflow.lightgbm

X = np.random.random((10, 10))
y = np.random.randint(2, size=10)
train_set = lgb.Dataset(X, y)

mlflow.lightgbm.autolog()

for _ in range(11):
    lgb.train({}, train_set, num_boost_round=1, verbose_eval=0)

```

## How is this patch tested?

NA

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/travis/run-large-python-tests.sh
+++ b/travis/run-large-python-tests.sh
@@ -15,6 +15,9 @@ else
   cat $SAGEMAKER_OUT;
 fi
 
+# Set matplotlib to not use the Xwindows backend which causes an error.
+echo "backend: Agg" > ~/.config/matplotlib/matplotlibrc
+
 # NB: Also add --ignore'd tests to run-small-python-tests.sh
 pytest tests --large --ignore=tests/examples --ignore=tests/h2o --ignore=tests/keras \
   --ignore=tests/pytorch --ignore=tests/pyfunc --ignore=tests/sagemaker --ignore=tests/sklearn \
--- a/travis/run-large-python-tests.sh
+++ b/travis/run-large-python-tests.sh
@@ -15,10 +15,6 @@ else
   cat $SAGEMAKER_OUT;
 fi
 
-# Set matplotlib to not use the Xwindows backend which causes an error.
-# Ref.: https://stackoverflow.com/questions/37604289/tkinter-tclerror-no-display-name-and-no-display-environment-variable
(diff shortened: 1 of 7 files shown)


travis/run-large-python-tests.sh
```diff
@@ -15,6 +15,10 @@ else
   cat $SAGEMAKER_OUT;
 fi
 
+# Set matplotlib to not use the Xwindows backend which causes an error.
```
smurching: Why set this config here instead of programmatically in test_xgboost_autolog as was done previously? The only issue with setting the config here is that if someone runs this script locally while developing a PR, we'll alter their matplotlib settings

smurching (commented):
Just one question, otherwise looks good

travis/run-large-python-tests.sh
harupy: @smurching I agree. I reverted the changes.

smurching (approved):
LGTM. BTW one observation, looks like XGBoost & LightGBM autologging will currently fail entirely (as soon as `autolog` is called) if matplotlib isn't installed, because they import `matplotlib` immediately. It might be better to import `matplotlib` only in the `log_feature_importance` function where it's used directly, so that only feature-importance-plot-logging fails if `matplotlib` is uninstalled. Would be happy to review that fix in a follow-up PR

harupy:
@smurching  Thanks for the review! I agree on moving `import matplotlib.pyplot as plt` in `log_feature_importance`. I'll open a follow-up PR.

PR #2393, opened 2020-02-05
Save metric plot state in URLs for easier link sharing

smurching:

Saves full metric plot state in URLs, to enable collaborators to share links to pre-configured metric plots. This enables e.g. sharing a zoomed-in, smoothed comparison of a particular set of metrics with a collaborator, e.g. to convey information like "Run A had a faster loss reduction than Run B initially" or "Run B had a lower final loss".

Previously, to share plots with each other, users needed to take screenshots or share instructions on how to configure the plots.

The primary implementation changes in this PR (will also leave comments on this PR to identify them as they appear) are:
* Removing component state from MetricPlotPanel & instead fetching state information from the URL (many replacements of `this.state` with `this.getUrlState()`)
* Overriding plotly event handlers for single & double-clicks of individual curves in the legend. The default plotly implementation is for single-clicking a curve in the legend to toggle its display on/off, and for double-clicking a curve to display that curve & hide all other curves. We re-implement this behavior & extend it to also update the page URL
* Updates to tests

### Example GIF
In the GIF below, we configure the plot & refresh a few times to show that plot configs persist across page refreshes (i.e. they're restored from the page URL):
gif

## How is this patch tested?
* Updates existing unit tests.
* Also tested manually, by running `python tests/generate_ui_test_data.py` to generate dummy UI data & viewing metric plots containing one or more run & one or more metrics.

Metric plot config is now saved in the URL to simplify sharing plots & collaborating

- [x] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


mlflow/server/js/.eslintrc.js
```diff
@@ -129,7 +129,7 @@ module.exports = {
     }],
     'max-lines': [2, 1000],
     'max-nested-callbacks': 2,
-    'max-params': [2, 8],
+    'max-params': [2, 12],
```
smurching: Needed because the number of params passed to `Routes.getMetricPageRoute` is now quite large

mlflow/server/js/src/Routes.js
```diff
@@ -14,11 +16,51 @@ class Routes {
...
+   * @param plotMetricKeys - Array of string metric keys to plot
+   * @param plotLayout - Object containing plot layout information in Plotly format. See
+   *   https://plot.ly/javascript/plotlyjs-events/#update-data for an idea of object structure
+   * @param selectedXAxis - Enum (string) describing type of X axis (wall time, relative time, step)
+   * @param yAxisLogScale - Boolean - if true, y axis should be displayed on a log scale
+   *   (y axis scale is assumed to be linear otherwise)
+   * @param lineSmoothness - Float, coefficient >= 0 describing how much line smoothing to apply
+   * @param showPoint - Boolean, whether or not to show dots at individual data points in the metric
+   *   line plot
+   * @param deselectedCurves - Array of strings where each string describes a curve that was
```
smurching: Feels a bit weird to expose `deselectedCurves` in the page URL, but not sure there's a better way of doing this (see the comment)

mlflow/server/js/src/Routes.js
```diff
@@ -14,11 +16,51 @@ class Routes {
...
+   *   deselected / toggled off by the user (a curve that should not be displayed in the metric
+   *   plot). Strings are of the form "<runId>-<metricKey>". We describe the plot in terms
+   *   of deselected curves as we don't know a-priori which runs from
+   *   runUuids contain which of the metric keys in plotMetricKeys
+   * @param lastLinearYAxisRange - Array containing most recent bounds of a linear-scale y axis.
```
smurching: Definitely feels weird to expose `lastLinearYAxisRange` in the URL, as it's really closely coupled to the implementation of the plot - open to suggestions here, keeping it in does improve the UX of toggling between log & linear Y axes and keeps this PR simpler (as we also keep track of `lastLinearYAxisRange` in master)

mlflow/server/js/src/components/MetricsPlotControls.js
```diff
@@ -23,6 +23,9 @@ export class MetricsPlotControls extends React.Component {
     handleYAxisLogScaleChange: PropTypes.func.isRequired,
```
smurching: The main updates in this component (which renders the plot controls e.g. the line smoothness slider) are to pass plot configs as props, so that we can initialize e.g. the line smoothness slider with the smoothness specified in the plot URL, toggle the y-axis-log-scale switch if a log-scale y axis is enabled, etc.

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -36,23 +36,27 @@ export class MetricsPlotPanel extends React.Component {
...
+  legendClickTimeout = null;
+  // Time (millis after Unix epoch) since last legend click - if two clicks occur in short
+  // succession, we trigger a double-click event & cancel the pending single-click.
+  prevLegendClickTime = Math.inf;
+
   constructor(props) {
     super(props);
-    const plotMetricKeys = Utils.getPlotMetricKeysFromUrl(props.location.search);
-    const selectedMetricKeys = plotMetricKeys.length ? plotMetricKeys : [props.metricKey];
     this.state = {
```
smurching: `state` removed in favor of fetching state from the URL

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -73,12 +77,21 @@ export class MetricsPlotPanel extends React.Component {
     return runs ? JSON.parse(runs).length > 1 : false;
   }
 
-  updateUrlWithSelectedMetrics(selectedMetricKeys) {
+  // Update page URL from component state. Intended to be called after React applies component
+  // state updates, e.g. in a setState callback
+  updateUrlState = (updatedState) => {
```
smurching: Calls to `setState` replaced with calls to `updateUrlState` (which updates the page URL)

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -97,8 +110,9 @@ export class MetricsPlotPanel extends React.Component {
 
   getMetrics = () => {
     /* eslint-disable no-param-reassign */
-    const selectedMetricsSet = new Set(this.state.selectedMetricKeys);
```
smurching: Most of the changes below are replacing accesses to `this.state` with accesses to URL-derived state from `this.getUrlState()`

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -253,38 +276,88 @@ export class MetricsPlotPanel extends React.Component {
...
       mergedLayout = {
         ...mergedLayout,
         yaxis: { autorange: true, type: axisType },
       };
     }
-    this.setState({ layout: mergedLayout, lastLinearYAxisRange });
+    this.updateUrlState({ layout: mergedLayout, lastLinearYAxisRange });
+  };
+
+  /**
```
smurching: Here are the single-click & double-click handlers for curves in the plot legend

mlflow/server/js/src/components/MetricsPlotView.js
```diff
@@ -41,10 +44,14 @@ export class MetricsPlotView extends React.Component {
...
   getPlotPropsForLineChart = () => {
-    const { metrics, xAxis, showPoint, lineSmoothness, isComparing } = this.props;
+    const { metrics, xAxis, showPoint, lineSmoothness, isComparing,
+      deselectedCurves } = this.props;
+    const deselectedCurvesSet = new Set(deselectedCurves);
     const data = metrics.map((metric) => {
-      const { metricKey, runDisplayName, history } = metric;
+      const { metricKey, runDisplayName, history, runUuid } = metric;
       const isSingleHistory = history.length === 0;
+      const visible = !deselectedCurvesSet.has(Utils.getCurveKey(runUuid, metricKey)) ?
```
smurching: Don't display curves that were deselected by the user

Zangr (commented):
Left some comment. Mostly around simplifying the update loop. 

Overall, I'd recommend considering the URL as the actual `state` for the plot options and treat other components as static UI components for visualizing that state.

mlflow/server/js/src/Routes.js
smurching: Yep, makes sense that something like an "MLflow board" with support for actually persisting plot/dashboard configs in the background is the ideal long-term solution

mlflow/server/js/.eslintrc.js
smurching: SGTM, will make the change

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -253,38 +276,88 @@ export class MetricsPlotPanel extends React.Component {
...
+    if (currentTime - this.prevLegendClickTime < 300) {
+      this.handleLegendDoubleClick({curveNumber, data});
+      this.prevLegendClickTime = Math.inf;
+    } else {
+      // Otherwise, record time of current click & trigger click event
+      // Wait full double-click window to trigger setting state, and only if there was no
+      // double-click do we run the single-click logic (we wait a little extra to be safe)
+      const curveKey = Utils.getCurveKey(data[curveNumber].runId, data[curveNumber].metricName);
+      this.legendClickTimeout = window.setTimeout(() => {
+        const existingDeselectedCurves = new Set(state.deselectedCurves);
```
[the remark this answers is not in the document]
smurching: Unfortunately not, AFAICT - it seems the only way to extract information on legend clicks is by handling `plotly_legendclick` and `plotly_legenddoubleclick` events as described in the [Plotly docs](https://plot.ly/javascript/plotlyjs-events/#legend-click-events).

The primary reason for maintaining our own state around click times / handling double-clicks ourselves is that Plotly (seemingly incorrectly) calls our single-legend-click event handler twice on double clicks. Thus we need to make our single-click handler (`handleLegendClick`) aware of double-clicks & delegate to the double-click handler (`handleLegendDoubleClick`) when there are two clicks in short succession

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -253,38 +276,88 @@ export class MetricsPlotPanel extends React.Component {
...
+        if (existingDeselectedCurves.has(curveKey)) {
+          existingDeselectedCurves.delete(curveKey);
+        } else {
+          existingDeselectedCurves.add(curveKey);
+        }
+        this.updateUrlState({deselectedCurves: Array.from(existingDeselectedCurves)});
+      }, 310);
```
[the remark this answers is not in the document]
smurching: Good question, can add a clarifying comment. I took 300 ms as the maximum delay between two clicks to constitute a double click from the Plotly default ([see docs](https://plot.ly/javascript/configuration-options/#double-click-delay)), and picked 310 here to allow the full window for double-click to expire before running the single-click event.

I can actually increase the value from 310 to something larger to allow enough time for the double-click handler to run & cancel the single-click event - currently, if a user performs two clicks ~300 ms apart, the double-click handler will have only 10ms to cancel the single-click event scheduled here

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -36,23 +36,27 @@ export class MetricsPlotPanel extends React.Component {
...
-      // that we can restore them when converting from log back to linear scale)
-      lastLinearYAxisRange: [],
     };
-    this.loadMetricHistory(this.props.runUuids, this.state.selectedMetricKeys);
+    const state = this.getUrlState();
+    this.loadMetricHistory(this.props.runUuids, state.selectedMetricKeys);
+  }
+
+  getUrlState() {
+    return Utils.getMetricPlotStateFromUrl(this.props.location.search);
```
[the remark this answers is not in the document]
smurching: Good question, AFAICT this works in the same way as MetricsView used to work, where updates to the page URL (`history.push`) cause a rerender of the entire component tree (but React intelligently reuses existing DOM elements where possible to avoid flashing large portions of the page). I experimentally verified this is the case by adding a `console.log` in MetricPage's `render` method, and observed `MetricPage.render` being called both when UI actions update the URL & when manually editing the URL myself

smurching:
Thanks for the review @Zangr! Will sync w/ you tomorrow to discuss the comments & make sure I understand the best next steps for addressing them as well as

> Overall, I'd recommend considering the URL as the actual state for the plot options and treat other components as static UI components for visualizing that state.

Zangr (approved):
Synced offline. We are actually not deriving component state from URL related props. It's just a local variable named `state` that confused me a little bit. The follow up we can do it wrapping the URL parsing to a wrapper component to increase code clarity and reusability. Currently, this PR has no issue blocking itself from being merged so LGTM!

Issue #2347, opened 2020-01-27
[FR]


smurching:
## Describe the proposal
Currently, users can delete runs from the experiment UI, which leads to a confirmation prompt like:
![image](https://user-images.githubusercontent.com/2358483/73196992-6edeb900-40e5-11ea-97ed-4e79b615bacf.png)

We should update this prompt for parent runs to allow for deleting the parent & its child runs, as this is usually the intended behavior. For example, if you run hyperparam tuning, you may generate a parent run to capture e.g. the best accuracy/model generated across all trials, and record the individual trials as child runs of the parent. If you delete the hyperparam tuning run, you likely want to delete all associated child runs (trials) as well.

We can implement child-run deletion in JS by searching for runs via a [SearchRuns API request](https://www.mlflow.org/docs/latest/rest-api.html#search-runs) with `filter_string="tags.`mlflow.parentRunId` = 'beb6b6bb00e24951873efa253ff1e813'"` (may need to iterate over successive pages of runs to ensure we delete all child runs).

### Proposed Changes
For user-facing changes, what APIs are you proposing to add or modify? For internal changes, what code paths will need to be modified?

albertotb:
Are there any updates on this? I see that the related PR was closed but not merged

Exitare:
Is there any update on this? Is really counterintuitive to not delete child runs when the parent-run is being deleted.

Jingnan-Jia:
This is really a nice feature. Any update?

chenmoneygithub:
This proposal is pretty valid to me, is this currently supported? @dbczumar

pdornian:
Nth recommendation. Obvs MLflow, is independent from Microsoft Fabric, but I noticed the behaviour there. I delete my parent level run and then the UI immediately gets clogged with its 20 child runs that get promoted to the top level in the view rather than deleted.

PR #2395, opened 2020-02-06
[WIP] Update deletion modal to allow / recommend deleting nested runs along with parent runs

smurching:
Addresses #2347  by updating run deletion modal to allow / recommend deleting nested runs along with parent runs. 

In particular, when selecting parent runs for deletion, the confirmation modal now exposes a button for deleting children of the selected runs:
image

If only childless runs are selected, the UX is the same as before:
image

## How is this patch tested?

So far, manual tests - need to write unit tests

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/server/js/src/components/ExperimentView.js
+++ b/mlflow/server/js/src/components/ExperimentView.js
@@ -356,6 +356,13 @@ export class ExperimentView extends Component {
     const deleteDisabled = Object.keys(this.state.runsSelected).length < 1;
     const restoreDisabled = Object.keys(this.state.runsSelected).length < 1;
     const noteInfo = NoteInfo.fromTags(experimentTags);
+    const runsExpanded = persistedState.runsExpanded;
+    const parentsWithChildren = ExperimentViewUtil.getRowRenderMetadata({runInfos, tagsList,
+      runsExpanded}).filter((row) => row.childrenIds && row.childrenIds.length > 0);
+    const parentToChildRunIds = {};
+    parentsWithChildren.forEach(
+        (rowMetadata) => parentToChildRunIds[rowMetadata.runId] = rowMetadata.childrenIds);
+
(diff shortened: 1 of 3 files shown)


gioa:
What happens if a user selects a child run which has its own children, will we delete its children even though they are not shown on the UI? Can we tell whether it has children?

The 3-buttons works but maybe we should explain that the user has selected at least a run with children hence the button "Delete selected and children". I wonder whether a default-selected checkbox makes more sense here.  It can display more text and is also a common pattern.

smurching:
@gioa good questions:

> What happens if a user selects a child run which has its own children, will we delete its children even though they are not shown on the UI? 

Just to confirm, the current UI behavior is to nest all descendants of top-level parent runs under the top-level parent run (that is, child & grandchild, great-grandchild, etc runs are all nested under their top-level parent run in a flat hierarchy). 

When the top-level parent is deleted, all descendants of the parent that have been loaded into the browser will be deleted. We can also make a search API request against the server & delete immediate children of the top-level parent - it's unfortunately too expensive to search on each child to ensure we clean up all grandchild & more-deeply nested runs.

The end behavior will be that a) all loaded descendants of parent runs will be deleted, no matter how deeply nested and b) all immediate children of parent runs will be deleted, even if not loaded (but grandchild & more deeply nested runs may be orphaned). I'm hoping this covers most use cases, as I believe it's uncommon for people to use more than one level of run nesting, but could be wrong (happy to discuss this point more).

Either way, seems like it'd make sense to have a tooltip explaining the exact deletion behavior. I also like the checkbox idea, happy to implement that

PR #2408, opened 2020-02-09
Prevent metrics plots from resetting zoom when line smoothness or other attributes change

smurching:

This PR contains part of https://github.com/mlflow/mlflow/pull/2393. Namely, it updates the parent component of the metric plot (MetricPlotPanel) to store the plot's x & y axis attributes within component state, so that they can be passed down to the Plotly component used to render the actual plot. Previously, we didn't preserve modifications to x & y axis state anywhere, so rerendering the metric plot (e.g. as occurs when updating line smoothness) would reset the x & y axes. It also fixes React warnings in a few other components/files.
### Before
before-gif

(note that the zoom state of the plot is reset when adjusting line smoothness or the Y axis type)

### After
after-gif
(note that the zoom state of the plot is preserved)

## How is this patch tested?
Unit tests to verify conversions between log & linear y axis scale. Also manually tested by running this snippet to generate data for viewing bar/line charts (augmented the UI test data generation script with this snippet):

```
import mlflow
import random
from random import random as rand
for i in range(3):
    with mlflow.start_run():
        for j in range(10):
            sign = random.choice([-1, 1])
            mlflow.log_metric("myReallyLongTimeSeriesMetricName-abcdefghijklmnopqrstuvwxyz",
                              random.random() * sign)
            mlflow.log_metric("Another Timeseries Metric", rand() * sign)
            mlflow.log_metric("Yet Another Timeseries Metric", rand() * sign)
        if i == 0:
            mlflow.log_metric("Special Timeseries Metric", rand() * sign)
        mlflow.log_metric("Bar chart metric", rand())
```

Fixes issue where metric plot's zoom state would reset when line smoothness or other plot attributes were modified.

- [x] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/server/js/src/components/MetricView.js
+++ b/mlflow/server/js/src/components/MetricView.js
@@ -20,7 +20,7 @@ class MetricView extends Component {
 
   render() {
     const { experiment, runUuids, runNames, metricKey, location } = this.props;
-    const plotMetricKeys = Utils.getPlotMetricKeysFromUrl(location.search);
+    const { selectedMetricKeys } = Utils.getPlotMetricKeysFromUrl(location.search);
     return (
       <div className='MetricView'>
         <div className='header-container'>
@@ -28,7 +28,7 @@ class MetricView extends Component {
             experiment={experiment}
             runNames={runNames}
             runUuids={runUuids}
-            title={<span>{plotMetricKeys.length > 1 ? 'Metrics' : plotMetricKeys[0]}</span>}
+            title={<span>{selectedMetricKeys.length > 1 ? 'Metrics' : selectedMetricKeys[0]}</span>}
           />
         </div>
         <MetricsPlotPanel {...{ runUuids, metricKey }} />
--- a/mlflow/server/js/src/components/MetricView.js
+++ b/mlflow/server/js/src/components/MetricView.js
@@ -20,7 +20,7 @@ class MetricView extends Component {
 
   render() {
     const { experiment, runUuids, runNames, metricKey, location } = this.props;
-    const selectedMetricKeys = Utils.getPlotMetricKeysFromUrl(location.search);
+    const plotMetricKeys = Utils.getPlotMetricKeysFromUrl(location.search);
     return (
       <div className='MetricView'>
         <div className='header-container'>
@@ -28,7 +28,7 @@ class MetricView extends Component {
             experiment={experiment}
             runNames={runNames}
             runUuids={runUuids}
-            title={<span>{selectedMetricKeys.length > 1 ? 'Metrics' : selectedMetricKeys[0]}</span>}
+            title={<span>{plotMetricKeys.length > 1 ? 'Metrics' : plotMetricKeys[0]}</span>}
           />
         </div>
         <MetricsPlotPanel {...{ runUuids, metricKey }} />
--- a/mlflow/server/js/src/components/MetricView.js
+++ b/mlflow/server/js/src/components/MetricView.js
@@ -20,7 +20,7 @@ class MetricView extends Component {
 
   render() {
     const { experiment, runUuids, runNames, metricKey, location } = this.props;
-    const { selectedMetricKeys } = Utils.getPlotMetricKeysFromUrl(location.search);
+    const selectedMetricKeys = Utils.getPlotMetricKeysFromUrl(location.search);
     return (
       <div className='MetricView'>
         <div className='header-container'>
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -131,11 +131,13 @@ export class MetricsPlotPanel extends React.Component {
     const oldYRange = oldLayout.yaxis.range;
     if (oldLayout.yaxis.type !== 'log') {
       if (oldYRange[0] < 0) {
-        // When converting to log scale, handle negative values as follows:
-        // If bottom of old Y range is negative, then tell plotly to infer the y-axis scale
+        // When converting to log scale, handle negative values (which have no log-scale
+        // representation as taking the log of a negative number is not possible) as follows:
+        // If bottom of old Y range is negative, then tell plotly to infer the log y-axis scale
         // (set 'autorange' to true), and preserve the old range in the 'range' attribute of
         // the layout so that we can restore it if the user converts back to a linear-scale
-        // y axis.
+        // y axis. We defer to Plotly's autorange here under the assumption that it will produce
+        // a reasonable y-axis log scale for plots containing negative values.
         newLayout.yaxis = {
           type: 'log',
           range: oldYRange,
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -45,7 +45,6 @@ export class MetricsPlotPanel extends React.Component {
       selectedMetricKeys,
       showPoint: false,
       historyRequestIds: [],
-      yAxisLogScale: false,
       lineSmoothness: 0,
     };
     this.loadMetricHistory(this.props.runUuids, this.state.selectedMetricKeys);
@@ -117,24 +116,12 @@ export class MetricsPlotPanel extends React.Component {
     // If yaxis was already explicitly specified, convert range to appropriate coordinates
     // for log axis (base 10), and vice versa. When converting to log scale, handle negative values
     // by deferring to plotly autorange
-    console.log("@SID got layout state " + JSON.stringify(state.layout));
-
-    // Possible layout types
-    // {yaxis: {range: []}}
-    // {yaxis: {autorange: true}}
-    // {yaxis: {autorange: true, range: [...]}}
-
-    //
-    // if (state.layout.yaxis && state.layout.yaxis.range) {
-    //   this.updateExistingYAxisLogScale()
-    // } else {
-    //   newLayout.yaxis = {autorange: true, type: "log"};
-    // }
+    const newAxisType = yAxisLogScale ? "log" : "linear";
 
     // If plot previously had no y axis range configured, simply set the axis type to log or
     // linear scale appropriately
     if (!state.layout.yaxis || !this.state.layout.yaxis.range) {
-      newLayout.yaxis = {type: yAxisLogScale ? "log" : "linear", autorange: true};
+      newLayout.yaxis = {type: newAxisType, autorange: true};
       this.setState({layout: newLayout});
       return;
     }
@@ -143,7 +130,6 @@ export class MetricsPlotPanel extends React.Component {
     // Convert the range to/from log scale as appropriate
     const oldYRange = state.layout.yaxis.range;
     if (state.layout.yaxis.type === 'log') {
-      console.log("Old state was log scale, converting back to linear scale");
       // We at this point know that there was an old y scale, so should convert it to/from
       // log scale.
       // When converting from log scale to linear scale, only apply conversion if autorange
@@ -160,9 +146,7 @@ export class MetricsPlotPanel extends React.Component {
         };
       }
     } else {
-      console.log("Old state was linear scale, converting to log scale");
       if (oldYRange[0] < 0) {
-        console.log("Linear-to-log: Y axis range negative, applying autorange");
         // When converting to log scale, handle negative values as follows:
         // If bottom of old Y range is negative, then simply autorange the plot, so that we
         // can convert back. Otherwise, do the conversion
@@ -172,14 +156,12 @@ export class MetricsPlotPanel extends React.Component {
           autorange: true,
         };
       } else {
-        console.log("Linear-to-log: Y axis range positive, applying standard conversion");
         newLayout.yaxis = {
           type: 'log',
           range: [Math.log(oldYRange[0]) / Math.log(10), Math.log(oldYRange[1]) / Math.log(10)],
         };
       }
     }
-    console.log("Setting layout state to " + JSON.stringify(newLayout));
     this.setState({layout: newLayout});
   };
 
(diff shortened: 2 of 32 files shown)


Zangr:
jenkins build-classic

mlflow/server/js/src/components/MetricView.js
```diff
@@ -20,15 +20,15 @@ class MetricView extends Component {
 
   render() {
     const { experiment, runUuids, runNames, metricKey, location } = this.props;
-    const plotMetricKeys = Utils.getPlotMetricKeysFromUrl(location.search);
+    const selectedMetricKeys = Utils.getPlotMetricKeysFromUrl(location.search);
```
andychow-db: Why the name switch?

harupy:
This is awesome. #2295 could be much simpler.

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,13 +111,117 @@ export class MetricsPlotPanel extends React.Component {
...
+        type: axisType,
+      },
+    };
+    this.setState({ selectedXAxis: e.target.value, layout: newLayout });
+  };
+
+  handleLayoutChange = (newLayout) => {
+    // Unfortunately, we need to parse out the x & y axis range changes from the onLayout event...
+    // see https://plot.ly/javascript/plotlyjs-events/#update-data
+    const newXRange0 = newLayout["xaxis.range[0]"];
```
Zangr: Does it have to be accessed using a path? If not we can simplify this as 
```
const [newXRangeStart, newXRangeEnd] = newLayout.xaxis;
```

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,13 +111,117 @@ export class MetricsPlotPanel extends React.Component {
...
+        ...mergedLayout,
+        yaxis: {
+          range: [newYRange0, newYRange1],
+        },
+      };
+    }
+    if (newLayout["xaxis.autorange"]) {
+      mergedLayout = {
+        ...mergedLayout,
+        xaxis: {autorange: true},
```
Zangr:
```suggestion
        xaxis: { autorange: true },
```

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,13 +111,117 @@ export class MetricsPlotPanel extends React.Component {
...
+      };
+    }
+    if (newLayout["yaxis.autorange"]) {
+      const axisType = this.state.layout && this.state.layout.yaxis &&
+        this.state.layout.yaxis.type === 'log' ? "log" : "linear";
+      mergedLayout = {
+        ...mergedLayout,
+        yaxis: {autorange: true, type: axisType},
```
Zangr:
```suggestion
        yaxis: { autorange: true, type: axisType },
```

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,13 +111,117 @@ export class MetricsPlotPanel extends React.Component {
...
+      };
+    }
+    this.setState({layout: mergedLayout});
   };
 
+
```
Zangr:
```suggestion

```

mlflow/server/js/src/test-utils/Fixtures.js
```diff
@@ -1,7 +1,7 @@
 import { Experiment } from '../sdk/MlflowMessages';
 
 const createExperiment = ({
-  experiment_id = '0',
+  experiment_id = 0,
```
Zangr: I think currently experiment_id is a string in the proto, we need to adjust the propTypes to `string`.

tests/generate_ui_test_data.py
```diff
@@ -49,6 +49,20 @@ def rand_str(max_len=40):
             log_params(parameters)
             log_metrics(metrics)
 
+    # Runs with multiple values for a single metric so that we can QA the time-series metric
+    # plot
+    for i in range(3):
```
Zangr: Thanks for adding this to test-gen script!

Zangr (commented):
Changes look pretty clean. I did a round of superficial checks. Could we have a before/after gif to help better understand the use cases this PR addresses?

mlflow/server/js/src/components/MetricsPlotPanel.js
smurching: Unfortunately it does - the actual key is called `"xaxis.range[0]"`

mlflow/server/js/src/test-utils/Fixtures.js
smurching: So it actually seems that `experiment_id` is an int still in the proto-generated JS messages (even though the protobuf itself has since been updated to make `experiment_id` a string). Thus the real correct long-term fix here is to update the proto-generated JS messages to reflect the current set of protobuf types. I'll revert this Fixtures.js change as it's incompatible with that future change (but will defer updating the protobuf types to a future PR as it seems a bit risky & I don't want to bundle it in with this other code).

AveshCSingh (changes requested):
Nice! I like how this PR utilizes plotly's existing props to preserve the axis attributes.

I have a bunch of suggestions and questions inline. PTAL.

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,11 +110,116 @@ export class MetricsPlotPanel extends React.Component {
...
+        range: [Math.pow(10, oldYRange[0]), Math.pow(10, oldYRange[1])],
+      };
+    }
+    this.setState({layout: newLayout});
   };
 
   handleXAxisChange = (e) => {
-    this.setState({ selectedXAxis: e.target.value });
+    // Set axis value type, & reset axis scaling via autorange
+    const axisType = e.target.value === X_AXIS_WALL ? "date" : "linear";
```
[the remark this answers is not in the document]
smurching: Oops, yeah good catch it does - will fix this, thanks!

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,11 +110,116 @@ export class MetricsPlotPanel extends React.Component {
...
+      this.setState({ layout: newLayout });
+      return;
+    }
+
+    // At this point, we know the plot previously had a y axis specified with range
+    // Convert the range to/from log scale as appropriate
+    const oldLayout = this.state.layout;
+    const oldYRange = oldLayout.yaxis.range;
+    if (oldLayout.yaxis.type !== 'log') {
+      if (oldYRange[0] < 0) {
```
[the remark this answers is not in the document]
smurching: Yeah definitely - was hoping to be able to use a symlog scale, but Plotly doesn't support it yet: https://github.com/plotly/plotly.js/issues/221. Will add a TODO linking to the Plotly issue!

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,11 +110,116 @@ export class MetricsPlotPanel extends React.Component {
...
+    // Unfortunately, we need to parse out the x & y axis range changes from the onLayout event...
+    // see https://plot.ly/javascript/plotlyjs-events/#update-data
+    const newXRange0 = newLayout["xaxis.range[0]"];
+    const newXRange1 = newLayout["xaxis.range[1]"];
+    const newYRange0 = newLayout["yaxis.range[0]"];
+    const newYRange1 = newLayout["yaxis.range[1]"];
+    let mergedLayout = {
+      ...this.state.layout,
+    };
+    if (newXRange0) {
```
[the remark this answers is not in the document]
smurching: Yeah good question, [Plotly docs](https://plot.ly/javascript/plotlyjs-events/#update-data) are a little ambiguous on this, stating: 

> For Cartesian (2D) plots, plotly_relayout emits only the xaxis and yaxis ranges which were directly changed by the triggering event

But not whether partial ranges are provided in the event only one coordinate changes. AFAICT though, they provide the full range whenever either range bound changes, e.g. as in [this GIF](https://recordit.co/U4bunEcs5n).

To ensure we don't end up with weird `undefined` values, we can check that both range values are provided (i.e. that `newXRange0` and `newXRange1` are both defined - will also update the check to explicitly check that the values are defined to properly handle the case where they're set to 0). 

More defensive would be to merge each of `newXRange0` and `newXRange1` into the existing X axis range (if one exists) & then set the resulting range - but would prefer not to introduce this complexity

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -106,11 +110,116 @@ export class MetricsPlotPanel extends React.Component {
...
+        // When converting to log scale, handle negative values (which have no log-scale
+        // representation as taking the log of a negative number is not possible) as follows:
+        // If bottom of old Y range is negative, then tell plotly to infer the log y-axis scale
+        // (set 'autorange' to true), and preserve the old range in the 'range' attribute of
+        // the layout so that we can restore it if the user converts back to a linear-scale
+        // y axis. We defer to Plotly's autorange here under the assumption that it will produce
+        // a reasonable y-axis log scale for plots containing negative values.
+        newLayout.yaxis = {
+          type: 'log',
+          range: oldYRange,
```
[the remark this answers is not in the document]
smurching: Yeah good call, it seems to work but definitely doesn't seem safe based on that doc - will update to store the old y range in a separate `state` attribute :)

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -109,37 +114,57 @@ export class MetricsPlotPanel extends React.Component {
...
+    // lastLinearYAxisRange contains the last range used for a linear-scale y-axis
+    // We reset this state attribute each time the y-axis log scale changes, unless we're explicitly
+    // converting from a linear axis with negative bounds to a log scale axis.
+    let lastLinearYAxisRange = [];
+
+    // Handle special case of a linear y-axis scale with negative values converted to log scale &
+    // now being restored to linear scale, by restoring the old linear-axis range from
+    // this.state.linearYAxisRange
+    if (!yAxisLogScale && this.state.lastLinearYAxisRange &&
+        this.state.lastLinearYAxisRange.length > 0) {
```
AveshCSingh: How do you know that the linear y-axis scale we're changing to has negative values?

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -109,37 +114,57 @@ export class MetricsPlotPanel extends React.Component {
...
+      newLayout.yaxis = {
+        type: "linear",
+        range: this.state.lastLinearYAxisRange,
+      };
+      this.setState({ layout: newLayout, lastLinearYAxisRange });
```
AveshCSingh: I think `lastLinearYAxisRange == []` at this point. Is that right?

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -148,27 +173,29 @@ export class MetricsPlotPanel extends React.Component {
...
+   */
   handleXAxisChange = (e) => {
     // Set axis value type, & reset axis scaling via autorange
-    const axisType = e.target.value === X_AXIS_WALL ? "date" : "linear";
+    const axisEnumToPlotlyType = {
+      [X_AXIS_WALL]: "date",
+      [X_AXIS_RELATIVE]: "linear",
+      [X_AXIS_STEP]: "linear",
+    };
+    const axisType = axisEnumToPlotlyType[e.target.value] || "linear";
```
AveshCSingh: Nice!

mlflow/server/js/src/components/MetricsPlotPanel.js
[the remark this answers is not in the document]
AveshCSingh: Ah, good to know!

mlflow/server/js/src/components/MetricsPlotPanel.js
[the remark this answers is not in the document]
AveshCSingh: Ah my mistake. I like the that you created `axisEnumToPlotlyType` to avoid this mistake in the future!

mlflow/server/js/src/components/MetricsPlotPanel.js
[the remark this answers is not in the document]
AveshCSingh: Yeah, agreed that that's overkill since the library currently provides the full range even if only one bound changes.

AveshCSingh (commented):
Thanks for addressing my comments! Two other quick questions, otherwise lgtm

mlflow/server/js/src/components/MetricsPlotPanel.js
smurching: I guess here we're assuming that if `lastLinearYAxisRange` is non-empty, it contains a linear y axis range with negative values. Maybe I can update the name/add a comment to make this clearer...

mlflow/server/js/src/components/MetricsPlotPanel.js
smurching: Yep that's right, can make this clearer by just explicitly setting to []

mlflow/server/js/src/components/MetricsPlotPanel.js
AveshCSingh: Ah got it. Your comment below really clears this up --thanks! (` We set this state attribute if and only if we're converting from ...`)

AveshCSingh (approved):
Thanks for answering my last few questions. LGTM!

PR #2415, opened 2020-02-10
Project backend plugin system

jdlesage:

This pull request implements a plugin system to be able to propose new MLflow project backends.
It is a sort of RFC: there is doc modification, the new interface and the function to load a new backend (relying on python `entrypoint`)

It copies most of the logic of other MLflow plugins such as ArtifactStores, etc.

Databricks backend is the reference implemenation

## How is this patch tested?

1/ Install the mlflow-test-plugin project. It contains a "dummy" backend
2/ Go to examples/sklearn_elasticnet_wine and run `mlflow run --no-conda --backend dummy-backend .`
3/ It generates an entry in default experiment of MLflow. `'mlflow.project.backend` tag is  `dummy`

It introduces a plugin system to extend MLFlow Projects backends.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [X] Docs
- [ ] Tracking
- [X] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


jdlesage:
Updated the "how is this patch tested?` with the dummy example

mlflow/projects/backend/abstract_backend.py
```diff
@@ -0,0 +1,21 @@
...
+
+class AbstractBackend():
+    """
+    Abstract class for MLflow Project Execution Backend
+    A backend must implement the run method to launch a new job
+    """
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def run(self, active_run, uri, entry_point, work_dir, parameters,
```
smurching: Would you mind adding a docstring for this method describing the params & their types? 

Also, it'd be good to write out the possible options for this method signature & the pros/cons of the different options, e.g. another option could be to accept **kwargs here, but we might prefer that users just specify custom backend-specific arguments via `cluster_spec` (which BTW I think should be renamed to `backend_config`?)

It'd also be good to write out a high-level description of how this interface would work for the kubernetes & local backends, given that it excludes arguments like `storage_dir` that are passed to the local backend (maybe we just special-case such arguments & handle them before delegating to the plugin?

mlflow/projects/backend/abstract_backend.py
```diff
@@ -0,0 +1,34 @@
...
+    def run(self, active_run, uri, entry_point, parameters,
+            backend_config):
+        """
+        Submit an entrypoint. It must returns a SubmittedRun object to track the execution
+
+        :param active_run: Current MLflow run object. Everything will registred in this run.
+        :param uri: uri to the project (could be local or git uri).
+                    This is the parameter given to mlflow run command.
+        :param entry_point: name of the entry point to execute.
+        :param parameters: Dict of parameters to pass to the entry point
```
smurching: I wonder if we should keep `project_dir` (the local filesystem directory containing the fetched project) as a param here - it is common across backends & may not be obvious to plugin implementors otherwise that they don't actually need to fetch the project from `uri` in order to get access to its contents

mlflow/projects/backend/abstract_backend.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def run(self, active_run, uri, entry_point, parameters,
+            backend_config):
+        """
+        Submit an entrypoint. It must returns a SubmittedRun object to track the execution
+
+        :param active_run: Current MLflow run object. Everything will registred in this run.
```
smurching: Could `active_run` be replaced with `run_id` containing just the ID of the MLflow run? And then to fetch the experiment ID, individual backends can call `run = mlflow.get_run(run_id); experiment_id = run.info.run_id`? Would be nice to keep the interface simpler for backend implementors (so that they don't necessarily need to learn the fields of `active_run`)

mlflow/projects/backend/abstract_backend.py
smurching:
```suggestion
        :param params: Dict of parameters to pass to the entry point
```

Just a small nit/suggestion on the argument name, for conciseness

mlflow/projects/backend/abstract_backend.py
```diff
@@ -0,0 +1,34 @@
...
+        :param uri: uri to the project (could be local or git uri).
```
smurching:
```suggestion
        :param project_uri: URI of the project (could be a local or git URI).
```

mlflow/projects/backend/abstract_backend.py
```diff
@@ -0,0 +1,34 @@
...
+                    This is the parameter given to mlflow run command.
+        :param entry_point: name of the entry point to execute.
+        :param parameters: Dict of parameters to pass to the entry point
+        :param backend_config: Dict to pass parameters to the backend
+
+        :return: A SubmittedRun. This function is asynchronous, it should only trigger the run and
```
smurching:
```suggestion
        :return: A :py:class:`mlflow.projects.SubmittedRun`. This function is expected to run the project asynchronously, i.e. it should trigger project execution and then immediately return a `SubmittedRun` to track execution status.
```

mlflow/projects/backend/abstract_backend.py
jdlesage: done

mlflow/projects/backend/abstract_backend.py
jdlesage: done

mlflow/projects/backend/abstract_backend.py
jdlesage: done

mlflow/projects/backend/abstract_backend.py
jdlesage: done

mlflow/projects/backend/abstract_backend.py
jdlesage: done

smurching (commented):
Thanks for all the updates @jdlesage! One high-level comment, in order to modularize the changes & facilitate review, could we split this PR up into the following sub-PRs? I think they can be sequenced as follows:
   1. Adding the AbstractBackend interface & plugin infrastructure, refactoring the test plugin & updating plugin docs
  2. conftest.py changes (these are awesome btw! should save other developers a ton of time)
  3. Refactoring the Databricks backend to extend the AbstractBackend

I think 1 & 2 can actually be opened in parallel, with 3 to follow merging both of them (& after 3, we can merge an additional PR containing the refactoring you've already done of the k8s and local backends).

I'd expect 2 & 3 to be pretty easy to get merged, with 1 requiring the most (if any) iteration.

Thanks again for working on this! It'll be a great enhancement for MLflow & enable a ton of new use cases, and let me know if you'd like any help splitting up the PRs etc

aarondav:
@jdlesage Did you get a chance to look at @smurching's comments?

jdlesage:
Hi. I took a break last week... I didn't progress on this PR 

I will rebase the PR to the lastest version (to solve conflicts)  and take in account @smurching remarks asap.

jdlesage:
Hi @aarondav 
This PR has been split. The one that will be merged is https://github.com/mlflow/mlflow/pull/2566. It contains "only" the plugin part.

This PR is confusing, I close it.

PR #2430, opened 2020-02-14
[mlflow/R] Update R client to call record logged model 

tomasatdatabricks:
Followup to #2369
Update R client to call new api to record logged model metadata with the tracking server.

## How is this patch tested?
Unit test.

## Release Notes
n/a

Followup to previous PR, adds functionality to R client.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [x] R
- [ ] Java
- [ ] Python


mlflow/R/mlflow/R/model.R
```diff
@@ -28,8 +29,22 @@ mlflow_save_model <- function(model, path, ...) {
...
-  mlflow_save_model(model, path = temp_path, ...)
-  mlflow_log_artifact(path = temp_path, artifact_path = artifact_path)
+  model_spec <- mlflow_save_model(model, path = temp_path, model_spec = list(
+    utc_time_created = mlflow_timestamp(),
+    run_id = mlflow_get_active_run_id(),
+    artifact_path = artifact_path,
+    flavors = list()
+  ), ...)
+  res <- mlflow_log_artifact(path = temp_path, artifact_path = artifact_path)
+  mlflow_record_logged_model(model_spec)
```
smurching: Should we try-catch this to handle the case where the server doesn't have the log-model endpoint implemented?

smurching (commented):
Just one comment around adding a try-catch, looks good otherwise!

mlflow/R/mlflow/R/model.R
tomasatdatabricks: yes, we should! good catch

mlflow/server/handlers.py
```diff
@@ -453,7 +453,6 @@ def _log_batch():
 def _log_model():
     request_message = _get_request_message(LogModel())
     try:
-        _logger.error(request_message.model_json)
```
tomasatdatabricks: This was forgotten in earlier python PR, we should remove it.

smurching (approved):
LGTM

PR #2432, opened 2020-02-14
Show only differing parameter for runs

tilman151:

In some projects (e.g. hyperparameter search for deep learning) a large number of parameters are logged to each run. The UI always displays all parameters although most of them are the same for all runs. This makes comparing runs difficult, as one run alone fills the whole height of the screen.

This pull request adds an optional parameter to the search_runs API: `parameter_diff`. If this is set to true, only the parameters that are different are returned for the runs. For paginated search requests, only the current page is considered. This functionality is usable in the Python API. The functionality is usable in the UI, too.

The main contribution lies in the backend code handling the parameter diff. The integration in the UI is only as an example because I am not that knowledgable in React.

## How is this patch tested?

Tests of the correctness of the functionality were added in all Python API levels:

- test_file_store.py
- test_rest_store.py
- test_sqlalchemy_store.py
- test_abstract_store.py
- test_client.py
- test_tracking.py

The tests cover the following scenarios:

- The returned parameters for a single run are empty, as this is the trivial case
- The returned parameters for runs with all the same parameter keys and values are empty
- For runs with different parameters, only the different ones are returned
- The parameter_key -> parameter_value pairings for the diffed parameters are consistent with the undiffed ones

- [ ] No.
- [x] Yes.

The search_runs function can now filter the parameters of the runs to include only the ones that differ. This functionality is added to the UI, too.

- [x] UI
- [ ] CLI
- [x] API
- [x] REST-API
- [ ] Examples
- [ ] Docs
- [x] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python


tilman151:
@tomasatdatabricks 

The CI pipeline fails at the large Python 2.7 tests while installing dependencies with pip. Is there an issue in the pipeline or does this stem from my PR?

tomasatdatabricks:
Hi @tilman151, it looks like it's not related to your PR. I'll take a look today.

tilman151:
Sorry for bothering but is  there any progress with the pipeline?

tomasatdatabricks:
Hi @tilman151, sorry I forgot to post back here!

The tests should be fixed now. 

I have also briefly looked at your PR - it looks good and I think this feature would be really valuable, however, I also think it belongs to the ui code instead of the api. What do you think @Zangr @gioa?

aarondav:
Does the "Columns" dropdown help with this use-case from the frontend side, which lets you hide those columns that you don't want to see?

image

tilman151:
Yes, this feature really helps and it's how I do it right now. The problem there is, that I have to check the columns manually every time I start the UI. This is really not fun, when you have many hyperparameters and have to remember exactly which ones you varied. Especially when you started your experiments two weeks ago.

Setting them automatically would be my preferred solution.

aarondav:
Makes sense @tilman151! What if the settings for Columns were persisted in local storage, so you only have to specify them once? @Zangr

tilman151:
This could be a good compromise. It leaves the possibility open to automatically determine which columns should be active in a later PR.

jhagege:
> Makes sense @tilman151! What if the settings for Columns were persisted in local storage, so you only have to specify them once? @Zangr

Hi, sounds a very interesting feature. 
Do you know if it is available yet ?
If so, how to activate ?

Issue #1945, opened 2019-10-16
[FR] Allow passing in bool parameters in MLproject entrypoint


tanmaymathur89:
## Proposal
```name: My Project

conda_env: my_env.yaml

entry_points:
  main:
    parameters:
      data_file: path
      regularization: {type: float, default: 0.1}
    command: "python script.py -r {regularization} {data_file}"
  validate:
    parameters:
      data_file: path
    command: "python validate.py {data_file}"
```
Only allows you to pass a string, float, path or uri. 
However, python allows you to pass bool command line arguments that we cannot fit into this format.

### Motivation
Command I want to execute is `python script.py --train -r {regularization} {data_file}`
where --train is an optional boolean argument.

### Proposed Changes
```name: My Project

conda_env: my_env.yaml

entry_points:
  main:
    parameters:
      data_file: path
      regularization: {type: float, default: 0.1}
      train : {type: bool, default: False}      
    command: "python script.py -r  {regularization} {train} {data_file} "
  validate:
    parameters:
      data_file: path
    command: "python validate.py {data_file}"
```
MLflow run execution: `mlflow run . -P data_file=data.csv train=True`
translates to `python script.py --train -r 0.1  data.csv`

lorenzwalthert:
I proposed this in #1572 and it was rejected for reasons I still don't understand.

andrewbrooks-o:
Also looking for boolean parameter support.

juntai-zheng:
Hi @tanmaymathur89, @lorenzwalthert, and @andrewbrooks-o! Thanks for your feedback - this seems like something that would be nice to support. Would any of you be willing to make a PR to add in this capability?

tanmaymathur89:
@juntai-zheng I can create a PR for it. Thanks

dbczumar:
@tanmaymathur89 @lorenzwalthert @andrewbrooks-o I think that adding `boolean` parameter support sounds great. #2453 currently introduces support for an `action` parameter, which appears to be relatively Python or bash-specific; @tanmaymathur89 can we refactor this to introduce support for True/False `boolean` parameters rather than `actions`?

sn1c:
@dbczumar is this still open? If @tanmaymathur89 won't do it, I'd prepare a PR for adding boolean parameter support.

tanmaymathur89:
> @dbczumar is this still open? If @tanmaymathur89 won't do it, I can prepare a PR for adding boolean parameter support.

Go ahead @sn1c.
[...]

PR #2453, opened 2020-02-20
Add action type to parameter

tanmaymathur89:
Add action type to MLProjects parameters. Currently there is support for string, uri, path and float. 

Related issue:
https://github.com/mlflow/mlflow/issues/1945

However, to make this consistent with current parameter support there is no support for making the parameters optional but rather allow users to pass default value.

## How is this patch tested?
Unit tested and tests are included in test_entry_point in projects tests.

Add support for action type.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [x] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### DIFF
--- a/docs/source/projects.rst
+++ b/docs/source/projects.rst
@@ -171,7 +171,8 @@ file in YAML syntax, to the project's root directory. The following is an exampl
         parameters:
           data_file: path
           regularization: {type: float, default: 0.1}
-        command: "python train.py -r {regularization} {data_file}"
+          args: {type: action, default: '--earlystopping'}
+        command: "python train.py -r {regularization} {data_file} {args}"
       validate:
         parameters:
           data_file: path
@@ -308,8 +309,12 @@ uri
     that know how to read from distributed storage (e.g., programs that use Spark).
 
 action
-    An action string. MLflow validates that the parameter is valid. Pass an empty string
-    to skip the argument
+    An action string. Pass an empty string to skip the argument.
+    More information on usage of action parameter can be found
+    `here https://docs.python.org/dev/library/argparse.html#action`_.
+    MLflow validates that the action parameter value is a
+    `valid identifier <https://docs.python.org/3/reference/lexical_analysis.html#identifiers>`_
+    that is prefixed either with single/double hyphen.
 
 .. _running-projects:
 
--- a/docs/source/projects.rst
+++ b/docs/source/projects.rst
@@ -307,6 +307,10 @@ uri
     relative paths to absolute paths, as in the ``path`` type. Use this type for programs
     that know how to read from distributed storage (e.g., programs that use Spark).
 
+action
+    An action string. MLflow validates that the parameter is valid. Pass an empty string
+    to skip the argument
+
 .. _running-projects:
 
 Running Projects
--- a/mlflow/data.py
+++ b/mlflow/data.py
@@ -65,6 +65,12 @@ def is_uri(string):
     return len(parsed_uri.scheme) > 0
 
 
+def is_action(action_string):
+    pattern = re.compile("^-{1,2}[a-zA-z]+")
+    return pattern.match(action_string)
+
+
+
 def download_uri(uri, output_path):
     if DBFS_REGEX.match(uri):
         _fetch_dbfs(uri, output_path)
--- a/mlflow/data.py
+++ b/mlflow/data.py
@@ -10,6 +10,7 @@ from mlflow.utils import process
 DBFS_PREFIX = "dbfs:/"
 S3_PREFIX = "s3://"
 GS_PREFIX = "gs://"
+ACTION_REGEX = re.compile("^-{1,2}[^\d\W]\w*\Z")
 DBFS_REGEX = re.compile("^%s" % re.escape(DBFS_PREFIX))
 S3_REGEX = re.compile("^%s" % re.escape(S3_PREFIX))
 GS_REGEX = re.compile("^%s" % re.escape(GS_PREFIX))
@@ -66,8 +67,9 @@ def is_uri(string):
 
 
 def is_action(action_string):
-    pattern = re.compile("^-{1,2}[a-zA-z]+")
-    return pattern.match(action_string)
+    if len(action_string.strip()) > 0:
+        return ACTION_REGEX.match(action_string)
+    return True
 
 
 def download_uri(uri, output_path):
--- a/mlflow/data.py
+++ b/mlflow/data.py
@@ -70,7 +70,6 @@ def is_action(action_string):
     return pattern.match(action_string)
 
 
-
 def download_uri(uri, output_path):
     if DBFS_REGEX.match(uri):
         _fetch_dbfs(uri, output_path)
--- a/mlflow/data.py
+++ b/mlflow/data.py
@@ -10,7 +10,7 @@ from mlflow.utils import process
 DBFS_PREFIX = "dbfs:/"
 S3_PREFIX = "s3://"
 GS_PREFIX = "gs://"
-ACTION_REGEX = re.compile("^-{1,2}[^\d\W]\w*\Z")
+ACTION_REGEX = re.compile(r"^-{1,2}[^\d\W]\w*\Z")
 DBFS_REGEX = re.compile("^%s" % re.escape(DBFS_PREFIX))
 S3_REGEX = re.compile("^%s" % re.escape(S3_PREFIX))
 GS_REGEX = re.compile("^%s" % re.escape(GS_PREFIX))
--- a/mlflow/projects/_project_spec.py
+++ b/mlflow/projects/_project_spec.py
@@ -13,7 +13,7 @@ from mlflow.utils.string_utils import is_string_type
 
 MLPROJECT_FILE_NAME = "mlproject"
 DEFAULT_CONDA_FILE_NAME = "conda.yaml"
-EMPTY_STRING_REGEX = "\s+((\"\")|(\'\'))"
+EMPTY_STRING_REGEX = r"\s+((\"\")|(\'\'))"
 
 
 def _find_mlproject(directory):
--- a/mlflow/projects/_project_spec.py
+++ b/mlflow/projects/_project_spec.py
@@ -2,7 +2,7 @@
 
 import os
 import yaml
-
+import re
 from six.moves import shlex_quote
 
 from mlflow import data
@@ -13,6 +13,7 @@ from mlflow.utils.string_utils import is_string_type
 
 MLPROJECT_FILE_NAME = "mlproject"
 DEFAULT_CONDA_FILE_NAME = "conda.yaml"
+EMPTY_STRING_REGEX = "\s+((\"\")|(\'\'))"
 
 
 def _find_mlproject(directory):
@@ -164,11 +165,17 @@ class EntryPoint(object):
         command_with_params = self.command.format(**params)
         command_arr = [command_with_params]
         command_arr.extend(["--%s %s" % (key, value) for key, value in extra_params.items()])
-        return " ".join(command_arr)
+        command = " ".join(command_arr)
+        return self._sanitize_command(command)
 
     @staticmethod
     def _sanitize_param_dict(param_dict):
         return {str(key): shlex_quote(str(value)) for key, value in param_dict.items()}
+
+    @staticmethod
+    def _sanitize_command(command):
+        # replace '' or "" with empty space, special case for action type parameter
+        return re.sub(EMPTY_STRING_REGEX, "", command).strip()
 
 
 class Parameter(object):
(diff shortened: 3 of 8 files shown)


mlflow/data.py
```diff
@@ -65,6 +65,11 @@ def is_uri(string):
     return len(parsed_uri.scheme) > 0
 
 
+def is_action(action_string):
+    pattern = re.compile("^-{1,2}[a-zA-z]+")
```
juntai-zheng: This doesn't cover all the valid Python identifiers (e.g. --action_, --act10n, etc.). I would recommend checking [this SO post](https://stackoverflow.com/questions/5474008/regular-expression-to-confirm-whether-a-string-is-a-valid-identifier-in-python) for a more thorough check.

mlflow/projects/_project_spec.py
```diff
@@ -202,10 +201,18 @@ def _compute_path_value(self, user_param_value, storage_dir):
             data.download_uri(uri=user_param_value, output_path=dest_path)
         return os.path.abspath(dest_path)
 
+    def _compute_action_value(self, user_param_value):
+        if not data.is_action(user_param_value):
+            raise ExecutionException("Expected parameter %s but got "
```
juntai-zheng:
```suggestion
            raise ExecutionException("Expected action %s but got "
```

tests/resources/example_project/MLproject
```diff
@@ -20,6 +20,12 @@ entry_points:
     parameters:
       duration: int
     command: "sleep {duration}"
+  greeter_compute_params:
```
juntai-zheng: Could we just add `args` to the `greeter` entry point and not have this separate entry point?

mlflow/data.py
juntai-zheng: Ideally, we would be able to cover the case where we have an action that is toggled by an argument being present or not (e.g. passing in `--earlystopping` as an argument if we want early stopping, and not passing it in if we don't want early stopping). For example, if I wanted to specify that I don't want early stopping, I would call 
```
mlflow run . -P data_file=data.csv -P stop=''
```

 and if I wanted early stopping I would call
 
```
mlflow run . -P data_file=data.csv -P stop=--earlystopping
```

It'd be good if you could add handling of an indicator (something like an empty string) that signals that the user does not want to pass in the action to the command.

tests/projects/test_entry_point.py
```diff
@@ -15,17 +15,17 @@ def test_entry_point_compute_params():
     Tests that EntryPoint correctly computes a final set of parameters to use when running a project
     """
     project = load_project()
-    entry_point = project.get_entry_point("greeter")
+    entry_point = project.get_entry_point("greeter_compute_params")
```
juntai-zheng: We should be adding tests that include cases where we don't want to include the action. We should also add in tests that check for failures when the action is used improperly to make sure the feature errors out as expected.

juntai-zheng (commented):
Hi @tanmaymathur89, thanks for your work! I left some comments, plus I have a few other remarks:

- Tests aren't fully covering this feature yet. We'll have to update `test_projects_cli.py` and `test_project_spec.py` to test that the new change works.
   - Your tests should test that actions work if they are included, behave as expected when not included, and error out properly when used incorrectly (this is for branch coverage). 
- Do you think you could update the docs to properly reflect the feature you're adding? It'd be great to have your work be documented!

Let me know if you have any questions or need help in implementing this!

mlflow/data.py
juntai-zheng: An alternative to this implementation (instead of having to deal with actions) I can think of is creating a new `type` (maybe called `flag`?) that takes in `True` or `False` (or equivalent values), and appends the flag to the command run, depending on the value of the `flag`.

For example, with an example MLProject of
```
...
parameters:
      data_file: path
      regularization: {type: float, default: 0.1}
      train: {type: flag, default: False}
    command: "python script.py -r  {regularization} {train} {data_file}"
...
```
a call of `mlflow run . data_file=data.csv` would translate to `python script.py --train -r 0.1 data.csv`, just like your proposal from #1945 proposes.

tanmaymathur89:
> Hi @tanmaymathur89, thanks for your work! I left some comments, plus I have a few other remarks:
> 
> * Tests aren't fully covering this feature yet. We'll have to update `test_projects_cli.py` and `test_project_spec.py` to test that the new change works.
>   
>   * Your tests should test that actions work if they are included, behave as expected when not included, and error out properly when used incorrectly (this is for branch coverage).
> * Do you think you could update the docs to properly reflect the feature you're adding? It'd be great to have your work be documented!
> 
> Let me know if you have any questions or need help in implementing this!

Thank you for your commen

tanmaymathur89:
Thank you for your comments @juntai-zheng. I'll address them shortly.

mlflow/data.py
tanmaymathur89: Good point, addressed this in the new commit. Passing an empty string will allow you to toggle the argument

mlflow/projects/_project_spec.py
tanmaymathur89: Changed.

tests/resources/example_project/MLproject
tanmaymathur89: Fair point, moved to greeter

tests/projects/test_entry_point.py
tanmaymathur89: Added more cases. Let me know if these are sufficient

tanmaymathur89:
@juntai-zheng do you have any additional comments?

docs/source/projects.rst
```diff
@@ -307,6 +307,10 @@ uri
     relative paths to absolute paths, as in the ``path`` type. Use this type for programs
     that know how to read from distributed storage (e.g., programs that use Spark).
 
+action
+    An action string. MLflow validates that the parameter is valid. Pass an empty string
+    to skip the argument
```
juntai-zheng:
```suggestion
    to skip the argument.
```

docs/source/projects.rst
```diff
@@ -307,6 +307,10 @@ uri
     relative paths to absolute paths, as in the ``path`` type. Use this type for programs
     that know how to read from distributed storage (e.g., programs that use Spark).
 
+action
+    An action string. MLflow validates that the parameter is valid. Pass an empty string
```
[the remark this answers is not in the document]
juntai-zheng: Would also be useful to explain what a 'valid' parameter is.

tests/projects/test_entry_point.py
```diff
@@ -41,12 +49,15 @@ def test_entry_point_compute_command():
...
+        assert command == "python greeter.py hi friend --goodbye --excitement 10"
+        command = entry_point.compute_command({"name": "friend", "excitement": 10, "args": ''},
+                                              storage_dir)
         assert command == "python greeter.py hi friend --excitement 10"
         with pytest.raises(ExecutionException):
             entry_point.compute_command({}, storage_dir)
         # Test shell escaping
         name_value = "friend; echo 'hi'"
-        command = entry_point.compute_command({"name": name_value}, storage_dir)
+        command = entry_point.compute_command({"name": name_value, "args": ""}, storage_dir)
```
juntai-zheng: Would be good to either add in a case or modify one of the existing tests where `args` has a value in it (e.g. `--goodbye`).

tests/projects/test_entry_point.py
```diff
@@ -127,36 +139,52 @@ def test_params():
...
 
+    user_7 = {"alpha": 0.004, "gamma": 0.89, "args": ""}
+    expected_final_7 = {"alpha": '0.004', "l1_ratio": '0.1', "l2_ratio": '0.0003',
+                        "random_str": "hello", "args": "''"}
+    expected_extra_7 = {"gamma": "0.89"}
+    final_7, extra_7 = entry_point.compute_parameters(user_7, None)
+    assert expected_extra_7 == extra_7
+    assert expected_final_7 == final_7
+
+    user_8 = {"alpha": 0.004, "gamma": 0.89, "args": "-early-stopping"}
```
juntai-zheng: This fails because of the hyphen between `early` and `stopping`. Just wanted to double check, that's the intention of the test, right?

tests/projects/test_entry_point.py
```diff
@@ -127,36 +139,52 @@ def test_params():
...
+    with pytest.raises(ExecutionException):
+        entry_point.compute_parameters(user_8, None)
+
+    user_9 = {"alpha": 0.004, "gamma": 0.89, "args": "early-stopping"}
```
juntai-zheng: This test fails both because there are no hyphens before the argument and because there is a hyphen in the middle of the argument name. Since the test above already checks for failure due to the hyphen being in the middle of the argument name, can we make this change so we're only testing for missing hyphens before the argument?

```suggestion
    user_9 = {"alpha": 0.004, "gamma": 0.89, "args": "earlystopping"}
```

tests/projects/test_project_spec.py
```diff
@@ -13,15 +13,18 @@ def test_project_get_entry_point():
...
     # Validate parameters
-    assert set(entry_point.parameters.keys()) == set(["name", "greeting"])
+    assert set(entry_point.parameters.keys()) == set(["name", "greeting", "args"])
     name_param = entry_point.parameters["name"]
     assert name_param.type == "string"
     assert name_param.default is None
     greeting_param = entry_point.parameters["greeting"]
     assert greeting_param.type == "string"
     assert greeting_param.default == "hi"
+    greeting_param = entry_point.parameters["args"]
```
juntai-zheng:
```suggestion
    action_param = entry_point.parameters["args"]
```

Nit: recommend making this separate from the variable name above.

tests/projects/test_projects_cli.py
```diff
@@ -23,9 +23,14 @@
...
     excitement_arg = 2
     name = "friend"
+    action_args = "--goodbye"
     invoke_cli_runner(cli.run, [TEST_PROJECT_DIR, "-e", "greeter", "-P",
                                 "greeting=hi", "-P", "name=%s" % name,
                                 "-P", "excitement=%s" % excitement_arg])
+    invoke_cli_runner(cli.run, [TEST_PROJECT_DIR, "-e", "greeter", "-P",
+                                "greeting=hi", "-P", "name=%s" % name,
+                                "-P", "excitement=%s" % excitement_arg, "-P",
+                                "args=%s" % action_args])
```
juntai-zheng: Maybe add one more test just like this one with `args=''`?

juntai-zheng (changes requested):
Thanks for your edits! The PR is looking a lot closer to mergability. Left some comments. Do you think you could add a concrete example of how to use the action to the code block here: https://mlflow.org/docs/latest/projects.html#mlproject-file ? It should be in `docs/source/projects.rst`.

Let me know if you have any questions or need help, and thanks again :)

tests/projects/test_entry_point.py
```diff
@@ -20,12 +20,20 @@ def test_entry_point_compute_params():
...
         assert extra_params == {"excitement": "10"}
         # Don't pass extra "excitement" param, pass value for `greeting`
         params, extra_params = entry_point.compute_parameters(
             {"name": "friend", "greeting": "hello"}, storage_dir)
-        assert params == {"name": "friend", "greeting": "hello"}
+        assert params == {"name": "friend", "greeting": "hello", "args": "--goodbye"}
+        # Don't pass extra "excitement" param, pass value for `args`
+        params, extra_params = entry_point.compute_parameters(
+            {"name": "friend", "greeting": "hello", "args": "--goodbye"}, storage_dir)
+        assert params == {"name": "friend", "greeting": "hello", "args": "--goodbye"}
```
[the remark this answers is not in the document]
tanmaymathur89: good point, ill add it!

docs/source/projects.rst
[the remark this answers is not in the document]
tanmaymathur89: Sounds like a good idea.

tests/projects/test_entry_point.py
tanmaymathur89: Makes sense

tests/projects/test_entry_point.py
tanmaymathur89: Thats right!

dbczumar:
@juntai-zheng @tanmaymathur89 I think that adding boolean parameter support sounds great, but I'm wary of adding `actions`, which appears to be relatively Python or bash-specific; @tanmaymathur89 can we refactor this to introduce support for True/False boolean parameters rather than actions?

tanmaymathur89:
> @juntai-zheng @tanmaymathur89 I think that adding boolean parameter support sounds great, but I'm wary of adding `actions`, which appears to be relatively Python or bash-specific; @tanmaymathur89 can we refactor this to introduce support for True/False boolean parameters rather than actions?

Is there a specific advantage of introducing support for boolean parameters instead of using language specific parameter types? I believe we will be adding additional complexity to mlflow instead of making it consistent with whats already out there since mlflow is essentially passing through the values.

PR #2456, opened 2020-02-21
[UI] Better error messages for model registry

aarondav:

Before this PR, attempting to register a model from the UI would result in an opaque "INTERNAL_SERVER_ERROR". This changes it to return the same message as would be returned by calling the API.

image

Note that there are 2 shown because we make 2 API calls from the JS...


mlflow/server/handlers.py
```diff
@@ -62,9 +63,7 @@ def _get_sqlalchemy_store(cls, store_uri, artifact_uri):
 class ModelRegistryStoreRegistryWrapper(ModelRegistryStoreRegistry):
     def __init__(self):
         super(ModelRegistryStoreRegistryWrapper, self).__init__()
-        # Model Registry does not support file based stores
-        self.register('', lambda store_uri: None)
-        self.register('file', lambda store_uri: None)
```
mparkhe: Awesome!

mparkhe (approved):
Neat change!

PR #2459, opened 2020-02-22
Fix handleLayoutChange to update dragmode

harupy:
Fixes a bug introduced in https://github.com/mlflow/mlflow/pull/2408, namely modifying  `handleLayoutChange` in `MetricsPlotPanel` to update `dragmode` of the plot.

Before (`dragmode` can't be changed)
relayout-before

After (`dragmode`can be changed.)
relayout-after

## How is this patch tested?

(Details)

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -221,7 +221,6 @@ export class MetricsPlotPanel extends React.Component {
   handleLayoutChange = (newLayout) => {
     // Unfortunately, we need to parse out the x & y axis range changes from the onLayout event...
     // see https://plot.ly/javascript/plotlyjs-events/#update-data
-    // debugger;
     const {
       "xaxis.range[0]": newXRange0,
       "xaxis.range[1]": newXRange1,
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -51,6 +51,7 @@ export class MetricsPlotPanel extends React.Component {
       // and then back to linear scale (we save the initial negative linear y-axis bounds so
       // that we can restore them when converting from log back to linear scale)
       lastLinearYAxisRange: [],
+      layout: {},
     };
     this.loadMetricHistory(this.props.runUuids, this.state.selectedMetricKeys);
   }
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -260,6 +260,14 @@ export class MetricsPlotPanel extends React.Component {
         yaxis: { autorange: true, type: axisType },
       };
     }
+
+    if (newLayout.dragmode) {
+      mergedLayout = {
+        ...mergedLayout,
+        dragmode: newLayout.dragmode,
+      };
+    }
+
     this.setState({ layout: mergedLayout, lastLinearYAxisRange });
   };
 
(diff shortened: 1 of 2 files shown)


smurching (approved):
Thanks @harupy for the fix, looks good! Do you know if there are any other plotly layout fields that we might inadvertently be excluding here / should make sure to include?

smurching:
Looking at the [Plotly API docs](https://plot.ly/javascript/reference/), it seems there are a number of fields nested under "layout" (fields with parent "layout"). Thus I wonder if the right approach here is to:
* Include `newLayout` in `mergedLayout`
* Exclude/convert fields from `newLayout` that aren't valid, e.g. convert `newLayout['xaxis.range[0]']` and `newLayout['xaxis.range[0]']` into an actual range array, as the current logic already does.

Let me know if that makes sense, thanks!

harupy:
Yes, that makes sense. I will fix the code.

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -220,12 +221,19 @@ export class MetricsPlotPanel extends React.Component {
...
+      "yaxis.range[1]": newYRange1,
+      "xaxis.autorange": xAxisAutorange,
+      "yaxis.autorange": yAxisAutorange,
+      ...restFields
+    } = newLayout;
+
     let mergedLayout = {
       ...this.state.layout,
+      ...restFields,
     };
```
harupy: @smurching 
I hope this is what you meant in https://github.com/mlflow/mlflow/pull/2459#issuecomment-589999741

smurching: @harupy yeah looks good!

smurching:
@harupy thanks for the updates - pushed an update to properly handle changes in the `showspikes` property used to control whether spike lines appear (which unfortunately also must be parsed out of the `newLayout`)

smurching:
Would appreciate your review of my changes, otherwise LGTM!

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -220,46 +221,58 @@ export class MetricsPlotPanel extends React.Component {
...
-          range: [newXRange0, newXRange1],
-        },
-      };
+      newXAxis.range = [newXRange0, newXRange1];
+      newXAxis.autorange = false;
+    }
+    if (xAxisShowSpikes) {
+      newXAxis.showspikes = true;
+    }
+    if (xAxisAutorange === true) {
```
harupy: Is `=== true` necessary?

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -220,46 +221,58 @@ export class MetricsPlotPanel extends React.Component {
...
-    if (newLayout["xaxis.autorange"] === true) {
-      mergedLayout = {
-        ...mergedLayout,
-        xaxis: { autorange: true },
-      };
+    if (yAxisShowSpikes) {
+      newYAxis.showspikes = true;
     }
-    if (newLayout["yaxis.autorange"] === true) {
+    if (yAxisAutorange === true) {
```
harupy: same here

harupy:
@smurching Just small comments otherwise LGTM!

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -220,46 +221,58 @@ export class MetricsPlotPanel extends React.Component {
...
-          range: [newXRange0, newXRange1],
-        },
-      };
+      newXAxis.range = [newXRange0, newXRange1];
+      newXAxis.autorange = false;
+    }
+    if (xAxisShowSpikes) {
+      newXAxis.showspikes = true;
+    }
+    if (xAxisAutorange) {
```
harupy: @smurching Thanks for the fix!

PR #2496, opened 2020-02-26
Optimize docker image size by creating docker image based on the model flavor

### LINKED ISSUE #2426: [FR] Reduce Docker Image Size for Predict
### Motivation
Docker Image produced by MLFlow for predicting the models is supposed to be the most portable format but unfortunately, the docker image produced by MLFlow is of huge size. We tried with a simple model file with 1kb size (a simple regression model) but the resultant docker image is 3.3 GB.  We are trying to deploy these docker images (about 60) in Kubernetes but since K8s nodes will have limited resources, docker images of this size make mlflow a non-viable option. 
### Proposed Changes
Please let us know how the docker image size for predicting the model can be reduced.

ambiSanthu:
Fixes #2426 : Optimize docker image size

## How is this patch tested?

Created a docker image using below command:
mlflow models build-docker -m model -n docker_image_name --install-mlflow --no-conda

## Release Notes
With this new feature, docker image is created either for python flavor or java flavor based on the model for which docker image is being created for. 

Also added a new arguments to build-docker ``--no-conda`` that doesn't create a new conda environment "custom_env" instead uses the default miniconda base environment. 

The current docker image size for sklearn iris example is 3.21GB and with the new changes docker image size is reduced to 2.14GB and by using ``--no-conda`` option (ie., using base miniconda environment), docker image size can be further reduced to 887MB.

- [X] Yes. Reduces the docker image size.

- [ ] UI
- [X] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [X] Java
- [X] Python


QuentinAmbard:
Looking good, however I think we should take the opportunity to take this a bit further.

Here are some ideas, let me know what you think:

- For java serving flavor, we could create a builder image with maven to serve the compiled jar and avoid having maven installed in the final image. 
- Maybe implement the same logic for conda: create a builder to download and install conda, and then just copy conda from this builder (something like this: https://github.com/databricks/containers/blob/master/ubuntu/python/Dockerfile)
- We could also automatically get the serving flavor from the ML model to know if we have to ship java or conda with the build (instead of relying on a manual parameter like "python-only")
- There are a few `print(opt_image)`we should remove too 

What's the value of `--no-conda`? If I'm correct it still create the default miniconda env but uses pip to install the dependencies, is it to reduce the image size? Do you have some number with and without this option?

ambiSanthu:
Please find my comments below.

> * For java serving flavor, we could create a builder image with maven to serve the compiled jar and avoid having maven installed in the final image.

I hadn't focused on java flavour. As suggested below for conda "copying conda from builder", I can have a java builder that does maven build and copy only required jars for java deployment. I will implement this feature.
> * Maybe implement the same logic for conda: create a builder to download and install conda, and then just copy conda from this builder (something like this: 
https://github.com/databricks/containers/blob/master/ubuntu/python/Dockerfile)

I will try this option. Looks good.
> * We could also automatically get the serving flavor from the ML model to know if we have to ship java or conda with the build (instead of relying on a manual parameter like "python-only")

"python-only" flag is passed only for build-docker command. mlflow serve doesn't need this flag as it takes from the flavor from the docker image (actually from MLmodel file)
> * There are a few `print(opt_image)`we should remove too

I will remove these print statements.

I will implement your suggestions and update the PR.

QuentinAmbard:
For the "python-only" what I meant is that we could get the **serving** flavor of the model in the backend.py and automatically define if we need conda or java when we build the final image, instead of relying on a manual flag.

mlflow/models/container/__init__.py
```diff
@@ -92,22 +93,53 @@ def _install_pyfunc_deps(model_path=None, install_mlflow=False):
...
+                            if isinstance(lib, dict):
+                                for i in lib["pip"]:
+                                    if i not in exclude:
+                                        req.write(i+'\n')
+                            else:
+                                if lib.split('=')[0] not in exclude:
+                                    req.write(lib.replace("=", "==")+'\n')
+
+                    req.close()
+                env_file.close()
```
AlexMooney: Seems like this whole chunk should be extracted to a method.

```python
def _write_requirements_from_env(base_path="/opt/mlflow"):
    with open(os.path.join(base_path, conf[pyfunc.ENV])) as env_file:
        env_yaml = yaml.full_load(env_file)
        dependencies = env_yaml["dependencies"]
        exclude = ["python", "mlflow"]
        with open(os.path.join(base_path, "requirements.txt"), "w") as req_file:
            for dependency in dependencies:
                if isinstance(dependency, dict):
                    req_file.write(
                        "\n".join(d for d in dependency["pip"] if d not in exclude)
                        + "\n"
                    )
                else:
                    if dependency.split("=")[0] not in exclude:
                        req_file.write(dependency.replace("=", "==") + "\n")
```

Also, no need to manually `close` the files when using `with`.

mlflow/models/container/__init__.py
```diff
@@ -121,10 +153,18 @@ def _install_pyfunc_deps(model_path=None, install_mlflow=False):
 def _serve_pyfunc(model):
     conf = model.flavors[pyfunc.FLAVOR_NAME]
     bash_cmds = []
-    if pyfunc.ENV in conf:
+    if os.environ.get('OPTIMIZED_IMAGE'):
+        opt_image = os.environ.get('OPTIMIZED_IMAGE')
+    else:
+        opt_image = "false"
+    print(opt_image)
```
AlexMooney:
```suggestion
    opt_image = os.environ.get('OPTIMIZED_IMAGE') or "false"
```

mlflow/models/docker_utils.py
```diff
@@ -43,42 +43,77 @@
...
+        else:
+            return (
+                "COPY {mlflow_dir} /opt/mlflow\n"
+                "RUN pip install /opt/mlflow\n"
+                "RUN cd /opt/mlflow/mlflow/java/scoring && "
+                "mvn --batch-mode package -DskipTests && "
+                "mkdir -p /opt/java/jars && "
+                "mv /opt/mlflow/mlflow/java/scoring/target/"
+                "mlflow-scoring-*-with-dependencies.jar /opt/java/jars\n"
+                ).format(mlflow_dir=mlflow_dir)
```
AlexMooney:
```suggestion
        install_string = (
            "COPY {mlflow_dir} /opt/mlflow\n"
            "RUN pip install /opt/mlflow\n"
        )
        if not python_only:
            install_string += (
                "RUN cd /opt/mlflow/mlflow/java/scoring && "
                "mvn --batch-mode package -DskipTests && "
                "mkdir -p /opt/java/jars && "
                "mv /opt/mlflow/mlflow/java/scoring/target/"
                "mlflow-scoring-*-with-dependencies.jar /opt/java/jars\n"
                )
        return install_string.format(mlflow_dir=mlflow_dir)
```

mlflow/models/docker_utils.py
```diff
@@ -43,42 +43,77 @@
...
-            " -Dartifact=org.mlflow:mlflow-scoring:{version}:jar"
-            " -DoutputDirectory=/opt/java/jars\n"
-            "RUN cp /opt/java/mlflow-scoring-{version}.pom /opt/java/pom.xml\n"
-            "RUN cd /opt/java && mvn "
-            "--batch-mode dependency:copy-dependencies -DoutputDirectory=/opt/java/jars\n"
-        ).format(version=mlflow.version.VERSION)
-
-
-def _build_image(image_name, entrypoint, mlflow_home=None, custom_setup_steps_hook=None):
+        if python_only:
```
AlexMooney: Similar to above, build the string, add to it `if not python_only`, and then format and return.

mlflow/models/docker_utils.py
```diff
@@ -91,16 +126,24 @@ def _build_image(image_name, entrypoint, mlflow_home=None, custom_setup_steps_ho
...
+            with open(os.path.join(cwd, "Dockerfile"), "w") as f:
+                f.write(_DOCKERFILE_PYTHON_TEMPLATE.format(
+                    install_mlflow=install_mlflow, custom_setup_steps=custom_setup_steps,
+                    entrypoint=entrypoint))
+        else:
+            with open(os.path.join(cwd, "Dockerfile"), "w") as f:
+                f.write(_DOCKERFILE_TEMPLATE.format(
+                    install_mlflow=install_mlflow, custom_setup_steps=custom_setup_steps,
+                    entrypoint=entrypoint))
```
AlexMooney: No need for the duplication.

```suggestion
        dockerfile_template = _DOCKERFILE_PYTHON_TEMPLATE if python_only else _DOCKERFILE_TEMPLATE
        with open(os.path.join(cwd, "Dockerfile"), "w") as f:
            f.write(
                dockerfile_template.format(
                    install_mlflow=install_mlflow,
                    custom_setup_steps=custom_setup_steps,
                    entrypoint=entrypoint,
                )
            )
```

AlexMooney (commented):
Really excited for a smaller docker container!  Looks like this PR can be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)ed out quite a bit.

ambiSanthu:
> For the "python-only" what I meant is that we could get the **serving** flavor of the model in the backend.py and automatically define if we need conda or java when we build the final image, instead of relying on a manual flag.

As suggested, got the flavor from MLmodel file and now docker image will be created based on the flavor defined in MLmodel file. Removed "python-only" option.

mlflow/models/docker_utils.py
ambiSanthu: Thank you for the feedback. Made the suggested code changes and testing is in progress.

ambiSanthu: Testing completed.

ambiSanthu:
@AlexMooney I have made the suggested code changes. Please review.

AlexMooney (approved):
Nice!  I'm not at all experienced with this codebase, but this is looks good to me.

ambiSanthu:
@mparkhe Please review the code changes.

harupy:
The following files should be removed.

- t10k-images-idx3-ubyte.gz
- t10k-labels-idx1-ubyte.gz
- train-images-idx3-ubyte.gz
- train-labels-idx1-ubyte.gz

ambiSanthu:
> The following unrelated files should be removed.
> 
> * t10k-images-idx3-ubyte.gz
> * t10k-labels-idx1-ubyte.gz
> * train-images-idx3-ubyte.gz
> * train-labels-idx1-ubyte.gz

@harupy Removed the files. But all the tests are failing with some generic error. Hence pulled in the fix from  Pull Request #2567.

sanketrk:
Can someone with write access review & approve? Everything looks good here.

sanketrk (approved):
We tested this and is working fine

smurching (commented):
@ambiSanthu thanks for the PR! & thanks to all the reviewers for chiming in here :)

As discussed in #2426, my primary feedback on this is:

> @ambiSanthu the approach of adding dependencies based on the model's flavor makes sense. I wonder if instead of adding a --no-conda option, it would make sense to simply remove the base miniconda image to save space (not sure how many models can be served solely using the base miniconda image). Thoughts?

ambiSanthu:
> @ambiSanthu thanks for the PR! & thanks to all the reviewers for chiming in here :)
> 
> As discussed in #2426, my primary feedback on this is:
> 
> > @ambiSanthu the approach of adding dependencies based on the model's flavor makes sense. I wonder if instead of adding a --no-conda option, it would make sense to simply remove the base miniconda image to save space (not sure how many models can be served solely using the base miniconda image). Thoughts?

As per the docker image creation flow, a new conda env called "custom_env" is being created and dependencies from conda.yml is installed in this custom_env. Instead with `--no-conda` option, I am not creating the custom_env but parsing the conda.yml and installing the dependencies directly into base env.  Hence I make sure required dependencies are installed for the model to run.

smurching:
@ambiSanthu apologies for the delay, the behavior in the `--no-conda` case makes sense - could we just do that by default (i.e. run `conda env update -f model-environment.yml` to update the base env rather than create a new env named `custom_env`), rather than introduce a `--no-conda` option?

I suppose the main thing we'd need to worry about would be unintended behavior changes, but I think we could test that e.g. restoring an env with a different python version from the base environment etc continues to work, and using `conda env update -f` to update an existing environment from the contents of a conda.yml file seems pretty standard: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#updating-an-environment

mlflow/models/container/__init__.py
```diff
@@ -76,13 +76,34 @@ def _serve():
...
 
 
-def _install_pyfunc_deps(model_path=None, install_mlflow=False):
+def _install_pyfunc_deps(model_path=None, install_mlflow=False, no_conda=False):
     """
     Creates a conda env for serving the model at the specified path and installs almost all serving
     dependencies into the environment - MLflow is not installed as it's not available via conda.
     """
     # If model is a pyfunc model, create its conda env (even if it also has mleap flavor)
+    def install_dependency_from_condafile(env_path_dst):
```
smurching, line 85: Could we achieve this by simply executing `["conda env update -f {env_path_dist}".format(env_path_dist=env_path_dist)]` rather than attempting to parse the conda env into a requirements.txt & installing that?

smurching:
Hi @ambiSanthu, just wanted to check in on this - did you have any questions on the latest comments/were you interested in continuing with the PR? Thanks!

mlflow/models/container/__init__.py
ambiSanthu: Hi @smurching , Thank you for the feedback. Please find my thoughts below.

Usually conda.yaml file has a name field and in that case when we run "conda env update" command, it usually creates a new conda environment if it doesn't exists. Again it adds up more space and the docker image size would increase. Hence I feel `conda env update` will not work in this case.

However I can try `conda env update -n base` but not sure how it would work and i afraid if name field in conda.yaml would override the env name. I will do some experiments on this and update.

ambiSanthu:
> Hi @ambiSanthu, just wanted to check in on this - did you have any questions on the latest comments/were you interested in continuing with the PR? Thanks!

Sorry for the delayed response. Please find my comments above.

mlflow/models/container/__init__.py
smurching: Hi @ambiSanthu, yep exactly, the idea is to run `conda env update` while the `base` environment is activated - I'd tried this on the anaconda docker image while the base image was activated and found it to work, let me know if you hit any issues!

no-blockers: If conda is a must, Can we use miniconda instead of anaconda?

no-blockers:
May I know what's the status of this PR? I would love to contribute if facing lack of resources.

mlflow/models/container/__init__.py
ambiSanthu: > Hi @ambiSanthu, yep exactly, the idea is to run `conda env update` while the `base` environment is activated - I'd tried this on the [anaconda docker image](https://hub.docker.com/r/continuumio/anaconda/) while the base image was activated and found it to work, let me know if you hit any issues!

But if conda.yaml file have "name" field then it creates a new conda environment with the name mentioned in conda.yaml file. Again it might not be an optimized image. In such cases, we might need to first remove the line that contains "name" in conda.yaml which I feel an ugly fix. Please share your thoughts on this @smurching

ambiSanthu: > If conda is a must, Can we use miniconda instead of anaconda?

We are already using miniconda in docker templates.

ambiSanthu: @smurching please let me know your thoughts on this.

> > Hi @ambiSanthu, yep exactly, the idea is to run `conda env update` while the `base` environment is activated - I'd tried this on the [anaconda docker image](https://hub.docker.com/r/continuumio/anaconda/) while the base image was activated and found it to work, let me know if you hit any issues!
> 
> But if conda.yaml file have "name" field then it creates a new conda environment with the name mentioned in conda.yaml file (even though "-n base" is mentioned). Again it might not be an optimized image. In such cases, we might need to first remove the line that contains "name" in conda.yaml which I feel an ugly fix. Please share your thoughts on this @smurching

bhushanbrb:
thanks @ambiSanthu , Hi All , any updates on this !!!

pocman:
Hi, do we have any updates on that topic ?

PR #2508, opened 2020-02-29
[UI] Truncate run id in the run name when comparing multiple runs

harupy:

This PR truncates run id in the run name when comparing multiple runs in order to:
- Prevent the legends from narrowing the plot area
- Make the popover compact

Before
Screen Shot 2020-02-29 at 12 38 01

After
Screen Shot 2020-02-29 at 12 36 51

## How is this patch tested?

Manually tested

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


mlflow/server/js/src/utils/Utils.js
```diff
@@ -290,7 +290,7 @@ class Utils {
    * @param runTags Object of tag name to MlflowMessages.RunTag instance
    */
   static getRunDisplayName(runTags, runUuid) {
-    return Utils.getRunName(runTags) || "Run " + runUuid;
+    return Utils.getRunName(runTags) || "Run " + runUuid.slice(0, 6);
```
mparkhe: Does it make sense to show trailing "..." so this truncation is not confusing.
Thoughts on this: @gioa and @smurching

gioa: Are the shorten run IDs distinct enough? I am worried they show up the same for different runs. 

Agree with @mparkhe, maybe we should truncate it with text-overflow: ellipsis. I also expect users hover can see the full string, or do a triple-click can copy the whole thing.

harupy: @mparkhe.
Agree on showing trailing `...`.

@gioa 
> Are the shorten run IDs distinct enough?

I think they are.

> I also expect users hover can see the full string, or do a triple-click can copy the whole thing.

What is the use case of this feature?

gioa: Doesn't the user need the full run IDs to refer to in the API calls? So they may copy it right there? I might be wrong though. 

Also when I see a trailing ..., I would expect to be able to copy the full string, just a pattern I get used to in other places.

harupy: How about putting a button on the popover that copies the full run id to the user's clipboard?

harupy: Example:

copy-button

gioa: That works too! Might be a little overkill. Could add a run ID in the top info section of a run detail page and support copy there.

harupy: @Zangr Any thoughts?

harupy:
Added trailing `...`.

Screen Shot 2020-03-03 at 23 22 54

mlflow/server/js/src/utils/Utils.js
harupy: Tried a copy button on the popover.

overkill

gioa: What do you think? It seems too much going on in this popup. I am okay without it. 

Btw I was talking about adding this in the top info section area. See a mock here:
MLFlow_RunDetail (1)

harupy: Yes, it's too much and confusing. The mock looks great!

mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -521,7 +521,10 @@ const mapStateToProps = (state, ownProps) => {
   // Flat array of all metrics, with history and information of the run it belongs to
   // This is used for underlying MetricsPlotView & predicting chartType for MetricsPlotControls
   const metricsWithRunInfoAndHistory = _.flatMap(runUuids, (runUuid) => {
-    const runDisplayName = Utils.getRunDisplayName(getRunTags(runUuid, state), runUuid);
+    const runDisplayName = Utils.getRunDisplayName(
+      getRunTags(runUuid, state),
+      runUuid.slice(0, 6) + '...'
```
Zangr: What do you think about setting the run name wrapper to `text-overflow: ellipsis`? That way, we won't lose any information in the DOM.
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

harupy: @Zangr Thanks for the comment! If we use `text-overflow`, do we need to set `max-width`?

aarondav:
I like the change - what's the current state of this PR?

mlflow/server/js/src/components/MetricsPlotPanel.js
harupy: @Zangr It seems this can't be applied to the legends.

aarondav:
@Zangr Mind reviewing based on @harupy's latest update?

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
```diff
@@ -523,7 +523,10 @@ const mapStateToProps = (state, ownProps) => {
   // Flat array of all metrics, with history and information of the run it belongs to
   // This is used for underlying MetricsPlotView & predicting chartType for MetricsPlotControls
   const metricsWithRunInfoAndHistory = _.flatMap(runUuids, (runUuid) => {
-    const runDisplayName = Utils.getRunDisplayName(getRunTags(runUuid, state), runUuid);
+    const runDisplayName = Utils.getRunDisplayName(
+      getRunTags(runUuid, state),
+      runUuid.slice(0, 6) + '...'
```
Zangr, line 528: We were using `Utils.truncateString(runDisplayName, MAX_RUN_NAME_DISPLAY_LENGTH)`, could you please check if you can achieve this by use and decrease MAX_RUN_NAME_DISPLAY_LENGTH?

aarondav:
@harupy Mind addressing Richard's feedback?

harupy:
Decreasing `MAX_RUN_NAME_DISPLAY_LENGTH` seems to work. Thanks @Zangr!

Before (MAX_RUN_NAME_DISPLAY_LENGTH = 36):

Screen Shot 2020-04-26 at 17 07 18

After (MAX_RUN_NAME_DISPLAY_LENGTH = 24):

Screen Shot 2020-04-26 at 17 06 41

Zangr (approved):
LGTM

PR #2510, opened 2020-02-29
[UI] Fix bug for RunLinksPopover

harupy:

The run links popover unintentionally shows up on relayout by double click (please take a look at the attached gif). This PR fixes it.

demo

## How is this patch tested?

Manually tested

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -260,6 +260,7 @@ export class MetricsPlotPanel extends React.Component {
    * and schema.
    */
   handleLayoutChange = (newLayout) => {
+    this.isClicked = false;
     const state = this.getUrlState();
     // Unfortunately, we need to parse out the x & y axis range changes from the onLayout event...
     // see https://plot.ly/javascript/plotlyjs-events/#update-data
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -71,6 +71,7 @@ export class MetricsPlotPanel extends React.Component {
       popoverY: 0,
       popoverRunItems: [],
     };
+    this.isClicked = false;
     this.loadMetricHistory(this.props.runUuids, this.getUrlState().selectedMetricKeys);
   }
 
--- a/mlflow/server/js/src/components/MetricsPlotPanel.js
+++ b/mlflow/server/js/src/components/MetricsPlotPanel.js
@@ -71,7 +71,7 @@ export class MetricsPlotPanel extends React.Component {
       popoverY: 0,
       popoverRunItems: [],
     };
-    this.isClicked = false;
+    this.isPlotClicked = false;
     this.loadMetricHistory(this.props.runUuids, this.getUrlState().selectedMetricKeys);
   }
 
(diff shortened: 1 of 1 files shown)


mlflow/server/js/src/components/MetricsPlotPanel.js
```diff
@@ -260,6 +260,7 @@ export class MetricsPlotPanel extends React.Component {
    * and schema.
    */
   handleLayoutChange = (newLayout) => {
+    this.isClicked = false;
```
harupy: I haven't yet figured out what causes this bug, but setting `isClicked` to `false` in `handleLayoutChange` fixes the bug.

smurching (commented):
@harupy thanks for the fix! After some investigation, this particular fix seems to depend on Plotly invoking the `handleLayoutChange` callback before the `updatePopover` callback - in practice, this seems to always be the case, but it'd be good to think if there's a more robust solution that would work if the callback ordering were reversed.

In particular, it seems the following occurs (after putting `debugger`s in `handleLayoutChange` and `updatePopover`) when the plot is double-clicked to reset the layout
* The plot range is reset (i.e. Plotly resets its own plot range - this change is visible in the UI)
* `updatePopover` is called. At this point, the plot range has already updated, so it seems Plotly mistakenly believed a point was clicked & has hence called `updatePopover`. We set `this.isClicked = !this.isClicked`, which in this case sets `this.isClicked` to `true`, then trigger a timeout that will update the popover state accordingly in 300ms
* `handleLayoutChange` is called and sets `this.isClicked` to false
* The timeout created in `updatePopover` runs, observes that `this.isClicked` is false, and closes the popover

smurching:
If a more robust solution isn't possible, I don't think it needs to block this PR, would just be good to spend some time investigating

harupy:
@smurching Thanks for the investigation!

I checked when plotly calls `onClick`.

- When the plot is double-clicked, `onClick` is called twice.
- The first `onClick` is called **before** the relayout.
- The second `onClick` is called **after** the relayout.
- `onClick` is called only when a point is selected.

If the first click is performed with a point selected, and the second one without, `onClick` is called only once,  which sets `isClicked` to `true`, then the popover shows up.

harupy:
Summarized when the popover shows up.

| point selected (1st click) | point selected (2nd click) | popover shows up? |
| :------------------------- | :------------------------- | :---------------- |
| yes                        | no                         | yes               |
| no                         | yes                        | yes               |
| yes                        | yes                        | no                |
| no                         | no                         | no                |

harupy:
I also investigated the execution order of `onClick`, `onDoubleClick`, and `onRelayout`. The second `onClick` is called after `onDoubleClick`. To prevent the popover from showing up, we need to set `isClicked` in `onRelayout` (which is `handleLayoutChange`).

order-callbacks

smurching (commented):
Got it, yeah sounds good to merge this fix as is - maybe we can rename `isPlotClicked` to `displayPopover` for clarity? Otherwise LGTM

harupy:
`displayPopover` sounds better. Thanks!

smurching (approved):
LGTM

harupy:
@smurching tests passed

PR #2514, opened 2020-03-01
Add a new API GetLatestProdModel

tongtong1990:

An new API is added to pull artifact for GetLatestProdModel. I am from Pinterest ML Platform team and we are using MLFlow as our model hub. One gap we need to close is to integrate with different deployment system, and now we use a pull mode so that we need an API to expose latest prod version for a model.

## How is this patch tested?

This PR is only for proto change. Implementation PR will come later in a separate one and will be tested by unit tests.

- [ ] UI
- [ ] CLI
- [ x] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/java/client/src/main/java/org/mlflow/api/proto/ModelRegistry.java
+++ b/mlflow/java/client/src/main/java/org/mlflow/api/proto/ModelRegistry.java
@@ -26017,6 +26017,1527 @@ public final class ModelRegistry {
 
   }
 
+  public interface GetLatestProdModelOrBuilder extends
+      // @@protoc_insertion_point(interface_extends:mlflow.GetLatestProdModel)
+      com.google.protobuf.MessageOrBuilder {
+
+    /**
+     * <pre>
+     * Name of the registered model
+     * </pre>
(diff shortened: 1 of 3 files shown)


mparkhe:
`mlflow/registered-models/get-latest-versions` is designed to solve this exact problem. It takes `stages` as argument--you can supply `["Production"]` for your use case. 

Another approach is to use the models URI path in load_model API. Here's an [example of this API in sklearn model flavor](https://mlflow.org/docs/latest/python_api/mlflow.sklearn.html#mlflow.sklearn.load_model):

```
model = mlflow.sklearn.load_model("models:/user_personalization/Production")
```

mparkhe:
Closing this out, since there are [two other mechanism](https://github.com/mlflow/mlflow/pull/2514#issuecomment-593055439) to achieve this.

PR #2556, opened 2020-03-06
Move_tracking_uri mock fixture into conftest.py

jdlesage:

conftest.py is the way in pytest to share fixtures. This PR moves the most used fixture `tracking_uri_mock` into this confest.py

It also resolves lot of lint warning (unused-arguments/imports)

## How is this patch tested?
Run tests

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,9 @@
+import os
+
 import pytest
+
+import mlflow
+from mlflow.utils.file_utils import path_to_local_sqlite_uri
 
 
 @pytest.fixture
@@ -14,3 +19,12 @@ def reset_mock():
     for obj, attr, value in cache:
         setattr(obj, attr, value)
     cache[:] = []
+
+
+@pytest.fixture()
+def tracking_uri_mock(tmpdir):
+    try:
+        mlflow.set_tracking_uri(path_to_local_sqlite_uri(os.path.join(tmpdir.strpath, 'mlruns')))
+        yield tmpdir
+    finally:
+        mlflow.set_tracking_uri(None)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,10 +21,12 @@ def reset_mock():
     cache[:] = []
 
 
-@pytest.fixture()
-def tracking_uri_mock(tmpdir):
+@pytest.fixture(autouse=True)
+def tracking_uri_mock(tmpdir, request):
     try:
-        mlflow.set_tracking_uri(path_to_local_sqlite_uri(os.path.join(tmpdir.strpath, 'mlruns')))
+        if 'notrackingurimock' not in request.keywords:
+            mlflow.set_tracking_uri(path_to_local_sqlite_uri(
+                os.path.join(tmpdir.strpath, 'mlruns')))
         yield tmpdir
     finally:
         mlflow.set_tracking_uri(None)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,8 +25,12 @@ def reset_mock():
 def tracking_uri_mock(tmpdir, request):
     try:
         if 'notrackingurimock' not in request.keywords:
-            mlflow.set_tracking_uri(path_to_local_sqlite_uri(
-                os.path.join(tmpdir.strpath, 'mlruns')))
+            tracking_uri = path_to_local_sqlite_uri(
+                os.path.join(tmpdir.strpath, 'mlruns'))
+            mlflow.set_tracking_uri(tracking_uri)
+            os.environ["MLFLOW_TRACKING_URI"] = tracking_uri
         yield tmpdir
     finally:
         mlflow.set_tracking_uri(None)
+        if 'notrackingurimock' not in request.keywords:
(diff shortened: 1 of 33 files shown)


smurching (commented):
@jdlesage awesome thank you for splitting this out! I wonder if we could simply set up `tracking_uri_mock` to be an `autouse`'d fixture, and then avoid having to even reference it in most tests? See https://docs.pytest.org/en/latest/fixture.html#order-higher-scoped-fixtures-are-instantiated-first for an example of declaring an `autouse` fixture with function scope (which is what I think we'd want in this case)

jdlesage:
Yes, autoclose will remove lot of useless code. My concern is if I add autoclose, then all the tests will use it (because the fixture is defined at the top-level of the project). It means: even the integration/release tests that should use a real tracking server will instantiate the fixture...
I agree to add the autoclose if I can disable it on some folders or python files. I will have a look into the documentation to check if this feature exists.

smurching:
@jdlesage good question, and good point - thankfully it seems `autouse` fixtures are instantiated before non-autouse fixtures at the same scope level (from [the pytest docs](https://docs.pytest.org/en/latest/fixture.html#order-higher-scoped-fixtures-are-instantiated-first)).

Thus in cases where we intend to use a real tracking server (e.g. in [this integration test suite](https://github.com/mlflow/mlflow/blob/889e46269256c3bef5c9744312fd860c0136521a/tests/tracking/test_rest_tracking.py#L95) that verifies Python client behavior against the different OSS REST API backends), I'd hope that `tracking_uri_mock` would be instantiated first, and then the more-specific tracking-uri mock would be instantiated & take precedence over the global `tracking_uri_mock`

jdlesage:
Some tests failed because of the mock. As far I understand the failures, some tests don't use the fluent api and instantiate a singleton with the client (such as tests on xgboost, lightgbm). Client is already created when the fixture is executed.. 
Other tests make the assumption it is a file store and the mock uses a sqllite backend.

Therefore, I added a way to disable the fixture based on the answer of this question: https://stackoverflow.com/questions/38748257/disable-autouse-fixtures-on-specific-pytest-marks
I updated CONTRIBUTING.rst to explain that.

jdlesage:
Hopefully https://github.com/mlflow/mlflow/pull/2567 will fix the tests that are broken due to a new version of click

jdlesage:
@smurching There is 1 test failing during some installs by pip (if I understand correctly the call stack). I don't think it is related to this change. Could we ignore this error ? Thank you!

smurching (approved):
@jdlesage apologies for the delay! Just pushed a commit that I'm hoping will remove the need for many usages of `notrackingurimock`, will leave some explanatory comments - otherwise LGTM!

tests/gluon_autolog/test_gluon_autolog.py
```diff
@@ -15,8 +15,6 @@
 import mlflow
 import mlflow.gluon
 
-client = mlflow.tracking.MlflowClient()
```
smurching: The main change needed for all these per-framework tests was to compute the client within the test, after the tracking URI mock was already instantiated, instead of globally

tests/conftest.py
```diff
@@ -25,8 +25,12 @@ def set_mock(obj, attr, mock):
 def tracking_uri_mock(tmpdir, request):
     try:
         if 'notrackingurimock' not in request.keywords:
-            mlflow.set_tracking_uri(path_to_local_sqlite_uri(
-                os.path.join(tmpdir.strpath, 'mlruns')))
+            tracking_uri = path_to_local_sqlite_uri(
+                os.path.join(tmpdir.strpath, 'mlruns'))
+            mlflow.set_tracking_uri(tracking_uri)
+            os.environ["MLFLOW_TRACKING_URI"] = tracking_uri
```
smurching: This is what was causing tests that launch subprocesses to fail (e.g. the CLI tests that we marked with `pytest.mark.notrackingurimock`). The specific problem is that we'd set the tracking URI, but not set the MLFLOW_TRACKING_URI environment variable, so subprocesses would not run with our mock tracking URI (& instead would actually log to whatever the user's default tracking URI was, usually ./mlruns within the current directory).

tests/tracking/_model_registry/test_utils.py
```diff
@@ -10,6 +10,11 @@
 from mlflow.tracking._model_registry.utils import _get_store
 from mlflow.tracking._tracking_service.utils import _TRACKING_URI_ENV_VAR
 
+# Disable mocking tracking URI here, as we want to test setting the tracking URI via
```
smurching: Just added an explanatory comment for cases where we do want to use `notrackingurimock`

smurching:
Thanks @jdlesage! Merged

Issue #2560, opened 2020-03-07
[PROPOSAL] Standardize code style for UI using prettier


harupy:
### Describe the proposal

Introduce [Prettier](https://prettier.io/) to standardize code style for the UI.

### Motivation

Prettier allows everyone to format code in the same way and helps improve development productivity. We can avoid wasted efforts during code review.

### Proposed Changes

I'm planning to follow the steps proposed in https://github.com/kubeflow/pipelines/issues/2406.

- [ ] Make a reasonable prettier config that's consistent with existing code style.
- [ ] Transform all code with prettier after syncing with people having open PRs.
- [ ] Set up CI check.
- [ ] Update CONTRIBUTING doc with related information.

smurching:
Thanks @harupy! Some additional context from offline discussion: `prettier` (when applied via the `prettier-eslint` package) is capable of applying the same rules as `eslint`, & basically can fix a superset of the linter errors that can be fixed via `eslint --fix`, as per discussion in https://github.com/prettier/prettier-eslint/issues/101

harupy:
@smurching Thanks! It seems [prettier/eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) can do the same thing.

harupy:
I tried the instruction below which seems to work.

https://prettier.io/docs/en/integrating-with-linters.html#eslint

harupy:
Should we foramt css files (`prettier-eslint` and `prettier-plugin-eslint` don't format css)? 

If we do, we can use the commands below.
```
# format
prettier 'src/**/*.css'

# check
prettier --check 'src/**/*.css'
```

harupy:
Resolved by #2561

PR #2561, opened 2020-03-07
[UI] Introduce prettier

harupy:

- Part of #2560.
- Install `prettier`, `eslint-config-prettier`, and `eslint-plugin-prettier` ([ref](https://prettier.io/docs/en/integrating-with-linters.html)).
- Add prettier config and commands for lint check and fix.

## How is this patch tested?

(Details)

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


mlflow/server/js/.prettierrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  printWidth: 100,
+  trailingComma: 'es5',
+  tabWidth: 2,
+  semi: true,
+  singleQuote: true,
+  arrowParens: 'always',
```
harupy: |Rule|Description|
|:-|:-|
|`printWidth`|Specify the line length that the printer will wrap on.|
|`trailingComma`|Print trailing commas wherever possible when multi-line|
|`tabWidth`|Specify the number of spaces per indentation-level.|
|`singleQuote`|Use single quotes instead of double quotes.|
|`arrowParens`|Include parentheses around a sole arrow function parameter.|

https://prettier.io/docs/en/options.html#tab-width

harupy:
@smurching @Zangr 

#2638 is merged. Should I wait for other PRs related to UI to be merged?

aarondav:
Looks like the config is working and there's a lot of places to update!

harupy:
@aarondav Thanks for the comment. If the current config is ok, I'll run `eslint --fix .` to apply the fixes.

@smurching @Zangr Any thoughts on the current prettier config?

aarondav:
@harupy Would you mind running the fix so we can see what the output of the options looks like? If some have weird effects, we can discuss them!

Zangr:
Thank you @harupy, current diffs look good. Could you please make sure we include all the rules specified in [tinyurl/mlflowjs](https://tinyurl.com/mlflowjs)?

harupy:
@Zangr Sure. I'll update the config. I have a question.

harupy:
@Zangr I have a question. Do you think we should format non-JS files (e.g. html, css, json) using the command below.


```
prettier --write 'src/**/*.{html,json,css}
```

Zangr:
@harupy let's keep this PR focusing on JS lint only, which makes reviewing a bit easier. We can consider other types when we find it necessary.

harupy:
@Zangr Updated the config to include the rules in tinyurl/mlflowjs.

harupy:
The following ESlint errors need to be fixed (or disabled).
- no-confusing-arrow
- indent-legacy
- max-len
- no-mixed-operators

https://travis-ci.org/github/mlflow/mlflow/jobs/676785830#L490

Zangr (approved):
@harupy the set of changes generated looks pretty good. Would you mind rerun it on the latest master?

harupy:
@Zangr synced with the master and reran `npm run lint:fix` again.

mlflow/server/js/.eslintrc.js
```diff
@@ -146,7 +148,7 @@ module.exports = {
     'no-catch-shadow': "off",
     'no-class-assign': 2,
     'no-cond-assign': 2,
-    'no-confusing-arrow': [2, { 'allowParens': true }],
+    'no-confusing-arrow': 0,
```
harupy, line 151: @Zangr I disabled some ESlint rules that conflict with prettier. Are you fine with these changes?

Zangr: Could you please comment on each of the disable eslintrc rules with the conflicting prettier rule?

mlflow/server/js/.eslintrc.js
```diff
@@ -359,7 +361,7 @@ module.exports = {
     'valid-typeof': 2,
     'vars-on-top': 0,
     'wrap-iife': 2,
-    'wrap-regex': 2,
+    'wrap-regex': 0,
```
harupy, line 364: Prettier removes `()` here, which conflicts with [wrap-regex](https://eslint.org/docs/rules/wrap-regex).

Screen Shot 2020-04-26 at 17 32 13

mlflow/server/js/.eslintrc.js
harupy: Got it. Prettier removes () here, which conflicts with [no-confusing-arrow](https://eslint.org/docs/rules/no-confusing-arrow).

Screen Shot 2020-04-26 at 17 37 10

mlflow/server/js/.eslintrc.js
```diff
@@ -122,10 +123,11 @@ module.exports = {
     'max-depth': [2, 4],
     'max-len': ['error', 100, 2, {
       'ignoreUrls': false,
-      'ignoreComments': false,
+      'ignoreComments': true,
       'ignoreRegExpLiterals': true,
       'ignoreStrings': false,
       'ignoreTemplateLiterals': false,
+      'ignorePattern': "^import\\s.+\\sfrom\\s'.+';",  // ignore import statements
```
harupy, line 130: Screen Shot 2020-04-26 at 17 46 59

mlflow/server/js/.eslintrc.js
```diff
@@ -110,9 +113,7 @@ module.exports = {
     'import/prefer-default-export': 0,
     'import/unambiguous': 0,
     'indent': 0,
-    'indent-legacy': ['error', 2, {
-      'SwitchCase': 1,
-    }],
+    'indent-legacy': 0,
```
harupy, line 116: Prettier indents the highlighted lines, which conflicts with [indent-legacy](https://eslint.org/docs/rules/indent-legacy)

Screen Shot 2020-04-26 at 17 57 18

Zangr: This should look like the following
```
    ? Promise.resolve()
    : wrapDeferred(Services.deleteModelVersion, {
      name: modelName,
      version: version,
    }),
```
Can we tune the prettier config to make indentation like this?

mlflow/server/js/.eslintrc.js
Zangr: We need to keep a max length of 100. Can we try to find some extra config on the prettier side to work around this?

mlflow/server/js/.eslintrc.js
Zangr: This lint rule makes sense. Let's keep it by removing the conflicting prettier rule.

mlflow/server/js/.eslintrc.js
Zangr: This lint rule makes sense. Let's keep it by removing the conflicting prettier rule.

Zangr (commented):
I think those lint rules are mostly valid. Let's try to keep them as is by either removing/adding config on the prettier side.

mlflow/server/js/.eslintrc.js
harupy: Prettier removes unnecessary parentheses and it's not possible to keep `()` here. There are three options to get around with this.

1. Don't use Prettier.
2. Insert `// prettier-ignore` as a comment to make Prettier ignore these lines.
3. Disable ESLint's no-confusing-arrow

mlflow/server/js/.eslintrc.js
Zangr: I just checked the output. This is actually fine to be indented like:
```
export const deleteModelVersionApi = (modelName, version, id = getUUID(), localUpdateOnly) => ({
  type: DELETE_MODEL_VERSION,
  payload: localUpdateOnly
    ? Promise.resolve()
    : wrapDeferred(Services.deleteModelVersion, {
        name: modelName,
        version: version,
      }),
  meta: { id, modelName, version },
});
```

mlflow/server/js/.eslintrc.js
Zangr: I see. I think overall the benefit prettier provides overrules adjustments of that legacy lint rules. As far as we run `npm run lint:fix` for each commit, it should be fine.

Zangr (approved):
LGTM

harupy:
@Zangr 
Thanks for merging!
I'm planning to apply Prettier to CSS and HTML files as well.
Please let me know if you have any thoughts.

PR #2565, opened 2020-03-09
highlight changes in CompareRunView

gabrielbretschner:

This PR introduces highlighting of different parameter values when comparing runs. This is achieved by adding a class `row-changed` to the parameter row and styling it with CSS. I just chose a color myself as I was unsure which colors should be used. 

## How is this patch tested?

Tested in local installation

When comparing runs, parameters are highlighted where the runs differ.

- [x] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


mlflow/server/js/src/components/CompareRunView.js
```diff
@@ -153,7 +156,21 @@ class CompareRunView extends Component {
     });
 
     return keys.map(k => {
-      return <tr key={k}>
+      let row_class = undefined;
+      if (highlightChanges) {
+        const all_equal = data[k].reduce((acc, currentValue) => {
```
AveshCSingh, line 161:
```suggestion
        data[k].every(x => x == data[k][0])
```

I think the reduce is unnecessarily confusing. Could you just compare each value to the 0'th value?

AveshCSingh (changes requested):
Thank you for the PR, @gabrielbretschner. This looks good to me, modulo one small request inline.

@gioa, do you know if MLflow has a predefined color palette?

Screen Shot 2020-04-03 at 4 22 37 PM

gioa:
> @gioa, do you know if MLflow has a predefined color palette?

@gabrielbretschner Thank you for the work! The orange looks too prominent for me. Maybe use a light yellow for the background? See the attached mock, I use a rgba(249, 237, 190, 0.5) and it passes the [WCAG color contrast](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast) guideline. 

Diff

Another related question, what do you think of adding a switch to show different parameters only? Would that be useful for people to better compare runs?

mlflow/server/js/src/components/CompareRunView.js
gabrielbretschner: Good point. I simplified the reduce

gabrielbretschner:
> > @gioa, do you know if MLflow has a predefined color palette?
> 
> @gabrielbretschner Thank you for the work! The orange looks too prominent for me. Maybe use a light yellow for the background? See the attached mock, I use a rgba(249, 237, 190, 0.5) and it passes the [WCAG color contrast](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast) guideline.
> 
> Diff

Thank you for your suggestion. That looks much better.

> Another related question, what do you think of adding a switch to show different parameters only? Would that be useful for people to better compare runs?

That would be great, though I don't know how to implement this (I'm not a webdev...)

AveshCSingh (commented):
LGTM. I don't believe the JS test failure is related, so I've retriggered the test suite. Once it passes, you can submit.

Thanks @gabrielbretschner!

AveshCSingh (approved):

Issue #2445, opened 2020-02-19
[FR] Plugins loader should be lazy


jdlesage:
## Describe the proposal
Change the different plugin registries by lazy load functions. It will change all plugins except RunContextProviders

### Motivation
Currently the registries load all plugin installed in current virtual environment. If one plugin fails to be loaded, the error is caught (https://github.com/mlflow/mlflow/blob/master/mlflow/tracking/context/registry.py#L33) It makes the debug of the plugin more difficult.
The suggestion is to load the plugin on demand and don't catch the error if the plugin fails to be loaded.

### Proposed Changes
Remove Registry class. Replace by loader functions that use `entrypoints`

PR #2566, opened 2020-03-09
Add plugin interface for executing MLflow projects against custom backends

jdlesage:
This PR is a split from https://github.com/mlflow/mlflow/pull/2415

This one contains:
* Abstract_backend interface
* Documentation
* Example in mlflow-plugin (+ split differents plugin to avoid circular imports)
* In utils functions that fetches and prepare runs. These methods are in this PR call at the beginning of _run function. In a next PR, they will be called inside the backend implementations.

## How is this patch tested?
Running tests

Adds pluggable backend for running MLflow projects against custom targets (e.g. YARN or other job schedulers)

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [X] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python


mlflow/projects/__init__.py
```diff
@@ -93,41 +85,26 @@ def _resolve_experiment_id(experiment_name=None, experiment_id=None):
...
-        for tag in [MLFLOW_GIT_REPO_URL, LEGACY_MLFLOW_GIT_REPO_URL]:
-            tracking.MlflowClient().set_tag(active_run.info.run_id, tag, repo_url)
-
-    # Add branch name tag if a branch is specified through -version
-    if _is_valid_branch_name(work_dir, version):
-        for tag in [MLFLOW_GIT_BRANCH, LEGACY_MLFLOW_GIT_BRANCH_NAME]:
-            tracking.MlflowClient().set_tag(active_run.info.run_id, tag, version)
-
-    if backend == "databricks":
+    active_run = get_or_create_run(run_id, uri, experiment_id, work_dir, entry_point)
```
dbczumar: @jdlesage @smurching Had we come to a conclusion that each backend should be responsible for fetching project contents using predefined utility functions? (this is my preferred approach). If so, I think  we should move this logic into implementations of `backend.run()` rather than leaving them in `mlflow.projects._run`.

mlflow/projects/__init__.py
```diff
@@ -93,41 +85,26 @@ def _resolve_experiment_id(experiment_name=None, experiment_id=None):
...
+    log_project_params_and_tags(active_run, project, work_dir, entry_point, parameters, version)
```
dbczumar: In a similar vein to my question further down about whether the MLflow client or the project backend should be responsible for MLflow run creation. I'm also wondering whether the project backend or the client should be responsible for logging project parameters, the working directory, etc. 

For example, this approach seems to require the client to always fetch the project in order to determine the working directory name. It seems feasible for backends to resolve projects directly in the future, independently of the client (e.g., this would probably be the case if we support the execution of Databricks notebooks as MLflow projects). @smurching do you have thoughts here?

mlflow/projects/backend/loader.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+# Statically register backend defined in mlflow
+MLFLOW_BACKENDS = dict()
+
+
+def load_backend(backend_name):
+    # Static backends
+    if backend_name in MLFLOW_BACKENDS:
+        return MLFLOW_BACKENDS[backend_name]()
```
dbczumar: @jdlesage @smurching It seems reasonable for some backends to be referenced by URI rather than a human-readable name. For example, this would likely be the case if we introduced support for REST-based project backends. We may want to consider support for regex here was well (e.g., registering a backend to use for URIs of the form `http://.*` or `https://.*`.

Matching on the full backend name seems fine for now. If we were to add suppot of RESTful backends, it might be possible that

mlflow/projects/backend/loader.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # backends from plugin
+    try:
+        backend_builder = entrypoints.get_single(ENTRYPOINT_GROUP_NAME,
```
dbczumar: Nit: @jdlesage Can we resolve these entrypoints at module import time? (similar to what we do for the Tracking backend store registry here: https://github.com/mlflow/mlflow/blob/dea20035fc804654b552b520942eae5ed52f48d4/mlflow/tracking/_tracking_service/utils.py#L99-L102)

dbczumar (changes requested):
@jdlesage @smurching I left a few drive-by comments here that I think should be considered. Please let me know what your thoughts are or if you have any questions!

mlflow/projects/__init__.py
smurching: Thanks @dbczumar & agreed, that was the conclusion we'd reached - would be good to move towards an implementation like that of the original PR ([link](https://github.com/mlflow/mlflow/pull/2415/files#diff-16104674981d54b31e320402a143eb19R100-R114)), where the backend has the option to fetch the project, log params, perform param validation, etc

mlflow/projects/backend/abstract_backend.py
```diff
@@ -0,0 +1,30 @@
...
+
+class AbstractBackend():
+    """
+    Abstract class for MLflow Project Execution Backend
+    A backend must implement the run method to launch a new job
+    """
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def run(self, run_id, experiment_id, project_uri, entry_point, params,
```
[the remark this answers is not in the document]
smurching: Hm seems reasonable to allow the possibility of the custom backend (rather than the local Python client) creating a run, which will be returned indirectly via the `SubmittedRun` returned by `run`. The one advantage I see of the current approach is that it guarantees that the local client can access the run created for the project execution.

On the other hand, the current behavior of creating the run on the client puts the onus on backend implementors to guarantee that the backend (k8s, YARN, etc) has the appropriate credentials to log to the run. I could also see it being nice to be able to say "if you can authenticate to project backend X, you can run a project" & not have to worry about also configuring auth to a tracking server.

Regarding the conceptual relationship, in my mind it's more along the lines of "a run is a representation of project execution" than "project execution is a sub-operation of a run". That being said, I've seen use cases where users ran multiple projects within the same run, so that all the params for the different project executions would be surfaced within a single top-level run (i.e. in this case it's helpful to be able to say "backend X, please log to run 'abc'"). If backends had the flexibility to create their own run / ignore the user-supplied run ID to `mlflow.projects.run`, this use case would be difficult to achieve against those backends. 

After writing this out I'd still lean towards creating the run in the client / allowing the client to specify the run ID for the project execution, but definitely open to your alternative - were there other use cases/thoughts you had in mind to motivate this?

mlflow/projects/backend/loader.py
smurching: I think your comment might have been cut off :P - I guess with the current scheme, we'd handle HTTP backends via a custom backend type, e.g. `mlflow-http` (probably named something better) and then the actual URI of the backend would be a field in the backend config

mlflow/projects/backend/abstract_backend.py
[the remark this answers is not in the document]
smurching: I like this idea, makes sense to me

mlflow/projects/backend/loader.py
dbczumar: Whoops! Accidentally left some trailing content at the end there. Ah, good point - putting the URL in the config sounds fine to me!

mlflow/projects/backend/abstract_backend.py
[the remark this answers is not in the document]
dbczumar: Got it! My concern is that, if we were to add this later, existing plugins may ignore the argument entirely, even if users supply non-`None` values.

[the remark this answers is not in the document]
dbczumar: @smurching I think you really nailed it with this explanation of the tradeoffs. In retrospect, I think my primary motivation for raising this was to determine whether or not we think the project backend and client should be tied to the same particular MLflow Tracking server. Given that the current structure of MLflow Projects operates under the assumption that both the client and the project executor are referring to Runs in the same location, I think it's fine to continue delegating run creation to the client and to stipulate that the Projects backend must communicate with the same MLflow Tracking server.

mlflow/projects/__init__.py
dbczumar: Given the discussion in https://github.com/mlflow/mlflow/pull/2566#discussion_r394840637, I think it's fair to delegate these responsibilities to the MLflow client, as is done here.

dbczumar: As an outcome of discussion here: https://github.com/mlflow/mlflow/pull/2566/files#r394638164, I think we agree that client run creation makes sense. However, we also agree that project backends should manage the resolution of projects; this presents a bit of a wrinkle with respect to the tag populated by `work_dir`, which isn't determined until the project is fetched. Perhaps backends should set this tag instead, leaving everything else to the client?

mlflow/projects/backend/loader.py
jdlesage: I think it is better to keep them lazy and don't import the plugin if it is not needed. You don't control the plugin implementations, and some of them can contain bugs. If mlflow imports all existing plugins to register them and one of these plugins have a bug then mlflow will not start.
There is already this problem with other registries. If you install in your virtual env the dummy project located in `tests/resources`, there are these warnings when mlflow is imported:
```
/home/jd.lesage/workspace_python/mlflow/mlflow/tracking/_model_registry/utils.py:61: UserWarning: Failure attempting to register store for scheme "file-plugin": bad magic number in 'mlflow_test_plugin.sqlalchemy_store': b'\x03\xf3\r\n'
  _model_registry_store_registry.register_entrypoints()
```
To workaround this problem, currently, these exceptions are caught and transformed in warnings.  I think it is safer to raise exceptions (stop if the plugin is not correct) and load only necessary plugins. 

I created a week ago a issue to make all registries lazy: https://github.com/mlflow/mlflow/issues/2445

dbczumar: @jdlesage Great point. In that case, I agree that we should stick with lazy entrypoint reading and move other store registries (e.g, Tracking) to this paradigm as well.

mlflow/projects/__init__.py
jdlesage: I created in utils some helper functions to fetch, validate, log and even tag. So backends can call them.
I did a mistake in my PR. I just pushed a change to fix it:  these functions are called in the client only if the backend is not managed by a plugin.
When all the existing backends will implement the abstract_class, I will remove all these calls in the `_run` method. As we agreed to split the PR (1 for abstract_class then 1 PR per backend), I will remove them in another PR.
Do you agree with this change ?

mlflow/projects/backend/abstract_backend.py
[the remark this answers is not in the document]
jdlesage: Agree to add this parameter. But in `projects/__init__.py`, the plumbing to inject the tracking uri does not exist yet. Do you agree to
* Add the parameter to the interface
* in `_run` method always injects the result of `get_tracking_uri` method ? (= don't change the interface of the `_run` method)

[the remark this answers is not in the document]
smurching: @jdlesage that makes sense to me (if we later added a less-fluent API for project execution, users could supply the tracking URI via that API, but makes sense to just use the current tracking URI in the case where the user's invoking `mlflow.projects.run`)

mlflow/projects/__init__.py
```diff
@@ -91,19 +91,21 @@ def _run(uri, experiment_id, entry_point="main", version=None, parameters=None,
     Helper that delegates to the project-running method corresponding to the passed-in backend.
     Returns a ``SubmittedRun`` corresponding to the project run.
     """
+    if backend_name:
+        backend = loader.load_backend(backend_name)
+        if backend:
+            submitted_run = backend.run(run_id, experiment_id, uri, entry_point, parameters,
+                                        version, backend_config)
+            tracking.MlflowClient().set_tag(submitted_run.run_id, MLFLOW_PROJECT_BACKEND,
```
smurching: Nit: to be safe, set the tag prior to calling `backend.run`? Just in case `backend.run` throws for whatever reason

mlflow/projects/utils.py
```diff
@@ -80,3 +100,186 @@ def _is_valid_branch_name(work_dir, version):
         except GitCommandError:
             return False
     return False
+
+
+def generate_env_vars_to_attach_to_run(run):
```
smurching: Nit: Could we remove this? Doesn't seem to currently be used anywhere, and seems like it'd only be used by the local project execution backend (let me know if I missed something), so would be better not to expose it in this `mlflow.projects.utils` module, given that we expect plugin developers to call functions from this module.

mlflow/projects/utils.py
```diff
@@ -80,3 +100,186 @@ def _is_valid_branch_name(work_dir, version):
...
+    if source_version is not None:
+        tags[MLFLOW_GIT_COMMIT] = source_version
+    if parent_run_id is not None:
+        tags[MLFLOW_PARENT_RUN_ID] = parent_run_id
+
+    active_run = tracking.MlflowClient().create_run(experiment_id=experiment_id, tags=tags)
+    return active_run
+
+
+def log_project_params_and_tags(run, project, work_dir, entry_point, parameters, version):
```
smurching: Should this be internal (`_log_project_params_and_tags`) if the primary API we expect plugin developers to call is `prepare_project_and_run`?

mlflow/projects/utils.py
```diff
@@ -80,3 +100,186 @@ def _is_valid_branch_name(work_dir, version):
...
+    # Add branch name tag if a branch is specified through -version
+    if _is_valid_branch_name(work_dir, version):
+        for tag in [MLFLOW_GIT_BRANCH, LEGACY_MLFLOW_GIT_BRANCH_NAME]:
+            tracking.MlflowClient().set_tag(run.info.run_id, tag, version)
+
+
+FetchedProject = collections.namedtuple("FetchedProject", ["project", "work_dir", "run"])
+
+
+def prepare_project_and_run(project_uri, experiment_id, run_id, entry_point, parameters, version):
```
smurching: Would be good to add an end-to-end test for this method btw, in tests/projects/utils.py. I also wonder if it'd be easier for plugin implementors if we just exposed the component utility functions that comprise this method (`fetch_and_validate_project`, `get_or_create_run`, `log_project_params_and_tags`) and showed an example of how to call the three functions as we do here in order to fetch the project, create a run, and set appropriate tags - it may be confusing for plugin implementors to reason about what exactly happens in this method & its NamedTuple return value. What do you think?

smurching (commented):
@jdlesage thank you for the updates! This largely looks good IMO, leaving some comments (& would be good to address comment around exposing a `tracking_uri` argument)

mlflow/projects/__init__.py
jdlesage: Here run_id can be None before the run method. Or backend implementation can create a new run_id. Therefore, to tag the run, the run_id must be the one from the submitted_run object.

mlflow/projects/backend/abstract_backend.py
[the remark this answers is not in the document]
jdlesage: Done

mlflow/projects/utils.py
jdlesage: Done

mlflow/projects/__init__.py
jdlesage: This method `log_project_params_and_tags` is supposed to be called by project backends. It is stilled called in `_run` method to keep the previous behavior for databricks, local and k8s backends.
It will be removed as soon they backends will implement the new interfaces (in a next PR)

FYI I will implement also @smurching's suggestion to remove this method and to give access to the methods that create runs, fetch the project & validate the config, etc.

jdlesage:
I closed the remarks I fixed. There are still the last remarks about utils methods that I have still to do.

mlflow/projects/utils.py
jdlesage: I remove the global method and created a test to validate the calls of these 3 methods.

mlflow/projects/utils.py
jdlesage: As the `prepare_project_and_run` is removed (to take in account your other comment). We should now keep them public as they will call by backend developers.

jdlesage:
All checks have passed and I have implemented all the suggestions. 🎉 Could you please have a look ? Thanks!

mlflow/projects/__init__.py
dbczumar: @smurching @jdlesage I was under the impression from prior conversation (https://github.com/mlflow/mlflow/pull/2566/files#r394638164) that we agreed to require the client to create the MLflow Run prior to beginning project backend execution and, therefore, that `run_id` must be defined prior to `backend.run()`. However, this dialogue seems to imply that the backend may be responsible for creating the run.

@smurching @jdlesage Please let me know if I am misinterpreting the outcome of https://github.com/mlflow/mlflow/pull/2566/files#r394638164. For consistency with preexisting behavior, I would prefer if the client be responsible for creating the MLflow run before beginning backend execution.

jdlesage: Hi @dbczumar , Sorry, I misunderstood your first remark. 
The problem by creating the run in the client is the validation databrick test (this one: https://github.com/mlflow/mlflow/blob/master/tests/projects/test_databricks.py#L179) cannot be implemented.
In my first attempt, I created the runs in the client. But there are additionnal checks done by the databricks backend. If the backend_config is wrong, the run is not created to not "spam" the tracking server. To implement this behavior, it is needed that the databricks backend controls the run creation.
I don't have strong opinion about this. Could you confirm the client creates the run before to call the backend, please ? In this case, databricks projects will generate FAILED runs in the tracking server if the backend config is invalid.

smurching: Ah, interesting...given this I'm wondering if we should reconsider forcing the client to create the run, the alternative would be to add a `validate_backend_config` method to the AbstractBackend API. Will sync with @dbczumar tomorrow on this & get back to you!

mlflow/projects/__init__.py
```diff
@@ -204,13 +184,13 @@ def _run(uri, experiment_id, entry_point="main", version=None, parameters=None,
...
     raise ExecutionException("Got unsupported execution mode %s. Supported "
-                             "values: %s" % (backend, supported_backends))
+                             "values: %s" % (backend_name, supported_backends))
 
 
 def run(uri, entry_point="main", version=None, parameters=None,
         experiment_name=None, experiment_id=None,
         backend=None, backend_config=None, use_conda=True,
-        storage_dir=None, synchronous=True, run_id=None):
+        storage_dir=None, synchronous=True, run_id=None, tracking_store_uri=None):
```
smurching, line 193:
```suggestion
        storage_dir=None, synchronous=True, run_id=None):
```

Sorry for the confusion on this one, I don't think we should necessarily expose a `tracking_store_uri` argument to the top-level `mlflow.projects.run` API, or at least not in this PR :). Seems reasonable to have it in the backend though.

mlflow/projects/utils.py
```diff
@@ -80,3 +99,150 @@ def _is_valid_branch_name(work_dir, version):
...
+
+def fetch_and_validate_project(uri, version, entry_point, parameters):
+    parameters = parameters or {}
+    work_dir = fetch_project(uri=uri, force_tempdir=False, version=version)
+    project = _project_spec.load_project(work_dir)
+    project.get_entry_point(entry_point)._validate_parameters(parameters)
+    return project, work_dir
+
+
+def fetch_project(uri, force_tempdir, version=None):
```
smurching, line 112:
```suggestion
def fetch_project(uri, version=None):
```

I think we can remove force_tempdir, as all usages pass `force_tempdir=False`

smurching (commented):
@jdlesage thanks for the updates, pushed a commit to simplify some of the project utils & update tests accordingly. I think the one remaining question is around where run creation logic lives, will get an answer for you on that tomorrow. 

We should also update the docs under `docs/source/python_api/projects.rst` to show the `mlflow.projects.utils` module, by adding a sphinx automodule (see [sphinx docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-automodule)) directive like:

```
.. automodule:: mlflow.projects
    :members:
    :undoc-members:
    :show-inheritance:
```

mlflow/projects/__init__.py
smurching: Sync'd offline with @dbczumar on this, it seems reasonable to allow the backend to create runs. Could we update the backend interface to match that in flow 2 of [this gist](https://gist.github.com/smurching/6fe7368c6075ee3acf0220c672185b5e#file-project-backend-py-L57-L58)? Let me know if you have any questions, thanks!

dbczumar: @smurching @jdlesage We should also include `tracking_uri` as part of this interface. (didn't see it in the gist).

smurching: Sorry, I wrote the gist in a confusing way - the  top-level `def run` method definition in the gist is actually meant to mimic `mlflow.projects.run` (updated the gist to be clearer about that), hence why it doesn't accept a `tracking_uri` argument. To confirm:

* We should not add the `tracking_uri` argument to the user-facing `mlflow.projects.run` API (at least not in this PR)
* The `AbstractBackend.run` API called by `mlflow.projects.run` should accept a `tracking_uri` argument, as @dbczumar said.

jdlesage: I did your last comment.
1/ AbstractBackend.run accepts a tracking_uri
2/ `tracking_uri` is not in the signature of `mlflow.projects.run`

mlflow/projects/__init__.py
jdlesage: Done

smurching (approved):
@jdlesage thanks for the hard work on this! Pushed some doc fixes after which I think we can merge - 

I'll follow up with a PR with additional doc updates & to refactor some of the project utils to make implementing a custom plugin easier (if you look at the code for the existing built-in project backends, you can see they use many internal / undocumented APIs under `mlflow.projects._project_spec` etc).

Once that PR's in, would welcome your help with follow-up PRs that refactor the existing execution backends to use the plugin interface - thanks!

PR #2578, opened 2020-03-12
Add a check in cli gc command

jcuquemelle:

Without this check, the mlflow server API can be passed as argument and
is resolved as a rest_store

This would allow anyone without knowledge of credentials on the
 backend_store to delete artifacts associated with runs

## How is this patch tested?

(Details)

- [ ] UI
- [x] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -304,6 +304,9 @@ def gc(backend_store_uri, run_ids):
     This command deletes all artifacts and metadata associated with the specified runs.
     """
     backend_store = _get_store(backend_store_uri, None)
+    if not hasattr(backend_store, '_hard_delete_run'):
+        raise MlflowException(
+            "This cli can only be used with a backend that allows hard-deleting runs")
     if not run_ids:
         run_ids = backend_store._get_deleted_runs()
     else:


smurching (approved):
LGTM

mlflow/cli.py
```diff
@@ -304,6 +304,9 @@ def gc(backend_store_uri, run_ids):
     This command deletes all artifacts and metadata associated with the specified runs.
     """
     backend_store = _get_store(backend_store_uri, None)
+    if not hasattr(backend_store, '_hard_delete_run'):
```
dbczumar: @jdlesage @smurching I'm a bit confused as to why this is necessary. The `_hard_delete_run` method is not implemented in the REST store, so this CLI would not work when targeted against the MLflow Server directly. Further, client-side changes won't address server-side gaps if they were to exist (e.g., users can simply remove these new lines of code and run `mlflow gc`). 

Regardless, I think this is a nice improvement to the UX because it raises a clear error message when an invalid store is supplied - thanks @jdlesage. Please feel free to correct me if I'm missing something here.

jcuquemelle: @dbczumar I think you've mistakenly pinged jd.lesage ;-) 

The issue was that the error about not implemented _hard_delete_run would be thrown after having already deleted artifacts in the artifacts store. 
This check avoids to mistakenly wipe artifacts.

PR #2592, opened 2020-03-17
Set rootID at run creation time and passed it down to all descendants.

arjundc-db:
Nested Run Deletion: Setting rootRunId at run creation time in the Python client.

## How is this patch tested?
Unit tested using pytest.
(Details)

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ x] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -135,15 +135,17 @@ def start_run(run_id=None, experiment_id=None, run_name=None, nested=False):
     else:
         if len(_active_run_stack) > 0:
             parent_run_id = _active_run_stack[-1].info.run_id
+            root_run_id = _active_run_stack[0].info.run_id
         else:
             parent_run_id = None
 
         exp_id_for_run = experiment_id if experiment_id is not None else _get_experiment_id()
 
         user_specified_tags = {}
+
         if parent_run_id is not None:
             user_specified_tags[MLFLOW_PARENT_RUN_ID] = parent_run_id
-            user_specified_tags[MLFLOW_ROOT_RUN_ID] = _active_run_stack[0].info.run_id
+            user_specified_tags[MLFLOW_ROOT_RUN_ID] = root_run_id
         else:
             user_specified_tags[MLFLOW_ROOT_RUN_ID] = None
 
@@ -451,7 +453,7 @@ def search_runs(experiment_ids=None, filter_string="", run_view_type=ViewType.AC
                 params[key].append(PARAM_NULL)
         new_params = set(run.data.params.keys()) - param_keys
         for p in new_params:
-            params[p] = [PARAM_NULL]*i  # Fill in null values for all previous runs
+            params[p] = [PARAM_NULL] * i  # Fill in null values for all previous runs
             params[p].append(run.data.params[p])
 
         # Metrics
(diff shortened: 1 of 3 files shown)


mlflow/tracking/fluent.py
```diff
@@ -146,6 +147,13 @@ def start_run(run_id=None, experiment_id=None, run_name=None, nested=False):
         if run_name is not None:
             user_specified_tags[MLFLOW_RUN_NAME] = run_name
 
+        if _active_run_stack:
+            user_specified_tags[MLFLOW_ROOT_RUN_ID] = _active_run_stack[0].data.tags[MLFLOW_ROOT_RUN_ID]
```
smurching:
```suggestion
            user_specified_tags[MLFLOW_ROOT_RUN_ID] = _active_run_stack[-1].data.tags[MLFLOW_ROOT_RUN_ID]
```

Given that we append to the active run stack, I think you want the most recent item on the stack, not the first one. Would be good to add a test case that verifies this behavior! (i.e. a test case that would fail if the code were as is, and passes with the code updated to get the most-recent item on the stack)

mlflow/tracking/fluent.py
```diff
@@ -146,6 +147,13 @@ def start_run(run_id=None, experiment_id=None, run_name=None, nested=False):
...
+        else:
+            user_specified_tags[MLFLOW_ROOT_RUN_ID] = uuid.uuid4().hex
```
smurching:
```suggestion
            user_specified_tags[MLFLOW_ROOT_RUN_ID] = uuid.uuid4().hex
```

We want to be able to look up all runs that are descendants of a root run based on its ID, so this needs to be equal to the ID of the created run if we're creating a top-level run. Since the run ID is generated by the server at run creation time (as part of handling a CreateRun request), we must instead get the run ID from `active_run_obj` below https://github.com/mlflow/mlflow/pull/2592/files#diff-581b8f8c85a274ab9d0317aec207b364R159 and set it as a tag on the newly-created run

smurching (commented):
Thanks @arjundc-db for the PR! Left some comments, also would be good to add test cases to test_tracking.py to verify that appropriate tags are set for nested runs - thanks!

smurching:
In particular, with regards to tests it's important that we have good [branch coverage](https://en.wikipedia.org/wiki/Code_coverage#Basic_coverage_criteria) for the value of the MLFLOW_ROOT_RUN_ID tag. Happy to discuss more tomorrow.

mlflow/tracking/fluent.py
```diff
@@ -146,6 +147,13 @@ def start_run(run_id=None, experiment_id=None, run_name=None, nested=False):
...
+
+
+
         tags = context_registry.resolve_tags(user_specified_tags)
 
         active_run_obj = MlflowClient().create_run(
```
smurching: MlflowClient().set_tag(run_id=active_run_obj.info.run_id, key=MLFLOW_ROOT_RUN_ID, value=active_run_obj.info.run_id)

arjundc-db: Not needed, since second RPC is not made

PR #2594, opened 2020-03-17
Updated azure-storage-blob to v12

Trollgeir:

azure-storage-(blob) is updated to version 12.

This version has breaking changes in the API, hence a lot of code rewriting.

## How is this patch tested?

Unittests (rewritten)
pylint

- [x] No.

### DIFF
--- a/docs/source/tracking.rst
+++ b/docs/source/tracking.rst
@@ -557,7 +557,7 @@ To store artifacts in Azure Blob Storage, specify a URI of the form
 MLflow expects Azure Storage access credentials in the
 ``AZURE_STORAGE_CONNECTION_STRING`` or ``AZURE_STORAGE_ACCESS_KEY`` environment variables (preferring
 a connection string if one is set), so you must set one of these variables on both your client
-application and your MLflow tracking server. Finally, you must run ``pip install azure-storage``
+application and your MLflow tracking server. Finally, you must run ``pip install azure-storage-blob``
 separately (on both your client and the server) to access Azure Blob Storage; MLflow does not declare
 a dependency on this package by default.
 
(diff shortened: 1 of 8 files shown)


tomasatdatabricks (commented):
Thanks for the contribution! I need to test it out with actual blob storage but in general the PR looks good. 

I have a slight concern about breaking dependencies of other azure libraries. For example, it looks like azure python sdk requires azure-blob-storage~=1.3.  That would mean that mlflow users who also use azure sdk would not be able to use the new repository. I am not sure if this was already the case (i.e. if the current mlflow client would work with 1.3 or not) but I will test that as well. Do you have any context on this by any chance?

Trollgeir:
> Thanks for the contribution! I need to test it out with actual blob storage but in general the PR looks good.
> 
> I have a slight concern about breaking dependencies of other azure libraries. For example, it looks like azure python sdk requires azure-blob-storage~=1.3. That would mean that mlflow users who also use azure sdk would not be able to use the new repository. I am not sure if this was already the case (i.e. if the current mlflow client would work with 1.3 or not) but I will test that as well. Do you have any context on this by any chance?

From what I can see at https://azure.github.io/azure-sdk-for-python/ref/Storage.html#azure-storage-blob - azure python SDK 2.0 is up to date with azure-storage-blob 12.x. Are you referring to SDK 1.0 ?

tomasatdatabricks:
I am talking about azure pip package:
```
pip show azure             
Name: azure
Version: 4.0.0
Summary: Microsoft Azure Client Libraries for Python
Home-page: https://github.com/Azure/azure-sdk-for-python
Author: Microsoft Corporation
Author-email: azpysdkhelp@microsoft.com
License: MIT License
```

Trollgeir:
I understand. Although I'd advocate that the MLFlow should not guarantee any support for azure 4.0.0  
`azure-blob-storage` is as minimalistic of a dependency as you'd get. Given the big change to drop python 2 in the upcoming 1.8, I take it people are prepared for changes in their current environment.

tomasatdatabricks (approved):
lgtm! Thanks for the contributiuon!

PR #2601, opened 2020-03-18
UI: Display all columns in the experiment view

### LINKED ISSUE #2475: [BUG] Columns disappear after sort in UI
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Centos
- **MLflow version (run ``mlflow --version``)**: latest
- **Python version**: 3.6

### Describe the problem
Columns (metrics, params, tags) in the ExperimentView are computed from the run displayed in the page. If a column is not present in the first X runs, it will not be in the UI..
It does not work well with the sorting & fitlering features. Following the runs that have been extracted from DB, the columns will change. 
columnsmlflow

### Code to reproduce issue
Sort on a experiments with numerous runs but different tags,metrics and params

jdlesage:

This ticket is about to fix https://github.com/mlflow/mlflow/issues/2475

Add a new service that list all columns of a experiment
In javascript, call this service when a new search is triggered.

For the future: Tags, params and metrics list are not injected in the ExperimentView component. It could be reuse to implement autocompletion in the search bar.

## How is this patch tested?

Tested on our data + unit test

all_columns

All columns (parameters, tags, or metrics) are displayed in the experiment view even if the first runs does not contain values for these columns.

- [X] UI
- [ ] CLI
- [ ] API
- [X] REST-API
- [ ] Examples
- [X] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python


mlflow/entities/columns.py
```diff
@@ -0,0 +1,40 @@
+from mlflow.entities._mlflow_object import _MLflowObject
+from mlflow.protos.service_pb2 import ListAllColumns
+
+
+class Columns(_MLflowObject):
```
mparkhe, line 5: This is not a ML run entity, but encodes meta data.

mlflow/protos/service.proto
```diff
@@ -918,6 +936,20 @@ message GetRun {
   }
 }
 
+message ListAllColumns {
+  option (scalapb.message).extends = "com.databricks.rpc.RPC[$this.Response]";
+
+  // List of experiment IDs to search over.
+  required string experiment_id = 1;
```
mparkhe, line 943: This API is very specific/custom API for showing columns in UI and not particularly generalizable over multiple experiments ... etc

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -604,6 +604,21 @@ def delete_tag(self, run_id, key):
...
+        with self.ManagedSessionMaker() as session:
+            tags = [r[0] for r in session.query(SqlTag.key).join(SqlRun)
+                    .filter(SqlRun.experiment_id == experiment_id,
+                            SqlRun.lifecycle_stage.in_(stages)).distinct().all()]
+            metrics = [r[0] for r in session.query(SqlLatestMetric.key).join(SqlRun)
+                       .filter(SqlRun.experiment_id == experiment_id,
+                               SqlRun.lifecycle_stage.in_(stages)).distinct().all()]
+            params = [r[0] for r in session.query(SqlParam.key).join(SqlRun)
+                      .filter(SqlRun.experiment_id == experiment_id,
+                              SqlRun.lifecycle_stage.in_(stages)).distinct().all()]
```
mparkhe, line 618: These queries can be super expensive if the experiment has a lot of runs or runs have lots of params/tags/... etc

mparkhe (changes requested):
I am not sure I understand the exact use cases and motivations behind this systemic change. From the description in #2475, it is not super clear to me, if other solutions were explored for the specific issues (there may be other ways of solving these problems):
 - columns changing: one of ways that can be fixed is pinning columns down and can be achieved as a UI only change. ditto with features for sorting and filtering columns.
 - column ordering: when more runs are loaded (with pagination), all the sorting happens in the UI

What you are implementing is a derived view from information that is already available. For cases like these we have deliberately avoided added additional APIs and banked on other APIs (eg: list_run_infos -- which depends on paginated search_runs). One of the ways of implementing this would be to call search runs in a paginated manners and extracting columns names--this is not very efficient UI code in cases large run data, but those performance challenges would affect this solution as well.

A few specific concerns about the implementation and summarizing comments:

1. `Column` REST and python entity is too specific as a view attribute  in the UI (encodes metadata) and is not really an ML run entity. This is a new type of entity in the tracking server data model. 
2. The `ListAllColumns` API is specific to UI and not really usable for generic use cases.
3. This breaks pagination model for pull large datasets. In cases of experiments with a large set of runs or with runs having large # of params/tags/metric.. etc listing all columns can be super expensive. In the UI, this can cause several seconds of delay if the user inadvertently clicks on the columns (as the feature is designed). MLflow UI and APIs are used by a lot of different backend implementations and depending on how backend developers have implemented their backend, this could be prohibitive and in some cases may not be feasible.
4. For cases, where an experiment has runs with a lot of different param/tag names -- this can case a very sparse looking table and user may have to scroll back and forth, up and down leading to unpleasant UX.
5. A lot of backend code between search runs and list all columns is duplicated. Particularly, adding new entities/updates are very expensive. Any changes in the data model, additional entities, backend -- needs to be updated in both places. Additionally, such changes affect the entire stack: REST, backend, front end, clients, ... and would lead to pervasive changes similar to this PR, so warrants some thought and consideration about future updatability.

---

As I mentioned before, we want to be careful and deliberate about adding new APIs and extending the data-model. I am happy to set up some time and catch up over voice/video chat and understand some of the design choices here.

rom1504:
Hello there,
I have some comments about the use cases for being able to know what are metrics, tags and params for an experiment.

> The ListAllColumns API is specific to UI and not really usable for generic use cases.

being able to list all columns for a given experiment is a very basic feature that is needed to be able to do any kind of analysis on the data.
To give a few examples : 
* I'm given an mlflow experiments with a lot of runs, a lot of different people having started these runs, there are hundred of columns ; I want to be able to compute correlation between metrics and params => I need to know what are all these metrics and params first
* I want to get some runs by filtering by some params and tags but I don't know the exact names of these params and tags => I need to know what are these metrics and params

and as pointed out by this PR name, it also simply fixes a bug in the current UI and enable creating more powerful UIs (such as what was proposed there https://github.com/mlflow/mlflow/issues/2084 )

mlflow/store/tracking/sqlalchemy_store.py
rom1504: these queries are not actually expensive compared to what search runs is doing.
search runs today doesn't scale beyond 2000 runs and 100 of params/tags/metrics
These queries definitely scale way beyond that as they do not contain any join, so it's definitely in the range of scale that mlflow currently support

mlflow/protos/service.proto
rom1504: Mlflow today doesn't support comparison between different experiments.
This PR follows this pattern, is this something that's meant to be changed in mlflow in the future ?

Issue #2341, opened 2020-01-23
[FR] Add usage of model registry to examples under examples directory


smurching:
## Describe the proposal
We should add an example that uses the model registry component to the set of open-source examples under the [`examples` directory](https://github.com/mlflow/mlflow/tree/master/examples), and link to it from the Tutorials & Examples doc page ([see docs source code](https://github.com/mlflow/mlflow/blob/master/docs/source/tutorials-and-examples/index.rst))

We could achieve this by updating the existing scikit-learn wine example (https://github.com/mlflow/mlflow/tree/master/examples/sklearn_elasticnet_wine) to call `log_model` with the `registered_model_name` argument set, which will result in registration of a new model version.

In a follow-up PR, we can also add a README under the `sklearn_elastic_wine` directory explaining what the example does & how to view registered models generated by the example in the UI, etc

### Motivation
An example that uses the model registry would make it more discoverable / make it easier for users to get started.

netanel246:
@smurching Hi, I would like to pick this issue if no one else is working on it.

smurching:
@netanel246 that'd be great! cc-ing @dmatrix, who recently worked on adding some model-registry code snippets to the model registry doc page. 

I imagine it'd still be very useful to update the existing MLflow examples to log models to the registry. The main nice thing is that now, in the READMEs for different examples, we can link to @dmatrix's new docs in https://mlflow.org/docs/latest/model-registry.html#ui-workflow rather than explaining how to view the UI ourselves.

Any thoughts @dmatrix?

dmatrix:
There are three primary ways  [Model Registry API workflows](https://mlflow.org/docs/latest/model-registry.html#api-workflow) to work with the model registry. None is a default one, and its usage and choice will depend on at what point in your experiment runs you will want to create or register a model. 

Instead of advising or showing examples to register the model while using mlflow.<model_flavor>.log_modelI(..), we should rather suggest the three ways you can register a model. 

This way a developer can choose at what point during his or her experiments are they ready to commit the model to the registry and start collaborating with the rest. 

If we update all the MLflow examples to register the model while logging the model, then we are dictating or setting the best practice, which may not always be the case. The workflows or usage I envision among developers is as follows:

**Scenario 1:** I know exactly which model I want to register as version 1, do it while running experiments, and let it create multiple versions, and I'll choose the best version later to transition to "Stage"
- create a model, log all metrics, and register as part of mlflow.log_model(...)

[...]

PR #2609, opened 2020-03-20
Added usage of the model registry to the log_model function

netanel246:
Add to the sklearn_elasticnet_wine example the ability to register the model,
if the user uses tracking URL store other then "file"

This refers to #2341

## How is this patch tested?

Ran tests.

Add to the sklearn_elasticnet_wine example the ability to register the model,
if the user uses tracking URL store other then "file"

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [X] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/examples/sklearn_elasticnet_wine/train.py
+++ b/examples/sklearn_elasticnet_wine/train.py
@@ -72,4 +72,15 @@ if __name__ == "__main__":
         mlflow.log_metric("r2", r2)
         mlflow.log_metric("mae", mae)
 
-        mlflow.sklearn.log_model(lr, "model")
+        tracking_url_type_store = mlflow.get_tracking_uri().split(":")[0]
+
+        # Model registry does not work with file store
+        if tracking_url_type_store != "file":
+
+            # Register the model
+            # There are other ways to use the Model Registry, which depends on the use case,
+            # please refer to the doc for more information:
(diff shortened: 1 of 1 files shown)


examples/sklearn_elasticnet_wine/train.py
```diff
@@ -72,4 +72,15 @@ def eval_metrics(actual, pred):
         mlflow.log_metric("r2", r2)
         mlflow.log_metric("mae", mae)
 
-        mlflow.sklearn.log_model(lr, "model")
+        tracking_url_type_store = mlflow.get_tracking_uri().split(":")[0]
```
sueann: Cool, thanks! another way to do this would be 
```
urllib.parse.urlparse(mlflow.get_tracking_uri()).scheme
```
but this look okay to me in an example context.

sueann (approved):

examples/sklearn_elasticnet_wine/train.py
netanel246: That's much more cleaner.
I prefer to change that to your suggestion if you don't mind :)
Thanks.

PR #2620, opened 2020-03-22
Change line smoothing on Exponential Moving Average

### LINKED ISSUE #2596: [FR] Line smoothing based on moving average rather than on splines
The current implementation (based on Catmull-Rom splines) doesn't actually show the overall trend. I think it is better to use some variant of moving average smoothing (https://en.wikipedia.org/wiki/Moving_average). 

E.g. here splines just make interpolation:
image
image

P.S. Tensorboard uses namely exponential moving average.

Valentyn1997:

Resolves https://github.com/mlflow/mlflow/issues/2596. 
Added method to smooth the array of values with EMA, changed Smoothing Slider range (now it corresponds to the inverted degree of weighting, from 1 to 100).

## How is this patch tested?

Visual inspection (I don't know, how to write tests for JS):
image
image
image

- [X] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


mlflow/server/js/src/components/MetricsPlotView.js
```diff
@@ -1,13 +1,25 @@
...
-import PropTypes from 'prop-types';
+import PropTypes, {array} from 'prop-types';
 import { X_AXIS_STEP, X_AXIS_RELATIVE } from './MetricsPlotControls';
 import { CHART_TYPE_BAR } from './MetricsPlotPanel';
 import Plot from 'react-plotly.js';
 
 const MAX_RUN_NAME_DISPLAY_LENGTH = 36;
 
+function EMA(mArray, mRange) {
+  const k = 2 / (mRange + 1);
```
dbczumar: @Valentyn1997 I tried out this EMA code and found that the behavior is a little bit counterintuitive in certain cases. I've gone ahead and prototyped an alternative routine inspired by TensorBoard. Here are some examples of the before (the EMA routine introduced by this PR) and after (the prototyped EMA routine):

Before:
before1
After:
after1

Before:
before2
After:
after2

Before:
before3
After:
after3

In these three cases, I believe that the `after` examples provide more accurate smoothing behavior. Here's the proposed routine, which performs a bias correction ("debiasing") routine to ensure that the first few elements of the metric value set aren't artificially skewed lower:

```
function EMA(mArray, smoothingWeight) {
  // If all elements in the set of metric values are constant, or if
  // the degree of smoothing is set to the minimum value, return the
  // original set of metric values
  if (smoothingWeight <= 1 || mArray.every((v) => v === mArray[0])) {
    return mArray;
  }

  const smoothness = smoothingWeight / 101;
  const smoothedArray = [];
  let biasedElement = 0;
  for (let i = 0; i < mArray.length; i++) {
    biasedElement = (biasedElement * smoothness) + ((1 - smoothness) * mArray[i]);
    // To avoid biasing earlier elements toward smaller-than-accurate values, we divide
    // all elements by a `debiasedWeight` that asymptotically increases and approaches
    // 1 as the element index increases
    const debiasWeight = 1.0 - Math.pow(smoothness, i + 1);
    const debiasedElement = biasedElement / debiasWeight;
    smoothedArray.push(debiasedElement);
  }
  return smoothedArray;
}
```

Please let me know how you feel about using this routine instead (we may also want to replace the constant `101` with `max + 1` where `max` is the maximum value configured in the smoothness slider).

cc @smurching and @Zangr for a review here as well.

dbczumar (changes requested):
@Valentyn1997 Thanks so much for this contribution. I've left a couple of comments; the biggest one is a proposed change to the EMA calculation.

mlflow/server/js/src/components/MetricsPlotView.js
smurching: Makes sense / looks reasonable to me!

smurching (commented):
Thanks @Valentyn1997 & @dbczumar, agree with @dbczumar's suggestion to use the TensorBoard-based EMA algorithm (& would be great to add tests for that algorithm under  `mlflow/server/js/src/components/MetricsPlotView.test.js`).

One UX question I have, since we're actually changing the values of the Y axis in the graph - if the user enables smoothing, can we enhance the plot tooltips that appear when we hover over an individual point to show both the original & smoothed values? e.g. see the tooltip in this TensorBoard example:

image

It seems possible to customize tooltip content by passing an extra `text` value into the `data` array passed to Plotly [here](https://github.com/mlflow/mlflow/blob/9ec25f157a548fa8e60ee13e3be5c3d2d513a3d5/mlflow/server/js/src/experiment-tracking/components/MetricsPlotView.js#L56), as per [the Plotly docs](https://plotly.com/javascript/hover-text-and-formatting/), I'd recommend giving that a try - thanks!

mlflow/server/js/src/components/MetricsPlotView.js
Valentyn1997: Yes, debiased version is much better!

Valentyn1997: Added - thanks for the function code!

mlflow/server/js/src/components/MetricsPlotView.js
```diff
@@ -61,10 +73,10 @@ export class MetricsPlotView extends React.Component {
           }
           return MetricsPlotView.parseTimestamp(entry.timestamp, history, xAxis);
         }),
-        y: history.map((entry) => entry.value),
+        y: EMA(history.map((entry) => entry.value), lineSmoothness),
         type: 'scatter',
         mode: isSingleHistory ? 'markers' : 'lines+markers',
-        line: { shape: 'spline', smoothing: lineSmoothness },
+        // line: { shape: 'spline', smoothing: lineSmoothness },
```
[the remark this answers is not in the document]
Valentyn1997: Done

Valentyn1997:
> Thanks @Valentyn1997 & @dbczumar, agree with @dbczumar's suggestion to use the TensorBoard-based EMA algorithm (& would be great to add tests for that algorithm under `mlflow/server/js/src/components/MetricsPlotView.test.js`).
> 
> One UX question I have, since we're actually changing the values of the Y axis in the graph - if the user enables smoothing, can we enhance the plot tooltips that appear when we hover over an individual point to show both the original & smoothed values? e.g. see the tooltip in this TensorBoard example:
> 
> image
> 
> It seems possible to customize tooltip content by passing an extra `text` value into the `data` array passed to Plotly [here](https://github.com/mlflow/mlflow/blob/9ec25f157a548fa8e60ee13e3be5c3d2d513a3d5/mlflow/server/js/src/experiment-tracking/components/MetricsPlotView.js#L56), as per [the Plotly docs](https://plotly.com/javascript/hover-text-and-formatting/), I'd recommend giving that a try - thanks!

I think it could be better to with `hovertemplate` rather than with `text`, as described here -   https://plotly.com/javascript/hover-text-and-formatting/.

So after the last commit, it should look like:
smoothing

dbczumar:
@Valentyn1997 Thanks so much for addressing comments! @smurching and I will give this another pass within the next day.

smurching:
@Valentyn1997 pushed a small change to fix the text that shows up in the metrics popover when an individual point is clicked in the plot (just showing the original y-axis value instead of the smoothed value)

Before:
image

After:
image

With that change, LGTM!

dbczumar (approved):
LGTM! I will follow up with a PR shortly that introduces a shadow curve with the true metric value:

Screen Shot 2020-04-15 at 1 11 35 PM

dbczumar:
Thanks for this excellent contribution @Valentyn1997!

Valentyn1997:
You are welcome! I also thought about the shadow curve, but I think, it will make rendering of many metrics harder for browser. MetricPlotView is freezing already when there are more than 30-40 curves.

dbczumar:
@Valentyn1997 Thanks for raising that concern - it would be awesome to investigate approaches for improving the performance here. Hopefully this isn't a performance limitation of Plotly.

PR #2639, opened 2020-03-25
Unblock Travis tests

tomasatdatabricks:
2 test fixes:

1.  we were running large python tests instead of the small tests due to a typo. 
     renamed stage "small" to "Small" so that it's recognized in the shell script that runs tests.
2.  spark test were failing due to misconfigured rights. 
   It looks like that executable spark scripts such as spark-submit, spark-class etc get installed in two 
   places and only on of them is executable. Spark happens to use the location which is not 
   executable. This PR fixes it by chmoding all spark-* executables. This is a quick fix, we should look into it more later, but I think we should unblock tests first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
     # Nightly builds run the tests/examples end-to-end test suite.
     # Does not include Windows since Projects/Models are mostly not compatible with Windows as of now.
     - stage: Nightly
-    - stage: small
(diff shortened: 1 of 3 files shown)


aarondav (approved):
It's hacky, but it makes our tests pass, so let's use it for now!

harupy:
@tomasatdatabricks Thanks a lot for resolving this issue!

Issue #2603, opened 2020-03-18
[BUG] Flower Classification example not serving model


tfurmston:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md)
for information on what types of issues we address. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

Please fill in this template and do not delete it unless you are sure your issue is outside its scope.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS Sierra
- **MLflow installed from (source or binary)**:  
- **MLflow version (run ``mlflow --version``)**: mlflow, version 1.7.0
- **Python version 3.5.6:
- **npm version, if running the dev UI**:
- **Exact command to reproduce**: mlflow models serve -m mlruns/0/ec90976429dc4c99b8785347e0f515a0/artifacts/model/   -p 54321

### Describe the problem
Describe the problem clearly here. Include descriptions of the expected behavior and the actual behavior.

I have been trying to run the flower_classifier example. 

Firstly, the command `mlflow run examples/flower_classifier` fails. I believe this is because the `conda.yaml` doesn't pin the version of `tensorflow-mkl` and so tensorflow 2 is installed and the tutorial is written for tensorflow 1. I have fixed this issue by pinning the version as follows: `tensorflow==1.13.1`. Training is now working and completes successfully. 

I now try to run the serving of the model with the command, `mlflow models serve -m mlruns/0/ec90976429dc4c99b8785347e0f515a0/artifacts/model/   -p 54321`. 

I get the following error:

`
mlflow models serve -m mlruns/0/ec90976429dc4c99b8785347e0f515a0/artifacts/model/ -p 54321

2020/03/18 16:09:22 INFO mlflow.models.cli: Selected backend for flavor 'python_function'
2020/03/18 16:09:23 INFO mlflow.pyfunc.backend: === Running command 'source /miniconda3/bin/../etc/profile.d/conda.sh && conda activate mlflow-7e358be9432b86abb8743e4042a7cbdfef5d583d 1>&2 && gunicorn --timeout=60 -b 127.0.0.1:54321 -w 1 ${GUNICORN_CMD_ARGS} -- mlflow.pyfunc.scoring_server.wsgi:app'
[...]

PR #2646, opened 2020-03-27
Update _mlflow_conda_env to add pip when necessary

tfurmston:

I have updated _mlflow_conda_env to automatically add pip to the conda dependencies when there exist pip dependencies but pip is not specified. 

This PR addresses one of the issues found here, [https://github.com/mlflow/mlflow/issues/2603](https://github.com/mlflow/mlflow/issues/2603)

## How is this patch tested?

Additional unit tests have been added.

- [ X ] No. You can skip the rest of this section.

It's fixes a bug in the automatic generation of dependencies

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ X ] Python

- [ X ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes

### DIFF
--- a/mlflow/utils/environment.py
+++ b/mlflow/utils/environment.py
@@ -19,16 +19,22 @@ def _mlflow_conda_env(path=None, additional_conda_deps=None, additional_pip_deps
                  in dictionary format.
     :param additional_conda_deps: List of additional conda dependencies passed as strings.
     :param additional_pip_deps: List of additional pip dependencies passed as strings.
-    :param additional_channels: List of additional conda channels to search when resolving packages.
+    :param additional_conda_channels: List of additional conda channels to search when resolving packages.
     :return: ``None`` if ``path`` is specified. Otherwise, the a dictionary representation of the
              Conda environment.
     """
-    env = yaml.safe_load(_conda_header)
-    env["dependencies"] = ["python={}".format(PYTHON_VERSION)]
     pip_deps = (["mlflow"] if install_mlflow else []) + (
         additional_pip_deps if additional_pip_deps else [])
-    if additional_conda_deps is not None:
-        env["dependencies"] += additional_conda_deps
+
+    pip_not_specified = (additional_conda_deps is None
+                         or not [i for i in additional_conda_deps if "pip==" in i or "pip" == i])
+    conda_deps = (additional_conda_deps if additional_conda_deps else []) + (
+        ["pip"] if pip_deps and pip_not_specified else [])
+
+    env = yaml.safe_load(_conda_header)
+    env["dependencies"] = ["python={}".format(PYTHON_VERSION)]
+    if conda_deps is not None:
+        env["dependencies"] += conda_deps
     env["dependencies"].append({"pip": pip_deps})
     if additional_conda_channels is not None:
         env["channels"] += additional_conda_channels
--- a/mlflow/utils/environment.py
+++ b/mlflow/utils/environment.py
@@ -19,7 +19,8 @@ def _mlflow_conda_env(path=None, additional_conda_deps=None, additional_pip_deps
                  in dictionary format.
     :param additional_conda_deps: List of additional conda dependencies passed as strings.
     :param additional_pip_deps: List of additional pip dependencies passed as strings.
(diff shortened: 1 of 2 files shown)


mlflow/utils/environment.py
```diff
@@ -19,16 +19,23 @@ def _mlflow_conda_env(path=None, additional_conda_deps=None, additional_pip_deps
...
     """
-    env = yaml.safe_load(_conda_header)
-    env["dependencies"] = ["python={}".format(PYTHON_VERSION)]
     pip_deps = (["mlflow"] if install_mlflow else []) + (
         additional_pip_deps if additional_pip_deps else [])
-    if additional_conda_deps is not None:
-        env["dependencies"] += additional_conda_deps
+
+    pip_not_specified = (additional_conda_deps is None
+                         or not [i for i in additional_conda_deps if "pip==" in i or "pip" == i])
```
dbczumar: @tfurmston I'm not sure that this logic accounts for the full suite of possible version specification formats used in a conda env file (e.g., `pip>=`, `pip<=`, `pip20.0.02|20.0.2*`). Other examples for `numpy` are listed here (these apply to all packages): https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html

Given that it's difficult to accurately determine whether `pip` is already included in a given environment, I think it would be safer to either:

1. Revert changes to this method and update each of the module-specific `DEFAULT_CONDA_ENV`s to include the `pip` dependency

or

2. Always add an unversioned `pip` dependency to the conda environment, regardless of whether or not `pip` is already specified in the list of `additional_pip_deps` passed to this method. In fact, we could update the docs to explicitly indicate that `pip` is automatically added as a dependency if `additional_pip_deps` is non-empty and therefore should not be specified within `additional_pip_deps`. 

I'm in favor of option 2. cc'ing @tomasatdatabricks for an opinion here.

tfurmston: Um, that is a good point. I had not considered those cases. 

Yeah, in that case it does seem like it would be brittle to handle all of these different cases. 

Personally, I would also prefer option 2, but am very happy to do whatever you and @tomasatdatabricks think is best. :)

tfurmston: I suppose an alternative would be to follow the approach from conda for this type of check?

See here, [https://github.com/conda/conda/blob/master/conda_env/env.py#L54](https://github.com/conda/conda/blob/master/conda_env/env.py#L54)

What do you think?

tomasatdatabricks: I also like option 2. Adding dependency on pip without version is safe even if pip is already included in dependencies with or without version.

The conda code looks ok but I don't think we need it, better to keep this simple.

tfurmston: ok, no problem. I've made the change. 

My last point was was about why this functionality in conda is not actually being called. Part of that functionality seems to address the same point as this MR, but obviously for conda in general. 

Maybe I am overthinking it, but I just thought it was worth highlighting.

tomasatdatabricks: Ah I see, yeah that is a good question. Maybe you can file a ticket with conda as well. 

But I think adding unversioned pip is a good idea regardless. At the very least it will get rid of the warning.

tfurmston: ok, cool. That makes sense. 

Thanks

tomasatdatabricks (commented):
Looks good! Can you also document the new behavior in the docstring for _mlflow_conda_env?

tfurmston:
great, thanks!

I've updated the docstring.

tfurmston:
Hi, 

Is there a way to retry the CI? This error seems to be just a networking issue from what I can see.

Thanks

tomasatdatabricks:
I restarted the tests.

tomasatdatabricks (approved):
lgtm, thanks @tfurmston

tfurmston:
great, thanks!

Issue #2263, opened 2020-01-04
[FR] Saving torchscript models


hhsecond:
## Describe the proposal
Option to save torchscript model using `torch.jit.save` instead of `torch.save` which enables the deployment toolkits to pickup the optimized torchscript model for production

### Motivation
Mlflow currently doesn't distinguish between native pytorch model (subclass of `torch.nn.Module`) and torchscript model (subclass of `torch.nn.ScriptModule`). This is crucial since the production deployment system works efficiently on models saved as torchscript models (using `torch.jit.save`)
This is not possible to achieve with the current `pytorch.save_model` since we use `torch.save` internally. 

### Proposed Changes
I have two proposals in mind after talking to @mateiz in slack

1. Use the existing `pytorch.save_model` itself with an argument user can control to tell mlflow that whether it's a torchscript model or a `torch.nn.Module` model (or we can check the type of the incoming model and decide which flavor but explicit is better than implicit, I guess). The only constraint here is the same `pytorch` file will now have to save two flavors -> `pytorch` or `torchscript`. This sort of doesn't follow the MLFlow API design where every flavor gets one `save`, `load` and `log` function. 
2. Use another flavor and call it as `torchscript` and make a module for that. The API here becomes `mlflow.torchscript`

I have coded up a draft form of the second approach [here](https://github.com/hhsecond/mlflow-redisai/blob/master/mlflow_redisai/torchscript.py) but looking for suggestions before I make a PR for this to mlflow

smurching:
@hhsecond do you know what the relationship is between Torchscript & native PyTorch models & how Torchscript is commonly used? Just skimmed through the [Torchscript tutorial](https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html), which suggests that Torchscript is an IR for native PyTorch models that's most-often used to score PyTorch models in non-Python contexts, rather than load them back for scoring in Python.

If that's the case, then I think a variant of option 1) may make more sense, where we call both `torch.save` and `torch.jit.save` in `pytorch.save_model` to generate both `torchscript` & `pytorch` model flavors.

Will give this a more detailed look tomorrow, but if you have any pointers to docs/explanations on this topic that'd be welcome - thanks :)

hhsecond:
@smurching Your understanding is right. There are two decisions I am not sure about if we are letting  `pytorch.save_model` saves both flavors.
[...]

PR #2662, opened 2020-04-03
[WIP] Torchscript as another flavor

hhsecond:
Adding torchscript as another flavor that exposes necessary functionalities especially `log_model`, `save_model`, `load_model` and `_load_pyfunc`. There have been a discussion about whether torchscript needs another flavor or it just has to be an optional format under the current pytorch flavor. But finally, we agreed to keep it as another flavor. More details can be found at the [issue page](https://github.com/mlflow/mlflow/issues/2263)
## How is this patch tested?

(Details)

Introduces torchscript as another flavor for exporting pytorch models for production (nonpython) runtime

- [ ] UI
- [x] CLI
- [x] API
- [ ] REST-API
- [x] Examples
- [x] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### DIFF
--- a/examples/torchscript/MLproject
+++ b/examples/torchscript/MLproject
@@ -1,0 +1,25 @@
+name: pytorch_tutorial
+
+conda_env: conda.yaml
+
+entry_points:
+  main:
+    parameters:
+      batch-size: {type: int, default: 16}
+      test-batch-size: {type: int, default: 1000}
+      epochs: {type: int, default: 10}
+      lr: {type: float, default: 0.01}
+      momentum: {type: float, default: 0.5}
+      enable-cuda: {type: string, default: 'False'}
+      seed: {type: int, default: 5}
+      log-interval: {type: int, default: 100}
+    command: |
+          python mnist_tensorboard_artifact.py \
+            --batch-size {batch-size} \
+            --test-batch-size {test-batch-size} \
+            --epochs {epochs} \
+            --lr {lr} \
+            --momentum {momentum} \
+            --enable-cuda {enable-cuda} \
+            --seed {seed} \
+            --log-interval {log-interval}
--- a/examples/torchscript/mnist_tensorboard_artifact.py
+++ b/examples/torchscript/mnist_tensorboard_artifact.py
@@ -156,8 +156,7 @@ def log_scalar(name, value, step):
 
 with mlflow.start_run() as run:
     # Log our parameters into mlflow
-    for key, value in vars(args).items():
-        mlflow.log_param(key, value)
+    mlflow.log_params(vars(args))
 
     # Create a SummaryWriter to write TensorBoard events locally
     output_dir = dirpath = tempfile.mkdtemp()
--- a/examples/torchscript/mnist_tensorboard_artifact.py
+++ b/examples/torchscript/mnist_tensorboard_artifact.py
@@ -24,8 +24,9 @@ from torchvision import datasets, transforms
 from torch.autograd import Variable
 from tensorboardX import SummaryWriter
 
+
 # Command-line arguments
-parser = argparse.ArgumentParser(description='PyTorch MNIST Example')
+parser = argparse.ArgumentParser(description='Torchscript MNIST Example')
 parser.add_argument('--batch-size', type=int, default=64, metavar='N',
                     help='input batch size for training (default: 64)')
 parser.add_argument('--test-batch-size', type=int, default=1000, metavar='N',
@@ -102,8 +103,6 @@ if args.cuda:
     model.cuda()
 
 optimizer = optim.SGD(model.parameters(), lr=args.lr, momentum=args.momentum)
-
-writer = None # Will be used to write TensorBoard events
 
 
 def train(epoch):
--- a/examples/torchscript/mnist_tensorboard_artifact.py
+++ b/examples/torchscript/mnist_tensorboard_artifact.py
@@ -1,0 +1,180 @@
+#
+# Trains an MNIST digit recognizer using PyTorch, and uses tensorboardX to log training metrics
+# and weights in TensorBoard event format to the MLflow run's artifact directory. This stores the
+# TensorBoard events in MLflow for later access using the TensorBoard command line tool.
+#
+# NOTE: This example requires you to first install PyTorch (using the instructions at pytorch.org)
+#       and tensorboardX (using pip install tensorboardX).
+#
+# Code based on https://github.com/lanpa/tensorboard-pytorch-examples/blob/master/mnist/main.py.
+#
+
+from __future__ import print_function
+import argparse
+import os
+import mlflow
+import mlflow.pytorch
+import mlflow.torchscript
+import tempfile
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+from torchvision import datasets, transforms
+from torch.autograd import Variable
+from tensorboardX import SummaryWriter
+
+# Command-line arguments
+parser = argparse.ArgumentParser(description='PyTorch MNIST Example')
+parser.add_argument('--batch-size', type=int, default=64, metavar='N',
+                    help='input batch size for training (default: 64)')
+parser.add_argument('--test-batch-size', type=int, default=1000, metavar='N',
+                    help='input batch size for testing (default: 1000)')
+parser.add_argument('--epochs', type=int, default=10, metavar='N',
+                    help='number of epochs to train (default: 10)')
+parser.add_argument('--lr', type=float, default=0.01, metavar='LR',
+                    help='learning rate (default: 0.01)')
+parser.add_argument('--momentum', type=float, default=0.5, metavar='M',
+                    help='SGD momentum (default: 0.5)')
+parser.add_argument('--enable-cuda', type=str, choices=['True', 'False'], default='True',
+                    help='enables or disables CUDA training')
+parser.add_argument('--seed', type=int, default=1, metavar='S',
+                    help='random seed (default: 1)')
+parser.add_argument('--log-interval', type=int, default=100, metavar='N',
+                    help='how many batches to wait before logging training status')
+args = parser.parse_args()
+
+enable_cuda_flag = True if args.enable_cuda == 'True' else False
+
+args.cuda = enable_cuda_flag and torch.cuda.is_available()
+
+torch.manual_seed(args.seed)
+if args.cuda:
+    torch.cuda.manual_seed(args.seed)
+
+kwargs = {'num_workers': 1, 'pin_memory': True} if args.cuda else {}
+train_loader = torch.utils.data.DataLoader(
+    datasets.MNIST('../data', train=True, download=True,
+                   transform=transforms.Compose([
+                       transforms.ToTensor(),
+                       transforms.Normalize((0.1307,), (0.3081,))
+                   ])),
+    batch_size=args.batch_size, shuffle=True, **kwargs)
+test_loader = torch.utils.data.DataLoader(
+    datasets.MNIST('../data', train=False, transform=transforms.Compose([
+                       transforms.ToTensor(),
+                       transforms.Normalize((0.1307,), (0.3081,))
+                   ])),
+    batch_size=args.test_batch_size, shuffle=True, **kwargs)
+
+
+class Net(nn.Module):
+    def __init__(self):
+        super(Net, self).__init__()
+        self.conv1 = nn.Conv2d(1, 10, kernel_size=5)
+        self.conv2 = nn.Conv2d(10, 20, kernel_size=5)
+        self.conv2_drop = nn.Dropout2d()
+        self.fc1 = nn.Linear(320, 50)
+        self.fc2 = nn.Linear(50, 10)
+
+    def forward(self, x):
+        x = F.relu(F.max_pool2d(self.conv1(x), 2))
+        x = F.relu(F.max_pool2d(self.conv2_drop(self.conv2(x)), 2))
+        x = x.view(-1, 320)
+        x = F.relu(self.fc1(x))
+        x = F.dropout(x, training=self.training)
+        x = self.fc2(x)
+        return F.log_softmax(x, dim=0)
+
+    def log_weights(self, step):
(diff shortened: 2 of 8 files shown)


examples/torchscript/MLproject
```diff
@@ -0,0 +1,25 @@
+name: pytorch_tutorial
```
smurching, line 1: Could we also add a README to the examples/torchscript directory providing an overview of the example / how to run it? It'd also be nice to link to the new example from https://github.com/mlflow/mlflow/blob/master/docs/source/tutorials-and-examples/index.rst

examples/torchscript/mnist_tensorboard_artifact.py
```diff
@@ -0,0 +1,180 @@
...
+
+def log_scalar(name, value, step):
+    """Log a scalar value to both MLflow and TensorBoard"""
+    writer.add_scalar(name, value, step)
+    mlflow.log_metric(name, value)
+
+
+with mlflow.start_run() as run:
+    # Log our parameters into mlflow
+    for key, value in vars(args).items():
```
smurching, line 160: Could we use `mlflow.log_params(var(args))` to log all these params in a single line / API request? Just to encourage best practices for efficient logging in our examples

mlflow/torchscript.py
```diff
@@ -0,0 +1,108 @@
...
+def get_default_conda_env():
+    """
+    :return: The default Conda environment for MLflow Models produced by calls to
+             :func:`save_model()` and :func:`log_model()`.
+    """
+    import torch
+
+    return _mlflow_conda_env(
+        additional_conda_deps=[
+            "pytorch={}".format(torch.__version__),
```
smurching, line 30: Is it sufficient to install just `pytorch`? From the [pytorch homepage](https://pytorch.org/) it seems for conda installations we need to also install `torchvision`, and if on linux or windows to install `cpuonly` as well

mlflow/torchscript.py
```diff
@@ -0,0 +1,108 @@
...
+    conda_env_subpath = "conda.yaml"
+    if conda_env is None:
+        conda_env = get_default_conda_env()
+    elif not isinstance(conda_env, dict):
+        with open(conda_env, "r") as f:
+            conda_env = yaml.safe_load(f)
+    with open(path / conda_env_subpath, "w") as f:
+        yaml.safe_dump(conda_env, stream=f, default_flow_style=False)
+    mlflow_model.add_flavor(FLAVOR_NAME, pytorch_version=torch.__version__)
+    pyfunc.add_to_model(mlflow_model, loader_module="mlflow.pytorch", env=conda_env_subpath)
```
smurching, line 62:
```suggestion
    pyfunc.add_to_model(mlflow_model, loader_module="mlflow.torchscript", env=conda_env_subpath)
```

mlflow/torchscript.py
```diff
@@ -0,0 +1,108 @@
...
+
+def log_model(model, artifact_path, conda_env=None, registered_model_name=None, **kwargs):
+    Model.log(artifact_path=artifact_path, flavor=mlflow.torchscript, model=model,
+              conda_env=conda_env, registered_model_name=registered_model_name, **kwargs)
+
+
+def save_model(model, path, conda_env=None, mlflow_model=Model(), **kwargs):
+    if not hasattr(model, 'state_dict') or not hasattr(model, 'save'):
+        # If it walks like a duck and it quacks like a duck, then it must be a duck
+        raise TypeError("Argument 'model' should be a TorchScript model")
```
smurching, line 45: Is there a type we can explicitly check against, vs checking attributes of the class? Just in case e.g. similar attributes (e.g. `state_dict`) are added to other classes in Pytorch?

mlflow/torchscript.py
smurching: e.g. in the [pytorch flavor](https://github.com/mlflow/mlflow/blob/7969f9140874e0644770f255d106f7a729150aad/mlflow/pytorch/__init__.py#L49-L51), we also install `torchvision`

smurching (commented):
Some small comments but largely looks reasonable, thanks @hhsecond!

smurching (commented):
Of course note that we should also add tests, e.g. see https://github.com/mlflow/mlflow/blob/master/tests/pytorch/test_pytorch_model_export.py as a reference

mlflow/torchscript.py
hhsecond: There are multiple types associated with the `jit`ed models (it's possible with `script` ing, `trace`ing and to make the things complicated, users can also make `script`ed python functions)

mlflow/torchscript.py
hhsecond: `torchvision` is required only for `vision` tasks, so it's not a mandatory requirement for pytorch
`cpuonly` is required to choose the non-CUDA version of pytorch. Perhaps it makes more sense to keep the CPU only version default. I'll add that

examples/torchscript/mnist_tensorboard_artifact.py
hhsecond: Sure, I have copied this script from [the pytorch example](https://github.com/mlflow/mlflow/blob/d8e6a8f07ce8fa940e26a68f0fe19f9fda1fb1da/examples/pytorch/mnist_tensorboard_artifact.py#L150). I am just wondering if the author had a reason to do it this way

examples/torchscript/MLproject
hhsecond: Will do

mlflow/torchscript.py
hhsecond: I think we have a relatively (than I thought) easier way to check this. Updated the code

examples/torchscript/mnist_tensorboard_artifact.py
smurching: Got it sounds good! Yeah I actually wonder if we could just update the pytorch example to log torchscript models, vs introducing a new example snippet that's a copy of the pytorch snippet (because we do want to encourage pytorch users to log torchscript models)

mlflow/torchscript.py
smurching: Sounds good, yeah let me know if you hit issues related to this! How expensive of a dependency to install is `torchvision`? Given that we install it in the conda env for the `pytorch` flavor, it would be nice to install it here as well for consistency (otherwise users may be confused that they need to specify the extra `torchvision` dependency when using `mlflow.torchvision.log_model` but not when using `mlflow.pytorch.log_model`)

mlflow/torchscript.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def predict(self, data, device='cpu'):
+        if not isinstance(data, pd.DataFrame):
+            raise TypeError("Input data should be pandas.DataFrame")
+        self.torchscript_model.to(device)
+        self.torchscript_model.eval()
+        with torch.no_grad():
+            input_tensor = torch.from_numpy(data.values.astype(np.float32)).to(device)
+            preds = self.torchscript_model(input_tensor)
+            if not isinstance(preds, torch.Tensor):
```
smurching, line 120: BTW, is there any way to validate that the Pytorch model produces a single output tensor in `save_model` and raise an exception there? It seems like it could be a bit frustrating for the user to save the model only to find that at inference time that it cannot be scored as a pyfunc. Adding such validation in `save_model` has the downside that you can't e.g. run code like:

```
# Save model
mlflow.torchscript.save_model(model_with_2d_output, path)
# Load model back as a torch.jit model rather than as a pyfunc
loaded_model = mlflow.torchscript.load_model(path)
```

for users who aren't interested in the pyfunc flavor. But I think it'd be better to be consistent here / fail at saving time if we can't use the model with the pyfunc flavor, given that much of MLflow's built-in serving tools rely on the pyfunc flavor

mlflow/torchscript.py
smurching: Hm sorry, I think because of the force push I can't see the commits where you added the new check - looks like we're still checking attributes of the object rather than its type?

smurching (commented):
Mostly looks good, thanks for adding the tests @hhsecond! Left some more comments

examples/torchscript/mnist_tensorboard_artifact.py
hhsecond: That's a fair reasoning. Let me know what's the final decision and I'll change it accordingly. In my opinion, you need to show both (you can put both in the same example) but then that would negate the structuring of the project, which is one folder for every flavour.

mlflow/torchscript.py
hhsecond: Fair enough

mlflow/torchscript.py
hhsecond: IMO, that's a hard (and avoidable) restriction to put on users.
In any case, I don't know a way to check this reliably. What we could probably do is to mention this in the docstring explicitly as `mlflow.pytorch` does

mlflow/torchscript.py
```diff
@@ -0,0 +1,313 @@
...
+            y_pred = model(x_data)
+            ...
+        # Save the model
+        with mlflow.start_run() as run:
+            mlflow.log_param("epochs", 500)
+            scripted_model = torch.jit.script(model)
+            mlflow.torchscript.save_model(scripted_model, pytorch_model_path)
+    """
+    if not hasattr(model, 'state_dict') or not hasattr(model, 'save'):
+        # If it walks like a duck and it quacks like a duck, then it must be a duck
```
smurching, line 207: Could we replace this with:

```
if not isinstance(model, torch.jit.ScriptModule) and not isinstance(model, torch.jit.ScriptFunction):
            raise TypeError("Argument 'model' should be a TorchScript model. Use torch.jit.script() to convert your model into a TorchScript model before passing it to this function.")
```

It looks like `torch.jit.script`'s documented behavior is to return either a `torch.jit.ScriptModule` or `torch.jit.ScriptFunction`: https://pytorch.org/docs/stable/jit.html#torch.jit.script

tests/torchscript/test_model_export.py
```diff
@@ -0,0 +1,403 @@
+import torch
```
smurching, line 1: Read through a couple of Pytorch/Torchscript references:
* https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html
* https://discuss.pytorch.org/t/when-should-i-use-tracing-rather-than-scripting/53883

It seems we should also have tests verifying that we can save/load TorchScript models created via `torch.script.trace`?

mlflow/torchscript.py
```diff
@@ -0,0 +1,313 @@
...
+            raise TypeError("Input data should be pandas.DataFrame")
+        self.torchscript_model.to(device)
+        self.torchscript_model.eval()
+        with torch.no_grad():
+            input_tensor = torch.from_numpy(data.values.astype(np.float32)).to(device)
+            preds = self.torchscript_model(input_tensor)
+            if not isinstance(preds, torch.Tensor):
+                raise TypeError("Expected PyTorch model to output a single output tensor, "
+                                "but got output of type '{}'".format(type(preds)))
+            predicted = pd.DataFrame(preds.numpy())
```
smurching, line 311: BTW, what do you think about this output format? It matches what the Pytorch flavor produces, but is a bit odd - we assume that the output tensor is one-or-two-dimensional and create a DataFrame where individual elements of each tensor are put into separate columns:

```
>>> oned_tensor
array([1, 2, 3])
>>> twod_tensor
array([[1, 2, 3],
       [4, 5, 6]])
>>> pd.DataFrame(oned_tensor)
   0
0  1
1  2
2  3
>>> pd.DataFrame(twod_tensor
... )
   0  1  2
0  1  2  3
1  4  5  6
>>> threed
array([[[1, 2, 3],
        [4, 5, 6]],

       [[1, 2, 3],
        [4, 5, 6]]])
>>> pd.DataFrame(threed)
  File "/Users/sid.murching/miniconda3/envs/mlflow-python3/lib/python3.6/site-packages/pandas/core/frame.py", line 361, in __init__
    copy=copy)
  File "/Users/sid.murching/miniconda3/envs/mlflow-python3/lib/python3.6/site-packages/pandas/core/frame.py", line 513, in _init_ndarray
    values = _prep_ndarray(values, copy=copy)
  File "/Users/sid.murching/miniconda3/envs/mlflow-python3/lib/python3.6/site-packages/pandas/core/frame.py", line 6257, in _prep_ndarray
    raise ValueError('Must pass 2-d input')
ValueError: Must pass 2-d input
```

Packaging the output into a DataFrame has the advantage that we can use it with MLflow's ``mlflow.pyfunc.spark_udf`` API, which expects the pyfunc to produce DataFrame-able output, but I wonder if we should just return a numpy array here to avoid constraining what the torch model can produce as output.

smurching: On the other hand, this approach is consistent with `mlflow.pytorch` and we could argue models usually produce 1D output (for regression) or 2D output (for classification when outputting probabilities across classes), in which case this approach is fine. Let me ask some folks more familiar with Pytorch use cases for opinions - currently I'm thinking this is fine as is, but if we do e.g. extend MLflow's spark UDF to better support multidimensional outputs, maybe we should consider the numpy array approach

mlflow/torchscript.py
hhsecond: Done

tests/torchscript/test_model_export.py
hhsecond: Done

mlflow/torchscript.py
hhsecond: Right, It is similar to "allowing users to pass only one input" scenario. My experience is mostly working with DL algorithms and hence my viewpoint is that restricting any of the DL frameworks to output DF objects is not nice. But then, that needs to be changed across the MLflow project. Let me know your decision.

HenryMaguire:
Hi, does anyone know what the status is with this PR?

harupy:
Closing this because `log_model`, `save_model`, and `load_model` now support logging/loading script models.

PR #2665, opened 2020-04-03
Allow numerical parameters with values >= 1'000 in R mlflow::mlflow_run()

### LINKED ISSUE #2664: [BUG] Numeric parameter with value >= 1'000 can't be supplied in R mlflow::mlflow_run()
### System information

- **Have I written custom code (as opposed to using a stock example script provided in MLflow): Yes.**
- macOS 10.15.3.
- **MLflow installed from source (GitHub)**:
- **MLflow version 1.7.3 (dev)**:
- **Python version: 3.6**:

```r
mlflow::mlflow_run('.', entry_point = 'R/script.R', parameters = list(
  large_dbl = 1000000,
  other = 3
))
```

### Describe the problem

It's impossible to supply a parameter value larger or equal to 1'000 by default in R mlflow when running `mlflow::mlflow_run()` because the scientific formatting will kick in and the number won't be recognised as numeric anymore somewhere downstream. The root of the problem can be observed in the R console:

```r
10000000
#> [1] 1e+07
```

With the above command, I get the following error:

```
2020/04/03 18:03:13 INFO mlflow.projects: === Created directory /var/folders/6g/bg5mcmsj7s51r33snbxhm1400000gn/T/tmp1zqfgvss for downloading remote URIs passed to arguments of type 'path' ===
2020/04/03 18:03:13 INFO mlflow.projects: === Running command 'source /Users/lorenz/opt/miniconda3/bin/../etc/profile.d/conda.sh && conda activate mlflow-da39a3ee5e6b4b0d3255bfef95601890afd80709 1>&2 && Rscript -e "mlflow::mlflow_source('R/script.R')" --args --large_dbl 1e+06 --other 3' in run with ID '6e1420faad164528b654a16b08c2da80' ===

Error: Provided value for `large_dbl` cannot be casted to type numeric: `x` cannot be casted to a double vector.
```

This is what happens: When any number is supplied to `mlflow::mlflow_run()`, it may be displayed in scientific notation, remains of class numeric. Then, to prepare the call to `mlflow::mlflow_source()`, it's `paste0()`d and hence converted to character. Like all other numbers. Then, it is passed to `mlflow::mlflow_source()` as command line arguments. The difference to "normal" numbers is that in `parse_command_line()`,where `yaml::yaml.load()` is used, strings like "3" will be turned into numeric again, but not scientific notation. I see two solutions:

- convert all numeric input to non-scientific before passing it to `mlflow::mlflow_source()`. Drawback: If a user supplies 1e6, it will be logged  as 1000000. This may take a lot of space, in particular in the UI when comparing many runs with big numers.

- recover scientific notation in `parse_command_line()`. Drawback: We might turn  a parameter that was supplied as string into numeric, i.e. when you supplied  "1e3" initially, this will be turned into 1000 and the user has to see how to  turn that back into 1e3.

I think the second approach is preferred and I submited #2665 to implement it.
### Code to reproduce issue

Run the above command in [this](https://github.com/lorenzwalthert/mlflow-r-casting-bug/tree/master/R) minimally reproducible git repo, file `R/runner.R`.

lorenzwalthert:

Convert all numeric parameter values to non-scientific before passing it to `mlflow::mlflow_source()` in `mlflow::mlflow_run()` as suggested in #2664 to enable parameter values of 1'000 or larger. Closes #2664. 

## How is this patch tested?

Added relevant unit tests. I first add a test to show how the current version of mlflow [fails](https://travis-ci.org/github/lorenzwalthert/mlflow/builds/670680510) to handle the case and then commit the fix to make tests pass.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [x] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/R/mlflow/R/project-run.R
+++ b/mlflow/R/mlflow/R/project-run.R
@@ -53,7 +53,7 @@ mlflow_run <- function(uri = ".", entry_point = NULL, version = NULL, parameters
     uri <- fs::path_expand(uri)
 
   param_list <- if (!is.null(parameters)) parameters %>%
-    purrr::imap_chr(~ paste0(.y, "=", .x)) %>%
+    purrr::imap_chr(~ paste0(.y, "=", format(.x, scientific = FALSE))) %>%
     purrr::reduce(~ mlflow_cli_param(.x, "--param-list", .y), .init = list())
 
   args <- list(uri) %>%
--- a/mlflow/R/mlflow/tests/testthat/test-run.R
+++ b/mlflow/R/mlflow/tests/testthat/test-run.R
@@ -46,3 +46,18 @@ test_that("mlflow run uses active experiment if not specified", {
       })
   })
 })
+
+
+test_that("mlflow_run passes all numbers as non-scientific", {
+  # we can only be sure conversion is actively avoided
+  # if default formatting turns into scientific.
+  expect_equal(as.character(10e4), "1e+05")
+  with_mock(.env = "mlflow", mlflow_cli = function(...){
+    args <- c(...)
+    expect_equal(sum("scientific=100000" == args), 1)
+    expect_equal(sum("non_scientific=30000" == args), 1)
+    list(stderr = "=== Run (ID '48734e7e2e8f44228a11c0c2cbcdc8b0') succeeded ===")
+  }, {
+    mlflow_run("project", parameters = c(scientific = 10e4, non_scientific = 30000))
+  })
+})


tomasatdatabricks (approved):
lgtm, thanks @lorenzwalthert !

PR #2670, opened 2020-04-04
Fix bug causing erroneous deletion of TensorBoard logging directory

### LINKED ISSUE #2644: [BUG] MLflow tf.autolog() deletes TensorBoard records if both services are used.
### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Google Colab
- **MLflow installed from (source or binary)**: source
- **MLflow version (run ``mlflow --version``)**:1.7.2
- **Python version**: 3.6.9
- **Exact command to reproduce**: https://colab.research.google.com/drive/1-bn6HGKoXEg5qGzWjvTyW34fBRgA6TJv

### Describe the problem
Whenever TensorBoard and MLflow are used in parallel, a MLflow clean up hook deletes all metrics logged to TensorBoard at the end of training. See the attached Colab for an example.

### Code to reproduce issue
https://colab.research.google.com/drive/1-bn6HGKoXEg5qGzWjvTyW34fBRgA6TJv

dbczumar:

This PR fixes #2644 by introducing a conditional check to prevent deleting the destination directory for TensorBoard logs if this directory was specified by an explicit TensorBoard callback passed to `model.fit()` or `model.fit()` generator.

## How is this patch tested?

This PR includes unit tests for TensorFlow 1.x and TensorFlow 2.0. Further, I manually verified that the workflow included in #2644 (https://colab.research.google.com/drive/1-bn6HGKoXEg5qGzWjvTyW34fBRgA6TJv) now produces TensorBoard logs that persist after the `model.fit()` call completes with autologging enabled.

## Release Notes

Fixes an autologging bug that caused accidental deletion of the TensorBoard logging directory when users specified an explicit TensorBoard callback as part of `model.fit()` or `model.fit_generator()`.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [X] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/tests/tensorflow_autolog/test_tensorflow2_autolog.py
+++ b/tests/tensorflow_autolog/test_tensorflow2_autolog.py
@@ -48,12 +48,6 @@ def manual_run(request, tracking_uri_mock):
     mlflow.end_run()
 
 
-@pytest.fixture(scope='function', autouse=True)
-def end_run_if_necessary():
-    yield
-    mlflow.end_run()
-
-
 def create_tf_keras_model():
     model = tf.keras.Sequential()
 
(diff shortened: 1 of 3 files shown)


tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -412,3 +412,59 @@ def test_duplicate_autolog_second_overrides(duplicate_autolog_tf_estimator_run):
     client = mlflow.tracking.MlflowClient()
     metrics = client.get_metric_history(duplicate_autolog_tf_estimator_run.info.run_id, 'loss')
     assert all((x.step - 1) % 4 == 0 for x in metrics)
+
+
+@pytest.mark.large
```
juntai-zheng, line 417: Nit: I would move these tests a bit up in the file with the other `tf.keras` tests since the tests are split by `tf.estimator` and `tf.keras` models.

tests/tensorflow_autolog/test_tensorflow_autolog.py
```diff
@@ -436,3 +436,59 @@ def test_duplicate_autolog_second_overrides(duplicate_autolog_tf_estimator_run):
     client = mlflow.tracking.MlflowClient()
     metrics = client.get_metric_history(duplicate_autolog_tf_estimator_run.info.run_id, 'loss')
     assert all((x.step - 1) % 4 == 0 for x in metrics)
+
+
+@pytest.mark.large
```
juntai-zheng, line 441: Nit: same as last comment

juntai-zheng (approved):
Two tiny nits but otherwise LGTM!

dbczumar:
@juntai-zheng Thanks for the review! Pushed changes to move the test cases to a more appropriate location. Will plan to merge when tests pass!

tests/tensorflow_autolog/test_tensorflow_autolog.py
```diff
@@ -288,6 +288,62 @@ def test_tf_keras_autolog_non_early_stop_callback_no_log(tf_keras_random_data_ru
...
+            generator(), epochs=10, steps_per_epoch=1, callbacks=[tensorboard_callback])
+    else:
+        model.fit(data, labels, epochs=10, callbacks=[tensorboard_callback])
+
+    assert os.path.exists(tensorboard_callback_logging_dir_path)
+
+
+@pytest.mark.large
+@pytest.mark.parametrize('fit_variant', ['fit', 'fit_generator'])
+def test_tf_keras_autolog_logs_to_and_deletes_temporary_directory_when_tensorboard_callback_absent(
```
juntai-zheng, line 320: Seems like [tests are failing](https://travis-ci.org/github/mlflow/mlflow/jobs/671795781#L4273) because runs are not ended after the new tests. Adding the fixture `manual_run` to these tests might fix it.

tests/tensorflow_autolog/test_tensorflow_autolog.py
```diff
@@ -423,16 +480,10 @@ def test_tf_estimator_autolog_model_can_load_from_artifact(tf_estimator_random_d
                                          "/model", session)
 
 
-@pytest.fixture
-def duplicate_autolog_tf_estimator_run(tmpdir, manual_run, export):
-    mlflow.tensorflow.autolog(every_n_iter=23)  # 23 is prime; no false positives in test
-    run = tf_estimator_random_data_run(tmpdir, manual_run, export)
```
dbczumar, line 429: This call to a fixture as a function is invalid. Interestingly, `pytest` did not pick up this issue until implicit `end_run()` calls were added to the new `tf_keras` tests. I've removed this fixture since it's only used in a single test case, and I've modified the test case to call `autolog(every_n_iter=23)` so that the same state is achieved.

PR #2619, opened 2020-03-21
Add autologging support for Fastai

antoniomdk:
## What changes are proposed in this pull request?

This PR includes a new flavor for Fastai v1. It supports model exporting and autologging.
The autologging functionality keeps track of most used callbacks: EarlyStopping and OneCycleScheduler. 

It's also included an improvement of log_fn_args_as_params function that replaces the deprecated function getargspec in favor of getfullargspec which supports keyword-only parameters.

Closes #2087

> Note: Pyfunc wrapper only works for tabular learners

## How is this patch tested?

Some tests similar to the Keras ones have been written to ensure model exporting and autologging works in most cases.

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

Added support for Fastai models.

### What component(s) does this PR affect?

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [ ] Examples
- [] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

### How should the PR be classified in the release notes? Choose one:

- [] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [ ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [x] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #2689, opened 2020-04-11
Add Fastai flavor

### LINKED ISSUE #2087: [FR] Autologging functionality for PyTorch/fastai
## Describe the proposal

It'd be nice to add an mlflow.pytorch.autolog() API for automatically logging metrics, params & models generated via PyTorch. In particular, I'd like this to be incorporated with the fastai library (built on PyTorch), which also generates artifacts such as learning rate finder plots, and autolog those as well. 

### Motivation
PyTorch is heavily used in research for Deep Learning and is becoming very prevalent in industry as well with the fastai library. The fastai community has over 450 contributors, and it incorporates practical deep learning best practices + cutting edge algorithms in the library. In the fast.ai course that is run by USF, they use Excel spreadsheets to keep track of their runs. If we could just autolog everything once they import mlflow, it will help bring the adoption of MLflow to the fastai community. 

### Proposed Changes
See above - we propose adding a new mlflow.pytorch.autolog API

antoniomdk:

This PR includes a new flavor for Fastai v1. It supports model exporting.

It's also included an improvement of log_fn_args_as_params function that replaces the deprecated function getargspec in favor of getfullargspec which supports keyword-only parameters.

Closes #2087

> Note: Pyfunc wrapper only works for tabular learners

## How is this patch tested?

Some tests similar to the Keras ones have been written to ensure model exporting and autologging works in most cases.

Added support for Fastai models.

- [ ] UI
- [ ] CLI
- [x] API
- [ ] REST-API
- [ ] Examples
- [] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python

- [] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section


tests/fastai/test_fastai_model_export.py
```diff
@@ -0,0 +1,287 @@
...
+    model_path = _download_artifact_from_uri(artifact_uri=model_uri)
+    pyfunc_conf = _get_flavor_configuration(model_path=model_path, flavor_name=pyfunc.FLAVOR_NAME)
+    conda_env_path = os.path.join(model_path, pyfunc_conf[pyfunc.ENV])
+    with open(conda_env_path, "r") as f:
+        conda_env = yaml.safe_load(f)
+
+    assert conda_env == mlflow.fastai.get_default_conda_env()
+
+
+@pytest.mark.release
```
smurching, line 271: We tend not to run these tests at all, so I think it's fine to delete it - thanks!

mlflow/fastai.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class _FastaiModelWrapper:
+    def __init__(self, learner):
+        self.learner = learner
+
+    def predict(self, dataframe):
+        test_data = TabularList.from_df(dataframe, cont_names=self.learner.data.cont_names)
+        self.learner.data.add_test(test_data)
+        return self.learner.get_preds(DatasetType.Test)[0].numpy()
```
smurching, line 160: Thanks for opening a new PR! To continue the discussion from https://github.com/mlflow/mlflow/pull/2619#discussion_r406965943, maybe we ought to just return both the probabilities & labels given that that's what fastai returns / to reduce user confusion? I would imagine that for most use cases, a user would care about the labels rather than the probabilities, so seems nice to return the labels in addition to the probabilities for that reason as well .

 Also, do you know what happens in the regression case? Spent some time reading through the [fastai regression tutorial](https://docs.fast.ai/tutorial.data.html#A-regression-example), but couldn't quite find an example of using `get_preds` in the regression case - thanks!!

antoniomdk: For the regression case, it returns a tuple but the first tensor is the predictions and the second is the actual target (ground truth), so I think it's not useful for inference.

https://forums.fast.ai/t/tabular-regression-prediction-phase-returns-2-tensors/28158

The reason why I chose to return only probabilities is that the behavior changes between regression and classification as you can see.

Btw, I've created a small example for you to try:

https://gist.github.com/antoniomdk/1f2728ddcc32af5a715cd8a6a011015c

tests/fastai/test_fastai_model_export.py
antoniomdk: Done!

mlflow/fastai.py
smurching: Thanks! From what I understand (from adding a `help(learn.get_preds)` to the end of your example) it seems `get_preds` returns both predictions & targets (which are all 0 in the regression case) on the dataset:

```
get_preds(ds_type:fastai.basic_data.DatasetType=<DatasetType.Valid: 2>, activ:torch.nn.modules.module.Module=None, with_loss:bool=False, n_batch:Union[int, NoneType]=None, pbar:Union[fastprogress.fastprogress.ProgressBar, NoneType]=None) -> List[torch.Tensor] method of fastai.basic_train.Learner instance
    Return predictions and targets on `ds_type` dataset.
```

I wonder if we could actually return a Dataframe from this method containing `predictions` and `targets` columns, so that users can access whichever column they care about (e.g. directly get the targets if that's relevant for their use case, or get the raw predictions). Thoughts?

antoniomdk: We can always return a Dataframe with two columns because that's what ` learn.get_preds` returns. If we want to return always two columns we can do two things:

1. Return the data as it is
2. Return the data as it is for classification tasks and duplicate the first column for regression. In that way, the target vector contains predictions too.

IMO, the first option is better to maintain the consistency between the Learn method and the wrapper.

smurching: :+1: option 1 (return the data as it is) sounds good to me!

antoniomdk: I agree. I'll work on that during this weekend.

antoniomdk: I've changed the wrapper to return both arrays.

mlflow/fastai.py
```diff
@@ -157,7 +158,10 @@ def __init__(self, learner):
     def predict(self, dataframe):
         test_data = TabularList.from_df(dataframe, cont_names=self.learner.data.cont_names)
         self.learner.data.add_test(test_data)
-        return self.learner.get_preds(DatasetType.Test)[0].numpy()
+        preds = self.learner.get_preds(DatasetType.Test)
+        preds = [pred.numpy() for pred in preds]
+        preds = [pred if len(pred.shape) > 1 else pred.reshape((len(pred), 1)) for pred in preds]
+        return np.concatenate(preds, axis=1)
```
smurching, line 164: @antoniomdk thanks for the update! I was actually thinking here that we'd return a pandas DataFrame with explicit `prediction` and `target` columns, corresponding to the first & second elements of the tuple returned by `self.learner.get_preds`, respectively. I don't think we'd want to concatenate the predictions and targets together into a numpy single array, as this makes it hard for the user of the model to interpret the result (they need to know how many output classes the model produces & write logic to parse out the predictions/targets separately). 

For example, given a binary classification model with:

```
>>> preds, targets = self.learner.get_preds(DatasetType.Test)
>>> print(preds)
[[0.1 0.9]
 [0.9 0.1]]
>>> preds.shape
(2, 2)
>>> print(targets)
[0 1]
>>> targets.shape
(2,)
```

We'd want to return a Pandas dataframe where:
* The `prediction` column contains 1D arrays with class probabilities, e.g. in the example above the dataframe would contain two rows - one row with `prediction` value `[0.1, 0.9]` and one row with value `[0.9, 0.1]`. 
* The `target` column contains scalar target values, e.g. in the example above the target column would contain values `[0, 1]`

Note that packaging the result in this way is actually important in order for it to work properly with MLflow's Spark UDF functionality ([link](https://github.com/mlflow/mlflow/blob/402381702e8c4b8383a1910062dc1907761c05ac/mlflow/pyfunc/__init__.py#L453-L455)), which will attempt to convert the result of `predict` directly into a Pandas DataFrame if it isn't already one. In the current implementation, this would result in:

```
>>> concatenated = np.concatenate([preds, targets.reshape(len(targets), 1)], axis=1)
>>> concatenated
array([[0.1, 0.9, 0. ],
       [0.9, 0.1, 1. ]])
>>> pd.DataFrame(data=concatenated)
     0    1    2
0  0.1  0.9  0.0
1  0.9  0.1  1.0
```

Thus the current implementation would cause the Spark UDF to return a DataFrame with three columns for the above example, rather than one array-valued column containing class probabilities & another scalar column containing the actual targets.

antoniomdk: Ah okay, I got it wrong.  Thanks for the clarification.  I wasn't sure how it would work if the wrapper returns a non-numerical column (dtype=object). I will change that, I just hope it won't cause any serialization issue.

mlflow/fastai.py
```diff
@@ -157,7 +158,10 @@ def __init__(self, learner):
     def predict(self, dataframe):
         test_data = TabularList.from_df(dataframe, cont_names=self.learner.data.cont_names)
         self.learner.data.add_test(test_data)
-        return self.learner.get_preds(DatasetType.Test)[0].numpy()
+        preds, target = self.learner.get_preds(DatasetType.Test)
+        preds = pd.Series(map(tuple, preds.numpy()), name='predictions')
```
smurching, line 162:
```suggestion
        preds = pd.Series(map(np.array, preds.numpy()), name='predictions')
```

This will make the resulting DataFrame convertible into a Spark DataFrame when used with the Spark UDF functionality,

tests/fastai/test_fastai_model_export.py
```diff
@@ -0,0 +1,259 @@
...
+def fastai_custom_env(tmpdir):
+    conda_env = os.path.join(str(tmpdir), "conda_env.yml")
+    _mlflow_conda_env(
+        conda_env,
+        additional_pip_deps=["fastai", "pytest"])
+    return conda_env
+
+
+@pytest.mark.large
+def test_model_save_load(fastai_model, model_path):
```
smurching, line 61: Could we update this test to verify that the reloaded model produces the same raw output as the original model, in addition to verifying correctness of the `_FastAiModelWrapper` output? e.g. something like:

```
def test_model_save_load(fastai_model, model_path):
    model = fastai_model.model

    mlflow.fastai.save_model(fastai_learner=model, path=model_path)
    reloaded_model = mlflow.fastai.load_model(model_uri=model_path)
    # Verify reloaded model computes same predictions as original model
    test_data = TabularList.from_df(fastai_model.inference_dataframe)
    model.data.add_test(test_data)
    real_preds, real_target = map(lambda output: output.numpy(), model.get_preds(DatasetType.Test))
    reloaded_model.data.add_test(test_data)
    reloaded_preds, reloaded_target = map(lambda output: output.numpy(), reloaded_model.get_preds(DatasetType.Test))
    np.testing.assert_array_almost_equal(real_preds, reloaded_preds)
    np.testing.assert_array_almost_equal(real_target, reloaded_target)

    # Verify pyfunc produces same output as raw model
    reloaded_pyfunc = pyfunc.load_model(model_uri=model_path)
    pyfunc_predictions = reloaded_pyfunc.predict(fastai_model.inference_dataframe)

    assert len(pyfunc_predictions) == len(real_preds)
    assert set(pyfunc_predictions.columns) == {"predictions", "target"}
    for i, prediction_arr in enumerate(pyfunc_predictions["predictions"]):
        np.testing.assert_array_almost_equal(real_preds[i], prediction_arr)
    for i, target in enumerate(pyfunc_predictions["target"]):
        assert real_target[i] == target
```

smurching (commented):
@antoniomdk some more small comments, otherwise looks good!

smurching (approved):
LGTM

harupy:
@smurching @antoniomdk 

I just noticed that the fastai flavor is not added to [run-python-flavor-tests.sh](https://github.com/mlflow/mlflow/blob/master/travis/run-python-flavor-tests.sh).

smurching:
Ah @harupy thanks for the catch!

@antoniomdk would you be able to file a follow-up PR adding a `pytest --verbose tests/fastai --large` to https://github.com/mlflow/mlflow/blob/master/travis/run-python-flavor-tests.sh#L22? Ran the tests locally & they passed, so simply adding that one line should be sufficient. 

Thanks both!

mlflow/fastai.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+    import fastai
+    from pathlib import Path
+
+    path = os.path.abspath(path)
+    if os.path.exists(path):
+        raise MlflowException("Path '{}' already exists".format(path))
+    model_data_subpath = "model.fastai"
+    model_data_path = os.path.join(path, model_data_subpath)
+    model_data_path = Path(model_data_path)
```
harupy, line 90: @antoniomdk 

Why do you convert `model_data_path` to a path object here?

antoniomdk:
@smurching I'll add it on the Autologging PR

mlflow/fastai.py
harupy: Thanks for the clear explanation! I understand.

Issue #2692, opened 2020-04-13
[BUG] Having to manually _active_run_stack to bypass exceptions when using jupyter


ahmed-shariff:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Arch linux
- **MLflow installed from (source or binary)**: pip
- **Environment**: jupyter
- **MLflow version (run ``mlflow --version``)**: 1.7.2
- **Python version**: 3.8

### Describe the problem
I was playing around with using mlflow on jupyter and was using the `MLflowClient` to check the results. At one point I had manually deleted the mlruns folder. After that, wheen try to start a run again I kept getting the 
```
Run with UUID {} is already active. To start a new run, first end the current run with mlflow.end_run(). To start a nested run, call start_run with nested=True
```
When I try to call `end_run`, it kept throwing `MLflowException` since the directory doesn't exist.
I had to manually set the `_active_run_stack`:
```
mlflow.tracking.fluent._active_run_stack = []
```

A potential solution:
wrap the line https://github.com/mlflow/mlflow/blob/63f1333c0436f897bcbb2de0ad65e21390f9434d/mlflow/tracking/fluent.py#L164 in a try catch?

juntai-zheng:
Thanks for identifying the bug and putting in the fix! Closing the issue as it's resolved now.

PR #2693, opened 2020-04-13
Fix for exception when run file has been deleted

ahmed-shariff:
Fix for #2692

surrounding https://github.com/mlflow/mlflow/blob/63f1333c0436f897bcbb2de0ad65e21390f9434d/mlflow/tracking/fluent.py#L164 with a try catch block

## How is this patch tested?
Manually tested on a jupyter notebook

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -161,10 +161,10 @@ def end_run(status=RunStatus.to_string(RunStatus.FINISHED)):
     """End an active MLflow run (if there is one)."""
     global _active_run_stack
     if len(_active_run_stack) > 0:
-        MlflowClient().set_terminated(_active_run_stack[-1].info.run_id, status)
         # Clear out the global existing run environment variable as well.
         env.unset_variable(_RUN_ID_ENV_VAR)
-        _active_run_stack.pop()
+        run = _active_run_stack.pop()
+        MlflowClient().set_terminated(run.info.run_id, status)
 
 
 atexit.register(end_run)


mlflow/tracking/fluent.py
```diff
@@ -161,7 +161,10 @@ def end_run(status=RunStatus.to_string(RunStatus.FINISHED)):
     """End an active MLflow run (if there is one)."""
     global _active_run_stack
     if len(_active_run_stack) > 0:
-        MlflowClient().set_terminated(_active_run_stack[-1].info.run_id, status)
```
aarondav, line 164: Could we just reorder the lines appropriately so that we call the set_terminated after unsetting the global state? It seems OK to re-throw the exception, but the run should be ended afterwards.

aarondav (changes requested):
See comment above, seems like a good fix though!

mlflow/tracking/fluent.py
ahmed-shariff: The problem I was running into was when the files related to the run don't exist anymore, resulting in a deadlock, where I can't start a new run or end the run mlflow thinks is active.

ahmed-shariff: in retrospective, it might be better idea to handle this in the `file_store`?

dmatrix: Can you catch a more specific Exception like FileNotFound rather than a wider MLflowException?

mlflow/tracking/fluent.py
```diff
@@ -161,7 +162,12 @@ def end_run(status=RunStatus.to_string(RunStatus.FINISHED)):
...
+        try:
+            MlflowClient().set_terminated(_active_run_stack[-1].info.run_id, status)
+        except MlflowException as e:
+            # if the run does not exist keep going
+            if e.error_code != RESOURCE_DOES_NOT_EXIST:
```
ahmed-shariff, line 169: @dmatrix @aarondav

mlflow/tracking/fluent.py
ahmed-shariff: The problem I ran into comes from: https://github.com/mlflow/mlflow/blob/7c9ef36c338fb6eb33d8b9f5eb23b9ca556731a9/mlflow/store/tracking/file_store.py#L442

So I am passing over only that case

aarondav:
The new patch does not seem to address #2692. Namely, set_terminated will probably still throw in the case where the runs were deleted.

I think the simplest solution is to move the `MlflowClient().set_terminated(_active_run_stack[-1].info` line roughly 3 lines down, so we first unset the active_run stack. That way, after calling end_run, you can start a new run.

aarondav (approved):

aarondav:
Merged, thanks for the fix!

PR #2698, opened 2020-04-13
Add  model signature

tomasatdatabricks:
This PR introduces signatures and input examples for MLflow models and allows users to log them with their models.

Model signature defines schema of model input and output. The schema is based on a simplified data
frame model defined by a list of (optionally) named and (mandatory) typed columns. The currently
supported data types consists of basic scalar types (bool, integer, long, float, double) and string and binary string. Neither arrays (and multidimensional arrays) nor structured types (struct, map) are supported at this time.

Input example specifies one or more valid inputs (observations) to the model. The example is stored
as json and so the examples are limited to jsonable content with the exception of binary data
provided the signature is defined. If the example includes binary data and model signature is
defined, the binary data is base64 encoded before producing output.

User can construct schema by hand or infer it from a dataset. Input examples are provided as is
(pandas.DataFrame, numpy.ndarray, dictinary or list) and are json serialized  and stored as part of
the model artifacts.

 
## How is this patch tested?
Unit tests in tests/models/test_model_signatures_and_examples.

## Release Notes
Added definition and utilities to allow users to include model signature definition as well as example(s) of input data with their models. Both can be inspected at model deployment time and provide additional information about the model interface. In addition, model signatures can be used by MLflow model deployment e.g. when parsing data (json) or producing typed output (spark udf). 

This PR introduces signatures and input examples for MLflow models and allows users to log them with their models.

Model signature defines schema of model input and output. The schema is based on a simplified data
frame model defined by a list of (optionally) named and (mandatory) typed columns. The currently
supported data types consists of basic scalar types (bool, integer, long, float, double) and string and binary string. Neither arrays (and multidimensional arrays) nor structured types (struct, map) are supported at this time.

Input example specifies one or more valid inputs (observations) to the model. The example is stored
as json and so the examples are limited to jsonable content with the exception of binary data
provided the signature is defined. If the example includes binary data and model signature is
defined, the binary data is base64 encoded before producing output.

User can construct schema by hand or infer it from a dataset. Input examples are provided as is
(pandas.DataFrame, numpy.ndarray, dictinary or list) and are json serialized  and stored as part of
the model artifacts.

Users can now include model signature definition and/ or input example data with model logged with mlflow. These can provide additional information to the users of the model as well as model utilities (safer parsing, type checking).

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [x] Python


aarondav:
Is this ready for review?

tomasatdatabricks:
It's ready for review. I am still fixing 2 test issues but that is separate.

aarondav:
(testing out leaving a comment)

aarondav:
Neato!!

mlflow/models/signature.py
```diff
@@ -0,0 +1,464 @@
...
+            return cls(inputs)
+
+    def __eq__(self, other) -> bool:
+        return self.inputs == other.inputs and self.outputs == other.outputs
+
+    def __repr__(self) -> str:
+        return json.dumps({"ModelSignature": self.to_dict()}, indent=2)
+
+
+ModelInputExample = TypeVar('ModelInputExample', pd.DataFrame, np.ndarray, dict, list)
```
aarondav, line 192: Neat: `TypeError: Unexpected type of input_example. Expected one of (pandas.DataFrame, numpy.ndarray, dict, list), got <class 'int'>`

mlflow/models/signature.py
```diff
@@ -0,0 +1,464 @@
...
+class ColSpec(object):
+    """
+    Specification of a column used in model signature.
+    Declares data type and optionally a name.
+    """
+
+    def __init__(self, name: str, type: DataType):  # pylint: disable=redefined-builtin
+        self.name = name
+        try:
+            self.type = getattr(DataType, type) if isinstance(type, str) else type
```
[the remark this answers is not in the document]
tomasatdatabricks: We are using isinstance to determine if the passed value is a string. The getattr(DataType, type) is translating string to Datatype, e.g. getattr(DataType, "boolean") returns DataType.boolean.

mlflow/models/signature.py
```diff
@@ -0,0 +1,464 @@
...
+        return self._numpy_type
+
+
+class ColSpec(object):
+    """
+    Specification of a column used in model signature.
+    Declares data type and optionally a name.
+    """
+
+    def __init__(self, name: str, type: DataType):  # pylint: disable=redefined-builtin
```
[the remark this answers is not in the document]
tomasatdatabricks: type, we are only redefining it in the constructor. I could call it data_type but we call it type in json and this allows us to to ColSpec(**json.loads(...)). I think it's ok to redefine cause it's only in the context of the constructor which does not do that much.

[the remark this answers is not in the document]
tomasatdatabricks: I left name as optional so users can pass unnamed data such as numpy array.

mlflow/models/signature.py
```diff
@@ -0,0 +1,464 @@
...
+    Data types supported by Mlflow model signature.
+    """
+
+    def __new__(cls, value, numpy_type):
+        res = object.__new__(cls)
+        res._value_ = value
+        res._numpy_type = numpy_type
+        return res
+
+    boolean = (1, np.bool)
```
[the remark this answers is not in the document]
tomasatdatabricks: yeah that is just a python enum thing. value is the value of this item in the enum. You do not have to set it and let it be auto assigned. I assigned it explicitly cause I thought it was simpler.

mlflow/models/__init__.py
```diff
@@ -36,28 +43,51 @@ class Model(object):
...
+        if not isinstance(other, Model):
+            return False
+        return self.__dict__ == other.__dict__
+
     def add_flavor(self, name, **params):
         """Add an entry for how to serve the model in a given format."""
         self.flavors[name] = params
         return self
 
+    def get_flavor_conf(self, flavor_name: str):
```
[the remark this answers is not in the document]
tomasatdatabricks: yeah I will remove this from this pr

mlflow/utils/model_utils.py
```diff
@@ -7,6 +7,17 @@
 from mlflow.utils.uri import append_to_uri_path
 
 
+def _get_model_configuration(model_path):
```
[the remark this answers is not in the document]
tomasatdatabricks: I will remove this from this PR

mlflow/models/__init__.py
```diff
@@ -92,12 +127,24 @@ def log(cls, artifact_path, flavor, registered_model_name=None, **kwargs):
...
+                              what data to feed the model.
         :param kwargs: Extra args passed to the model flavor.
         """
         with TempDir() as tmp:
             local_path = tmp.path("model")
             run_id = mlflow.tracking.fluent._get_or_start_run().info.run_id
             mlflow_model = cls(artifact_path=artifact_path, run_id=run_id)
+            if input_example is not None:
+                input_schema = model_signature.inputs if model_signature is not None else None
+                mlflow_model.input_example = save_example(tmp.path(), input_example, input_schema)
```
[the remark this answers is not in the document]
tomasatdatabricks: sorry I missed this comment before. The schema was being used for data conversion but this is no longer the case.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+from typing import List, Dict, TypeVar
+
+from mlflow.exceptions import MlflowException
+from mlflow.utils.proto_json_utils import NumpyEncoder
+
+
+class TensorsNotSupportedException(MlflowException):
+    def __init__(self, msg):
+        super().__init__("Multidimensional arrays (aka tensors) are not supported. "
+                         "{}".format(msg))
```
aarondav, line 38: just `msg`?

mlflow/models/signature.py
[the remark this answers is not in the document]
aarondav: If name is optional, why is this not `def __init__(self, type: DataType, name: str = None)`?

mlflow/models/signature.py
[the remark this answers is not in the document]
aarondav: Haha - can't we do `DataType[type]`? from https://www.tutorialspoint.com/enum-in-python

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+
+    def __init__(self, name: str, type: DataType):  # pylint: disable=redefined-builtin
+        self.name = name
+        try:
+            self.type = getattr(DataType, type) if isinstance(type, str) else type
+        except AttributeError:
+            raise MlflowException("Unsupported type '{0}', expected instance of DataType or "
+                                  "one of {1}".format(type, [t.name for t in DataType]))
+
+    def to_dict(self) -> Dict[str, str]:
```
aarondav, line 81: nit: `Dict[str, str]` is technically overly restrictive, even if it's OK here. JSON does not have string values.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+        names_eq = (self.name is None and other.name is None) or self.name == other.name
+        return names_eq and self.type == other.type
+
+    def __repr__(self) -> str:
+        return "{name}: {type}".format(name=self.name, type=self.type)
+
+
+class Schema(object):
+    """
+    Schema specifies column types (:py:class:`DataType`) in a dataset.
```
aarondav, line 98: Maybe document that there is one "input schema" and one "output schema" for a model, to help understandability and differentiating the concepts.

mlflow/models/signature.py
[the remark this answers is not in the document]
tomasatdatabricks: ah yeah, good point, I'll change that.

mlflow/models/signature.py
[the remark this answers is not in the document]
tomasatdatabricks: oh we can? let me try, this was my first time using enums, but it makes sense this should be possible :)

mlflow/models/signature.py
```diff
@@ -0,0 +1,476 @@
...
+
+    def __eq__(self, other) -> bool:
+        return self.inputs == other.inputs and self.outputs == other.outputs
+
+    def __repr__(self) -> str:
+        return json.dumps({"ModelSignature": self.to_dict()}, indent=2)
+
+
+ModelInputExample = TypeVar('ModelInputExample', pd.DataFrame, np.ndarray, dict, list)
+MlflowModelDataset = TypeVar('MlflowModelDataset', pd.DataFrame, np.ndarray, Dict[str, np.ndarray])
```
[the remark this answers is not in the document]
tomasatdatabricks: Do you mean why do we have different type for datasets and examples?

That's because of slight difference in accepted types.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+    Infer an MLflow model signature from the training data (input) and model predictions (output).
+    This method captures the column names and data types from the user data. The signature
+    represents model input and output as dataframes with (optionally) named columns and data type
+    specified as one of types defined in :py:class:`DataType`. This method will raise
+    an exception if the user data contains incompatible types or is not passed in one of the
+    supported formats (containers).
+
+    Supported input:
+
+    container types:
```
aarondav, line 214: I think container types is common parlance in R, but haven't seen it in python/DataFrame world. Maybe can just that "the input should be one of these", and that "each element should be one of these".

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+      - pandas.DataFrame
+      - dictionary of { name -> numpy.ndarray}
+      - numpy.ndarray
+      - pyspark.sql.DataFrame
```
aarondav, line 218: Do we actually support this one? seems like the TypeVar does not include it.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+
+def _infer_schema(data: MlflowModelDataset) -> Schema:
+    if hasattr(data, "__len__") and len(data) == 0:
+        return Schema([])
+    if isinstance(data, dict):
+        res = []
+        for col in data.keys():
+            ary = data[col]
+            if not isinstance(ary, np.ndarray):
+                raise TypeError("Data in the dictionary must be of type numpy.ndarray")
```
aarondav, line 418: Why is one of these a TypeError and the other an MlflowException?

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+        for i in binary_cols:
+            col = df.columns[i]
+            df[col] = np.array(df[col].map(_base64decode), dtype=np.bytes_)
+            return df
+    else:
+        return pd.read_json(path_or_str, orient=pandas_orient, dtype=False)
+
+
+def _map_numpy_dtype(col: np.ndarray) -> DataType:
+    if not isinstance(col, np.ndarray):
```
aarondav, line 343: Is this possible? Don't python type ascriptions check this?

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+        raise TypeError("Expected numpy.ndarray, got '{}'.".format(type(col)))
+    if len(col.shape) > 1:
+        raise MlflowException("Expected 1d array, got array with shape {}".format(col.shape))
+    if col.dtype.kind == "b":
+        return DataType.boolean
+    elif col.dtype.kind == "i" or col.dtype.kind == "u":
+        if col.dtype.itemsize < 4 or col.dtype.kind == "i" and col.dtype.itemsize == 4:
```
aarondav, line 350: Mind parentheses?

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+            return DataType.integer
+        elif col.dtype.itemsize < 8 or col.dtype.kind == "i" and col.dtype.itemsize == 8:
```
aarondav, line 352: Same here for the parentheses

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+            return DataType.long
+    elif col.dtype.kind == "f":
+        if col.dtype.itemsize <= 4:
+            return DataType.float
+        elif col.dtype.itemsize <= 8:
+            return DataType.double
+    elif col.dtype.kind == "U":
+        return DataType.string
+    elif col.dtype.kind == "S":
```
aarondav, line 361: Do we have unit tests for all these cases? Would be good to just create a variety of ndarrays and confirm they all work, so I don't have to read the numpy documentation.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+        return DataType.binary
+    elif col.dtype.kind == "O":
+        first_elem = col[0]
+        if isinstance(first_elem, bytearray) or isinstance(first_elem, bytes) and all(
+                [isinstance(x, bytearray) or isinstance(x, bytes) for x in col]):
+            return DataType.binary
+        elif isinstance(first_elem, str) and all([isinstance(x, str) for x in col]):
```
aarondav, line 368: Is `first_elem` a performance thing? I think `all` short-circuits, so seems unnecessary.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+
+def _is_spark_df(x) -> bool:
+    try:
+        return isinstance(x, importlib.import_module("pyspark.sql.dataframe").DataFrame)
+    except ImportError:
+        return False
+
+
+def _infer_schema(data: MlflowModelDataset) -> Schema:
+    if hasattr(data, "__len__") and len(data) == 0:
```
aarondav, line 411: Ideally we don't need to do this

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+            schema = _infer_schema(df)
+            return Schema([ColSpec(None, t) for t in schema.column_types()])
+        else:
+            if len(data.shape) == 2:
+                array_type = _map_numpy_dtype(data[0])
+                return Schema([ColSpec(None, array_type)] * data.shape[1])
+            else:
+                array_type = _map_numpy_dtype(data)
+                return Schema([ColSpec(None, array_type)])
+    elif _is_spark_df(data):
```
aarondav, line 444: Is this possible given the type signature?

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+
+        for i in binary_cols:
+            col = df.columns[i]
+            df[col] = np.array(df[col].map(_base64decode), dtype=np.bytes_)
+            return df
+    else:
+        return pd.read_json(path_or_str, orient=pandas_orient, dtype=False)
+
+
+def _map_numpy_dtype(col: np.ndarray) -> DataType:
```
aarondav, line 342: Why is this called map as opposed to `_infer_numpy_dtype`? Maybe just a doc comment would be helpful.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+    elif col.dtype.kind == "O":
+        first_elem = col[0]
+        if isinstance(first_elem, bytearray) or isinstance(first_elem, bytes) and all(
+                [isinstance(x, bytearray) or isinstance(x, bytes) for x in col]):
+            return DataType.binary
+        elif isinstance(first_elem, str) and all([isinstance(x, str) for x in col]):
+            return DataType.string
+        elif isinstance(first_elem, int) and all([isinstance(x, int) for x in col]):
+            return DataType.long
+        elif isinstance(first_elem, float) and all([isinstance(x, float) for x in col]):
```
aarondav, line 372: How do nulls interact with these conditonals?

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+    elif isinstance(x, pyspark.sql.types.BinaryType):
+        return DataType.binary
+    else:
+        raise Exception("Unsupported Spark Type '{}', MLflow schema is only supported for scalar "
+                        "Spark types.".format(type(x)))
+
+
+def _is_spark_df(x) -> bool:
+    try:
+        return isinstance(x, importlib.import_module("pyspark.sql.dataframe").DataFrame)
```
aarondav, line 405: I think that we can just call `import`.

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+    :param model_input: Valid input to the model. E.g. (a subset of) training dataset.
+    :param model_output: Valid model output. Model predictions for the (subset of) training dataset.
+    :return: ModelSignature
+    """
+    inputs = _infer_schema(model_input)
+    outputs = _infer_schema(model_output) if model_output is not None else None
+    return ModelSignature(inputs, outputs)
+
+
+def save_example(path: str, input_example: ModelInputExample) -> str:
```
aarondav, line 234: Would it make sense to split this file up at all? Right now, it handles a few different concerns -- examples and signature/schema inference seem pretty independent. Also some random stuff about loading DataFrames from JSON. Could have `example.py`, `signature.py`, and `utils.py`, for example, might make it a bit easier to read and understand.

mlflow/models/signature.py
tomasatdatabricks: Hmm, you mean dropping the shared part of the message altogether? Yeah I get it makes sense.

mlflow/models/signature.py
tomasatdatabricks: Makes sense!

mlflow/models/signature.py
tomasatdatabricks: We do support it. The typevar can not have it because it requires importing spark. The typevar is more for a hint, the types are not enforced. I guess alternatively we could mode infer)signature with spark frame into mfllow.spark package but I think this one is nicer, despite the missing type. Maybe I can make the typevar conditional depending on whether pyspark is available.

mlflow/models/signature.py
tomasatdatabricks: The other is the "must me 1 dimensional" exception? Technically that one is not a TypError as it is the right type (numpy array) but wrong value.

mlflow/models/signature.py
tomasatdatabricks: The types are not enforced. It's just an annotation for user and for tools such as IDE.

mlflow/models/signature.py
tomasatdatabricks: Make sense, will add them.

mlflow/models/signature.py
tomasatdatabricks: yeah it's a perf thing, the list that is passed to all would be materialized which I wanted to avoid.

mlflow/models/signature.py
tomasatdatabricks: yes

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+                                      "got shape {}".format(ary.shape))
+        return Schema(res)
+
+    if isinstance(data, pd.DataFrame):
+        return Schema([ColSpec(col, _map_numpy_dtype(data[col].values)) for col in data.columns])
+    elif isinstance(data, np.ndarray):
+        if len(data.shape) > 2:
+            raise MlflowException("Multidimensional arrays (aka tensors) are not supported, "
+                                  "got array with shape {}".format(data.shape))
+        if data.dtype == np.dtype("O"):
```
aarondav, line 433: Maybe document this piece of code and the three cases

mlflow/models/signature.py
```diff
@@ -0,0 +1,449 @@
...
+            df = pd.DataFrame(data).infer_objects()
+            schema = _infer_schema(df)
+            return Schema([ColSpec(None, t) for t in schema.column_types()])
+        else:
```
aarondav, line 437: nit: Probably could elif this

mlflow/models/signature.py
tomasatdatabricks: good question, there is no reason for that! I'll rename it.

mlflow/models/signature.py
tomasatdatabricks: kk makes sense.

mlflow/models/signature.py
tomasatdatabricks: Good question, they return false. I will update the logic. Unclear what to do about integers though. Integers do not have a way to represent NaNs or missing values.

mlflow/models/signature.py
tomasatdatabricks: makes sense, will do

docs/source/python_api/mlflow.models.utils.rst
```diff
@@ -0,0 +1,6 @@
+mlflow.models.utils
+===================
+.. automodule:: mlflow.models.utils
```
aarondav, line 3: Do we really want to expose `mlflow.models.utils`? Couldn't we export important functions in `mlflow.models`? Probably same for `signature`.

docs/source/python_api/mlflow.types.schema.rst
```diff
@@ -0,0 +1,6 @@
+mlflow.types.schema
```
aarondav, line 1: similar here for types.schema, seems like we shouldn't be introducing so many exposed modules for this PR

docs/source/python_api/mlflow.models.utils.rst
aarondav: Btw, I think save_example definitely shouldn't be public initially. It is only available through mlflow model log -- seems fine.

docs/source/python_api/mlflow.models.utils.rst
```diff
@@ -0,0 +1,6 @@
...
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
aarondav, line 6: Documentation looks weird around the experimental warning:

image

docs/source/python_api/mlflow.models.utils.rst
aarondav: Would recommend inlining a subset of the `save_example` documentation. We don't need to assert as many properties (e.g., 2D arrays being unsupported).

Maybe could simplify most of the paragraphs as "The given example will be converted to a Pandas DataFrame and then serialized using the Pandas split-oriented format. Bytes are base64-encoded."

tomasatdatabricks: We do not have to expose mlflow.model.utils if we hide save_example.

docs/source/python_api/mlflow.types.schema.rst
tomasatdatabricks: Yeah I agree, I don't know how to make stuff visible to documentation without exporting the module though.

docs/source/python_api/mlflow.types.schema.rst
```diff
@@ -0,0 +1,6 @@
...
+===================
+.. automodule:: mlflow.types.schema
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
aarondav, line 6: Doesn't look production-ready:

image

docs/source/python_api/mlflow.types.schema.rst
aarondav: Can we embed it into types? We should also get feedback from others on if we should release with these packages, but we can do that post-merge. For now let's just make sure it looks reasonable.

tests/types/test_schema.py
```diff
@@ -0,0 +1,151 @@
...
+        schema = _infer_schema(pandas_df_with_all_types)
+        assert schema == Schema([ColSpec(x, x) for x in pandas_df_with_all_types.columns])
+        spark_session = pyspark.sql.SparkSession(pyspark.SparkContext.getOrCreate())
+        spark_schema = StructType(
+            [StructField(t.name, _parse_datatype_string(t.name), True)
+             for t in schema.column_types()])
+        sparkdf = spark_session.createDataFrame(pandas_df_with_all_types, schema=spark_schema)
+        schema = _infer_schema(sparkdf)
+        assert schema == Schema([ColSpec(x, x) for x in pandas_df_with_all_types.columns])
+    except ImportError:
```
aarondav, line 150: super sketch, why not only run when it's available (e.g., large tests)?

tomasatdatabricks: oh, good catch, I have no idea why it's not just marked large

docs/source/python_api/mlflow.types.schema.rst
tomasatdatabricks: done

aarondav (approved):

PR #2705, opened 2020-04-15
Change stalebot for "Needs-Author-Feedback" use-case

aarondav:

Goal of this change is that a reviewer can add the label "Needs-Author-Feedback" which is auto-cleared when the PR is interacted with -- and therefore someone sweeping for PRs which need action taken can ignore all the PRs that are in this camp.


smurching (approved):
LGTM

PR #2707, opened 2020-04-15
Use LibYAML by default to increase performance in FileUtils

Higgcz:
I've noticed significant performance issues when using *file store* tracking backend compared to other frameworks. After profiling the code, I've noticed the `log_metric` method ([file_store.py:642](https://github.com/mlflow/mlflow/blob/master/mlflow/store/tracking/file_store.py#L642)) is loading YAML file with the run information to see if the run is currently active or not. This check is happening every time the `mlflow.log_metric` function is called.

By default, Python's YAML is using Python parser, which is significantly slower than its C++ counterpart. See [PyYAML Documentaion](https://pyyaml.org/wiki/PyYAMLDocumentation).

I've added a try-except block into the `mlflow/utils/file_utils.py`, which tries to import the `CSafeLoader` and `CSafeDumper` to replace the Python's default ones. If this fails, it means *LibYAML* is not installed on the system and `file_utils.py` will import the default ones.

I also added small paragraph about *File store performance* under **Storage** section, see the file `docs/source/tracking.rst`. Mainly to inform users that if the LibYAML is not installed and they experience some performance issues, they can easily install it.

## How is this patch tested?

- tested it on macOS
- tested in Docker - Debian with and without LibYAML

Improves performance of reading experiment data from local filesystem when LibYAML is installed

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [x] Docs
- [x] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/utils/file_utils.py
+++ b/mlflow/utils/file_utils.py
@@ -14,9 +14,9 @@ from six.moves import urllib
 
 import yaml
 try:
-    from yaml import CSafeLoader as SafeLoader, CSafeDumper as SafeDumper
+    from yaml import CSafeLoader as YamlSafeLoader, CSafeDumper as YamlSafeDumper
 except ImportError:
-    from yaml import SafeLoader, SafeDumper
+    from yaml import SafeLoader as YamlSafeLoader, SafeDumper as YamlSafeDumper
 
 from mlflow.entities import FileInfo
 from mlflow.exceptions import MissingConfigException
@@ -142,7 +142,9 @@ def write_yaml(root, file_name, data, overwrite=False):
     try:
         with codecs.open(yaml_file_name, mode='w', encoding=ENCODING) as yaml_file:
             yaml.dump(data, yaml_file,
-                      default_flow_style=False, allow_unicode=True, Dumper=SafeDumper)
+                      default_flow_style=False,
+                      allow_unicode=True,
+                      Dumper=YamlSafeDumper)
     except Exception as e:
         raise e
 
@@ -165,7 +167,7 @@ def read_yaml(root, file_name):
         raise MissingConfigException("Yaml file '%s' does not exist." % file_path)
     try:
         with codecs.open(file_path, mode='r', encoding=ENCODING) as yaml_file:
-            return yaml.load(yaml_file, Loader=SafeLoader)
+            return yaml.load(yaml_file, Loader=YamlSafeLoader)
     except Exception as e:
         raise e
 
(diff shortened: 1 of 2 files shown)


pogil (changes requested):

mlflow/utils/file_utils.py
```diff
@@ -137,7 +141,7 @@ def write_yaml(root, file_name, data, overwrite=False):
 
     try:
         with codecs.open(yaml_file_name, mode='w', encoding=ENCODING) as yaml_file:
-            yaml.safe_dump(data, yaml_file, default_flow_style=False, allow_unicode=True)
+            yaml.dump(data, yaml_file, default_flow_style=False, allow_unicode=True, Dumper=SafeDumper)
```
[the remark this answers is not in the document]
Higgcz: That’s a good question, let me check. I think I was just assuming it doesn’t use LibYAML by default, even if it’s available, but actually it makes more sense if it does.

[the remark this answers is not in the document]
Higgcz: I've checked and unfortunately, those changes are necessary. PyYAML will install LibYAML bindings if LibYAML is available, but you still need to explicitly use `CLoader` and `CDumper`.

Higgcz:
I've updated the code to satisfy linter, unfortunately we need to keep these changes in `file_utils.py`.

pogil (approved):
Thanks, looks good to me!

mlflow/utils/file_utils.py
```diff
@@ -13,6 +13,10 @@
 from six.moves import urllib
 
 import yaml
+try:
+    from yaml import CSafeLoader as SafeLoader, CSafeDumper as SafeDumper
+except ImportError:
+    from yaml import SafeLoader, SafeDumper
```
mparkhe, line 19: Can you use a different alias? Eg.

```
try:
    from yaml import CSafeLoader as YamlSafeLoader, CSafeDumper as YamlSafeDumper
except ImportError:
    from yaml import SafeLoader as YamlSafeLoader, SafeDumper as YamlSafeDumper
```

and use `YamlSafeLoader` and `YamlSafeDumper` in load and dump method.

Higgcz: Done. One more question, should I squash the commits?

mparkhe: No worries. I can squash commits when merging.

PR #2709, opened 2020-04-16
Fix autologging of Keras optimizers for TF.Keras 1.X

juntai-zheng:

Originally, MLflow would only autolog TensorFlow optimizers when using Tensorflow.Keras models. This change allows for the autologging of Keras optimizers as well.

## How is this patch tested?

Changed the test in `tests/tensorflow_autolog/test_tensorflow_autolog.py` to reflect this change.

Keras optimizers are now correctly logged when autologging with TensorFlow.Keras 1.X models.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [X] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [X] Python

### DIFF
--- a/mlflow/tensorflow.py
+++ b/mlflow/tensorflow.py
@@ -471,10 +471,12 @@ class __MLflowTfKerasCallback(Callback):
 
     def on_train_begin(self, logs=None):  # pylint: disable=unused-argument
         opt = self.model.optimizer
-        # Secondary checks are if the optimizer is a TensorFlow optimizer rather than a Keras one.
         if hasattr(opt, '_name'):
             try_mlflow_log(mlflow.log_param, 'optimizer_name', opt._name)
+        # Elif checks are if the optimizer is a TensorFlow optimizer rather than a Keras one.
         elif hasattr(opt, 'optimizer'):
+            # TensorFlow optimizer parameters are associated with the inner optimizer variable.
+            # Therefore, we assign opt to be opt.optimizer for logging parameters.
             opt = opt.optimizer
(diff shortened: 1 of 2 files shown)


mlflow/tensorflow.py
```diff
@@ -471,13 +471,23 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
     def on_train_begin(self, logs=None):  # pylint: disable=unused-argument
         opt = self.model.optimizer
-        if hasattr(opt, 'optimizer'):
+        # Secondary checks are if the optimizer is a TensorFlow optimizer rather than a Keras one.
```
smurching, line 474: Can we pull the reassignment of `opt` up into its own if-statement before logging any params & explain why we do it? e.g.

```
# Explain why we do this
if hasattr(opt, 'optimizer'):
    opt = opt.optimizer
# Log optimizer attributes...
if hasattr(opt, '_name'):
```

Or if there's a reason we bundle assigning `opt` & logging the `name` together in the same branch, could we add a clarifying comment (e.g. because we want to log the name of the outer optimizer, but extract the inner optimizer so that we can log its learning rate & epislon)? Or maybe just log the name first, and then reassign `opt` afterwards? I think either would help make this clearer

smurching (commented):
One comment, otherwise looks good

smurching (approved):
LGTM

PR #2710, opened 2020-04-16
Refactor tests: split out sagemaker, tensorflow, and other model flavors from python large

mparkhe:
## What changes are proposed in this pull request?

Splitting long running job in stage `Large Python` into parallel jobs:
- `TensorFlow`: tf tests
- `SageMaker`: build, push, tests
- all other `Flavors`, and 
- `Large Python` (everything else)
#### Note: Runtimes can be optimized further by splitting out library dependencies for each job.

Stages `Small` and `Nightly`: not changed

### Before
image

### After
image

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s) does this PR affect?

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [ ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [ ] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #2716, opened 2020-04-16
Use pytest.ini to specify `--color=yes`

harupy:

Instead of adding `--color=yes` to all the pytest commands, we can write it in `pytest.ini`.

## How is this patch tested?

Verified the pytest output is colored.

https://github.com/mlflow/mlflow/pull/2716/checks?check_run_id=592800991

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+addopts = --color=yes
--- a/travis/run-large-python-tests.sh
+++ b/travis/run-large-python-tests.sh
@@ -16,7 +16,7 @@ else
 fi
 
 # NB: Also add --ignore'd tests to run-small-python-tests.sh
-pytest tests --color=yes --large --ignore=tests/examples --ignore=tests/h2o --ignore=tests/keras \
+pytest tests --large --ignore=tests/examples --ignore=tests/h2o --ignore=tests/keras \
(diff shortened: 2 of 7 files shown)


pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+addopts = --color=yes
```
harupy, line 2: With this configuration, `pytest` is equivalent to `pytest --color=yes`.

https://docs.pytest.org/en/latest/customize.html#adding-default-options

harupy: It seems all the pytest commands contain `--verbose`. Should it also be included in `addopts`?

aarondav: I guess the small test runner doesn't have `--verbose`, so can probably just keep this here.

aarondav:
@harupy Thanks for the change! By the way, Mani's PR went in which moved around the large tests, mind rebasing for that? #2710

aarondav (approved):

aarondav:
LGTM! Will merge after tests pass.

PR #2724, opened 2020-04-17
Move all small tests to GitHub Actions

aarondav:

Adds lint, Java, JS, and proto tests to GitHub actions, and removes them from Travis. Travis doesn't allow us to run very many concurrent builds (despite our longest step taking 20 minutes, and most steps running in parallel, builds typically take ~50 minutes to run -- and if you have 2 PRs in progress, the second one will block behind the first).

I tried and failed to get the R build to work (complaints about `curl` and `gh` not being installed by `usethis`), so gave up and just moved R to the large tests in Travis.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -51,19 +51,13 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - uses: r-lib/actions/setup-r@v1
-      with:
-        r-version: '3.3.0'
     - name: Install dependencies
       run: |
-        export GITHUB_WORKFLOW=1
-        # cached packages dramatically decreases build time, but it must not include mlflow
-        rm -rf $HOME/R/Library/mlflow
-        cd mlflow/R/mlflow
-        Rscript -e 'install.packages("devtools")'
-        Rscript -e 'devtools::install_deps(dependencies = TRUE, upgrade = FALSE)'
+        install.packages("devtools")
+        remotes::install_deps(dependencies = TRUE, upgrade = FALSE)
+      shell: Rscript {0}
     - name: Run tests
       run: |
-        export GITHUB_WORKFLOW=1
         source ./travis/install-common-deps.sh
         export PATH="$HOME/miniconda/bin:$PATH"
         source activate test-environment
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -7,6 +7,26 @@ on:
     branches: [ master ]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
+    - name: Install dependencies
+      run: |
+        export GITHUB_WORKFLOW=1
+        INSTALL_LARGE_PYTHON_DEPS=true INSTALL_SMALL_PYTHON_DEPS=true source ./travis/install-common-deps.sh
+        pip install -r ./travis/lint-requirements.txt
+    - name: Run tests
+      run: |
+        export GITHUB_WORKFLOW=1
+        export PATH="$HOME/miniconda/bin:$PATH"
+        source activate test-environment
+        ./lint.sh
+
   python-small:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


aarondav:
All small tests completed after 10 minutes end-to-end. On [Travis](https://travis-ci.org/github/mlflow/mlflow/builds/675848171), the tests individually take the same time (Lint is the most at 10 minutes), but the stage from start to finish took 22 minutes because of the limited number of concurrent runners (there we are limited to 5 _across all PRs).

.github/workflows/master.yml
```diff
@@ -7,6 +7,26 @@ on:
...
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
```
mparkhe, line 17: 👍

.github/workflows/master.yml
```diff
@@ -25,3 +45,53 @@ jobs:
...
+        source ./travis/install-common-deps.sh
+    - name: Run tests
+      run: |
+        cd mlflow/server/js
+        npm i
+        npm run lint
+        set -e
+        for file in $(find src | grep test.js | sort) ; do npm run test -- $file; done
+
+  protos:
```
mparkhe, line 87: 👍

.travis.yml
```diff
@@ -10,17 +10,9 @@ matrix:
     # Nightly builds run the tests/examples end-to-end test suite.
     # Does not include Windows since Projects/Models are mostly not compatible with Windows as of now.
     - stage: Nightly
-    - stage: Small
```
mparkhe, line 13: corresponding section for `TRAVIS_BUILD_STAGE_NAME` from `install:` line 138-141 (old version) can be cleaned out. Also remove `"$TRAVIS_BUILD_STAGE_NAME" == "Small"` in `./travis/stage-python3.sh`

mparkhe (commented):
remove all references for small in `.travis.yml` and `./travis/stage-python3.sh`

.github/workflows/master.yml
```diff
@@ -25,3 +45,53 @@ jobs:
...
+        source activate test-environment
+        cd mlflow/java
+        mvn clean package -q
+
+  js:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Node
+      uses: actions/setup-node@v1
```
smurching, line 74: Looks like this currently uses node 10.x (from github actions JS [build logs](https://github.com/mlflow/mlflow/pull/2724/checks?check_run_id=595900760)):

```
Run actions/setup-node@v1
  with:
    always-auth: false
    node-version: 10.x
```


Maybe we should pin to that version? Couldn't figure out how the nodejs action picks which version to install if none is specified (looking at [this code](https://github.com/actions/setup-node/blob/339aa41a8d72adcf661974ca2a947fa1e144530e/src/setup-node.ts#L12-L18), you'd expect it to not install nodejs at all if no `version` or `node-version` attribute is specified)

smurching (commented):
Looks great once @mparkhe's comments are addressed, just had a question RE the node version used in the JS build - thanks @aarondav!!!

.github/workflows/master.yml
```diff
@@ -7,6 +7,26 @@ on:
...
+      with:
+        python-version: 3.6
+    - name: Install dependencies
+      run: |
+        export GITHUB_WORKFLOW=1
+        INSTALL_LARGE_PYTHON_DEPS=true INSTALL_SMALL_PYTHON_DEPS=true source ./travis/install-common-deps.sh
+        pip install -r ./travis/lint-requirements.txt
+    - name: Run tests
+      run: |
+        export GITHUB_WORKFLOW=1
```
harupy, line 25: Why overwrite `GITHUB_WORKFLOW` with `1`?

https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

aarondav: oh, I didn't realize there was already such an environment variable set! Haha.

.travis.yml
aarondav: Removed!

.github/workflows/master.yml
aarondav: Done!

.github/workflows/master.yml
```diff
@@ -7,6 +7,24 @@ on:
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
+    - name: Install dependencies
+      run: |
+        INSTALL_LARGE_PYTHON_DEPS=true INSTALL_SMALL_PYTHON_DEPS=true source ./travis/install-common-deps.sh
+        pip install -r ./travis/lint-requirements.txt
```
harupy, line 21:
```suggestion
      env:
        INSTALL_LARGE_PYTHON_DEPS: true
        INSTALL_SMALL_PYTHON_DEPS: true
      run: |
        ./travis/install-common-deps.sh
        pip install -r ./travis/lint-requirements.txt
```

We can use [the env context](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#env) here to set `INSTALL_LARGE_PYTHON_DEPS` and `INSTALL_SMALL_PYTHON_DEPS`.

aarondav:
@harupy Do you know why we install fbprophet as part of large tests?

harupy:
@aarondav I think it's installed to run this example in [test_examples](https://github.com/mlflow/mlflow/blob/master/tests/examples/test_examples.py):

https://github.com/mlflow/mlflow/tree/master/examples/prophet

aarondav:
@harupy That one seems to have a conda.yaml that includes prophet -- let's see if the large tests fail without it. I saw a built fail to install prophet because of pystan not being there, which doesn't make a lot of sense since pystan is also part of our requirements.

smurching (approved):
LGTM

PR #2726, opened 2020-04-17
Add mlflow.note.content to tags and documentation

crflynn:

Adds an internal mlflow tag `mlflow.note.content`  to `mlflow_tags.py`. Updates the System Tags documentation to describe this tag.

## How is this patch tested?

There are no tests here, however, the mlflow tag that is used is the same from [NoteUtils.js](https://github.com/mlflow/mlflow/blob/bf31f01a3b3302752fcbd96093c7a7ac96f9c197/mlflow/server/js/src/experiment-tracking/utils/NoteUtils.js#L3) which applies the note to the run with the browser editor.

Adds a system tag constant which is used by the model run Note feature.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [x] Docs
- [x] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


aarondav:
I'm not sure this use-case warrants entirely new top-level `mlflow.set_note` / `mlflow.delete_note` methods. This tag should probably be documented [here](https://www.mlflow.org/docs/latest/tracking.html#system-tags), so that people know how to overwrite the note per our public documentation (as opposed to reading code). What do you think?

crflynn:
I think that alternative is fine and would look something like this:

```python
import mlflow
from mlflow.utils.mlflow_tags import MLFLOW_RUN_NOTE

mlflow.start_run()
# training the model...
mlflow.set_tag(key=MLFLOW_RUN_NOTE, value="this is a note about what we are training")
mlflow.end_run()
```

The goal here being to be able to clearly set the note for a run without having to use the UI. I'd be happy with either implementation and can make the suggested changes if this alternative is preferable.

aarondav:
That looks great to me! So the proposed changes are:
1. Add this tag to mlflow_tags.py
2. Update the tracking docs table to include this tag (and mention that the user may overwrite this value, which is special compared to most other system tags).

Also, I noticed that the comment at the top of `mlflow_tags` is outdated ("See REST API documentation") -- let's update that to point to the System Tags table.

Thanks!

aarondav (approved):
Looks great! Checked out the markdown locally and all looks good there.

PR #2731, opened 2020-04-18
Bugfix projects on windows

MynherVanKoek:

I've had some problems running `mlflow run` on Windows. The error appeared when the project's conda environment was activated within the `mlflow run` workflow, i.e., `conda activate` throw the error. I executed `conda init cmd.exe` beforehand, though. More interestingly, `mlflow models serve` does work, which follows the same workflow as `mlflow run`: optionally creating a conda environment and activating it. After some research, I found the difference in the implementation of both modules: the call of the `Popen` method. The backend for `mlflow models serve` is using `["cmd", "/c", command]` as argument, which is working fine. `mlflow run`, however, is only using `command`, which does not work. I changed that.

## How is this patch tested?

I changed the code in my local MLflow installation on my Windows laptop and `mlflow run` finally worked smoothly. I refrained from executing the complete unit test procedure since only the activation of the conda environment on Windows laptops is affected and no MLflow component is touched. Furthermore, I only copied pre-existing and already working code from the MLflow model-serving part and did not invent anything new. 

## Release Notes

Bug fix: Enale MLflow Projects to run on Windows

- [ ] UI
- [x] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [x] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python


tomasatdatabricks (approved):
thanks for the fix!

PR #2742, opened 2020-04-22
Onnx.predict handles list response

andychow-db:

MLflow assumes that an onnx model will return a numpy array. A user
discovered this is not the case and in fact, the model can return a
python list.

The result is that the service throws an error when it attempts to call
reshape() on a python list and that function does not exist.

The fix is to look convert a list to numpy array before calling reshaping it.

## How is this patch tested?

1. There is a new python.large test that validates that it works with an sklearn based onnx model.
2. I manually ran the mlflow model using `mlflow models -m models.onnx` and crafted a request to verify it works successfully. 

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/onnx.py
+++ b/mlflow/onnx.py
@@ -175,9 +175,8 @@ class _OnnxModelWrapper:
         def format_output(data):
             # Output can be list and it should be converted to a numpy array
             # https://github.com/mlflow/mlflow/issues/2499
-            if isinstance(data, list):
-                data = np.asarray(data)
-            return data
+            data = np.asarray(data)
+            return data.reshape(-1)
         response = pd.DataFrame.from_dict({c: format_output(p)
                                            for (c, p) in zip(self.output_names, predicted)})
         return response
--- a/mlflow/onnx.py
+++ b/mlflow/onnx.py
@@ -178,7 +178,8 @@ class _OnnxModelWrapper:
             if isinstance(data, list):
                 data = np.asarray(data)
             return data
-        response = pd.DataFrame.from_dict({c: format_output(p) for (c, p) in zip(self.output_names, predicted)})
+        response = pd.DataFrame.from_dict({c: format_output(p)
+                                           for (c, p) in zip(self.output_names, predicted)})
         return response
 
 
--- a/mlflow/onnx.py
+++ b/mlflow/onnx.py
@@ -170,10 +170,16 @@ class _OnnxModelWrapper:
             }
         else:
             feed_dict = {self.inputs[0][0]: dataframe.values}
-
         predicted = self.rt.run(self.output_names, feed_dict)
-        return pd.DataFrame.from_dict(
-            {c: p.reshape(-1) for (c, p) in zip(self.output_names, predicted)})
+
+        def format_output(data):
+            # Output can be list and it should be converted to a numpy array
+            # https://github.com/mlflow/mlflow/issues/2499
+            if isinstance(data, list):
+                data = np.asarray(data)
+            return data
+        response = pd.DataFrame.from_dict({c: format_output(p) for (c, p) in zip(self.output_names, predicted)})
+        return response
 
 
 def _load_pyfunc(path):
--- a/tests/onnx/test_onnx_model_export.py
+++ b/tests/onnx/test_onnx_model_export.py
@@ -77,6 +77,7 @@ def onnx_sklearn_model(sklearn_model):
 def predicted(model, data):
     return model.predict(data[0])
 
+
 @pytest.fixture(scope='module')
 def tf_model_multiple_inputs_float64():
     graph = tf.Graph()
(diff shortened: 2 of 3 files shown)


mlflow/onnx.py
```diff
@@ -170,10 +170,17 @@ def predict(self, dataframe):
...
-            {c: p.reshape(-1) for (c, p) in zip(self.output_names, predicted)})
+
+        def format_output(data):
+            # Output can be list and it should be converted to a numpy array
+            # https://github.com/mlflow/mlflow/issues/2499
+            if isinstance(data, list):
+                data = np.asarray(data)
+            return data
+        response = pd.DataFrame.from_dict({c: format_output(p)
+                                           for (c, p) in zip(self.output_names, predicted)})
```
dbczumar, line 182: @andychow-db Looks like we're no longer reshaping the numpy array output. This is necessary in order to flatten multi-dimensional arrays into Pandas DataFrame columns. I would expect pyfunc tests to fail without this reshaping logic.

mlflow/onnx.py
```diff
@@ -170,10 +170,17 @@ def predict(self, dataframe):
...
-
         predicted = self.rt.run(self.output_names, feed_dict)
-        return pd.DataFrame.from_dict(
-            {c: p.reshape(-1) for (c, p) in zip(self.output_names, predicted)})
+
+        def format_output(data):
+            # Output can be list and it should be converted to a numpy array
+            # https://github.com/mlflow/mlflow/issues/2499
+            if isinstance(data, list):
+                data = np.asarray(data)
```
dbczumar, line 179: Can we attempt to do this unconditionally? There are other types (e.g., tuple) that we may want to convert to numpy arrays before attempting to reshape. I've confirmed that invoking `np.asarray()` on a numpy array correctly yields a numpy array:
```
In [1]: import numpy as np

In [2]: np.asarray(np.array([1,2,3]))
Out[2]: array([1, 2, 3])
```

tests/onnx/test_onnx_model_export.py
```diff
@@ -424,3 +443,17 @@ def test_model_log_without_specified_conda_env_uses_default_env_with_expected_de
         conda_env = yaml.safe_load(f)
 
     assert conda_env == mlflow.onnx.get_default_conda_env()
+
+
+@pytest.mark.large
+def test_OnnxModelWrapper_predit_handles_list_output(onnx_sklearn_model,  model_path, data):
```
dbczumar, line 449: Nit: typo - `predit` -> `predict`.

rather than referring to `OnnxModelWrapper` in the test case name, I think we should refer to the pyfunc representation of the ONNX model. I.e.:
```
test_pyfunc_predict_supports_models_with_list_outputs
```

tests/onnx/test_onnx_model_export.py
```diff
@@ -424,3 +443,17 @@ def test_model_log_without_specified_conda_env_uses_default_env_with_expected_de
...
+    """
+    https://github.com/mlflow/mlflow/issues/2499
+    User encountered issue where an sklearn model, converted to onnx, would return a list response.
+    The issue resulted in an error because MLflow assumed it would be a numpy array. Therefore,
+    the this test validates the service does not receive that error when using such a model.
+    """
+    x, y = data
+    mlflow.onnx.save_model(onnx_sklearn_model, model_path)
+    wrapper = mlflow.onnx._OnnxModelWrapper(model_path + '/model.onnx')
```
dbczumar, line 458: Rather than instantiating the wrapper in this test directly, we should instead load the model's pyfunc representation via `wrapper = mlflow.pyfunc.load_model(...)`.

dbczumar (changes requested):
@andychow-db Thanks for submitting this! Left a few comments - nothing too major

mlflow/onnx.py
andychow-db: Nice catch! Readded the reshape(-1) function.

mlflow/onnx.py
andychow-db: Done

tests/onnx/test_onnx_model_export.py
andychow-db: Done

tests/onnx/test_onnx_model_export.py
andychow-db: Done

tests/onnx/test_onnx_model_export.py
```diff
@@ -424,3 +443,17 @@ def test_model_log_without_specified_conda_env_uses_default_env_with_expected_de
...
+
+
+@pytest.mark.large
+def test_pyfunc_predict_supports_models_with_list_outputs(onnx_sklearn_model,  model_path, data):
+    """
+    https://github.com/mlflow/mlflow/issues/2499
+    User encountered issue where an sklearn model, converted to onnx, would return a list response.
+    The issue resulted in an error because MLflow assumed it would be a numpy array. Therefore,
+    the this test validates the service does not receive that error when using such a model.
+    """
```
dbczumar, line 455: I think we might have to import `mlflow.onnx` here for this test to pass

dbczumar (approved):
LGTM. Thanks @andychow-db

andychow-db:
The new test returned an error: `ImportError: cannot import name 'RunOptions'`

From researching the issue, it looks like an issue with a DLL. Therefore, I've bumped the onnxruntime version from 0.3.0 to 1.2.0 in our test dependendencies. 

https://github.com/microsoft/onnxruntime/issues/1404

PR #2743, opened 2020-04-23
Update compatibility with later versions of pyarrow (up to 0.17.0)

fhoering:

- Remove hdfs3 support because current code fails with pyarrow 0.17.0
"TypeError: connect() got an unexpected keyword argument 'driver'"

- Related arrow commit where hdfs3 is deactivated
(see https://github.com/apache/arrow/commit/4e53749097ba687afd5e000067925def2e2802c9)

- Setting ARROW_PRE_0_15_IPC_FORMAT=1 in tests for spark udf:
https://spark.apache.org/docs/latest/sql-pyspark-pandas-with-arrow.html#compatibiliy-setting-for-pyarrow--0150-and-spark-23x-24x

- Even though the tests have been updated it still keeps working with older pyarrow versions, driver will just take 'libhdfs' as default arg

## How is this patch tested?

Tested on our Hadoop cluster

mlflow now supports all versions of arrow until the current latest version 0.17.0.  In order to maintain newer arrow versions we have dropped support for the hdfs3 driver, libhdfs will be chosen as a driver for HDFS access.

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ x] Python

- [ x] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section

### DIFF
--- a/docs/source/tracking.rst
+++ b/docs/source/tracking.rst
@@ -652,13 +652,7 @@ There are also two ways to authenticate to HDFS:
 Most of the cluster contest settings are read from ``hdfs-site.xml`` accessed by the HDFS native
 driver using the ``CLASSPATH`` environment variable.
 
-Optionally you can select a different version of the HDFS driver library using:
-
-.. code-block:: bash
-
-  export MLFLOW_HDFS_DRIVER=libhdfs3
-
-The default driver is ``libhdfs``.
+The used HDFS driver is ``libhdfs``.
 
 
 Networking
--- a/mlflow/store/artifact/hdfs_artifact_repo.py
+++ b/mlflow/store/artifact/hdfs_artifact_repo.py
@@ -164,7 +164,6 @@ def hdfs_system(host, port):
     """
     import pyarrow as pa
 
-    driver = os.getenv('MLFLOW_HDFS_DRIVER') or 'libhdfs'
     kerb_ticket = os.getenv('MLFLOW_KERBEROS_TICKET_CACHE')
     kerberos_user = os.getenv('MLFLOW_KERBEROS_USER')
     extra_conf = _parse_extra_conf(os.getenv('MLFLOW_PYARROW_EXTRA_CONF'))
@@ -172,7 +171,6 @@ def hdfs_system(host, port):
     connected = pa.hdfs.connect(host=host or 'default',
                                 port=port or 0,
                                 user=kerberos_user,
-                                driver=driver,
                                 kerb_ticket=kerb_ticket,
                                 extra_conf=extra_conf)
     yield connected
--- a/tests/pyfunc/test_spark.py
+++ b/tests/pyfunc/test_spark.py
@@ -22,10 +22,7 @@ types = [np.int32, np.int, np.str, np.float32, np.double]
 
 
 def score_model_as_udf(model_uri, pandas_df, result_type="double"):
-    spark = pyspark.sql.SparkSession.builder\
-        .config(key="spark.python.worker.reuse", value=True)\
-        .master("local-cluster[2, 1, 1024]")\
-        .getOrCreate()
+    spark = get_spark_session(pyspark.SparkConf())
     spark_df = spark.createDataFrame(pandas_df)
     pyfunc_udf = spark_udf(spark=spark, model_uri=model_uri, result_type=result_type)
     new_df = spark_df.withColumn("prediction", pyfunc_udf(*pandas_df.columns))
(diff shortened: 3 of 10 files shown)


docs/source/tracking.rst
```diff
@@ -652,13 +652,7 @@ There are also two ways to authenticate to HDFS:
 Most of the cluster contest settings are read from ``hdfs-site.xml`` accessed by the HDFS native
 driver using the ``CLASSPATH`` environment variable.
 
-Optionally you can select a different version of the HDFS driver library using:
```
tomasatdatabricks, line 655: Can we still leave this as an option? Just mention that if you use anything else than libhdfs it will not work with newer pyarrow?

mlflow/store/artifact/hdfs_artifact_repo.py
```diff
@@ -164,15 +164,13 @@ def hdfs_system(host, port):
     """
     import pyarrow as pa
 
-    driver = os.getenv('MLFLOW_HDFS_DRIVER') or 'libhdfs'
```
tomasatdatabricks, line 167: can we still look if the environment is set and if it's set to something other than libhdfs? If that is the case, can we check the pyarrow version and raise an exception if it's incompatible?

tests/pyfunc/test_spark.py
```diff
@@ -52,14 +49,21 @@ def configure_environment():
     os.environ["PYSPARK_PYTHON"] = sys.executable
 
 
-@pytest.fixture
-def spark():
+def get_spark_session(conf):
+    os.environ["ARROW_PRE_0_15_IPC_FORMAT"] = "1"
```
tomasatdatabricks, line 53: Can you clarify why is this needed in a comment?

tomasatdatabricks (commented):
Hi @fhoering, thanks for the contribution!

Can we still leave an option of setting hdfs driver for people with older pyarrow?

Other than that it looks great!

docs/source/tracking.rst
fhoering: Thanks for the review. 

If we leave the option for hdfs3 it would introduce code like:
```
if pyarrow > 0.16:
    connected = pa.hdfs.connect(host=host or 'default',
                                port=port or 0,
                                user=kerberos_user,
                                kerb_ticket=kerb_ticket,
                                extra_conf=extra_conf)
else:
    connected = pa.hdfs.connect(host=host or 'default',
                                port=port or 0,
                                user=kerberos_user,
                                driver=driver,
                                kerb_ticket=kerb_ticket,
                                extra_conf=extra_conf)
```

It would make the code less clean.

Is anyone using hdfs3 for real ? What about removing and if somebody complains discuss again to eventually put it back ?

At Criteo we were using hdfs3 with https://github.com/dask/hdfs3 for a long time. It is pretty nice but it is unmaintained and doesn't support viewfs for example. 

I also made some tests and the throughput of libhdfs vs hdfs3 is mostly the same.
The only reason for hdfs is that it doesn't need the JVM mounted via JNI. So 'time to first byte' when starting a new python interpreter is better. But that is really is the only reason.

So we decided to move to libdhfs via arrow's implementation.

In a next PR I would also like to commit https://github.com/apache/arrow/pull/5957 to add viewfs support to mlflow (which only works for libhdfs and not for hdfs3)

Also nobody noticed so far that hdfs3 is broken in later pyarrow versions.

tomasatdatabricks: Those are good points. Can you include a summary of this conversation in the release notes section of the PR description? I.e. say something like - in order to maintain newer pyarrow version we have dropped support for hdfs3 driver.

tomasatdatabricks (commented):
Looks generally good. 
 - There is still a question I left on the code about why do we need to set os.environ["ARROW_PRE_0_15_IPC_FORMAT"] = "1" in a test. Can you add a comment? 
- Can you add to the release notes a sentence about dropping support for non standard hdfs drivers?

After those 2 points are addressed we can merge. Thanks!

fhoering:
>     There is still a question I left on the code about why do we need to set os.environ["ARROW_PRE_0_15_IPC_FORMAT"] = "1" in a test. Can you add a comment?

Done

>     Can you add to the release notes a sentence about dropping support for non standard hdfs drivers?
I added in the desription:
`mlflow now supports all versions of arrow until the current latest version 0.17.0. In order to maintain newer arrow versions we have dropped support for the hdfs3 driver, libhdfs will be chosen as a driver for HDFS access.`

And i've put 'breaking-change' but feel free to reclassify.

tests/pyfunc/test_spark.py
tomasatdatabricks: Awesome, thanks!

tomasatdatabricks:
Awesome, thank you @fhoering !

tomasatdatabricks:
I have restarted the failed Sagemaker test, will merge after it's green.

tomasatdatabricks (approved):
lgtm!

fhoering:
Thanks

PR #2753, opened 2020-04-26
Absolute path at server startup

### LINKED ISSUE #2671: [BUG] Wrong path for artifacts when using http server
When the server starts, the path is set via the command or created ./mlruns/
The mlflow.get_artifact_uri() function should receive the absolute path from the server, if it is a local path. Or return the FTP path, if specified in the settings.
```python
import mlflow
mlflow.set_tracking_uri("http://127.0.0.1:5000/")
mlflow.set_experiment("AutoPytorch")
with mlflow.start_run(run_name="runtest"):
    print(mlflow.get_artifact_uri()) # ./mlruns/1/fbad648257e549b0843bcb996e91a175/artifacts
```
When using mlflow.log_artifact(), the files are saved in the script folder, not the server.

The function mlflow.set_artifact_uri() is missing.

Roffild:
Closed #2671
```python
import mlflow
mlflow.set_tracking_uri("http://127.0.0.1:5000/")
mlflow.set_experiment("AutoPytorch")
with mlflow.start_run(run_name="runtest"):
    print(mlflow.get_artifact_uri()) # ./mlruns/1/fbad648257e549b0843bcb996e91a175/artifacts
```

### DIFF
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -22,7 +22,8 @@ from mlflow.server import _run_server
 from mlflow.server.handlers import initialize_backend_stores
 from mlflow.store.tracking import DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
 from mlflow.store.artifact.artifact_repository_registry import get_artifact_repository
-from mlflow.tracking import _get_store, _get_artifact_uri
+from mlflow.tracking import _get_store
+from mlflow.tracking._tracking_service.utils import _get_artifact_uri
 from mlflow.utils import cli_args, experimental
 from mlflow.utils.logging_utils import eprint
 from mlflow.utils.process import ShellCommandException
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -22,7 +22,7 @@ from mlflow.server import _run_server
 from mlflow.server.handlers import initialize_backend_stores
 from mlflow.store.tracking import DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
 from mlflow.store.artifact.artifact_repository_registry import get_artifact_repository
-from mlflow.tracking import _get_store, get_tracking_uri
+from mlflow.tracking import _get_store, _get_artifact_uri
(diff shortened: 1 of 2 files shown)


tomasatdatabricks (changes requested):
Hi @Roffild , thanks for the contribution!

I am not sure if this is the right way to fix this problem though. The default artifact uri should not point to the tracking uri as that can be set to anything. E.g. if it's set to http://127.0.01:5000 (as in your example) you would not want the artifact uri to be http://127.0.01:5000/..., or did I misunderstand something?

I believe the right fix is what you mentioned in the github issue - to return the absolute path from the file store ( and also from the sql alchemy store which I think has the same problem).

Let me know what you think. If you agree, can you update your PR to do that instead  and add tests? Thanks!

Roffild:
There is a normal API to get TRACKING_URI. For ARTIFACT_PATH only chaos in different files.
Everything is tied to this variable:
```python
DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH = "./mlruns"
```
Now the tests fail on:
```python
assert store.artifact_root_uri == "./mlruns"
```
Due to the relative path, artifacts do not work.
In the description of mlflow.get_artifact_uri()
```python
:return: An *absolute* URI referring to the specified artifact or the currently adtive run's artifact root.
```
This was not true due to "./mlruns".

file_store.py:
```python
_TRACKING_DIR_ENV_VAR = "MLFLOW_TRACKING_DIR"

def _default_root_dir():
    return get_env(_TRACKING_DIR_ENV_VAR) or os.path.abspath(DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH)
```
Nowhere else MLFLOW_TRACKING_DIR not used.

My fix closes only one problem. It is necessary to extract all the APIs for ARTIFACT_PATH into a separate file.

sylus:
I have tried this P.R. and doesn't seem to work for me, any chance on this getting worked on for when the artifact store is powered through Azure storage?

nickvazz:
Any updates?

PR #2754, opened 2020-04-27
Parameterize Python Path for Entry Points

adeandrade:

MLflow calls `python` when running `*.py` entry points. In most environments, `python` might be linked to a Python 2 installation, which is deprecated. Aliasing `python3` as `python` is discouraged as per PEP 394 (https://legacy.python.org/dev/peps/pep-0394). In this PR we change the command to be run from `python` to `python3`. We also allow an environment variable to define the path to a customized Python installation.

## How is this patch tested?

Unit tests have been updated.

Parameterize Python command used to run entry points.

- [ ] UI
- [X] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [X] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/projects/_project_spec.py
+++ b/mlflow/projects/_project_spec.py
@@ -99,7 +99,10 @@ class Project(object):
         if entry_point in self._entry_points:
             return self._entry_points[entry_point]
         _, file_extension = os.path.splitext(entry_point)
-        ext_to_cmd = {".py": "python", ".sh": os.environ.get("SHELL", "bash")}
+        ext_to_cmd = {
+            ".py": os.environ.get("PYTHON", "python3"),
+            ".sh": os.environ.get("SHELL", "bash"),
+        }
         if file_extension in ext_to_cmd:
             command = "%s %s" % (ext_to_cmd[file_extension], shlex_quote(entry_point))
             if not is_string_type(command):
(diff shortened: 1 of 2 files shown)


mlflow/projects/_project_spec.py
```diff
@@ -99,7 +99,10 @@ def get_entry_point(self, entry_point):
         if entry_point in self._entry_points:
             return self._entry_points[entry_point]
         _, file_extension = os.path.splitext(entry_point)
-        ext_to_cmd = {".py": "python", ".sh": os.environ.get("SHELL", "bash")}
+        ext_to_cmd = {
+            ".py": os.environ.get("PYTHON", "python3"),
```
pogil, line 103: Is this a best practice for setting an environment variable for which python to use? I just want to make sure this doesn't stomp on or miss some existing convention.

pogil (commented):
One test to fix, and one question

adeandrade:
I don't think I like this approach anymore. Moreover, PEP 394 has been updated to reflect their intention to point `python` to `python3`.

PR #1536, opened 2019-07-01
Add feature flag for enabling use of absolute Ajax URLs in the MLflow UI

smurching:
## What changes are proposed in this pull request?
 
Follow-up to #1413, which changes the behavior of the MLflow UI Javascript to use relative AJAX URLs when making API requests, to simplify hosting the OSS server behind an nginx proxy (e.g. you can host the UI at `http://my-proxy/mlflow-ui`, and AJAX requests will be made to `http://my-proxy/mlflow-ui/ajax-api/...` instead of `http://my-proxy/ajax-api/...`). 

This PR adds a feature flag (setting the `USE_ABSOLUTE_AJAX_URLS` environment variable to `true`, off by default) to enable the old behavior of using absolute Ajax URLs when serving the MLflow UI
 
## How is this patch tested?
 
Tested manually - in particular:
* Ran `npm run build` from the mlflow/server/js directory
* Ran the server via `mlflow server --static-prefix /mlflow-server`, verified the UI could be loaded at http://localhost:5000/mlflow-server (& that AJAX API requests were made to `http://localhost:5000/mlflow-server/ajax-api/...`)
image

* Rebuilt OSS JS with the feature flag flipped, e.g. `USE_ABSOLUTE_AJAX_URLS=true npm run build`, and relaunched the server via `mlflow server --static-prefix /mlflow-server`
* Verified that AJAX API requests were made to `http://localhost:5000/ajax-api/...` (resulting in a broken UI)
 
image

### Is this a user-facing change? 

- [x] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.
 
(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)
 
### What component(s) does this PR affect?
 
- [ ] UI
- [ ] CLI 
- [ ] API 
- [ ] REST-API 
- [ ] Examples 
- [ ] Docs
- [ ] Tracking
- [ ] Projects 
- [ ] Artifacts 
- [ ] Models 
- [ ] Scoring 
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### How should the PR be classified in the release notes? Choose one:
 
- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [ ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [ ] `rn/feature` - A new user-facing feature worth mentioning in the release notes
[...]

Issue #1120, opened 2019-04-11
Unable to run mlflow behind jupyter-server-proxy


mckev-amazon:
### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Amazon SageMaker (Amazon Linux)
- **MLflow installed from (source or binary)**: pip install
- **MLflow version (run ``mlflow --version``)**: latest
- **Python version**: 3
- **npm version (if running the dev UI): n/a
- **Exact command to reproduce**: See https://stackoverflow.com/questions/55593128/how-to-run-the-mlflow-web-based-user-interface-from-amazon-sagemaker

### Describe the problem
In a hosted environment such as Amazon SageMaker or JupyterHub, a user might want to run mlflow behind the popular [jupyter-server-proxy](https://github.com/jupyterhub/jupyter-server-proxy) extension.  This allows them to use Jupyter alongside mlflow on the same host.

Unfortunately, the way this works is that mlflow is proxied to a path such as http://myjupyter-server.com/proxy/5000 instead of http://127.0.0.1:5000.  Because there are a lot of references to paths that don't exist, the UI doesn't load.

I was almost able to get mlflow up and running, I get as far as the "Oops! Something went wrong." page, but nothing beyond that.

### Source code / logs

1. Create a SageMaker Notebook Instance.
2. !pip install mlflow
3. !mlflow
4. Open http://mynotebook.notebook.region.sagemaker.aws/proxy/5000/
5. "Oops! Something went wrong." (Niagra falls.)

freefrag:
+1 

I'd like to expose MlFlow from behind an Nginx Server, with an application path `/mlflow` and the hard-coded references to `$HOST:$PORT` disallow this.

cbg:
+1

vermanurag:
+1

cfrazier91:
+1

Jomonsugi:
+1

murphycrosby:
+1

johnDorian:
The reason why the website fails, is that there are api calls in the javascript code which are executedon the client side. The definition of these api calls are in mlflow/server/js/src/sdk/MlflowService.js - which is automatically generated based on the proto definitions. 

When you load the website on the client side, the api endpoints (/ajax-api/...) in the javascript code are combined with the hostname and the port number, but not the path of the url. 

[...]

PR #116, opened 2018-07-03
Add static-prefix option to mlflow ui and server

andrewmchen:
With this change, we start serving the UI from either `/` or `static-prefix` if it is set from the command line.

In addition, this change will change all existing routes to a hash based one: `http://localhost:5000/#/experiments/0/runs/a50907d0e0034aa790187f39b502de6c` for example

PR #2767, opened 2020-04-29
Changed HashRouter to BrowserRouter in app.js

t-henri:

Change the front end router from HashRouter to BrowserRouter.

## How is this patch tested?

Tested manually in chrome, firefox, IE11 and Edge. Navigating the UI and refreshing the pages with different routes, and checking that the history is working

Using BrowserRouter instead of HashRouter remove the /#/ between the host and the path in the url.

- [X] UI
- [ ] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [ ] Python


mlflow/server/js/package.json
```diff
@@ -58,7 +58,7 @@
     "eslint-config-airbnb-base": "~12.0.0",
     "eslint-config-standard": "10.2.1",
     "eslint-import-resolver-webpack": "0.8.4",
-    "eslint-loader": "1.9.0",
+    "eslint-loader": "^2.1.2",
```
Zangr, line 61: Is this a necessary change?

Zangr (commented):
I took a brief look and search in the change history. Changes in this PR seems to have conflicts with some of the previous relative path efforts. @t-henri could you please take a look at those PRs listed below?
https://github.com/mlflow/mlflow/pull/116
https://github.com/mlflow/mlflow/pull/1536
CC @smurching as you may have more context on those two issues?

mlflow/server/js/package.json
t-henri: No sorry it was due to my local environment, just reverted this

t-henri:
Hello @Zangr, thanks for having a look, it looks like this PR indeed conflicts with the two other ones. I'd be interested to have more context from @smurching.

smurching (commented):
@t-henri sorry to ask this naive question, but what's the motivation for this change (i.e. why switch from HashRouter to BrowserRouter)?

mlflow/server/js/src/common/utils/Utils.js
```diff
@@ -545,10 +545,7 @@ class Utils {
   }
 
   static getAjaxUrl(relativeUrl) {
-    if (process.env.USE_ABSOLUTE_AJAX_URLS === 'true') {
-      return '/' + relativeUrl;
-    }
-    return relativeUrl;
```
smurching, line 551: Previously, we allowed using relative URLs here to enable use cases like https://github.com/mlflow/mlflow/issues/1120 - what's the motivation for removing support for relative URLs here?

t-henri:
Hello @smurching , we discussed it a bit on slack already but my use case is that I use a technology called openam to setup authentication and this technology doesn't support # in urls, and with hash router we always have a # in url.
Also, it is specified in the documentation that using hashrouter is not recommanded and that every application should aim to use BrowserRouter.

mlflow/server/js/src/common/utils/Utils.js
t-henri: Sorry I didn't have this use case in mind, but relative urls don't work with BrowserRouter that's why I removed them, but I can put them back if needed.

PR #2775, opened 2020-05-01
Add schema and example logging to save_model methods of all supported flavors.

tomasatdatabricks:
Model schema followup:
1. save schema and examples in save_model (previsouly only done in log model)
2. fix bug with default Model value in save_model of flavors (the default value is shared between invocations)

## How is this patch tested?
Added tests for all flavors.

- [ ] UI
- [ ] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [ ] Python


pogil:
I know this is still WIP - let's be sure we also update docs and add release notes for this feature! Maybe also change the PR title to describe what the PR does

tomasatdatabricks:
Thanks @pogil. I have updated the title. I don't think this PR needs release notes, those should come from the previous schema PR (this one is a more of a fix).

mlflow/fastai.py
```diff
@@ -90,6 +111,13 @@ def save_model(fastai_learner, path, conda_env=None, mlflow_model=Model(), **kwa
...
 
+    if mlflow_model is None:
+        mlflow_model = Model()
+    if signature is not None:
+        mlflow_model.signature = signature
+    if input_example is not None:
+        save_example(mlflow_model, input_example, path)
+
     # Save an Learner
     fastai_learner.export(model_data_path, **kwargs)
```
smurching, line 122: No action needed, just a note that I confirmed that `signature` & `input_example` are not valid arguments to [export](https://docs.fast.ai/basic_train.html#Learner.export)

mlflow/fastai.py
```diff
@@ -77,6 +79,25 @@ def save_model(fastai_learner, path, conda_env=None, mlflow_model=Model(), **kwa
...
+                      The model signature can be :py:func:`inferred <mlflow.models.infer_signature>`
+                      from datasets with valid model input (e.g. the training dataset) and valid
+                      model output (e.g. model predictions generated on the training dataset),
+                      for example:
+
+                      .. code-block:: python
+
+                        from mlflow.models.signature import infer_signature
+                        train = df.drop_column("target_label")
+                        signature = infer_signature(train, model.predict(train))
```
smurching, line 94: Not sure that this example will make sense to fastai users (where there's no `model.predict(pandas_df)` API) - maybe we could make it `fastai` specific (preferable but more work) or rewrite it to be generic to ML frameworks (i.e. write something like):

```
                        from mlflow.models.signature import infer_signature
                        train_pandas_df = train_pandas_df.drop_column("target_label")
                        output_pandas_df = ... # compute model predictions & convert to a pandas dataframe
                        signature = infer_signature(input_pandas_df, output_pandas_df)
```

I do think per-framework usage examples would be ideal (make it easiest for users of the framework to actually pass a signature/example), but could come in a follow-up PR.

mlflow/fastai.py
```diff
@@ -77,6 +79,25 @@ def save_model(fastai_learner, path, conda_env=None, mlflow_model=Model(), **kwa
                             ]
                         }
     :param mlflow_model: MLflow model config this flavor is being added to.
+
+    :param signature: (Experimental) :py:class:`ModelSignature <mlflow.models.ModelSignature>`
+                      describes model input and output :py:class:`Schema <mlflow.types.Schema>`.
+                      The model signature can be :py:func:`inferred <mlflow.models.infer_signature>`
+                      from datasets with valid model input (e.g. the training dataset) and valid
```
smurching, line 86:
```suggestion
                      from datasets with valid model input (e.g. the training dataset with target column omitted) and valid
```

mlflow/keras.py
```diff
@@ -116,6 +118,24 @@ def save_model(keras_model, path, conda_env=None, mlflow_model=Model(), custom_o
                          attempt to infer the Keras module based on the given model.
     :param kwargs: kwargs to pass to ``keras_model.save`` method.
 
+    :param signature: (Experimental) :py:class:`ModelSignature <mlflow.models.ModelSignature>`
```
smurching, line 121: No action needed, just confirmed that `signature` and `input_example` are not already arguments to [`keras_model.save`](https://www.tensorflow.org/api_docs/python/tf/keras/Model#save) - there's actually an argument called `signatures`, but no `signature` :P

mlflow/fastai.py
```diff
@@ -77,6 +79,25 @@ def save_model(fastai_learner, path, conda_env=None, mlflow_model=Model(), **kwa
                             ]
                         }
     :param mlflow_model: MLflow model config this flavor is being added to.
+
+    :param signature: (Experimental) :py:class:`ModelSignature <mlflow.models.ModelSignature>`
```
smurching, line 83: BTW, do we also plan to document these params for `log_model` in a follow-up (IMO we should)?

mlflow/gluon.py
```diff
@@ -114,6 +115,26 @@ def save_model(gluon_model, path, mlflow_model=Model(), conda_env=None):
...
+                      from datasets with valid model input (e.g. the training dataset) and valid
+                      model output (e.g. model predictions generated on the training dataset),
+                      for example:
+
+                      .. code-block:: python
+
+                        from mlflow.models.signature import infer_signature
+                        train = df.drop_column("target_label")
+                        signature = infer_signature(train, model.predict(train))
+    :param input_example: (Experimental) Input example provides one or several instances of valid
```
smurching, line 130: It would be nice if there were some way to get the list of supported types to show up in this docstring, e.g. when I ran 

```
>>> import mlflow.gluon
>>> help(mlflow.gluon.save_model)
```

I couldn't actually see what valid types I could pass for `input_example`. Do you know offhand if there's a way? If not I can help look for one

mlflow/models/model.py
```diff
@@ -22,18 +22,18 @@ class Model(object):
     """
 
     def __init__(self, artifact_path=None, run_id=None, utc_time_created=None, flavors=None,
-                 signature: ModelSignature = None, input_example: ModelInputExample = None,
+                 signature: ModelSignature=None, input_example: Dict[str, Any]=None,
```
smurching, line 25: Why change this type away from `ModelInputExample`?

mlflow/models/model.py
```diff
@@ -22,18 +22,18 @@ class Model(object):
...
         if run_id:
             self.run_id = run_id
             self.artifact_path = artifact_path
         self.utc_time_created = str(utc_time_created or datetime.utcnow())
         self.flavors = flavors if flavors is not None else {}
-        if signature is not None:
-            self.signature = signature
-        if input_example is not None:
-            self.input_example = input_example
+        self._signature = None
```
smurching, line 33: Hm don't we want to assign `self._signature` to `signature` here?

mlflow/models/model.py
```diff
@@ -86,9 +106,7 @@ def from_dict(cls, model_dict):
         return cls(**model_dict)
 
     @classmethod
-    def log(cls, artifact_path, flavor, registered_model_name=None,
-            signature: ModelSignature = None, input_example: ModelInputExample = None,
```
smurching, line 90: What's the motivation for removing `signature` and `input_examples` as explicit params to this method btw?

mlflow/models/model.py
```diff
@@ -124,15 +142,8 @@ def log(cls, artifact_path, flavor, registered_model_name=None,
             local_path = tmp.path("model")
             run_id = mlflow.tracking.fluent._get_or_start_run().info.run_id
             mlflow_model = cls(artifact_path=artifact_path, run_id=run_id)
-            if signature is not None:
```
smurching, line 127: To make sure I understand - before, we could only save signature/examples via `log_model`, which saved them here (the operation of saving a signature/example is flavor-agnostic). Now, we've enabled saving signature/examples in `save_model`, which doesn't have a common code-path across flavors (i.e. each flavor defines its own `save_model` implementation), hence why we duplicate the code for saving signature/examples across the flavors.

mlflow/models/utils.py
```diff
@@ -104,3 +105,41 @@ def save(self, parent_dir_path: str):
         """Save the example as json at ``parent_dir_path``/`self.info['artifact_path']`.  """
         with open(os.path.join(parent_dir_path, self.info["artifact_path"]), "w") as f:
             json.dump(self.data, f, cls=NumpyEncoder)
+
+
+def save_example(mlflow_model: Model, input_example: ModelInputExample, path: str):
```
smurching, line 110: Nit, should these be private methods (prefixed with `_`)? We've run into issues in the past where undocumented but public APIs (e.g. APIs that aren't prefixed with underscores) are assumed to be stable / meant for public use

smurching (commented):
Left some questions/comments, thanks @tomasatdatabricks!

mlflow/models/model.py
tomasatdatabricks: The input example here is example metadata rather than the actual example. I admit it is confusing. I can rename it or create a type for it to make it more clear.

mlflow/models/model.py
tomasatdatabricks: yeah, this is confusing. I wanted to use setter but the linter was complaining that _signature was not defined in the __init__ so I just did declare it as None before assigning the user value vis the setter. I will either assign directly to _signature here (the setter does not really do anything else right now) or remove the _signature assignment and disable the warning.

mlflow/models/model.py
tomasatdatabricks: They are not used by the log function anymore, instead they are passed to the flaver.save_model. So, if the flavor does not support signature / example, neither would Model.log_model.

mlflow/models/model.py
tomasatdatabricks: yes, exactly.

mlflow/models/utils.py
tomasatdatabricks: yes that makes sense.

mlflow/fastai.py
tomasatdatabricks: what do you mean? You mean to update the models.rst docs?

mlflow/fastai.py
tomasatdatabricks: I agree that per framework example would be the best but I don't think it's super important. 
Showing more generic way of generating model predictions makes sense.

mlflow/gluon.py
tomasatdatabricks: Hmm, if I do help(mlflow.gluon.save_model) it expands the ModelInputExample into Union:

..., input_example: Union[pandas.core.frame.DataFrame, numpy.ndarray, dict, list] = None, ...

mlflow/fastai.py
smurching: Basically, we probably want to update the docs for [`mlflow.fastai.log_model`](https://github.com/mlflow/mlflow/blob/66a54d5aa5829abe06b8020a36e328ccb031397f/mlflow/fastai.py#L109) & analogous methods in other flavors to mention that `signature` and `input_example` are valid arguments (& actually in the fastai case, that they won't be passed to `fastai.Learner.export` [as the docs currently state](https://github.com/mlflow/mlflow/blob/66a54d5aa5829abe06b8020a36e328ccb031397f/mlflow/fastai.py#L135))

smurching: Oops, sorry I somehow missed that this PR also updates the log_model docstrings - nvm!

mlflow/gluon.py
smurching: Ah got it, yep I see it now (sorry for the noise, noob mistake getting used to seeing types in the actual Python method signature vs in documentation for each individual param :P)

mlflow/models/model.py
smurching: Thanks for clarifying, maybe we can also add a docstring for the constructor to explain what the passed-in values are? I know there isn't an existing one, but would be nice given that this class [is publicly documented](https://mlflow.org/docs/latest/python_api/mlflow.models.html#mlflow.models.Model)

mlflow/models/model.py
smurching: Got it makes sense! I'll actually open a follow-up PR to update this docstring to state a bit more clearly that `Model.log` calls the passed-in flavor's `save_model` & that [``kwargs``](https://github.com/mlflow/mlflow/pull/2775/files#diff-c470a0d352438df0c82bed9873047b3bR140) are passed to the flavor's `save_model` method.

mlflow/models/model.py
smurching: Also happy to do that in a follow-up^, as this PR otherwise LGTM

smurching (approved):
LGTM, besides one small comment around a docstring update that we can address in a follow-up IMO

mlflow/models/model.py
tomasatdatabricks: Makes sense!

mlflow/models/model.py
tomasatdatabricks: Cool, thx!

tomasatdatabricks:
Thanks for reviewing Sid!

PR #2779, opened 2020-05-03
Support selecting different parameters in the parallel coordinates plot

### LINKED ISSUE #2273: [FR] Parallel Coordinates Plot  - Select/Deselect all parameters/metrics
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md)
for information on what types of issues we address.

Please fill in this template and do not delete it unless you are sure your issue is outside its scope.

-------
## Describe the proposal
When comparing runs with Parallel Coordinates Plot it shows all parameters by default. There is no option to select or deselect all parameters or metrics.  A button to select/deselect all parameters/metrics should be added.

### Motivation
Most experiments/runs have many parameters, but only few are changed between runs or are relevant for run comparison analysis. Since all parameters are enabled in Parallel Coordinates Plot by default, it costs much time to deselect irrelevant parameters. Furthermore, having many parameters enabled in Parallel Coordinates Plots, the UI is less responsive, which increases time further to set up the plot as wanted. 

### Proposed Changes
Add buttons for parameters and metrics to toggle selection of parameters/metrics.

harupy:

- Support selecting different parameters in the parallel coordinates plot.
- Resolves #2273

select-diff

## How is this patch tested?

(Details)

- [x] UI
- [ ] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
+++ b/mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
@@ -85,7 +85,7 @@ export class ParallelCoordinatesPlotPanel extends React.Component {
     if (indeterminate) {
       this.setState({
         selectedParamKeys: allParamKeys,
-        selectAll: { checked: true },
+        selectAll: { indeterminate: false, checked: true },
         selectDiff: this.getSelectDiffState(allParamKeys),
       });
     } else if (checked) {
--- a/mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
+++ b/mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import ParallelCoordinatesPlotView from './ParallelCoordinatesPlotView';
-import ParallelCoordinatesPlotControls from './ParallelCoordinatesPlotControls';
+import { ParallelCoordinatesPlotControls } from './ParallelCoordinatesPlotControls';
 import {
   getAllParamKeysByRunUuids,
   getAllMetricKeysByRunUuids,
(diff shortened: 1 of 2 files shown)


dbczumar:
@Zangr @gioa Can you provide some input on the proposed UI design here? This appears to differ a bit from the mocks shown in the feature request back in January (https://github.com/mlflow/mlflow/issues/2273#issuecomment-572233183).

gioa:
@dbczumar Thanks for pinging!

@harupy Is there any reason not doing what's proposed before? It seems a little weird to group the parameters in the two sections (different/constant parameters). The common use cases are select all, select different parameters, or select individual parameters manually. With the one you implement here, users need to think about what each section means; for selecting all, they need to check both. It works, but breaks the design principle of "Don't make me think".

harupy:
@gioa Thanks for the comment. I fixed the design. What do you think?


fixed

gioa:
@harupy It looks great, thank you for the fix!

dbczumar:
@harupy Is this ready for review? I noticed the PR is still marked `Draft`; looks like there are some minor JS test failures. More than happy to review this at any time!

harupy:
@dbczumar Changed the status to `Ready to review`.

mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
```diff
@@ -25,36 +25,126 @@ export class ParallelCoordinatesPlotPanel extends React.Component {
...
+    diffParamKeys: PropTypes.arrayOf(String).isRequired,
   };
 
-  state = {
-    // Default to select all parameters
-    selectedParamKeys: this.props.sharedParamKeys,
-    // Default to select the first metric key.
-    // Note that there will be no color scaling if no metric is selected.
-    selectedMetricKeys: this.props.sharedMetricKeys.slice(0, 1),
+  constructor(props) {
```
aarondav, line 32: Coolio! Can we make "Select Diff Only" the default, _if_ there are non-constant keys?

aarondav:
A little weird, if all parameters differ, we have both checked, and selecting either one checks or unchecks the other:

image

This is mainly weird since it says "Select Diff Only", and I would expect "Only" to be exclusive. Maybe just "Differing Values" would be sufficient?

mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
```diff
@@ -25,36 +25,126 @@ export class ParallelCoordinatesPlotPanel extends React.Component {
     sharedParamKeys: PropTypes.arrayOf(String).isRequired,
     // An array of metric keys shared by all runs
     sharedMetricKeys: PropTypes.arrayOf(String).isRequired,
+    // An array of different parameter keys across runs
```
aarondav, line 28: I was a little confused by this, since it seems to be in opposition of "sharedParamKeys", but it doesn't mean param keys which don't exist in some runs. Maybe this should say

"A subset of sharedParamKeys where the values of the parameters differ between runs."

mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
```diff
@@ -25,36 +25,126 @@ export class ParallelCoordinatesPlotPanel extends React.Component {
...
+      selectAll: this.getSelectAllState(paramValues),
+      selectDiff: this.getSelectDiffState(paramValues),
+    });
   };
 
   handleMetricsSelectChange = (metricValues) => {
     this.setState({ selectedMetricKeys: metricValues });
   };
 
+  handleSelectAllChange = () => {
```
aarondav, line 81: I expected this method to just say

if-checked, set selectedParamKeys=[] otherwise selectedParamKeys= allParamKeys.

I think there should be some other method which recomputes the selectAll/selectDiff derived values. For example, what if `this.selectAllState` was just a method that read `selectedParamKeys` from the current state?

aarondav:
Do we have any unit tests we can add for this component?

This feature is awesome, btw, really should make the parallel coordinates plot easier to use, especially once we make the diff-only default!

mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
harupy: I think we can. I'll add this change.

mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
harupy: That is better. I'll fix the comment.

mlflow/server/js/src/experiment-tracking/components/ParallelCoordinatesPlotPanel.js
harupy: That makes sense. I'll fix this.

harupy:
@aarondav Thanks for the comments!

> A little weird, if all parameters differ, we have both checked, and selecting either one checks or unchecks the other. This is mainly weird since it says "Select Diff Only", and I would expect "Only" to be exclusive.

Could you elaborate on this?

> Do we have any unit tests we can add for this component?

I'll add unit tests.

PR #2780, opened 2020-05-03
Remove imports from `__future__`

harupy:

Remove imports from `__future__` because we no longer support Python 2.

## How is this patch tested?

Verified `git grep "from __future__ import" | wc -l` returns `0`.

- [ ] UI
- [ ] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/examples/hyperparam/train.py
+++ b/examples/hyperparam/train.py
@@ -6,8 +6,6 @@ Validation data is used to select the best hyperparameters, test set performance
 at epochs which improved performance on the validation dataset. The model with best validation set
 performance is logged with MLflow.
 """
-from __future__ import print_function
-
 import warnings
 
 import math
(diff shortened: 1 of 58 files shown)


dbczumar:
@harupy Thanks so much for filing this - makes sense to remove uses of `__future__`; I'll go ahead and review this shortly.

dbczumar (approved):
LGTM! I've confirmed that MLflow doesn't rely on any `__future__` modules that are optional in 3.x versions of Python (e.g., `generator_stop` and `annotations`, as listed here: https://docs.python.org/3/library/__future__.html). Accordingly, it's safe to remove all existing imports from `__future__`, as they were made mandatory in Python 3.0. Thanks @harupy!

PR #2782, opened 2020-05-04
Support HDFS federation viewfs

fhoering:

Support logging artifacts with Hadoop viewfs using the same fix as in Apache arrow:
https://github.com/apache/arrow/pull/5957

Link to viewfs in Hadoop:
https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/ViewFs.html

## How is this patch tested?

- tested on our Criteo Hadoop cluster having viewfs set up

- [] Yes. Give a description of this change to be included in the release notes for MLflow users.

If you are using HDFS you can now use federation with viewfs to log artifacts.

- [ ] UI
- [ ] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [x ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [x ] Python

- [x ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section

### DIFF
--- a/mlflow/store/artifact/hdfs_artifact_repo.py
+++ b/mlflow/store/artifact/hdfs_artifact_repo.py
@@ -189,6 +189,7 @@ def _resolve_connection_params(artifact_uri):
 
     return parsed.scheme, parsed.hostname, parsed.port, parsed.path
 
+
 def _resolve_base_path(path, artifact_path):
     if path == artifact_path:
         return path
--- a/mlflow/store/artifact/hdfs_artifact_repo.py
+++ b/mlflow/store/artifact/hdfs_artifact_repo.py
@@ -20,7 +20,8 @@ class HdfsArtifactRepository(ArtifactRepository):
     """
 
     def __init__(self, artifact_uri):
-        self.host, self.port, self.path = _resolve_connection_params(artifact_uri)
+        self.scheme, self.host, self.port, self.path = _resolve_connection_params(
+            artifact_uri)
         super(HdfsArtifactRepository, self).__init__(artifact_uri)
 
     def log_artifact(self, local_file, artifact_path=None):
@@ -31,7 +32,7 @@ class HdfsArtifactRepository(ArtifactRepository):
         """
         hdfs_base_path = _resolve_base_path(self.path, artifact_path)
 
-        with hdfs_system(host=self.host, port=self.port) as hdfs:
+        with hdfs_system(scheme=self.scheme, host=self.host, port=self.port) as hdfs:
             _, file_name = os.path.split(local_file)
             destination = posixpath.join(hdfs_base_path, file_name)
             with hdfs.open(destination, 'wb') as output:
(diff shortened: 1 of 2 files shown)


dbczumar:
@fhoering Thanks for filing this; I'll go ahead and review it shortly. Do you have any tips for setting up and testing ViewFS?

fhoering:
I guess this is the general problem with Hadoop and why it is declining. Nothing is easy.
https://github.com/apache/arrow/pull/5957
I

fhoering:
I guess this is the general problem with Hadoop, nothing is easy. 

When doing https://github.com/apache/arrow/pull/5957 I had a look at the integration tests. 

They use an external docker image (seems like it is this one https://github.com/parrot-stream/docker-impala) for hdfs with a very basic hdfs config. So it would have meant adding viewfs support to this which basically means following the viewfs documentation and setting up multiple namenodes. So, as I have never done it either (as it is already set up on our cluster) and given that it was a considerable amount of work I decided not to do it.

In general, it would be nice also to have real integration tests with HDFS for mlflow. This project seems pretty nice: https://github.com/jcrist/hadoop-test-cluster, but viewfs also has to be configured as a feature. I still have this on my todo list to add hadoop integration tests to our github projects that use Hadoop. So then it would be easy to also just upstream some hdfs integration tests in mlflow.

So to sum it up. The only solution I see is setting up a hadoop cluster with docker compose and following the viewfs documentation until it works.

To come back to the fix here, basically the trick is to inject the schema to libhdfs driver. I can confirm it doesn't create a regression because pyarow integrations tests still work, it works on our cluster and in https://github.com/apache/arrow/pull/5957 there is a link to the Hadoop libhdfs code to see how the schema is handled.

fhoering:
@dbczumar 

OK. After having asked the experts it turns out it is quite easy. You can just reuse one namenode and add the mappings to core-site.xml:
```
    <property>
        <name>fs.defaultFS</name>
        <value>viewfs://root</value>
    </property>
    <property>
        <name>fs.viewfs.mounttable.root.link./tmp</name>
        <value>hdfs://CLUSTER/tmp</value>
    </property>
    <property>
        <name>fs.viewfs.mounttable.root.link./user</name>
        <value>hdfs://CLUSTER/user</value>
    </property>
    <property>
        <name>fs.viewfs.mounttable.root.link./var</name>
        <value>hdfs://CLUSTER/var</value>
    </property>
    <property>
        <name>fs.viewfs.rename.strategy</name>
        <value>SAME_FILESYSTEM_ACROSS_MOUNTPOINT</value>
    </property>
```

dbczumar (approved):
LGTM @fhoering ! Sorry for the delay - I tested this out with single-node HDFS and ViewFS. I've also confirmed that HDFS URIs are always of the form `scheme://authority/path` (as discussed here: https://hadoop.apache.org/docs/r2.4.1/hadoop-project-dist/hadoop-common/FileSystemShell.html), so the URI construction on line 174 appears to be safe.

Thank you for your contribution!

fhoering:
Thanks

PR #2788, opened 2020-05-05
Utilize model signature during model scoring.

tomasatdatabricks:
1. Introduce PyFuncModel - wrapper around pyfunc model implementation (existing pyfunc models) and model metadata. Pyfunc models will now be loaded as instances of PyFuncModel. 

2. Add type enforcement to pyfunc predict. PyFuncModel will enforce input schema for models that contain input signature. The schema enforcement checks column names as well as column types. 

**Column Name Enforcement:**
- Extra columns (not delcared in input schema) are ingored
-  Error is raised if some expected input columns are missing.
-  The input is reordered to match the order in input schema.

**Type Enforcement:** Types must match exactly, except:
- we allow upcasting for compatible types, so int -> long and float -> double 
- binary and strings can be passed as objects. 

3. Utilize model signature in json parsing inside of scoring_server.
Currently we use pandas json parser (pd.read_json). If the model includes model signature, we will pass data type hints based on the input schema to the parser. The parser is expected (but not required) to respect the types if possible. The parser currently does not enforce the types. 

4. Autofill column names in spark_udf from with names from input schema if it is available. 

## How is this patch tested?
Unit tests.

## Release Notes
The following MLflow python model apis have been updated with model signatures: 
 - Python function models loaded with mlflow.pyfunc.load_model will contain model metadata in addition to the model implementation. The metadata can be used to get additional details about the model such as it's signature.
- Python function models that include model signature will enforce model input schema in predict. Schema enforcement includes column names (all input columns specified in the schema must be provided, any extra columns not specified in schema will be ignored, input is reordered according to the ordering declared in schema) and colum types (incompatible column types will result in an error). 
- MLflow models loaded as spark_udf via mlflow.pyfunc.spark_udf will autofill column names from the model input schema if available.  

 

MLflow python model apis have been updated to use model signatures if available. 

- [ ] UI
- [ ] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [x] Python


mlflow/azureml/__init__.py
```diff
@@ -308,12 +308,10 @@ def deploy(model_uri, workspace, deployment_config=None, service_name=None, mode
 
     model_pyfunc_conf, model = _load_pyfunc_conf_with_model(model_path=absolute_model_path)
     model_python_version = model_pyfunc_conf.get(pyfunc.PY_VERSION, None)
-    run_id = None
-    run_id_tag = None
-    try:
-        run_id = model.run_id
-        run_id_tag = run_id
-    except AttributeError:
```
smurching, line 316: Is it safe to remove this check?

mlflow/pyfunc/__init__.py
```diff
@@ -265,7 +268,51 @@ def _load_model_env(path):
     return _get_flavor_configuration(model_path=path, flavor_name=FLAVOR_NAME).get(ENV, None)
 
 
-def load_model(model_uri, suppress_warnings=True):
+PyFuncOutput = Union[pandas.DataFrame, pandas.Series, np.ndarray, list]
+
+
+class PyFuncModel(object):
```
smurching, line 274: Do we want this class to be public?

mlflow/pyfunc/__init__.py
```diff
@@ -265,7 +268,51 @@ def _load_model_env(path):
...
+    """
+    MLflow 'python function' model.
+
+    Wrapper around model implementation and metadata loaded with py:func:`mlflow.pyfunc.load_model`.
```
smurching, line 278:
```suggestion
    Wrapper around model implementation and metadata. Instances of this class are returned from py:func:`mlflow.pyfunc.load_model`.
```

mlflow/pyfunc/__init__.py
```diff
@@ -265,7 +268,51 @@ def _load_model_env(path):
...
+    The model implementation can be arbitrary python object returned by the loader_module.
```
smurching, line 279: Wonder if it's worth trying to explain this out a bit more for the end-user (maybe we don't need to, since the class is generally constructed for them / they aren't expected to construct it themselves), e.g. via something like:

```suggestion
    ``model_impl`` can be any Python object that implements the `Pyfunc interface <https://mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#pyfunc-inference-api>`_, and is loaded in `load_model` by invoking the appropriate method within ``loader_module``.
    
    ``model_meta`` is loaded by `:py:func:load_model` from the model metadata (MLmodel) file
```

mlflow/pyfunc/__init__.py
```diff
@@ -438,26 +492,45 @@ def spark_udf(spark, model_uri, result_type="double"):
...
     def predict(*args):
         model = SparkModelCache.get_or_load(archive_path)
-        schema = {str(i): arg for i, arg in enumerate(args)}
+        input_schema = model.metadata.get_input_schema()
         pdf = None
+
+        def convert_type(series: pandas.Series, type_: np.dtype):
+            if series.dtype == type_:
+                return series
+            return series.astype(type_, errors="ignore")
```
smurching, line 501: Should we consider using errors="coerce" here to convert invalid values into NaNs as described in [this SO post](https://stackoverflow.com/a/50858819)? Or actually, maybe we should just throw if the series cannot be converted into the expected type?

mlflow/pyfunc/__init__.py
```diff
@@ -438,26 +492,45 @@ def spark_udf(spark, model_uri, result_type="double"):
...
+                pdf = pandas.DataFrame(data=data, columns=[str(i) for i in range(len(args))])
+            else:
+                if len(args) != len(input_schema.columns):
+                    raise MlflowException("Model input is not compatible with model schema. "
+                                          "Expected  {0} input columns, got {1}".format(
+                                            len(input_schema.columns), len(args)))
+                names = input_schema.column_names()
+                types = input_schema.numpy_types()
+                data = dict([(names[i], convert_type(x, types[i])) for i, x in enumerate(args)])
+                pdf = pandas.DataFrame(data=data, columns=names[:len(args)])
```
smurching, line 522: Nit: why do we slice `names` here? Shouldn't it be the same length as `args`?

mlflow/utils/proto_json_utils.py
```diff
@@ -97,13 +95,26 @@ def _dataframe_from_json(path_or_str, schema: Schema = None,
...
+        df = pd.read_json(path_or_str, orient=pandas_orient, dtype=dtypes)
+
+        names = schema.column_names()
+        expected_names = set(names)
+        actual_names = set(df.columns)
+        if actual_names != expected_names:
+            missing_cols = expected_names - actual_names
+            extra_columns = actual_names - expected_names
+            raise MlflowException("Incompatible data. {0} {1}".format(
+                "{}.".format("Missing columns {}".format(missing_cols) if missing_cols else ""),
```
smurching, line 108:
```suggestion
                "Missing columns {}".format(missing_cols) if missing_cols else "",
```

Nit: remove redundant call to `.format`? Same with the next line

mlflow/pyfunc/__init__.py
```diff
@@ -438,26 +492,45 @@ def spark_udf(spark, model_uri, result_type="double"):
...
         if pdf is None:
-            # Explicitly pass order of columns to avoid lexicographic ordering (i.e., 10 < 2)
-            columns = [str(i) for i, _ in enumerate(args)]
-            pdf = pandas.DataFrame(schema, columns=columns)
+            if input_schema is None:
+                data = dict([(str(i), x) for i, x in enumerate(args)])
+                # Explicitly pass order of columns to avoid lexicographic ordering (i.e., 10 < 2)
+                pdf = pandas.DataFrame(data=data, columns=[str(i) for i in range(len(args))])
+            else:
+                if len(args) != len(input_schema.columns):
```
smurching, line 515: To confirm, are we sure we want to throw if there are extra columns in the input? e.g. should we worry about the potential use case where you're applying a model as part of a pipeline & there are some extra columns that are unused by the current model but are used further down the pipeline?

smurching: Same goes for the serving use case, e.g. the protobuf convention is to ignore unknown fields in REST API input, so maybe we should do the same?

mlflow/pyfunc/__init__.py
tomasatdatabricks: Yes we do. I think it makes sense to make this the official pyfunc  interface. It has consequences though - e.g. some of our tests were passing in model_input=..., where model_input is only known to PythonModel. This will no longer work. We could consider passing in generic arguments instead to preserve such workloads, but I think it's better to enforce unified interface.

mlflow/pyfunc/__init__.py
tomasatdatabricks: That's a good question. I think both raising exception or conversion to NaNs are pretty invasive changes so if we want to do it we should do it before releasing this. Alternatively, we can add this behavior later protected by user controlled flag - e.g. setting --enforce_schema when loading/deploying model. 

I prefer the latter for the following reasons: 
a) We do not want to add barriers to starting using schemas - e.g. to avoid situation when adding schema breaks your workflow. 
b) If we ever record schema automatically (e.g. as part of autologging or autoextract it from tf models) we could break models that would otherwise work.

mlflow/pyfunc/__init__.py
tomasatdatabricks: Oh, good catch. Originally I was allowing a subset of the columns but I decided it was better to require exact match. I'll remove the slice.

mlflow/utils/proto_json_utils.py
tomasatdatabricks: It's not redundant.

mlflow/pyfunc/__init__.py
tomasatdatabricks: Hmm I did not get the pipeline bit - these are the args passed to the model itself, the subsequent pipeline elements can still receive additional stuff?

As for whether we should follow protobuf convention - that's a good question. Maybe we should.  Also, I've realized this is a similar question to whether we should raise exception if there are incompatible types and we should be consistent in handling it.

mlflow/pyfunc/__init__.py
```diff
@@ -265,7 +268,58 @@ def _load_model_env(path):
...
+
+    Wrapper around model implementation and metadata. This class is not meant to be constructed
+    directly. Instead, instances of this class are constructed and returned from
+    py:func:`mlflow.pyfunc.load_model`.
+
+    ``model_impl`` can be any Python object that implements the `Pyfunc interface
+    <https://mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#pyfunc-inference-api>`_, and is
+    by invoking the model's ``loader_module``.
+    
+    ``model_meta`` is loaded by `:py:func:load_model` from the model metadata (MLmodel) file
```
aarondav, line 286: Somewhere can we link to what model_meta is? Is it just the JSON form of the MLmodel file or some other type?

tomasatdatabricks: It's the Model object. I'll update the docstring.

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,117 @@ def _load_model_env(path):
...
+    def predict(self, data: pandas.DataFrame) -> PyFuncOutput:
+        """
+        Generate model predictions.
+        :param data: Model input as pandas.DataFrame.
+        :return: Model predictions as one of pandas.DataFrame, pandas.Series, numpy.ndarray or list.
+        """
+        input_schema = self._model_meta.get_input_schema()
+
+        if input_schema is not None:
+            def convert_type(name, values: pandas.Series, t: DataType):
```
aarondav, line 306: Could we put this in a private helper? A bit hard to read deeply nested conditionals in python.

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,117 @@ def _load_model_env(path):
...
+                if values.dtype == np.object and t not in (DataType.binary, DataType.string):
+                    values = values.infer_objects()
+
+                if values.dtype in (t.to_pandas(), t.to_numpy()):
+                    return values
+
+                if t == DataType.binary and values.dtype.kind == t.binary.to_numpy().kind:
+                    #  NB: bytes in numpy have variable itemsize and would fail the upcast check.
```
aarondav, line 314: This seems to be a forward reference to some logic. Maybe we could say earlier what we plan to do given an input schema?

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,117 @@ def _load_model_env(path):
...
+        :return: Model predictions as one of pandas.DataFrame, pandas.Series, numpy.ndarray or list.
+        """
+        input_schema = self._model_meta.get_input_schema()
+
+        if input_schema is not None:
+            def convert_type(name, values: pandas.Series, t: DataType):
+                if values.dtype == np.object and t not in (DataType.binary, DataType.string):
+                    values = values.infer_objects()
+
+                if values.dtype in (t.to_pandas(), t.to_numpy()):
```
aarondav, line 310: This is the short circuit in the case where the type is already appropriate, right? Maybe we can comment that.

mlflow/pyfunc/__init__.py
tomasatdatabricks: I am not sure what do you mean? The reason this is necessary here is that (and tbh I don't think this happens very often) if you add a column encoded as numpy bytes, the data type will be S<itemsize> where itemsize is the length of the longest item. The itemsize of our numpy type for binary is 0, so the upcast logic would not work here.

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,117 @@ def _load_model_env(path):
...
+                        raise MlflowException(
+                            "Failed to convert column {0} from type {1} to {2}.".format(
+                              name, values.dtype, t)
+                        )
+
+                numpy_type = t.to_numpy()
+                is_compatible_type = values.dtype.kind == numpy_type.kind
+                is_upcast = values.dtype.itemsize <= numpy_type.itemsize
+                if is_compatible_type and is_upcast:
+                    # conversions between compatible types and ints -> floats are allowed.
```
aarondav, line 336: Wouldn't `is_compatible_type` be false if one is int and one is float?

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,117 @@ def _load_model_env(path):
...
+                    return values.astype(numpy_type, errors="raise")
+                else:
+                    # NB: conversion between incompatible types (e.g. floats -> ints or
```
aarondav, line 339: or double -> float, may be good to point out

mlflow/pyfunc/__init__.py
tomasatdatabricks: yup, I will add a comment.

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,117 @@ def _load_model_env(path):
...
+            col_types = input_schema.column_types()
+            expected_names = set(input_schema.column_names())
+            actual_names = set(data.columns)
+            missing_cols = expected_names - actual_names
+            extra_columns = actual_names - expected_names
+            if missing_cols:
+                message = "Model input is missing columns {0}.".format(missing_cols)
+                raise MlflowException(message)
+            if extra_columns:
+                _logger.warning("Ignoring unexpected columns %s", extra_columns)
```
aarondav, line 356: I don't think we should bother printing this. It could actually be more expensive than doing the actual prediction.

mlflow/pyfunc/__init__.py
tomasatdatabricks: it would, I thought we've decided not to allow int to float conversions.

mlflow/pyfunc/__init__.py
tomasatdatabricks: kk makes sense

mlflow/pyfunc/__init__.py
tomasatdatabricks: ok, makes sense.

mlflow/pyfunc/__init__.py
tomasatdatabricks: Makes sense, will do

mlflow/pyfunc/__init__.py
```diff
@@ -439,26 +560,41 @@ def spark_udf(spark, model_uri, result_type="double"):
...
                 pdf = x
         if pdf is None:
-            # Explicitly pass order of columns to avoid lexicographic ordering (i.e., 10 < 2)
-            columns = [str(i) for i, _ in enumerate(args)]
-            pdf = pandas.DataFrame(schema, columns=columns)
+            args = list(args)
+            if input_schema is None:
+                names = [str(i) for i in range(len(args))]
+            else:
+                names = input_schema.column_names()
```
aarondav, line 577: What is the value of column_names if the schema provided no column names?

tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
```diff
@@ -115,6 +118,111 @@ def test_signature_and_examples_are_saved_correctly(sklearn_knn_model, iris_data
...
+    assert "Incompatible input types" in str(ex)
+    pdf["a"] = pdf["a"].astype(np.int32)
+
+    pdf["b"] = pdf["b"].astype(np.float64)
+    with pytest.raises(MlflowException) as ex:
+        pyfunc_model.predict(pdf)
+    pdf["b"] = pdf["b"].astype(np.int64)
+    assert "Incompatible input types" in str(ex)
+
+    # 7. objects work
```
aarondav, line 216: nice tests!

tests/pyfunc/test_scoring_server.py
```diff
@@ -244,6 +270,47 @@ def test_split_oriented_json_to_df():
...
+
+    schema = Schema([ColSpec(c, c) for c in pandas_df_with_all_types.columns])
+    df = _shuffle_pdf(pandas_df_with_all_types)
+    with TempDir(chdr=True):
+        with mlflow.start_run() as run:
+            mlflow.pyfunc.log_model("model", python_model=TestModel(),
+                                    signature=ModelSignature(schema))
+        response = pyfunc_serve_and_score_model(
+            model_uri="runs:/{}/model".format(run.info.run_id),
+            data=json.dumps(df.to_dict(orient="split"), cls=NumpyEncoder),
```
aarondav, line 300: Mm, this test is a bit weird since we're starting with a Pandas DF. Could we at least include ones where the input is JSON so we can make it obvious how we deal with converting types? E.g., what happens if I specify

9007199254740992

for something that specifies float as its data type? This is valid as a double or as a long, but not as an int or a float.

mlflow/pyfunc/__init__.py
tomasatdatabricks: range from 0..num_cols

tests/pyfunc/test_scoring_server.py
tomasatdatabricks: Makes sense. I'll add a test.

tomasatdatabricks: Actually, it looks like pandas is happy parsing 9007199254740992 as float into 9.007199e+15 which contradicts our rules about only upcasts is ok. I don't think we can do much about it though.  

It's  even worse for integers as it parses as 0 ( I think this changes with pandas version). Again, I don't think it makes too much sense to do something about it for now. 

So, I could add the test, but it would fail.

mlflow/pyfunc/__init__.py
aarondav: Right, sorry, I just meant that it wasn't clear at this point in the code that we were going to do an upcast later. You never mention what the code looks like, so until you read 30 lines later, you don't know why this "would fail the upcast check".

tests/pyfunc/test_scoring_server.py
aarondav: Add the test please, as documentation for the current behavior. You can note that it is weird and subject to change.

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,139 @@ def _load_model_env(path):
...
 
-def load_model(model_uri, suppress_warnings=True):
+def _enforce_type(name, values: pandas.Series, t: DataType):
+    """
+    Enforce the input column type matches the declared in model input schema.
+
+    The following type conversions are allowed:
+
+    1. np.object -> string
+    2. int -> long
```
aarondav, line 280:
```suggestion
    2. int -> long (upcast)
```

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,139 @@ def _load_model_env(path):
...
+    3. float -> double.
```
aarondav, line 281:
```suggestion
    3. float -> double (upcast)
```

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,139 @@ def _load_model_env(path):
...
+            raise MlflowException(
+                "Failed to convert column {0} from type {1} to {2}.".format(
+                  name, values.dtype, t)
+            )
+
+    numpy_type = t.to_numpy()
+    is_compatible_type = values.dtype.kind == numpy_type.kind
+    is_upcast = values.dtype.itemsize <= numpy_type.itemsize
+    if is_compatible_type and is_upcast:
+        # upcasting compatible types is ok
```
aarondav, line 315: the above if statement has great variable names, so no need for this comment :)

mlflow/pyfunc/__init__.py
```diff
@@ -286,16 +422,23 @@ def load_model(model_uri, suppress_warnings=True):
                               loading process will be suppressed. If ``False``, these warning
                               messages will be emitted.
     """
-    local_model_path = _download_artifact_from_uri(artifact_uri=model_uri)
-    conf = _get_flavor_configuration(model_path=local_model_path, flavor_name=FLAVOR_NAME)
+    local_path = _download_artifact_from_uri(artifact_uri=model_uri)
```
aarondav, line 425: What are the effective changes here?

mlflow/pyfunc/__init__.py
```diff
@@ -266,7 +270,139 @@ def _load_model_env(path):
...
+    For column names, we check there are no missing columns and reorder the columns to match the
+    ordering declared in schema if necessary. Any extra columns are ignored.
+
+    For column types, we make sure the types match schema or can be safely converted to match the
+    input schema.
+    """
+    col_names = input_schema.column_names()
+    col_types = input_schema.column_types()
+    expected_names = set(col_names)
+    actual_names = set(pdf.columns)
```
aarondav, line 340: Just to confirm, we have a test that ensures that this works for unnamed columns?

mlflow/pyfunc/__init__.py
tomasatdatabricks: :)

tests/pyfunc/test_scoring_server.py
tomasatdatabricks: kk will do

mlflow/pyfunc/__init__.py
tomasatdatabricks: The old code loaded conf for the given flavor in a function call. 
Now we also need MLmodel file to pass to the PyFuncModel, so I parse the MLmodel file here and extract the flavor conf from there.

mlflow/pyfunc/__init__.py
tomasatdatabricks: Good catch, we don't! In fact, now that I am thinking about it, it probably makes more sense to not check column names at all if schema did not declare them. We should only check that we got at least the expected number of columns. We actually enforce them to be an integer range which is weird. I'll update the logic and add a test.

mlflow/types/schema.py
```diff
@@ -2,34 +2,50 @@
...
 import numpy as np
+import pandas as pd
 from typing import Dict, Any, List, Union, Optional
 
+from pandas.core.dtypes.dtypes import PandasExtensionDtype
+
 from mlflow.exceptions import MlflowException
 
 
+def _pandas_string_type():
```
aarondav, line 13: What's going on here?

mlflow/types/schema.py
```diff
@@ -2,34 +2,50 @@
...
         res._value_ = value
         res._numpy_type = numpy_type
+        res._spark_type = spark_type
+        res._pandas_type = pandas_type if pandas_type is not None else numpy_type
         return res
 
-    boolean = (1, np.bool)
+    # NB: We only use pandas extension type for strings. There are also pandas extension types for
+    # integers and boolean values. We do not use them here for now as most downstream tools are
+    # most likely to use / expect native numpy types and owuld not be compatible with the extension
```
aarondav, line 34:
```suggestion
    # most likely to use / expect native numpy types and would not be compatible with the extension
```

mlflow/types/schema.py
```diff
@@ -119,6 +145,22 @@ def numpy_types(self) -> List[np.dtype]:
...
+    def pandas_types(self) -> List[np.dtype]:
+        """ Convenience shortcut to get the datatypes as pandas types."""
+        return [x.type.to_pandas() for x in self.columns]
+
+    def as_spark_schema(self):
+        """Convert to Spark schema. If this schema is a single unnamed column, it is converted
+        directly the corresponding spark data type, otherwise it's returned as a struct (missing
+        column names are filled with an integer sequence).
+        """
+        if len(self.columns) == 1 and self.columns[0].name is None:
```
aarondav, line 157: Do we have test coverage for these various cases?

mlflow/types/schema.py
tomasatdatabricks: String extention type is only available for pandas >= 1.0 and is recommended over object. We will just use object for pandas < 1.0.

mlflow/types/schema.py
tomasatdatabricks: Added tests.

tests/pyfunc/test_scoring_server.py
tomasatdatabricks: Added a test above (testing only the parse in scoring server itself).

aarondav:
Merged in #2920!

Issue #2789, opened 2020-05-05
[BUG] Tensorflow 1.x example conda ResolvePackageNotFound


jlangballe-miovision:
### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix
for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [x] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 10.15.4
- **MLflow installed from (source or binary)**: binary
- **MLflow version (run ``mlflow --version``)**:  1.8.0
- **Python version**: 3.7.7
- **npm version, if running the dev UI**:
- **Exact command to reproduce**:
```
cd examples/tensorflow/tf1
mlflow run .
```

### Describe the problem
Trying to run the tf1 examples gives an error installing dependencies.

```
ResolvePackageNotFound: 
  - python==3.6
```

### Code to reproduce issue
```
cd examples/tensorflow/tf1
mlflow run .
```

### Other info / logs
This is caused by a typo with dependencies in the tf1/conda.yaml file.

```
name: tensorflow-example
channels:
  - defaults
  - anaconda
dependencies:
  - python==3.6 # <-- Should be single =
  - tensorflow==1.13.1 # Here as well
  - pip:
    - mlflow
```

Changing the == to = fixes the issue.

Here's the full output of running the run command.
```

$ mlflow run .                                
2020/05/05 15:19:57 INFO mlflow.projects: === Creating conda environment mlflow-680bb0a93589120b486c611d395fa96a6ec264e5 ===
Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - python==3.6

Traceback (most recent call last):
  File "/usr/local/bin/mlflow", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 829, in __call__
[...]

Issue #1796, opened 2019-08-29
[SETUP-BUG] ResolvePackageNotFound - python=3.5.2


FredYao:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md)
for information on what types of issues we address.
  
Please fill in this template and do not delete it unless you are sure your issue is outside its scope.

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 16.04
- **MLflow installed from (source or binary)**:  binary (pip install mlflow)
- **MLflow version (run ``mlflow --version``)**: 1.2.0
- **Python version**: 3.5.2
- **Exact command to reproduce**: mlflow models build-docker -m /path/to/model -n "my-model" 

### Describe the problem
mlflow models build-docker -m /path/to/model -n "my-model"

### Other info / logs
Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your condadependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Pleaseadd an explicit pip dependency.  I'm adding one for you, but still nagging you.
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed

ResolvePackageNotFound:
  - python=3.5.2

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/miniconda/lib/python3.7/site-packages/mlflow/models/container/__init__.py", line 102, in _install_pyfunc_deps
    raise Exception("Failed to create model environment.")
Exception: Failed to create model environment.
creating and activating custom environment
The command '/bin/sh -c python -c                 'from mlflow.models.container import _install_pyfunc_deps; _install_pyfunc_deps("/opt/ml/model", install_mlflow=False)'' returned a non-zero code: 1

AlistairLR112:
bumping as I am experiencing this too

AlistairLR112:
Python 3.5.2 is no longer available on the conda-forge repo

AlistairLR112:
It seems like databricks clusters (if you built it on Databricks) run python 3.5.2

PR #2792, opened 2020-05-06
Fix example tests and add spacy example test to CRON job

juntai-zheng:
Addresses https://github.com/mlflow/mlflow/issues/2789

Several examples have been broken due to conda yamls suddenly not accepting `python==3.6`. These have been changed to `python=3.6`. Currently do not know why this is the case, but this hopefully fixes the examples.

This PR also adds the `spacy` example to the suite of nightly tests.

## How is this patch tested?

Travis tests.

- [ ] UI
- [ ] Command Line Interface
- [ ] API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Model Registry
- [ ] Scoring
- [ ] R
- [ ] Java
- [ ] Python

### DIFF
--- a/examples/flower_classifier/conda.yaml
+++ b/examples/flower_classifier/conda.yaml
@@ -2,6 +2,7 @@ name: flower_classifier
 channels:
   - defaults
   - anaconda
+  - conda-forge
 dependencies:
   - python=3.7
   - pandas=1.0.3
--- a/examples/flower_classifier/conda.yaml
+++ b/examples/flower_classifier/conda.yaml
@@ -3,12 +3,12 @@ channels:
   - defaults
   - anaconda
 dependencies:
-  - python==3.7
-  - pandas==1.0.3
-  - scikit-learn==0.22.1
-  - tensorflow==1.13.1
-  - keras==2.3.1
-  - pillow==7.0.0
-  - pip==20.0.2
+  - python=3.7
+  - pandas=1.0.3
+  - scikit-learn=0.22.1
+  - tensorflow=1.13.1
+  - keras=2.3.1
+  - pillow=7.0.0
+  - pip=20.0.2
   - pip:
     - mlflow>=1.6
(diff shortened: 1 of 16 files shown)


harupy:
@juntai-zheng 

I was investigating this issue too and here's what I found.

[This anaconda doc](https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications) says:

> <, >, <=, >=, ==, and != are relational operators on versions, **which are compared using PEP-440**.

And [PEP-440](https://www.python.org/dev/peps/pep-0440/#version-matching) says:

> By default, the version matching operator (==) is based on a strict equality comparison: the specified version must be exactly the same as the requested version. **The only substitution performed is the zero padding of the release segment to ensure the release segments are compared with the same length.**

So `python==3.6` matches `python==3.6` or `python==3.6.0`.

juntai-zheng:
@harupy thanks for the investigation! That makes sense, though it's unfortunate that Anaconda decided to delete those versions with little notification...

dbczumar:
@juntai-zheng @juntai-zheng This is the same class of issue described here: https://github.com/mlflow/mlflow/issues/1796 (in this case, Python versions specified in our auto-generated conda environments that we persist with MLflow models are no longer available from the Anaconda channel). I'm going to explore fixing this by adding the `conda-forge` channel to our default conda environments (Conda forge lists some Python versions that aren't present in the Anaconda channel, and vice versa: https://anaconda.org/conda-forge/python/files?version=3.7.6 , https://anaconda.org/anaconda/python/files?version=3.7.4)

dbczumar:
Filed https://github.com/mlflow/mlflow/pull/2797. Please take a look!

dbczumar (approved):
LGTM! Thanks @juntai-zheng!

Issue #2574, opened 2020-03-11
[FR] Add possibility to use client cert. with tracking api


PhilipMay:
I would like to use client certs with tracking API.
Could you add that?

You basically have to add the possibility to add a filename to the cert param of the `requests.request` function. See here: https://requests.readthedocs.io/en/master/api/

Here are the 2 calls that need to be extended:
- https://github.com/mlflow/mlflow/blob/master/mlflow/utils/rest_utils.py#L51
- https://github.com/mlflow/mlflow/blob/master/mlflow/utils/rest_utils.py#L62

dbczumar:
@PhilipMay both https://github.com/mlflow/mlflow/blob/master/mlflow/utils/rest_utils.py#L51 and https://github.com/mlflow/mlflow/blob/master/mlflow/utils/rest_utils.py#L62 already accept a `**kwargs` parameter that can be used to forward cert arguments to the `requests.request` function. However, these are internal methods meant to be called by the `RestStore` during MLflow Tracking operations; users are not expected to call these directly.

I think we might want to add a cert path field to the `MlflowHostCreds` abstraction (https://github.com/mlflow/mlflow/blob/74465b29c396e9c442031f7b0d3720f1312442bc/mlflow/utils/rest_utils.py#L143) instead, along with an environment variable that can be used to specify a cert path (similar to the variables defined here: https://github.com/mlflow/mlflow/blob/476dcd63a50d739ca8f4272bfd0215880419ce30/mlflow/tracking/_tracking_service/utils.py#L20-L23.

cc @aarondav for input here.

dbczumar:
@PhilipMay Would you be willing to contribute this type of enhancement?

dbczumar:
@PhilipMay Just checking in again! Any thoughts here?

PhilipMay:
I will have a look into it. 
Maybe I will just add "client cert. with tracking api" with the possibility to set an env-variable.

PhilipMay:
Well - I did some code changes but I am blocked by this bug at the moment: https://github.com/mlflow/mlflow/issues/2812

dbczumar:
Hi @PhilipMay , I responded to #2812. Please let me know if this unblocks you!

PhilipMay:
> Hi @PhilipMay , I responded to #2812. Please let me know if this unblocks you!

I found a workaround.

PhilipMay:
Started a PR: #2843

PhilipMay:
[...]

PR #2843, opened 2020-05-17
Add possibility to use client cert. with tracking API.

PhilipMay:
see #2574 

## TODO
- [x] tests
- [x] documentation - https://mlflow.org/docs/latest/tracking.html#logging-to-a-tracking-server
- [x] python docstring of `MlflowHostCreds`


PhilipMay:
Before I continue with this PR I would like to have some feedback.
In this PR I only plan to introduce client cert usage with environment variable.
I do not plan to open up the python tracking API.

Disregarding the missing documentation and tests: Is this implementation ok or do you see more needed changes on other places?

dbczumar:
@PhilipMay Sorry for the delay here. This implementation looks great! Let me know if you can add some tests and documentation - more than happy to get this merged once those have been added! Thank you for your contribution!

PhilipMay:
I am finished with this PR.
- Tests written - positive and negative
- Docstring for python
- Documentation

Can you please review this and give feedback?

Thanks,
Philip

PhilipMay:
Hi. Any comments on this PR?

mlflow/utils/rest_utils.py
```diff
@@ -153,13 +159,26 @@ class MlflowHostCreds(object):
...
+        function (see https://requests.readthedocs.io/en/master/api/).
+        If this is set ``ignore_tls_verification`` must be false.
     """
     def __init__(self, host, username=None, password=None, token=None,
-                 ignore_tls_verification=False):
+                 ignore_tls_verification=False, client_cert_path=None, server_cert_path=None):
         if not host:
             raise MlflowException("host is a required parameter for MlflowHostCreds")
+        if ignore_tls_verification and (server_cert_path is not None):
+            raise MlflowException("if 'MLFLOW_TRACKING_INSECURE_TLS' is set to true "
```
dbczumar, line 176: @PhilipMay This is a really minor point of feedback, but can we start this error message by indicating that `client_cert_path` and `server_cert_path` should not both be specified and include the environment variable part afterwards as a hint? Users or some other piece of backend code may try to construct this object directly, without the use of environment variables.

Also, can we specify an error code of `INVALID_PARAMETER_VALUE` here? See https://github.com/mlflow/mlflow/blob/master/mlflow/utils/uri.py#L55 for an example.

dbczumar (commented):
@PhilipMay Sorry for the delay again - this looks awesome! I left one small comment about the error message behavior. I'm going to give this a manual test tomorrow and should be ready to approve / merge it by EOD tomorrow!

mlflow/utils/rest_utils.py
PhilipMay: The exception message has been improved. The error code is still missing.

2 lines above my exception is an other `MlflowException` without error code. Since I am not familiar with the proto stuff like in the example I am not sure what to do...
Could you maybe add this after the PR?

PhilipMay:
I improved the message but did not add an error ID. See comment above.
@dbczumar What do you think?

PhilipMay:
> I'm going to give this a manual test tomorrow and should be ready to approve / merge it by EOD tomorrow!

If you have problems with the manual test I can provide an howto. I start the server with this command:

`mlflow server --gunicorn-opts "--keyfile=/home/mlflow/mlflow-cert/server.key --certfile=/home/mlflow/mlflow-cert/server.crt --cert-reqs=2 --ca-certs=/home/mlflow/mlflow-cert/client.crt" --backend-store-uri postgresql://mlflow:<password>@127.0.0.1/mlflow?sslmode=require --default-artifact-root /home/mlflow/default-artifact-root -h 0.0.0.0
```

dbczumar:
@PhilipMay I've pushed the tweaks you mentioned for the `INVALID_PARAMETER_VALUE` error code; I've tested this manually and found that it works as expected! I'll go ahead and merge this once the tests pass!

dbczumar (approved):
LGTM! Thanks so much @PhilipMay !

PR #2845, opened 2020-05-18
Add support for cluster library definition in MLflow project execution

### LINKED ISSUE #2800: [FR] Add support for cluster library definition in MLflow project execution
Hi!

MLFlow redefines library dependencies on databricks cluster creation. In my project, extra libraries are needed on cluster nodes, but when I'm specifying them in cluster json file, they are ignored and replaced by the single mlflow dependency. 

I suppose this happens in this line of code: https://github.com/mlflow/mlflow/blob/58a6053cad8d6c01f095a4e3dac8a1b560dc4a61/mlflow/projects/databricks.py#L182

This is not very flexible and doesn't allow me to install extra libraries on mlflow cluster nodes.
So, better behaviour would be to add mlflow dependency to dependency given in cluster specification file.

Below is the example of my cluster.json:
```
{
  "spark_conf": {
    "spark.databricks.hive.metastore.glueCatalog.enabled": "true"
  },
  "libraries": [
    {
      "whl": "dbfs:/path_to_my_lib.whl"
    }
  ],
  "num_workers": 20
}
```

But my whl dependency is ignored on cluster creation. Below is the resulting cluster properties:
```
  "cluster_spec": {
    "new_cluster": {
    "libraries": [
      {
        "pypi": {
          "package": "'mlflow<=1.7.2'"
        }
      }
    ]
  },
```
Thanks!

pogil:

Add support for Databricks cluster specification, including libraries, to Databricks backend configuration. Fixes #2800 using [Approach 2](https://github.com/mlflow/mlflow/issues/2800#issuecomment-629399933).

## How is this patch tested?

New test.
Manual verification of projects installing wheels on driver and workers on Databricks backend.

Extended support for library specification in Databricks backend config to match 
the syntax in [ClusterSpec](https://docs.databricks.com/dev-tools/api/latest/jobs.html#clusterspec) in addition to the previously supported syntax of [NewCluster](https://docs.databricks.com/dev-tools/api/latest/jobs.html#jobsclusterspecnewcluster)

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/docs/source/projects.rst
+++ b/docs/source/projects.rst
@@ -394,7 +394,7 @@ If you do need to install libraries on the worker, use the "cluster specificatio
   .. code-block:: json
 
     {
-      "new-cluster": {
+      "new_cluster": {
         "spark_version": "6.4.x-scala2.11",
         "node_type_id": "i3.xlarge",
         "aws_attributes": {"availability": "ON_DEMAND"},
--- a/docs/source/projects.rst
+++ b/docs/source/projects.rst
@@ -237,7 +237,7 @@ Docker container environment
   .. rubric:: Example 3: Image in a remote registry
 
   .. code-block:: yaml
-    
+
     docker_env:
       image: 012345678910.dkr.ecr.us-west-2.amazonaws.com/mlflow-docker-example-environment:7.0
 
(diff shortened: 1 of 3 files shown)


docs/source/projects.rst
```diff
@@ -375,18 +375,59 @@ in the Databricks docs
...
       "aws_attributes": {"availability": "ON_DEMAND"},
       "num_workers": 4
     }
 
+If you do need to install libraries on the worker, use the "cluster specification" format. For example:
+
+  .. code-block:: json
+
+    {
+      "new-cluster": {
```
smurching, line 397:
```suggestion
      "new_cluster": {
```

pogil: Yep, true! Code and tests expect that too!

docs/source/projects.rst
```diff
@@ -375,18 +375,59 @@ in the Databricks docs
 of how to use the feature is as follows:
 
 1. Create a JSON file containing the
-`new cluster specification <https://docs.databricks.com/api/latest/jobs.html#jobsclusterspecnewcluster>`_
-for your run. For example:
+`new cluster specification <https://docs.databricks.com/dev-tools/api/latest/jobs.html#jobsclusterspecnewcluster>`_
+or `cluster specification <https://docs.databricks.com/dev-tools/api/latest/jobs.html#clusterspec>`_
+for your run. If you do not need to specify libraries installed on the Spark worker nodes, you can use the
```
smurching, line 380:
```suggestion
for your run. Note that running projects against an existing cluster is not supported. If you do not need to specify libraries installed on the Spark worker nodes, you can use the
```

I think it'd be helpful to clarify this in the docs, as [ClusterSpec](https://docs.databricks.com/dev-tools/api/latest/jobs.html#clusterspec) allows you to specify an existing cluster ID, although users will also see an error message if they attempt to run against an existing cluster (that error message I suspect will say something cryptic about "shell command tasks" instead of "mlflow projects", will file a separate PR to fix that...)

mlflow/projects/databricks.py
```diff
@@ -162,12 +162,26 @@ def _run_shell_command_job(self, project_uri, command, env_vars, cluster_spec):
...
                  Databricks
                  `Runs Get <https://docs.databricks.com/api/latest/jobs.html#runs-get>`_ API.
         """
+        # NB: We use <= on the version specifier to allow running projects on pre-release
+        # versions, where we will select the most up-to-date mlflow version available.
+        # Also note, that we escape this so '<' is not treated as a shell pipe.
+        libraries = [{"pypi": {"package": "'mlflow<=%s'" % VERSION}}]
+
+        # Check syntax of JSON - if it contains a libraries and a cluster_spec, pull those out
+        if 'libraries' in cluster_spec and 'new_cluster' in cluster_spec:
```
smurching, line 181: Given that we document that a `new_cluster_spec` with [these fields](https://docs.databricks.com/dev-tools/api/latest/jobs.html#clusterspec) is valid, I suspect that we may want to allow specifying a `new_cluster` without libraries, e.g via logic like:

```suggestion
    if 'new_cluster' in cluster_spec:
        if 'libraries' in cluster_spec:
            libraries.extend(cluster_spec['libraries'])
        cluster_spec = cluster_spec['new_cluster']
```

We'd also then want a unit test similar to [the one you already added](https://github.com/mlflow/mlflow/pull/2845/files#diff-e5f3a2fbd5bd330b3069ea935f1718eeR273-R296) for specifying `new_cluster` without `libraries`. I'm open to adding this support in a follow-up PR - if we go that route, we should probably raise an exception if the user specifies `new_cluster` but no `libraries`, as we otherwise currently just pass an object like `{"new_cluster": {...}}` to the REST API, which I think will fail to run the job. Let me know what you think, thanks!

smurching (commented):
A few comments but otherwise looks good, thanks @pogil!

mlflow/projects/databricks.py
pogil: Thanks, @smurching! I've incorporated your feedback to make `libraries` optional in the specification and added a test covering that.

smurching (approved):
LGTM

PR #2884, opened 2020-05-26
Removing down pinning of `sqlalchemy` version

lokesh005:

Removing down pinning of `sqlalchemy` version so that it is compatible with the latest versions.

## How is this patch tested?

I updated the version of the package and ran mlflow (as mentioned in the Readme file), it is working fine with no issue.

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
         'docker>=4.0.0',
         'entrypoints',
         'sqlparse',
-        'sqlalchemy>=1.3.0',
+        'sqlalchemy',
         'gorilla',
         'prometheus-flask-exporter',
     ],
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
         'docker>=4.0.0',
         'entrypoints',
         'sqlparse',
-        'sqlalchemy<=1.3.13',
+        'sqlalchemy>=1.3.0',
         'gorilla',
         'prometheus-flask-exporter',
     ],


mparkhe:
@lokesh005 : This was pinned to `1.3.13` or below in PR #2587 to avoid issues that arise when installing it using `pep517`. Is there a reason you wish to import a later version?

cc: @smurching

lokesh005:
I followed the instruction mentioned in [CONTRIBUTING.rst](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst#id11) script to install all the packages and I did not face any issue. Also, it is always better to upgrade the pin instead of downgrading it, so that it can be compatible with the latest changes in the package. I am using `sqlalchemy==1.3.15` so the down pinning is causing an issue for me when I am trying to install the latest version of `mlflow`. 

Also, I googled about the `pep517` issue, it says that while installing `sqlalchemy` we can add argument `--no-use-pep517` so as to avoid any issue. It will be worth going through [pip#7874](https://github.com/pypa/pip/issues/7874) and [pip#5207](https://github.com/sqlalchemy/sqlalchemy/issues/5207) issues.

mparkhe:
Since MLflow is used in several environments and backends, we are very deliberate about upgrade library versions and do it as a part of major version release. Also while upgrade library dependencies like `sqlalchemy` we want to make sure that performance and functionality is not changed from expected behavior. Adding `--no-use-pep517` is a workaround, but we would need to test on how that affects the other libraries loaded as requirements. If we an help avoid customizations, makes installation easy for all.

Since you are having version conflicts, I would recommend [creating an independent conda environment](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst#prerequisites) for MLflow. This way you can have the latest `sqlalchemy` version on your machine and a pinned version in the conda environment.

lokesh005:
I tried upgrading the `sqlalchemy` version to the latest version in my local system and it is working as is, with no issues. Is there anything that I can do (or contribute) to make sure that performance and functionality are not deviating from expected behavior so that we can remove down pinning of version in the future release. Also, I think that there were no issues with `sqlalchemy` version prior to PR[#2587](https://github.com/mlflow/mlflow/pull/2587).

PR #2891, opened 2020-05-28
Making the instructions clearer in the Exception description "Could n…

Hadrien-Cornier:
…ot find Conda executable" and fixing a small bug with MLFLOW_CONDA_HOME

The problem happens whenever the user runs in the exception line 350 inside "_get_or_create_conda_env", which corresponds to a failure to find the conda bin executable.

If the user has not yet modified the MLFLOW_CONDA_HOME environment variable, the exception message tells the user to modify the  MLFLOW_CONDA_HOME environment variable.

A user might be tempted to set it by using "mlflow.projects.MLFLOW_CONDA_HOME " after reading the description of '_get_conda_bin_executable' at line 315, but in fact doing this will not solve the problem because  at line 325 what will happen is that :  

os.environ.get(MLFLOW_CONDA_HOME) will return None

In fact the variable MLFLOW_CONDA_HOME is somewhat misleading because it cannot be anything else than the string 'MLFLOW_CONDA_HOME' otherwise the line of code 'conda_home = os.environ.get(MLFLOW_CONDA_HOME)' will not work .

So the current situtation is that we have a variable declared that in fact cannot be allowed to vary.

In fact the way to solve the conda bin not found problem is to do : 
os.environ['MLFLOW_CONDA_HOME']= "/path/to/anaconda3/"

I indicated that in the exception message, but I also added the capacity to read a path set by user through mlflow.projects.MLFLOW_CONDA_HOME

## How is this patch tested?

(Details)

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/projects/__init__.py
+++ b/mlflow/projects/__init__.py
@@ -322,6 +322,8 @@ def _get_conda_bin_executable(executable_name):
     ``mlflow.projects.MLFLOW_CONDA_HOME`` is unspecified, this method simply returns the passed-in
     executable name.
     """
+    if MLFLOW_CONDA_HOME != "MLFLOW_CONDA_HOME":
+        return os.path.join(MLFLOW_CONDA_HOME, "/bin/%s" % executable_name)
     conda_home = os.environ.get(MLFLOW_CONDA_HOME)
     if conda_home:
         return os.path.join(conda_home, "bin/%s" % executable_name)
@@ -353,7 +355,8 @@ def _get_or_create_conda_env(conda_env_path, env_id=None):
                                  "user-guide/install/index.html. "
                                  "You can also configure MLflow to look for a specific "
                                  "Conda executable by setting the {1} environment variable "
-                                 "to the path of the Conda executable"
+                                 "to the path of the Conda executable, by doing "
+                                 "os.environ['MLFLOW_CONDA_HOME']= '/path/to/anaconda3/'" 
                                  .format(conda_path, MLFLOW_CONDA_HOME))
     (_, stdout, _) = process.exec_cmd([conda_path, "env", "list", "--json"])
     env_names = [os.path.basename(env) for env in json.loads(stdout)['envs']]


mlflow/projects/__init__.py
```diff
@@ -322,6 +322,8 @@ def _get_conda_bin_executable(executable_name):
     ``mlflow.projects.MLFLOW_CONDA_HOME`` is unspecified, this method simply returns the passed-in
     executable name.
     """
+    if MLFLOW_CONDA_HOME != "MLFLOW_CONDA_HOME":
```
mparkhe, line 325: This would be a no-op since `MLFLOW_CONDA_HOME` constant is set to `"MLFLOW_CONDA_HOME"` string @ [line 45](https://github.com/mlflow/mlflow/blob/03f9c82e3ed2915e2dfc6cb2d32ee49c04be84f2/mlflow/projects/__init__.py#L45)

mlflow/projects/__init__.py
```diff
@@ -353,7 +355,8 @@ def _get_or_create_conda_env(conda_env_path, env_id=None):
                                  "user-guide/install/index.html. "
                                  "You can also configure MLflow to look for a specific "
                                  "Conda executable by setting the {1} environment variable "
-                                 "to the path of the Conda executable"
+                                 "to the path of the Conda executable, by doing "
+                                 "os.environ['MLFLOW_CONDA_HOME']= '/path/to/anaconda3/'"
```
mparkhe, line 359: Why set an environment variable in python? Easier to set is using `export` or 

```
MLFLOW_CONDA_HOME='...' mlflow run ...
```

mparkhe:
Per documentation on [mlflow.org](https://mlflow.org/docs/latest/projects.html#project-environments), `mlflow projects` will pick up path to `conda` is resolved form default system path. If user wants to override that define an environment variable `MLFLOW_CONDA_HOME`.

In order to make mlflow very generic on different linux, mac, windows ... etc environments the paths are not hard coded inside python and have left to the user to define overrides using system variables. The error message is reporting that `conda` executable is not available at the resolved path

mlflow/projects/__init__.py
Hadrien-Cornier: not if the user does : 

`mlflow.projects.MLFLOW_CONDA_HOME = "path/to/anaconda.."`

it might seem more intuitive from the user's standpoint to communicate the conda path to mlflow in a way that changes a variable inside mlflow rather than going through the system's environment variables.

PR #2919, opened 2020-06-09
Created cluster context provider to set cluster ID as part of run tags

willzhan-db:

Create cluster context provider to set cluster ID as part of run tags.

## How is this patch tested?

With unit tests and manual test in dogfood

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/entities/source_type.py
+++ b/mlflow/entities/source_type.py
@@ -1,10 +1,11 @@
 class SourceType(object):
     """Enum for originating source of a :py:class:`mlflow.entities.Run`."""
-    NOTEBOOK, JOB, PROJECT, LOCAL, UNKNOWN = range(1, 6)
+    NOTEBOOK, JOB, CLUSTER, PROJECT, LOCAL, UNKNOWN = range(1, 6)
 
     _STRING_TO_SOURCETYPE = {
         "NOTEBOOK": NOTEBOOK,
         "JOB": JOB,
+        "CLUSTER": CLUSTER,
         "PROJECT": PROJECT,
         "LOCAL": LOCAL,
         "UNKNOWN": UNKNOWN
--- a/mlflow/entities/source_type.py
+++ b/mlflow/entities/source_type.py
@@ -1,11 +1,10 @@
 class SourceType(object):
     """Enum for originating source of a :py:class:`mlflow.entities.Run`."""
-    NOTEBOOK, JOB, CLUSTER, PROJECT, LOCAL, UNKNOWN = range(1, 6)
+    NOTEBOOK, JOB, PROJECT, LOCAL, UNKNOWN = range(1, 6)
 
     _STRING_TO_SOURCETYPE = {
         "NOTEBOOK": NOTEBOOK,
         "JOB": JOB,
-        "CLUSTER": CLUSTER,
         "PROJECT": PROJECT,
         "LOCAL": LOCAL,
         "UNKNOWN": UNKNOWN
--- a/mlflow/tracking/context/registry.py
+++ b/mlflow/tracking/context/registry.py
@@ -5,6 +5,7 @@ from mlflow.tracking.context.default_context import DefaultRunContext
 from mlflow.tracking.context.git_context import GitRunContext
 from mlflow.tracking.context.databricks_notebook_context import DatabricksNotebookRunContext
 from mlflow.tracking.context.databricks_job_context import DatabricksJobRunContext
+from mlflow.tracking.context.databricks_cluster_context import DatabricksClusterContext
 
 
 class RunContextProviderRegistry(object):
@@ -47,6 +48,7 @@ _run_context_provider_registry.register(DefaultRunContext)
 _run_context_provider_registry.register(GitRunContext)
 _run_context_provider_registry.register(DatabricksNotebookRunContext)
 _run_context_provider_registry.register(DatabricksJobRunContext)
+_run_context_provider_registry.register(DatabricksClusterContext)
 
 _run_context_provider_registry.register_entrypoints()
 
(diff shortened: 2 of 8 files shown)


mlflow/utils/databricks_utils.py
```diff
@@ -1,10 +1,12 @@
 import os
 import logging
 import subprocess
+import mlflow._spark_autologging
```
smurching, line 4: Could we actually add a file like `mlflow/_spark_utils.py` containing common utils for working with Spark and move the `get_active_spark_session` into that file? We could then import and use the `get_active_spark_session` method here and in `mlflow/_spark_autologging.py` as follows:


```suggestion
from mlflow._spark_utils import get_active_spark_session
```

tests/tracking/context/test_databricks_cluster_context.py
```diff
@@ -0,0 +1,26 @@
...
+    patch_cluster_id = mock.patch("mlflow.utils.databricks_utils.get_cluster_id")
+
+    with patch_cluster_id as cluster_id_mock:
+        assert DatabricksClusterRunContext().tags() == {
+            MLFLOW_DATABRICKS_CLUSTER_ID: patch_cluster_id.return_value
+        }
+
+
+def test_databricks_notebook_run_context_tags_nones():
+
```
smurching, line 25:
```suggestion
```

tests/tracking/context/test_databricks_cluster_context.py
```diff
@@ -0,0 +1,26 @@
...
+from mlflow.tracking.context.databricks_cluster_context import DatabricksClusterRunContext
+
+
+def test_databricks_cluster_run_context_in_context():
+    with mock.patch("mlflow.utils.databricks_utils.is_in_cluster") as in_cluster_mock:
+        assert DatabricksClusterRunContext().in_context() == in_cluster_mock.return_value
+
+
+def test_databricks_cluster_run_context_tags():
+
```
smurching, line 15:
```suggestion
```

Removing extra/unneeded newlines

tests/tracking/context/test_databricks_cluster_context.py
```diff
@@ -0,0 +1,26 @@
...
+    patch_cluster_id = mock.patch("mlflow.utils.databricks_utils.get_cluster_id")
+
```
smurching, line 17:
```suggestion
```

smurching (commented):
One comment around avoiding duplicate code, otherwise LGTM! I would also make sure to take a look at the lint / Python small test failures

mlflow/utils/databricks_utils.py
willzhan-db: I think this is better than brutally importing the mlflow._spark_autologging. Will make these functions more organized. Will do!

mlflow/entities/source_type.py
```diff
@@ -1,10 +1,11 @@
 class SourceType(object):
     """Enum for originating source of a :py:class:`mlflow.entities.Run`."""
-    NOTEBOOK, JOB, PROJECT, LOCAL, UNKNOWN = range(1, 6)
+    NOTEBOOK, JOB, CLUSTER, PROJECT, LOCAL, UNKNOWN = range(1, 6)
 
     _STRING_TO_SOURCETYPE = {
         "NOTEBOOK": NOTEBOOK,
         "JOB": JOB,
+        "CLUSTER": CLUSTER,
```
smurching, line 8: I don't think we need to add `CLUSTER` as a source type, it'll show up separately from the "source" that generated the run in the MLflow UI (i.e. a notebook/job/local file will still be the run's source). Thus I'd undo changes to this file.

mlflow/tracking/context/registry.py
```diff
@@ -5,6 +5,7 @@
 from mlflow.tracking.context.git_context import GitRunContext
 from mlflow.tracking.context.databricks_notebook_context import DatabricksNotebookRunContext
 from mlflow.tracking.context.databricks_job_context import DatabricksJobRunContext
+from mlflow.tracking.context.databricks_cluster_context import DatabricksClusterContext
```
smurching, line 8:
```suggestion
from mlflow.tracking.context.databricks_cluster_context import DatabricksClusterRunContext
```

(let's use DatabricksClusterRunContext as the name as opposed to DatabricksClusterContext)

mlflow/utils/databricks_utils.py
```diff
@@ -98,6 +116,11 @@ def get_notebook_path():
     return _get_extra_context("notebook_path")
 
 
+def get_cluster_id():
+    spark_session = get_active_spark_session()
```
smurching, line 120: Add check: if `spark_session` is `None`, should return `None` here to be safe, rather than trying to access `spark_session.conf`

willzhan-db: Original thought is that it might be redundant as we will do "is_in_cluster_check" first, which has the logic that returns false if spark_session is none, or spark_session exists while spark_session.conf is none.
But as a standalone function, we should add that check as you mentioned

mlflow/utils/_spark_utils.py
```diff
@@ -0,0 +1,8 @@
+from pyspark.sql import SparkSession
```
smurching, line 1: Oh this is kinda subtle, can we move this line down into the try-catch block below, so that it looks like:

```
def _get_active_spark_session():
    try:
        from pyspark.sql import SparkSession
    except ImportError:
        # Return None if user doesn't have PySpark installed
        return None
    try:
        # getActiveSession() only exists in Spark 3.0 and above
        return SparkSession.builder.getActiveSession()
    except Exception:  # pylint: disable=broad-except
        # Fall back to this internal field for Spark 2.x and below.
        return SparkSession._instantiatedSession
```

That way we handle the case where this method is called but PySpark isn't installed

smurching (approved):
Pushed a small edit to be extra safe in ensuring that `is_in_cluster` does not throw, as it's called on each run creation - otherwise LGTM, thanks @willzhan-db!

PR #2930, opened 2020-06-12
Testing comment trigger for actions

pogil:

Add a check to trigger lint by adding a comment containing `trigger lint`.

## How is this patch tested?

This will be the test. 

Components 
Model Registry

Interface 

Language 

Integrations

Just a test, won't be committed.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -5,7 +5,6 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-    
 
 jobs:
   lint:
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -5,15 +5,11 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-  issue_comment:
-    types: [ created ]
-    
     
 
 jobs:
   lint:
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' || (github.event_name == 'issue_comment' && contains(github.event.comment.html_url, '/pull/') && contains(github.event.comment.body, 'trigger lint')) }}
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python 3.6
@@ -35,7 +31,6 @@ jobs:
 
   python-small:
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' || (github.event_name == 'issue_comment' && contains(github.event.comment.html_url, '/pull/') && contains(github.event.comment.body, 'trigger python-small')) }}
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python 3.6
@@ -55,7 +50,6 @@ jobs:
 
   java:
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' || (github.event_name == 'issue_comment' && contains(github.event.comment.html_url, '/pull/') && contains(github.event.comment.body, 'trigger java')) }}
     steps:
     - uses: actions/checkout@v2
     - name: Set up Java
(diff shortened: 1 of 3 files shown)


.github/workflows/master.yml
```diff
@@ -5,10 +5,15 @@ on:
...
     branches: [ master ]
+  issue_comment:
+    types: [ created ]
+    
+    
 
 jobs:
   lint:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' || (github.event_name == 'issue_comment' && contains(github.comment.html_url, '/pull/') && contains(github.event.comment.body, 'trigger lint')) }}
```
pogil, line 16: We would need to add similar lines for other jobs. Unfortunately docs suggest you can't do an `if` in the `on` block, nor can you an `if` on the list of all jobs.

pogil:
trigger lint

pogil:
trigger docs

pogil:
trigger lint

harupy:
I found a related post: https://github.community/t/trigger-action-via-pull-request-message/17630. It seems pretty complicated to achieve this feature.

harupy:
We could use the [re-run-workflow](https://developer.github.com/v3/actions/workflow-runs/#re-run-a-workflow) API (note that this API re-runs the whole workflow and has no option which allows us to re-run a specific job). This approach would be like:

1. Use `issue_comment` event as a trigger.
2. Find the id of a workflow that we want to re-run (not sure if this is really possible)
3. Hit the re-run-workflow API with the workflow id found in the previous step.

pogil:
@harupy thanks for doing the research here! I had came to a similar conclusion but hadn't found the re-run-workflow API yet. I'll take a look

pogil:
1. Get pull request of issue comment from `github.event.issue.pull_request.url` in workflow for issue_comment type actions
2. Look up that pull request
> curl 'https://api.github.com/repos/mlflow/mlflow/pulls/2930'
3. Get version from the pull request from 'head.sha' in JSON
4. Use the sha to look up check suites for that commit
> curl -H'Accept: application/vnd.github.antiope-preview+json' 'https://api.github.com/repos/mlflow/mlflow/commits/cbb29bc1aa61c8e7ed4c53aab16b1fc519c2f581/check-suites'

^^ this could in theory return multiple check suites, but in practice it may be broken. There's a valid scenario where you may open the same pull requests against different branches. It looks like in the UI with a test I did the check suites get shared across the two branches rather than trying the merge request for each branch separately. That simplifies bug validation, I guess? It's kind hard to know from information in the check suite which branch it was run on, because it just has the before/after commit hashes. 

5. Trigger a rerun using the check_suites `id`:
> curl -uuser:token -X POST -H'Accept: application/vnd.github.antiope-preview+json' 'https://api.github.com/repos/mlflow/mlflow/check-suites/791794316/rerequest'

pogil:
trigger checks

harupy:
It seems we need to merge the PR to the master branch to enable this action. [The doc](https://help.github.com/en/actions/reference/events-that-trigger-workflows#issue-comment-event-issue_comment) says:

> Note: This event (`issue_comment`) will only trigger a workflow run if the workflow file is on the master or default branch.

harupy:
When I ran the check-suite re-request API **on GitHub Action**, it didn't work (tested on [my dummy repo](https://github.com/harupy/trigger-action-by-comment)). The check-suite was successfully re-requested, but its status never changes from "queued" and the GitHub UI doesn't seem to update.

Screen Shot 2020-06-17 at 22 15 14

https://github.com/harupy/trigger-action-by-comment/pull/2/checks?check_run_id=779661702

pogil:
Yeah I expect I will hit a dead end using GitHub for this due to permissions. We could put a token in a secret, but then the commenter would have to have edit access to the workflow file to read the secret during workflow execution if I read the docs right.

Looks like the only way might be some externally hosted service that + webhooks

harupy:
Or make our own bot using GitHub Apps and give it permission required to do the job.

harupy:
I used my personal token instead of `secrets.GITHUB_TOKEN` and it worked.

harupy:
GitHub Apps don't seem to work either. As reported [here](https://github.com/probot/probot/issues/1065#issuecomment-576444183), check-suites created by GitHub Actions can't be updated from other GitHub Apps.

harupy:
@pogil Would it be useful to introduce a bot that automatically adds labels based on a PR description?

PR #2934, opened 2020-06-15
Add Local and Kubernetes execution backends

jdlesage:

This PR follows https://github.com/mlflow/mlflow/pull/2566 (introduction of execution backend). This PR encapsulates local execution and Kubernetes submission code into a class that implements the AbstractBackend interface

## How is this patch tested?

Unit tests

Components 
Model Registry
- [X] `area/projects`: MLproject format, project running backends

Interface 

Language 

Integrations


mlflow/projects/__init__.py
```diff
@@ -91,25 +68,25 @@ def _run(uri, experiment_id, entry_point, version, parameters,
     Returns a ``SubmittedRun`` corresponding to the project run.
     """
     tracking_store_uri = tracking.get_tracking_uri()
+    backend_config[PROJECT_USE_CONDA] = use_conda
+    backend_config[PROJECT_SYNCHRONOUS] = synchronous
+    backend_config[PROJECT_DOCKER_ARGS] = docker_args
+    backend_config[PROJECT_STORAGE_DIR] = storage_dir
     # TODO: remove this check once local, databricks, kubernetes execution have been refactored
```
smurching, line 75:
```suggestion
    # TODO: remove this check once databricks execution has been refactored into its own built-in
    # project execution backend
```

mlflow/projects/__init__.py
```diff
@@ -91,25 +68,25 @@ def _run(uri, experiment_id, entry_point, version, parameters,
...
     # into their own built-in project execution backends.
```
smurching, line 76:
```suggestion
```

mlflow/projects/__init__.py
```diff
@@ -91,25 +68,25 @@ def _run(uri, experiment_id, entry_point, version, parameters,
...
-    if backend_name not in {"local", "databricks", "kubernetes"}:
+    if backend_name not in {"databricks", "kubernetes"}:
```
smurching, line 77: Question: shouldn't we remove kubernetes from this list? Given that we've refactored it into its own backend in this PR

smurching: Unfortunate (but not the fault of this PR) that tests didn't catch this - there seems to be no test that calls `mlflow.run` with `backend="databricks"`, could we add one to this PR? We might just be able to modify one of the existing tests which already mocks out all the actual k8s job-submission logic, e.g. [this one](https://github.com/mlflow/mlflow/blob/f291f5bc8230cb44cfe25748505d3189dafe8aaf/tests/projects/test_kubernetes.py#L89) that directly calls the `run` method of the `KubernetesBackend`, and simply call `mlflow.projects.run` instead to more-closely verify things work end-to-end.

mlflow/projects/backend/loader.py
```diff
@@ -1,13 +1,24 @@
...
 __logger__ = logging.getLogger(__name__)
 
 
 # Statically register backend defined in mlflow
-MLFLOW_BACKENDS = dict()
+MLFLOW_BACKENDS = {
+    'local': LocalBackend,
+}
+
+try:
```
smurching, line 17: I'm guessing we perform this check to avoid `ModuleNotFoundErrors` when e.g. the `kubernetes` Python library isn't available. It makes sense to avoid such errors, but could we instead try to `import kubernetes` within the actual `run` method of the `KubernetesBackend` (so that we only throw a `ModuleNotFoundError` if you try to actually run a k8s project)? It seems a bit odd to not register the kubernetes backend at all if the `kubernetes` python library isn't present, as this will cause the backend name to not appear in help documentation / the list of supported backends generated [here](https://github.com/mlflow/mlflow/pull/2934/files#diff-16104674981d54b31e320402a143eb19R101)

mlflow/projects/utils.py
```diff
@@ -245,3 +253,74 @@ def _create_run(uri, experiment_id, work_dir, version, entry_point, parameters):
                    list(final_params.items()) + list(extra_params.items())]
     tracking.MlflowClient().log_batch(active_run.info.run_id, params=params_list)
     return active_run
+
+
+def get_conda_command(conda_env_name):
```
smurching, line 258: Given that this method is also called by `mlflow models serve` and other APIs unrelated to projects, IMO it'd be better to put them in a `mlflow/utils/conda_utils.py` module - what do you think? The thinking is that it's a bit odd to have modules under `mlflow.pyfunc` depend on `mlflow.projects.utils` - thanks! Same goes for `get_conda_bin_executable`

smurching (commented):
@jdlesage thanks so much for doing this refactor! Left some initial comments - I also wonder, would it be too much work to split this PR up into separate ones? e.g. maybe we can start by refactoring the k8s backend in a separate PR (should be easier to review/merge as the k8s backend has few dependencies on other modules), and then review/merge this PR, which should become smaller once the k8s changes have already been merged into master

smurching (commented):
Aside from those^ comments, this largely looks good BTW - but I'd like to take another pass again once we've reduced the diff size, if possible - thanks again!

mlflow/projects/__init__.py
jdlesage: Agree. I will add one in the PR related to kubernetes

mlflow/projects/utils.py
jdlesage: I don't not see the usage of this method in scoring server, but I agree about to put this method in a specific module.

jdlesage:
@smurching I started to split this PR https://github.com/mlflow/mlflow/pull/3124. I found easier to start by local backend. I took in consideration your remarks related to conda

I agree also on remark about the kubernetes import. I will address it in the next PR about kubernetes backend. 
I suggest to first have the local backend be merged before to start the kubernetes one. I keep this PR open just to keep track your comments.

jinzhang21:
Close for too many conflicts with master. Please update and reopen if you are still interested in working on this PR.

PR #2921, opened 2020-06-10
[ML-11010] add a search registered models database query and tests

ankit-db:
## What changes are proposed in this pull request?

This PR introduces a database query that can search for registered models via an equality operator, `LIKE`, and `ILIKE`. This is going to eventually be a part of a new API introduced to search for registered models.

## How is this patch tested?

Unit tests

### Is this a user-facing change?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [x] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [x] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations

### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [x] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [ ] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #2935, opened 2020-06-15
[Search RM Pagination] Paginate the search registered models database endpoint

ankit-db:

This PR adds onto #2921 and adds a pagination implementation of the search registered models endpoint.

## How is this patch tested?

Tested with unit tests

Components 
Model Registry

Interface 

Language 

Integrations


mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -100,11 +102,12 @@ def _save_to_db(self, session, objs):
             # single object
             session.add(objs)
 
-    def create_registered_model(self, name):
+    def create_registered_model(self, name, creation_time=None):
```
ankitmathur-db, line 105: I chose this approach because of its similarity to how we account for flakiness when testing search_runs pagination: https://github.com/mlflow/mlflow/blob/HEAD/tests/store/tracking/test_sqlalchemy_store.py#L1381

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -398,82 +411,132 @@ def test_search_registered_models(self):
...
+
+    def test_search_registered_model_pagination(self):
+        rms = [self._rm_maker("RM" + str(i), creation_time=i).name for i in range(50)]
+        # reverse because we return in order of the newest matches
+        rms.reverse()
+        # test that pagination will return all valid results in sorted order
+        # by last updated timestamp
+        result, next_pt = self.search_registered_models("name LIKE 'RM%'", max_results=5)
+        self.assertNotEqual(next_pt, None)
+        self.assertEqual(result, rms[0:5])
```
ankitmathur-db, line 508: We don't use sets here because there is an expected order

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -105,6 +107,7 @@ def create_registered_model(self, name):
         Create a new registered model in backend store.
 
         :param name: Name of the new model. This is expected to be unique in the backend store.
+        :param creation_time: Time when the registered model is created - for use in tests
```
mparkhe, line 110: remove this

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -382,98 +382,139 @@ def search_versions(filter_string):
...
         assert mvds[0].description == "Online prediction model!"
 
+    def search_registered_models(self,
+                                 filter_string,
+                                 page_token=None,
+                                 max_results=10):
+        return self.store.search_registered_models(filter_string, page_token, max_results)
+
+    def get_rm_names(self, rms):
+        return [rm.name for rm in rms]
```
mparkhe, line 392: looks like these 2 methods are always used together. would it make sense to always convert registered models -> names and token from this `_search_registered_models` method. I would make it the `_` private method prefix

mlflow/store/model_registry/__init__.py
```diff
@@ -0,0 +1,2 @@
+SEARCH_REGISTERED_MODEL_MAX_RESULTS_DEFAULT = 1000
+SEARCH_REGISTERED_MODEL_MAX_RESULTS_THRESHOLD = 50000
```
mparkhe, line 2: let's lower the defaults

```
SEARCH_REGISTERED_MODEL_MAX_RESULTS_DEFAULT = 100
SEARCH_REGISTERED_MODEL_MAX_RESULTS_THRESHOLD = 1000
```

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -382,98 +382,139 @@ def search_versions(filter_string):
...
+            "name LIKE 'RM%'", page_token=result.token, max_results=20)
+        self.assertNotEqual(result.token, None)
+        self.assertEqual(self.get_rm_names(result), rms[15:35])
+
+        result = self.search_registered_models(
+            "name LIKE 'RM%'", page_token=result.token, max_results=100)
+        # assert that page token is None
+        self.assertEqual(result.token, None)
+        self.assertEqual(self.get_rm_names(result), rms[35:])
+
```
mparkhe, line 509: these tests look good -- i would write an additional test that starts with a `None` token; and repeated calls search registered models with say `max_results = 5` until no more. In each round, collect the results and ensure that backend returns all 50 RMs

mlflow/store/model_registry/sqlalchemy_store.py
ankitmathur-db: Whoops sorry

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -247,9 +272,14 @@ def search_registered_models(self, filter_string):
         with self.ManagedSessionMaker() as session:
             if self.db_type == SQLITE:
                 session.execute("PRAGMA case_sensitive_like = true;")
-            sql_registered_models = session.query(SqlRegisteredModel).filter(*conditions).all()
+            sql_registered_models = session.query(SqlRegisteredModel)\
+                .filter(*conditions)\
+                .order_by(SqlRegisteredModel.name.asc())\
+                .offset(offset)\
```
mparkhe, line 278: is `offset(0)` is the same as not applying the offset. Or is it better to conditionally apply that clause?

```
query = session.query(SqlRegisteredModel).filter(*conditions).order_by(SqlRegisteredModel.name.asc()).limit(max_results)
if page_token:
  query = query.offset(SearchUtils.parse_start_offset_from_page_token(page_token))
sql_registered_models = query.all()
```

Is there a functional / performance difference one way or another?

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -382,98 +382,152 @@ def search_versions(filter_string):
...
+        # test flow with fixed max_results
+        returned_rms = []
+        result = self._search_registered_models(
+            "name LIKE 'RM%'", page_token=None, max_results=5)
+        returned_rms.extend(result)
+        while result.token:
+            result = self._search_registered_models(
+                "name LIKE 'RM%'", page_token=result.token, max_results=5)
+            returned_rms.extend(result)
+        self.assertEqual(set(rms), set(returned_rms))
```
mparkhe, line 498: Why did you need to convert this list to a set? Would be beneficial to compare with expected ordering as well.

mparkhe (commented):
Looks good! One small thing that I missed yesterday about applying `offset` clause conditionally.

mlflow/store/model_registry/sqlalchemy_store.py
ankitmathur-db: There doesn't appear to be a functional difference, but I can't find anything definitive about a performance impact. Regardless, it seems reasonable to not apply the offset if the offset isn't attainable, so I changed it to do that.

Notably, I didn't take your entire suggestion, since we use the offset variable to compute the `next_page_token`, so I'd rather still call the parsing function every time.

tests/store/model_registry/test_sqlalchemy_store.py
ankitmathur-db: Yup - fixed

mparkhe (approved):
LGTM

PR #2939, opened 2020-06-16
[Search RM Pagination] switch listRegisteredModels to be paginated

ankit-db:

See chain of PRs that are adding a paginated version of listRegisteredModels 

## How is this patch tested?

Unit testing

Components 
Model Registry

Interface 

Language 

Integrations


mlflow/java/client/src/main/java/org/mlflow/api/proto/ModelRegistry.java
```diff
@@ -10965,6 +10965,32 @@ public GetRegisteredModel parsePartialFrom(
...
     long getMaxResults();
+
+    /**
+     * <pre>
+     * Pagination token to go to the next page based on a previous query.
+     * </pre>
+     *
+     * <code>optional string page_token = 2;</code>
+     */
+    boolean hasPageToken();
```
andychow-db, line 10976: Nit: `page` is a word and not really understood as an abbreviation for `pagination`. Perhaps `hasPaginationToken` would be better?

mlflow/store/model_registry/rest_store.py
```diff
@@ -95,16 +95,26 @@ def delete_registered_model(self, name):
...
+        :return: PagedList of :py:class:`mlflow.entities.model_registry.RegisteredModel` objects.
         """
-        req_body = message_to_json(ListRegisteredModels())
+        req_body = message_to_json(ListRegisteredModels(
+            page_token=page_token,
+            max_results=max_results))
         response_proto = self._call_endpoint(ListRegisteredModels, req_body)
-        return [RegisteredModel.from_proto(registered_model)
-                for registered_model in response_proto.registered_models]
+        print(response_proto)
```
andychow-db, line 113: Remove `print`

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -125,30 +125,83 @@ def test_delete_registered_model(self):
...
+        # test that providing too large of a max_results throws
+        with self.assertRaises(MlflowException) as exception_context:
+            self._list_registered_models(page_token="evilhax", max_results=1e15)
+            assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+        self.assertIn("Invalid value for request parameter max_results",
+                      exception_context.exception.message)
+        # list should not return deleted models
+        self.store.delete_registered_model(name=f"RM{0:03}")
+        self.assertEqual(set(self._list_registered_models(max_results=100)),
+                         set(rms[1:]))
```
andychow-db, line 203: Nit: I'd prefer that a test function test only one scenario rather than have one test function validate multiple scenarios.

tests/tracking/test_model_registry.py
```diff
@@ -109,6 +109,25 @@ def test_create_and_query_registered_model_flow(mlflow_client, backend_store_uri
     assert_is_between(start_time, end_time, registered_model_detailed.creation_timestamp)
     assert_is_between(start_time, end_time, registered_model_detailed.last_updated_timestamp)
     assert [name] == [rm.name for rm in mlflow_client.list_registered_models() if rm.name == name]
+    # clean up test
+    mlflow_client.delete_registered_model(name)
+
+
+def test_create_and_query_registered_model_flow_complex(mlflow_client, backend_store_uri):
```
andychow-db, line 116: Test name is not clear. Add a description of what "complex" means.

andychow-db (commented):
Looking good. Just a minor change and some things to consider

mlflow/java/client/src/main/java/org/mlflow/api/proto/ModelRegistry.java
andychow-db: This might be a larger discussion. Feel free to just say no here because it's already been implemented this way in other parts of our code.

mlflow/store/model_registry/abstract_store.py
```diff
@@ -69,9 +69,12 @@ def delete_registered_model(self, name):
         pass
 
     @abstractmethod
-    def list_registered_models(self):
+    def list_registered_models(self, page_token, max_results):
```
sueann, line 72: nit: `(self, max_results, page_token)` seems more intuitive to me and it matches the ordering in the proto. we might as well make it a convention to follow the ordering in the proto.

mlflow/store/model_registry/rest_store.py
```diff
@@ -95,16 +95,26 @@ def delete_registered_model(self, name):
             name=name))
         self._call_endpoint(DeleteRegisteredModel, req_body)
 
-    def list_registered_models(self):
+    def list_registered_models(self,
+                               page_token,
+                               max_results):
```
sueann, line 100: nit: should fit in one line

mlflow/tracking/client.py
```diff
@@ -399,13 +400,17 @@ def delete_registered_model(self, name):
         self._get_registry_client().delete_registered_model(name)
 
     @experimental
-    def list_registered_models(self):
+    def list_registered_models(self,
+                               page_token=None,
+                               max_results=SEARCH_REGISTERED_MODEL_MAX_RESULTS_DEFAULT):
         """
         List of all registered models.
```
sueann, line 408: params? i think this docstring will show up in the API docs.

mlflow/java/client/src/main/java/org/mlflow/api/proto/ModelRegistry.java
ankitmathur-db: It's definitely worth considering!

For the time being, though, we do have `page_token` pretty much everywhere in our paginated API. It does appear to be somewhat? canonical in other APIs, like [this](https://googleapis.dev/python/google-api-core/latest/page_iterator.html) and [this](https://developers.google.com/google-ads/api/docs/concepts/retrieving-objects)

ankitmathur-db:
Thanks for the review @andychow-db ! PR updated with regards to your comments

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -125,30 +125,83 @@ def test_delete_registered_model(self):
...
+        rms = [self._rm_maker(f"RM{i:03}").name for i in range(50)]
+
+        # test flow with fixed max_results
+        returned_rms = []
+        result = self._list_registered_models(page_token=None, max_results=5)
+        returned_rms.extend(result)
+        while result.token:
+            result = self._list_registered_models(page_token=result.token, max_results=5)
+            returned_rms.extend(result)
+        self.assertEqual(set(rms), set(returned_rms))
```
sueann, line 180: this seems largely redundant with the checks below, except that here we are testing a case where the last call will return `max_results` registered models. so can we
1. make max_results maybe 25, so it doesn't loop so much and 
2. verify after the loop that `result.token` is `None` and `result` has `max_results` elements? we can also check that we loop the expected number of times.

sueann (commented):
looks mostly good to me - just some small comments inline.

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -155,7 +155,19 @@ def test_list_registered_model(self):
...
+
+    def test_list_registered_model_paginated_ascending_max_results(self):
```
sueann, line 170: i think something like `test_list_registered_model_paginated_returns_in_correct_order` would be more clear

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -155,7 +155,19 @@ def test_list_registered_model(self):
         self.assertEqual(set(self._list_registered_models()),
                          set(["A", "BB", "AB", "BBC"]))
 
-    def test_list_registered_model_paginated(self):
+    def test_list_registered_model_paginated_standard(self):
```
sueann, line 158: maybe `test_list_registered_model_paginated_last_page` (after previous suggestions)

mlflow/java/client/src/main/java/org/mlflow/api/proto/ModelRegistry.java
andychow-db: Oh interesting! I think it's fine then if Google is doing this convention.

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -125,30 +125,97 @@ def test_delete_registered_model(self):
...
+        rms = [self._rm_maker(f"RM{i:03}").name for i in range(50)]
+
+        # test flow with fixed max_results
+        returned_rms = []
+        result = self._list_registered_models(page_token=None, max_results=5)
+        returned_rms.extend(result)
+        while result.token:
+            result = self._list_registered_models(page_token=result.token, max_results=5)
+            returned_rms.extend(result)
+        self.assertEqual(set(rms), set(returned_rms))
```
sueann, line 168: we can remove 173-180 now right?

mlflow/store/model_registry/abstract_store.py
ankitmathur-db: Updated the function signature!

tests/store/model_registry/test_sqlalchemy_store.py
ankitmathur-db: Added this test at Mani's suggestion earlier - updated the test so that it has a higher max_results and verified None token and that the result is of the right size

tests/store/model_registry/test_sqlalchemy_store.py
ankitmathur-db: Yup sorry that was an oversight

sueann (approved):
lgtm - thanks for addressing all the comments!

sueann:
Actually - it looks like some list_registered_models tests are failing in travis - we need to fix them before merging.

sueann (changes requested):
we should figure out the last page business

ankitmathur-db:
@sueann - replied on the original comment with an explanation for what's happening. What do you think the right course of action is? If we want to not return a `next_page_token` if there are no results left, we could also do something like querying with `limit + 1`

sueann (approved):
lgtm - thanks!

PR #2946, opened 2020-06-16
Remove deprecated fields on client APIs

mparkhe:
What changes are proposed in this pull request?

These fields had been marked as deprecated since 1.7 after new APIs were added. Removing these fields in next release.

 - update_model_version: stage
 - update_registered_model: new_name

## How is this patch tested?

Fixed tests

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/tests/tracking/test_client.py
+++ b/tests/tracking/test_client.py
@@ -176,37 +176,31 @@ def test_client_registry_operations_raise_exception_with_unsupported_registry_st
             assert exc.value.error_code == ErrorCode.Name(FEATURE_DISABLED)
 
 
-def test_update_registered_model_compatibility_layer(mock_registry_store):
+def test_update_registered_model(mock_registry_store):
     """
-    This test makes sure that old (now deprecated) apis work as expected after api update.
-    Update registered model no longer accepts new name, but the client should translate it to
-    rename call.
+    Update registered model no longer supports name change.
     """
     expected_return_value = "some expected return value."
     mock_registry_store.rename_registered_model.return_value = expected_return_value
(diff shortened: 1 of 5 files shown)


tests/tracking/test_client.py
```diff
@@ -176,37 +176,31 @@ def test_client_registry_operations_raise_exception_with_unsupported_registry_st
             assert exc.value.error_code == ErrorCode.Name(FEATURE_DISABLED)
 
 
-def test_update_registered_model_compatibility_layer(mock_registry_store):
+def test_update_registered_model(mock_registry_store):
     """
-    This test makes sure that old (now deprecated) apis work as expected after api update.
-    Update registered model no longer accepts new name, but the client should translate it to
-    rename call.
+    Update registered model no longer supports name change.
```
sueann, line 181: do we really need these deprecation tests - as something that will run forever? makes sense for this PR, but seems unnecessary for the future (we have other tests calling these methods, and if someone tries to use the old arguments, they will see python error thrown).

sueann (approved):
lgtm - just one question for you

tests/tracking/test_client.py
mparkhe: I removed the deprecated part so now these test are only the legit non-deprecated arguments to `update_*` methods. Maybe these methods can move elsewhere.

sueann: Ah, I see, these are the actual only tests for `update_registered_model` and `update_model_version`. I was confused because of the comment that suggested that we were specifically (and only) testing the deprecated behavior.

PR #2959, opened 2020-06-18
Fix incorrect usage of nested `PropTypes`

harupy:

Fix incorrect usage of nested `PropTypes`.

```javascript
class Example extends Component {
  static proptypes = {

    // wrong: this does not raise a warning for an invalid prop type
    strArray: PropTypes.arrayOf(String).isRequired;

    // correct
    strArray: PropTypes.arrayOf(PropTypes.string).isRequired;
  }
}
```

## How is this patch tested?

- The existing tests pass.
- Manually verify that invalid-prop warnings are not raised in the dev console when running the UI

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/server/js/src/experiment-tracking/components/CompareRunContour.js
+++ b/mlflow/server/js/src/experiment-tracking/components/CompareRunContour.js
@@ -14,10 +14,10 @@ import CompareRunUtil from './CompareRunUtil';
 
 export class CompareRunContour extends Component {
   static propTypes = {
-    runInfos: PropTypes.arrayOf(RunInfo).isRequired,
-    metricLists: PropTypes.arrayOf(Array).isRequired,
-    paramLists: PropTypes.arrayOf(Array).isRequired,
-    runDisplayNames: PropTypes.arrayOf(String).isRequired,
+    runInfos: PropTypes.arrayOf(PropTypes.instanceOf(RunInfo)).isRequired,
+    metricLists: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.object)).isRequired,
+    paramLists: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.object)).isRequired,
+    runDisplayNames: PropTypes.arrayOf(PropTypes.string).isRequired,
   };
 
   // Size limits for displaying keys and values in our plot axes and tooltips
--- a/mlflow/server/js/src/common/components/tables/EditableFormTable.js
+++ b/mlflow/server/js/src/common/components/tables/EditableFormTable.js
@@ -54,8 +54,8 @@ class EditableCell extends React.Component {
 
 export class EditableTable extends React.Component {
   static propTypes = {
-    columns: PropTypes.arrayOf(Object).isRequired,
-    data: PropTypes.arrayOf(Object).isRequired,
+    columns: PropTypes.arrayOf(PropTypes.object).isRequired,
+    data: PropTypes.arrayOf(PropTypes.object).isRequired,
     onSaveEdit: PropTypes.func.isRequired,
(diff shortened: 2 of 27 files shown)


ankitmathur-db:
Out of curiosity, why is this incorrect?

harupy:
@ankitmathur-db Thanks for the comment! As explained in [the official doc](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes), when we say:

```
PropTypes.arrayOf(String)
```

`PropTypes` thinks `String` is a validator function and emits a warning if the function throws an error, but `String` just casts a given value to a string and DOES NOT throw an error. This causes `PropTypes` to ignore incorrect prop types.
```
> String(1)
'1'
```

You can confirm this behavior using the example below:

```javascript
// Sample.js
import React, { Component } from 'react';
import PropTypes from 'prop-types';

export class Sample extends Component {
  static propTypes = {
    stringArray1: PropTypes.arrayOf(String).isRequired, // wrong
    stringArray2: PropTypes.arrayOf(PropTypes.string).isRequired, // correct
  };

  render() {
    const { stringArray1, stringArray2 } = this.props;
    return (
      <div>
        <div>{stringArray1.join('\n')}</div>
        <div>{stringArray2.join('\n')}</div>
      </div>
    );
  }
}

// Sample.test.js
import React from 'react';
import { mount } from 'enzyme';
import { Sample } from './Sample';

describe('Sample', () => {
  test('test', () => {
    const props = {
      stringArray1: [1, 2, 3], // DOES NOT raise a warning
      stringArray2: [1, 2, 3], // raises a warning
    };
    const wrapper = mount(<Sample {...props} />);
    expect(wrapper.length).toBe(1);
  });
});
```

```
> npm run test

 PASS  src/common/components/IncorrectPropTypes.test.js
  Sample
    ✓ test (20ms)

  console.error node_modules/react/node_modules/prop-types/checkPropTypes.js:20
    Warning: Failed prop type: Invalid prop `stringArray2[0]` of type `number` supplied to `Sample`, expected `string`.
        in Sample (at IncorrectPropTypes.test.js:11)
```

harupy:
cc @Zangr

mlflow/server/js/src/experiment-tracking/components/CompareRunContour.js
```diff
@@ -14,10 +14,10 @@ import CompareRunUtil from './CompareRunUtil';
 
 export class CompareRunContour extends Component {
   static propTypes = {
-    runInfos: PropTypes.arrayOf(RunInfo).isRequired,
-    metricLists: PropTypes.arrayOf(Array).isRequired,
-    paramLists: PropTypes.arrayOf(Array).isRequired,
-    runDisplayNames: PropTypes.arrayOf(String).isRequired,
+    runInfos: PropTypes.arrayOf(PropTypes.instanceOf(RunInfo)).isRequired,
+    metricLists: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.object)).isRequired,
+    paramLists: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.object)).isRequired,
```
harupy, line 19: We can be more strict here using `PropTypes.shape`.

Zangr (approved):
Thanks for the fix @harupy ! LGTM

mlflow/server/js/src/experiment-tracking/components/CompareRunContour.js
harupy: > Hopefully going forward we can spend some time converting to typescript

Agree with using typescript which should make the UI development much easier.

> Hopefully going forward we can spend some time converting to typescript。

I'm actually working on this. I'll submit a PR when it's ready.

PR #2975, opened 2020-06-19
Specify a minimum Pandas version that is compatible with model schemas

dbczumar:
## What changes are proposed in this pull request?

When we added support for Model Schemas to MLflow, a later version of Pandas than the one specified in this example became a hard dependency because we now (interestingly) import the `mlflow/types/schema.py` file when performing MLflow client operations (stacktrace below). This module imports `pandas.core.dtypes.dtypes.PandasExtensionDtype`, which does not exist in older versions of Pandas (e.g., `0.22`). Accordingly, users running older versions of Pandas are likely to see failures when using the latest version of MLflow 1.9. To fix this issue, this PR introduces a minimum Pandas version requirement for MLflow of `1.0.0`.

## How is this patch tested?

This PR has been tested by creating a conda environment with an old Pandas version (`0.22.0`), installing MLflow, ensuring that `Pandas` was upgraded, and ensuring that training code from the `mlflow-example` (`train.py`) repository executes correctly: https://github.com/mlflow/mlflow-example/blob/master/train.py.

## Release Notes

Specify a minimum Pandas version requirement to fix model-schema-related failures encountered when using MLflow with old versions of Pandas.

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [X] Yes. Give a description of this change to be included in the release notes for MLflow users.

See release note

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry
- [X] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients

Integrations
[...]

PR #2988, opened 2020-06-22
Remove use of too-new Pandas API

aarondav:

When we added schema, we accidentally added a dependency on `pandas.core.dtypes.dtypes. PandasExtensionDtype`, which was added in Pandas 1.0. Since we don't really need this type, we simply remove the import here.

## How is this patch tested?

Followed the same steps as #2975 -- ran [this file](https://github.com/mlflow/mlflow-example/blob/master/train.py) using a conda env that had pandas 0.22. Without this change, it failed, and with this change it works.


smurching (approved):
LGTM

PR #3000, opened 2020-06-24
[docs] Fetching a MLflow model from the registry

andychow-db:
Updated the model registry documentation to show how to fetch a registered model using `load_model` and the `models:/` model_uri syntax. Secondly, added documentation about how to serve a model registry model via the CLI.

## How is this patch tested?

Screen Shot 2020-06-29 at 4 56 59 PM

Screen Shot 2020-06-29 at 4 53 35 PM

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -196,7 +196,6 @@ After you have registered an MLflow model, you can serve the model as a service 
 
     #!/usr/bin/env sh
 
-    echo "Deploying Production model name=sk-learn-random-forest-reg-model"
     # Set environment variable for the tracking URL where the Model Registry resides
     export MLFLOW_TRACKING_URI=http://localhost:5000
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -193,6 +193,7 @@ Serve a MLflow Model from Model Registry
 After you have registered an MLflow model, you can serve the model as a service on your host.
 
 .. code-block:: bash
+
     #!/usr/bin/env sh
 
     echo "Deploying Production model name=sk-learn-random-forest-reg-model"
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -155,7 +155,7 @@ While the method above creates an empty registered model with no version associa
 Fetching an MLflow Model from the Model Registry
 ------------------------------------------------
 
-After you have registered an MLflow model, you can fetch that model using :meth:`~mlflow.pyfunc.load_model`. 
+After you have registered an MLflow model, you can fetch that model using :meth:`~mlflow.<model_flavor>.load_model`, or more generally, :meth:`~mlflow.pyfunc.load_model`. 
 
 **Fetch a specific model version**
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -157,7 +157,7 @@ Run a MLflow Model from the Model Registry
 
 After you have registered an MLflow model, you can run that model using :meth:`~mlflow.sklearn.load_model`. 
 
-**Run a specific model version**
+**Load a specific model version**
 
 To fetch a specific model version, just supply that version name as part of the model URI.
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -172,6 +172,8 @@ To fetch a specific model version, just supply that version number as part of th
         model_uri=f"models:/{model_name}/{model_version}
     )
 
+    model.predict(data)
+
 **Fetch the latest model version in a specific stage**
 
 To fetch a model version by stage, simply provide the model stage as part of the model URI, and it will fetch the most recent version of the model in that stage.
@@ -186,6 +188,8 @@ To fetch a model version by stage, simply provide the model stage as part of the
     model = mlflow.pyfunc.load_model(
         model_uri=f"models:/{model_name}/{stage}
     )
+
+    model.predict(data)
 
 Serving an MLflow Model from Model Registry
 -------------------------------------------
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -155,7 +155,7 @@ While the method above creates an empty registered model with no version associa
 Run a MLflow Model from the Model Registry
 ------------------------------------------
 
-After you have registered an MLflow model, you can run that model using :meth:`~mlflow.sklearn.load_model`. 
+After you have registered an MLflow model, you can load to run that model using :meth:`~mlflow.<model_flavor>.load_model`. 
 
 **Load a specific model version**
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -219,7 +219,7 @@ You can download the model by using :meth:`~mlflow.tracking.MlflowClient.get_mod
    print(model_artifact_uri)
 
 
-After obtaining the download uri, you can download it via the Click
+After obtaining the download uri, you can download it via the MLflow CLI.
 
 .. code-block:: py
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -155,7 +155,7 @@ While the method above creates an empty registered model with no version associa
 Fetching an MLflow Model from the Model Registry
 ------------------------------------------------
 
-After you have registered an MLflow model, you can fetch that model using :meth:`~mlflow.<model_flavor>.load_model`, or more generally, :meth:`~mlflow.pyfunc.load_model`. 
+After you have registered an MLflow model, you can fetch that model using ``mlflow.<model_flavor>.load_model()``, or more generally, :meth:`~mlflow.pyfunc.load_model`. 
 
 **Fetch a specific model version**
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -174,7 +174,7 @@ To fetch a specific model version, just supply that version name as part of the 
 
 **Run the latest model version in a specific stage**
 
-Model Registry allows you to transition multiple versions of the model into the same or different stages. To fetch the model in its stage name, simply provide the model stage label as part of the model URI, and it will fetch the most recent version of the model in that stage.
+To fetch a model version by stage, simply provide the model stage as part of the model URI, and it will fetch the most recent version of the model in that stage.
 
 .. code-block:: py
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -152,43 +152,43 @@ While the method above creates an empty registered model with no version associa
         run_id="d16076a3ec534311817565e6527539c0"
     )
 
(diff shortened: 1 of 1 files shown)


docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
         run_id="d16076a3ec534311817565e6527539c0"
     )
 
+Run a MLflow Model from the Model Registry
+------------------------------------------
+
+After you have registered an MLflow model, you can run that model using :meth:`~mlflow.sklearn.load_model`.
```
sueann, line 158:
```suggestion
After you have registered an MLflow model, you can load to run that model using :meth:`~mlflow.<model_flavor>.load_model`. 
```

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
         run_id="d16076a3ec534311817565e6527539c0"
     )
 
+Run a MLflow Model from the Model Registry
+------------------------------------------
+
+After you have registered an MLflow model, you can run that model using :meth:`~mlflow.sklearn.load_model`. 
+
+**Run a specific model version**
```
sueann, line 160:
```suggestion
**Load a specific model version**
```

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
...
+
+To fetch a specific model version, just supply that version name as part of the model URI.
```
sueann, line 162:
```suggestion
To fetch a specific model version, just supply that version number as part of the model URI.
```

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
...
+    import mlflow.sklearn
+
+    model_name = "sk-learn-random-forest-reg-model"
+    model_version = 1
+
+    model = mlflow.sklearn.load_model(
+        model_uri=f"models:/{model_name}/{model_version}
+    )
+
+**Run the latest model version in a specific stage**
```
sueann, line 175:
```suggestion
**Load the latest model version in a specific stage**
```

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
...
+
+Model Registry allows you to transition multiple versions of the model into the same or different stages. To fetch the model in its stage name, simply provide the model stage label as part of the model URI, and it will fetch the most recent version of the model in that stage.
```
sueann, line 177:
```suggestion
To fetch a model version by stage, simply provide the model stage as part of the model URI, and it will fetch the most recent version of the model in that stage.
```

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
         run_id="d16076a3ec534311817565e6527539c0"
     )
 
+Run a MLflow Model from the Model Registry
```
sueann, line 155:
```suggestion
Use an MLflow Model from the Model Registry
```

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
...
+    import mlflow.sklearn
+
+    model_name = "sk-learn-random-forest-reg-model"
+    stage = 'Staging'
+
+    model = mlflow.sklearn.load_model(
+        model_uri=f"models:/{model_name}/{stage}
+    )
+
+Serve a MLflow Model from Model Registry
```
sueann, line 190:
```suggestion
Serve an MLflow Model from Model Registry
```

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
...
+----------------------------------------
+
+After you have registered an MLflow model, you can serve the model as a service on your host.
+
+.. code-block:: bash
+
+    #!/usr/bin/env sh
+
+    echo "Deploying Production model name=sk-learn-random-forest-reg-model"
```
sueann, line 199: do you need this line?

docs/source/model-registry.rst
```diff
@@ -152,6 +152,57 @@ While the method above creates an empty registered model with no version associa
         run_id="d16076a3ec534311817565e6527539c0"
     )
 
+Run a MLflow Model from the Model Registry
+------------------------------------------
+
+After you have registered an MLflow model, you can run that model using :meth:`~mlflow.sklearn.load_model`. 
+
```
sueann, line 159: Can we also mention `MlflowClient.get_model_version_download_uri(name, version)` as an explicit way to get the URI and download it? Might be also worth mentioning the java version. We don't need to provide an example for these I think because that distracts from the main flow - but we could link to the docs perhaps.

sueann (commented):
looks mostly good to me - some nits and 1 real comment inline. cc @stbof @andreakress

docs/source/model-registry.rst
andychow-db: I suppose not. We can just remove it.

docs/source/model-registry.rst
andychow-db: I added a new section with an example.

andychow-db: Perhaps we should add the param `artifact_uri` to `MlflowClient.downloadArtifacts`. It would be trivial to do so. What do you think?

sueann: `MlflowClient.download_artifacts` is a run-specific API. Are you thinking we need an API to download any artifact URI? I think we used to have one but we cleaned it up as part of the 1.0 release. There is the internal `mlflow.tracking.artifact_utils._download_artifact_from_uri` (which i think eventually powers the CLI command below)... it would probably be useful to have a public method for that, but I think the community just hasn't asked for it (AFAIK).

docs/source/model-registry.rst
dmatrix: This was part of the workshop I was doing, but not needed here.

docs/source/model-registry.rst
dmatrix: This section is tailored more toward the Model Registry Workflow API; we might not want to conflate it. It could serve a better purpose to make this part of the API calls in the `MLflowclient` docs.

dmatrix: A couple of observations.

1. "Run an MLflow Model from the Model Registry"

This heading is not consistent with the use of gerunds in the other subheadings under this workflow. Also, the verb "Run", not sure it quite fits with model: you serve a mode; you train a model; you score (or infer from) a model. "Run" usage may be more appropriate with, say, something like "Run" an MLflow Project. 

I think more fitting to be consistent with the flow of the workflow APIs would be "Fetching (or Retrieving) an MLflow model" 
(We don't need to use the prepositional phrase "from the Model Registry,"  as we are already in the section). 

2. Second, you can fetch models from the registry using both native `load_model` and `pyfunc_load_model`. I think this is imperative because to serve a model from the CLI, we only support pyfunc flavors.

docs/source/model-registry.rst
```diff
@@ -152,6 +152,81 @@ While the method above creates an empty registered model with no version associa
...
+    import mlflow.sklearn
+
+    model_name = "sk-learn-random-forest-reg-model"
+    stage = 'Staging'
+
+    model = mlflow.sklearn.load_model(
+        model_uri=f"models:/{model_name}/{stage}
+    )
+
+Serve an MLflow Model from Model Registry
```
andychow-db, line 190: Serving

andychow-db:
Actions

1. remove download 
2. add get latest across all stages
3. gerand fix
4. create task to add download apis to python (same as java client)

docs/source/model-registry.rst
```diff
@@ -152,6 +152,60 @@ While the method above creates an empty registered model with no version associa
         run_id="d16076a3ec534311817565e6527539c0"
     )
 
+Fetching an MLflow Model from the Model Registry
+------------------------------------------------
+
+After you have registered an MLflow model, you can fetch that model using :meth:`~mlflow.pyfunc.load_model`.
```
sueann, line 158: i think in the code example below using `mlflow.pyfunc.load_model` is fine, but here we should say `mlflow.<model_flavor>.load_model` to match `mlflow.<model_flavor>.log_model` above (line 92).

docs/source/model-registry.rst
sueann: ok looking above at the section intro, this is meant to be for API workflow using the python `MlflowClient` and `models` methods, so it makes sense not to include the model downloading stuff here. 

I still think it is very valuable for those who don't want to load the model using `<flavor>.load_model` -- could it make sense to include the downloading example as a whole different section for how to use model registry at inference time in arbitrary contexts? then we can list the python (`MlflowClient.get_model_version_download_uri`), REST API, and java ways. too much?

dmatrix: Yes, true, we should evolve these docs, as part of the model registry general usage, with expanded sections: Accessing or downloading model artifacts: a) models b) gifs/pdfs c) schemas d) tags/values, etc.

docs/source/model-registry.rst
dmatrix: Might be worth to mention, there are two ways to fetch a model: `<model_flavor>.load_model` and `mlflow.pyfunc.load_model`. I think @andychow-db and I discussed it, but we felt it might confuse the developer. However, we do mention above that there are two ways to register the model:`mlflow.<model_flavor>.log_model()` and `mlflow.register_model()`. Likewise, it's not unreasonable to mention that there are two ways to fetch or retrieve a registered model. But focus on the example showing `mlflow.pyfunc.load_model`, because only the `mlflow.pyfunc.load_model` can be served using `mlflow serve models -m URI`

@sueann is that reasonable?

Other than that it LGTM.

sueann: mm my comment was even simpler - just that we should be consistent in this paragraph with line 92. will add a suggestion here

sueann, line 158:
```suggestion
After you have registered an MLflow model, you can fetch that model using :meth:`~mlflow.<model_flavor>.load_model`, or more generally, :meth:`~mlflow.pyfunc.load_model`. 
```

docs/source/model-registry.rst
sueann: ok we can do this in another PR at a later date

docs/source/model-registry.rst
dmatrix: Yes, I was making an argument to mention both ways. And your simple suggestion below is spot on.

dmatrix: LGTM!

docs/source/model-registry.rst
dmatrix: Yes, we can!

sueann (approved):
lgtm - thanks!

Issue #2991, opened 2020-06-23
[BUG] AzureBlobArtifactRepository.list_artifacts() wrong


wuqunfei:
### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [x] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **OS Platform and Distribution (Mac/Linux/Docker)**:
- **MLflow version (1.9.0/1.8.0)**:
- **Python version(3.6/3.7/3.8)**:
- **SetUp AzureStorage**:

### Describe the problem
Azure Storage aritficats can't download from UI,  in 1.8.0 and 1.9.0 Version, because all AzureBlobArtifactRepository get list method is wrong. 

    def list_artifacts(self, path=None):

alwary return files >=1, even the path is a file. It can impact is_directory method

    class ArtifactRepository:
        def _is_directory(self, artifact_path):
        listing = self.list_artifacts(artifact_path)
        return len(listing) > 0
In the end, in azure download file, in backend, always create direction instead of file name in ArtifactRepository.py abastract class.

    if self._is_directory(artifact_path) 
        return download_artifact_dir(artifact_path)
    else:
        return download_file(artifact_path)

@Asgeir Berland <asgeir.berland@gmail.com>, in this commit. Maybe something broken this logic https://github.com/mlflow/mlflow/pull/2594

### Code to reproduce issue

 1. install azure-storage-blob==12.3.2 and mlflow==1.9.0
 2. Set up AZURE_STORAGE_ARTIFACT in env
 3. Set up AZURE_STORAGE_ACCESS_KEY in env
 4. `mlflow server \
    --backend-store-uri /yourdisk \
    --default-artifact-root $AZURE_STORAGE_ARTIFACT/ \
    --host 0.0.0.0`
  
 5. Training once and try to download the artifact from one of your experiments http://localhost:5000/#/experiments/1

### Other info / logs
Get Artifact
    http://localhost:5000/get-artifact?path=random-forest-model%2Fmodel.pkl&run_uuid=79379700d86e45cd8aac1a33ebc7c651

Error Display

      File "/usr/local/lib/python3.8/site-packages/mlflow/server/handlers.py", line 182, in get_artifact_handler
[...]

PR #3014, opened 2020-06-29
Fix Azure Blob Artifact Repo not returning single files

Trollgeir:

The `list_artifact()` method in the Azure Blob Artifact Repo did not conform with the ArtifactRepository's `list_artifact()` specifications. More precisely, it did not return an empty list when the target was a single file. This resulted in the inability to download and view individual artifacts from the UI (see https://github.com/mlflow/mlflow/issues/2991).

## How is this patch tested?

Local testing with private azure repository.
Unittest.

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/store/artifact/azure_blob_artifact_repo.py
+++ b/mlflow/store/artifact/azure_blob_artifact_repo.py
@@ -109,6 +109,11 @@ class AzureBlobArtifactRepository(ArtifactRepository):
             else:  # Just a plain old blob
                 file_name = posixpath.relpath(path=r.name, start=artifact_path)
                 infos.append(FileInfo(file_name, False, r.size))
+        # The list_artifacts API expects us to return an empty list if the
+        # the path references a single file.
+        rel_path = dest_path[len(artifact_path)+1:]
+        if (len(infos) == 1) and not infos[0].is_dir and (infos[0].path == rel_path):
+            return []
         return sorted(infos, key=lambda f: f.path)
 
     def _download_file(self, remote_file_path, local_path):
--- a/tests/store/artifact/test_azure_blob_artifact_repo.py
+++ b/tests/store/artifact/test_azure_blob_artifact_repo.py
@@ -86,6 +86,17 @@ def test_list_artifacts_empty(mock_client):
     assert repo.list_artifacts() == []
 
 
+def test_list_artifacts_single_file(mock_client):
+    repo = AzureBlobArtifactRepository(TEST_URI, mock_client)
+
+    # Evaluate single file
+    blob_props = BlobProperties()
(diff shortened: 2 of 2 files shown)


arjundc-db:
Hi @Trollgeir, thank you for finding the bug and implementing the fix.
Could you also add unit tests for this change.

Thanks again.

Trollgeir:
@arjundc-db done!

wuqunfei (approved):
It looks fine

mlflow/store/artifact/azure_blob_artifact_repo.py
```diff
@@ -109,6 +109,9 @@ def list_artifacts(self, path=None):
             else:  # Just a plain old blob
                 file_name = posixpath.relpath(path=r.name, start=artifact_path)
                 infos.append(FileInfo(file_name, False, r.size))
+        # If the path is directed at a single file, return an empty list.
+        if (len(infos) == 1) and (infos[0].is_dir is False):
```
arjundc-db, line 113: Sorry for missing this earlier
```suggestion
        if (len(infos) == 1) and (infos[0].is_dir is False) and (infos[0].path == path):
```
This is because if a directory contains a single file and list_artifacts is called with the `path=directory` , the function should return:
[<FileInfo: file_size=<some-size>, is_dir=False, path='<directory>/<file>'>].
However, without the path comparison, the function would return an empty list, which would be incorrect.
(Thanks @dbczumar for pointing this out)

dbczumar: @arjundc-db Are we sure that `path` is the correct right-hand-side for this comparison? Are we sure it isn't `dest_path`?

Trollgeir: This doesn't work as the `FileInfo.path` has been stripped of its root path, which needs to be compared to `dest_path`.
The logic can be done before creation of the `FileInfo` object, as shown in https://github.com/mlflow/mlflow/pull/3014/commits/b628d2db4d76823ed3b50ff90aef3ef4d56e17ec
(At this point in the code we know it's not a directory, and if the path has an exact match there can't be multiple files)

Trollgeir: Sorry, I was overthinking this. Reverted back to @dbczumar 's suggestion and added the relative path:
https://github.com/mlflow/mlflow/pull/3014/commits/28be18c5c903250a186919bfb43c33ca27722071

wuqunfei (approved):

arjundc-db:
Awesome, thank you for the contribution @Trollgeir.

arjundc-db (approved):

wuqunfei:
thanks a lot @Trollgeir, wonderful. I will test it after a new version release.

PR #3023, opened 2020-06-29
[pagination] followups to clean up search registered models

### LINKED ISSUE #2976: [BUG] `mlflow.search_runs` does not raise error or warning when passed invalid `order_by` expression
`mlflow.search_runs` does not raise error or warning when passed invalid `order_by` expression. I think this behavior is dangerous. Let's consider a scenario where a user wants to select the best model from a bunch of runs and writes the code below:

```python
order_by = ["metrics.accuracy DECS"]  # "DECS" is a typo, should be "DESC"
sorted_runs = mlflow.search_runs(..., order_by=order_by)
best_run = sorted_runs.iloc[0]  # select the frist row
best_model = load_model_from_run(best_run)
```

What this code does is:
- The `order_by` expression `"metrics.accuracy DECS"` doesn't end with ` desc` or ` asc` so it's parsed as a metric named `accuracy DECS`.
- `search_runs` sorts runs by the metric `accuracy DECS` which doesn't exist and returns unsorted runs.
- `sorted_runs.iloc[0]` is obviously not the best run, **but no error or warning is raised**.

ankit-db:

This PR cleans up some issues with testing and edge cases with search registered models:

1. It adds tests to verify that lowercase and uppercase ordering keys work.
2. It changes search `order_by` parsing in `search_runs` and `search_registered_models` to throw if an invalid token is passed in

per discussion with @mparkhe and @sueann 

## How is this patch tested?

Unit tests and manual

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -371,13 +371,14 @@ class SearchUtils(object):
             raise MlflowException(f"Invalid order_by clause '{order_by}'. Could not be parsed.",
                                   error_code=INVALID_PARAMETER_VALUE)
         elif len(tokens) == 2:
-            if tokens[1].lower() not in cls.VALID_ORDER_BY_TAGS:
+            order_token = tokens[1].lower()
+            if order_token not in cls.VALID_ORDER_BY_TAGS:
                 raise MlflowException(f"Invalid ordering key in order_by clause '{order_by}'.",
                                       error_code=INVALID_PARAMETER_VALUE)
-            if tokens[1].lower() == cls.DESC_OPERATOR:
+            if order_token == cls.DESC_OPERATOR:
                 is_ascending = False
                 token_value = tokens[0]
-            elif tokens[1].lower() == cls.ASC_OPERATOR:
+            elif order_token == cls.ASC_OPERATOR:
                 token_value = tokens[0]
         return token_value, is_ascending
 
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -375,11 +375,8 @@ class SearchUtils(object):
             if order_token not in cls.VALID_ORDER_BY_TAGS:
                 raise MlflowException(f"Invalid ordering key in order_by clause '{order_by}'.",
                                       error_code=INVALID_PARAMETER_VALUE)
-            if order_token == cls.DESC_OPERATOR:
-                is_ascending = False
-                token_value = tokens[0]
-            elif order_token == cls.ASC_OPERATOR:
-                token_value = tokens[0]
+            is_ascending = (order_token == cls.ASC_OPERATOR)
+            token_value = tokens[0]
         return token_value, is_ascending
 
     @classmethod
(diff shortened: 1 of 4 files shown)


mlflow/utils/search_utils.py
```diff
@@ -363,11 +366,20 @@ def _validate_order_by_and_generate_token(cls, order_by):
...
+        elif len(tokens) == 2:
+            order_token = tokens[1].lower()
+            if order_token not in cls.VALID_ORDER_BY_TAGS:
+                raise MlflowException(f"Invalid ordering key in order_by clause '{order_by}'.",
+                                      error_code=INVALID_PARAMETER_VALUE)
+            if order_token == cls.DESC_OPERATOR:
+                is_ascending = False
+                token_value = tokens[0]
+            elif order_token == cls.ASC_OPERATOR:
+                token_value = tokens[0]
```
sueann, line 382: nit: could do
```
is_ascending  = (order_token == cls.ASC_OPERATOR)
token_value = tokens[0]
```

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -515,6 +515,11 @@ def test_search_registered_models(self):
         rms, _ = self._search_registered_models(f"name iLike '%blah%'")
         self.assertEqual(rms, [])
 
+        # confirm that ILIKE works for empty query
+        rms, _ = self._search_registered_models(f"name iLike '%%'")
+        # raise Exception()
```
sueann, line 520: ?

mlflow/utils/search_utils.py
```diff
@@ -363,11 +366,20 @@ def _validate_order_by_and_generate_token(cls, order_by):
...
-        elif token_value.lower().endswith(" asc"):
-            token_value = token_value[0:-len(" asc")]
+        tokens = token_value.split()
+        if len(tokens) > 2:
+            raise MlflowException(f"Invalid order_by clause '{order_by}'. Could not be parsed.",
+                                  error_code=INVALID_PARAMETER_VALUE)
+        elif len(tokens) == 2:
+            order_token = tokens[1].lower()
+            if order_token not in cls.VALID_ORDER_BY_TAGS:
+                raise MlflowException(f"Invalid ordering key in order_by clause '{order_by}'.",
```
sueann, line 376: this does cause a behavior change for `search_runs` (so this is technically a breaking change), right? i see that lines 355-362 above check the ordering keywords in some way, so just wanted to confirm.

sueann (commented):
just a few nits + 1 question, otherwise lgtm - thanks!

sueann:
also, i looked at the unit tests a bit more and we should do a couple of things:
* add the new error cases to https://github.com/mlflow/mlflow/blob/230c7d722c9d52522a1b7b4ce76b79b86273c190/tests/utils/test_search_utils.py#L323-L326
* add a similar test case for `parse_order_by_for_search_registered_models` in the same test suite^

mlflow/utils/search_utils.py
ankitmathur-db: Yeah that's true - it is breaking

tests/store/model_registry/test_sqlalchemy_store.py
ankitmathur-db: This is the query that the UI sends when the page first loads - the comment raising Exception is something I missed while debugging. Removed now

ankitmathur-db:
Thanks for the review @sueann - updated the PR based on your comments

sueann (approved):
lgtm!

Issue #3010, opened 2020-06-26
[FR] logging metrics via log_batch using SqlAlchemyStore takes very long


creinders:
## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [ ] Yes. I can contribute this feature independently.
- [x] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

Logging multiple metrics via log_batch takes very long when using SQLAlchemy storage. I did some experiments using a mysql database. For logging 1000 metrics it takes about 9 seconds on average (on a Intel(R) Core(TM) i7-7820X CPU @ 3.60GHz). The sample code is attached at the end.

The reason that the function is very slow is that each metric entry is handled individually. I think this can be speed up by introducing a batch-wise processing of the data. I did some tests and was able to speed up the processing to 2 seconds (factor 4.5) e.g. by initializing the session only once and updating the latest metric at the end.

 With some additional improvements (batch-wise insertion of the data and checking the integrity at the end), I was able to decrease the processing time to 0.13 seconds (factor 69).

## Motivation
- What is the use case for this feature?
Speed up logging multiple metrics via log_batch when using SQLAlchemy storage.
- Why is this use case valuable to support for MLflow users in general?
Makes adding metrics in batch mode much faster.
- Why is this use case valuable to support for your project(s) or organization?
We developed an import to mysql storage and notices that it takes a while. 
- Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient)
A couple of initializations and checks (e.g. session or check if is running) are performed multiple times. 

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
[...]

PR #3024, opened 2020-06-30
Speed up logging of multiple metrics

creinders:

Logging multiple metrics via log_batch takes very long when using SQLAlchemy storage (as described here #3010). 

With this pr, metrics are added batch-wise. In this way, only one session is opened which saved a lot of time. Additionally, storing the latest metrics is performed once at the end, so that only one database write is necessary.

It should be noted that metrics with the same run-id, step, timestamp, and value should be added only once. I assume that this conflict happens not that often. The check has originally been performed when adding the metric. However, it has turned out that this can be speed up by adding all metrics and catching possible integrity errors. If an integrity error appears, the metrics will be added in the slower mode (on after another).

Inserting 1000 new metrics now takes about 0.2 seconds instead of 9 seconds. If some metrics are duplicated, it takes about 2 seconds in the slower mode.

## How is this patch tested?

Covered by unit tests.

Components 
Model Registry

Interface 

Language 

Integrations


creinders:
@ankitmathur-db as proposed here https://github.com/mlflow/mlflow/issues/3010. What do you think?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -457,30 +457,81 @@ def _get_deleted_runs(self):
...
+                    # ToDo: Consider prior checks for null, type, metric name validations, ... etc.
+
+                    # Entries will be added to session. If conflicts are detection during saving,
+                    # metrics are inserted one after another using _get_or_create
+                    if auto_new:
+                        logged_metric = SqlMetric(run_uuid=run_id, key=metric.key, value=value,
+                                                  timestamp=metric.timestamp, step=metric.step,
+                                                  is_nan=is_nan)
+                        self._save_to_db(session, logged_metric)
+                        just_created = True
```
mparkhe, line 496: why was this special logic required?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -457,30 +457,81 @@ def _get_deleted_runs(self):
...
+
+                    else:
+                        logged_metric, just_created = self._get_or_create(
+                            model=SqlMetric, session=session, run_uuid=run_id, key=metric.key,
+                            value=value, timestamp=metric.timestamp, step=metric.step,
+                            is_nan=is_nan)
+
+                    if just_created:
+                        if metric.key not in metrics_per_key:
```
mparkhe, line 505: these conditions sound contradictory. They are prolly required due to shared session. Please add a comment to clarify that

mparkhe: Also please add a comment on why _all_ metrics are being grouped by the metric name -- a bit hard for reader to follow until they see call to `_update_latest_metric_if_necessary`

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -457,30 +457,81 @@ def _get_deleted_runs(self):
...
-            if just_created:
-                self._update_latest_metric_if_necessary(logged_metric, session)
+            try:
+                run = self._get_run(run_uuid=run_id, session=session)
+                self._check_run_is_active(run)
+
+                metrics_per_key = {}
+
+                for metric in metrics:
+
```
mparkhe, line 474: **nit** can you remove these additional blank lines -- makes the function unnecessarily long.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -457,30 +457,81 @@ def _get_deleted_runs(self):
...
+                # Conditionally update the ``latest_metrics`` table if the logged metric  was not
+                # already present in the ``metrics`` table. If the logged metric was already
+                # present, we assume that the ``latest_metrics`` table already accounts for
+                # its presence
+
+                for logged_metric_list in metrics_per_key.values():
+                    self._update_latest_metric_if_necessary(None,
+                                                            logged_metrics=logged_metric_list,
+                                                            session=session)
```
mparkhe, line 518: please put a comment/ToDo that this part is currently not done in batch due to complexity in `_update_latest_metric_if_necessary` step

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -489,6 +540,14 @@ def _compare_metrics(metric_a, metric_b):
             return (metric_a.step, metric_a.timestamp, metric_a.value) > \
                    (metric_b.step, metric_b.timestamp, metric_b.value)
 
+        if logged_metrics is not None:
+            for m in logged_metrics:
+                if logged_metric is None or _compare_metrics(m, logged_metric):
+                    logged_metric = m
+
+        if logged_metric is None:
+            return
```
mparkhe, line 549: `None` metrics should be an error case and should have been handled in the caller function

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -457,30 +457,81 @@ def _get_deleted_runs(self):
...
+                        #  with max/min 64b float value
+                        value = 1.7976931348623157e308 if metric.value > 0 \
+                            else -1.7976931348623157e308
+                    else:
+                        value = metric.value
+
+                    # ToDo: Consider prior checks for null, type, metric name validations, ... etc.
+
+                    # Entries will be added to session. If conflicts are detection during saving,
+                    # metrics are inserted one after another using _get_or_create
```
mparkhe, line 490: I am not sure I understand this comment. Also unclear why `auto_new` is required

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -668,8 +727,7 @@ def log_batch(self, run_id, metrics, params, tags):
         try:
             for param in params:
                 self.log_param(run_id, param)
-            for metric in metrics:
-                self.log_metric(run_id, metric)
+            self.log_metrics(run_id, metrics)
             for tag in tags:
                 self.set_tag(run_id, tag)
```
mparkhe, line 732: this change only batches metrics -- params and tags are still added one at a time.

I am curious -- was there a particular reason you chose not to implement batch insert for params and tags. Would be easier than metrics.

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -1470,9 +1470,11 @@ def _raise_exception_fn(*args, **kwargs):  # pylint: disable=unused-argument
 
         package = "mlflow.store.tracking.sqlalchemy_store.SqlAlchemyStore"
         with mock.patch(package + ".log_metric") as metric_mock, \
+                mock.patch(package + ".log_metrics") as metrics_mock, \
                 mock.patch(package + ".log_param") as param_mock, \
                 mock.patch(package + ".set_tag") as tags_mock:
             metric_mock.side_effect = _raise_exception_fn
+            metrics_mock.side_effect = _raise_exception_fn
```
mparkhe, line 1477: we should also add some additional tests to make sure that batched metric insertion is done correctly. For instance multiple metrics with the same key in a single batch, metrics with non-increasing steps, metrics from the batch **do not overwrite** existing metrics (with different keys or same keys if the existing ones have higher step values or timestamp) and **do overwrite** (existing metrics with same keys and lower timestamps and step values), ... etc.

mparkhe (changes requested):
I added a few comments inlined. Would help to add some new recommended tests to verify functionality has not changed.

Also not sure why `auto_new` is required

mparkhe:
Hi @creinders:  I have added some comments and requested changes to your pull request. Would you be able to address this in the coming week? This would a good change to be included in the next release.

creinders:
> Hi @creinders: I have added some comments and requested changes to your pull request. Would you be able to address this in the coming week? This would a good change to be included in the next release.

Thanks for the comments, I will push an update the next days

mlflow/store/tracking/sqlalchemy_store.py
creinders: The name of the variable was not very clear. I renamed auto_new to batch_mode. In batch mode, all metrics are added to the session and the constraints are checked during saving. This is much faster.

mlflow/store/tracking/sqlalchemy_store.py
creinders: I added a comment in the code for that. Regarding your first comment, I not sure I understand it correctly. In line 504, it will be checked if the metric is new (this might be not the case, if batch mode is disabled. In this case the metric with same step, value, ts, and name is already stored). Line 505: the metrics are store in a dictionary with the name as the key. If the key does not exist, an empty list will be added to the dictionary to collect all metrics.

mlflow/store/tracking/sqlalchemy_store.py
creinders: Done :)

mlflow/store/tracking/sqlalchemy_store.py
creinders: I added a todo

mlflow/store/tracking/sqlalchemy_store.py
creinders: The method will now check if an error should be raised

mlflow/store/tracking/sqlalchemy_store.py
creinders: I improved the comments and added some details.

mlflow/store/tracking/sqlalchemy_store.py
creinders: Yes definitely - this can be done in the same way for params and tags as well. I just started with metrics because it influences the runtime the most. Handling the constraints might be a little bit more tricky for params and tags.

tests/store/tracking/test_sqlalchemy_store.py
creinders: I added a couple of tests.

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -656,6 +656,226 @@ def test_log_null_metric(self):
...
+        assert len(metric_histories) == 4
+
+        metrics = [(m.key, m.value, m.timestamp, m.step) for m in metric_histories]
+        assert set(metrics) == set(metric_tuples)
+
+        run_data = self.store.get_run(run_id).data
+        run_metrics = run_data.metrics
+
+        assert len(run_metrics) == 1
+        assert run_metrics[metric_name] == 40
```
mparkhe, line 685: also assert that timestamp and step values are correct

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -656,6 +656,226 @@ def test_log_null_metric(self):
...
+
+    def test_log_metrics_deactivated_batch_mode(self):
```
mparkhe, line 687: i don't see the deactivation logic. looks like this test is exactly the same as above

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -656,6 +656,226 @@ def test_log_null_metric(self):
...
+
+        last_metric_tuples = [
+            (metric_name_a, 50, 5, 5),
+            (metric_name_b, 7, 8, 8),
+            (metric_name_c, -10, 3, 3),
+        ]
+        last_metrics = [(m.key, m.value, m.timestamp, m.step) for m in run_data._metric_objs]
+        assert set(last_metrics) == set(last_metric_tuples)
+
+    def test_log_metrics_multiple_names_deactivated_batch_mode(self):
```
mparkhe, line 815: same comment as before. i am not sure how this is deactivating batch_mode

mparkhe (commented):
The tests look good. The only thing that I am confused about is why support both `batch_mode` and non-`batch_mode`. Can the code be simplified. Please find more comments around `log_metrics` method.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -506,37 +506,104 @@ def _get_deleted_runs(self):
...
+
+                    # ToDo: Consider prior checks for null, type, metric name validations, ... etc.
+
+                    if batch_mode:
+                        # In batch mode, all metrics are added to the session.
+                        # This is much faster than checking for each metric if a metric with the
+                        # same name, value, ts, and step already exists and adding them otherwise.
+                        # Conflicts will be checked during saving. If conflicts are detection
+                        # during saving, metrics are inserted one after another
+                        # using _get_or_create
```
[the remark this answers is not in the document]
mparkhe: oh! i see ... this is `sqlalchemy.exc.IntegrityError` section

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -506,37 +506,104 @@ def _get_deleted_runs(self):
...
+                        logged_metric, just_created = self._get_or_create(
+                            model=SqlMetric,
+                            session=session,
+                            run_uuid=run_id,
+                            key=metric.key,
+                            value=value,
+                            timestamp=metric.timestamp,
+                            step=metric.step,
+                            is_nan=is_nan,
+                        )
```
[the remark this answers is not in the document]
creinders: Unfortunately, `_get_or_create` is much slower because an additional query operation is performed for every metric (checking if it already exists). The same check is performed however when saving a session (due to the constraint check).

Using only `_get_or_create` takes about 2 seconds for 2000 points while inserting in this way takes only 0.2 seconds (if no conflicts are in the data which I guess is the case most of the time).

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -548,6 +615,20 @@ def _compare_metrics(metric_a, metric_b):
...
+        if logged_metrics is not None:
+            for m in logged_metrics:
+                if logged_metric is None or _compare_metrics(m, logged_metric):
+                    logged_metric = m
+
+        if logged_metric is None:
+            raise MlflowException(
+                "Invalid parameters for _update_latest_metric_if_necessary. "
+                "Please set logged_metrics",
+                INVALID_PARAMETER_VALUE,
```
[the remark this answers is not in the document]
creinders: Actually, the error will never arise due to previous checks in log_metric(s). `_update_latest_metric_if_necessary` is only called when it has at least one metric so that this case never occurs. Should when then switch back to just returning? Or do you have another recommendation?

tests/store/tracking/test_sqlalchemy_store.py
creinders: I added the checks

tests/store/tracking/test_sqlalchemy_store.py
creinders: Batch mode is deactivate because there are two metrics with same value, ts, step. In this way, the integrity check is tested.
In this test, the comment was missing which I added now to the definition of the metrics.

tests/store/tracking/test_sqlalchemy_store.py
creinders: There is a comment in line 848. Please also see the previous comment.

creinders:
Hi @mparkhe, I addressed all points a while ago. What do you think?

nickvazz:
Hi, I was wondering if this is still being worked on. Thank you!

creinders:
Actually I was waiting for a response. The feature was fully implemented.

mparkhe:
I will look at this again shortly.

@creinders : Thank you for your patience. Do you mind if i make some small edits to the pull request (if any required) before merging?

creinders:
> @creinders : Thank you for your patience. Do you mind if i make some small edits to the pull request (if any required) before merging?
Sure you can make changes to the pull request

lostless13:
Any chance of progressing this? We're really struggling with performance in this area.

prakhar19:
Any updates on this? It is a much-needed feature. MLFlow is almost unusable for large batch logging.

WeichenXu123:
@prakhar19 
I merge this PR code with master branch.
You can install the mflow with this PR fix by:
```
pip install git+https://github.com/creinders/mlflow.git@master
```

I will take over this PR to complete it.

jaymegordo:
Does this PR only handle batch logging `metrics` but not `params`? Shouldn't they be handled the same way?

WeichenXu123:
@jaymegordo
Yes. We should both handle metric and params. I will do that. Though metrics are more likely to be the perf bottleneck.

erensahin:
Hi, I was thinking about adding this support and found this one. I can take over this PR and continue the work on it if no-one interests.

I also think anything should be logged in batches, not only metrics and params, also tags etc.

harupy:
Addressed by https://github.com/mlflow/mlflow/pull/5460

PR #3028, opened 2020-06-30
Add example illustrating the use of RAPIDS with MLFlow

drobison00:

Adds an example workflow for utilizing RAPIDS with MLFlow.

## How is this patch tested?

Code elements verified in a conda environment, supplied with the example.

- [ X ] Yes. Give a description of this change to be included in the release notes for MLflow users.

Adds a basic example workflow, in CLI and Jupyter notebook forms, for training and deploying an MLFlow model using RAPIDS.

Components 
- [X] `area/examples`: Example code
Model Registry

Interface 

Language 

Integrations


examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,39 @@
+### Train and Publish Locally With RAPIDS and MLFlow
```
smurching, line 1:
```suggestion
### Train and Deploy Models with RAPIDS and MLflow
```

Seems like model training & subsequent deployment with MLflow are the focus of the example? Although correct me if I'm wrong :)

examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,39 @@
...
+**[RAPIDS](https://rapids.ai/)** is a suite of open source libraries for GPU-accelerated analytics. 
+
+**[RAPIDS cuML](https://github.com/rapidsai/cuml)** matches the scikit-learn API, so it can build on MLFlow's existing support for scikit-learn-like models to support 
+persistence and deployment."
+
```
smurching, line 6: Maybe add a sentence describing what our example workflows do, e.g:

```suggestion
The example workflows below train RAPIDs regression models to predict airline flight delays, using MLflow to log models and deploy them as local REST API endpoints for real-time inference.
```

^assuming that such a description is accurate :)

examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,39 @@
...
+#### Jupyter Notebook Workflow
+[Jupyter Notebook](notebooks/rapids_mlflow.ipynb)
+
+#### CLI Based Workflow
+1. Create a new conda environment.
+    1. `$ conda create -f envs/conda.yaml`
```
smurching, line 12: Shouldn't `mlflow run` create this conda environment for the user? We might be able to remove this step

examples/rapids/mlflow_project/src/rf_test/train.py
```diff
@@ -0,0 +1,122 @@
...
+                      algo=algorithm,
+                      max_evals=2,
+                      trials=trials)
+
+        print("===========")
+        fn = partial(train, fpath=args.fpath, hyperopt=False)
+        final_model = fn(tuple(argmin.values()))
+
+        conda_data = ""
+        if (args.conda_env.startswith("http")):
```
smurching, line 110: Interesting, is the ultimate goal here to allow users to log the current conda environment as the model's conda environment? I wonder if it'd work to simply run `conda env export` in a subprocess to get the current conda env in YAML form (we could prob also make it easier to achieve this via MLflow APIs)

examples/rapids/mlflow_project/src/sample_server_query.sh
```diff
@@ -0,0 +1,2 @@
+curl -X POST -H "Content-Type:application/json; format=pandas-split" --data '{"columns":["Year", "Month", "DayofMonth", "DayofWeek", "CRSDepTime", "CRSArrTime", "UniqueCarrier", "FlightNum", "ActualElapsedTime", "Origin"    , "Dest", "Distance", "Diverted"],"data":[[1987, 10, 1, 4, 1, 556, 0, 190, 247, 202, 162, 1846, 0]]}' http://127.0.0.1:55756/invocations
```
smurching, line 1: Nice, thanks for adding this example input :)

We actually recently added support for logging example input ([see docs](https://mlflow.org/docs/latest/python_api/mlflow.sklearn.html#mlflow.sklearn.log_model)) along with MLflow models in 1.9.0. You can use syntax like the following to log a model with example input:

```
mlflow.sklearn.log_model(model,
    path=path,
    signature=signature,
    input_example=example)
```

And access the example input when loading it back via:

```
model = mlflow.pyfunc.load_model(path)
example_info = model.metadata.saved_input_example_info
example_path = example_info["artifact_path"] # also has fields type -> dataframe, pandas_orient -> split etc
example_df = pd.read_json(example_path, orient = example_info.get("pandas_orient"))
example_prediction = model.predict(example_df)
```

@tomasatdatabricks do you think it'd be worth having such an example / does the above usage pattern look right? We may want to include a section in the MLflow docs explaining how to use the new model schemas/input-examples

examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,39 @@
...
+1. Train the model
```
smurching, line 13:
```suggestion
1. Train the model
  1. In your shell, run:
   shell script
# Download the file
wget -N https://rapidsai-cloud-ml-sample-data.s3-us-west-2.amazonaws.com/airline_small.parquet
# Launch the job
mlflow run mlflow_project -e hyperopt \
         -P conda-env=$PWD/envs/conda.yaml \
         -P fpath=airline_small.parquet
  1. The above command kicks off RAPIDS model training using MLflow's projects feature. MLflow will create the necessary conda environment for model training and pass parameters specified via `-P` to our training script. See https://mlflow.org/docs/latest/projects.html# for more info
```

Maybe we could switch to something like the above for brevity?

smurching: I think it'd also be worth it to try logging results to a local SQLite file (e.g. encourage users to run `MLFLOW_TRACKING_URI=./mlflow-db.sqlite mlflow run ...`). Then we could encourage users to use the model registry when deploying their model (e.g. they can just run `mlflow models serve models:/myRapidsModel/1` to serve their model), as opposed to getting the artifact path via the UI and plugging it into the `mlflow models serve` command - let me know if that makes sense!

examples/rapids/mlflow_project/notebooks/rapids_mlflow.ipynb
```diff
@@ -0,0 +1,310 @@
+{
```
smurching, line 1: The notebook is great! Some questions/comments:
* Why is the output-helper necessary for displaying the output of `mlflow models serve`? I guess IPython won't display server output anywhere otherwise? For simplicity, could we instead encourage the user to run `mlflow models serve` via the CLI in a separate terminal window after they've trained their model in the notebook?
* Similar to the other example, I wonder if we can actually log to a local sqlite file in order to use model registry features (as usage of SQL-backed storage and model registry is the encouraged best-practice), rather than working around it

smurching (commented):
Thanks @drobison00 for adding this example! Left a few comments :)

examples/rapids/mlflow_project/src/rf_test/train.py
drobison00: This is somewhat vestigial in this context, to allow for external environment files to be specified that can be packaged with the saved model.

examples/rapids/mlflow_project/notebooks/rapids_mlflow.ipynb
drobison00: We could direct the user to run the model serve call in a separate terminal; however, I think they already get that workflow by following the CLI instructions. This just provides a more self-contained example, fully self-contained after migrating to the sqlite storage solution, since it lets us look up the artifact uri.

All examples will use sqllite in the next PR update.

examples/rapids/mlflow_project/src/sample_server_query.sh
drobison00: I'll update the notebook version to use this process. I do think it makes sense to leave in the curl example for the CLI workflow though.

examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,47 @@
+### Train and Publish Locally With RAPIDS and MLFlow
+**[RAPIDS](https://rapids.ai/)** is a suite of open source libraries for GPU-accelerated analytics. 
+
+**[RAPIDS cuML](https://github.com/rapidsai/cuml)** matches the scikit-learn API, so it can build on MLFlow's existing support for scikit-learn-like models to support
```
harupy, line 4: @drobison00 Can you replace `MLFlow` with `MLflow`?

examples/rapids/mlflow_project/notebooks/rapids_mlflow.ipynb
```diff
@@ -0,0 +1,301 @@
...
+ "cells": [
```
harupy, line 2: Can we run this notebook on Google Colab?

drobison00: You can run in colab, and install rapids using the csp helpers (https://github.com/rapidsai/rapidsai-csp-utils); however, it does require at least a Pascal level GPU, which may not be satisfied if colab gives you a k80.

datametrician:
Hi @harupy, anything left to do to merge this?

smurching:
@drobison00 @datametrician apologies for the delay, I'll take another pass tonight!

examples/rapids/mlflow_project/src/rf_test/train.py
```diff
@@ -94,8 +94,13 @@ def train(params, fpath, hyperopt=False):
     fn = partial(train, fpath=args.fpath, hyperopt=True)
     experid = 0
 
+    artifact_path = "Airline-Demo"
+    artifact_uri = None
+
+    mlflow.set_tracking_uri(uri='sqlite:////tmp/mlflow-db.sqlite')
     with mlflow.start_run():
-        mlflow.set_tag("mlflow.runName", "RAPIDS-Hyperopt")
+        mlflow.set_tag("mlflow.runName", "(CLI) RAPIDS-Hyperopt")
```
smurching, line 102: To set the run name, you can actually pass the `run_name` argument to `mlflow.start_run` ([see docs](https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.start_run))

smurching: Maybe the name should just be `RAPIDS-Hyperopt`? Since the script isn't necessarily run via the CLI

examples/rapids/mlflow_project/notebooks/rapids_mlflow.ipynb
smurching: @drobison00 got it, I see where you're coming from re striving for a self-contained example. However, IMO it'd still be better to recommend using the CLI to serve models - in our examples, we want to encourage best practices for using RAPIDS and MLflow, and currently, the recommended & well-supported way to serve models is to invoke the `mlflow models serve` CLI via terminal, rather than attempt to run it as a subprocess and process the output via a helper thread etc. It'd also make the notebook shorter, which is always nice to have in an example :)

smurching: Let me know if that makes sense btw, happy to discuss further!

examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,47 @@
...
+        wget -N https://rapidsai-cloud-ml-sample-data.s3-us-west-2.amazonaws.com/airline_small.parquet
+        ```
+1. Set MLflow tracking uri
+    1. ```shell script
+        export MLFLOW_TRACKING_URI=sqlite:////tmp/mlflow-db.sqlite
+       ```
+1. Train the model using a single run.
+    1. ```shell script
+       # Launch the job
+       mlflow run . -b local -e hyperopt \
```
smurching, line 26: We should be able to remove the `-b local` from this `mlflow run` invocation and the one below, as the default backend for project execution is "local"

examples/rapids/mlflow_project/src/rf_test/train.py
drobison00: It just indicates which example the result came from, so people exploring can easily identify whether the run was logged from the CLI example or the Notebook.

datametrician:
> @drobison00 @datametrician apologies for the delay, I'll take another pass tonight!

No worries at all... we're just excited to join this community.

drobison00:
@smurching Are there any remaining changes you'd like to see?

examples/rapids/mlflow_project/src/rf_test/train.py
smurching: Ah got it, thanks for clarifying! Would it suffice to look at the source column in the UI for that? Ideally, the "source" would be what users look at to identify the source of the run (i.e. whether it's the CLI example or notebook)

examples/rapids/mlflow_project/src/rf_test/train.py
```diff
@@ -0,0 +1,117 @@
...
+    trials = Trials()
+    algorithm = tpe.suggest if args.algo == 'tpe' else None
+    fn = partial(train, fpath=args.fpath, hyperopt=True)
+    experid = 0
+
+    artifact_path = "Airline-Demo"
+    artifact_uri = None
+
+    mlflow.set_tracking_uri(uri='sqlite:////tmp/mlflow-db.sqlite')
+    with mlflow.start_run():
```
smurching, line 101:
```suggestion
    with mlflow.start_run(run_name="RAPIDS-Hyperopt"):
```

Could we use the `run_name` argument to `start_run` instead of setting the tag directly?

examples/rapids/mlflow_project/src/rf_test/train.py
```diff
@@ -0,0 +1,117 @@
...
+        mlflow.set_tag("mlflow.runName", "(CLI) RAPIDS-Hyperopt")
```
smurching, line 102:
```suggestion
    with mlflow.start_run():
```

examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,47 @@
...
+#### Jupyter Notebook Workflow
+[Jupyter Notebook](notebooks/rapids_mlflow.ipynb)
+
+#### CLI Based Workflow
+1. Download data
+    1. `cd examples/rapids/mlflow_project`
+        ```shell script
+        # Download the file
+        wget -N https://rapidsai-cloud-ml-sample-data.s3-us-west-2.amazonaws.com/airline_small.parquet
+        ```
```
smurching, line 18: We should include a test for the CLI-based workflow - it should be doable by adding an item to the list of `(directory, args_to_mlflow_run)` tuples in the existing [examples tests](https://github.com/mlflow/mlflow/blob/91dca77a7e850ec25a88dd79736cbab2bfd740a3/tests/examples/test_examples.py#L59). For example, you should be able to add an item like `(os.path.join("rapids", "mlflow_project"), ["-e", "simple"])` and the test will invoke `mlflow run rapids/mlflow_projects -e simple` and verify that it succeeds.

It might be easiest to test this if we update the project so that `fpath` is actually an optional parameter and when unspecified, the training script downloads the data via logic like:

```
import requests
import tempfile

# Data will ultimately be written to the file whose path is stored in fname
_, fname = tempfile.mkstemp()
url = "https://rapidsai-cloud-ml-sample-data.s3-us-west-2.amazonaws.com/airline_small.parquet"
airline_data_response = requests.get(url)
with open(fname , 'wb') as handle:
    handle.write(airline_data_response.content)
```

We can defer this to a follow-up PR, but would be good to have - otherwise, this example may break in the future.

examples/rapids/mlflow_project/MLproject
```diff
@@ -0,0 +1,17 @@
+name: cuML RF test
+
+conda_env: envs/conda.yaml
+
+entry_points:
+    hyperopt:
+        parameters:
+            fpath: {type: str}
```
smurching, line 8:
```suggestion
```
To make fpath an optional parameter as described in [this comment](https://github.com/mlflow/mlflow/pull/3028/files#r469043050), we can just remove it from the MLproject file. Then when a user invokes `mlflow run rapids/mlflow_project -e hyperopt`, we will simply not pass the --fpath parameter to our training script, but if the user invokes `mlflow run rapids/mlflow_project -e hyperopt -P fpath=my-airline-data.parquet`, we'll still pass `--fpath my-airline-data.parquet`

examples/rapids/mlflow_project/README.md
```diff
@@ -0,0 +1,47 @@
...
+    1. ```shell script
+        export MLFLOW_TRACKING_URI=sqlite:////tmp/mlflow-db.sqlite
+       ```
+1. Train the model using a single run.
+    1. ```shell script
+       # Launch the job
+       mlflow run . -e simple\
+                --experiment-name RAPIDS-CLI \
+                -P max_depth=10 -P max_features=0.75 -P n_estimators=500 \
+                -P conda-env=$PWD/envs/conda.yaml \
```
smurching, line 29: We no longer need to pass conda-env here right?
```suggestion
```

Also, I like the idea of having an easy-to-run "simple" entry point - could we add default values for the max_depth, max_features, n_estimators params in the MLproject file, as shown in [the docs](https://mlflow.org/docs/latest/projects.html#mlproject-file), and mention that the user can simply run `mlflow run . -e simple`?

examples/rapids/mlflow_project/src/rf_test/train_simple.py
```diff
@@ -0,0 +1,89 @@
...
+
+    experid = 0
+
+    artifact_path = "Airline-Demo"
+    artifact_uri = None
+    experiment_name = "RAPIDS-CLI"
+    experiment_id = None
+
+    mlflow.set_tracking_uri(uri='sqlite:////tmp/mlflow-db.sqlite')
+    with mlflow.start_run(run_name="(CLI) RAPIDS-MLFlow"):
```
smurching, line 80:
```suggestion
    with mlflow.start_run(run_name=""RAPIDS-MLFlow"):
```

Assuming the reasoning in https://github.com/mlflow/mlflow/pull/3028/files#r469031980 makes sense

smurching (commented):
@drobison00 thanks for the updates, the IPython notebook looks great!

Leaving a few more comments, but this looks good once those are addressed - let me know if you have any questions, esp around testing etc :) I may be able to help push some edits to expedite getting this merged.

Oh, and we should update https://github.com/mlflow/mlflow/blob/master/docs/source/tutorials-and-examples/index.rst to include a link to the new RAPIDS example!

examples/rapids/mlflow_project/README.md
smurching: Ah, just discussed offline with @harupy - looks like our underlying CI provider (Github actions) doesn't support GPUs, so we unfortunately can't actually test this example - feel free to disregard the above comment.

smurching (approved):
Thanks @drobison00 for the hard work on this! Pushed a small README update - let me know if it makes sense/if you have any questions, but otherwise LGTM, let's merge and we can additionally iterate on this as needed :)

In particular, I think it might help ease adoption/usage of this example if we simplify the workflow for users who don't have an NVIDIA GPU with CUDA enabled, e.g. by updating the notebook workflow so that it can easily be run on a GPU-enabled colab instance. But we can definitely make such a change in a follow-up PR!

Would you be able to sign off your commits as described in https://github.com/mlflow/mlflow/pull/3028/checks?check_run_id=1001812724? Then the DCO check will pass and we can merge - thanks again!

smurching:
@drobison00 looks like the commits are still not signed off for whatever reason - if you're able to provide verbal approval via a comment on this PR, I'm happy to mark the commit as signed-off and merge - thanks!

drobison00:
@smurching Yes, I'll sign off on this. For some reason one of the commits doesn't seem to work.

PR #3054, opened 2020-07-06
Fix sklearn deserialization with latest cloudpickle when pickle5 is installed

AnesBenmerzoug:
In the latest version of [cloudpickle](https://github.com/cloudpipe/cloudpickle/blob/v1.5.0/CHANGES.md) a [feature was added](https://github.com/cloudpipe/cloudpickle/pull/370) to optionally use [pickle5](https://pypi.org/project/pickle5/) (a backport of all features and APIs added in the pickle module in Python 3.8.3 ) when installed. This currently works fine for serialization, but when deserializing
an sklearn model the following exception is raised:

```python
ValueError: unsupported pickle protocol: 5
```

This PR adds an optional serialization_format argument to the functions used to load an sklearn model that is set to use 
cloudpickle by default. There shouldn't be any issues with this issues since the default serialization format is already set to cloudpickle.

## How is this patch tested?

Since the exception only happens when pickle5, an optional dependency, is installed, I tested the changes manually using the following reproducer (Make sure to install [pickle5](https://pypi.org/project/pickle5/)):

```python
import os
import tempfile

import mlflow.sklearn
import mlflow.pyfunc
import sklearn.datasets as datasets
import sklearn.neighbors as knn

if __name__ == "__main__":
    iris = datasets.load_iris()
    X = iris.data[:, :2]  # we only take the first two features.
    y = iris.target
    knn_model = knn.KNeighborsClassifier()
    knn_model.fit(X, y)
    with tempfile.TemporaryDirectory() as tmp_dir:
        model_path = os.path.join(tmp_dir, "model")
        mlflow.sklearn.save_model(sk_model=knn_model, path=model_path)
        reloaded_knn_model = mlflow.sklearn.load_model(model_uri=model_path)
        reloaded_knn_pyfunc_model = mlflow.pyfunc.load_model(model_uri=model_path)

```

On the current Master branch it raises this exception:

```python
  File "reproducer.py", line 17, in <module>
    reloaded_knn_model = mlflow.sklearn.load_model(model_uri=model_path)
  File "/home/mlflow/mlflow/sklearn.py", line 336, in load_model
    return _load_model_from_local_file(path=sklearn_model_artifacts_path)
  File "/home/mlflow/mlflow/sklearn.py", line 271, in _load_model_from_local_file
    return pickle.load(f)
ValueError: unsupported pickle protocol: 5
```

Whereas on this branch it runs successfully.

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/sklearn.py
+++ b/mlflow/sklearn.py
@@ -309,7 +309,7 @@ def _load_pyfunc(path):
         # specify the ``data`` field within the pyfunc flavor configuration. For these newer
         # models, the ``path`` parameter of ``load_pyfunc()`` refers to the top-level MLflow
         # Model directory. In this case, we parse the model path from the MLmodel's pyfunc
-        # flavor configuration and attempt to fetch the serialization format from the 
+        # flavor configuration and attempt to fetch the serialization format from the
         # scikit-learn flavor configuration
         pyfunc_flavor_conf = _get_flavor_configuration(
             model_path=path, flavor_name=pyfunc.FLAVOR_NAME)
--- a/mlflow/sklearn.py
+++ b/mlflow/sklearn.py
@@ -9,6 +9,7 @@ Python (native) `pickle <https://scikit-learn.org/stable/modules/model_persisten
     Produced for use by generic pyfunc-based deployment tools and batch inference.
 """
 import os
+import logging
 import pickle
 import yaml
 
@@ -33,6 +34,8 @@ SUPPORTED_SERIALIZATION_FORMATS = [
     SERIALIZATION_FORMAT_PICKLE,
     SERIALIZATION_FORMAT_CLOUDPICKLE
 ]
+
+_logger = logging.getLogger(__name__)
 
 
 def get_default_conda_env(include_cloudpickle=False):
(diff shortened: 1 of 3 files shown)


mlflow/sklearn.py
```diff
@@ -261,23 +261,44 @@ def log_model(sk_model, artifact_path, conda_env=None,
                      input_example=input_example)
 
 
-def _load_model_from_local_file(path):
-    """Load a scikit-learn model saved as an MLflow artifact on the local file system."""
+def _load_model_from_local_file(path, serialization_format):
```
dbczumar, line 264: @AnesBenmerzoug The serialization format used to persist the model is stored in the `mlflow.sklearn` flavor configuration within the MLmodel file. This attribute is added here: https://github.com/mlflow/mlflow/blob/136669748dc0a1aeef92ece3f125fc56c6094e29/mlflow/sklearn.py#L174

Instead of adding a `serialization_format` argument, can we read the value from the flavor configuration and use that format accordingly?

mlflow/sklearn.py
```diff
@@ -261,23 +261,44 @@ def log_model(sk_model, artifact_path, conda_env=None,
...
+                    " of the following supported formats: {supported_formats}.".format(
+                        serialization_format=serialization_format,
+                        supported_formats=SUPPORTED_SERIALIZATION_FORMATS)),
+                error_code=INVALID_PARAMETER_VALUE)
     with open(path, "rb") as f:
-        # Models serialized with Cloudpickle can be deserialized using Pickle; in fact,
-        # Cloudpickle.load() is just a redefinition of pickle.load(). Therefore, we do
-        # not need to check the serialization format of the model before deserializing.
-        return pickle.load(f)
+        # Models serialized with Cloudpickle cannot necessarily be deserialized using Pickle;
```
dbczumar, line 282: Thanks for fixing this! This seems to be the crux of the issue (we should not have assumed that `pickle.load()` would continue to remain compatible with Cloudpickle as the package evolves).

dbczumar (changes requested):
@AnesBenmerzoug Thanks so much for addressing this. I left one comment that I hope will allow us to simplify things.

mlflow/sklearn.py
dbczumar: If the MLmodel file does not contain a serialization format entry (which might be the case in very old versions of MLflow), we can assume that the value is `pickle`.

AnesBenmerzoug: Thanks @dbczumar for the information. I did not know about this.
This should make it easier for the load_model function, but what about the _load_pyfunc ? For what I see, it does not have access to the configuration file

dbczumar: @AnesBenmerzoug That's a really good point! Ideally, it would be nice if the path to the directory containing the MLmodel configuration were passed to `_load_pyfunc`. We can make this happen for all future models by removing the `data` argument here - https://github.com/mlflow/mlflow/blob/master/mlflow/sklearn.py#L169 - and replacing it with a different key name (e.g., `model_path`). 

The MLflow pyfunc loader has a special interpretation of the `data` field in the MLmodel flavor configuration; if it's present, only the path referred to by `data` is passed to `_load_pyfunc`. Otherwise, the path to the model directory is passed to `_load_pyfunc` (see https://github.com/mlflow/mlflow/blob/136669748dc0a1aeef92ece3f125fc56c6094e29/mlflow/pyfunc/__init__.py#L472)

If we make this change and remove `data`, the MLmodel file will become accessible within `mlflow.sklearn._load_pyfunc`. We'll still want to remain backwards compatible with older models that specify the `data` field, so our `mlflow.sklearn._load_pyfunc` function should check to see if `path` refers to an MLflow model directory or to a file and perform the appropriate routine.

Let me know if this makes sense!

AnesBenmerzoug:
@dbczumar Is this what you had in mind?
I have updated the reproducer to also load the pyfunc model.

dbczumar (approved):
@AnesBenmerzoug I pushed a few small tweaks (most significantly, the serialization format  must be read from the sklearn flavor configuration in `_load_pyfunc`; in some rare edge cases, this may not exist, so I added a try/catch). I've also added a test for legacy models to ensure that they're loaded correctly. With these changes, this PR looks great! Will intend to merge when tests pass. Thank you for your contribution, and let me know if you have any questions!

AnesBenmerzoug:
@dbczumar Thanks for your help and for taking care of writings the comments and the test.
If I ever run into a similar case I'll make sure to contribute.

mlflow/sklearn.py
pierreglaser: `cloudpickle` developer here. You are right, when `pickle5` is install ,`cloudpickle`-generated pickle strings cannot be unserialized using `pickle.load`. They **can** be unserialized using either

- `cloudpickle.load`
- `pickle5.load`

Just to clarify: `cloudpickle` did not deviate from the `pickle` API, it simply uses `pickle5` as a drop-in replacement of `pickle` (hence the compatibility with `pickle5.load)`. However, this is a breaking change and deserves at least a message in the `README`.

dbczumar:
Hi @AnesBenmerzoug, since MLflow has joined the Linux Foundation, we now require each PR to be signed using `git commit -s`. Can you follow the instructions here so that the `DCO` tests will pass?: https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst#sign-your-work

mlflow/sklearn.py
AnesBenmerzoug: @pierreglaser you're right, this deserves to be mentioned as it may cause issues for certain people.

AnesBenmerzoug:
@dbczumar Should I just edit my commit messages and force push the changes?

dbczumar:
@AnesBenmerzoug Creating a new, signed commit with `git commit --allow-empty -s` should be sufficient! Let me know if that doesn't work for some reason!

AnesBenmerzoug:
@dbczumar It seems I have to go over the previous commits too. I will sign my previous commits and force push the changes.

AnesBenmerzoug:
@dbczumar I did that, but the DCO check seems to fail because of your commits.

PR #3061, opened 2020-07-07
Add support for boolean flags for arguments

### LINKED ISSUE #2898: [FR] Run Docker with -t argument
**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute

- [ ] Yes. I can contribute this feature independently.
- [X] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

Please enable the -t flag for docker run.

## Motivation
To enable colored output inside a run Docker container I need the `-t` flag. This requires a tty.

Colored output is very helpful to discern status and error messages when training models.

Components 
Model Registry

Interfaces
- [X] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models

Languages 

Integrations

## Details

The corresponding code is here: https://github.com/mlflow/mlflow/blob/6ef63a7f33ebfe3481ea57cacd94a8a32fc7efd8/mlflow/projects/__init__.py#L528

The cmd needs to (optionally) expand from `docker run --rm` to `docker run --rm -t`

Zethson:

Closes #2898 

Added a special docker argument 'tty', which can be passed as all other Docker arguments via `-A tty=true`.
This enables pseudo TTY support -> color support for example.
I did not want to add a parameter somewhere else, since I think this would get confusing quickly and require more code changes. 

## How is this patch tested?

Tested locally with a container that has colored output. 

Enabled pseudo TTY support for Docker based projects via special Docker argument `-A tty=true`.

Components 
- [X] `area/docs`: MLflow documentation pages
Model Registry

Interface 
- [X] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models

Language 

Integrations

### DIFF
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -144,8 +144,8 @@ def _user_args_to_dict(arguments, argument_type='P'):
         # Docker arguments such as `t` don't require a value
         # -> set them to True if specified
         if len(split) == 1:
-                name = split[0]
-                value = True
+            name = split[0]
+            value = True
         elif len(split) == 2:
             name = split[0]
             value = split[1]
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -53,7 +53,8 @@ def cli():
                    "corresponding entry point as command-line arguments in the form `--name value`")
 @click.option("--docker-args", "-A", metavar="NAME=VALUE", multiple=True,
               help="A `docker run` flag or argument, of the form -A name=value. Where `name` "
-              "will then be propagated as `docker run --name value`.")
+              "will then be propagated as `docker run --name value`. "
+              "Pass -A tty=true to allocate a pseudo-TTY.")
 @click.option("--experiment-name", envvar=tracking._EXPERIMENT_NAME_ENV_VAR,
               help="Name of the experiment under which to launch the run. If not "
                    "specified, 'experiment-id' option will be used to launch run.")
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -54,7 +54,7 @@ def cli():
 @click.option("--docker-args", "-A", metavar="NAME=VALUE", multiple=True,
               help="A `docker run` flag or argument, of the form -A name=value. Where `name` "
               "will then be propagated as `docker run --name value`. "
-              "Pass -A tty=true to allocate a pseudo-TTY.")
+              "Pass -A t=true to allocate a pseudo-TTY.")
 @click.option("--experiment-name", envvar=tracking._EXPERIMENT_NAME_ENV_VAR,
               help="Name of the experiment under which to launch the run. If not "
                    "specified, 'experiment-id' option will be used to launch run.")
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -53,8 +53,8 @@ def cli():
                    "corresponding entry point as command-line arguments in the form `--name value`")
 @click.option("--docker-args", "-A", metavar="NAME=VALUE", multiple=True,
               help="A `docker run` argument or flag, of the form -A name=value (e.g. -A gpus=all) "
-                   "or -A name (e.g. -A t). The argument will then be passed as `docker run --name value` "
-                   "or `docker run -name` respectively. ")
+                   "or -A name (e.g. -A t). The argument will then be passed as "
+                   "`docker run --name value` or `docker run -name` respectively. ")
 @click.option("--experiment-name", envvar=tracking._EXPERIMENT_NAME_ENV_VAR,
               help="Name of the experiment under which to launch the run. If not "
                    "specified, 'experiment-id' option will be used to launch run.")
--- a/mlflow/cli.py
+++ b/mlflow/cli.py
@@ -54,7 +54,7 @@ def cli():
 @click.option("--docker-args", "-A", metavar="NAME=VALUE", multiple=True,
               help="A `docker run` flag or argument, of the form -A name=value. Where `name` "
               "will then be propagated as `docker run --name value`. "
-              "Pass -A t=true to allocate a pseudo-TTY.")
+              "Pass -A t to allocate a pseudo-TTY.")
 @click.option("--experiment-name", envvar=tracking._EXPERIMENT_NAME_ENV_VAR,
               help="Name of the experiment under which to launch the run. If not "
                    "specified, 'experiment-id' option will be used to launch run.")
@@ -104,7 +104,7 @@ def run(uri, entry_point, version, param_list, docker_args, experiment_name, exp
         sys.exit(1)
 
     param_dict = _user_args_to_dict(param_list)
-    args_dict = _user_args_to_dict(docker_args, flag_name='A')
+    args_dict = _user_args_to_dict(docker_args, argument_type='A')
 
     if backend_config is not None and os.path.splitext(backend_config)[-1] != ".json":
         try:
@@ -137,16 +137,23 @@ def run(uri, entry_point, version, param_list, docker_args, experiment_name, exp
         sys.exit(1)
 
 
-def _user_args_to_dict(user_list, flag_name='P'):
+def _user_args_to_dict(arguments, argument_type='P'):
     user_dict = {}
-    for s in user_list:
-        index = s.find("=")
-        if index == -1:
+    for arg in arguments:
+        split = arg.split('=')
+        # Docker arguments such as `t` don't require a value
+        # -> set them to True if specified
+        if len(split) == 1:
+            if argument_type == 'A':
+                name = split[0]
+                value = True
+        elif len(split) == 2:
+            name = split[0]
+            value = split[1]
+        else:
             eprint("Invalid format for -%s parameter: '%s'. "
-                   "Use -%s name=value." % (flag_name, s, flag_name))
+            "Use -%s name=value." % (argument_type, arg, argument_type))        
             sys.exit(1)
-        name = s[:index]
-        value = s[index + 1:]
         if name in user_dict:
             eprint("Repeated parameter: '%s'" % name)
             sys.exit(1)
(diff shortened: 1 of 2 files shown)


Zethson:
Requesting review from @dbczumar .

Zethson:
Kindly requesting review from @dbczumar.

mlflow/cli.py
```diff
@@ -53,7 +53,8 @@ def cli():
                    "corresponding entry point as command-line arguments in the form `--name value`")
 @click.option("--docker-args", "-A", metavar="NAME=VALUE", multiple=True,
               help="A `docker run` flag or argument, of the form -A name=value. Where `name` "
-              "will then be propagated as `docker run --name value`.")
+              "will then be propagated as `docker run --name value`. "
+              "Pass -A tty=true to allocate a pseudo-TTY.")
```
dbczumar, line 57: @Zethson Instead of renaming / aliasing the `-t` argument of `docker run` as `-tty`, can we continue to use `-t`? This way, we don't have to do the special casing in `mlflow.projects.__init__`. This will likely be clearer to users who are familiar with the `-t` flag in Docker.

Zethson: Sure. Will not look that descriptive: `-A t=true`, but fine with me if you prefer it.

dbczumar: Awesome! Thanks!

mlflow/cli.py
```diff
@@ -53,7 +53,8 @@ def cli():
                    "corresponding entry point as command-line arguments in the form `--name value`")
 @click.option("--docker-args", "-A", metavar="NAME=VALUE", multiple=True,
               help="A `docker run` flag or argument, of the form -A name=value. Where `name` "
-              "will then be propagated as `docker run --name value`.")
+              "will then be propagated as `docker run --name value`. "
+              "Pass -A t=true to allocate a pseudo-TTY.")
```
dbczumar, line 57: @Zethson Ah, I see the difficulty now. I was hoping to allow users to simply pass the `-t` flag as `--docker-args t`, but it looks like our `NAME=VALUE` metavar structure prevents that from happening. Rather than special casing `t=true` or `tty=true`, perhaps we can revise `--docker-args` to take either a pair of `NAME=VALUE` or a boolean flag of the form `NAME`. That way, users could enable a tty using one of the following:

```
mlflow run --docker-args tty
mlflow run --docker-args t
```

This would automatically provide support for all other boolean flags as well. Does that approach make sense?

Zethson: Yes, of course. This is also how I would have done it initially, but I didn't quite know how many things I am allowed to touch :)
Will implement tomorrow.

mlflow/cli.py
```diff
@@ -136,16 +137,23 @@ def run(uri, entry_point, version, param_list, docker_args, experiment_name, exp
...
-        index = s.find("=")
-        if index == -1:
+    for arg in arguments:
+        split = arg.split('=')
+        # Docker arguments such as `t` don't require a value
+        # -> set them to True if specified
+        if len(split) == 1:
+            if argument_type == 'A':
+                name = split[0]
+                value = True
```
Zethson, line 149: Fine with you if I work with booleans here or do you want me to use 'true'?

dbczumar: @Zethson How do we intend to map True (or `true`) to the proper command flag (which is just `-t` rather than `-t True` or `-t true`)?

dbczumar: Perhaps the user args to do not necessarily have to be a dictionary, since certain args do not have value elements. We may consider changing the return type of this to a list of `UserArgument` objects with two properties: `name` (required) and `value` (optional).

Zethson: Not sure whether I understand the first question.
If only a single parameter is specified without `=value`, it is automatically set to the value True. This only happens when the parameter is passed. All command line interfaces treat options like this. Note that this behavior is only enabled for Docker arguments and not for mlflow -P parameters. 

About the second question: If mlflow were at Python 3.7 we could use a pretty [data class](https://docs.python.org/3/library/dataclasses.html). However, since this is not the case I guess it would not make the code much prettier. 

>  since certain args do not have value elements. 

Well, yes and no. They don't have an associated value when passed by the user, but they automatically get the value True inside mlflow. At runtime every argument has a name AND a value.

I am fine with the current design, but if you have more specific wishes (and reasoning) I am of course happy to adjust the implementation.

dbczumar: > If only a single parameter is specified without `=value`, it is automatically set to the value True. 

I think the issue is that this behavior is incompatible with downstream CLI tools like Docker. For example, if `_user_args_to_dict` returns a dictionary with a string -> boolean mapping which is then passed to `_get_docker_command`, the resulting Docker command will contain an argument of the form `--name True`, which is not the correct behavior for boolean flags on the commandline.

`docker run --interactive True ...` is *invalid*
`docker run --interactive ...` is valid.

I think defining `user_args_to_dict` like this is fine, but the downstream consumer of this dictionary (i.e., `_get_docker_command`) needs to handle boolean values in the dictionary. Specifically, the string `True` or `False` should not be passed to the `docker run` command. Let me know if this makes sense.

mlflow/projects/__init__.py
```diff
@@ -534,6 +534,13 @@ def _get_docker_command(image, active_run, docker_args=None, volumes=None, user_
     cmd = [docker_path, "run", "--rm"]
 
     if docker_args:
+        # Allocate a pseudo-TTY if specified
+        if 't' in docker_args:
```
dbczumar, line 538: Once the changes described in https://github.com/mlflow/mlflow/pull/3061/files#r451860120 have been made, I don't think we'll need this special casing logic for `t`. We can probably remove this code.

mlflow/cli.py
Zethson: Yeah, I'm aware of that. Hence, the manual setting of the -t in the command.

The only thing that makes this more complicated is that all the options that may contain boolean values need to be passed BEFORE the rest of the a=b arguments.

Hence, I suggest that I write some code that filters all pairs of name:booleanvalue -> set them first by appending `-name` to the docker run command -> removing them from the dict (or use a prefiltered one) and then append the rest as is business now.
Since the dict is always super small there is no performance penalty anyways.

Does this sound fine to you?

Zethson: This would of course also allow for all boolean options (not just -t like now).
Though I fail to see why you would ever pass -i  :)

dbczumar: Hmmm...I'm probably missing something, but I don't understand why boolean arguments need to come first. Can you clarify?

Zethson: Huh, actually I thought so, but this doesn't seem to be the case.
Well yeah, then things get easier of course.

Zethson: I need to sleep, but will get back to you with a new, certainly cleaner looking draft tomorrow.

dbczumar: Awesome!

mlflow/projects/__init__.py
```diff
@@ -534,8 +534,13 @@ def _get_docker_command(image, active_run, docker_args=None, volumes=None, user_
     cmd = [docker_path, "run", "--rm"]
 
     if docker_args:
-        for key, value in docker_args.items():
-            cmd += ['--' + key, value]
+        for name, value in docker_args.items():
+            # Passed just the name as boolean flag
+            if isinstance(value, bool) and value:
+                cmd += ['-' + name]
```
dbczumar, line 540: I think we need to use a double hyphen for long options (`name` >= 2 chars) and a single hyphen for short options (`name` = 1 char)

Zethson:
@dbczumar `-A t and -A tty` now both work.

mlflow/cli.py
```diff
@@ -52,8 +52,9 @@ def cli():
                    "are not in the list of parameters for an entry point will be passed to the "
                    "corresponding entry point as command-line arguments in the form `--name value`")
 @click.option("--docker-args", "-A", metavar="NAME=VALUE", multiple=True,
-              help="A `docker run` flag or argument, of the form -A name=value. Where `name` "
-              "will then be propagated as `docker run --name value`.")
+              help="A `docker run` argument or flag, of the form -A name=value (e.g. -A gpus=all) "
+                   "or -A name (e.g. -A t). The argument will then be passed as "
+                   "`docker run --name value` or `docker run -name` respectively. ")
```
dbczumar, line 57: Nit:
```suggestion
                   "`docker run --name value` or `docker run --name` respectively. ")
```

mlflow/projects/__init__.py
```diff
@@ -534,8 +534,16 @@ def _get_docker_command(image, active_run, docker_args=None, volumes=None, user_
     cmd = [docker_path, "run", "--rm"]
 
     if docker_args:
-        for key, value in docker_args.items():
-            cmd += ['--' + key, value]
+        for name, value in docker_args.items():
+            # Passed just the name as boolean flag
+            if isinstance(value, bool) and value:
+                if len(name) == 1:
+                    cmd += ['-' + name]
```
dbczumar, line 541: We should do this length check when `value` is not a bool as well. For example, docker's `-u/--user` argument requires a string value; the current implementation would produce `--u value` as opposed to `-u value`, causing an error.

mlflow/cli.py
```diff
@@ -136,16 +137,22 @@ def run(uri, entry_point, version, param_list, docker_args, experiment_name, exp
...
+def _user_args_to_dict(arguments, argument_type='P'):
     user_dict = {}
-    for s in user_list:
-        index = s.find("=")
-        if index == -1:
+    for arg in arguments:
+        split = arg.split('=')
+        # Docker arguments such as `t` don't require a value
+        # -> set them to True if specified
+        if len(split) == 1:
```
dbczumar, line 146: can we re-introduce the check against `argument_type` so that valueless inputs are only acceptable for `argument_type=A`? We don't currently support valueless inputs for project parameters - if we wanted to do so, it would be best to create a separate PR that also updates the documentation for project parameters.

dbczumar (changes requested):

dbczumar:
@Zethson This is looking great - really close. Excited to get this merged once comments are addressed. Thanks for your contribution!

dbczumar (approved):
LGTM! Thanks @Zethson !

PR #3068, opened 2020-07-08
Fix bug with renaming registered model with sqlite as backend store

zhidongqu-db:

Currently, when we use sqlite as backend store, renaming registered model does not cascade update its model version names and therefore cause errors. Solution is to explicitly enforce foreign key constrains on rename operation when using sqlite 

## How is this patch tested?

Unit test

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -169,6 +169,7 @@ class SqlAlchemyStore(AbstractStore):
         :return: A single updated :py:class:`mlflow.entities.model_registry.RegisteredModel` object.
         """
         with self.ManagedSessionMaker() as session:
+            session.execute("PRAGMA foreign_keys = ON;")
             sql_registered_model = self._get_registered_model(session, name)
             try:
                 updated_time = now()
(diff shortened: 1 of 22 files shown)


mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -169,6 +169,8 @@ def rename_registered_model(self, name, new_name):
         :return: A single updated :py:class:`mlflow.entities.model_registry.RegisteredModel` object.
         """
         with self.ManagedSessionMaker() as session:
+            if self.db_type == SQLITE:
+                session.execute("PRAGMA foreign_keys = ON;")
```
dbczumar, line 173: @zhidongqu-db I'm wondering if it makes sense to set `PRAGMA foreign_keys = ON;` in `self.ManagedSessionMaker()` for SQLITE databases. We already set this pragma for search operations in the sqlalchemy store for the model registry and for tracking, which provides a nice proof-of-concept (afaik, we have not encountered issues due to setting this pragma). Otherwise, I think we're less likely to miss `PRAGMA foreign_keys = ON` for a subsequent API, which could lead to broken behavior like it did here. Thoughts?

zhidongqu-db: @dbczumar I would agree as in other DBs like mysql, postgres, the foreign key constraint is enabled by default, it make sense to enable this for sqlite as well. Double checked that PRAGMA foreign_keys = ON; does not seems to do anything strange other than enabling the constraint. I will try to make this change and see if it can pass all the unit tests we have.

zhidongqu-db: @dbczumar I just double checked, looks like we are using `PRAGMA case_sensitive_like = true;` in the searches of runs and models. `rename_registered_model` is the only place that needs `PRAGMA foreign_keys = ON;` currently, however, I still it is good to move this logic to `__init__` and apply to all sqlalchemy sessions. Please let me know what you think!

PR #3072, opened 2020-07-08
Add set_registry_uri, get_registry_uri

sueann:

Similarly to `set_tracking_uri` and `get_tracking_uri`, provide a way to set the registry URI for fluent APIs.

Note that we are not adding support for this in R here (and do not plan to unless there is community feedback).

## How is this patch tested?

New unit tests added

Now you can set the model registry URI for fluent APIs via `mlflow.set_registry_uri`, if you want it to to be different from the tracking URI. You can also retrieve the current value of the registry URI via `mlflow.get_registry_uri`.

Components 
- [X] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/_model_registry/registry.py
+++ b/mlflow/tracking/_model_registry/registry.py
@@ -29,7 +29,7 @@ class ModelRegistryStoreRegistry(StoreRegistry):
         :return: An instance of `mlflow.store.model_registry.AbstractStore` that fulfills the
                  store URI requirements.
         """
-        from mlflow.tracking._tracking_service import utils
-        store_uri = store_uri if store_uri is not None else utils.get_tracking_uri()
+        from mlflow.tracking._model_registry import utils
+        store_uri = utils._resolve_registry_uri(store_uri)
         builder = self.get_store_builder(store_uri)
         return builder(store_uri=store_uri)
--- a/mlflow/tracking/_model_registry/utils.py
+++ b/mlflow/tracking/_model_registry/utils.py
@@ -4,7 +4,8 @@ from mlflow.store.db.db_types import DATABASE_ENGINES
 from mlflow.store.model_registry.rest_store import RestStore
 from mlflow.tracking._model_registry.registry import ModelRegistryStoreRegistry
 from mlflow.tracking._tracking_service.utils import _TRACKING_USERNAME_ENV_VAR, \
-    _TRACKING_PASSWORD_ENV_VAR, _TRACKING_TOKEN_ENV_VAR, _TRACKING_INSECURE_TLS_ENV_VAR
+    _TRACKING_PASSWORD_ENV_VAR, _TRACKING_TOKEN_ENV_VAR, _TRACKING_INSECURE_TLS_ENV_VAR, \
+    _resolve_tracking_uri, get_tracking_uri
 from mlflow.utils import rest_utils
 from mlflow.utils.databricks_utils import get_databricks_host_creds
 from mlflow.utils.uri import get_db_profile_from_uri
@@ -46,14 +47,23 @@ def set_registry_uri(uri):
     _registry_uri = uri
 
 
+def _get_registry_uri_from_context():
+    global _registry_uri
+    # in the future, REGISTRY_URI env var support can go here
+    return _registry_uri
+
+
 def get_registry_uri():
     """
-    Get the current registry URI specified by `set_registry_uri`.
+    Get the current registry URI. If none has been specified, defaults to the tracking URI.
 
     :return: The registry URI.
     """
-    global _registry_uri
-    return _registry_uri
+    return _get_registry_uri_from_context() or get_tracking_uri()
+
+
+def _resolve_registry_uri(registry_uri=None, tracking_uri=None):
+    return registry_uri or _get_registry_uri_from_context() or _resolve_tracking_uri(tracking_uri)
 
 
 def _get_sqlalchemy_store(store_uri):
--- a/mlflow/tracking/_model_registry/utils.py
+++ b/mlflow/tracking/_model_registry/utils.py
@@ -31,7 +31,8 @@ _registry_uri = None
 
 def set_registry_uri(uri):
     """
-    Set the registry server URI.
+    Set the registry server URI. This method is especially useful if you have a registry server
+    that's different from the tracking server.
 
     :param uri:
 
--- a/mlflow/tracking/_model_registry/utils.py
+++ b/mlflow/tracking/_model_registry/utils.py
@@ -12,11 +12,11 @@ from mlflow.utils.uri import get_db_profile_from_uri
 
 # NOTE: in contrast to tracking, we do not support the following ways to specify
 # the model registry URI:
-#  - via a utility method like tracking.tracking.utils.set_tracking_uri
 #  - via environment variables like MLFLOW_TRACKING_URI, MLFLOW_TRACKING_USERNAME, ...
 # We do support specifying it
 #  - via the ``model_registry_uri`` parameter when creating an ``MlflowClient`` or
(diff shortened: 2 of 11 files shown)


mlflow/tracking/_model_registry/utils.py
```diff
@@ -25,7 +25,36 @@
...
+
+
+def get_registry_uri():
+    """
+    Get the current registry URI.
+
+    :return: The tracking URI.
+    """
+    global _registry_uri
+    return _registry_uri
```
sueann, line 56: One concern I have is that `get_registry_uri` is not semantically equivalent to `get_tracking_uri` which has logic to return default values if there isn't a tracking uri specified through `set_tracking_uri`. For use in `MlflowClient` this is important, because we want to set the registry URI to be:
```
(registry_uri parameter value) or get_registry_uri() or (tracking_uri parameter value) or get_tracking_uri()
``` 
. However, if `get_registry_uri` were to be used in another context as a stand-alone method to get the _final_ registry uri, you'd have to apply the logic^ yourself. If you are familiar with `get_tracking_uri`, you would expect it to do it for you.

One option is to make this function:
```
def get_registry_uri(tracking_uri=None):
    """
    Get the current registry URI.
    :return: The tracking URI.
    """
    global _registry_uri
    return _registry_uri or tracking_uri or get_tracking_uri()
```
though that seems a bit convoluted and replicates the logic for tracking_uri setting which will need to be maintained in sync with other places where we set the tracking_uri (though may be unlikely to change).

mlflow/tracking/_model_registry/utils.py
```diff
@@ -12,11 +12,11 @@
 
 # NOTE: in contrast to tracking, we do not support the following ways to specify
 # the model registry URI:
-#  - via a utility method like tracking.tracking.utils.set_tracking_uri
 #  - via environment variables like MLFLOW_TRACKING_URI, MLFLOW_TRACKING_USERNAME, ...
```
mparkhe, line 15: is this comment about `MLFLOW_TRACKING_URI` still accurate?

mlflow/tracking/_model_registry/utils.py
mparkhe: can we have lower level utility functions that all callers call for tracking and registry URI resolution. something like...

```
def resolve_tracking_uri(tracking_uri=None):
  return tracking_uri or _tracking_uri

def resolve_registry_uri(registry_uri=None, tracking_uri=None):
  return registry_uri or _registry_uri or resolve_tracking_uri(tracking_uri)
```

So for instances, client can call the utility method with provided parameters and utility methods can resolve the ordering.

sueann: I think there are several uses of `get_tracking_uri` and `get_registry_uri`:
1. For the fluent API user, `get_*_uri` should tell you which tracking or registry server we're currently using.
2. In projects - not relevant for `get_registry_uri`
3. In `MlflowClient`, to resolve the user-provided params with the values set previously.

I think `get_registry_uri` should return considering the fallback options in order to satisfy 1. For 3, it doesn't seem like a common use case, though if someone wants to implement their own version of `MlflowClient` for some reason it'd be useful to have methods to support it. So I propose:

`mlflow/tracking/_model_registry/utils.py`
```
def _get_registry_uri_from_context():
    global _registry_uri
    # in the future, REGISTRY_URI env var support can go here
    return _registry_uri

def get_registry_uri():
    """
    Get the current registry URI specified by `set_registry_uri`.
    :return: The registry URI.
    """
    tracking_uri = get_tracking_uri()
    return _get_registry_uri_from_context() or (tracking_uri if not is_filestore_uri(tracking_uri) else None)    

def _resolve_registry_uri(registry_uri=None, tracking_uri=None):
    return registry_uri or _get_registry_uri_from_context() or _resolve_tracking_uri(tracking_uri)
```

`mlflow/tracking/_tracking_service/utils.py`
```
def _resolve_tracking_uri(tracking_uri=None):
    return tracking_uri or get_tracking_uri()
```

And in `MlflowClient`:
```
def __init__(self, tracking_uri=None, registry_uri=None):
        final_tracking_uri = utils._resolve_tracking_uri(tracking_uri)
        self._registry_uri = registry_utils._resolve_registry_uri(registry_uri, tracking_uri)
```

I kind of want to also refactor the util files so all tracking and registry URI methods are in `tracking/server_uri_utils.py` now that there's a dependency between the two, but it's technically a breaking change so I don't need to do it... or do I :-)

sueann: cc @aarondav

aarondav: Okkie dokkie! Makes sense to be more careful with the publicly-exposed method. Should still update the docstring :)

mlflow/tracking/_model_registry/utils.py
sueann: yeah. maybe it would be less confusing if it said `MLFLOW_REGISTRY_URI` ?

mlflow/tracking/_model_registry/registry.py
```diff
@@ -29,7 +29,7 @@ def get_store(self, store_uri=None):
         :return: An instance of `mlflow.store.model_registry.AbstractStore` that fulfills the
                  store URI requirements.
         """
-        from mlflow.tracking._tracking_service import utils
-        store_uri = store_uri if store_uri is not None else utils.get_tracking_uri()
+        from mlflow.tracking._model_registry import utils
+        store_uri = utils._resolve_registry_uri(store_uri)
```
sueann, line 33: this does change the behavior such that `store_uri = ""` is not a legit URI anymore. that seems okay - could it maybe affect a plug-in that is defined with `uri = ""`? maybe...

mlflow/tracking/_tracking_service/registry.py
```diff
@@ -32,6 +32,6 @@ def get_store(self, store_uri=None, artifact_uri=None):
                  requirements.
         """
         from mlflow.tracking._tracking_service import utils
-        store_uri = store_uri if store_uri is not None else utils.get_tracking_uri()
+        store_uri = utils._resolve_tracking_uri(store_uri)
```
sueann, line 35: same comment as for `mlflow/tracking/_model_registry/registry.py` on `store_uri = ""`

tests/tracking/_model_registry/test_utils.py
```diff
@@ -19,14 +19,39 @@
...
+        assert get_registry_uri() is None
+
+
+def test_default_get_registry_uri_with_tracking_uri_set():
+    tracking_uri = "databricks://tracking_werohoz"
+    with mock.patch("mlflow.tracking._model_registry.utils.get_tracking_uri") \
+            as get_tracking_uri_mock:
+        get_tracking_uri_mock.return_value = tracking_uri
+        set_registry_uri(None)
+        assert get_registry_uri() == tracking_uri
```
sueann, line 55: did not add tests for `_resolve_registry_uri` as it's fairly well-covered by the client tests

mlflow/tracking/_model_registry/utils.py
```diff
@@ -25,7 +26,45 @@
 # This means the following combinations are not supported:
 #  - Tracking RestStore & Model Registry RestStore that use different credentials.
 
-# NOTE: SqlAlchemyStore code is commented out here - we can add it back once it's available
+_registry_uri = None
+
+
+def set_registry_uri(uri):
+    """
+    Set the registry server URI.
```
aarondav, line 34:
```suggestion
    Set the registry server URI. This method is especially useful if have a registry server 
    that's different from the tracking server.
```

aarondav (approved):

PR #3082, opened 2020-07-09
[Multi-Workspace Registry] allow specifying hostcreds with Databricks utility

ankit-db:

This PR will allow users to specify a remote registry by setting a `tracking_uri` or a `registry_uri` of the form `databricks://scope/prefix`. This PR will modify `get_databricks_host_creds` so that if a profile by the name of `scope` is not found in the Databricks CLI configuration, we will use host credentials stored in the Databricks Secrets service under the scope `scope` and with the key `prefix + "host"` and `prefix + "token"`. For example, if I specified `databricks://prod-registry/ankit` then secrets for the host and token would be fetched from scope: `prod-registry` with keys `ankithost` and `ankittoken`.

Edge cases to consider:

- What happens if a user just doesn't want to set a prefix?

## How is this patch tested?

See unit tests and DM about how to test

Components 
Model Registry

Interface 

Language 

Integrations


mlflow/projects/__init__.py
```diff
@@ -781,7 +781,8 @@ def _get_databricks_env_vars(tracking_uri):
         return {}
 
     db_profile = get_db_profile_from_uri(tracking_uri)
-    config = databricks_utils.get_databricks_host_creds(db_profile)
+    db_path_info = get_db_path_info_from_uri(tracking_uri)
```
ankitmathur-db, line 784: I changed all the locations where we get the host creds by default, but I'm not sure that we actually need this everywhere. We definitely need it in the tracking and registry utils to fetch artifacts and to do all other registry operations, but I honestly don't have enough context on projects to know if it's right to put it here

mlflow/utils/databricks_utils.py
```diff
@@ -162,6 +163,20 @@ def get_databricks_host_creds(profile=None):
...
     else:
         config = provider.get_config()
+    # if a path is specified, that implies a Databricks tracking URI of the form:
+    # databricks://profile-name/path-specifier
+    if (not config or not config.host) and path:
+        dbutils = _get_dbutils()
+        # Prefix differentiates users and is provided as path information in the URI
+        key_prefix = path
+        if dbutils:
+            host = dbutils.secrets.get(scope=profile, key=key_prefix + "host")
```
ankitmathur-db, line 173: Should we delimit the prefix with a `_` or `-` here?

mlflow/projects/__init__.py
sueann: hm yeah. i think it'd make sense to support it here. how about if we made `get_databricks_host_creds ` take in the URI instead, since all callsites essentially have and use the URI anyway? In `mlflow/projects/databricks.py` there are a few places where we store `databricks_profile` and use it to call `get_databricks_host_creds` but it doesn't seem like we use it for anything else, so if we stored the URI itself instead of the parsed profile name I don't think there would be a problem.

cc @smurching - in projects, `DatabricksSubmittedRun` and `DatabricksJobRunner` use `databrick_profile` to get the hostcreds. if we change the `DatabricksJobRunner.databricks_profile` field from a string to the tracking URI itself (and possibly rename it), do you think there would be a problem? those objects seem to be used just within `mlflow/projects/databricks.py` (but i don't know if people have plug-ins or anything)

mlflow/utils/databricks_utils.py
sueann: I thought about this very briefly and had just let it be... but it seems more natural to have a delimiter -- let's do `-`

mlflow/utils/databricks_utils.py
```diff
@@ -162,6 +163,20 @@ def get_databricks_host_creds(profile=None):
         config = provider.ProfileConfigProvider(profile).get_config()
     else:
         config = provider.get_config()
+    # if a path is specified, that implies a Databricks tracking URI of the form:
+    # databricks://profile-name/path-specifier
+    if (not config or not config.host) and path:
```
sueann, line 168: what if path is ""? should we support that?

mlflow/utils/databricks_utils.py
```diff
@@ -162,6 +163,20 @@ def get_databricks_host_creds(profile=None):
...
+        dbutils = _get_dbutils()
+        # Prefix differentiates users and is provided as path information in the URI
+        key_prefix = path
```
sueann, line 171: nit: this can go inside `if dbutils:` right?

tests/utils/test_uri.py
```diff
@@ -27,6 +27,12 @@ def test_get_db_profile_from_uri_casing():
     assert get_db_profile_from_uri('databricks://aAbB') == 'aAbB'
 
 
+def test_get_db_path_info_from_profile():
+    assert get_db_path_info_from_uri('databricks://profile/prefix') == 'prefix'
+    assert get_db_path_info_from_uri('nondatabricks://profile/prefix') is None
+    assert get_db_path_info_from_uri('databricks://profile') is None
```
sueann, line 33: try `'databricks://profile/'`, `'databricks://'`, `'databricks:/profile/prefix'`, `'databricks'`  ?

sueann (commented):
looks pretty good to me! left some comments inline ^

mlflow/utils/databricks_utils.py
ankitmathur-db: Yeah, good point! We currently don't support that, due to the path truthiness stuff. 

I think the problem with supporting it is that an incorrect profile passed in would throw an error related to the lack of a path. So, for example, if you don't have a configured profile `databricks://profile1/`, we currently throw and say that you misconfigured your URI and you should change it. If we supported "", we would attempt to fetch secrets from the Secrets service and throw on the lack of that scope being defined or on the lack of presence of a "host" or "token" key and that case would be difficult to distinguish from scenarios where people weren't even trying to use this feature or multi-workspace registry at all (for example, just in tracking)

mlflow/utils/databricks_utils.py
ankitmathur-db: Agreed - will update that

mlflow/utils/databricks_utils.py
ankitmathur-db: Yup - fixed

tests/utils/test_uri.py
ankitmathur-db: Changed the test

One point of clarification: I'm pretty sure we currently don't support `databricks:/profile` because in `get_db_profile_from_uri`, if there aren't 2 slashes, then the netloc is empty, and that's the profile that we're supposed to read in.

mlflow/utils/databricks_utils.py
sueann: ok so we just need to document that the secret key prefix cannot be empty

tests/utils/test_uri.py
sueann: mm yeah.. i guess it'd be better to test that we throw something reasonable for that in `get_databricks_host_creds` (if we change it to take in the URI as the param)

mlflow/projects/__init__.py
ankitmathur-db: Yeah that would make sense @sueann ! Will wait until @smurching weighs in before I do the refactor

smurching: They're internal classes, seems fine to refactor :)

ankitmathur-db:
@sueann updated with the refactor - also got rid of the redundant URI functions and simplified some other logic as well.

I changed the interface of DatabricksJobRunner to take in a databricks_profile_uri, and I also made a utility in the uri file that constructs Databricks URI's from profiles so that any previous users of this API can easily switch to using this. I considered just keeping the interface to DatabricksJobRunner as the profile, but this wouldn't actually solve the problem that users need to be able to specify the prefix for the remote URL.

mlflow/projects/databricks.py
```diff
@@ -20,7 +20,8 @@
 from mlflow.exceptions import ExecutionException
 from mlflow.utils.mlflow_tags import MLFLOW_DATABRICKS_RUN_URL, MLFLOW_DATABRICKS_SHELL_JOB_ID, \
     MLFLOW_DATABRICKS_SHELL_JOB_RUN_ID, MLFLOW_DATABRICKS_WEBAPP_URL
-from mlflow.utils.uri import get_db_profile_from_uri, is_databricks_uri, is_http_uri
+from mlflow.utils.uri import get_db_info_from_uri, is_databricks_uri, is_http_uri,\
```
sueann: `get_db_info_from_uri` is not used?

mlflow/utils/databricks_utils.py
```diff
@@ -143,7 +144,7 @@ def _fail_malformed_databricks_auth(profile):
                           "https://github.com/databricks/databricks-cli." % profile)
 
 
-def get_databricks_host_creds(profile=None, path=None):
+def get_databricks_host_creds(server_uri=None):
     """
     Reads in configuration necessary to make HTTP requests to a Databricks server. This
     uses the Databricks CLI's ConfigProvider interface to load the DatabricksConfig object.
```
sueann, line 150: Let's add that if no Databricks CLI profile matching the ``server_uri`` is available, we will also try to retrieve it from the Databricks secret manager if available, and in that case ``server_uri`` needs to have the path part which is the key prefix, etc. we should document how to use the secret manager for mlflow in databricks docs, but within the MLflow repo this is likely the only place that will explain it so let's make it more comprehensive.

mlflow/utils/uri.py
```diff
@@ -27,29 +27,22 @@ def is_databricks_uri(uri):
...
     returns None.
     """
     parsed_uri = urllib.parse.urlparse(uri)
     if parsed_uri.scheme == "databricks":
-        if parsed_uri.path:
-            if parsed_uri.path[0] == '/':
-                return parsed_uri.path[1:]
-            return parsed_uri.path
-    return None
+        parsed_path = parsed_uri.path.lstrip('/') or None
```
sueann, line 42: what does `or None` do here?

tests/utils/test_databricks_utils.py
```diff
@@ -49,7 +49,7 @@ def test_databricks_params_custom_profile(ProfileConfigProvider):
     mock_provider.get_config.return_value = \
         DatabricksConfig("host", "user", "pass", None, insecure=True)
     ProfileConfigProvider.return_value = mock_provider
-    params = databricks_utils.get_databricks_host_creds("profile")
+    params = databricks_utils.get_databricks_host_creds("databricks://profile")
```
sueann, line 52: did you want to use `construct_db_uri_from_profile` here too?

tests/utils/test_uri.py
```diff
@@ -23,14 +24,19 @@ def test_extract_db_type_from_uri():
...
 
-def test_get_db_profile_from_uri_casing():
-    assert get_db_profile_from_uri('databricks://aAbB') == 'aAbB'
-
-
-def test_get_db_path_info_from_profile():
-    assert get_db_path_info_from_uri('databricks://profile/prefix') == 'prefix'
-    assert get_db_path_info_from_uri('nondatabricks://profile/prefix') is None
-    assert get_db_path_info_from_uri('databricks://profile') is None
+def test_get_db_info_from_uri():
```
sueann, line 27: could just parameterize and check both the profile and path values for each one?

mlflow/utils/databricks_utils.py
```diff
@@ -143,16 +144,18 @@ def _fail_malformed_databricks_auth(profile):
...
 
 
-def get_databricks_host_creds(profile=None):
+def get_databricks_host_creds(server_uri=None):
     """
     Reads in configuration necessary to make HTTP requests to a Databricks server. This
     uses the Databricks CLI's ConfigProvider interface to load the DatabricksConfig object.
     This method will throw an exception if sufficient auth cannot be found.
 
     :param profile: Databricks CLI profile. If not provided, we will read the default profile.
```
sueann, line 153: profile -> server_uri

sueann (commented):
thanks for the refactor! looks great. just some small comments.

mlflow/utils/databricks_utils.py
ankitmathur-db: Got it - will add documentation around this.

mlflow/utils/uri.py
ankitmathur-db: Previously, if the parsed path was '' it would return None, but if it was '/', it would return ''. To make it consistent, `or None` basically ensures that if the returned value is falsey (''), that it is converted into None.

tests/utils/test_databricks_utils.py
ankitmathur-db: Ah yes! Good point

mlflow/projects/databricks.py
ankitmathur-db: I believe this was updated in one of the lint commits

sueann: ah makes sense. i was confused why lint wasn't failing on it :-)

ankitmathur-db:
@sueann updated based on your comments

sueann (approved):
lgtm!

PR #3076, opened 2020-07-09
Add support for `archive_existing_versions` [Backend API]

harupy:
## What changes are proposed in this pull request?

- Add support for  `archive_existing_versions` in `transition_model_version_stage`.

## How is this patch tested?

Unit tests

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

`transition_model_verstion_stage` now supports the `archive_existing_versions` flag which allows archiving all existing models in the stage.

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [x] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [x] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations

### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [ ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [x] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #3095, opened 2020-07-14
Add `archive_existing_versions` to model registry client

harupy:

A follow-up for #3076 

## How is this patch tested?

(Details)

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -501,7 +501,7 @@ class MlflowClient(object):
         :param version: Registered model version.
         :param stage: New desired stage for this model version.
         :param archive_existing_versions: If this flag is set, all existing model
-        versions in the stage will be atomically moved to the "archived" stage.
+            versions in the stage will be atomically moved to the "archived" stage.
 
         :return: A single :py:class:`mlflow.entities.model_registry.ModelVersion` object.
         """
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -500,8 +500,9 @@ class MlflowClient(object):
         :param name: Registered model name.
         :param version: Registered model version.
         :param stage: New desired stage for this model version.
-        :param archive_existing_versions: If this flag is set, all existing model
-            versions in the stage will be atomically moved to the "archived" stage.
+        :param archive_existing_versions: If this flag is set to ``True``, all existing model
+            versions in the stage will be automically moved to the "archived" stage. Only valid
+            when ``stage`` is ``"staging"`` or ``"production"`` otherwise an error will be raised.
 
         :return: A single :py:class:`mlflow.entities.model_registry.ModelVersion` object.
         """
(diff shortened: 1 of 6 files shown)


mlflow/tracking/client.py
```diff
@@ -493,17 +493,21 @@ def update_model_version(self, name, version, description=None):
...
-    def transition_model_version_stage(self, name, version, stage):
+    def transition_model_version_stage(self, name, version, stage, archive_existing_versions=False):
         """
         Update model version stage.
 
         :param name: Registered model name.
         :param version: Registered model version.
         :param stage: New desired stage for this model version.
+        :param archive_existing_versions: If this flag is set, all existing model
+            versions in the stage will be atomically moved to the "archived" stage.
```
andychow-db, line 504: I think you should mention that this should only be applied for staging and production stages otherwise it'll throw an error.

mlflow/tracking/client.py
```diff
@@ -493,17 +493,21 @@ def update_model_version(self, name, version, description=None):
                                                                 description=description)
 
     @experimental
-    def transition_model_version_stage(self, name, version, stage):
+    def transition_model_version_stage(self, name, version, stage, archive_existing_versions=False):
```
andychow-db, line 496: I think we should add some tests here: https://github.com/mlflow/mlflow/blob/master/tests/tracking/test_client.py

harupy: I'll add tests.

mlflow/tracking/client.py
```diff
@@ -493,17 +493,22 @@ def update_model_version(self, name, version, description=None):
...
         """
         Update model version stage.
 
         :param name: Registered model name.
         :param version: Registered model version.
         :param stage: New desired stage for this model version.
+        :param archive_existing_versions: If this flag is set to ``True``, all existing model
+            versions in the stage will be automically moved to the "archived" stage. Only valid
+            when ``stage`` is ``"staging"`` or ``"production"`` otherwise an error will be raised.
```
harupy, line 505: @andychow-db cc @sueann 
Fixed the description for `archive_existing_versions`. After making sure this is fine, I'll apply the change to other `transition_model_version_stage`s.

andychow-db (approved):
LGTM

mlflow/tracking/client.py
sueann: @andychow-db - i think we had a discussion (a really long time ago) about what to set the default for this API. one concern was that if we set it to be `False` it doesn't match the UI which is `True` by default. Do you remember what we decided there? i assume we want to set it to `False` here to not break any existing usages.

andychow-db: I vaguely remember we said the UI would be true and the client would default to False to preserve behavior.

sueann: ok

Issue #3117, opened 2020-07-16
[BUG]Cannot parse "Order by a metric" if it contains a space


jdlesage:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for additional information about bug reports. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

**Please fill in this bug report template to ensure a timely and thorough response.**

### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [X] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: no
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: centos 7
- **MLflow installed from (source or binary)**: source
- **MLflow version (run ``mlflow --version``)**: master branch
- **Python version**: 3.6
- **npm version, if running the dev UI**: 
- **Exact command to reproduce**:

### Describe the problem + how to reproduce the issue
Create a run with a metric that contains a space in name. Such as "Mean Square Error". 
Order following this column from the UI
Got the following error: `Invalid order_by clause 'metrics.`Mean Square Error` DESC'. Could not be parsed.`

### Other info / logs
Error seems to be on backend (sqlalchemy store in my case) . The POST search request has the correct order_by attribute in payload but response code from backend is 400 + INVALID_PARAMETER_VALUE

### What component(s), interfaces, languages, and integrations does this bug affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [X] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
[...]

PR #3118, opened 2020-07-16
Fix order by parsing when name contains a space

jdlesage:

A fix to https://github.com/mlflow/mlflow/issues/3117. (parse orderby clause with a space in name)

The solution is to use shlex.split instead of string split that is able to manage quotes.

## How is this patch tested?

* Adding a unit test for this case
* Test ordering from the UI (columns with space or not)

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -368,7 +368,6 @@ class SearchUtils(object):
         token_value = cls._validate_order_by_and_generate_token(order_by)
         is_ascending = True
         tokens = shlex.split(token_value.replace("`", "\""))
-        print(tokens)
         if len(tokens) > 2:
             raise MlflowException(f"Invalid order_by clause '{order_by}'. Could not be parsed.",
                                   error_code=INVALID_PARAMETER_VALUE)
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -2,6 +2,7 @@ import base64
 import json
 import operator
 import re
+import shlex
 
 import sqlparse
 from sqlparse.sql import Identifier, Token, Comparison, Statement
@@ -366,7 +367,8 @@ class SearchUtils(object):
     def _parse_order_by_string(cls, order_by):
         token_value = cls._validate_order_by_and_generate_token(order_by)
         is_ascending = True
-        tokens = token_value.split()
+        tokens = shlex.split(token_value.replace("`", "\""))
+        print(tokens)
         if len(tokens) > 2:
             raise MlflowException(f"Invalid order_by clause '{order_by}'. Could not be parsed.",
                                   error_code=INVALID_PARAMETER_VALUE)
--- a/tests/utils/test_search_utils.py
+++ b/tests/utils/test_search_utils.py
@@ -328,6 +328,14 @@ def test_invalid_order_by_search_runs(order_by, error_message):
     assert error_message in e.value.message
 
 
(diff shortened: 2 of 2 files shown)


mlflow/utils/search_utils.py
```diff
@@ -366,7 +367,8 @@ def _validate_order_by_and_generate_token(cls, order_by):
     def _parse_order_by_string(cls, order_by):
         token_value = cls._validate_order_by_and_generate_token(order_by)
         is_ascending = True
-        tokens = token_value.split()
+        tokens = shlex.split(token_value.replace("`", "\""))
+        print(tokens)
```
harupy, line 371: @jdlesage Could you remove `print` here?

harupy, line 371:
```suggestion
```
@jdlesage Could you remove print here?

jdlesage:
@harupy I removed the print statement. Could you please review it ? Thank you very much

tests/utils/test_search_utils.py
```diff
@@ -328,6 +328,14 @@ def test_invalid_order_by_search_runs(order_by, error_message):
     assert error_message in e.value.message
 
 
+def test_space_order_by_search_runs():
+    identifier_type, identifier_name, ascending = SearchUtils.parse_order_by_for_search_runs(
+        "metrics.`Mean Square Error`")
+    assert identifier_type == "metric"
+    assert identifier_name == "Mean Square Error"
+    assert ascending
```
harupy, line 336:
```suggestion
@pytest.mark.parametrize("order_by, ascending_expected", [
    ("metrics.`Mean Square Error`", True),
    ("metrics.`Mean Square Error` ASC", True),
    ("metrics.`Mean Square Error` DESC", False),
])
def test_space_order_by_search_runs(order_by, ascending_expected):
    identifier_type, identifier_name, ascending = SearchUtils.parse_order_by_for_search_runs(
        order_by)
    assert identifier_type == "metric"
    assert identifier_name == "Mean Square Error"
    assert ascending == ascending_expected
```
@jdlesage Sorry for the delay. Can we add two more test cases to ensure that order_by query with `ASC|DESC` annotation works too?

jdlesage: Done

harupy: @jdlesage Thanks!

harupy (approved):
LGTM

mlflow/utils/search_utils.py
```diff
@@ -366,7 +367,7 @@ def _validate_order_by_and_generate_token(cls, order_by):
     def _parse_order_by_string(cls, order_by):
         token_value = cls._validate_order_by_and_generate_token(order_by)
         is_ascending = True
-        tokens = token_value.split()
```
mparkhe, line 369: One quick question @jdlesage : `filter_string` supports quote parsing. Was there a reason the code was not reused or consolidated? Seems like quote handling operations are different handled for these 2 arguments.

mparkhe (commented):
I know, this pull request has been merged already but was curious about the approach. Maybe we can consolidate the code for `filter` and `order_by`.

cc: @harupy

mlflow/utils/search_utils.py
jdlesage: There are no specific reasons. I didn't know `fitler_string` manages correctly the quote parsing (I didn't test it). I agree it could be a nice improvement to factorize the 2 ways to manage quotes.

PR #3136, opened 2020-07-20
Fix empty creator field in model version if basic authentication is set up

AnastasiaKol:

Use username from request's `authorization` header (if such exists), to fill in `Creator` field in model version page when create a new model version

## How is this patch tested?

Unit test + manual test

`Creator` field in model version page is extracted from request's `authorization` header (if such exists) when creating a new model version

Components 
- [X] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry

Interface 

Language 

Integrations


mlflow/server/handlers.py
```diff
@@ -152,6 +152,13 @@ def _get_request_message(request_message, flask_request=request):
     return request_message
 
 
+def _get_username(flask_request=request):
+    # Get username from basic authentication
+    if "authorization" in flask_request.headers:
```
andychow-db, line 157: Authentication might not be the right header here according to [MDN authorization header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) because it contains credentials.

Perhaps use the header "from". Source: [MDN from header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/From)?

AnastasiaKol: @andychow-db oh so you don't think using username `authorization` is a good idea? it's just the way we've set up our server, and i thought it might be useful for others who have similar setup with basic authentication and want to know which user created the model.

andychow-db: From the doc, authorization may contain credentials which will be saved into the tracking DB as the user.

andychow-db: It sounds like this might be more common practice than I recognize. @smurching , @dbczumar what do y'all think here?

smurching: Seems reasonable to me to detect the model version creator via the username provided via basic auth - I think this is what most other servers that implement the MLflow REST API (e.g. Databricks) do

andychow-db: @smurching Alright thank you.

andychow-db (approved):

mlflow/server/handlers.py
AnastasiaKol: @andychow-db Hi:) wondering if and when this pr is gonna be merged? and should i be keeping merging master branch to make sure there are no conflicts?

PR #3164, opened 2020-07-23
[Multi-Workspace Registry] add client changes for populating registry lineage through run-link

ankit-db:

In this PR, we introduce `run_link` as a parameter that can be passed into `create_model_version`. This will allow users to track the exact lineage of their model versions in settings where the Model Registry is hosted on a different MLflow instance than the tracking server. 

If using MLflow in a Databricks notebook where the tracking server is in the workspace of the notebook, the run link is automatically populated. Otherwise, we expect a profile for the tracking URI of the form: databricks://profile/prefix, where the host name, token, and workspace ID are stored as secrets in the Databricks Secret Manager with the scope=<profile> and the key being <prefix>-host, <prefix>-token, and <prefix>-workspace-id,

## How is this patch tested?

Unit testing and manual testing (please contact for instructions on how to manual test)

Introduces run_link to create_model_version to allow for lineage tracking in scenarios where the model registry is hosted in a different MLflow instance.

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -17,6 +17,10 @@ from mlflow.tracking.registry import UnsupportedModelRegistryStoreURIException
 from mlflow.tracking._tracking_service import utils
 from mlflow.tracking._tracking_service.client import TrackingServiceClient
 from mlflow.utils import experimental
+from mlflow.utils.databricks_utils import is_databricks_default_tracking_uri, \
+    is_in_databricks_notebook, get_workspace_info_from_dbutils, \
+    get_workspace_info_from_databricks_secrets
+from mlflow.utils.uri import is_databricks_uri, construct_run_url
 
 _logger = logging.getLogger(__name__)
 
@@ -488,7 +492,7 @@ class MlflowClient(object):
     # Model Version Methods
 
     @experimental
-    def create_model_version(self, name, source, run_id, tags=None):
+    def create_model_version(self, name, source, run_id, tags=None, run_link=None):
         """
         Create a new model version from given source or run ID.
 
@@ -497,10 +501,35 @@ class MlflowClient(object):
         :param run_id: Run ID from MLflow tracking server that generated the model
         :param tags: A dictionary of key-value pairs that are converted into
                      :py:class:`mlflow.entities.model_registry.ModelVersionTag` objects.
+        :param run_link: Link to the run from an MLflow tracking server that generated this model.
         :return: Single :py:class:`mlflow.entities.model_registry.ModelVersion` object created by
                  backend.
         """
-        return self._get_registry_client().create_model_version(name, source, run_id, tags)
+        tracking_uri = self._tracking_client.tracking_uri
+        # for Databricks backends, we support automatically populating the run link field
+        if is_databricks_uri(tracking_uri) and tracking_uri != self._registry_uri and not run_link:
+            # if using the default Databricks tracking URI and in a notebook, we can automatically
+            # figure out the run-link.
+            if is_databricks_default_tracking_uri(tracking_uri) and is_in_databricks_notebook():
+                # use DBUtils to determine workspace information.
+                workspace_host, workspace_id = get_workspace_info_from_dbutils()
+            else:
+                # in this scenario, we're not able to automatically extract the workspace ID
+                # to proceed, and users will need to pass in a databricks profile with the scheme:
+                # databricks://scope/prefix and store the host and workspace-ID as a secret in the
+                # Databricks Secret Manager with scope=<scope> and key=<prefix>-workspaceid.
+                workspace_host, workspace_id = \
+                    get_workspace_info_from_databricks_secrets(tracking_uri)
+            # retrieve experiment ID of the run for the URL
+            experiment_id = self.get_run(run_id).info.experiment_id
+            if workspace_host and run_id and experiment_id:
+                run_link = construct_run_url(workspace_host, experiment_id, run_id, workspace_id)
+        return self._get_registry_client().create_model_version(
+            name=name,
+            source=source,
+            run_id=run_id,
+            tags=tags,
+            run_link=run_link)
 
     @experimental
     def update_model_version(self, name, version, description=None):
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -522,10 +522,10 @@ class MlflowClient(object):
                     get_workspace_info_from_databricks_secrets(tracking_uri)
                 if not workspace_id:
                     print("No workspace ID specified; if your Databricks workspaces share the same"
-                          "host URL, you may want to specify the workspace ID (along with the host"
-                          "information in the secret manager) for run lineage tracking. For more"
-                          "details on how to specify this information in the secret manager,"
-                          "please refer to the model registry documentation.")
+                          " host URL, you may want to specify the workspace ID (along with the host"
+                          " information in the secret manager) for run lineage tracking. For more"
+                          " details on how to specify this information in the secret manager,"
+                          " please refer to the model registry documentation.")
             # retrieve experiment ID of the run for the URL
             experiment_id = self.get_run(run_id).info.experiment_id
             if workspace_host and run_id and experiment_id:
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -520,6 +520,12 @@ class MlflowClient(object):
                 # Databricks Secret Manager with scope=<scope> and key=<prefix>-workspaceid.
                 workspace_host, workspace_id = \
                     get_workspace_info_from_databricks_secrets(tracking_uri)
+                if not workspace_id:
+                    print("No workspace ID specified; if your Databricks workspaces share the same"
+                          "host URL, you may want to specify the workspace ID (along with the host"
+                          "information in the secret manager) for run lineage tracking. For more"
+                          "details on how to specify this information in the secret manager,"
+                          "please refer to the model registry documentation.")
             # retrieve experiment ID of the run for the URL
             experiment_id = self.get_run(run_id).info.experiment_id
             if workspace_host and run_id and experiment_id:
--- a/mlflow/utils/databricks_utils.py
+++ b/mlflow/utils/databricks_utils.py
@@ -151,14 +151,14 @@ def get_workspace_info_from_dbutils():
         workspace_host = context['extraContext']['api_url']
         workspace_id = context['tags']['orgId']
         return workspace_host, workspace_id
+    return None, None
 
 
 def get_workspace_info_from_databricks_secrets(tracking_uri):
-    profile, path = get_db_info_from_uri(tracking_uri)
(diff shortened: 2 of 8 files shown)


ankitmathur-db:
The linter appears to have mixed up the previous commit with this one?

mlflow/tracking/client.py
```diff
@@ -497,10 +501,35 @@ def create_model_version(self, name, source, run_id, tags=None):
...
+                # in this scenario, we're not able to automatically extract the workspace ID
+                # to proceed, and users will need to pass in a databricks profile with the scheme:
+                # databricks://scope/prefix and store the host and workspace-ID as a secret in the
+                # Databricks Secret Manager with scope=<scope> and key=<prefix>-workspaceid.
+                workspace_host, workspace_id = \
+                    get_workspace_info_from_databricks_secrets(tracking_uri)
+            # retrieve experiment ID of the run for the URL
+            experiment_id = self.get_run(run_id).info.experiment_id
+            if workspace_host and run_id and experiment_id:
+                run_link = construct_run_url(workspace_host, experiment_id, run_id, workspace_id)
```
sueann, line 526: it'd be good to put this in a utility function to make this method easier to read - but let me do that in my PR since the if condition will need to be refactored anyway.

tests/utils/test_uri.py
```diff
@@ -37,6 +37,27 @@ def test_get_db_info_from_uri(server_uri, result):
...
+    ('https://www.databricks.com/', '19201', '2231', '0',
+     'https://www.databricks.com/#mlflow/experiments/19201/runs/2231'),
+    (None, '19201', '2231', '0', 'Exception'),
+    ('https://www.databricks.com/', None, '2231', '0', 'Exception'),
+    ('https://www.databricks.com/', '19201', None, '0', 'Exception'),
+])
+def test_construct_run_url(hostname, experiment_id, run_id, workspace_id, result):
+    if result == 'Exception':
+        with pytest.raises(MlflowException):
+            construct_run_url(hostname, experiment_id, run_id, workspace_id)
```
sueann, line 56: probably cleaner to just separate it out into another test case like `test_construct_run_url_errors`

mlflow/utils/databricks_utils.py
```diff
@@ -138,6 +143,28 @@ def get_webapp_url():
...
 
+def get_workspace_info_from_dbutils():
+    dbutils = _get_dbutils()
+    if dbutils:
+        context = json.loads(
+            dbutils.notebook.entry_point.getDbutils().notebook().getContext().toJson())
+        workspace_host = context['extraContext']['api_url']
+        workspace_id = context['tags']['orgId']
+        return workspace_host, workspace_id
+
```
sueann, line 154: `return None, None` explicitly?

mlflow/tracking/client.py
sueann, line 526: I wonder if we should fail here if workspace_id is not given, or just give a warning. I guess throwing is okay - user can give a run_link explicitly to bypass this problem. I am not sure what the best way to document this is though - probably in the databricks docs for multi-workspace workflows.

mlflow/utils/databricks_utils.py
```diff
@@ -138,6 +143,28 @@ def get_webapp_url():
...
+
+def get_workspace_info_from_databricks_secrets(tracking_uri):
+    profile, path = get_db_info_from_uri(tracking_uri)
```
sueann, line 157: why not just name `path` `key_prefix`?

tests/tracking/test_client.py
```diff
@@ -252,3 +253,63 @@ def test_registry_uri_from_implicit_tracking_uri():
         get_tracking_uri_mock.return_value = tracking_uri
         client = MlflowClient()
         assert client._registry_uri == tracking_uri
+
+
+def test_create_model_version_normal(mock_registry_store):
```
sueann, line 258: `test_create_model_version_nondatabricks_source_no_runlink` maybe? should also add a test with `run_link` explicitly provided to `MlflowClient.create_model_version`.

tests/tracking/test_client.py
```diff
@@ -252,3 +253,63 @@ def test_registry_uri_from_implicit_tracking_uri():
...
+    hostname = 'https://workspace.databricks.com/'
+    workspace_id = '10002'
+    run_id = 'runid'
+    workspace_url = construct_run_url(hostname, experiment_id, run_id, workspace_id)
+    get_run_mock = mock.MagicMock()
+    get_run_mock.return_value = Run(RunInfo(run_id, experiment_id, 'userid', 'status', 0, 1, None),
+                                    None)
+    with mock.patch('mlflow.tracking.client.is_in_databricks_notebook', return_value=True), mock\
+            .patch('mlflow.tracking.client.get_workspace_info_from_dbutils',
+                   return_value=(hostname, workspace_id)):
```
sueann, line 283:
```suggestion
    with mock.patch('mlflow.tracking.client.is_in_databricks_notebook', return_value=True), \
            mock.patch('mlflow.tracking.client.get_workspace_info_from_dbutils',
                       return_value=(hostname, workspace_id)):
```

sueann (commented):
lgtm - just small comments inline

mlflow/tracking/client.py
ankitmathur-db: Agreed - sounds good

ankitmathur-db: Makes sense

tests/utils/test_uri.py
ankitmathur-db: yeah looking back on it, definitely agree. will update!

mlflow/tracking/client.py
ankitmathur-db: Yeah my concern was that ST users might not know what the workspace-id is, and so they might just not provide one in the secrets, so then that would just return None. In that scenario, it'd still be ok since it isn't needed in the URL anyways.

For MT users, if they didn't specify a workspace ID, would going to the URL just make them log in?

mlflow/utils/databricks_utils.py
ankitmathur-db: yeah that's better - was carrying over terminology in my head from the URLlib parsing, but there's no need

mlflow/tracking/client.py
sueann: mm yeah.. that's a good point. i think if they only have access to 1 workspace it will go to that one, and if not, put them in the workspace selector menu. maybe we can print out a message like "No workspace ID specified; if your Databricks workspaces share the same host URL, you may want to specify the workspace ID (along with the host information in the secret manager) for run lineage tracking." if no workspace_id is found.

tests/tracking/test_client.py
ankitmathur-db: Linters hate me today lol so I tried to do this and it complained about the continuation line being over-indented. Moving the second `mock` line back *1 space* has apparently fixed the problem and looks more readable than the original (though less clean than your suggestion)

mlflow/tracking/client.py
ankitmathur-db: Yeah makes sense - will add that

mlflow/tracking/client.py
```diff
@@ -520,6 +520,12 @@ def create_model_version(self, name, source, run_id, tags=None, run_link=None):
                 # Databricks Secret Manager with scope=<scope> and key=<prefix>-workspaceid.
                 workspace_host, workspace_id = \
                     get_workspace_info_from_databricks_secrets(tracking_uri)
+                if not workspace_id:
+                    print("No workspace ID specified; if your Databricks workspaces share the same"
+                          "host URL, you may want to specify the workspace ID (along with the host"
+                          "information in the secret manager) for run lineage tracking. For more"
+                          "details on how to specify this information in the secret manager,"
+                          "please refer to the model registry documentation.")
```
sueann, line 528: i think you need a space between the consecutive lines? or can just do a large print block ("""...""") that will have some newlines in it.

tests/utils/test_databricks_utils.py
```diff
@@ -85,6 +87,52 @@ def test_databricks_single_slash_in_uri_scheme_throws(get_config):
...
+
+def test_get_workspace_info_from_dbutils():
+    mock_dbutils = mock.MagicMock()
+    mock_dbutils.notebook.entry_point.getDbutils.return_value.notebook.return_value.getContext\
+        .return_value.toJson.return_value = '{"extraContext":' \
+                                            '{"api_url": "https://mlflow.databricks.com"},' \
+                                            '"tags": {"orgId" : "1111"}}'
+    with mock.patch("mlflow.utils.databricks_utils._get_dbutils", return_value=mock_dbutils):
+        workspace_host, workspace_id = \
+            get_workspace_info_from_dbutils()
```
sueann, line 110:
```suggestion
        workspace_host, workspace_id = get_workspace_info_from_dbutils()
```

tests/utils/test_databricks_utils.py
```diff
@@ -85,6 +87,52 @@ def test_databricks_single_slash_in_uri_scheme_throws(get_config):
...
+        assert workspace_host == 'https://mlflow.databricks.com'
+        assert workspace_id == '1111'
+
+
+def test_get_workspace_info_from_dbutils_when_no_dbutils_available():
+    with mock.patch("mlflow.utils.databricks_utils._get_dbutils", return_value=None):
+        workspace_host, workspace_id = \
+            get_workspace_info_from_dbutils()
```
sueann, line 118:
```suggestion
        workspace_host, workspace_id = get_workspace_info_from_dbutils()
```

tests/tracking/test_client.py
```diff
@@ -252,3 +253,84 @@ def test_registry_uri_from_implicit_tracking_uri():
...
+    experiment_id = 'test-exp-id'
+    hostname = 'https://workspace.databricks.com/'
+    workspace_id = '10002'
+    run_id = 'runid'
+    workspace_url = construct_run_url(hostname, experiment_id, run_id, workspace_id)
+    get_run_mock = mock.MagicMock()
+    get_run_mock.return_value = Run(RunInfo(run_id, experiment_id, 'userid', 'status', 0, 1, None),
+                                    None)
+    with mock.patch('mlflow.tracking.client.is_in_databricks_notebook',
+                    return_value=True),\
```
sueann, line 302:
```suggestion
                    return_value=True), \
```

sueann (approved):
lgtm - just one formatting issue to fix (and super-nit suggestions) before merging.

mlflow/tracking/client.py
ankitmathur-db: Added spaces! thanks for catching that

sueann:
oh no these suggestion commits fail the DCO check... 😢

ankitmathur-db:
It's ok - the force push to fix signoff has become second nature :)

Looks like it's just a deficiency in the Github Web UI (https://github.com/todogroup/gh-issues/issues/50)?

Issue #3043, opened 2020-07-03
[BUG] MLFlow version not specified in packages


mikeoconnor0308:
### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [x] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10
- **MLflow installed from (source or binary)**: Binary
- **MLflow version (run ``mlflow --version``)**: 1.9.1
- **Python version**: 3.6
- **npm version, if running the dev UI**:
- **Exact command to reproduce**:

### Describe the problem

MLFlow packages do not specify an MLFlow version in the produced conda.yaml: 

e.g.
```yaml
channels:
- defaults
- conda-forge
dependencies:
- python=3.7.7
- fastai=1.0.61
- pip
- pip:
  - mlflow
name: mlflow-env
```

This may be an intentional feature rather than a _bug_, but it strikes me as a bit odd to assume a package produced from one version of mlflow will work with a newer version? 

Is there some rationale of compatibility here, or should the ability to specify the mlflow version when saving a model by added?

dmatrix:
@mikeoconnor0308 

Thanks for filing this issue. 

Have you been able to save a model with a specific version of mlflow and run that as a project, to see if the versions are identical?

@dbczumar @tomasatdatabricks We don't seem to save mlfow version as part of the conda.yaml 
![Screen Shot 2020-07-06 at 10 47 47 AM](https://user-images.githubusercontent.com/1117597/86623298-32f22e80-bf76-11ea-8a41-4a4e127328e9.png)

Is there a particular reason why and how do we reproduce the environment during mlflow run  for projects?
I know that it works but where do we get the mlflow version from?

tomasatdatabricks:
@dmatrix @mikeoconnor0308 Yes this is indeed intentional and the assumption is that models logged with older versions of MLflow will continue to work with newer MLflow versions.

The main reason for this is that the MLflow dependency is also used to serve the model (i.e. it defines the web server) and if there are any changes to the model server in the future MLflow these would not apply to older models.

[...]

PR #3176, opened 2020-07-27
fix mlflow version when creating conda env

### LINKED ISSUE #2979: [BUG] mlflow version is not fixed in default conda env
### Willingness to contribute

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OS 10.15.2
- **MLflow version (run ``mlflow --version``)**: 1.5.0
- **Python version**: 3.7
- **Exact command to reproduce**: mlflow.sklearn.log_model(lr, "model")

### Describe the problem
When loging any kind of supported model, `mlflow` is added as a pip dependency to the `conda.yaml`. The issue is that it is added without any fixed version, which results in the conda env having the latest `mlflow` version even if the training environment had an older one.

### Code to reproduce issue
Any log model function uses `_mlflow_conda_env()` which adds the version less `mlflow` pip dependency.

https://github.com/mlflow/mlflow/blob/10f336a0c86cbabab3089fdfc3943884d2f807ca/mlflow/utils/environment.py#L13-L14

https://github.com/mlflow/mlflow/blob/10f336a0c86cbabab3089fdfc3943884d2f807ca/mlflow/utils/environment.py#L31-L32

for instance, the following code

```python
mlflow.sklearn.log_model(lr, "model")
```

will results in the following `conda.yaml`
```yaml
channels:
- defaults
dependencies:
- python=3.6.10
- scikit-learn=0.22.1
- pip:
  - mlflow
  - cloudpickle==1.4.1
name: mlflow-env
```

a simple fix would be to infer `mlflow` version similar to what is done with other dependencies. for instance:

https://github.com/mlflow/mlflow/blob/6e60e79a593fed2ff5a08aa637a79354bee1be0e/mlflow/sklearn.py#L46-L49

eavidan:

When loging any kind of supported model, mlflow is added as a pip dependency to the `conda.yaml`. The following change with fix the mlflow version to the one used to log the model.

Fixes: #2979 

## How is this patch tested?
```py
import numpy as np
import mlflow
import mlflow.sklearn

X = np.array([[1, 1], [1, 2], [2, 2], [2, 3]])
y = np.dot(X, np.array([1, 2])) + 3
reg = LinearRegression().fit(X, y)
reg.score(X, y)

with mlflow.start_run():
    mlflow.sklearn.log_model(sk_model=reg, artifact_path="model")
```

results in the following `conda.yaml`
```yaml
channels:
- defaults
- conda-forge
dependencies:
- python=3.6.10
- scikit-learn=0.20.2
- pip
- pip:
  - mlflow==1.9.2.dev0
  - cloudpickle==0.8.0
name: mlflow-env
```

unit test added

When logging a model the `conda.yaml` file will an `mlflow` pip dependency with the version used to log the model

Components 
Model Registry

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/environment.py
+++ b/mlflow/utils/environment.py
@@ -28,8 +28,12 @@ def _mlflow_conda_env(path=None, additional_conda_deps=None, additional_pip_deps
     :return: ``None`` if ``path`` is specified. Otherwise, the a dictionary representation of the
              Conda environment.
     """
-    pip_deps = (["mlflow"] if install_mlflow else []) + (
-        additional_pip_deps if additional_pip_deps else [])
+    pip_deps = []
+    if install_mlflow:
+        import mlflow
+        pip_deps.append("mlflow=={}".format(mlflow.__version__))
+    pip_deps.extend(additional_pip_deps) if additional_pip_deps else []
+
     conda_deps = (additional_conda_deps if additional_conda_deps else []) + (
         ["pip"] if pip_deps else [])
 
--- a/tests/utils/test_environment.py
+++ b/tests/utils/test_environment.py
@@ -53,3 +53,14 @@ def test_mlflow_conda_env_includes_pip_dependencies_and_pip_is_specified(pip_spe
         assert conda_dep in env["dependencies"]
     assert pip_specification in env["dependencies"]
     assert env["dependencies"].count("pip") == (2 if pip_specification == "pip" else 1)
+
+
+def test_mlflow_conda_env_includes_a_fixed_mlflow_version_by_default():
+    import mlflow
+    env = _mlflow_conda_env()
+    for dep in env["dependencies"]:
+        if isinstance(dep, dict) and dep.get("pip") is not None:
+            assert "mlflow=={}".format(mlflow.__version__) in dep.get("pip")
+            return
+
+    assert False


harupy:
@eavidan Thanks for the PR, but we don't restrict mlflow version for forward compatibility (please take a look at [this issue](https://github.com/mlflow/mlflow/issues/3043))

eavidan:
@harupy I was sure it was a bug (#2979). 
I can understand your motivation for forward compatibility for old models. 

having said that, we are using mlflow in production and my issue was reported due to real world experience,
 where some of our models started failing due to dependencies collisions. Since the mlflow version was not fixed so does its dependencies.
for us this was an issue and we found ourselves writing a work around for each mlflow model deployed, fixing the mlflow version to the one existing on our mlflow server.

Do you have any suggestions on how to handle this issue better?

harupy:
@eavidan 
How about using `conda_env` argument in `mlflow.<flavor>.log_model`? For example,

```python
with mlflow.start_run():
    mlflow.sklearn.log_model(
        sk_model=sk_model,
        artifact_path="model",
        conda_env=conda_env_dict_where_mlflow_version_is_specified,
 )
```

eavidan:
@harupy thanks for the suggestions, but this will require all our data scientists to start defining their serving dependencies which we wanted to avoid. Currently, I believe we will keep solving this on our deployment side.

If the reason is forward compatibility then maybe the way to go is to align the conda env with the mlflow server version during the invocation of the `load_model`.
This way we can get an alignment between the model and server mlflow versions. 
Getting the latest mlflow version for the model can conflict with an mlflow server with a fixed version as well.
does this make sense or am I am missing something?

dbczumar:
@eavidan Can you provide more details about the dependency collision / undesirable behavior you encountered? Did the latest version of MLflow specify minimum / maximum library version dependencies that conflicted with other libraries in your conda environment?

eavidan:
hi @dbczumar. We use `seldon-core 1.1` as our serving platform. Seldon core requires `azure-storage-blob >= 2.0.1, < 3.0.0`.
everything went well while we were working with `mlflow 1.5.0` as azure in general was an extra requirement, but `mlflow 1.9.0` requires `azure-storage-blob>=12.0`. Hence the conflict. The funny thing is that we don't use azure :).

as a side note, I would move these kind of the dependencies to the `extras_require` section and the import statement into the relevant methods, but I understand that this is not always possible.

Issue #3157, opened 2020-07-22
[FR] Pagination support for list_run_infos


trangevi:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [ ] Yes. I can contribute this feature independently.
- [x] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

`tracking.MlflowClient.list_run_infos` provides a convenience wrapper around `search_runs`, but doesn't surface the pagination support which `search_runs` provides. This FR is to surface the necessary parameters/return handling.

## Motivation
- What is the use case for this feature? This will be beneficial for users who have a significant number of runs.
- Why is this use case valuable to support for MLflow users in general? It will allow for the continued use of the `list_run_infos` convenience function even as users scale up in the number of runs they have.
- Why is this use case valuable to support for your project(s) or organization? As above, the continued use of the convenience function will be valuable to users at scale.
- Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient). Users are currently forced to change how they interact with MLflow as their use case scales, which while not impossible is detrimental to user experience.

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [x] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interfaces
[...]

PR #3183, opened 2020-07-28
Add Pagination to list_run_infos

trangevi:

Add pagination handling to `list_run_infos`, and add it to the fluent API to match `search_runs`. The function in the fluent API will perform list calls if it receives a token back until it has retrieved `max_results` number of results, same as `search_runs`. The function in the store only makes a single call and returns the token as a part of the result.

Resovles #3157

## How is this patch tested?

Added unit tests matching the `search_runs` pagination tests.

Adding `list_run_infos` to the fluent API and adding pagination support. 

Components 
Model Registry

Interface 

Language 

Integrations


AveshCSingh (changes requested):
Thanks for implementing this solution to #3157!

I've left some comments inline. PTAL, and feel free to reply or Slack me with any follow-up questions.

mlflow/tracking/fluent.py
```diff
@@ -502,6 +503,38 @@ def _get_paginated_runs(experiment_ids, filter_string, run_view_type, max_result
     return all_runs
 
 
+def list_run_infos(experiment_id, run_view_type=ViewType.ACTIVE_ONLY,
```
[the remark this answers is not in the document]
trangevi: did some refactoring, please take a look and let me know what you think.

tests/tracking/test_fluent.py
```diff
@@ -476,62 +492,64 @@ def test_search_runs_no_arguments():
     mock_experiment_id = mock.Mock()
     experiment_id_patch = mock.patch("mlflow.tracking.fluent._get_experiment_id",
                                      return_value=mock_experiment_id)
-    get_paginated_runs_patch = mock.patch('mlflow.tracking.fluent._get_paginated_runs',
-                                          return_value=[])
+    get_paginated_runs_patch = mock.patch('mlflow.tracking.fluent._paginate', return_value=[])
```
trangevi, line 495: I don't like the changes in this test, as I don't feel like they are actually testing anything really, at least not really what the name implies, but I couldn't figure out a good way to validate a lambda short of what I added, which only validates that it was a callable, not anything about it. If you have any ideas of how to improve this please let me know.

mlflow/tracking/fluent.py
```diff
@@ -482,24 +486,52 @@ def search_runs(experiment_ids=None, filter_string="", run_view_type=ViewType.AC
...
+
+    :param paginated_fn:
+    :type paginated_fn: This function is expected to take in the number of results to retrieve
+        per page and a pagination token, and return a PagedList object
+    :param max_results_per_page:
+    :type max_results_per_page: The maximum number of results to retrieve per page
+    :param max_results:
+    :type max_results: The maximum number of results to retrieve overall
+    :return:
+    :rtype:
```
AveshCSingh, line 519: Could you fill in the `:return` and `:rtypes`?

tests/tracking/test_fluent.py
AveshCSingh: Wow you're right, this is a complicated test to write. 

I know this is lame, but what do you think of just asserting that `_get_experiment_id` was called? I think this does validate the behavior the test is trying to verify. (If you decide to do this, you should also update the docstring to ` When no experiment ID is specified, it should try to get the implicit one`.)

mlflow/tracking/fluent.py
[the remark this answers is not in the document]
AveshCSingh: Beautiful! I think this is great.

AveshCSingh (commented):
I really like the refactor. I've added a few minor comments, otherwise this looks good.

trangevi:
Had to do a rebase and force push for the DCO thing, hopefully that doesn't mess with the review much.

tests/tracking/test_fluent.py
AveshCSingh: IMO you should actually remove this test, since it's not really testing anything except that the lambda gets called.

AveshCSingh (approved):
LGTM! I'll merge this once the tests pass. The CircleCI test should be fixed by merging in master. I've just done this.

For the lint test, could you remove the _LambdaValidator test? If you have a compelling reason to keep it, you can instead keep it & update the code to pass the linter (define a function rather than a lambda).

PR #2575, opened 2020-03-12
Fix for running MLflow Projects on Windows

juntai-zheng:
## What changes are proposed in this pull request?

Addresses #1670.

## How is this patch tested?

Travis

### Is this a user-facing change?

- [X] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s) does this PR affect?

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [ ] Models
- [ ] Scoring
- [ ] Serving
- [ ] R
- [ ] Java
- [ ] Python

### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [X] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [ ] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #3185, opened 2020-07-28
Re-enable + fix projects tests for Windows

apurva-koti:

#2575, re-done with the current state of master (with a considerable refactor) in mind.
Also re-enables the windows project test in github actions.

## How is this patch tested?

Github actions

Components 
Model Registry

Interface 
- [x] `area/windows`: Windows support

Language 

Integrations


apurva-koti:
The majority of these fixes are related to cosmetic differences, ie. forward slash in linux, backslash in windows, `urlparse` incorrectly identifying `C:\Users\...` as having scheme = `C`.

Some docker tests are leftover and review/discussion is welcomed.

mlflow/data.py
```diff
@@ -68,7 +68,10 @@ def parse_gs_uri(uri):
 
 def is_uri(string):
     parsed_uri = urllib.parse.urlparse(string)
-    return len(parsed_uri.scheme) > 0
+    if os.name == 'nt':
+        return not re.match('^[a-zA-Z]:\\\\.*', string)
```
harupy, line 72:
```suggestion
        return not re.match(r'^[a-zA-Z]:\\.*', string)
```
You can use raw string here.

harupy: And can we put this in a function (named like `is_windows_drive_path`) and write a test?

apurva-koti: @harupy, when I tried raw string with two slashes the function errored out with “Expected string or bytes-like object”.
Function has been moved, will write test.

travis/install-common-deps.sh
```diff
@@ -19,14 +19,14 @@ conda create -q -n test-environment python=3.6
 source activate test-environment
 
 python --version
-pip install --upgrade pip==19.3.1
+pip install --upgrade pip
```
apurva-koti, line 22: Written in to get flaky dependency installation to work, to verify windows tests would pass. This should be factored out into another PR, if it is indeed the fix.

mlflow/projects/backend/local.py
```diff
@@ -229,14 +230,13 @@ def _get_local_artifact_cmd_and_envs(artifact_repo):
...
-        win_user_dir = os.environ["USERPROFILE"]
-        aws_path = os.path.join(win_user_dir, ".aws")
+        aws_path = ntpath.expanduser(os.path.join("~", ".aws"))
     else:
         aws_path = posixpath.expanduser("~/.aws")
 
     volumes = []
     if posixpath.exists(aws_path):
-        volumes = ["-v", "%s:%s" % (str(aws_path), "/.aws")]
+        volumes = ["-v", "%s:%s" % (str(aws_path), os.path.sep+".aws")]
```
smurching, line 239:
```suggestion
        volumes = ["-v", "%s:%s" % (os.path.join(str(aws_path), ".aws"))]
```

mlflow/data.py
harupy: Looks like you passed `None` to `string`. Raw string is not the cause of the error.

Screen Shot 2020-08-07 at 2 59 23

https://github.com/mlflow/mlflow/runs/951541560#step:5:162

mlflow/utils/file_utils.py
```diff
@@ -384,6 +389,8 @@ def local_file_uri_to_path(uri):
     No-op if the uri does not have the expected scheme.
     """
     path = urllib.parse.urlparse(uri).path if uri.startswith("file:") else uri
+    if os.name == 'nt' and not path:
```
smurching, line 392: To reduce my own ignorance :P do you have an example (maybe one of our tests fails without this?) that illustrates the need for this change? What sort of input do we pass to this function that requires special handling for Windows / why does the solution in https://stackoverflow.com/a/43925228/12111347 (which seems to be what we had before) not work?

tests/projects/test_docker_projects.py
```diff
@@ -130,10 +131,15 @@ def test_docker_invalid_project_backend_local():
 
 
 @pytest.mark.parametrize("artifact_uri, host_artifact_uri, container_artifact_uri, should_mount", [
-    ("/tmp/mlruns/artifacts", "/tmp/mlruns/artifacts", "/tmp/mlruns/artifacts", True),
+    (os.path.join(os.path.sep, "tmp", "mlruns", "artifacts"),
```
smurching, line 134:
```suggestion
    os.path.abspath(os.path.join("tmp", "mlruns", "artifacts")),
```

To construct an absolute path here and elsewhere (vs trying to do it by prefixing the path with the path-separator)

tests/projects/test_docker_projects.py
```diff
@@ -147,8 +153,13 @@ def test_docker_mount_local_artifact_uri(artifact_uri, host_artifact_uri,
     image.tags = ["image:tag"]
 
     docker_command = _get_docker_command(image, active_run)
-
-    docker_volume_expected = "-v {}:{}".format(host_artifact_uri, container_artifact_uri)
+    if os.name == 'nt' and (host_artifact_uri is not None) and \
+            (not is_valid_windows_path(host_artifact_uri)):
```
smurching, line 157: Why do we need to check ``(host_artifact_uri is not None)` and `(not is_valid_windows_path(host_artifact_uri)`? It might be better at some point to refactor this into separate tests vs have this sort of branching logic in the test body, as it makes it hard to understand the test

tests/projects/utils.py
```diff
@@ -49,7 +49,12 @@ def docker_example_base_image():
         mlflow_dir = _copy_project(
             src_path=mlflow_home, dst_path=cwd)
         import shutil
-        shutil.copy(os.path.join(TEST_DOCKER_PROJECT_DIR, "Dockerfile"), tmp.path("Dockerfile"))
+        if os.name == "nt":
+            dockerfile = os.path.join(TEST_DOCKER_PROJECT_DIR, "windows", "Dockerfile")
+        else:
+            dockerfile = os.path.join(TEST_DOCKER_PROJECT_DIR, "Dockerfile")
+        # dockerfile = os.path.join(TEST_DOCKER_PROJECT_DIR, "Dockerfile")
```
smurching, line 56:
```suggestion
```

tests/projects/backend/test_local.py
```diff
@@ -69,7 +71,7 @@ def test_docker_hdfs_artifact_cmd_and_envs_from_home():
 
 def test_docker_local_artifact_cmd_and_envs():
     host_path_expected = os.path.abspath("./mlruns")
-    container_path_expected = "/mlflow/projects/code/mlruns"
+    container_path_expected = os.path.sep + os.path.join("mlflow", "projects", "code", "mlruns")
```
smurching, line 74: Same comment around using `os.path.abspath` here

mlflow/projects/utils.py
```diff
@@ -134,7 +134,8 @@ def _fetch_project(uri, version=None):
     if _is_zip_uri(parsed_uri):
         if _is_file_uri(parsed_uri):
             parsed_file_uri = urllib.parse.urlparse(urllib.parse.unquote(parsed_uri))
-            parsed_uri = os.path.join(parsed_file_uri.netloc, parsed_file_uri.path)
+            parsed_uri = parsed_file_uri.netloc if (os.name == "nt") \
```
smurching, line 137: Would you be able to share an example that illustrates the need for this check?

mlflow/projects/utils.py
```diff
@@ -86,7 +86,7 @@ def _is_file_uri(uri):
 
 def _is_local_uri(uri):
     """Returns True if the passed-in URI should be interpreted as a path on the local filesystem."""
-    return not _GIT_URI_REGEX.match(uri)
+    return not _GIT_URI_REGEX.match(uri) or (os.name == 'nt' and re.match('^[a-zA-Z]:\\\\.*', uri))
```
smurching, line 89: Can we call `is_valid_windows_path` here?

smurching: Seems surprisingly hard to do this correctly...see https://stackoverflow.com/a/34102855/12111347

At least in our case, it seems we always use `_is_local_uri` to check whether a string is a file URI (`file:///Users/apurva/...`) or a local path (`/Users/apurva/...`) _that actually exists_ (lmk if I missed something!). Given that we expect the file to exist in these cases, maybe we can actually remove this function and update usages of it to check `os.path.isfile(uri) or os.path.isdir(uri)` etc. The one downside is that if you pass a nonexistent local path (e.g. `/Users/not-apurva/...`) we may attempt to treat it as a remote URI & git clone it etc, resulting in an odd error message - we may just want to audit what error messages show up and ensure they're clear.

smurching (commented):
Thanks @apurva-koti! Leaving some comments

mlflow/data.py
apurva-koti: Good catch, thanks!

mlflow/utils/file_utils.py
apurva-koti: image

Here is one example:

I believe `urlparse` incorrectly parses the windows path as an entire `netloc` instead of a `path`.

dbczumar: Windows `\`-style paths aren't valid URIs. Afaik, the guidance for windows users has been to use file:/// URIs instead of backslashed paths with MLflow. I may be mistaken, though

PR #3231, opened 2020-08-05
Trigger auto-labeling action when new issue or PR has been opened

harupy:

GitHub has introduced a new workflow event `pull_request_target` ([link](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/)) where we can run actions with read/write permission. This PR udpates the auto-labeling action to get triggered when a new issue or PR has been opened.

## How is this patch tested?

Verified on my auto-labeling repo: https://github.com/harupy/auto-labeling

Components 

Interface 

Language 

Integrations

### DIFF
--- a/.github/workflows/labeling.yml
+++ b/.github/workflows/labeling.yml
@@ -4,9 +4,11 @@ on:
   issues:
     types:
       - opened
+      - edited
   pull_request_target:
     types:
       - opened
+      - edited
 
 jobs:
   labeling:
--- a/.github/workflows/labeling.yml
+++ b/.github/workflows/labeling.yml
@@ -1,9 +1,12 @@
 name: Labeling
 
 on:
-  schedule:
-    # Run this action daily at 7:00 UTC
-    - cron: "0 7 * * *"
+  issues:
+    types:
+      - opened
+  pull_request_target:
+    types:
+      - opened
 
 jobs:
   labeling:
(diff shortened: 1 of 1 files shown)


.github/workflows/labeling.yml
```diff
@@ -14,5 +19,3 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           label-pattern: '- \[(.*?)\] ?`(.+?)`'
-          # Label issues from last week
-          offset: '7d'
```
aarondav, line 18: What's the default offset?

.github/workflows/labeling.yml
```diff
@@ -1,9 +1,14 @@
 name: Labeling
 
 on:
-  schedule:
```
aarondav, line 4: Should we also keep the schedule? Just in case no one opens an issue or PR in a day?

.github/workflows/labeling.yml
harupy: The default offset is one month, but this argument is only valid for the schedule event.

aarondav: Ah, I see. So this will just cause autolabelling happen only as things are opened or edited. Is that sufficient to capture all events that would cause us to change labels?

.github/workflows/labeling.yml
harupy: I don't think we should because with both `issues` and `pull_request_target` events enabled, the action should be triggered right after a new issue or PR has been opened.

.github/workflows/labeling.yml
harupy: Yes because only `opened` and `edited` events updates an issue (or PR) description.

aarondav (approved):

PR #3220, opened 2020-08-03
MlflowClient should pass in the tracking uri info to ArtifactRepository

sueann:
## What changes are proposed in this pull request?

Previously, even if you specified the `tracking_uri` explicitly in initializing `MlflowClient`, its artifact methods (`log_artifact`, `log_artifacts`, `list_artifacts`, `download_artifacts`) did not talk to the correct tracking server if the artifacts were in DBFS. Now those methods instantiate an `ArtifactRepository` with the tracking URI to ensure that we talk to the correct Databricks workspace the DBFS location belongs to.

Also includes a few small fixes around dealing with different tracking and registry URIs:
* Allow uploading the files from the same run to a Databricks DBFS location multiple times in `artifact_utils._upload_artifacts_to_databricks`.
* In `MlflowClient.create_model_version`:
  * Change the messaging about model artifact copy to `eprint` (which seems to show up better in more contexts).
  * Support Databricks job context when constructing the `run_link`.

## How is this patch tested?

Manual tests - tried the artifact method with `tracking_uri` pointing to both local and remote contexts.

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [X] Yes. Give a description of this change to be included in the release notes for MLflow users.

Writing and reading artifacts via ``MlflowClient`` to a DBFS location in a Databricks tracking server specified through the ``tracking_uri`` parameter during the initialization of ``MlflowClient`` now works properly.

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [X] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [X] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
[...]

PR #3232, opened 2020-08-05
[Multiworkspace-Registry] use colon instead of slash to delimit scope and prefix

ankit-db:

Switch databricks://profile/prefix -> databricks://profile:prefix

## How is this patch tested?

Unit tests and testing notebook within Databricks

Components 

Interface 

Language 

Integrations


mlflow/utils/uri.py
```diff
@@ -42,9 +42,15 @@ def get_db_info_from_uri(uri):
     """
     parsed_uri = urllib.parse.urlparse(uri)
     if parsed_uri.scheme == "databricks":
-        parsed_path = parsed_uri.path.lstrip('/') or None
-        parsed_profile = parsed_uri.netloc
-        return parsed_profile, parsed_path
+        profile_tokens = parsed_uri.netloc.split(':')
+        parsed_profile = profile_tokens[0]
+        if len(profile_tokens) == 1:
```
ankitmathur-db, line 47: There are shorter possible ways to write this in code (the `len(profile_tokens) == 1` case is actually not necessary, since `:.join(..)` will not include a `:` if the length of the join list is 1), but I think it may be better to just be explicit about what's happening.

mlflow/utils/uri.py
```diff
@@ -93,12 +98,15 @@ def add_databricks_profile_info_to_artifact_uri(artifact_uri, databricks_profile
...
             (profile, key_prefix) = get_db_info_from_uri(databricks_profile_uri)
-            if ':' in profile:
-                raise MlflowException("Unsupported Databricks profile name: %s." % profile +
-                                      " Profile names cannot contain ':'.")
-            if key_prefix and '/' in key_prefix:
-                raise MlflowException("Unsupported Databricks profile key prefix: %s."
-                                      % key_prefix + " Key prefixes cannot contain '/'.")
+            # Both profile and key_prefix should not contain special chars for URIs, like '/'
+            # and ':'.
+            for c in ['/', ':']:
```
ankitmathur-db, line 103: I don't think it's actually possible given the current implementation to get a `/` into either of these, since that would mean that it falls into the `path` when being parsed by the URI parser, but logically, I think it's probably still true that from a validation perspective, we want to make sure neither of those chars happens to exist.

tests/utils/test_uri.py
```diff
@@ -355,7 +356,8 @@ def test_add_databricks_profile_info_to_artifact_uri(artifact_uri, profile_uri,
 
 @pytest.mark.parametrize("artifact_uri, profile_uri", [
     ('dbfs:/path/a/b', 'databricks://not:legit:auth'),
-    ('dbfs:/path/a/b/', 'databricks://scope/key/'),
```
ankitmathur-db, line 358: ^In this scenario, we will just interpret the "scope" component and ignore the rest, so it no longer throws.

mlflow/utils/uri.py
sueann: Why do we need to enforce that ':' is not part of the prefix? It looks like actually you cannot have either character in the secret scope name, but the key name is not as restrictive: https://docs.databricks.com/security/secrets/secret-scopes.html#overview (also we don't enforce the same restriction in other uri functions e.g. `get_databricks_profile_uri_from_artifact_uri`)

sueann (commented):
looks mostly good to me - the main comment is on the restriction on the prefix name

mlflow/utils/databricks_utils.py
```diff
@@ -177,7 +177,7 @@ def get_databricks_host_creds(server_uri=None):
     uses the Databricks CLI's ConfigProvider interface to load the DatabricksConfig object.
     If no Databricks CLI profile is found corresponding to the server URI, this function
     will attempt to retrieve these credentials from the Databricks Secret Manager. For that to work,
-    the server URI will need to be of the following format: "databricks://profile/prefix". In the
+    the server URI will need to be of the following format: "databricks://profile:prefix". In the
     Databricks Secret Manager, we will query for a secret in the scope "<profile>" for secrets with
```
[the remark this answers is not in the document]
ankitmathur-db: Good point - I do think that'd be clearer. Will update

mlflow/utils/uri.py
ankitmathur-db: The logic for this was that I came across scenarios like this:

Original DBFS path: 'dbfs:/path/a/b', Databricks Profile URI: 'databricks://not:legit:auth` which would result in dbfs:/not:legit:auth@databricks. 

You pointed out though, that, if we document then the scope isn't allowed to have a `:`, then the only possible way to parse this would be to include the second `:` in the second portion (the prefix)

ankitmathur-db:
@sueann added the validation in the `get_db_info_from_uri` function - it felt like the cleanest way to validate everywhere without adding the validator code in every function. One question I wanted to get feedback on: should we only do this validation if `key_prefix` is defined? That way, we change nothing at all about existing workflows

tests/utils/test_uri.py
```diff
@@ -277,14 +278,14 @@ def test_is_databricks_acled_artifacts_uri():
     ('dbfs://@databricks:port/path/to/nowhere', 'databricks'),
     ('dbfs://profile@databricks/path/to/nowhere', 'databricks://profile'),
     ('dbfs://profile@databricks:port/path/to/nowhere', 'databricks://profile'),
-    ('dbfs://scope:key_prefix@databricks/path/abc', 'databricks://scope/key_prefix'),
-    ('dbfs://scope:key_prefix@databricks:port/path/abc', 'databricks://scope/key_prefix'),
+    ('dbfs://scope:key_prefix@databricks/path/abc', 'databricks://scope:key_prefix'),
+    ('dbfs://scope:key_prefix@databricks:port/path/abc', 'databricks://scope:key_prefix'),
     # Treats secret key prefixes with ":" to be valid
-    ('dbfs://incorrect:netloc:format@databricks/path/a', 'databricks://incorrect/netloc:format'),
+    ('dbfs://incorrect:netloc:format@databricks/path/a', 'databricks://incorrect:netloc:format'),
```
sueann, line 284: we don't want to allow this anymore, right?

sueann: i think for `test_remove_databricks_profile_info_from_artifact_uri` it's okay to remove the info in this invalid case.

sueann: also should add the invalid case to `test_is_valid_dbfs_uri` below

tests/utils/test_uri.py
```diff
@@ -325,7 +326,7 @@ def test_remove_databricks_profile_info_from_artifact_uri(uri, result):
     ('dbfs:/path/a/b/', 'databricks://', 'dbfs://@databricks/path/a/b/'),
     ('dbfs:/path/a/b/', 'databricks://Profile', 'dbfs://Profile@databricks/path/a/b/'),
     ('dbfs:/path/a/b/', 'databricks://profile/', 'dbfs://profile@databricks/path/a/b/'),
-    ('dbfs:/path/a/b/', 'databricks://scope/key', 'dbfs://scope:key@databricks/path/a/b/'),
+    ('dbfs:/path/a/b/', 'databricks://scope:key', 'dbfs://scope:key@databricks/path/a/b/'),
     ('dbfs:/path/a/b/', 'nondatabricks://profile', 'dbfs:/path/a/b/'),
```
sueann, line 330: can we add the test case `'databricks://scope:key/random_stuff'` here? we will just ignore "random_stuff" right?

sueann (commented):
> One question I wanted to get feedback on: should we only do this validation if key_prefix is defined? That way, we change nothing at all about existing workflows

The way we are parsing in `get_db_info_from_uri`, I think we are already breaking the existing workflows that contain ":" in the profile name. Since we're making the assumption that a profile can't have ":" in the artifact URI code, I think it's cleanest to mark this PR as a breaking change and make sure we note it in the release notes. 


I've added comments on other places we should enforce this restriction. One big one I didn't comment inline on is `is_databricks_uri` -- it could make sense to add it there, but since it's not `is_valid_databricks_uri` we could skip it. Whatever flow is using that method is likely to hit one of the places we're enforcing the restriction here.

tests/utils/test_uri.py
ankitmathur-db: Agreed - I think the remove stuff probably doesn't need the validation

sueann:
Cool! lgtm - i merged #3220 though so i think you need to fix that one thing.

ankitmathur-db:
@sueann just fixed thaT - do you think you could approve the PR?

sueann (approved):
yep, lgtm - thanks!

PR #3263, opened 2020-08-13
Add `SklearnTrainigSession` for sklearn autologging

harupy:

Add `SklearnTrainigSession` for sklearn autologging. 

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/sklearn.py
+++ b/mlflow/sklearn.py
@@ -440,6 +440,33 @@ class _SklearnTrainingSession(object):
         :param clazz: A class object that this session originates from.
         :param allow_children: If True, allows autologging in child sessions.
                                If False, disallows autologging in all descendant sessions.
+
+        Example:
+
+        >>> class Parent: pass
+        >>> class Child: pass
+        >>> class Grandchild: pass
+
+        >>> with _SklearnTrainingSession(Parent, False) as p:
+        ...     with _SklearnTrainingSession(Child, True) as c:
+        ...         with _SklearnTrainingSession(Grandchild, True) as g:
+        ...             print(p.should_log())
+        ...             print(c.should_log())
+        ...             print(g.should_log())
+        True
+        False
+        False
+
+        >>> with _SklearnTrainingSession(Parent, True) as p:
+        ...     with _SklearnTrainingSession(Child, False) as c:
+        ...         with _SklearnTrainingSession(Grandchild, True) as g:
+        ...             print(p.should_log())
+        ...             print(c.should_log())
+        ...             print(g.should_log())
+        True
+        True
+        False
+
         """
         self.allow_children = allow_children
         self.clazz = clazz
--- a/mlflow/sklearn.py
+++ b/mlflow/sklearn.py
@@ -426,10 +426,21 @@ def load_model(model_uri):
     )
 
 
+# NOTE: The current implementation doesn't guarantee thread-safety, but that's okay for now because:
+# 1. We don't currently have any use cases for allow_children=True.
+# 2. The list append & pop operations are thread-safe, so we will always clear the session stack
+#    once all _SklearnTrainingSessions exit.
 class _SklearnTrainingSession(object):
     _session_stack = []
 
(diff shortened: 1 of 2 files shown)


tests/sklearn/test_sklearn_training_session.py
```diff
@@ -0,0 +1,50 @@
+import pytest
```
harupy, line 1: Created a file that focuses on `_SklearnTrainingSession` to simplify test function names

harupy: I'll add more tests

dbczumar: Looks great! Thanks so much. On the subject of additional testing, it'd be awesome to include a case verifying that, if we create a session with `allow_children=False` and a nested session with `allow_children=True`, additional nested sessions should still not log (i.e., the top-level session's `allow_children=True` should override the second level's `allow_children=False`).

mlflow/sklearn.py
```diff
@@ -424,3 +424,29 @@ def load_model(model_uri):
     return _load_model_from_local_file(
         path=sklearn_model_artifacts_path, serialization_format=serialization_format
     )
+
+
+class _SklearnTrainingSession(object):
```
dbczumar, line 429: Can we document the class constructor with information about the `clazz` and `allow_children` parameters & also document the `should_log()` method with information about the behavior?

dbczumar, line 429: Can we include some information about the lack of thread safety guarantees for this class and why that's okay for now? I.e., 1. we don't currently have any use cases for `allow_children=True` and 2. the list `append` & `pop` operations are thread-safe, so we will always clear the session stack once all `SklearnTrainingSessions` exit.

In the future, we'll need to figure out how to handle multithreaded backends.

mlflow/sklearn.py
```diff
@@ -424,3 +424,46 @@ def load_model(model_uri):
...
+    _session_stack = []
+
+    def __init__(self, clazz, allow_children=True):
+        """
+        A session manager for nested autologging runs.
+
+        :param clazz: A class object that this session originates from.
+        :param allow_children: If True, allows autologging in child sessions.
+                               If False, disallows autologging in all descendant sessions.
```
harupy, line 442: @dbczumar What do you think about providing some examples here like below?

```python

>>> with _SklearnTrainingSession(Parent, False) as p:
...     with _SklearnTrainingSession(Child, True) as c:
...         with _SklearnTrainingSession(Grandchild, True) as g:
...             print(p.should_log())  # -> True
...             print(c.should_log())  # -> False
...             print(g.should_log())  # -> False

>>> with _SklearnTrainingSession(Parent, True) as p:
...    with _SklearnTrainingSession(Child, False) as c:
...        with _SklearnTrainingSession(Grandchild, True) as g:
...            print(p.should_log())  # -> True
...            print(c.should_log())  # -> True
...            print(g.should_log())  # -> False
```

dbczumar: Sure! An example like this seems valuable to include

dbczumar (approved):
LGTM! (pending optional inclusion of the example described in https://github.com/mlflow/mlflow/pull/3263/files#r470774807). Thanks @harupy!

mlflow/sklearn.py
```diff
@@ -424,3 +424,73 @@ def load_model(model_uri):
...
+
+        >>> with _SklearnTrainingSession(Parent, False) as p:
+        ...     with _SklearnTrainingSession(Child, True) as c:
+        ...         with _SklearnTrainingSession(Grandchild, True) as g:
+        ...             print(p.should_log())
+        ...             print(c.should_log())
+        ...             print(g.should_log())
+        True
+        False
+        False
```
harupy, line 458: Used doctest syntax here so we can test this example by running:

```
pytest --doctest-modules mlflow/sklearn.py
```

PR #3282, opened 2020-08-16
make log_model with save_format=tf possible

t-lou:

mlflow uses "model.h5" as temp file for saving model, but then it will always be saved as hdf5 (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/saving/save.py line 120), even if save_format='tf' is given. Then subclass model cannot be saved.

## How is this patch tested?

(Details)

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -217,6 +217,9 @@ def save_model(
     # save keras model to path/data/model.h5
     model_subpath = os.path.join(data_subpath, _MODEL_SAVE_PATH)
     model_path = os.path.join(path, model_subpath)
+    # if the path ends with ".h5", save will use hdf5
+    if "save_format" in kwargs and kwargs["save_format"] == "tf" and model_path.endswith(".tf"):
+        model_path = model_path[:-3]
     if path.startswith("/dbfs/"):
         # The Databricks Filesystem uses a FUSE implementation that does not support
         # random writes. It causes an error.


mlflow/keras.py
```diff
@@ -217,6 +217,9 @@ def _is_tf_keras(model):
     # save keras model to path/data/model.h5
     model_subpath = os.path.join(data_subpath, _MODEL_SAVE_PATH)
     model_path = os.path.join(path, model_subpath)
+    # if the path ends with ".h5", save will use hdf5
+    if "save_format" in kwargs and kwargs["save_format"] == "tf" and model_path.endswith(".tf"):
```
skylarbpayne, line 221: Since `_MODEL_SAVE_PATH` wasn't modified, the model path will actually end with `.h5` and never `.tf`, so the body of this if statement can never be triggered, right?

mlflow/keras.py
```diff
@@ -217,6 +217,9 @@ def _is_tf_keras(model):
...
+        model_path = model_path[:-3]
```
skylarbpayne, line 222: Right now this PR only seems to address how to save the model; what happens when we later try to load this model? If I understand correctly, the load path will still look for the file with `.h5` suffix; which the file would _not_ have if it were successfully saved with `save_format='tf'`

We probably also want to make sure that this change is backwards compatible (i.e. so that models with the `.h5` suffix can be loaded successfully with changes)

harupy:
Close this RP since #3552 has addressed #3324.

PR #3287, opened 2020-08-16
Add autologging for scikit-learn

harupy:

Add autologging for scikit-learn

## How is this patch tested?

unite tests

Components 

Interface 

Language 

Integrations


mlflow/sklearn.py
```diff
@@ -500,3 +503,67 @@ def should_log(self):
...
+
+    patch_settings = gorilla.Settings(allow_hit=True, store_hit=True)
+    func_names_to_patch = ["fit", "fit_transform", "fit_predict"]
+    for _, class_def in sklearn.utils.all_estimators():
+        for func_name in func_names_to_patch:
+            if hasattr(class_def, func_name):
+                original = getattr(class_def, func_name)
+                patch_func = create_patch_func(func_name)
+                # preserve original function attributes
+                patch_func = functools.wraps(original)(patch_func)
```
harupy, line 567: Used `functools.wraps` to preserve the original function attributes.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -0,0 +1,183 @@
...
+    mlflow.sklearn.autolog()
+
+    model = sklearn.cluster.KMeans()
+    Xy = get_iris()
+
+    with mock.patch("mlflow.log_params") as mock_log_params, mock.patch(
+        "mlflow.log_metric"
+    ) as mock_log_metric, mock.patch("mlflow.set_tags") as mock_set_tags, mock.patch(
+        "mlflow.sklearn.log_model"
+    ) as mock_log_model:
```
harupy, line 152: This looks a bit ugly but is equivalent:

```
    with mock.patch("mlflow.log_params") as mock_log_params, \
            mock.patch("mlflow.log_metric") as mock_log_metric, \
            mock.patch("mlflow.set_tags") as mock_set_tags, \ 
            mock.patch("mlflow.sklearn.log_model") as mock_log_model:
```

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -0,0 +1,183 @@
...
+    return mlflow.sklearn.load_model("runs:/{}/model".format(run_id))
+
+
+def stringify_dict_values(d):
+    return {k: str(v) for k, v in d.items()}
+
+
+@pytest.fixture(params=["fit", "fit_transform", "fit_predict"])
+def fit_func_name(request):
+    return request.param
```
harupy, line 54: Parametrized fixture to DRY the code a bit.

mlflow/sklearn.py
```diff
@@ -500,3 +503,67 @@ def should_log(self):
         return (self._parent is None) or (
             self._parent.allow_children and self._parent.clazz != self.clazz
         )
+
+
+def autolog():
+    """
+    Enable autologging for scikit-learn.
```
harupy, line 510: I'll update the docstring.

mlflow/sklearn.py
```diff
@@ -500,3 +506,184 @@ def should_log(self):
...
         )
+
+
+def _is_old_version():
+    from sklearn import __version__
+
+    return LooseVersion(__version__) < LooseVersion("0.20.3")
+
+
+def all_estimators(type_filter=None):
```
harupy, line 517: maybe we should create a directory named `sklearn` and seprate code into multiple files otherwise `mlflow/sklearn.py` would become very large and harder to maintain.

mlflow/sklearn.py
```diff
@@ -8,10 +8,14 @@
 :py:mod:`mlflow.pyfunc`
```
harupy, line 8: maybe we should create a directory named `sklearn` and seprate code into multiple files otherwise `mlflow/sklearn.py` would become very large and harder to maintain.

dbczumar: +1. Let's do this!

mlflow/sklearn.py
```diff
@@ -500,3 +505,183 @@ def should_log(self):
...
+        )
+
+    def fit_mlflow(self, func_name, *args, **kwargs):
+        active_run_exists = bool(mlflow.active_run())
+        if not active_run_exists:
+            try_mlflow_log(mlflow.start_run)
+
+        # TODO: We should not log nested estimator parameters for
+        # parameter search estimators (GridSearchCV, RandomizedSearchCV)
+        try_mlflow_log(mlflow.log_params, self.get_params(deep=True))
```
dbczumar, line 640: In the rare / unlikely case where the model has more than 100 top-level parameter keys or produces more than 1MB in aggregate parameter output, this may fail due to limits of the underlying `Log Batch` API (https://mlflow.org/docs/latest/rest-api.html#log-batch). Can we handle this case by chunking the set of parameters?

dbczumar (changes requested):
@harupy This is awesome! Left some initial comments. Great work!

mlflow/sklearn.py
harupy: @corey Createa a separate PR to address this: https://github.com/mlflow/mlflow/pull/3294

mlflow/sklearn/__init__.py
```diff
@@ -500,3 +507,90 @@ def should_log(self):
...
+            },
+        )
+
+        original_fit = gorilla.get_original_attribute(self, func_name)
+        try:
+            fit_output = original_fit(*args, **kwargs)
+        except Exception as e:
+            _logger.warning("{} failed: {}".format(original_fit.__qualname__, str(e)))
+            mlflow.end_run(RunStatus.to_string(RunStatus.FAILED))
+            return
```
harupy, line 553: Should we raise an exception here?

dbczumar: Yep! We'll want to re-raise the scikit exception so the user can observe / handle it. We should add a test case for this behavior too

dbczumar: Also, we'll only want to end the run if the autologging routine started it. Otherwise, the following code won't behave as desired:

```
import mlflow.sklearn
mlflow.sklearn.autolog()

with mlflow.start_run():
    try:
        throwing_model.fit(X, y)
    catch Exception:
        # Do something else, run should continue...
```

harupy: got it! I'll fix the code

.github/workflows/sklearn.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          python-version: "3.6"
+      - name: Install dependencies
+        run: |
+          pip install -e .
+          pip install -r dev/small-requirements.txt
+          pip install scikit-learn==${{ matrix.version }}
+      - name: Run tests
+        run: |
+          pytest --verbose tests/sklearn/test_sklearn_autolog.py
```
harupy, line 30: Created this file to iterate faster and compare results between old and new sklearns. I'll delete this before merging.

harupy:
Maybe we should set `print_changed_only` to `False` otherwise `__repr__` would return a long string which might cause `log_batch` to fail due to the maximum param value length limit (250).

https://scikit-learn.org/stable/whats_new/v0.23.html#id13 says:

> The default setting `print_changed_only` has been changed from False to True. This means that the repr of estimators is now more concise and only shows the parameters whose default value has been changed when printing an estimator. You can restore the previous behaviour by using sklearn.set_config(print_changed_only=False)

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -0,0 +1,366 @@
...
+    with pytest.warns(
+        UserWarning, match="Autologging utilities may not work properly on scikit-learn"
+    ):
+        mlflow.sklearn.autolog()
+
+
+def test_autolog_does_not_terminate_active_run():
+    mlflow.sklearn.autolog()
+    mlflow.start_run()
+    sklearn.cluster.KMeans().fit(*get_iris())
```
harupy, line 105: After calling `sklearn.utils.all_estimators`, we can sklearn's submodules without importing them.

mlflow/sklearn/utils.py
```diff
@@ -0,0 +1,186 @@
...
+
+    return Xy
+
+
+def _chunk_dict(data, chunk_size=100):
+    # Copied from: https://stackoverflow.com/a/22878842
+
+    it = iter(data)
+    for _ in range(0, len(data), chunk_size):
+        yield {k: data[k] for k in islice(it, chunk_size)}
```
harupy, line 70: A function to chunk a dict by # of keys. I need to create another one to chunk by bytes size.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -0,0 +1,392 @@
...
+        ESTIMATOR_NAME: model.__class__.__name__,
+        ESTIMATOR_CLASS: model.__class__.__module__ + "." + model.__class__.__name__,
+    }
+    assert "model" in artifacts
+
+    loaded_model = load_model_by_run_id(run_id)
+    np.testing.assert_array_equal(loaded_model.predict(Xy[0]), model.predict(Xy[0]))
+
+
+def test_call_fit_with_arguments_score_does_not_accept():
```
dbczumar, line 212: Awesome! Can we also add a small test for the case where `X` and `y` are specified as kwargs?

dbczumar (commented):
@harupy This looks awesome! Left a few more small comments - looking forward to getting this merged!

mlflow/sklearn/__init__.py
```diff
@@ -500,3 +505,106 @@ def should_log(self):
         return (self._parent is None) or (
             self._parent.allow_children and self._parent.clazz != self.clazz
         )
+
+
+def autolog():
```
dbczumar, line 510: We should add the `@experimental` decorator to this function.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -0,0 +1,392 @@
...
+def stringify_dict_values(d):
+    return {k: str(v) for k, v in d.items()}
+
+
+@pytest.fixture(params=FIT_FUNC_NAMES)
+def fit_func_name(request):
+    return request.param
+
+
+def test_autolog_preserves_original_function_attributes():
```
dbczumar, line 64: Autologging tests should be decorated with `@pytest.mark.large`.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -0,0 +1,516 @@
...
+
+@pytest.fixture(autouse=True, scope="function")
+def throw_if_try_mlflow_log_has_emitted_warnings():
+    # autolog contains multiple try_mlflow_log. They may hide unintentional errors (often caused by
+    # incorrect test settings) and allow tests that should not pass to pass (without us noticing).
+    # To prevent that, throw if try_mlflow_log has emitted warnings during test execution.
+    with warnings.catch_warnings(record=True) as ws:
+        yield
+        mlflow_warnings = [w for w in ws if str(w.message).startswith("Logging to MLflow failed")]
+        assert len(mlflow_warnings) == 0
```
harupy, line 98: A fixture to prevent `try_mlflow_log` from allowing tests that should not pass to pass.

mlflow/sklearn/utils.py
```diff
@@ -0,0 +1,238 @@
+from distutils.version import LooseVersion
+import inspect
+from itertools import islice
+import logging
+
+_logger = logging.getLogger(__name__)
+
+_SAMPLE_WEIGHT = "sample_weight"
+_MIN_SKLEARN_VERSION = "0.20.3"
```
harupy, line 9: Does the variable name `_MIN_SKLEARN_VERSION ` make sense?

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+@pytest.mark.xfail
+def test_throw_if_try_mlflow_log_has_emitted_warnings():
+    with mlflow.start_run():
+        # This line should emit a warning because a metric value must be a float value.
+        try_mlflow_log(mlflow.log_metric, "m", "m")
```
harupy, line 105: A test to ensure `throw_if_try_mlflow_log_has_emitted_warnings` works properly.

tests/sklearn/test_sklearn_autolog.py
harupy: Relying on a warning message doesn't seem to be a good approach though.

mlflow/sklearn/__init__.py
```diff
@@ -500,3 +506,118 @@ def should_log(self):
         return (self._parent is None) or (
             self._parent.allow_children and self._parent.clazz != self.clazz
         )
+
+
+@experimental
+def autolog():
```
harupy, line 512: @dbczumar This sounds like a follow-up task (because this PR is already gigantic), but we could emit a warning when a user has already imported estimators before enabling autolog.

```python
import warnings

from sklearn.linear_model import LinearRegression  # import estimator before enabling autolog

# mlflow.sklearn.autolog()

def is_estimator_class(obj):
    from sklearn.base import BaseEstimator

    return isinstance(obj, type) and (obj != BaseEstimator) and issubclass(obj, BaseEstimator)


estimators_imported = list(filter(is_estimator_class, globals().values()))

if len(estimators_imported):
    warnings.warn("< warning message to encourage users to enable autolog before importing estimators >")
```

dbczumar (approved):
@harupy Looks great - thanks so much! Left a few small comments that we should address prior to merging; with those addressed, this LGTM! 

I tried this out manually with a few different model types and confirmed that the expected tags / metrics / parameters are produced. I also confirmed that tests pass on `0.20.3` through `0.24.dev0` (dev version).

harupy:
I found that the doc for `sklearn.utils.all_estimators` says it does **NOT** return meta estimators, but it turns out this is a mistake (see [this issue report](https://github.com/scikit-learn/scikit-learn/issues/18229#issuecomment-678953092) that I filed).

https://scikit-learn.org/stable/modules/generated/sklearn.utils.all_estimators.html

> Get a list of all estimators from sklearn. This function crawls the module and gets all classes that inherit from BaseEstimator. Classes that are defined in test-modules are not included. **By default meta_estimators such as GridSearchCV are also not included**.

.github/workflows/sklearn.yml
```diff
@@ -0,0 +1,44 @@
...
+      matrix:
+        version:
+          # include this unsupported version to ensure autologging emits a warning properly.
+          - "0.20.2"
+          # unsupported versions
+          # --------------------
+          # supported versions
+          - "0.20.3"
+          - "0.21.2"
+          - "0.22.2"
```
harupy, line 29: Excluded `0.23.2` because the `flavors` tests run checks using it.

PR #3288, opened 2020-08-17
Expose model flavors under `mlflow` so users don't have to import them

harupy:

Expose model flavors under `mlflow` so users don't have to import them.

Before:

```python
import mlflow

mlflow.sklearn.log_model(...)  # raises an error without `import mlflow.sklearn`
```

After:

```python
import mlflow

mlflow.sklearn.log_model(...)  # works
```

## How is this patch tested?

Unit test

All model flavors are exposed under `mllfow`.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -183,7 +183,7 @@ jobs:
           source activate test-environment
           ./dev/run-python-flavor-tests.sh;
 
-  mlflow:
+  import:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -183,6 +183,20 @@ jobs:
           source activate test-environment
           ./dev/run-python-flavor-tests.sh;
 
+  mlflow:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
+      - name: Install mlflow
+        run: |
+          pip install -e .
+      - name: Verify mlflow can be imported without errors
+        run: |
+          python -c "import mlflow"
+
   tensorflow:
     runs-on: ubuntu-latest
     steps:
--- a/mlflow/__init__.py
+++ b/mlflow/__init__.py
@@ -1,3 +1,4 @@
+# pylint: disable=wrong-import-position
 """
 The ``mlflow`` module provides a high-level "fluent" API for starting and managing MLflow runs.
 For example:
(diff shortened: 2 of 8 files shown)


tomasatdatabricks (commented):
Will this require users to install dependencies (e.g. tensorflow) for all flavors to import mlflow?

harupy:
@tomasatdatabricks
No, they don't have to because all the import statements are hidden in functions.

harupy:
Our CI doesn't have a check that verifies `mlflow` still can be imported without flavor dependencies.

tomasatdatabricks:
Hmm, it looks like mlflow.tensorflow.py does import tensorflow and keras at the top of the module. So we probably need to fix that and add a test as you mentioned. Otherwise looks good!

harupy:
> mlflow.tensorflow.py does import tensorflow and keras at the top of the module.

Which lines?

tomasatdatabricks:
https://github.com/mlflow/mlflow/blob/master/mlflow/tensorflow.py#L25-L26
https://github.com/mlflow/mlflow/blob/master/mlflow/tensorflow.py#L29=
+ there are other suspicious imports such as  e.g. contextlib and gorilla?

harupy:
> https://github.com/mlflow/mlflow/blob/master/mlflow/tensorflow.py#L25
https://github.com/mlflow/mlflow/blob/master/mlflow/tensorflow.py#L29

These two import statements has been moved into functions.

harupy:
> there are other suspicious imports such as e.g. contextlib and gorilla

- contextlib should be fine because it's a standard library.
-

tomasatdatabricks:
@harupy Sorry I missed the tf changes! Yeah in that case it sounds good as long as we have a test.

Thanks for doing this!

.github/workflows/master.yml
```diff
@@ -183,6 +183,20 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
+      - name: Install mlflow
+        run: |
+          pip install -e .
+      - name: Verify mlflow can be imported without errors
+        run: |
+          python -c "import mlflow"
```
harupy, line 198: @tomasatdatabricks Added a new job to verify `mlflow` can be installed without errors.

tomasatdatabricks (approved):
lgtm!

.github/workflows/master.yml
tomasatdatabricks: Thanks!

harupy:
Hmm, the storage size for GitHub free account is only 500MB ([link](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions)). Our wheel size is about 15 MB so we could instantly reach this limit.

PR #3296, opened 2020-08-19
[Model Registry Search] Tags support for filter_string and order_by in searchRegisteredModels

zhidongqu-db:

Introducing tags support for filter_string ad order_by in searchRegisteredModels by extracting a vast amount of functions in `search_utils.py` and create new `search_runs_utils` and `search_models_utils` separately. Existing search logic in `search_registered_models` is also updated to allow tags as filter_string

## How is this patch tested?

Unit Tests, Manual Tests

This PR allows user to search registered models based on tags attached to it, user can also order search results with tag values. (e.g. filter_string = "name ILIKE '%neural net%' AND tag.algorithm = 'neural net'", order_by = ["name ASC", "tag.owner ASC"])

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -192,7 +192,7 @@ class SearchUtils(object):
 
         # search models may not provide table name, in that case default to attribute table
         # (e.g name = 'CNN' is equivalent to attribute.model = 'CNN')
-        column_only = not is_search_run and identifier.find(".") == -1
+        column_only = identifier.find(".") == -1
         entity_type, key = split_identifier(identifier, column_only)
         identifier = (
             cls._valid_run_entity_type(entity_type)
@@ -202,8 +202,9 @@ class SearchUtils(object):
         key = cls._trim_backticks(cls._strip_quotes(key))
         if identifier == cls._ATTRIBUTE_IDENTIFIER and key not in valid_attributes:
             raise MlflowException(
-                "Invalid attribute key '{}' specified. Valid keys "
-                " are '{}'".format(key, valid_attributes)
+                "Invalid attribute key '{}' specified. Valid keys"
+                " are {}".format(key, valid_attributes),
+                error_code=INVALID_PARAMETER_VALUE,
             )
         return {"type": identifier, "key": key}
 
@@ -361,7 +362,8 @@ class SearchUtils(object):
             if comparator not in cls.VALID_TAG_COMPARATORS:
                 raise MlflowException(
                     "Invalid comparator '%s' "
-                    "not one of '%s" % (comparator, cls.VALID_TAG_COMPARATORS)
+                    "not one of '%s" % (comparator, cls.VALID_TAG_COMPARATORS),
+                    error_code=INVALID_PARAMETER_VALUE,
                 )
             return True
         return False
@@ -372,7 +374,8 @@ class SearchUtils(object):
             if comparator not in cls.VALID_STRING_ATTRIBUTE_COMPARATORS:
                 raise MlflowException(
                     "Invalid comparator '{}' not one of "
-                    "'{}".format(comparator, cls.VALID_STRING_ATTRIBUTE_COMPARATORS)
+                    "'{}".format(comparator, cls.VALID_STRING_ATTRIBUTE_COMPARATORS),
+                    error_code=INVALID_PARAMETER_VALUE,
                 )
             return True
         return False
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -174,36 +174,49 @@ class SearchUtils(object):
             return entity_type
 
     @classmethod
-    def _get_identifier(cls, identifier, valid_attributes, is_search_run):
-        def split_identifier(full_column_name, missing_table_name):
-            if missing_table_name:
-                return "attribute", full_column_name
-            else:
-                try:
-                    table_name, column_name = full_column_name.split(".", 1)
-                    return table_name, column_name
-                except ValueError:
-                    raise MlflowException(
-                        "Invalid identifier '%s'. Columns should be specified as "
-                        "'attribute.<key>', 'metric.<key>', 'tag.<key>', or "
-                        "'param.'." % full_column_name,
-                        error_code=INVALID_PARAMETER_VALUE,
-                    )
+    def _split_identifier(cls, full_column_name, is_missing_table_name):
+        if is_missing_table_name:
+            return "attribute", full_column_name
+        else:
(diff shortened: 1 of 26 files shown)


tests/utils/test_search_utils.py
```diff
@@ -172,11 +299,32 @@ def test_error_comparison_clauses(filter_string, error_message):
...
         ("attribute.status IS 'RUNNING'", "Invalid clause(s) in filter string"),
+        ("status IS 'RUNNING'", "Invalid clause(s) in filter string"),
     ],
 )
-def test_bad_quotes(filter_string, error_message):
+def test_bad_quotes_for_run(filter_string, error_message):
     with pytest.raises(MlflowException) as e:
-        SearchUtils.parse_search_filter(filter_string)
+        SearchUtils.parse_filter_for_run(filter_string)
+    assert error_message in e.value.message
```
harupy, line 308: Hmm, this is out of the scope of this PR, but we should use the `match` argument for `pytest.raises` ([ref](https://docs.pytest.org/en/stable/reference.html#pytest-raises)) instead of `assert error_message in e.value.message`.

mlflow/utils/search_utils.py
```diff
@@ -430,20 +491,212 @@ def _parse_order_by_string(cls, order_by):
...
+                return SqlRegisteredModelTag
+            elif cls.is_attribute(key_type, comparator):
+                return None
+            else:
+                raise MlflowException(
+                    "Invalid identifier type '%s'" % key_type, error_code=INVALID_PARAMETER_VALUE,
+                )
+
+    @classmethod
+    def _get_order_by_clauses(cls, order_by_list, session, is_search_run):
```
dbczumar, line 537: As with the other backend PR, I think we should try to avoid these boolean flag-based inputs / behavioral modifiers for utilities. Rather than providing a `_get_order_by_clauses` method that is parameterized by the `is_search_run` boolean, I'd rather we extract out common logic for order by construction into one or more utility methods and then invoke these utility methods from `get_order_by_clauses_for_run` and `get_order_by_clauses_for_registered_model`. This will make it easier to understand what's going on (right now, the control flow of this method is complex / hard to reason about for various input cases).

This same comment applies to other utility methods in this module that fork behavior based on `is_search_run`; if we have to, I think it's better to duplicate a bit of code than create aggregate routines with lots of branching logic / complex control flow.

mlflow/utils/search_utils.py
```diff
@@ -430,20 +491,212 @@ def _parse_order_by_string(cls, order_by):
...
+            ) not in observed_order_by_clauses:
+                clauses.append(SqlRegisteredModel.name.asc())
+        return clauses, ordering_joins
+
+    @classmethod
+    def get_order_by_clauses_for_run(cls, order_by_list, session):
+        return cls._get_order_by_clauses(order_by_list, session, True)
+
+    @classmethod
+    def get_order_by_clause_for_registered_model(cls, order_by_list, session):
```
dbczumar, line 622: Nit:

```suggestion
    def get_order_by_clauses_for_registered_model(cls, order_by_list, session):
```

mlflow/utils/search_utils.py
```diff
@@ -5,16 +5,24 @@
...
+    SqlTag,
+    SqlLatestMetric,
+)
+from mlflow.store.model_registry.dbmodels.models import (
+    SqlRegisteredModel,
+    SqlRegisteredModelTag,
+)
 
 
 class SearchUtils(object):
```
dbczumar, line 28: Over all, it seems like this module is aggregating together a bunch of routines from the SQLAlchemy stores for model registry and tracking into methods of the form `my_search_util(*args, is_search_runs)`. I think it would be more readable / easier to reason about if we left logic specific to the model registry & tracking in the appropriate store classes and extracted out common pieces of logic / utility functions used by both of them into `search_utils`. Let me know if you have questions here (see my comments below for additional reasoning).

mlflow/utils/search_utils.py
```diff
@@ -31,40 +39,51 @@ class SearchUtils(object):
     VALID_REGISTERED_MODEL_SEARCH_COMPARATORS = CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS.union(
         {"="}
     )
-    VALID_SEARCH_ATTRIBUTE_KEYS = set(RunInfo.get_searchable_attributes())
-    VALID_ORDER_BY_ATTRIBUTE_KEYS = set(RunInfo.get_orderable_attributes())
+    VALID_SEARCH_RUN_ATTRIBUTE_KEYS = set(RunInfo.get_searchable_attributes())
```
dbczumar, line 42: Can we split out run-specific constants and registry-specific constants into separate classes? perhaps `RunsSearchUtils` and `ModelRegistrySearchUtils`?

dbczumar (changes requested):
@zhidongqu-db Awesome feature! I left a few comments about the utility refactoring. If it's too difficult to extract out common logic without branching, etc right now, I think it's perfectly acceptable to repeat a couple of routines across the model registry and tracking stores for now & follow up with refactoring later. Let me know what you think!

dbczumar (commented):
@zhidongqu-db This looks great! Left a few inline comments. I also have feedback about error specificity that's very similar to the feedback left on the other backend PR.

On `master`, specifying an invalid `order_by` column explicitly mentions `order_by`:

```
client.search_registered_models("name LIKE 'model_%'", order_by=["invalid"])

RestException: INVALID_PARAMETER_VALUE: Invalid order by key 'invalid' specified. Valid keys are '{'name', 'timestamp'}'
```

However, on the PR branch, the error message becomes less specific (`order_by` is not mentioned):

```
client.search_registered_models("name LIKE 'model_%'", order_by=["invalid"])

RestException: INVALID_PARAMETER_VALUE: Invalid attribute key 'invalid' specified. Valid keys are {'timestamp', 'name', 'last_updated_timestamp'}
```

Can we ensure that our error messages specify where a particular problem occurred? (E.g., in the `order_by` or `filter` clause?)

ZeroFuture:
lost github account `zhidongqu-db`, a new pull request with the exact same change and comments addressed is filed in https://github.com/mlflow/mlflow/pull/3320

ZeroFuture:
Original PR had some problem with git sign-off and rebasing, new PR filed at https://github.com/mlflow/mlflow/pull/3296

ZeroFuture:
Continue the work left with PR https://github.com/mlflow/mlflow/pull/3296

WeichenXu123:
@dbczumar Note that we haven't support order_by clauses for searching model / model version.

Issue #3173, opened 2020-07-27
[DOC-FIX] Description of filter argument syntax for MlflowClient.search_registered_models is missing


amesar:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for information on what types of issues we address.

**Please fill in this documentation issue template to ensure a timely and thorough response.**

### Willingness to contribute
The MLflow Community encourages documentation fix contributions. Would you or another member of your organization be willing to contribute a fix for this documentation issue to the MLflow code base?

- [ ] Yes. I can contribute a documentation fix independently.
- [ ] Yes. I would be willing to contribute a document fix with guidance from the MLflow community.
- [ x] No. I cannot contribute a documentation fix at this time.

### URL(s) with the issue:

https://mlflow.org/docs/latest/python_api/mlflow.tracking.html#mlflow.tracking.MlflowClient.search_registered_models

### Description of proposal (what needs changing):

We need a description of the syntax that the filter_string argument of MlflowClient.search_registered_models accepts. For example, search_registered_models and search_model_versions do provide descriptions.

Currently search_registered_models simply says "filter_string – Filter query string, defaults to searching all registered models."

dmatrix:
@amesar thanks for filling this. We'll work on providing code snippets and descriptions, not only to newly added APIs but to all APIs. This is an ongoing effort.

PR #3322, opened 2020-08-25
Added filter_string descripition and code snippet; address issue #3173

dmatrix:
Signed-off-by: Jules Damji <dmatrix@comcast.net>

Add filter_description for `search_registered_model` method and added code snippets how to use the API
This address issue #3173 as part of documentation fix.

Components 

- [x ] `area/docs`: MLflow documentation pages


mlflow/tracking/client.py
```diff
@@ -469,6 +471,43 @@ def search_registered_models(
...
                 that satisfy the search expressions. The pagination token for the next page can be
                 obtained via the ``token`` attribute of the object.
+
+        .. code-block:: python
+            :caption: Example
+
+            import mlflow
+
+            model_name="CordobaWeatherForecastModel"
+            filter = "name='{}'".format(model_name)
```
harupy, line 481:
```suggestion
            filter_string = "name='{}'".format(model_name)
```

because `filter` conflicts with python's built-in `filter` function.

mlflow/tracking/client.py
```diff
@@ -469,6 +471,43 @@ def search_registered_models(
...
+            print("-" * 80)
+            for res in results:
+                for mv in res.latest_versions:
+                    print("name={}; run_id={}; version={}".format(mv.name, mv.run_id, mv.version))
+            # Get all registered models and order them by ascending order of the names
+            results = client.search_registered_models(order_by=["name ASC"])
+            print("-" * 80)
+            for res in results:
+                for mv in res.latest_versions:
+                    print("name={}; run_id={}; version={}".format(mv.name, mv.run_id, mv.version))
```
harupy, line 495: Should we add examples for `LIKE` and `ILIKE` operators?

mlflow/tracking/client.py
dmatrix: good point. I'll change it!

mlflow/tracking/client.py
dmatrix: I'm going to do that with `search_runs()` as part of another PR that covers fluent `mflow` APIs. [Search runs](https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.search_runs) will have these `LIKE`and `ILIKE` operators.  For this issue, I quickly wanted to address it. 

Both these APIs are experimental; we might also add 'tag' too.

harupy: > I belive these APIs, for now, only support name= operators. If you use LIKE such as query = "name LIKE '%Azure%'", you'll get an exception.

I think `search_registered_models` supports `LIKE` and `ILIKE` operators. 

```python
import mlflow

# mkdir -p /tmp/mlflow/
mlflow.set_tracking_uri("sqlite:////tmp/mlflow/db")


class Model(mlflow.pyfunc.PythonModel):
    pass


mlflow.pyfunc.log_model("model", python_model=Model(), registered_model_name="model")
print(client.search_registered_models("name LIKE 'mode%'"))
```

`search_model_versions` doesn't though: https://github.com/mlflow/mlflow/search?q=%22only+supports+equality%22&unscoped_q=%22only+supports+equality%22

harupy: `search_model_versions` doesn't:

https://github.com/mlflow/mlflow/blob/3936705d9a41ada8e526c7981bb0a119862d53c3/mlflow/store/model_registry/sqlalchemy_store.py#L748-L752

dmatrix: Excellent point @harupy! Let me add that operator to `search_registered_models`. I think it's good to put that as en example
in the code snippet for `search_registered_models`.

mlflow/tracking/client.py
```diff
@@ -665,10 +704,41 @@ def search_model_versions(self, filter_string):
...
+            # Get all versions of the model filtered by name
+            results = client.search_model_versions(filter_name)
+            print("-" * 80)
+            for res in results:
+                print("name={}; run_id={}; version={}".format(res.name, res.run_id, res.version))
+            # Get the version of the model filtered by run_id
+            results = client.search_model_versions(filter_runid)
+            print("-" * 80)
+            for res in results:
+                print("name={}; run_id={}; version={}".format(res.name, res.run_id, res.version))
```
harupy, line 732:
```suggestion
            client = mlflow.tracking.MlflowClient()

            # Get all versions of the model filtered by name
            model_name = "CordobaWeatherForecastModel"
            filter_name = "name='{}'".format(model_name)
            results = client.search_model_versions(filter_name)
            print("-" * 80)
            for res in results:
                print("name={}; run_id={}; version={}".format(res.name, res.run_id, res.version))

            # Get the version of the model filtered by run_id
            run_id = "e14afa2f47a040728060c1699968fd43"
            filter_runid = "run_id='{}'".format(run_id)
            results = client.search_model_versions(filter_runid)
            print("-" * 80)
            for res in results:
                print("name={}; run_id={}; version={}".format(res.name, res.run_id, res.version))
```

harupy: Just make the code more readable

dmatrix: Indeed, makes it readable. Will change it.

mlflow/tracking/client.py
dmatrix: Done!

mlflow/tracking/client.py
dmatrix: done and committed all feedback.

mlflow/tracking/client.py
```diff
@@ -460,7 +460,11 @@ def search_registered_models(
         """
         Search for registered models in backend that satisfy the filter criteria.
 
-        :param filter_string: Filter query string, defaults to searching all registered models.
+        :param filter_string: Filter query string, defaults to searching all registered
+                models. Currently, it supports only a single filter condition as the name
+                of the model, for example, ``name = 'model_name'`` or a search expression
+                to match a pattern in the registered model name.
+                For example, ``name LIKE 'Boston%'`` or ``name ILIKE '%Forecast%'``.
```
harupy, line 467:
```suggestion
                For example, ``name LIKE 'Boston%'`` (case sensitive) or ``name ILIKE '%Forecast%'`` (case insensitive).
```

I think users who are not familiar with SQL would not understand what `ILIKE` means.

dmatrix: Indeed, make it more specific, as the API does not describe it. I'll add this too.

harupy:
The r check failed:  https://github.com/mlflow/mlflow/pull/3322/checks?check_run_id=1034060279#step:8:1068

harupy (approved):
LGTM!

dmatrix:
@harupy Can't figure out why the tests/r are failing?

harupy:
Merging the master branch should solve the issue because #3344 fixed it.

dmatrix:
@harupy Thanks for fixing the #3344; it seems to have fixed the problems, and all tests have passed. You can merge at your convenience.

harupy:
Merged!

PR #3326, opened 2020-08-25
Log model signature and input example in sklearn autologging

harupy:

Log model signature and input example in sklearn autologging

## How is this patch tested?

Unit tests.

Components 

Interface 

Language 

Integrations


mlflow/sklearn/__init__.py
```diff
@@ -668,7 +671,25 @@ def fit_mlflow(self, func_name, *args, **kwargs):
             else:
                 try_mlflow_log(mlflow.log_metric, "training_score", training_score)
 
-        try_mlflow_log(log_model, self, artifact_path="model")
+        SAMPLE_ROWS = 5
+        fit_arg_names = _get_arg_names(self.fit)
+        X_var_name, y_var_name = fit_arg_names[:2]
+        X_sample = _get_Xy(args, kwargs, X_var_name, y_var_name)[0][:SAMPLE_ROWS]
```
harupy, line 677: How many rows should we sample here?

dbczumar: I think that 5 rows makes sense to me! Can we try/catch this block on the off chance that slice-based sampling of the data fails? I know that sklearn indicates that data should be array-like, but it'd be nice to cover the unlikely case where it isn't.

mlflow/sklearn/__init__.py
```diff
@@ -668,7 +671,25 @@ def fit_mlflow(self, func_name, *args, **kwargs):
...
+        try:
+            model_output = self.predict(X_sample) if hasattr(self, "predict") else None
```
dbczumar, line 679: There are a couple of interesting questions that come to mind here:

1. If the model doesn't have a `predict()` method, should we still be recording the input portion of the inference schema? (I.e., if `predict()` isn't defined, then what inference routine / function is compatible with the stated inputs?)

2. Should we also consider attempting to fetch outputs for models that define `transform()` but not `predict()`? I.e. perhaps it makes sense to do the following:

- Check if `predict()` is defined. If so, output the schema based on `predict()`
- Else, check if `transform()` is defined. If so, output the schema based on `transform()`
- Else, do not output a schema (see point no.1)

This has the benefit of providing schema information for a wider breadth of models. However, there are a couple of downsides to consider: A. when serving / scoring sklearn models as pyfunc, we assume that the `predict()` method is defined and attempt to invoke it; if we output schemas for models that have `transform()` defined but not `predict()`, that may lead to confusing behavior ("why can't I perform pyfunc inference on my model? you claim it has an inference schema!"). B. Given a schema, users need to reason about which method it refers to (it refers to `predict()` if defined, else to `transform()`). Let me know what your thoughts are here. @tomasatdatabricks it would also be great to get your input here!

3. Can we handle the case where models define `fit_predict()` but do not define `predict()`? Interestingly, that seems to apply to a few estimators, such as: https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html#sklearn.cluster.AgglomerativeClustering.fit_predict

dbczumar (commented):
@harupy This is great! I left a couple of comments - the latter one probably requires some thought / discussion. Let me know if you'd like to discuss details via VC.

mlflow/sklearn/__init__.py
tomasatdatabricks: @dbczumar Those are good questions!  And a cool PR @harupy!

We do not really support models without predict method and we should probably not output PyFunc flavor for those at all until we fix it. If we fix it, I think we should map PyFunc.predict to the relevant method (e.g. transform) on the underlying model. I don't think we want to adapt distinction between predict and transform. 

So my take on this is, ideally we want the following: 
1) find out what should PyFunc.predict call on the model. If we don't find what to map to, do not output pyfunc at all.
2) use the method from step 1 to generate the output schema.

dbczumar: Makes perfect sense. Thanks @tomasatdatabricks !

harupy: makes sense! thanks @tomasatdatabricks

mlflow/sklearn/__init__.py
```diff
@@ -668,7 +674,50 @@ def fit_mlflow(self, func_name, *args, **kwargs):
...
+        # Example: https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html  # NOQA
+        # For models that don't define `predict`, use `fit_predict` to get the model prediction.
+        has_predict = hasattr(self, "predict") or hasattr(self, "fit_predict")
+
+        signature = None
+
+        if (X_sample is not None) and has_predict:
+            # If this function is `fit_predict`, reuse its output (which should be a numpy array).
+            if func_name == "fit_predict":
+                model_output = fit_output[:SAMPLE_ROWS]
```
harupy, line 700: Reuse the output of `fit_predict`.

mlflow/sklearn/__init__.py
```diff
@@ -668,7 +674,50 @@ def fit_mlflow(self, func_name, *args, **kwargs):
...
+            else:
+                try:
+                    model_output = (
+                        self.predict(X_sample)
+                        if hasattr(self, "predict")
+                        # use `clone` to prevent re-fitting
+                        else clone(self).fit_predict(X_sample)
```
harupy, line 707: Use `sklearn.base.clone` to prevent re-fitting if we have to use `fit_predict`.

dbczumar: If `hasattr(self, "predict")` is False, then, based on the previous discussion with @tomasatdatabricks, we should not output the pyfunc flavor or a model signature. Accordingly, I think we can get rid of this `else` case.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -672,3 +682,55 @@ def test_autolog_does_not_throw_when_mlflow_logging_fails(func_to_fail):
...
+    with mlflow.start_run() as run:
+        model.fit(X, y)
+        model_path = os.path.join(run.info.artifact_uri, MODEL_DIR)
+
+    model_conf = get_model_conf(run.info.artifact_uri)
+    input_example = _read_example(model_conf, model_path)
+    pyfunc_model = mlflow.pyfunc.load_model(model_path)
+
+    assert model_conf.signature == infer_signature(X, model.predict(X[:5]))
+    np.testing.assert_array_equal(pyfunc_model.predict(input_example), model.predict(X[:5]))
```
dbczumar, line 705: @harupy I tweaked this test case slightly to verify that the input example can be used to evaluate the pyfunc representation of the model. The example can be loaded as a dataframe using the `mlflow.models.utils._read_example` utility function.

mlflow/sklearn/__init__.py
```diff
@@ -720,7 +726,26 @@ def _log_posttraining_metadata(estimator, *args, **kwargs):
             else:
                 try_mlflow_log(mlflow.log_metric, "training_score", training_score)
 
-        try_mlflow_log(log_model, estimator, artifact_path="model")
+
+        input_example = None
+        signature = None
+        if hasattr(estimator, "predict"):
```
dbczumar, line 732: @harupy I simplified the logic here to encapsulate the entire row sampling + signature inference procedure into a single try/catch block. While the error messages may be a bit less granular, the key point users should be aware of is that the input example and signature are not logged in the event of an error; this structure makes that a bit clearer, and it improves the readability of the code.

mlflow/sklearn/__init__.py
```diff
@@ -188,12 +188,14 @@ def save_model(
     with open(os.path.join(path, conda_env_subpath), "w") as f:
         yaml.safe_dump(conda_env, stream=f, default_flow_style=False)
 
-    pyfunc.add_to_model(
-        mlflow_model,
-        loader_module="mlflow.sklearn",
-        model_path=model_data_subpath,
-        env=conda_env_subpath,
-    )
+    # `PyFuncModel` only works for sklearn models that define `predict()`.
```
dbczumar, line 191: Made a minor tweak to this comment to remove the double negative

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -672,3 +682,55 @@ def test_autolog_does_not_throw_when_mlflow_logging_fails(func_to_fail):
...
+
+
+def test_autolog_does_not_throw_when_failing_to_sample_X():
+    class ArrayThatThrowsWhenSliced(np.ndarray):
+        def __new__(cls, input_array):
+            return np.asarray(input_array).view(cls)
+
+        def __getitem__(self, key):
+            if isinstance(key, slice):
+                raise IndexError("DO NOT SLICE ME")
```
dbczumar, line 715: Love it!

dbczumar (approved):
LGTM! I pushed a couple small changes to the branch (comments inline) and would like to merge with your approval of these tweaks, @harupy. Let me know if things look good to you!

mlflow/sklearn/__init__.py
harupy: makes sense

tests/sklearn/test_sklearn_autolog.py
harupy: makes sense! Thanks!

tests/sklearn/test_sklearn_autolog.py
harupy: probably the best class I've ever written.

mlflow/sklearn/__init__.py
```diff
@@ -720,7 +726,30 @@ def _log_posttraining_metadata(estimator, *args, **kwargs):
...
+                # training data supplied to the training routine (e.g., `fit()`)
+                SAMPLE_ROWS = 5
+                fit_arg_names = _get_arg_names(estimator.fit)
+                X_var_name, y_var_name = fit_arg_names[:2]
+                input_example = _get_Xy(args, kwargs, X_var_name, y_var_name)[0][:SAMPLE_ROWS]
+
+                model_output = estimator.predict(input_example)
+                signature = infer_signature(input_example, model_output)
+            except Exception as e:
+                input_example = None
```
harupy, line 743: Added this line to prevent logging `input_example` when  `estimator.predict` or `infer_signature` fails.

dbczumar:
LGTM! Thanks @harupy !

PR #3327, opened 2020-08-25
Add addition metrics for sklearn auto logging

willzhan-db:

auto-log common metrics for classifiers & regressors

For classifier:
(1) accuracy_score
(2) log_loss (for classifier with **predict_proba** method)
(3) f1_score
(4) precision_score
(5) recall_score
(6) roc_auc_score (for classifier with **predict_proba** method, and with Sklearn version above **0.22.2**)

For regressor:
(1) mean_squared_error
(2) root_mean_squared_error
(3) r2_score
(4) mean_absolute_error

## How is this patch tested?

Through full coverage of unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -531,6 +531,56 @@ def autolog():
       **Metrics**
         - A training score obtained by ``estimator.score``. Note that the training score is
           computed using parameters given to ``fit()``.
+        - Common metrics for classifier:
+
+          - `precision score`_
+
+          .. _precision score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html
+
+          - `recall score`_
+
+          .. _recall score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.recall_score.html
+
+          - `f1 score`_
+
+          .. _f1 score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html
+
+          - `accuracy score`_
+
+          .. _accuracy score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html
+
+          If the classifier has method ``predict_proba``, we additionally log:
+
+          - `log loss`_
+
+          .. _log loss:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html
+
+          - `roc auc score`_
+
+          .. _roc auc score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html
+
+        - Common metrics for regressor:
+
+          - `(root) mean squared error`_
+
+          .. _(root) mean squared error:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html
+
+          - `mean absolute error`_
+
+          .. _mean absolute error:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_absolute_error.html
+
+          - `r2 score`_
+
+          .. _r2 score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html
 
       **Tags**
         - An estimator class name (e.g. "LinearRegression").
@@ -626,6 +676,7 @@ def autolog():
         _is_supported_version,
         _chunk_dict,
         _get_args_for_score,
+        _log_specialized_estimator_content,
         _get_Xy,
         _all_estimators,
         _truncate_dict,
@@ -730,6 +781,9 @@ def autolog():
             else:
                 try_mlflow_log(mlflow.log_metric, "training_score", training_score)
 
+        # log common metrics and artifacts for estimators (classifier, regressor)
+        _log_specialized_estimator_content(estimator, mlflow.active_run().info.run_id, args, kwargs)
+
         input_example = None
         signature = None
         if hasattr(estimator, "predict"):
@@ -829,6 +883,7 @@ def autolog():
     estimators_to_patch = set(estimators_to_patch).union(
         set(_get_meta_estimators_for_autologging())
     )
+
     for class_def in estimators_to_patch:
         for func_name in ["fit", "fit_transform", "fit_predict"]:
             if hasattr(class_def, func_name):
--- a/mlflow/sklearn/utils.py
+++ b/mlflow/sklearn/utils.py
@@ -1,8 +1,10 @@
+import collections
 from distutils.version import LooseVersion
-import inspect
 from itertools import islice
+import inspect
 import logging
 from numbers import Number
+import numpy as np
 import time
 
 from mlflow.entities import Metric, Param
@@ -24,7 +26,12 @@ _logger = logging.getLogger(__name__)
 # on scikit-learn older than this version.
 _MIN_SKLEARN_VERSION = "0.20.3"
 
+_METRICS_PREFIX = "training_"
 _SAMPLE_WEIGHT = "sample_weight"
+
+# _SklearnMetric represents a metric (e.g, precision_score) that will be computed and logged
+# during the autologging routine for a particular model type (eg, classifier, regressor).
+_SklearnMetric = collections.namedtuple("_SklearnMetric", ["name", "function", "arguments"])
 
 
 def _get_estimator_info_tags(estimator):
@@ -49,6 +56,16 @@ def _get_Xy(args, kwargs, X_var_name, y_var_name):
 
     # corresponds to: model.fit(<X_var_name>=X, <y_var_name>=y)
     return kwargs[X_var_name], kwargs[y_var_name]
+
+
+def _get_samples_labels_and_predictions(fitted_estimator, fit_args, fit_kwargs, fit_arg_names):
+    # In most cases, X_var_name and y_var_name become "X" and "y", respectively.
+    # However, certain sklearn models use different variable names for X and y.
+    X_var_name, y_var_name = fit_arg_names[:2]
+    X, y_true = _get_Xy(fit_args, fit_kwargs, X_var_name, y_var_name)
+    y_pred = fitted_estimator.predict(X)
+
+    return X, y_true, y_pred
 
 
 def _get_sample_weight(arg_names, args, kwargs):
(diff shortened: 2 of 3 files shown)


mlflow/sklearn/__init__.py
```diff
@@ -670,6 +676,10 @@ def fit_mlflow(self, func_name, *args, **kwargs):
 
         try_mlflow_log(log_model, self, artifact_path="model")
 
+        # If the estimator is of classifier type, then log corresponding metrics
+        if self._estimator_type == _CLASSIFIER:
```
dbczumar, line 680: According to the sklearn docs on this subject (https://scikit-learn.org/stable/developers/develop.html#estimator-types), we should use utility methods, such as `sklearn.base.is_classifier`, to identify different types of models when possible:

```
... When a meta-estimator needs to distinguish among estimator types, instead of checking _estimator_type directly, helpers like base.is_classifier should be used.
```

Under the hood, these methods do seem to check the `_estimator_type` attribute, but that might change. In the case of clustering models, which don't have a nice `sklearn.base.*` method, we can still check for this attribute.

Finally, we should handle the case where `_estimator_type` is not defined by using `getattr` instead (see https://github.com/scikit-learn/scikit-learn/blob/2fa4272b1deee0ac605e173445659fc5fec293ef/sklearn/base.py#L777).

dbczumar (changes requested):
@willzhan-db This is a great start! Left a few comments. Let me know if you have questions!

mlflow/sklearn/utils.py
```diff
@@ -1,7 +1,11 @@
-from distutils.version import LooseVersion
 import inspect
-from itertools import islice
 import logging
+import mlflow
+import sklearn
```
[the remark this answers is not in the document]
harupy: FYI: We have the check (added by #3288) to assert that `import mlflow` works with its default depedencies: https://github.com/mlflow/mlflow/blob/70b61e309a2a4755d616ea3dac4f506917e98439/.github/workflows/master.yml#L187-L199

Currently, this check doesn't fail because `import ... from mlflow.sklearn.utils` is executed in the `autolog` function, but as @dbczumar suggested, we should import these optional dependencies in functions.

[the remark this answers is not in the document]
willzhan-db: This is really a good catch! Thanks for the suggestion!

mlflow/sklearn/utils.py
```diff
@@ -77,6 +94,219 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+    import sklearn
+
+    print("In Clusterer!")
+    fit_arg_names = _get_arg_names(trained_estimator.fit)
+    X, y_true, y_pred = _get_fit_Xy(trained_estimator, fit_args, fit_kwargs, fit_arg_names)
+
+    # Maintain a metrics dictionary to store metrics info
+    # name_args_func_metrics_dict stores pairs of (function name, (function arguments, function call))
+    metrics_dict_clusterer = {
+        "completeness_score": ((y_true, y_pred), sklearn.metrics.completeness_score),
```
dbczumar, line 253: For many clustering models, `y_true` (which corresponds to the `y` argument of `fit()`) is an unused placeholder value. For example, see https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html#sklearn.cluster.KMeans.fit_predict.

Additionally, the number of classes reflected by `y_true` does not necessarily correspond to the number of clusters. I.e. there may be fewer clusters than labels or more clusters than labels.

Accordingly, I'm not sure that `y_true` is the correct value to pass in here. Is there a different value that should be used instead?

mlflow/sklearn/utils.py
```diff
@@ -25,6 +32,16 @@ def _get_Xy(args, kwargs, X_var_name, y_var_name):
     return kwargs[X_var_name], kwargs[y_var_name]
 
 
+def _get_fit_Xy(trained_estimator, fit_args, fit_kwargs, fit_arg_names):
+    # In most cases, X_var_name and y_var_name become "X" and "y", respectively.
+    # However, certain sklearn models use different variable names for X and y.
+    X_var_name, y_var_name = fit_arg_names[:2]
+    X, y_true = _get_Xy(fit_args, fit_kwargs, X_var_name, y_var_name)
+    y_pred = trained_estimator.predict(X)
```
dbczumar, line 40: Certain clustering models don't define `predict()`. For example: https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html#sklearn.cluster.SpectralClustering.fit.

We should probably handle the case where `predict()` is undefined. Let's talk about this tomorrow. In conjunction with my other comment about clustering down below (where `y_true` may not be the correct input for most of these metrics functions), our best option may be to forgo logging clustering metrics for now and focus solely on regression / classification.

dbczumar (changes requested):
@willzhan-db Awesome progress! I left some comments - let me know if you have questions!

mlflow/sklearn/utils.py
```diff
@@ -77,6 +94,219 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+    """
+    import sklearn
+
+    fit_arg_names = _get_arg_names(trained_estimator.fit)
+    X, y_true, y_pred = _get_fit_Xy(trained_estimator, fit_args, fit_kwargs, fit_arg_names)
+
+    # Maintain a metrics dictionary to store metrics info
+    # name_args_func_metrics_dict stores pairs of (function name, (function arguments, function call))
+    metrics_dict_classifier = {
+        "precision_score": ((y_true, y_pred, None, 1, "weighted"), sklearn.metrics.precision_score),
```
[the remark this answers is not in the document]
willzhan-db: Actually I have been thinking of a more readable way but cannot find one (not aware of this namedtuple template). Thanks for your suggestion!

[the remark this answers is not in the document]
willzhan-db: Before what I thought is that we have the whole control of how the arguments are ordered in the monkey patch, but I haven't thought about the future update might change the arguments ordering (and even the arguments' names) : (
Should have considered a more flexible way for the function call: )
Thanks for this reminder!

mlflow/sklearn/utils.py
```diff
@@ -77,6 +94,219 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+    import sklearn
+
+    name_score_dict = {}
+    if sklearn.base.is_classifier(trained_estimator):
+        name_score_dict = _get_classifier_metrics(trained_estimator, fit_args, fit_kwargs)
+    elif sklearn.base.is_regressor(trained_estimator):
+        name_score_dict = _get_regressor_metrics(trained_estimator, fit_args, fit_kwargs)
+    elif (
+        hasattr(trained_estimator, "_estimator_type")
+        and trained_estimator._estimator_type == "clusterer"
```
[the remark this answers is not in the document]
willzhan-db: From your last comments, I will remove the implementation part for **clusterer** for now.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -24,6 +24,17 @@
...
+PRECISION_SCORE = "precision_score"
+RECALL_SCORE = "recall_score"
+F1_SCORE = "f1_score"
+MSE = "mse"
+RMSE = "rmse"
+MAE = "mae"
+COMPLETENESS_SCORE = "completeness_score"
+HOMOGENEITY_SCORE = "homogeneity_score"
+V_MEASURE_SCORE = "v_measure_score"
+R2_SCORE = "r2_score"
```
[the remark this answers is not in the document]
willzhan-db: Actually for now I cannot think of a case to reuse these specific metrics. I would prefer remove them for now, and in the future to create these global variables when necessary.

mlflow/sklearn/utils.py
willzhan-db: Sounds reasonable!

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -621,7 +698,7 @@ def test_parameter_search_handles_large_volume_of_metric_outputs():
     with mlflow.start_run() as run:
         svc = sklearn.svm.SVC()
         cv_model = sklearn.model_selection.GridSearchCV(
-            svc, {"C": [1]}, n_jobs=1, scoring=metrics_to_log, refit=False
+            svc, {"C": [1]}, n_jobs=1, scoring=metrics_to_log, refit="score_0"
```
willzhan-db, line 701: I made this modification in order to incorporate my added unit tests.
I found that **GridSearchCV**'s base class is **Classifier**, to it will execute the classifier metrics logging, where I will always use the passed in estimator as an already trained model and call **predict()**
If you set refit=False, then the estimator will not refit itself according to the metric you put. So here I just arbitrarily pick **refit=score_0** so that the model can get refit.

Another way to solve this case is we can guard the **GridSearchCV** from going into the code path for metric logging, but this will be based on whether we want to log common metrics for **GridSearchCV** or not.

mlflow/sklearn/utils.py
```diff
@@ -103,6 +117,211 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+    print("Exception name {0}".format(func_name))
+    _logger.warning(msg)
+
+
+def _log_specialized_estimator_content(trained_estimator, run_id, fit_args, fit_kwargs):
+    import sklearn
+
+    name_score_dict = {}
+    # test case: [test_parameter_search_handles_large_volume_of_metric_outputs] will fail
+    # As GridSearchCV turns out to be a "classifier" in the end
```
willzhan-db, line 308: I think we should guard the GridSearchCV case, but just left a comment for now. Will appreciate your input :)

mlflow/sklearn/utils.py
```diff
@@ -103,6 +118,248 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+                ),
+                _SklearnMetric(
+                    name="roc_auc_score",
+                    function=sklearn.metrics.roc_auc_score,
+                    arguments=dict(
+                        y_true=y_true,
+                        y_score=y_pred_proba,
+                        average="weighted",
+                        sample_weight=sample_weight,
+                        multi_class="ovo",
```
willzhan-db, line 233: This is the attribute that is not existed in deprecated Sklearn version

mlflow/sklearn/utils.py
```diff
@@ -103,6 +118,248 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+    )
+    sample_weight = (
+        _get_sample_weight(fit_arg_names, fit_args, fit_kwargs)
+        if _SAMPLE_WEIGHT in fit_arg_names
+        else None
+    )
+
+    classifier_metrics = [
+        _SklearnMetric(
+            name="precision_score",
```
harupy, line 186:
```suggestion
            name=sklearn.metrics.precision_score.__name__
```

Can you de-hardcode `name` to prevent typos?

mlflow/sklearn/utils.py
```diff
@@ -103,6 +118,248 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+                        multi_class="ovo",
+                    ),
+                ),
+            ]
+        )
+
+    return _get_metrics_value_dict(classifier_metrics)
+
+
+def _get_regressor_metrics(trained_estimator, fit_args, fit_kwargs):
```
harupy, line 242:
```suggestion
def _get_regressor_metrics(fitted_estmator, fit_args, fit_kwargs):
```

to match sklearn's terminology?

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -468,16 +528,38 @@ def throwing_score(X, y=None, sample_weight=None):  # pylint: disable=unused-arg
...
+    """
+    mlflow.sklearn.autolog()
+
+    model = sklearn.svm.SVC()
+
+    @functools.wraps(model.score)
+    def throwing_score(y_true, y_pred):  # pylint: disable=unused-argument
+        raise Exception("EXCEPTION")
+
+    sklearn.metrics.precision_score = throwing_score
```
harupy, line 553: The reason I patched `model.score` this way was to preserve its signature (for [this line](https://github.com/mlflow/mlflow/blob/master/mlflow/sklearn/utils.py#L90)). sorry I should have put a comment on this. If you don't have to preserve a signature, you can just use `mock.patch` with `side_effect`.

mlflow/sklearn/__init__.py
```diff
@@ -819,6 +823,14 @@ def f(self, *args, **kwargs):
     estimators_to_patch = set(estimators_to_patch).union(
         set(_get_meta_estimators_for_autologging())
     )
+
+    # from sklearn.utils.testing import all_estimators
```
dbczumar, line 827: Can we remove this commented code?

mlflow/sklearn/utils.py
```diff
@@ -26,6 +28,9 @@
 
 _SAMPLE_WEIGHT = "sample_weight"
 
+# A organized template to store sklearn metrics
```
dbczumar, line 31: Nit: I think we could be more specific with this comment. _SklearnMetric represents a metric that will be computed and logged during the autologging routine for a particular model type

mlflow/sklearn/utils.py
```diff
@@ -103,6 +118,248 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+            arguments=dict(
+                y_true=y_true,
+                y_pred=y_pred,
+                sample_weight=sample_weight,
+                multioutput="uniform_average",
+            ),
+        ),
+    ]
+
+    # To be compatibale with deprecated versions of scikit-learn (below 0.22.2), where
```
dbczumar, line 312: Nit: small typo in "compatible". Also, these are older versions of sklearn, not necessarily deprecated versions

mlflow/sklearn/utils.py
dbczumar: Can we check the sklearn version and only add this metric to the set of metrics to compute if we're running a version that supports it?

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -477,16 +537,38 @@ def throwing_score(X, y=None, sample_weight=None):  # pylint: disable=unused-arg
...
             "Scoring error: EXCEPTION"
         )
 
-    metrics = get_run_data(run.info.run_id)[1]
-    assert metrics == {}
+
+def test_autolog_emits_warning_message_when_metric_fails():
+    """
+    Take precision_score metric from SVC as an example to test metric logging failure
+    :return: NULL
```
dbczumar, line 552: Nit: remove return NULL if we're not returning a value

mlflow/sklearn/utils.py
```diff
@@ -103,6 +118,248 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+    )
+    _logger.warning(msg)
+
+
+def _log_specialized_estimator_content(fitted_estimator, run_id, fit_args, fit_kwargs):
+    import sklearn
+
+    name_score_dict = {}
+    try:
+        if sklearn.base.is_classifier(fitted_estimator):
```
dbczumar, line 337: Awesome! Can we add a test for the case when model prediction fails to ensure that we log a warning but don't throw an exception?

dbczumar (commented):
@willzhan-db this is looking really good! Left a few more small comments - looking forward to getting this merged!

mlflow/sklearn/utils.py
willzhan-db: This makes sense to me!

mlflow/sklearn/__init__.py
willzhan-db: I thought I have deleted it lol. Will be more careful next time!

mlflow/sklearn/utils.py
willzhan-db: Added a test:)

dbczumar (approved):
LGTM! Thanks @willzhan-db  - this is awesome!

mlflow/sklearn/utils.py
```diff
@@ -103,6 +119,253 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+            ),
+        ),
+        _SklearnMetric(
+            name="training_accuracy_score",
+            function=sklearn.metrics.accuracy_score,
+            arguments=dict(
+                y_true=y_true, y_pred=y_pred, normalize=True, sample_weight=sample_weight
+            ),
+        ),
+    ]
```
harupy, line 214: Thse lines can be simplifed to:

```python
metric_funcs = {
    sklearn.metrics.precision_score: {"average": "weighted"},
    sklearn.metrics.accuracy_score: {"average": "weighted"},
    # ...
}

classifier_metrics = [
    _SklearnMetric(
        name="training_{}".format(func.__name__),
        function=func,
        arguments=dict(y_true=y_true, y_pred=y_pred, sample_weight=sample_weight, **kwargs),
    )
    for func, kwargs in metric_funcs.items()
]
```

mlflow/sklearn/utils.py
```diff
@@ -302,6 +565,16 @@ def _is_supported_version():
     return LooseVersion(sklearn.__version__) >= LooseVersion(_MIN_SKLEARN_VERSION)
 
 
+# Util function to check whether a metric is able to be computed in given sklearn version
+def _is_metric_supported(metric_name):
+    import sklearn
+
+    # This dict can be extended to store special metrics' specific supported versions
+    _metric_supported_version = {"training_roc_auc_score": "0.22.2"}
```
harupy, line 573:
```suggestion
    _metric_supported_version = {"roc_auc_score": "0.22.2"}
```
Should we exclude `training_` becase `"training_roc_auc_score": "0.22.2"` souds like scikit-learn 0.22.2 has a metric function named `traning_roc_auc_score`?

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -192,7 +192,75 @@ def test_estimator(fit_func_name):
...
+    # logging of (1) log_loss and (2) roc_auc_score.
+    model = sklearn.ensemble.RandomForestClassifier(max_depth=2, random_state=0, n_estimators=10)
+    X, y_true = get_iris()
+
+    with mlflow.start_run() as run:
+        model = fit_model(model, X, y_true, "fit")
+
+    y_pred = model.predict(X)
+    y_pred_prob = model.predict_proba(X)
+    run_id = run._info.run_id
```
harupy, line 215:
```suggestion
    run_id = run.info.run_id
```
nit

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -192,7 +192,75 @@ def test_estimator(fit_func_name):
...
+    mlflow.sklearn.autolog()
+    # use simple `LinearRegression`, which only implements `fit`.
+    model = sklearn.linear_model.LinearRegression()
+    X, y_true = get_iris()
+
+    with mlflow.start_run() as run:
+        model = fit_model(model, X, y_true, "fit")
+
+    y_pred = model.predict(X)
+    run_id = run._info.run_id
```
harupy, line 253:
```suggestion
    run_id = run.info.run_id
```

nit

mlflow/sklearn/utils.py
```diff
@@ -103,6 +119,253 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+    )
+    sample_weight = (
+        _get_sample_weight(fit_arg_names, fit_args, fit_kwargs)
+        if _SAMPLE_WEIGHT in fit_arg_names
+        else None
+    )
+
+    classifier_metrics = [
+        _SklearnMetric(
+            name="training_precision_score",
```
harupy, line 187: Should we put this `training` prefix into a constant variable and reuse it?

```python
_METRICS_PREFIX = "training"
```

mlflow/sklearn/utils.py
```diff
@@ -103,6 +119,253 @@ def _get_args_for_score(score_func, fit_func, fit_args, fit_kwargs):
...
+def _get_regressor_metrics(fitted_estimator, fit_args, fit_kwargs):
+    """
+    Compute and log various common metrics for regressors
+
+    For (1) (root) mean squared error:
+    https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html#sklearn.metrics.mean_squared_error
+    (2) mean absolute error:
+    https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_absolute_error.html#sklearn.metrics.mean_absolute_error
+    (3) r2 score:
+    https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html#sklearn.metrics.r2_score
```
harupy, line 257:
```suggestion
    For (1) (root) mean squared error:
    https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html
    (2) mean absolute error:
    https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_absolute_error.html
    (3) r2 score:
    https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html
```

harupy: I think these links work without a fragment.

mlflow/sklearn/utils.py
willzhan-db: Makes sense!

mlflow/sklearn/utils.py
willzhan-db: I have actually though about simplify these lines, but I kinda feel this optimization is a bit overkill and impair the readability. As there are some special cases that y_pred=y_pred_proba.
If we only simplify the above common metrics, it seems a bit inconsistent though.
Listing like now looks pretty straightforward, and if in the future we want to add other metrics we may not need think about refactor the code (As you suggestion suggest that all functions always take in y_true=y_true, y_pred=y_pred, sample_weight=sample_weight,)

harupy: Makes sense :)

harupy: Makes sense :)

mlflow/sklearn/__init__.py
```diff
@@ -531,6 +531,13 @@ def autolog():
       **Metrics**
         - A training score obtained by ``estimator.score``. Note that the training score is
           computed using parameters given to ``fit()``.
+        - Common metrics for classifier:
+          (1) precision score. (2) recall score. (3) f1_score. (4) accuracy score.
+          If the classifier has method `predict_proba`, we additionally log:
+          (5) log loss. (6) roc_auc_score
```
harupy, line 537: Can we use a nested list here?

Screen Shot 2020-09-01 at 9 24 16

code
```
          - precision score
          - recall score
          - f1 score
          - accuracy score

          If the classifier has method ``predict_proba``, we additionally log:

          - log loss
          - roc auc score
```

harupy: and can we add hyperlinks to these metrics to clarify which function we use to compute them? For example,

```
- `presicion score`_

.. _precision score:
    https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html
```

mlflow/sklearn/__init__.py
```diff
@@ -531,6 +531,30 @@ def autolog():
...
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html
+          .. _accuracy score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html
+
+          If the classifier has method ``predict_proba``, we additionally log:
+
+          .. _log loss:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html
+          .. _roc auc score:
+              https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html
```
harupy, line 549: Thanks for the fix, but just defining link targets doesn't generate links.

Screen Shot 2020-09-01 at 16 46 00

https://927-136202695-gh.circle-artifacts.com/0/docs/build/html/python_api/mlflow.sklearn.html

harupy: You can verify the doc is rendered properly by following the instruction here: https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst#writing-docs

harupy: you also verify the doc on CircleCI:

Sep-01-2020 16-55-20

willzhan-db: Gotcha. Before I thought this is just a pure comment and haven't realized that this comment is for documentation lol.

harupy:
@willzhan-db great work!

PR #3341, opened 2020-08-27
Remove dependency on SQLAlchemy to import mlflow

smurching:

Removes need to have `sqlalchemy` installed in order to `import mlflow`, to enable installing MLflow & using in environments where SQLAlchemy is not installed

## How is this patch tested?

Manually - uninstalled `sqlalchemy` via `pip uninstall sqlalchemy` and verified I could `import mlflow` and log metrics/params/tags. Existing tests to verify that no existing functionality is broken by this change.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -14,8 +14,6 @@ from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
 
 import math
 
-from mlflow.store.model_registry.dbmodels.models import SqlRegisteredModel
-
 
 class SearchUtils(object):
     LIKE_OPERATOR = "LIKE"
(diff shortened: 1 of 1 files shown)


mlflow/utils/search_utils.py
```diff
@@ -54,7 +52,7 @@ class SearchUtils(object):
     # Registered Models Constants
     ORDER_BY_KEY_TIMESTAMP = "timestamp"
     ORDER_BY_KEY_LAST_UPDATED_TIMESTAMP = "last_updated_timestamp"
-    ORDER_BY_KEY_MODEL_NAME = SqlRegisteredModel.name.key
```
mparkhe, line 57: Can we do 2 things.

1. Move some of the column name constants to a lower level utils or stores package as `column_names.py` and share it for search utils and also SQL Alchemy stuff.
2. These this columns as names on SqlAlchemyStore data models using [name param in constructor](https://docs.sqlalchemy.org/en/13/core/metadata.html#sqlalchemy.schema.Column.__init__). For example, 

```
class SqlRegisteredModel(Base):
    __tablename__ = "registered_models"

    name = Column(String(256), unique=True, nullable=False)

    creation_time = Column(BigInteger, default=lambda: int(time.time() * 1000))

    ...
```

becomes

```
from mlflow.stores.db.column_names import *

class SqlRegisteredModel(Base):
    __tablename__ = "registered_models"

    name = Column(REGISTERED_MODEL_NAME, String(256), unique=True, nullable=False)

    creation_time = Column(REGISTERED_MODEL_CREATION_TIME, BigInteger, default=lambda: int(time.time() * 1000))
```

I think it would involve some work for registry and tracking data models and changes to test code. Also, I expect some changes to search utils. Does not need to block the release, but would be awesome to DRY these constants.

mparkhe (approved):
LGTM... one comment

PR #3345, opened 2020-08-27
Allow to skip TLS verification of S3 endpoint

dolfinus:
There is a `MLFLOW_TRACKING_INSECURE_TLS` and `MLFLOW_TRACKING_SERVER_CERT_PATH` for skipping TLS verification or passing custom ca bundle for certificates checking. This allows users to access their local MLflow instances with self-signed SSL certificates, and that's quite useful feature.

Underlying AWS client library boto3 already have option of getting value of env variable `AWS_CA_BUNDLE` (of corresponding config option) which can be used for passing custom ca bundle.

But there is no env variables for disabling SSL certificate check completely, like `MLFLOW_TRACKING_INSECURE_TLS`, but for artifact storage.

This pull request adds new env variable `MLFLOW_S3_IGNORE_TLS` with just the same meaning.

## How is this patch tested?

Added few mock items to existing test.

Allow to skip TLS verification of S3 endpoint.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/docs/source/tracking.rst
+++ b/docs/source/tracking.rst
@@ -622,7 +622,7 @@ For example, if you have a MinIO server at 1.2.3.4 on port 9000:
 
   export MLFLOW_S3_ENDPOINT_URL=http://1.2.3.4:9000
 
-If MinIO server is configured with using SSL self-signed or signed using some internal-only CA certificate, you could set ``MLFLOW_S3_IGNORE_TLS`` or ``AWS_CA_BUNDLE`` variables (not both at the same time!) to disable certificate signature check or add use some custom CA bundle to perform this check, respectively:
+If the MinIO server is configured with using SSL self-signed or signed using some internal-only CA certificate, you could set ``MLFLOW_S3_IGNORE_TLS`` or ``AWS_CA_BUNDLE`` variables (not both at the same time!) to disable certificate signature check, or add a custom CA bundle to perform this check, respectively:
 
 .. code-block:: bash
 
--- a/docs/source/tracking.rst
+++ b/docs/source/tracking.rst
@@ -622,11 +622,20 @@ For example, if you have a MinIO server at 1.2.3.4 on port 9000:
 
   export MLFLOW_S3_ENDPOINT_URL=http://1.2.3.4:9000
 
+If MinIO server is configured with using SSL self-signed or signed using some internal-only CA certificate, you could set ``MLFLOW_S3_IGNORE_TLS`` or ``AWS_CA_BUNDLE`` variables (not both at the same time!) to disable certificate signature check or add use some custom CA bundle to perform this check, respectively:
+
+.. code-block:: bash
+
+  export MLFLOW_S3_IGNORE_TLS=true
+  #or
+  export AWS_CA_BUNDLE=/some/ca/bundle.pem
+
 Additionally, if MinIO server is configured with non-default region, you should set ``AWS_DEFAULT_REGION`` variable:
 
 .. code-block:: bash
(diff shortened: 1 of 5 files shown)


dolfinus:
CI/CD error is connected with some R code which I didn't change.

AveshCSingh:
Thank you for this contribution, and sorry for the late response @dolfinus.

It would be great if these arguments could be generalized to any artifact store, rather than just S3. 

One suggestion of how to do this: Remove the `MLFLOW_S3_SERVER_CERT_PATH` argument, as I believe you can use the boto3 environment variable `AWS_CA_BUNDLE` instead ([link](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html)). 

As for `MLFLOW_S3_IGNORE_TLS`, could this generalize to `MLFLOW_ARTIFACT_STORE_INSECURE_TLS`? You don't need to implement this functionality for all the stores in this PR, but it would be good to check that the environment variable is sensible for most stores & add code to fail if it is set for stores where it has not yet been implemented.

What do you think? Please feel free to push back if this does not seem like the right thing to do.

dolfinus:
Hello.

According to documentation, supported artifacts storages are; 
-    Amazon S3 and S3-compatible storage
-    Azure Blob Storage
-    Google Cloud Storage
-    FTP server
-    SFTP Server
-    NFS
-    HDFS

Only items 1-3 are using HTTPS as transport. Others use other protocols like FTP ans so on, and TLS verification does not make any sense here.

Amazon AWS, Azure and Google Cloud are cloud platforms, and their SSL certificates cannot be self-signed because it'll ruin the whole infrastructure. So even if there is an option to disable TLS verification, it will not affect anything, because these certificates are always correct.

So only first item (S3-compatible storage) can be deployed on some server within internal organization network there self-signed SSL certificates are possible.

That's why I've named env variable `MLFLOW_S3_IGNORE_TLS`. Also I think that it's better to have separated env variables for each artifact storage to give users option of building more flexible configuration. Like setting that just one particular storage is using non-secure connection and others are completely secure.

Regarding using `AWS_CA_BUNDLE` instead of `MLFLOW_S3_SERVER_CERT_PATH`, thank you for mentioning this variable, I didn't know it is already exist. I've removed it from this PR.

docs/source/tracking.rst
```diff
@@ -622,12 +622,21 @@ For example, if you have a MinIO server at 1.2.3.4 on port 9000:
 
   export MLFLOW_S3_ENDPOINT_URL=http://1.2.3.4:9000
 
+If MinIO server is configured with using SSL self-signed or signed using some internal-only CA certificate, you could set ``MLFLOW_S3_IGNORE_TLS`` or ``AWS_CA_BUNDLE`` variables (not both at the same time!) to disable certificate signature check or add use some custom CA bundle to perform this check, respectively:
```
AveshCSingh, line 625:
```suggestion
If the MinIO server is configured with using SSL self-signed or signed using some internal-only CA certificate, you could set ``MLFLOW_S3_IGNORE_TLS`` or ``AWS_CA_BUNDLE`` variables (not both at the same time!) to disable certificate signature check, or add a custom CA bundle to perform this check, respectively:
```

AveshCSingh:
Thanks for explaining this, @dolfinus. This change LGTM.

AveshCSingh (approved):

dolfinus:
Thank you!

dolfinus:
It looks like this PR caused a bug #4047. @AveshCSingh could you please take a look on that?

PR #3369, opened 2020-08-31
Add additional artifacts for auto logging

willzhan-db:

Compute and log artifacts for classifier including:
(1) confusion matrix
(2) precision recall curve
(3) roc curve

## How is this patch tested?

Unit tests.

Components 

Interface 

Language 

Integrations


harupy:
@willzhan-db Can you rebase on master?

mlflow/sklearn/utils.py
```diff
@@ -126,16 +133,38 @@ def _get_metrics_value_dict(metrics_list):
...
         else:
             metric_value_dict[metric.name] = metric_value
-
     return metric_value_dict
 
 
+# Remove duplicates from y_true while still preserving orders
+def _get_labels(y_true):
+    seen = set()
+    return [x for x in y_true if not (x in seen or seen.add(x))]
```
harupy, line 145: why do we need to preserve labels order?

willzhan-db: My thought:
As we only take in X and Y from users, sometimes I need to check whether this is a binary classifier problem so that to log respective metrics. 
Get the non-duplicated labels can help me to decide whether this is a binary problem or not.

mlflow/sklearn/utils.py
```diff
@@ -174,81 +203,169 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
                         ),
+                        title="",
                     ),
                 ]
             )
 
     return _get_metrics_value_dict(classifier_metrics)
 
 
+def _get_classifier_artifacts(fitted_estimator, fit_args, fit_kwargs):
```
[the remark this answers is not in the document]
willzhan-db: Hum.
I think "artifacts" is actually what we use more frequently across databricks? (No matter in the mlflow documentation, or in the sklean design doc). I am totally fine to change it to "plots".
Ask about Corey's opinion :)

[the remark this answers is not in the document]
dbczumar: I think artifacts makes sense in this case, since the routine is meant to log MLflow artifacts & the term artifact is part of the MLflow lexicon.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -206,37 +206,52 @@ def test_classifier():
...
+        _TRAINING_PREFIX + "accuracy_score": sklearn.metrics.accuracy_score(y_true, y_pred),
+        _TRAINING_PREFIX
         + "precision_score": sklearn.metrics.precision_score(y_true, y_pred, average="weighted"),
-        _METRICS_PREFIX
+        _TRAINING_PREFIX
         + "recall_score": sklearn.metrics.recall_score(y_true, y_pred, average="weighted"),
-        _METRICS_PREFIX + "f1_score": sklearn.metrics.f1_score(y_true, y_pred, average="weighted"),
-        _METRICS_PREFIX + "log_loss": sklearn.metrics.log_loss(y_true, y_pred_prob),
+        _TRAINING_PREFIX + "f1_score": sklearn.metrics.f1_score(y_true, y_pred, average="weighted"),
+        _TRAINING_PREFIX + "log_loss": sklearn.metrics.log_loss(y_true, y_pred_prob),
```
dbczumar, line 233: On second thought, I'd rather we use the full strings `training_accuracy_score, `training_f1_score` in our test assertions (it's fine to continue using `_TRAINING_PREFIX` in the actual implementation). Otherwise, someone could come along and change `_TRAINING_PREFIX` to `will_is_awesome`; these tests would still pass, but users would see strange metrics of the form `will_is_awesome_accuracy_score`.

mlflow/sklearn/utils.py
```diff
@@ -342,15 +465,17 @@ def _log_specialized_estimator_content(fitted_estimator, run_id, fit_args, fit_k
     import sklearn
 
     name_metric_dict = {}
+    artifact_dir = None
     try:
         if sklearn.base.is_classifier(fitted_estimator):
             name_metric_dict = _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs)
+            artifact_dir = _get_classifier_artifacts(fitted_estimator, fit_args, fit_kwargs)
```
dbczumar, line 472: If we fail to log metrics, can't we still try to log artifacts & vice versa? It seems like these should be separate try/catch blocks.

mlflow/sklearn/utils.py
```diff
@@ -58,14 +66,13 @@ def _get_Xy(args, kwargs, X_var_name, y_var_name):
...
 
-def _get_samples_labels_and_predictions(fitted_estimator, fit_args, fit_kwargs, fit_arg_names):
+def _get_samples_and_labels(fit_args, fit_kwargs, fit_arg_names):
     # In most cases, X_var_name and y_var_name become "X" and "y", respectively.
     # However, certain sklearn models use different variable names for X and y.
     X_var_name, y_var_name = fit_arg_names[:2]
     X, y_true = _get_Xy(fit_args, fit_kwargs, X_var_name, y_var_name)
-    y_pred = fitted_estimator.predict(X)
 
-    return X, y_true, y_pred
```
dbczumar, line 68: What was the motivation for this change? In every instance where we call `_get_samples_and_labels`, the subsequent line calls `y_pred = fitted_estimator.predict(X)`, which is effectively the same as keeping that logic in this function.

mlflow/sklearn/utils.py
```diff
@@ -26,12 +28,18 @@
...
 _SAMPLE_WEIGHT = "sample_weight"
 
-# _SklearnMetric represents a metric (e.g, precision_score) that will be computed and logged
-# during the autologging routine for a particular model type (eg, classifier, regressor).
-_SklearnMetric = collections.namedtuple("_SklearnMetric", ["name", "function", "arguments"])
+# _SklearnMetricOrArtifact represents a metric (e.g, precision_score)
+# or artifact (e.g confusion matrix) that will be computed and
+# logged during the autologging routine for a particular model type
+# (eg, classifier, regressor). "title" will only be used for recording artifacts
+_SklearnMetricOrArtifact = collections.namedtuple(
```
dbczumar, line 40: For readability and potential extensibility later, I think I'd rather maintain separate `_SklearnMetric` and `_SklearnArtifact` named tuple objects - especially because they contain different required fields (artifacts all specify a non-empty title) and are consumed by different routines that explicitly assume the object is an artifact or a metric. Can we split these up?

mlflow/sklearn/utils.py
dbczumar: Right, but I don't think we need to preserve any sort of ordering to perform this checking (which is @harupy's point). Because this routine is iterating over a list and adding elements to a set, I don't think it's order-preserving anyway. 

It should be sufficient to simplify this by returning `set(y_true)`. At that point, `_get_labels` is redundant because it simply wraps `set`. Accordingly, I think we can just remove this function and use `set()` where this used to be called.

mlflow/sklearn/utils.py
```diff
@@ -126,16 +133,38 @@ def _get_metrics_value_dict(metrics_list):
         try:
             metric_value = metric.function(**metric.arguments)
         except Exception as e:  # pylint: disable=broad-except
-            _log_warning_for_metrics(metric.name, metric.function, e)
+            _log_warning_for_metrics_or_artifacts(metric.name, metric.function, e, True)
```
dbczumar, line 136: I think it would be cleaner / more extensible to break this into `_log_warning_for_metrics` and `_log_warning_for_artifacts`. In general, this pattern of defining specialized / branching logic based on a boolean parameter is brittle & prone to complexity later on (e.g., what happens if we add a third type of entity being logged & the desired warning message is substantially different?)

mlflow/sklearn/utils.py
```diff
@@ -126,16 +133,38 @@ def _get_metrics_value_dict(metrics_list):
...
+def _get_labels(y_true):
+    seen = set()
+    return [x for x in y_true if not (x in seen or seen.add(x))]
+
+
+def _get_artifacts_list(artifacts_list):
+    artifacts_dir = tempfile.mkdtemp()
+    for artifact in artifacts_list:
+        try:
+            # print(_get_arg_names(artifact.function))
```
dbczumar, line 152: Can we remove extraneous `print()` statements? (e.g., this one and the ones on lines 158-159).

mlflow/sklearn/utils.py
```diff
@@ -174,81 +203,169 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+            function=sklearn.metrics.plot_confusion_matrix,
+            arguments=dict(
+                estimator=fitted_estimator,
+                X=X,
+                y_true=y_pred,
+                sample_weight=sample_weight,
+                normalize="true",
+                labels=labels,
+            ),
+            title="Confusion Matrix",
```
dbczumar, line 337: Nit: Each word in this title is capitalized, while the precision / recall curve plot titles are not. We should be consistent. My vote would be to only capitalize the first letter of the first word in the title; this is consistent with the axis labeling convention in sklearn.

mlflow/sklearn/utils.py
```diff
@@ -126,16 +133,38 @@ def _get_metrics_value_dict(metrics_list):
...
 
 
+# Remove duplicates from y_true while still preserving orders
+def _get_labels(y_true):
+    seen = set()
+    return [x for x in y_true if not (x in seen or seen.add(x))]
+
+
+def _get_artifacts_list(artifacts_list):
+    artifacts_dir = tempfile.mkdtemp()
```
dbczumar, line 149: Rather than using `tempfile`, can we use a `TempDir()` context (https://github.com/mlflow/mlflow/blob/master/mlflow/utils/file_utils.py#L176)?. With `TempDir()`, the artifact files will be immediately removed after logging.

mlflow/sklearn/utils.py
```diff
@@ -174,81 +203,169 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
                 [
-                    _SklearnMetric(
-                        name=_METRICS_PREFIX + "roc_auc_score",
+                    _SklearnMetricOrArtifact(
+                        name=_TRAINING_PREFIX + "roc_auc_score",
                         function=sklearn.metrics.roc_auc_score,
                         arguments=dict(
                             y_true=y_true,
-                            y_score=y_pred_proba,
+                            y_score=y_pred_proba_binary,
```
dbczumar, line 275: This throws a "variable referenced before assignment" exception when the classifier is non-binary (because `y_pred_proba_binary` never gets assigned in that case). Also, can you explain the motivation for making this change? I tried this out with logistic regression on the Iris dataset and got the exact same value for `roc_auc_score` when using `y_pred_proba` and `y_pred_proba_binary`. I'd rather not make this change if we don't have to, especially because we're making a lot of assumptions about `y_pred_proba` (i.e. it has a `shape` attribute and has the shape `(?, 2)`).

dbczumar (changes requested):
@willzhan-db This is great! I left some comments about code structure / documentation. Let me know if you have questions!

mlflow/sklearn/utils.py
willzhan-db: For scalability, yes!

mlflow/sklearn/utils.py
willzhan-db: Oh I messed up with the logic.
But regarding the motivation, we can talk about this during sync-up.

mlflow/sklearn/utils.py
willzhan-db: I create another separate try/catch block.

mlflow/sklearn/utils.py
```diff
@@ -246,9 +274,97 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+            arguments=dict(
+                estimator=fitted_estimator,
+                X=X,
+                y_true=y_pred,
+                sample_weight=sample_weight,
+                normalize="true",
+                cmap='Blues',
+                labels=list(labels),
+            ),
+            title="Confusion matrix",
```
harupy, line 328:
```suggestion
            title="Normalized confusion matrix",
```

to clarify we apply normalization?

mlflow/sklearn/utils.py
```diff
@@ -246,9 +274,97 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+        if _is_artifact_supported("roc_curve"):
+            classifier_artifacts.extend(
+                [
+                    _SklearnArtifact(
+                        name=_TRAINING_PREFIX + "roc_curve",
+                        function=sklearn.metrics.plot_roc_curve,
+                        arguments=dict(
+                            estimator=fitted_estimator, X=X, y=y_pred, sample_weight=sample_weight,
+                        ),
+                        title="Roc curve",
```
harupy, line 344:
```suggestion
                        title="ROC curve",
```

nit

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -206,37 +204,52 @@ def test_classifier():
...
+    # use binary datasets to cover the test for roc curve & precision recall curve
+    X, y_true = sklearn.datasets.load_breast_cancer(return_X_y=True)
 
     with mlflow.start_run() as run:
         model = fit_model(model, X, y_true, "fit")
 
     y_pred = model.predict(X)
     y_pred_prob = model.predict_proba(X)
+    # For bonary classification, y_score only accepts the probability of greater label
+    y_pred_prob_roc = [prob[1] for prob in y_pred_prob]
```
harupy, line 217:
```suggestion
    # For bonary classification, y_score only accepts the probability of greater label
    y_pred_prob_roc = y_pred_prob[:, 1]
```

willzhan-db: What is the span stuff?

harupy: `[:, 1]` means `grab the second column`. Does this answer your question?

mlflow/sklearn/utils.py
```diff
@@ -246,9 +274,93 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+                name=_TRAINING_PREFIX + "confusion_matrix",
+                function=sklearn.metrics.plot_confusion_matrix,
+                arguments=dict(
+                    estimator=fitted_estimator,
+                    X=X,
+                    y_true=y_pred,
+                    sample_weight=sample_weight,
+                    normalize="true",
+                    cmap="Blues",
+                    labels=list(labels),
```
harupy, line 328: Do we need to specify `labels`? [The doc](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.plot_confusion_matrix.html) says:

> If None is given, those that appear at least once in y_true or y_pred are used in sorted order.

tests/sklearn/test_sklearn_autolog.py
willzhan-db: Oh, not sure why but what I saw was a bunch of <span> stuff (looks like html). Now it looks normal again.
Thanks!

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -206,37 +205,59 @@ def test_classifier():
...
+    # use binary datasets to cover the test for roc curve & precision recall curve
+    X, y_true = sklearn.datasets.load_breast_cancer(return_X_y=True)
 
     with mlflow.start_run() as run:
         model = fit_model(model, X, y_true, "fit")
 
     y_pred = model.predict(X)
     y_pred_prob = model.predict_proba(X)
+    # For binary classification, y_score only accepts the probability of greater label
+    y_pred_prob_roc = [prob[1] for prob in y_pred_prob]
```
harupy, line 218:
```suggestion
    y_pred_prob_roc = y_pred_prob[:, 1]
```
same here

mlflow/sklearn/utils.py
```diff
@@ -246,9 +274,93 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+                ),
+                title="Normalized confusion matrix",
+            ),
+        ]
+
+        # The plot_roc_curve and plot_precision_recall_curve can only be
+        # supported for binary classifier
+        if len(labels) == 2:
+            classifier_artifacts.extend(
```
harupy, line 337: Can we add a test verifying this using a multi-class (> 2) classifier?

willzhan-db: Sure

mlflow/sklearn/utils.py
```diff
@@ -330,44 +441,80 @@ def _get_regressor_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+def _log_warning_for_artifacts(func_name, func_call, err):
+    msg = (
+        func_call.__qualname__
+        + " failed. The artifact "
+        + func_name
+        + "will not be recorded."
+        + " Artifact error: "
+        + str(err)
+    )
+    print("2: ", msg)
```
dbczumar, line 462: @willzhan-db Can we remove these println debugging statements?

mlflow/sklearn/utils.py
```diff
@@ -330,44 +441,80 @@ def _get_regressor_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+            + ". Logging error: "
+            + str(err)
+        )
+        print("4: ", msg)
+        _logger.warning(msg)
+    else:
+        if bool(name_artifact_dict):
+            with TempDir() as tmp:
+                for name, display in name_artifact_dict.items():
+                    filepath = os.path.join(tmp.path(), "{}.png".format(name))
```
dbczumar, line 514: Nit: I think we can simplify this as:

```suggestion
                    filepath = tmp.path("{}.png".format(name))
```

mlflow/sklearn/utils.py
```diff
@@ -219,18 +239,23 @@ def _get_classifier_metrics(fitted_estimator, fit_args, fit_kwargs):
...
                     function=sklearn.metrics.log_loss,
                     arguments=dict(y_true=y_true, y_pred=y_pred_proba, sample_weight=sample_weight),
                 ),
             ]
         )
 
         if _is_metric_supported("roc_auc_score"):
+            # For binary case, the parameter `y_score` expect scores must be
+            # the scores of the class with the greater label.
+            if y_pred_proba.shape[1] == 2:
```
dbczumar, line 252: I still don't think we want to assume that `y_pred_proba` has a `shape` attribute. It doesn't necessarily have to be a numpy array or a DataFrame. Could we instead check the length of the first element of `y_pred_proba`? This way, we're only assuming that `y_pred_proba` is an arbitrary 2D array, rather than a numpy array.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -608,12 +679,8 @@ def throwing_predict():  # pylint: disable=unused-argument
...
         )
         cv_model.fit(*get_iris())
-        mock_warning.assert_called_once()
-        mock_warning.called_once_with(
-            "Failed to autolog metrics for "
-            + sklearn.model_selection.GridSearchCV.__class__.__name__
-            + ". Logging error: EXCEPTION"
-        )
+        # Will be called twice, once for metrics, once for artifacts
+        mock_warning.assert_called()
```
dbczumar, line 683: If it's meant to be called twice, can we assert that it was called twice?

dbczumar (changes requested):
@willzhan-db This is looking really good - a few more small comments. Also, can we update the `mlflow.sklearn.autolog()` method documentation to list the set of artifacts logged for each type of model (i.e., just for classifiers right now).

mlflow/sklearn/utils.py
```diff
@@ -330,44 +441,80 @@ def _get_regressor_metrics(fitted_estimator, fit_args, fit_kwargs):
...
+            + str(err)
+        )
+        print("4: ", msg)
+        _logger.warning(msg)
+    else:
+        if bool(name_artifact_dict):
+            with TempDir() as tmp:
+                for name, display in name_artifact_dict.items():
+                    filepath = tmp.path("{}.png".format(name))
+                    display.figure_.savefig(filepath)
```
harupy, line 515: Can we close the figure after saving it to avoid the following matplotlib warning?

```
RuntimeWarning: More than 20 figures have been opened.
Figures created through the pyplot interface (`matplotlib.pyplot.figure`)
are retained until explicitly closed and may consume too much memory.
(To control this warning, see the rcParam `figure.max_open_warning`).
```

code to reproduce the warning:

```python
import numpy as np
from sklearn.linear_model import SGDClassifier

import mlflow

# enable autologging
mlflow.sklearn.autolog()

# prepare training data
X = np.array([[1, 1], [1, 2], [2, 2], [2, 3]])
y = np.random.randint(2, size=4)

# train a model
model = SGDClassifier()

for i in range(100):
    print(i)
        model.fit(X, y)
```

harupy:
https://github.com/mlflow/mlflow/pull/3423 took over this PR.

PR #3375, opened 2020-09-01
Initial support for pip projects

fhoering:

Add support for pip only projects (not using conda or docker)

## How is this patch tested?

Tested with mlflow projects local backend and mlflow-yarn backend.

Add support for pip only projects.

Components 
- [x ] `area/projects`: MLproject format, project running backends

Interface 

Language 

Integrations


fhoering:
@dbczumar @smurching @juntai-zheng 
Any news ? There not so many changes actually and it would unlock to be able to use only pip instead of conda.
I guess this could be also reused for other parts where having an isolated python env is important.

juntai-zheng:
Hey @fhoering, could you answer some questions regarding this new feature in regards to the motivation behind it? These are pulled from the motivation section for FRs:
- What is the use case for this feature?
- Why is this use case valuable to support for MLflow users in general?
- Why is this use case valuable to support for your project(s) or organization?
- Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient)

fhoering:
>     What is the use case for this feature?
>  Why is this use case valuable to support for your project(s) or organization?

We don't use conda at all in our organization. So what currently we can't use any feature of MLflow that makes use of conda for packaging. It is not an option to use conda becasue we have our own internal pypi mirror and it is used by many other Python projects. All our CI/Prod code is based on Pypi and pip. That's why we alignd our datascience tools to that.

>  Why is this use case valuable to support for MLflow users in general?
I'm not sure I understand the question. pip is widely used in the Python community. This would open up  mlflow to much more users than what is currently existing with conda.

>   Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient)
As you can see in the PR it is not that difficult and the maintenance effort seems limited but it would be cleaner to integrate this by default into MLflow.

jinzhang21:
We've added support to use pip-only for mlflow installation. Closing this PR. Thanks for the contribution!

PR #3386, opened 2020-09-03
Add inference of input/output schema and input example to LightGBM autologging

andrewnitu:

- LightGBM autologging will now automatically include the input and output type signatures in the MLmodel artifact
- LightGBM autologging will now automatically include input_example.json artifact

## How is this patch tested?

Unit testing & manual testing

The autolog function for LightGBM models will now automatically include an example input and the model signature as part of the generated artifacts. (They will follow the format listed in the docs at https://mlflow.org/docs/latest/models.html#model-signature https://mlflow.org/docs/latest/models.html#model-input-example)

Components 

Interface 

Language 

Integrations


mlflow/lightgbm.py
```diff
@@ -424,7 +429,32 @@ def log_feature_importance_plot(features, importance, importance_type):
             finally:
                 shutil.rmtree(tmpdir)
 
-        try_mlflow_log(log_model, model, artifact_path="model")
+        input_example = None
+        signature = None
+        try:
+            raw_data = train_data.get_data()
+            if isinstance(raw_data, str):
```
andrewnitu, line 436: i'm wondering if this is the best flow for this case (when the data is given as a path to a file)

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -312,3 +313,69 @@ def test_lgb_autolog_loads_model_from_artifact(bst_params, train_set):
...
+    input_example_filename = "input_example.json"
+    assert "model/" + input_example_filename in artifacts
+    input_example_path = os.path.join(artifacts_dir, "model", input_example_filename)
+
+    input_example = None
+    with open(input_example_path, "r") as f:
+        input_example = json.load(f)["data"]
+
+    for i in range(5):
+        assert (input_example[i] == iris.data[i][:2]).all
```
dbczumar, line 345: There's an internal `_read_example` method (https://github.com/mlflow/mlflow/blob/master/mlflow/models/utils.py#L136) that can be used to read an input example from an MLflow model as a Pandas DataFrame. I think we could simplify this logic by leveraging that utility and comparing the resulting dataframe to the `X` value you constructed on line 324.

Here's a usage example: https://github.com/mlflow/mlflow/blob/master/tests/sklearn/test_sklearn_autolog.py#L720

dbczumar: Following from ^, can we also evaluate the reloaded input example using the trained model's `pyfunc` representation (which can be obtained via `mlflow.pyfunc.load_model`) and confirm that inference is successful (i.e. the inference line should not throw an exception / fail).

dbczumar (commented):
@andrewnitu This is great! Left a few comments - let me know if you have any questions!

mlflow/lightgbm.py
```diff
@@ -379,6 +379,11 @@ def log_feature_importance_plot(features, importance, importance_type):
         else:
             kwargs["callbacks"] = [callback]
 
+        # We set free_raw_data to false on the Dataset object
+        # so that we can access the original data later.
+        train_data = args[1] if len(args) > 1 else kwargs["train_data"]
+        train_data.free_raw_data = False
+
```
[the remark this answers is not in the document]
harupy: Can we create a slice of `train_data.data` and use it after training instead of setting `free_raw_data` to `False`?

```python
sample = train_data.data[:5]

...

model = original(*args, **kwargs)  # == lgb.train

# do something using `sample`
```

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -312,3 +313,69 @@ def test_lgb_autolog_loads_model_from_artifact(bst_params, train_set):
...
+    ml_model_filename = "MLmodel"
+    assert "model/" + ml_model_filename in artifacts
+    ml_model_path = os.path.join(artifacts_dir, "model", ml_model_filename)
+
+    data = None
+    with open(ml_model_path, "r") as f:
+        data = yaml.load(f, Loader=yaml.FullLoader)
+
+    assert data
+    assert data["signature"]
```
harupy, line 372:
```suggestion
    assert data is not None
    assert "signature" in data
```

to clarify what we're asserting here?

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -312,3 +313,69 @@ def test_lgb_autolog_loads_model_from_artifact(bst_params, train_set):
...
+    mlflow.lightgbm.autolog()
+    lgb.train(bst_params, dataset)
+    run = get_latest_run()
+    run_id = run.info.run_id
+    artifacts_dir = run.info.artifact_uri.replace("file://", "")
+    client = mlflow.tracking.MlflowClient()
+    artifacts = [x.path for x in client.list_artifacts(run_id, "model")]
+
+    input_example_filename = "input_example.json"
+    assert "model/" + input_example_filename in artifacts
```
harupy, line 337:
```suggestion
    assert os.path.join("model", input_example_filename)
```

should we use `os.path.join` here so we can run this test on windows?

harupy: We currently don't run model flavor tests on windows though...

mlflow/lightgbm.py
```diff
@@ -424,7 +429,32 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+        signature = None
+        try:
+            raw_data = train_data.get_data()
+            if isinstance(raw_data, str):
+                raise Exception(
+                    "The training data was given as a path. Input example and"
+                    + " model signature inference is not supported for datasets specified by paths."
+                )
+
+            SAMPLE_ROWS = 5
```
harupy, line 442: @dbczumar should we put this constant in a separate file (for example `autologging_utils.py`) and reuse it across autlogging integrations?

tests/lightgbm/test_lightgbm_autolog.py
andrewnitu: might as well, so that it's consistent with the lines around it

mlflow/lightgbm.py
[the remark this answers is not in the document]
andrewnitu: @harupy I think we wanted to run the original train function before attempting to slice the data in case of invalid input, i.e. if they put None, 2, etc as the data. The train function would perform its own validation on it and it would fail with a useful error message for the user.

However, if we just wrap this in a try-catch I think it should work nicely! And it also would save a lot of memory

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -312,3 +313,69 @@ def test_lgb_autolog_loads_model_from_artifact(bst_params, train_set):
     np.testing.assert_array_almost_equal(
         model.predict(train_set.data), loaded_model.predict(train_set.data)
     )
+
```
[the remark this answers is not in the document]
andrewnitu: sounds good! do we have such a dataset in a file already? otherwise I can save one to a file during the test.

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -312,3 +313,69 @@ def test_lgb_autolog_loads_model_from_artifact(bst_params, train_set):
...
+    assert data["signature"]
+    assert (
+        data["signature"]["inputs"]
+        == '[{"name": "sepal length (cm)", "type": "double"}, '
+        + '{"name": "sepal width (cm)", "type": "double"}]'
+    )
+    assert (
+        data["signature"]["outputs"]
+        == '[{"type": "double"}, {"type": "double"}, {"type": "double"}]'
+    )
```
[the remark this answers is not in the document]
andrewnitu: totally agree with this point

just a side note it's interesting i've never actually seen a dictionary out of order, even though the spec says it could be :D

tests/lightgbm/test_lightgbm_autolog.py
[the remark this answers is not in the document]
andrewnitu: I ended up creating a basic CSV to work with

andrewnitu:
@harupy @dbczumar I have made the requested changes :)

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -312,3 +315,104 @@ def test_lgb_autolog_loads_model_from_artifact(bst_params, train_set):
...
+        {"type": "double"},
+        {"type": "double"},
+    ]
+
+
+@pytest.mark.large
+def test_lgb_autolog_continues_logging_even_if_signature_inference_fails():
+    # signature and input example inference should fail here since the dataset is given
+    #   as a file path
+    dataset = lgb.Dataset(os.path.join(os.path.dirname(__file__), "iris_training.csv"))
```
harupy, line 393: Can we use a temporary csv file instead of `iris_training.csv`?

```suggestion
@pytest.mark.large
def test_lgb_autolog_continues_logging_even_if_signature_inference_fails(tmpdir):
    # signature and input example inference should fail here since the dataset is given
    #   as a file path
    tmp_csv = tmpdir.join("data.csv")
    tmp_csv.write("foo\n1.0")
    dataset = lgb.Dataset(tmp_csv.strpath)
```

harupy: how `tmpdir` works:

https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture

mlflow/lightgbm.py
```diff
@@ -379,6 +380,22 @@ def log_feature_importance_plot(features, importance, importance_type):
...
             kwargs["callbacks"] = [callback]
 
+        # We set free_raw_data to false on the Dataset object
+        # so that we can access the original data later.
+        train_data = args[1] if len(args) > 1 else kwargs["train_data"]
+
+        input_example = None
+        try:
+            if isinstance(train_data.data, str):
+                raise Exception("The input data was of type string.")
```
harupy, line 390: should we use [MlflowException](https://github.com/mlflow/mlflow/blob/d743a40426d5dedbde395a4e6bbdeebadbccd4dc/mlflow/exceptions.py#L31) here?

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -312,3 +315,104 @@ def test_lgb_autolog_loads_model_from_artifact(bst_params, train_set):
...
+    mlflow.lightgbm.autolog()
+    lgb.train(bst_params, dataset)
+    run = get_latest_run()
+    run_id = run.info.run_id
+    artifacts_dir = run.info.artifact_uri.replace("file://", "")
+    client = mlflow.tracking.MlflowClient()
+    artifacts = [x.path for x in client.list_artifacts(run_id, "model")]
+
+    ml_model_filename = "MLmodel"
+    assert str(os.path.join("model", ml_model_filename)) in artifacts
```
harupy, line 409:
```suggestion
    assert os.path.join("model", ml_model_filename) in artifacts
```

nit

mlflow/lightgbm.py
```diff
@@ -379,6 +380,22 @@ def log_feature_importance_plot(features, importance, importance_type):
         else:
             kwargs["callbacks"] = [callback]
 
+        # We set free_raw_data to false on the Dataset object
+        # so that we can access the original data later.
+        train_data = args[1] if len(args) > 1 else kwargs["train_data"]
```
dbczumar, line 385: I don't think we ended up addressing the previous comment here; i.e., what happens if `len(args) <= 1` and `train_data` isn't in kwargs? In that case, it looks like MLflow will throw before LightGBM does, which creates a somewhat confusing UX.

mlflow/lightgbm.py
```diff
@@ -379,6 +380,22 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+
+        input_example = None
+        try:
+            if isinstance(train_data.data, str):
+                raise Exception("The input data was of type string.")
+
+            SAMPLE_ROWS = 5
+            input_example = deepcopy(train_data.data[:SAMPLE_ROWS])
+        except Exception as e:  # pylint: disable=broad-except
+            input_example = None
```
dbczumar, line 395: Nit: I think this line is unnecessary; if the assignment to `input_example` via `deepcopy()` fails, `input_example will still be `None` (as defined above). Alternatively, instead of removing this instance, you could also remove the one on line 387.

mlflow/lightgbm.py
andrewnitu: makes sense; for some reason i thought indexing would return None if not found

I switched it to .get instead so that it will not throw, and the code can pass to the original lightGBM train function and give the correct error message

andrewnitu:
i'll rebase to sign the missing commits after the code is final

dbczumar (approved):
LGTM with one documentation nit; let's get this merged once that nit is addressed. Thanks @andrewnitu !

PR #3398, opened 2020-09-07
Use allowlisted APIs to fetch information on clusters

smurching:

Use new, allowlisted APIs to fetch information on clusters when DataFrame ACLs are enabled. We fall back to using existing APIs for backwards compatibility with older runtime versions.

## How is this patch tested?

Updated unit tests

Components 

Interface 

Language 

Integrations


mlflow/utils/databricks_utils.py
```diff
@@ -148,17 +161,25 @@ def get_webapp_url():
...
 
 def get_workspace_info_from_dbutils():
     dbutils = _get_dbutils()
     if dbutils:
-        context = json.loads(
-            dbutils.notebook.entry_point.getDbutils().notebook().getContext().toJson()
-        )
-        workspace_host = context["extraContext"]["api_url"]
-        workspace_id = context["tags"]["orgId"]
+        print("Yo fro msid")
```
nfx, line 180: please remove this debug line

nfx (changes requested):

mlflow/utils/databricks_utils.py
smurching: Oops, haha thanks for the catch!

mlflow/utils/databricks_utils.py
```diff
@@ -35,18 +34,29 @@ def _get_java_dbutils():
...
     if tag_opt.isDefined():
         return tag_opt.get()
     else:
         return None
 
 
+def acl_path_of_acl_root():
+    try:
+        return _get_command_context().aclPathOfAclRoot().get()
+    except Exception:  # pylint: disable=broad-except
```
nfx, line 56: can you catch exact exception here and remove pylint disable?

mlflow/utils/databricks_utils.py
```diff
@@ -118,7 +128,10 @@ def get_notebook_path():
     path = _get_property_from_spark_context("spark.databricks.notebook.path")
     if path is not None:
         return path
-    return _get_extra_context("notebook_path")
+    try:
+        return _get_command_context().notebookPath().get()
+    except Exception:  # pylint: disable=broad-except
```
nfx, line 133: remove `# pylint: disable=broad-except` and catch exact exception

mlflow/utils/databricks_utils.py
```diff
@@ -148,17 +161,24 @@ def get_webapp_url():
...
+    try:
+        return _get_command_context().apiUrl().get()
+    except Exception:  # pylint: disable=broad-except
+        return _get_extra_context("api_url")
+
+
+def get_workspace_id():
+    try:
+        return _get_command_context().workspaceId().get()
+    except Exception:  # pylint: disable=broad-except
```
nfx, line 173: remove # pylint: disable=broad-except and catch exact exception

nfx (changes requested):
remove lint ignores and catch exact exceptions

smurching:
@nfx it's a good question whether we can catch a more specific exception type. It turns out Py4j (at least the current release of Py4j) will only ever raise instances of Py4jError ([docs](https://www.py4j.org/advanced_topics.html#py4j-exceptions)). However, since our goal is to support older runtime versions with the current MLflow client (which may have older versions of Py4j or other hard-to-test behavior), it seems the least risky option is to try the new behavior of using allowlisted APIs, and fall back to the existing behavior in case of any exceptions, rather than try to catch a specific exception and find that we've broken behavior on existing runtimes.

One other subtlety worth noting in case we revisit this in the future - `py4j` is intentionally not a dependency of MLflow. In order to catch `Py4jError`, we may be tempted to `from py4j.protocol import Py4jError` at the top level of the `databricks_utils` module, i.e. make it dependent on `py4j`. However since `databricks_utils` is occasionally used in contexts outside of Databricks (e.g. when triggering execution of MLflow projects remotely on Databricks from your laptop), we should instead perform all `py4j` imports inside functions where they're used & ensure these functions are only called on Databricks.

nfx:
@smurching the pain point of catching `Exception` is that it can catch OSError, KeyError, ZeroDivisionError and other dangerous things. if we explicitly don't want to add py4j dependency, then i'd agree with your comment.

tomasatdatabricks (approved):
lgtm

PR #3403, opened 2020-09-09
Add inference of input/output schema and input example to XGBoost autologging

andrewnitu:

- XGBoost autologging will now automatically include the input and output type signatures in the MLmodel artifact
- XGBoost autologging will now automatically include input_example.json artifact

## How is this patch tested?

Unit testing & manual testing

The autolog function for XGBoost models will now automatically include an example input and the model signature as part of the generated artifacts. (They will follow the format listed in the docs at https://mlflow.org/docs/latest/models.html#model-signature https://mlflow.org/docs/latest/models.html#model-input-example)

Components 

Interface 

Language 

Integrations


mlflow/xgboost.py
```diff
@@ -416,10 +437,36 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+            signature=signature,
+            input_example=input_example,
+        )
 
         if auto_end_run:
             try_mlflow_log(mlflow.end_run)
         return model
 
     wrap_patch(xgboost, "train", train)
+    wrap_patch(xgboost.DMatrix, "__init__", __init__)
```
harupy, line 472: Lools like it's impossible to extract given `data` from `xgboost.DMatrix`: https://github.com/dmlc/xgboost/issues/3173

mlflow/xgboost.py
```diff
@@ -416,10 +437,36 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+
+        input_example = None
+        signature = None
+        try:
+            input_example = dtrain.data_copy
+
+            # it is possible that the dataset was constructed before the patched
+            #   constructor was applied, so we cannot assume the data_copy exists
+            if input_example is None:
+                raise Exception("failed to gather example input.")
```
dbczumar, line 451: Can we provide a more specific / actionable message to the user? E.g., `"please ensure that autologging is enabled before you import or construct your training dataset"`?

mlflow/xgboost.py
```diff
@@ -345,6 +365,7 @@ def log_feature_importance_plot(features, importance, importance_type):
         # logging booster params separately via mlflow.log_params to extract key/value pairs
         # and make it easier to compare them across runs.
         params = args[0] if len(args) > 0 else kwargs["params"]
+
```
dbczumar, line 368: Nit: Revert empty line

mlflow/xgboost.py
```diff
@@ -286,6 +288,24 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
     import xgboost
     import numpy as np
 
+    def __init__(self, *args, **kwargs):
+        data = args[0] if len(args) > 0 else kwargs["data"]
```
dbczumar, line 292: what happens if `len(args) == 0` and `data` is not in `kwargs` (e.g., user error)?

mlflow/xgboost.py
```diff
@@ -416,10 +437,36 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+        dtrain = args[1] if len(args) > 1 else kwargs["dtrain"]
+
+        input_example = None
+        signature = None
+        try:
+            input_example = dtrain.data_copy
+
+            # it is possible that the dataset was constructed before the patched
+            #   constructor was applied, so we cannot assume the data_copy exists
+            if input_example is None:
```
dbczumar, line 450: In the case where the dataset was constructed prior to patching, won't we fail at the first line in this block (`input_example = dtrain.data_copy`) with an `AttributeError`? Instead, should we use `getattr()` to access the `data_copy` attribute to ensure that accessing the attribute doesn't throw?

mlflow/xgboost.py
```diff
@@ -286,6 +288,24 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
...
+        original = gorilla.get_original_attribute(xgboost.DMatrix, "__init__")
+
+        data_copy = None
+        try:
+            if isinstance(data, str):
+                raise Exception("The input data was of type string.")
+
+            data_copy = deepcopy(data[:5])
+        except Exception as e:
+            msg = "Failed to gather an input example: " + str(e)
```
dbczumar, line 303: It seems a bit odd that we're logging a warning message when the dataset is constructed rather than during the training process. Users may construct dataset objects outside of the training context; e.g., they may create an XGBoost dataset and serialize it in binary format to be shared / reloaded later. In that case, logging a warning message such as "Failed to gather an input example" may seem quite confusing.

Instead, perhaps we could hold on to the exception string via the `data_copy` attribute and use it to populate the warning message emitted in `train()` when an input example cannot be found? We could use an internal container object (e.g., `_InputExampleInfo`) which contains `input_example` and `error` fields that are populated in a mutually exclusive fashion.

mlflow/xgboost.py
```diff
@@ -416,10 +437,36 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+
+            # it is possible that the dataset was constructed before the patched
+            #   constructor was applied, so we cannot assume the data_copy exists
+            if input_example is None:
+                raise Exception("failed to gather example input.")
+
+            model_output = model.predict(xgb.DMatrix(input_example))
+            signature = infer_signature(input_example, model_output)
+        except Exception as e:  # pylint: disable=broad-except
+            msg = "Failed to infer the model signature: " + str(e)
```
dbczumar, line 456: In the case where `input_example` is assigned to a non-None value but `infer_signature` fails, we should reset `input_example` to `None` before passing it to `log_model()`. I noticed we have a test case for this (`test_xgb_autolog_continues_logging_even_if_signature_inference_fails`), but it doesn't seem to assert that `input_example` is not present when signature inference fails. We should add that for consistency with the behavior of `mlflow.skearn.autolog()` and `mlflow.lightgbm.autolog()`.

Though it does occur to me that `input_example` may be independently useful even if `signature` cannot be inferred. @harupy Can you weigh in here? I forgot why we decided to reset `input_example` in this case.

dbczumar (changes requested):
@andrewnitu This is great! Left a few comments - let me know if you have any questions!

mlflow/xgboost.py
harupy: @dbczumar 

> I forgot why we decided to reset input_example in this case.

I remember it's for simplicity.

harupy: Found a related comment: https://github.com/mlflow/mlflow/pull/3326#discussion_r478742287

mlflow/xgboost.py
andrewnitu: ah, that's correct. technically the autologging will still continue because it's in a try block, but I definitely wanted it to take the `if input_example is None` path instead

mlflow/xgboost.py
```diff
@@ -340,6 +365,8 @@ def log_feature_importance_plot(features, importance, importance_type):
                 plt.close(fig)
                 shutil.rmtree(tmpdir)
 
+        import xgboost as xgb
```
dbczumar, line 368: Looks like we already import xgboost at the top of the `autolog()` function. Do we need to import it again?

mlflow/xgboost.py
```diff
@@ -416,10 +443,51 @@ def log_feature_importance_plot(features, importance, importance_type):
             finally:
                 shutil.rmtree(tmpdir)
 
-        try_mlflow_log(log_model, model, artifact_path="model")
+        # dtrain must exist as the original train function already ran successfully
+        dtrain = args[1] if len(args) > 1 else kwargs.get("dtrain")
+
+        input_example_info = None
```
dbczumar, line 449: I think this line is unnecessary, since `input_example_info` is always reassigned in the `try` block to the output of `getattr(...)`.

mlflow/xgboost.py
```diff
@@ -286,6 +287,30 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
...
+    def __init__(self, *args, **kwargs):
+        data = args[0] if len(args) > 0 else kwargs.get("data")
+
+        if data is not None:
+            original = gorilla.get_original_attribute(xgboost.DMatrix, "__init__")
+
+            input_example_info = None
+            try:
+                if isinstance(data, str):
+                    raise Exception("The input data was of type string.")
```
dbczumar, line 304: Can we be more specific about the types of input data that are supported / indicate why we can't capture string data? More specifically, the problem is that `data` refers to a path, right?

mlflow/xgboost.py
```diff
@@ -286,6 +287,30 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
     import xgboost
```
dbczumar, line 287: Can we add docs to the `mlflow.xgboost.autolog()` method indicating that we also log the signature and input example when possible?

mlflow/xgboost.py
```diff
@@ -286,6 +287,30 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
...
     import numpy as np
 
+    class _InputExampleInfo:
+        def __init__(self, input_example=None, error_msg=None):
+            self.input_example = input_example
+            self.error_msg = error_msg
+
+    def __init__(self, *args, **kwargs):
```
dbczumar, line 295: Can we add documentation to this function explaining why we're patching it (i.e. in order to fetch an input example).

mlflow/xgboost.py
```diff
@@ -286,6 +287,30 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
...
+
+        if data is not None:
+            original = gorilla.get_original_attribute(xgboost.DMatrix, "__init__")
+
+            input_example_info = None
+            try:
+                if isinstance(data, str):
+                    raise Exception("The input data was of type string.")
+
+                input_example_info = _InputExampleInfo(input_example=deepcopy(data[:5]))
```
dbczumar, line 306: can we pull `5` into a variable defined in `autologging_utils.py` and use that variable throughout the various autologging integrations that support signature / input example collection? (i.e. sklearn, lightgbm, xgboost). cc @harupy

mlflow/xgboost.py
dbczumar, line 295: Can we add a test case ensuring that datasets can still be successfully serialized to disk and reloaded even with our patch modifications? It'd be painful if adding this extra attribute somehow interfered with that serialization / deserialization process.

mlflow/xgboost.py
```diff
@@ -416,10 +443,51 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+
+        if input_example is not None:
+            # input example collection succeeded, move on to signature prediction
+            try:
+                model_output = model.predict(xgb.DMatrix(input_example))
+                signature = infer_signature(input_example, model_output)
+            except Exception as e:  # pylint: disable=broad-except
+                input_example = None
+                msg = "Failed to infer the model signature: " + str(e)
+                _logger.warning(msg)
```
dbczumar, line 478: Can we consolidate this into the same try/catch block that's used for input example extraction? Ultimately, the user either ends up with both input example and signature or ends up with neither one. We could use a single warning message beginning with `Failed to gather example input and model signature:` to make this clear. Otherwise, if the input example is missing, we don't currently say anything about the signature being absent in our warning message.

dbczumar (changes requested):
@andrewnitu Awesome progress - I left a few more comments regarding documentation & code style

mlflow/xgboost.py
andrewnitu: it seems we have this already due to another test: https://github.com/mlflow/mlflow/pull/3403/files#diff-581561ae97c9c7aa194fbdc0a42fdc3fR304

is this the same kind of serialization/deserialization you are referring to?

harupy: I think corey's point is we should add a test like below.

```python
def test_xxx_yyy():
    mlflow.xgboost.autolog()
    dtrain = xgb.DMatrix(...)

    ...

    model = xgb.train()

    ...

    dtrain.save_binary(path)  # doesn't throw
    xgb.DMatrix(path)  # doesn't throw
```

tests/xgboost/test_xgboost_autolog.py
```diff
@@ -295,3 +298,108 @@ def test_xgb_autolog_loads_model_from_artifact(bst_params, dtrain):
...
+
+    xgb.train(bst_params, dataset)
+    run = get_latest_run()
+    run_id = run.info.run_id
+    artifacts_dir = run.info.artifact_uri.replace("file://", "")
+    client = mlflow.tracking.MlflowClient()
+    artifacts = [x.path for x in client.list_artifacts(run_id, "model")]
+
+    ml_model_filename = "MLmodel"
+    assert str(os.path.join("model", ml_model_filename)) in artifacts
```
harupy, line 349:
```suggestion
    assert os.path.join("model", ml_model_filename) in artifacts
```

nit

tests/xgboost/test_xgboost_autolog.py
```diff
@@ -295,3 +298,108 @@ def test_xgb_autolog_loads_model_from_artifact(bst_params, dtrain):
...
+    assert json.loads(signature["outputs"]) == [
+        {"type": "float"},
+        {"type": "float"},
+        {"type": "float"},
+    ]
+
+
+@pytest.mark.large
+def test_xgb_autolog_continues_logging_even_if_signature_inference_fails(bst_params, tmpdir):
+    tmp_csv = tmpdir.join("data2.csv")
```
harupy, line 377:
```suggestion
    tmp_csv = tmpdir.join("data.csv")
```

nit

tests/xgboost/test_xgboost_autolog.py
andrewnitu: ah, i think this is conflicting with the other XGBoost bugs PR. I will resolve the merges and this should be fixed

mlflow/xgboost.py
andrewnitu: ohhh i see, i am adding this now, thanks!

tests/xgboost/test_xgboost_autolog.py
```diff
@@ -329,3 +404,54 @@ def test_xgb_autolog_does_not_throw_if_importance_values_are_empty(bst_params, t
...
+    iris = datasets.load_iris()
+    X = pd.DataFrame(iris.data[:, :2], columns=iris.feature_names[:2])
+    y = iris.target
+    dataset = xgb.DMatrix(X, y)
+
+    xgb.train(bst_params, dataset)
+
+    dataset.save_binary("dataset_serialization_test")  # serialization should not throw
+    xgb.DMatrix("dataset_serialization_test")  # deserialization also should not throw
+    os.remove("dataset_serialization_test")
```
harupy, line 457: Should we use `tmpdir` here because if we keyboard-interrupt right after `xgb.DMatrix(...)`, the `"dataset_serialization_test"` file remains in the project root directory?

andrewnitu: true... I changed it :)

mlflow/xgboost.py
```diff
@@ -291,6 +299,38 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
...
+    #   to use as an input example and for inferring the model signature.
+    #   (there is no way to get the data back from a DMatrix object)
+    # We store it on the DMatrix object so the train function is able to read it.
+    def __init__(self, *args, **kwargs):
+        data = args[0] if len(args) > 0 else kwargs.get("data")
+
+        if data is not None:
+            original = gorilla.get_original_attribute(xgboost.DMatrix, "__init__")
+
+            input_example_info = None
```
dbczumar, line 317: Nit: I think this assignment is unnecessary

dbczumar (approved):
LGTM! Awesome work @andrewnitu !

harupy (approved):
LGTM :)

Issue #3258, opened 2020-08-11
[BUG] mlflow serve pytorch model -> no module named torch


Zethson:
### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [X ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: Y
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:Arch Linux
- **MLflow installed from (source or binary)**: Binary
- **MLflow version (run ``mlflow --version``)**: 1.10.0
- **Python version**: 3.7.6
- **Exact command to reproduce**: mlflow models serve -m MODELPATH

### Describe the problem
I successfully trained a model. Now, when trying to serve it I run into:
```
zeth@master /tmp> mlflow models serve -m /tmp/exploding_springfield/mlruns/0/f7c632e43f93437280cc72b88f279a56/artifacts/models                                                         (base) 
2020/08/11 14:47:26 INFO mlflow.models.cli: Selected backend for flavor 'python_function'
2020/08/11 14:47:28 INFO mlflow.pyfunc.backend: === Running command 'source /home/zeth/anaconda3/bin/../etc/profile.d/conda.sh && conda activate mlflow-dd325f076f6465c8205b2342fd8ab4531e905e1a 1>&2 && gunicorn --timeout=60 -b 127.0.0.1:5000 -w 1 ${GUNICORN_CMD_ARGS} -- mlflow.pyfunc.scoring_server.wsgi:app'
[2020-08-11 14:47:28 +0200] [20429] [INFO] Starting gunicorn 20.0.4
[2020-08-11 14:47:28 +0200] [20429] [INFO] Listening at: http://127.0.0.1:5000 (20429)
[2020-08-11 14:47:28 +0200] [20429] [INFO] Using worker: sync
[2020-08-11 14:47:28 +0200] [20435] [INFO] Booting worker with pid: 20435
[2020-08-11 14:47:29 +0200] [20435] [ERROR] Exception in worker process
Traceback (most recent call last):
  File "/home/zeth/anaconda3/lib/python3.7/site-packages/gunicorn/arbiter.py", line 583, in spawn_worker
[...]

Issue #3199, opened 2020-07-30
Unable to log pytorch models


Dhaval08:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for additional information about bug reports. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

**Please fill in this bug report template to ensure a timely and thorough response.**

### Describe the problem
I am training a pytorch model and trying to log it towards the end using the following command
```
mlflow.pytorch.log_model(model, "model")
```

### Code to reproduce issue
https://github.com/mlflow/mlflow/tree/master/examples/pytorch
I am using the exact same code by adding the log_model line at the bottom

### Other info / logs
```
Traceback (most recent call last):
  File "mnist_tensorboard_artifact.py", line 172, in <module>
    mlflow.pytorch.log_model(model, "model")
  File "/Users/dshah/Desktop/local/miniconda/envs/mlflow-3c24da0bdd94369e854abcee14bf75d1ad29c419/lib/python3.6/site-packages/mlflow/pytorch/__init__.py", line 179, in log_model
    signature=signature, input_example=input_example, **kwargs)
  File "/Users/dshah/Desktop/local/miniconda/envs/mlflow-3c24da0bdd94369e854abcee14bf75d1ad29c419/lib/python3.6/site-packages/mlflow/models/model.py", line 154, in log
    **kwargs)
  File "/Users/dshah/Desktop/local/miniconda/envs/mlflow-3c24da0bdd94369e854abcee14bf75d1ad29c419/lib/python3.6/site-packages/mlflow/pytorch/__init__.py", line 300, in save_model
    torch.save(pytorch_model, model_path, pickle_module=pickle_module, **kwargs)
  File "/Users/dshah/Desktop/local/miniconda/envs/mlflow-3c24da0bdd94369e854abcee14bf75d1ad29c419/lib/python3.6/site-packages/torch/serialization.py", line 328, in save
    _legacy_save(obj, opened_file, pickle_module, pickle_protocol)
  File "/Users/dshah/Desktop/local/miniconda/envs/mlflow-3c24da0bdd94369e854abcee14bf75d1ad29c419/lib/python3.6/site-packages/torch/serialization.py", line 401, in _legacy_save
    pickler.dump(obj)
[...]

PR #3408, opened 2020-09-10
Removing CloudPickle dependency for mlflow pytorch log/save model

shrinath-suresh:

https://pytorch.org/tutorials/beginner/saving_loading_models.html 

Pytorch now supports `pickle` module for serialization. Removing cloud pickle dependencies. 

## How is this patch tested?

Tested save, log and load model with pickle as serialization module. Models are getting saved and loaded. pickle_module_info.txt is showing pickle as well. 

Components 

Interface 

Language 

Integrations


mlflow/pytorch/__init__.py
```diff
@@ -52,12 +52,6 @@ def get_default_conda_env():
             "pytorch={}".format(torch.__version__),
             "torchvision={}".format(torchvision.__version__),
         ],
-        additional_pip_deps=[
-            # We include CloudPickle in the default environment because
-            # it's required by the default pickle module used by `save_model()`
-            # and `log_model()`: `mlflow.pytorch.pickle_module`.
-            "cloudpickle=={}".format(cloudpickle.__version__)
```
shrinath-suresh, line 59: will add pickle module here

dbczumar:
@shrinath-suresh @smurching With this change, can we still capture PyTorch model classes defined in the `__main__` scope or a notebook environment?

shrinath-suresh:
> @shrinath-suresh @smurching With this change, can we still capture PyTorch model classes defined in the `__main__` scope or a notebook environment?

Thanks for the comment. Looks like there are specific tests written for it - https://github.com/mlflow/mlflow/blob/326c5ecc77b16bc8b08ecc504d62dc5183be1cfa/tests/pytorch/test_pytorch_model_export.py#L506 . When i change the pickle_module from `cloudpickle` to `cloudpickle` , save_model is throwing exception. So the answer is `No`

shrinath-suresh:
> > @shrinath-suresh @smurching With this change, can we still capture PyTorch model classes defined in the `__main__` scope or a notebook environment?
> 
> Thanks for the comment. Looks like there are specific tests written for it -
> 
> https://github.com/mlflow/mlflow/blob/326c5ecc77b16bc8b08ecc504d62dc5183be1cfa/tests/pytorch/test_pytorch_model_export.py#L506
> 
> . When i change the pickle_module from `cloudpickle` to `pickle` , save_model is throwing exception. So the answer is `No`

@dbczumar @smurching Can you please suggest if we need to live with `cloudpickle` dependency (even for newer version of pytorch models) due to the above mentioned issue.

smurching (commented):
@shrinath-suresh it seems we may need to stick with cloudpickle for now for the reasons @dbczumar mentioned. 

Our existing pytorch example ([link](https://github.com/mlflow/mlflow/blob/1f637f888445df2e6a1e89d3c164d467de62eef0/examples/pytorch/mnist_tensorboard_artifact.py#L226)) actually calls `mlflow.pytorch.log_model` -> `torch.save` using `pickle` rather than `cloudpickle` (i.e. it uses `pickle`, as this PR attempts to do). 

However, this causes loading back the model afterwards with `mlflow.pytorch.load_model` to break:


```
$ python examples/pytorch/mnist_tensorboard_artifact.py --epochs=0
...
Logging the trained model as a run artifact...

The model is logged at:
./mlruns/0/6e67a69c815145ae9f438ce6ad4fc18f/artifacts/pytorch-model

$  python
>>> import mlflow.pytorch
>>> model = mlflow.pytorch.load_model("./mlruns/0/6e67a69c815145ae9f438ce6ad4fc18f/artifacts/pytorch-model")
  File "/Users/sid.murching/code/mlflow/mlflow/pytorch/__init__.py", line 451, in load_model
    return _load_model(path=torch_model_artifacts_path, **kwargs)
  File "/Users/sid.murching/code/mlflow/mlflow/pytorch/__init__.py", line 393, in _load_model
    return torch.load(model_path, **kwargs)
  File "/Users/sid.murching/miniconda3/envs/mlflow-python3/lib/python3.6/site-packages/torch/serialization.py", line 592, in load
    return _load(opened_zipfile, map_location, pickle_module, **pickle_load_args)
  File "/Users/sid.murching/miniconda3/envs/mlflow-python3/lib/python3.6/site-packages/torch/serialization.py", line 852, in _load
    result = unpickler.load()
AttributeError: Can't get attribute 'Net' on <module '__main__' (built-in)>

```

smurching:
It does seem users run into problems with cloudpickle quite often:
* https://github.com/mlflow/mlflow/issues/3199
* https://github.com/mlflow/mlflow/issues/3258 (underlying issue seems related to Docker, but user was confused by the fact that we call `torch.save` instead of saving the model's state_dict, which is the [recommended approach](https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-state-dict-recommended))

I wonder if we should try to support the `state_dict`-based persistence API & unify support for that & torchscript e.g something like:

```
import torch
import mlflow.pytorch
class MyCustomModelClass(torch.nn.Module):

# Train model
my_model = MyCustomModelClass(...)
model.train()
...
# Save model with MLflow. Saves both the model and its state dict. Generates a warning if unable to save the model via 
# torch.save, throws if saving the state_dict fails.
# Users can pass optional save_torchscript = True to additionally save the model in torchscript
mlflow.pytorch.log_model(my_model, "my-custom-path", save_torchscript=True)

# Load model back using the new recommended approach of loading via state dict.
# More robust, doesn't require having `MyCustomModelClass` in a file at some specific 
# path at model-load time, but puts the burden on the user to ensure `MyCustomModelClass` definition is available
mlflow.pytorch.load_model("my-custom-path", model_class=MyCustomModelClass)

# Or load pickled torch.save'd model back:
mlflow.pyotrch.load_model("my-custom-path")

```

shrinath-suresh:
@smurching For main scoped models using pickle i have added a new test case to validate the attribute error. Can you please review the code and lemme know your comments?

smurching:
@shrinath-suresh apologies for the delay & thanks for adding the test! MLflow's APIs are stable, so we unfortunately can't go with the solution currently in this PR as it will break existing user code, i.e. cause loading of main-scoped Pytorch models to fail.

To write it out, this is my current understanding of the use cases supported by MLflow, along with the corresponding approach recommended by Pytorch:

* Loading models back for retraining via `mlflow.pytorch.load_model`. For this use case, Pytorch recommends saving & loading models using their `state_dict` and class. This works well for users, as usually they’re loading models back in the context of existing training code/in the same Python environment in which the model was originally trained, so the class definition is readily available
* Loading models back for inference, e.g. for batch inference via `mlflow.pyfunc.spark_udf("/path/to/my/logged/pytorch/model"), or RESTful inference via `mlflow models serve /path/to/my/logged/pytorch/model`. Pytorch recommends using torchscript for inference, as it's both more performant and hermetic. The latter property is particularly important, as model inference often happens in different environments from model training, so having a fully self-contained model is critical.

Thus I’m wondering if the right approach is to try to use the right pytorch APIs/persistence format for the appropriate scenario. That is, if MLflow should try to both save the model’s state dict and its torchscript representation when `mlflow.pytorch.log_model` is called, so that `mlflow.pytorch.load_model` can attempt to load back the model using the state dict & a user-supplied class, and inference APIs like `mlflow models serve` can depend on the model’s self-contained torchscript representation. I think we'd ultimately need to still support saving/loading eager models via `mlflow.pytorch.log_model` and `mlflow.pytorch.load_model` to avoid breaking existing user code, but maybe we can extend these APIs to support state_dict and torchscript, as in the strawman proposal in https://github.com/mlflow/mlflow/pull/3408/#issuecomment-693013610. Note that I'm not wedded to the API proposed in my comment - for example, maybe an API like `mlflow.pytorch.load_state_dict(model_uri)` makes more sense than `mlflow.pytorch.load_model(model_uri, model_class=None)`

Let me know if that makes sense / what you think!

cc also @harupy

PR #3416, opened 2020-09-11
support h2o for R

yitao-li:

Support H2O models in R in the same way that they are currently supported in Python

## How is this patch tested?

unit tests in test-model-h2o.R

This PR implements the support for H2O models in R.

Components 

Interface 

Language 

Integrations


mlflow/R/mlflow/R/model-h2o.R
```diff
@@ -0,0 +1,88 @@
...
+  model_file <- basename(h2o_save_location)
+
+  settings <- list(
+    full_file = h2o_save_location,
+    model_file = model_file,
+    model_dir = model_data_path
+  )
+  yaml::write_yaml(settings, file.path(model_data_path, "h2o.yaml"))
+
+  version <- as.character(utils::packageVersion("h2o"))
```
lorenzwalthert, line 31: for xgboost we don't take patches because not every patch might exist in R / python xgboost. Should we do the same here?

mlflow/R/mlflow/R/model-h2o.R
```diff
@@ -0,0 +1,88 @@
...
+
+  conda_env <- if (!is.null(conda_env)) {
```
lorenzwalthert, line 33: This also exists in xgboost. I think we could factor this out into a function and use it in both xgboost and h2o and potentially others later.

mlflow/R/mlflow/R/model-h2o.R
```diff
@@ -0,0 +1,88 @@
...
+  h2o::h2o.loadModel(file.path(h2o_model_file_path, settings$model_file))
+}
+
+#' @export
+mlflow_predict.H2OModel <- function(model, data, ...) {
+  h2o_assert_installed()
+  as.data.frame(h2o::h2o.predict(model, h2o::as.h2o(data), ...))
+}
+
+h2o_assert_installed <- function() {
```
lorenzwalthert, line 84: this is also duplicated from xgboost and could be shared.

lorenzwalthert (commented):
Looks good overall. Not sure if there are some files in `tests/testthat` that should not be in version control.

lorenzwalthert:
To fix the built, I think you needed to install the h2o python package too, as we did for xgboost: https://github.com/mlflow/mlflow/blob/master/mlflow/R/mlflow/.run-tests.R#L11.
I suggest an additional test to see if you can print the loaded model (e.g. with `expect_error(print(model), NA)` (maybe `print()` can be omitted). I just realised that the xgboost loaded model cannot be printed in R because after I did my PR, the experiment id is also supposed to be logged in `MLmodel` (and that information is accessed). Since in the xgboost tests, we never print, we could not detect that problem. I'll provide a fix for this in another PR.

lorenzwalthert:
Another suggestion I have is to update [this](https://github.com/mlflow/mlflow/blob/d5a1c372bb4472ea9f040a2f2020beee1b8fd51e/docs/source/models.rst#h2o-h2o) doc page to say that h2o is now supported from R as well, similar to what we did for xgboost.

yitao-li:
@lorenzwalthert Hey thanks for the suggestions! At least all tests are passing now.

There were multiple gotchas that caused test failures in the first iteration of this PR:
- I did have `reticulate::conda_install("h2o", ...)`, but turns out I also needed the same for `h2o-py` (!!)
- `h2o` server's default port number is 54321, and another "randomly-chosen" port number in some test file was also 54321 (because great minds think alike), so that caused some odd, unexpected http requests/responses as well -- I think in another PR we should completely randomize all port numbers and also directory names to ensure isolation among the test cases

Re: code repetition and re-use: it will be trivial, assuming all `language/r` contributors of mlflow are familiar with the concept of [collate order](https://rdrr.io/cran/roxygen2/man/update_collate.html) in R and are OK with introducing a collate order in `mlflow/R/mlflow`. 

Why a collate order is needed : say if I have a `utils.R` file that contains helper functions needed by `model-*.R`, then the R interpreter needs to load `utils.R` before loading `model-*.R`. This is typically accomplished by having the `#` @include utils.R` annotation in `model-*.R`.

Meanwhile I'm working on your other suggestions now.

mlflow/R/mlflow/R/model-h2o.R
yitao-li: Looks like the answer is yes -- I don't see any guarantee anywhere that says the patch versions will be necessarily in sync between h2o-py and the R package of h2o.

lorenzwalthert:
Thanks for the explanation @yitao-li. I'll have another look once you make tests pass (I think we are not too far away). What I don't understand about the port numbers: Should we not aim for maximal isolation of the tests? I.e. can't we shut down the servers after tests are ran?

yitao-li:
@lorenzwalthert Tests are passing now :)


Re: shutting down servers -- yes completely agree that should be part of the PR for improving isolation among test cases.

Implementation-wise, I think we will need to explicitly specify (testthat >= 2.0) and use the teardown syntax described in https://www.tidyverse.org/blog/2017/12/testthat-2-0-0 for shutting down server eventually. I think that should suffice.

Picking different port numbers is still necessary to be completely fool-proof though, because shutting down a server won't necessarily free up the underlying TCP socket instantaneously (because the TCP TIME-WAIT state needs to last for 2*max segment life and things like that...)

javierluraschi (approved):

javierluraschi:
Great feature @yitao-li and review @lorenzwalthert!

mlflow/R/mlflow/R/model-h2o.R
```diff
@@ -0,0 +1,74 @@
...
+    model_file = model_file,
+    model_dir = model_data_path
+  )
+  yaml::write_yaml(settings, file.path(model_data_path, "h2o.yaml"))
+
+  # needed because R and python packages may not have the same patch number
+  version <- remove_patch_version(
+    as.character(utils::packageVersion("h2o"))
+  )
+
```
yitao-li, line 38: we did not require exact match on patch versions (or at least that was the plan originally : ) )

PR #3419, opened 2020-09-14
Some additional model handling to azureml.deploy

trangevi:

Adding some handling to mlflow.azureml.deploy to better align with handling within our tracking/registry plugins.

## How is this patch tested?

There are automated tests in `tests\azureml\test_deploy.py`, as well as more manual testing focused around the AML specific handling.

Deploying a model using `mlflow.azureml.deploy` now integrates better with the AzureML tracking/registry plugins if the user has been using them before calling deploy. Previously registered models will be reused instead of being registered again, and models provided from a run via a 'run://' model_uri will have proper run ID lineage.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/azureml/__init__.py
+++ b/mlflow/azureml/__init__.py
@@ -402,7 +402,7 @@ def deploy(
                     registered_model.name,
                     registered_model.version,
                 )
-        except Exception:
+        except Exception:  # pylint: disable=broad-except
             _logger.info(
                 "Unable to find model in AzureML with ID '%s', will register the model.", m_id
             )
--- a/mlflow/azureml/__init__.py
+++ b/mlflow/azureml/__init__.py
@@ -386,8 +386,8 @@ def deploy(
 
         try:
             if model_uri.startswith("models:/"):
-                m_name = model_uri.split('/')[-2]
-                m_version = int(model_uri.split('/')[-1])
+                m_name = model_uri.split("/")[-2]
+                m_version = int(model_uri.split("/")[-1])
                 m_id = "{}:{}".format(m_name, m_version)
                 registered_model = AzureModel(workspace, id=m_id)
 
@@ -397,19 +397,28 @@ def deploy(
                 m_id = "{}:{}".format(m.name, m.version)
                 registered_model = AzureModel(workspace, id=m_id)
 
-                _logger.info("Registered an Azure Model with name: `%s` and version: `%s`",
-                             registered_model.name, registered_model.version)
+                _logger.info(
+                    "Registered an Azure Model with name: `%s` and version: `%s`",
+                    registered_model.name,
+                    registered_model.version,
+                )
         except:
-            _logger.info("Unable to find model in AzureML with ID '{}', will register the "
-                         "model.".format(m_id))
+            _logger.info(
+                "Unable to find model in AzureML with ID '{}', will register the "
+                "model.".format(m_id)
+            )
 
         if not registered_model:
-            registered_model = AzureModel.register(workspace=workspace, model_path=tmp_model_path,
-                                                   model_name=model_name, tags=tags)
+            registered_model = AzureModel.register(
+                workspace=workspace, model_path=tmp_model_path, model_name=model_name, tags=tags
+            )
 
-            _logger.info("Registered an Azure Model with name: `%s` and version: `%s`",
-                         registered_model.name, registered_model.version)
-        
+            _logger.info(
+                "Registered an Azure Model with name: `%s` and version: `%s`",
+                registered_model.name,
+                registered_model.version,
+            )
+
         # Create an execution script (entry point) for the image's model server. Azure ML requires
         # the container's execution script to be located in the current working directory during
         # image creation, so we create the execution script as a temporary file in the current
--- a/mlflow/azureml/__init__.py
+++ b/mlflow/azureml/__init__.py
@@ -391,7 +391,7 @@ def deploy(
                 m_id = "{}:{}".format(m_name, m_version)
                 registered_model = AzureModel(workspace, id=m_id)
 
-                _logger.info("Found registered model in AzureML with ID '{}'".format(m_id))
+                _logger.info("Found registered model in AzureML with ID '%s'", m_id)
             elif model_uri.startswith("runs:/") and get_tracking_uri().startswith("azureml"):
                 m = mlflow_register_model(model_uri, model_name)
                 m_id = "{}:{}".format(m.name, m.version)
@@ -402,10 +402,9 @@ def deploy(
                     registered_model.name,
                     registered_model.version,
                 )
-        except:
+        except Exception:
             _logger.info(
-                "Unable to find model in AzureML with ID '{}', will register the "
-                "model.".format(m_id)
+                "Unable to find model in AzureML with ID '%s', will register the model.", m_id
             )
 
         if not registered_model:
--- a/mlflow/azureml/__init__.py
+++ b/mlflow/azureml/__init__.py
@@ -10,7 +10,7 @@ import uuid
 
 from distutils.version import StrictVersion
 
-from mlflow import get_tracking_uri
+from mlflow import get_tracking_uri, get_registry_uri
 from mlflow import pyfunc
 from mlflow import register_model as mlflow_register_model
 from mlflow.exceptions import MlflowException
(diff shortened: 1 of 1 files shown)


mlflow/azureml/__init__.py
```diff
@@ -380,16 +382,34 @@ def deploy(
...
+
+        try:
+            if model_uri.startswith("models:/"):
+                m_name = model_uri.split('/')[-2]
+                m_version = int(model_uri.split('/')[-1])
+                m_id = "{}:{}".format(m_name, m_version)
+                registered_model = AzureModel(workspace, id=m_id)
+
+                _logger.info("Found registered model in AzureML with ID '{}'".format(m_id))
+            elif model_uri.startswith("runs:/") and get_tracking_uri().startswith("azureml:/"):
```
eedeleon, line 395: We should check for startswith "azureml:" although this may not be compatible with our target of "azureml", 

startswith("azureml:") or  =="azureml" would be longer lived

mlflow/azureml/__init__.py
```diff
@@ -380,15 +382,41 @@ def deploy(
...
+                m = mlflow_register_model(model_uri, model_name)
+                m_id = "{}:{}".format(m.name, m.version)
+                registered_model = AzureModel(workspace, id=m_id)
+
+                _logger.info(
+                    "Registered an Azure Model with name: `%s` and version: `%s`",
+                    registered_model.name,
+                    registered_model.version,
+                )
+        except Exception:  # pylint: disable=broad-except
```
eedeleon, line 405: Do we need to catch Exception? If we could scope this down it will help against bugs leaking through from unhandled exceptions

eedeleon: I think we can pull the if elif out of the try to scope down possible buggy code

then 

if m_name, m_version, m_id are not none,

try:
 register_model
except AzureMLException or even Exception at this point

do the rest

mlflow/azureml/__init__.py
```diff
@@ -380,15 +382,41 @@ def deploy(
...
+            _logger.info(
+                "Unable to find model in AzureML with ID '%s', will register the model.", m_id
```
eedeleon, line 407: We should log the error info here so that users can send us it if they see a print that does not make sense

mlflow/azureml/__init__.py
```diff
@@ -380,15 +382,41 @@ def deploy(
...
+                m_name = model_uri.split("/")[-2]
+                m_version = int(model_uri.split("/")[-1])
+                m_id = "{}:{}".format(m_name, m_version)
+                registered_model = AzureModel(workspace, id=m_id)
+
+                _logger.info("Found registered model in AzureML with ID '%s'", m_id)
+            elif model_uri.startswith("runs:/") and get_tracking_uri().startswith("azureml"):
+                m = mlflow_register_model(model_uri, model_name)
+                m_id = "{}:{}".format(m.name, m.version)
+                registered_model = AzureModel(workspace, id=m_id)
```
eedeleon, line 398: My delete seems to have confused github.

We should try to scope down the try catch to only wrap this call with m_id

trangevi: Discussed offline.

eddiestudies (approved):
Looks good to me

mlflow/azureml/__init__.py
```diff
@@ -380,15 +382,45 @@ def deploy(
...
+        m_id = None
 
-        _logger.info(
-            "Registered an Azure Model with name: `%s` and version: `%s`",
-            registered_model.name,
-            registered_model.version,
-        )
+        if model_uri.startswith("models:/"):
+            m_name = model_uri.split("/")[-2]
+            m_version = int(model_uri.split("/")[-1])
```
dbczumar, line 390: The last component of `model_uri` may not always be an integer version. Users can specify the lifecycle stage of a model via the last component. For example, `models:/my-model/production` can be used to refer to the latest version of `my-model` in stage `production`.

Because the latest production version may change over time, `m_id` will not necessarily continue to refer to the same model version. A similar problem occurs if the user deletes a registered model and re-creates it with the same name: version `1` of `my-model` at time T is not guaranteed to refer to the same ML model as version `1` of `my-model` at time T + n if deletion and re-creation occurs in between.

Is `m_id` meant to be a unique identifier? Are we trying to enforce a 1:1 mapping between AzureModels and model versions in the AzureML Model Registry? It would be helpful to spell out the exact goals in the PR description.

dbczumar (changes requested):
@trangevi Thanks for filing this! Left a few comments / questions!

mlflow/azureml/__init__.py
```diff
@@ -380,15 +382,45 @@ def deploy(
...
+            m_id = "{}:{}".format(m_name, m_version)
+        elif model_uri.startswith("runs:/") and get_tracking_uri().startswith("azureml"):
+            m = mlflow_register_model(model_uri, model_name)
```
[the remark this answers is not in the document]
trangevi: It does not only support a runs:/ uri, it should support any of the valid artifact schemes that are in mlflow. However here we're just concerned with improving lineage, so we specifically care about it being a runs:/ uri for this case.

mlflow/azureml/__init__.py
trangevi: In AzureML, model ID is unique within a workspace, and is a composite of <name>:<version>. This doesn't necessarily mean that there is a 1:1 mapping between a model file and a model version, as the file could be registered multiple times, deleted and re-registered.etc.

Specifically for this however, we need to have the AzureML Model ID in order to deploy, that's just how the deployment functionality works currently. This function that I'm editing is a workaround that was introduced to allow MLFlow users a convenience method for deploying on Azure, and as such as limitations. I personally am fine with saying that one of those limitations is lack of support for `/production` type uris, however @akshaya-a may feel differently about that.

mlflow/azureml/__init__.py
```diff
@@ -380,15 +382,45 @@ def deploy(
...
 
-        _logger.info(
-            "Registered an Azure Model with name: `%s` and version: `%s`",
-            registered_model.name,
-            registered_model.version,
-        )
+        if model_uri.startswith("models:/"):
+            m_name = model_uri.split("/")[-2]
+            m_version = int(model_uri.split("/")[-1])
+            m_id = "{}:{}".format(m_name, m_version)
```
eedeleon: Is there a way to go from /production to /version @dbczumar ? We might be able to expand that out to avoid the issue you brought up

mlflow/azureml/__init__.py
```diff
@@ -380,15 +382,49 @@ def deploy(
...
-        _logger.info(
-            "Registered an Azure Model with name: `%s` and version: `%s`",
-            registered_model.name,
-            registered_model.version,
-        )
+            _logger.info(
+                "Registered an Azure Model with name: `%s` and version: `%s`",
+                registered_model.name,
+                registered_model.version,
+            )
```
dbczumar, line 404: I'm a bit confused by this logging statement. In this block, we're registering a model with the model registry, which I think is distinct from creating an `AzureModel`. Is this log message accurate?

mlflow/azureml/__init__.py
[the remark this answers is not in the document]
dbczumar: Got it. Can we add a comment to the code explaining why we filter on `runs:/` and `azureml` for interpretability later on?

mlflow/azureml/__init__.py
dbczumar: > In AzureML, model ID is unique within a workspace, and is a composite of :. This doesn't necessarily mean that there is a 1:1 mapping between a model file and a model version, as the file could be registered multiple times, deleted and re-registered.etc.

The model ID uniqueness part of this concerns me though. With the logic implemented in this PR, what happens if a user deploys a model with name `A`, deletes the associated registered model, and then tries to redeploy with name `A`? In that case, it seems like we'll pull down the old AzureModel because the ID matches and end up deploying content that the user didn't specify.

Also, do users expect to have their models registered with the MLflow Model Registry on Azure at deployment time? We should at least document this side-effect, but I wonder if that's the intended UX.

dbczumar: What are the drawbacks to using a UUID for the model ID instead?

mlflow/azureml/__init__.py
trangevi: This is the inherent issue with representing a cloud object. Registering the model with the model registry does create the azure model, but it's a cloud object, not an in memory object.

When we register the model, it creates a cloud object in the azure model registry, and then because we're registering through MLFlow, that function returns an MLFlow SDK model object. However, we need an AzureML SDK model object to be able to deploy (at this time, that will change once we fully integrate with the deployment handling in MLFlow), so we end up getting the AzureML SDK object in order to do the deployment (the AzureML SDK uses the object constructor as a getter, it's a long story).

mlflow/azureml/__init__.py
trangevi: A UUID can not be used for ID, the ID is not something that is set client side, it's set service side, and is a composite of name and version as I mention, that is out of my control.

If a user deploys a model with name A version 1, and then deletes that model, it won't exist anywhere at that point, so it couldn't pull down the old model. Unless I'm not understanding your concern correctly?

And yes, it is expected that the model will be registered at deployment time if it's not already. At the very least it is the expectation for our users and has been from the beginning.

dbczumar (approved):
LGTM! I've confirmed that deployment from models and runs URIs succeeds with these patches applied. Thanks @trangevi !

PR #3436, opened 2020-09-18
Adding support for logging additional artifacts when using mfllow pytorch log and save model

shrinath-suresh:

In the current version of mlflow pytorch log/save model, there is no mechanism available to push artifact files along with the pytorch model. The supporting files such as requirements.txt or vocabulary files would be really helpful in case of deployments or new environment setups. Hence, adding a new parameter artifacts to mlflow pytorch log and save model which would help user to log supporting files into mlflow.

## How is this patch tested?

Tested the code by logging artifact files along with log and save model. And also ensured that the existing scenario is working as expected(saving and logging model without artifact parameter). Added unit test cases as well.

Users can log additional artifact files such as requirements, vocabulary files along with the pytorch model when using mlflow.pytorch.log_model or mfllow.pytorch.save_model by adding them in artifacts parameter.

Components 

Interface 

Language 

Integrations


harupy:
@shrinath-suresh How would this PR simplify the deployments or new environment setups?

shrinath-suresh:
> @shrinath-suresh Could you clarify how this PR would simplify the deployments or new environment setups?

The use where we need to deploy bert sentiment analsyis with pytorch/pytorch lightning using torchserve. 

Along with the model, there are some extra files which would be really useful for deployment.  Ex: the vocabulary file generated out of the model or the requirements.txt file.Using  the artifacts parameter these files can be pushed along with the model.

On the deployment side, we can easily extract these files from the artifacts folder(say `models/artifacts/requirements.txt`) and perform necessary actions.

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +814,84 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+
+
+@pytest.fixture
+def create_artifact(tmpdir):
+    artifact_file_name = "requirements.txt"
+    fp = tmpdir.mkdir("artifacts").join(artifact_file_name)
+    test_string = "pytest"
+    fp.write(test_string)
+    yield fp
+    shutil.rmtree(str(tmpdir))
```
harupy, line 826: Can we remove this line (`shutil.rmtree(str(tmpdir))`)?

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +814,84 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+            conda_env=None,
+            artifacts={"requirements.txt": str(artifact_file_path)},
+        )
+
+        model_uri = "runs:/{run_id}/{model_path}".format(
+            run_id=mlflow.active_run().info.run_id, model_path="models"
+        )
+        with TempDir(remove_on_exit=True) as tmp:
+            model_path = _download_artifact_from_uri(model_uri, tmp.path())
+            assert os.path.isdir(os.path.join(model_path, "artifacts")) is True
```
harupy, line 844:
```suggestion
            assert os.path.isdir(os.path.join(model_path, "artifacts"))
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +814,84 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+    with mlflow.start_run(), pytest.raises(FileNotFoundError):
+        mlflow.pytorch.log_model(
+            pytorch_model=sequential_model,
+            artifact_path="models",
+            conda_env=None,
+            artifacts={
+                "requirements.txt": str(artifact_file_path).replace(
+                    "requirements", "requirements1"
+                )
+            },
```
harupy, line 884:
```suggestion
            artifacts={"requirements.txt": "inexistent_file.txt"},
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +814,84 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+def test_log_model_invalid_type(create_artifact, sequential_model):
+    artifact_file_path = create_artifact
+    with mlflow.start_run(), pytest.raises(TypeError) as exc_info:
+        mlflow.pytorch.log_model(
+            pytorch_model=sequential_model,
+            artifact_path="models",
+            conda_env=None,
+            artifacts=str(artifact_file_path).replace("requirements", "requirements1"),
+        )
+        assert "Argument artifacts should be a list" in str(exc_info)
```
harupy, line 897: Can you use the `match` argument here instead of `assert "Argument artifacts should be a list" in str(exc_info)`?

https://docs.pytest.org/en/stable/reference.html#pytest-raises

mlflow/pytorch/__init__.py
```diff
@@ -325,6 +357,23 @@ def save_model(
         f.write(pickle_module.__name__)
     # Save pytorch model
     model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
+    if artifacts:
+        if not isinstance(artifacts, dict):
+            raise TypeError("Argument artifacts should be a dict")
+
+        with TempDir() as tmp_artifacts_dir:
+            tmp_artifacts_config = {}
```
harupy, line 365: Do we need this variable (`tmp_artifacts_config`) which doesn't seem to be used?

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +814,84 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+    artifact_path = create_artifact
+    with mlflow.start_run(), TempDir(remove_on_exit=True) as tmp:
+        model_path = os.path.join(tmp.path(), "models")
+        mlflow.pytorch.save_model(
+            pytorch_model=sequential_model,
+            path=model_path,
+            artifacts={"requirements.txt": str(artifact_path)},
+        )
+
+        assert os.path.isdir(os.path.join(model_path, "artifacts")) is True
```
harupy, line 864:
```suggestion
        assert os.path.isdir(os.path.join(model_path, "artifacts"))
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +814,84 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+        requirements_file = os.path.join(model_path, "artifacts", "requirements.txt")
+        assert os.path.isfile(requirements_file) is True
```
harupy, line 866:
```suggestion
        assert os.path.isfile(requirements_file)
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +814,84 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+        model_uri = "runs:/{run_id}/{model_path}".format(
+            run_id=mlflow.active_run().info.run_id, model_path="models"
+        )
+        with TempDir(remove_on_exit=True) as tmp:
+            model_path = _download_artifact_from_uri(model_uri, tmp.path())
+            assert os.path.isdir(os.path.join(model_path, "artifacts")) is True
+            requirements_file = os.path.join(
+                model_path, "artifacts", "requirements.txt"
+            )
+            assert os.path.isfile(requirements_file) is True
```
harupy, line 848:
```suggestion
            assert os.path.isfile(requirements_file)
```

shrinath-suresh:
@harupy Addressed the review comments.

shrinath-suresh:
> @harupy Addressed the review comments.

On the linting failure. pylint and pycodestyle issues are resolved. I need to run black command. Should i do it in same review or raise a separate PR for the same ?

harupy:
> Should i do it in same review or raise a separate PR for the same ?

It should be done in this review :)

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +813,74 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
     np.testing.assert_array_almost_equal(
         deployed_model_preds.values[:, 0], sequential_predicted, decimal=4
     )
+
+
+@pytest.fixture
+def create_artifact(tmpdir):
+    artifact_file_name = "requirements.txt"
+    fp = tmpdir.mkdir("artifacts").join(artifact_file_name)
```
harupy, line 821:
```suggestion
    fp = tmpdir.join(artifact_file_name)
```

Do we need `mkdir("artifacts")` here?

shrinath-suresh: Done

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -813,3 +813,74 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
     )
+
+
+@pytest.fixture
+def create_artifact(tmpdir):
+    artifact_file_name = "requirements.txt"
+    fp = tmpdir.join(artifact_file_name)
+    test_string = "pytest"
+    fp.write(test_string)
+    yield fp
```
harupy, line 824:
```suggestion
    return fp
```

shrinath-suresh: done..

harupy:
@shrinath-suresh One of the aims of this PR is make it easier associate files required for [torch-model-archiver](https://github.com/pytorch/serve/tree/master/model-archiver#arguments) (e.g. `extra-files`, `requirements-file`). Is that correct?

shrinath-suresh:
> @shrinath-suresh One of the aims of this PR is make it easier to associate files required for [torch-model-archiver](https://github.com/pytorch/serve/tree/master/model-archiver#arguments) (e.g. `extra-files`, `requirements-file`) to the model. Is that correct?

Yes. You are right. It associates files required for `extra-files` argument. Apart from that, it also enables an option to push any other artifact file along with the model. For ex: (.mar file generated during the deployment process.. currently it doesnt work that way.. but why not ?)

harupy:
Thanks for the quick reply. Is `torch-model-archiver` something we must do before serving a model?

shrinath-suresh:
> Thanks for the quick reply. Is `torch-model-archiver` something we must do before serving a model?

Its a two step process. First step is to generate mar file using `torch-model-archiver' and the second step is to serve the model. https://github.com/pytorch/serve - Commands mentioned in `Store a model` section.

harupy:
btw does `torch-model-archiver` accept a model serialized by pickle? It looks like it requires a state_dict. `mlflow.pytorch.log_model` doesn't log a state_dict.

shrinath-suresh:
> btw does `torch-model-archiver` accept a model serialized by pickle? It looks like it doesn't and requires a state_dict instead. `mlflow.pytorch.log_model` doesn't log a state_dict.

We have tested couple of model deployments by logging the model into mlflow using `mlflow.pytorch.log_model`. The deployment process works with `mlflow.pytorch.log_model`. There is a separate activity going on for the plugin and its test.

harupy:
`mlflow.pytorch.log_model` currently serialize a model using pickle but `torchserve` requires state_dict or torchscript 
 model. You're trying to address this issue in https://github.com/chauhang/mlflow/pull/16. Is that correct

shrinath-suresh:
> @shrinath-suresh
> `mlflow.pytorch.log_model` currently serialize a model using pickle but `torchserve` requires state_dict or torchscript
> model. You're trying to address this issue in [chauhang#16](https://github.com/chauhang/mlflow/pull/16). Is that correct?

`mlflow.pytorch.log_model` currently uses `cloudpickle` to serialize the model. 

https://github.com/mlflow/mlflow/pull/3408 - This PR is to remove the dependency on cloudpickle and set pickle as default serialization module.

https://github.com/chauhang/mlflow/pull/16 - Currently there is no mechanism available in `mlflow.pytorch` library to log the `torch scripted` model. This PR is to log, load and save `torch scripted` models using `mlflow.pytorch` library.

shrinath-suresh:
@harupy Can you please review and approve the PR if you dont have any more comments?

smurching:
Thanks @shrinath-suresh for all the explanations! @harupy's gonna take another pass.

We had one follow-up question that we can maybe address after merging this PR, but just wanted to make sure we'd thought it through. @harupy & I read through the Torchserve example at  https://aws.amazon.com/blogs/aws/announcing-torchserve-an-open-source-model-server-for-pytorch/ - it makes sense that users may want to log extra artifacts alongside their pytorch model so that they can be made available at serving time. We're wondering how exactly this user workflow will work, for example if I train a model and log extra files using `mlflow.pytorch.log_model(model, artifacts={"vocab_file.json": "/path/to/vocab_file.json"})`, how will Torchserve deploy plugin logic know to pass `vocab_file.json` as `--extra-files` when ultimately invoking the torch-model-archiver CLI?

Do we want the TorchServe plugin to try to infer which extra files to include by convention? For example, given a command like `mlflow deployments create -t torchserve -n mymodel s3:/my-cool-bucket/model`, maybe the TorchServe plugin would include all files under `s3:/my-cool-bucket/model`  besides the `MLmodel` file and the persisted model itself as `--extra-files` when generating the model archive, and so automatically capture `vocab_file.json` above?

shrinath-suresh:
> Thanks @shrinath-suresh for all the explanations! @harupy's gonna take another pass :)
> 
> We had one follow-up question that we can maybe address after merging this PR, but just wanted to make sure we'd thought it through. @harupy & I read through the Torchserve example at https://aws.amazon.com/blogs/aws/announcing-torchserve-an-open-source-model-server-for-pytorch/ - it makes sense that users may want to log extra artifacts alongside their pytorch model so that they can be made available at serving time. We're wondering how exactly this user workflow will work, for example if I train a model and log extra files using `mlflow.pytorch.log_model(model, artifacts={"vocab_file.json": "/path/to/vocab_file.json"})`, how will the Torchserve deploy plugin know to pass `vocab_file.json` as `--extra-files` when ultimately invoking the `torch-model-archiver` CLI?
> 
> Do we want the TorchServe plugin to try to infer which extra files to include by convention? For example, given a command like `mlflow deployments create -t torchserve -n mymodel -m s3:/my-cool-bucket/model`, maybe the TorchServe plugin would include all files under `s3:/my-cool-bucket/model` besides the `MLmodel` file and the persisted model itself as `--extra-files` when generating the model archive, and so automatically capture `vocab_file.json` above?

Thank you very much for the detailed question. The user flow would be as follows.

1. The `artifacts` parameter enables user to log extra files along with the model. 
2. On Torchserve plugin, the extra files will be derived as follows
    2.1 Derive the list of files under `artifacts` folder and add it to `extra-files` argument
    2.2. Allow user to pass extra files using config and add it as extra files argument - This is needed for a scenario where the extra files were initially missed during log model and later needed for prediction. 
3. In the torchserve handler file, locate the extra files which were added and perform prediction. 

I am creating a simple example to demonstrate this end to end workflow. will keep you posted on the same.

shrinath-suresh:
@harupy @smurching I have added an example for [save-artifacts with torchserve-deployment plugin](https://github.com/chauhang/mlflow/tree/feature-torchserve-plugin/mlflow/pytorch/torchserve/examples/MNIST/example3)

Since, the save-artifacts and predict method implementation are in pending PR state, the example needs to be tested as follows.

1.  Clone save-artifacts branch - https://github.com/shrinath-suresh/mlflow/tree/pytorch-save-artifacts  
2. Build mlflow from source using `pip install .` 
3. Clone the torchserve-plugin repo - https://github.com/chauhang/mlflow/tree/feature-torchserve-plugin
4. move to `mlflow/mlflow/pytorch/torchserve/examples/MNIST/example3` and run the model file. Once training process is complete, model file and the mapping file (extra artifact file) will be pushed into mlflow. 
5. Move to the home of `feature-torchserve-plugin` and once again build mlflow from source - It is needed for predict method implementation of torchserve deployment plugin
6. Install torchserve-plugin and  run create , predict commands - setup and details available in Readme.md file - https://github.com/chauhang/mlflow/blob/feature-torchserve-plugin/mlflow/pytorch/torchserve/examples/MNIST/example3/README.md


Please find the commit for supporting extra files on deployment plugin - https://github.com/chauhang/mlflow/pull/8/commits/a9e4b4ecd198a316399c89ac1520eb318cd61dbe

harupy:
@shrinath-suresh Thanks for the example. With the current design, a user can log `extra-files` and `requirement-file` and the torchserve plugin automatically detects them. Can we do the same thing for `source-vocab`?

shrinath-suresh:
> @shrinath-suresh Thanks for the example. With the current imlementation, a user can log `extra-files` and `requirement-file` and the torchserve plugin automatically detects them by matching their names.

Yes. It automatically detects the extra files and requirement during deployment phase.

harupy:
@shrinath-suresh Today, Sid and I discussed an alternative approach:

```python
# Add new arguments to `log_model`

def log_model(
    model,
    ...,
    # Log these parameters in the MLmodel file
    # so we can look them up later in the torchserve plugin.
    model_file: str = None,
    handler: str = None,
    source_vocab: str = None,
    extra_files: List[str] = None,
    requirements_file: str = None,
):
    ...
```

- Pros:
  - Users don't need to rename files
  - Easier to look up later.
- Cons:
  - These parameters are specific to torchserve.
  - `log_model` may end up having lots of parameters if we later want to integrate other deployment tools.


What do you think?

shrinath-suresh:
> @shrinath-suresh Today, Sid and I discussed alternative approaches:
> 
> ### Option 1:
> ```python
> # Add new arguments to `log_model`
> 
> def log_model(
>     model,
>     ...,
>     # Log these parameters in the MLmodel file
>     # so we can look them up later in the torchserve plugin.
>     model_file: str = None,
>     handler: str = None,
>     source_vocab: str = None,
>     extra_files: List[str] = None,
>     requirements_file: str = None,
> ):
>     ...
> ```
> 
> * Pros:
>   
>   * Users don't need to rename files
>   * Easier to look up files later.
> * Cons:
>   
>   * These parameters are specific to torchserve. `log_model` may end up having lots of parameters if we later want to integrate other deployment tools.
> 
> ### Option 2:
> ```python
> # Wrapper API
> 
> # inject TorchServe specific configs via another API
> mlflow_model = mlflow.pytorch.add_torchserve_config(
>     model,
>     requirements_file="requirements.txt",
>     source_vocab="vocab.json",
> )
> 
> # log model
> mlflow.pytorch.log_model(mlflow_model, path)
> ```
> 
> * Pros:
>   
>   * Can address the con in option 1
> * Cons:
>   
>   * Users need to call `mlflow.pytorch.add_torchserve_config` to wrap a model.
> 
> What do you think?

As you mentioned, option1 would add too many parameters into `mlflow.pytorch.log_model` and option 2 looks interesting. But it is again too specific to torchserve and user has to call this method in every script before logging the model.  

Extra `artifacts` parameter is already available as part of [mlflow pyfunc model](https://github.com/mlflow/mlflow/blob/33792e6f11eb636e26bd7c6616e75f97e1de4832/mlflow/pyfunc/model.py#L149) and we have incorporated the same in pytorch. `artifacts` parameter is generic and user can log any type of file in this parameter. We will discuss on all three options and get back shortly.

harupy:
@shrinath-suresh 

> artifacts parameter is generic and user can log any type of file in this parameter.

I agree on this point, but the current approach seems error prone.
For example, the following code works and the user wouldn't notice the mistake until deploying the model.


```python
# "requirement.txt" is incorrect, should be "requirements.txt".
mlflow.pytorch.log_model(model, artifacts={"req": "requirement.txt"})

# later he noticed the torchserve plugin doesn't detect the requirement file because the file name is incorrect.
```

shrinath-suresh:
> ```python
> requirement
> ```

Yes. I agree to your point. `artifacts` parameter enforces the user to rename the requirement file to `requirements.txt`. Option 2 looks more reasonable as it is focused on torchserve specific arguments. We will evaluate all 3 designs. Thanks for the detailed suggestions.

shrinath-suresh:
@smurching @harupy We have updated the code. log and save model takes `requirements_file` and `extra_files` arguments for logging additional artifacts. Please reivew and let us know your comments.

mlflow/pytorch/__init__.py
```diff
@@ -325,6 +376,34 @@ def save_model(
...
+        with TempDir() as tmp_requirements_dir:
+            saved_requirements_dir_subpath = "requirements"
+            _download_artifact_from_uri(
+                artifact_uri=requirements_file, output_path=tmp_requirements_dir.path()
+            )
+            shutil.move(
+                tmp_requirements_dir.path(), posixpath.join(path, saved_requirements_dir_subpath)
+            )
+
+    if extra_files and len(extra_files) > 0:
```
shrinath-suresh, line 393: Single check is enough i believe. will remove `len(extra_files) > 0` in the next commit

mlflow/pytorch/__init__.py
```diff
@@ -325,6 +376,34 @@ def save_model(
         f.write(pickle_module.__name__)
     # Save pytorch model
     model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
+
+    if requirements_file:
+        if not isinstance(requirements_file, str):
+            raise TypeError("Path to requirements file should be a string")
+
+        with TempDir() as tmp_requirements_dir:
```
smurching, line 384: The logic here seems correct, but just wondering - do people tend to log requirements.txt files from remote artifact stores, e.g. s3, as part of their pytorch models?

We could simplify this by requiring requirements.txt to be a file on the local filesystem. With the current logic, in the seemingly-common case where requirements.txt is in the local filesystem, we'll copy it to another temporary directory on the same local filesystem, which seems a bit redundant

mlflow/pytorch/__init__.py
```diff
@@ -325,6 +376,34 @@ def save_model(
...
+
+        with TempDir() as tmp_extra_files_dir:
+            saved_extra_files_dir_subpath = "artifacts"
+            for extra_file in extra_files:
+                _download_artifact_from_uri(
+                    artifact_uri=extra_file, output_path=tmp_extra_files_dir.path()
+                )
+            shutil.move(
+                tmp_extra_files_dir.path(), posixpath.join(path, saved_extra_files_dir_subpath)
+            )
```
smurching, line 405: I have the same question as https://github.com/mlflow/mlflow/pull/3436/files#r513084383 for `extra_files` (do people log `extra_files` stored remotely in S3 alongside their pytorch models, or are the extra_files usually on the local filesystem)? Happy to defer to your judgment here on what's common for pytorch users

smurching (commented):
Thanks for the updates @shrinath-suresh! I just had two questions, otherwise this looks good - thanks for the hard work on the PR!

mlflow/pytorch/__init__.py
```diff
@@ -268,6 +298,27 @@ def save_model(
                           serialized to json using the Pandas split-oriented format. Bytes are
                           base64-encoded.
 
+    :param requirements_file: A string containing the path to requirements file. Remote URIs
```
harupy, line 301: Can we store `requirements_file` and `extra_files` in the `MLmodel` file so we can look them up later?

mlflow/pytorch/__init__.py
smurching: From offline discussion with @chauhang:

It varies based on the files. The requirements.txt will generally be local or from github. The vocab files will be from a shared storage like S3. Other files will be very model specific so we won't know where they will come from. Hence we need support for remote download of the files if they are not local

Thus let's stick with the logic as is for now - we can iterate on it later if we need to optimize for whatever reason

mlflow/pytorch/__init__.py
chauhang: @harupy Adding to MLmodel has performance implications during model inference loading so these files need to be kept as separate artifacts, but bundled along with the model for version tracking and matching artifacts

harupy: @charnger 

> these files need to be kept as separate artifacts

Yes, they will be kept as separate files and `MLmodel` file (which contains metadata of a logged model) should contains information of where they are located like below.

```yml
# Example of MLmodel file
artifact_path: models
flavors:
  python_function:
    data: data
    env: conda.yaml
    loader_module: mlflow.pytorch
    pickle_module_name: mlflow.pytorch.pickle_module
    python_version: 3.6.10
  pytorch:
    model_data: data
    pytorch_version: 1.4.0
    # new fields
    requirements_file: ...
    extra_files:
      - ...
      - ...
run_id: cd8c2872466c463790bc4ac70b9d161f
utc_time_created: '2020-10-27 23:22:47.135800'

```

smurching (approved):
Discussed offline with @harupy - let's go ahead & merge this, and make follow-up changes (like including metadata about the location of the logged artifacts in the MLmodel file) in a follow-up PR. Thanks @shrinath-suresh @chauhang for the hard work on this PR!

PR #3443, opened 2020-09-20
[ML-12132] Fix persistence of Spark models on passthrough-enabled environments

smurching:

Fixes ``mlflow.spark.log_model``, ``mlflow.spark.load_model`` APIs on passthrough-enabled environments against ACL'd artifact locations.

Previously, we had the following behavior:
* `mlflow.spark.save_model`: saves spark model to temporary directory on Spark's default distributed FS (e.g. DBFS on Databricks), then copies the persisted spark model directory to the desired local path using HDFS APIs called via Py4j. The latter step fails in passthrough environments as HDFS APIs aren't allowlisted
* `mlflow.spark.log_model`: attempts to write directly to the destination directory, which fails for ACL'd artifact locations. Then, falls back to calling `save_model` & `log_artifacts`, which fails as described above
* `mlflow.spark.load_model`: attempts to load model directly from the artifact location via Spark. If this fails, downloads model files locally, uploads them to Spark's distributed FS via HDFS APIs (fails due to lack of allowlisting), and then loads Spark model from the distributed FS

This PR makes the following changes when running on Databricks, for both passthrough-enabled & non-passthrough envs:
* `mlflow.spark.save_model`: copy model files locally using DBFS FUSE instead of denylisted HDFS APIs. This also fixes ``log_model``
* `mlflow.spark.load_model`: upload locally-downloaded model files to Spark's distributed FS via DBFS FUSE instead of denylisted HDFS APIs.

## How is this patch tested?

Manually

Components 

Interface 

Language 

Integrations


mlflow/spark.py
```diff
@@ -501,7 +508,14 @@ def save_model(
     tmp_path = _tmp_path(dfs_tmpdir)
     spark_model.save(tmp_path)
     sparkml_data_path = os.path.abspath(os.path.join(path, _SPARK_MODEL_PATH_SUB))
-    _HadoopFileSystem.copy_to_local_file(tmp_path, sparkml_data_path, remove_src=True)
+    # If spark DFS is DBFS and we're running on a Databricks cluster, copy to local FS
+    # via the FUSE mount
+    is_saving_to_dbfs = is_valid_dbfs_uri(tmp_path) or posixpath.abspath(tmp_path) == tmp_path
+    if is_saving_to_dbfs and databricks_utils.is_in_cluster():
+        tmp_path_fuse = dbfs_hdfs_uri_to_fuse_path(tmp_path)
+        shutil.move(src=tmp_path_fuse, dst=sparkml_data_path)
```
edgan8, line 516: When does the file get written to tmp_path fuse so we can move it? Are we assuming that the FUSE mount has been synced with HDFS? What if the sync hasn't finished yet?

mlflow/spark.py
```diff
@@ -513,17 +527,43 @@ def save_model(
...
+    local_model_path = _download_artifact_from_uri(model_uri)
     # Spark ML expects the model to be stored on DFS
     # Copy the model to a temp DFS location first. We cannot delete this file, as
     # Spark may read from it at any point.
-    model_path = _HadoopFileSystem.maybe_copy_from_uri(model_uri, tmp_path)
-    return PipelineModel.load(model_path)
+    fuse_dfs_tmpdir = dbfs_hdfs_uri_to_fuse_path(dfs_tmpdir)
+    os.mkdir(fuse_dfs_tmpdir)
+    # Workaround for inability to use shutil.copytree with DBFS FUSE due to permission-denied
+    # errors on passthrough-enabled clusters when attempting to copy permission bits for directories
```
edgan8, line 541: Can we factor out this copytree functionality into a more generic helper function?

mlflow/spark.py
```diff
@@ -501,7 +508,14 @@ def save_model(
     tmp_path = _tmp_path(dfs_tmpdir)
     spark_model.save(tmp_path)
     sparkml_data_path = os.path.abspath(os.path.join(path, _SPARK_MODEL_PATH_SUB))
-    _HadoopFileSystem.copy_to_local_file(tmp_path, sparkml_data_path, remove_src=True)
+    # If spark DFS is DBFS and we're running on a Databricks cluster, copy to local FS
+    # via the FUSE mount
+    is_saving_to_dbfs = is_valid_dbfs_uri(tmp_path) or posixpath.abspath(tmp_path) == tmp_path
```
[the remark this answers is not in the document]
smurching: Good Q, it's because persisting a spark model to a path like "/my-directory" actually saves to "dbfs:/my-directory".

I'll add this explanation/clarifying comment to the code, lmk if it makes sense:

```
    # If we're running on a Databricks cluster, we can detect whether the tempdir we're saving our
    # Spark model to is on DBFS, by checking if it either starts with dbfs:/
    # e.g. "dbfs:/my-directory" or is an absolute path like "/my-directory" (in which case Spark
    # saves to the default distributed FS => DBFS). If we're using a DBFS
    # tempdir, copy to local FS via the FUSE mount. Otherwise, use HDFS APIs to copy the model to
    # the local FS.
```

mlflow/spark.py
smurching: In this save_model method, the goal is to save the SparkML model to a local filesystem path

The flow here is:
* Save SparkML model to a temporary dbfs location (`dbfs:/tmp/my-model`)
* Move that file to the local filesystem using FUSE and a file move (move `/dbs/tmp/my-model` to the `/my/desired/local/path`)

We leverage FUSE to copy from DBFS -> local as there's no allowlisted JVM-side, HDFS API for copying from dbfs:/... to the local filesystem

smurching: The FUSE access has the same consistency properties as standard HDFS-style access to the same DBFS path. That is, writing to `dbfs:/tmp/my-model` and reading `/dbfs/tmp/my-model` is equivalent from a blob-storage-eventual-consistency perspective to writing to `dbfs:/tmp/my-model` and subsequently reading `dbfs:/tmp/my-model`

mlflow/spark.py
```diff
@@ -513,17 +530,50 @@ def save_model(
...
+def _shutil_copytree_without_file_permissions(src_dir, dst_dir):
+    """
+    Copies the directory src_dir into dst_dir, without preserving filesystem permissions
+    """
+    for (dirpath, dirnames, filenames) in os.walk(src_dir):
+        for dirname in dirnames:
+            relative_dir_path = os.path.relpath(os.path.join(dirpath, dirname), src_dir)
+            # Compute corresponding FUSE path of each local directory and create an equivalent
+            # FUSE directory
+            fuse_dir_path = os.path.join(dst_dir, relative_dir_path)
```
edgan8, line 542: This is no longer specific to fuse, it's just "absolute_dir_path" right?

smurching: 👍 yep that's right! I also added a unit test for this method, although it's an internal one used only in the `mlflow.spark` module, because it's somewhat complex

mlflow/spark.py
```diff
@@ -198,11 +207,11 @@ def log_model(
...
             await_registration_for=await_registration_for,
         )
-    # If Spark cannot write directly to the artifact repo, defer to Model.log() to persist the
-    # model
     model_dir = os.path.join(run_root_artifact_uri, artifact_path)
+    # Try to write directly to the artifact repo via Spark. If this fails, defer to Model.log()
+    # to persist the model
     try:
-        spark_model.save(os.path.join(model_dir, _SPARK_MODEL_PATH_SUB))
+        spark_model.save(posixpath.join(model_dir, _SPARK_MODEL_PATH_SUB))
```
smurching, line 214: Note: should be a posixpath join since HDFS URIs are posixpath style (e.g. we don't want to include backward slashes `\` here on platforms like Windows)

edgan8 (approved):
Nice refactoring! Can you update the PR description to clarify that this is changing dbfs access from HDFS to FUSE for all clusters (not just the passthrough ones that motivated this)?

PR #3449, opened 2020-09-22
Add input example and model signature options to Sklearn, XGBoost, LightGBM autologging, and change LightGBM input example and signature implementation

andrewnitu:
Signed-off-by: Andrew Nitu <andrewnitu@gmail.com>

1.
`Add configuration options log_input_example and log_model_signature to Sklearn, XGBoost, and LightGBM autologging.`
`By default log_input_example is False (so the user does not inadvertently log sensitive data) and log_model_signature is True.`
`The two options work independently of one another.`

2.
`Pull out the input example and signature inference logic to a separate module, where it can be easily tested for the correct behaviour`

3.
`Convert LightGBM autologging implementation to store the example input on the Dataset object, similarly to how it is implemented for xgboost. This allows a Dataset object to be reused for multiple trainings and still provide input_example and signature to each of those trained models, whereas before this was only possible once.`

## How is this patch tested?

- New unit tests for new configuration logic
- New unit tests for new input example and signature logic
- Unit tests for lightGBM continue passing
- Manual end-to-end testing for all 3 changes

**There are multiple changes in this PR, but this is the only user-facing one:**
Added configuration options log_input_example and log_model_signature to sklearn, XGBoost, and LightGBM autologging. By default, log_input_example is set to False and log_model_signature is set to True.

Components 

Interface 

Language 

Integrations


tests/xgboost/test_xgboost_autolog.py
```diff
@@ -454,3 +461,42 @@ def test_xgb_autolog_does_not_break_dmatrix_serialization(bst_params, tmpdir):
...
+    X = pd.DataFrame(iris.data[:, :2], columns=iris.feature_names[:2])
+    y = iris.target
+
+    with mlflow.start_run() as run:
+        mlflow.xgboost.autolog(log_input_example=False, log_model_signature=False)
+        dataset = xgb.DMatrix(X, y)
+        xgb.train(bst_params, dataset)
+    model_conf = get_model_conf(run.info.artifact_uri)
+    assert "saved_input_example_info" not in model_conf.to_dict()
+    assert "signature" not in model_conf.to_dict()
```
harupy, line 478: We can use `pytest.mark.parametrize` here.

```suggestion
@pytest.mark.large
@pytest.mark.parametrize("log_input_example", [True, False])
@pytest.mark.parametrize("log_model_signature", [True, False])
def test_lgb_autolog_configuration_options(bst_params):
    iris = datasets.load_iris()
    X = pd.DataFrame(iris.data[:, :2], columns=iris.feature_names[:2])
    y = iris.target
    
    assert ("saved_input_example_info" in model_conf.to_dict()) == log_input_example
    assert ("signature" in model_conf.to_dict()) == log_model_signature
```

harupy: With `pytest.mark.parametrize`, we can easily understand that this test verifies `mlflow.xgboost.autolog` works as expected for each `(log_input_example, log_model_signature)` combination.

mlflow/lightgbm.py
```diff
@@ -455,8 +459,9 @@ def log_feature_importance_plot(features, importance, importance_type):
             if input_example is None:
                 raise Exception("failed to gather example input.")
 
-            model_output = model.predict(input_example)
-            signature = infer_signature(input_example, model_output)
+            if log_model_signature:
+                model_output = model.predict(input_example)
+                signature = infer_signature(input_example, model_output)
         except Exception as e:  # pylint: disable=broad-except
             input_example = None
```
dbczumar, line 466: For lightgbm, xgboost, and sklearn, can we stop resetting this input example field to `None` in the case where signature is unspecified? I think these concepts are independent, especially given that they can be configured independently.

mlflow/lightgbm.py
```diff
@@ -292,6 +292,10 @@ def autolog():
         - inferred signature of the inputs and outputs of the model.
 
     Note that the `scikit-learn API`_ is not supported.
+
+    :param log_input_example: whether to log a sample of the training data
```
dbczumar, line 296: Nit: Are we using the full available line length here? the newline placement seems a bit odd

tests/config_helper.py
```diff
@@ -0,0 +1,2 @@
+def input_example_and_signature_on():
```
dbczumar, line 1: Is this necessary? I see we're passing it to most of the `mlflow.*.autolog()` calls throughout our test cases. In general, I think it makes sense to test the default set of arguments (i.e. leave the arguments unspecified) and only mutate the values when we're testing specific behavior. For example, `input_example_and_signature_on` doesn't seem essential for test cases like `test_lightgbm_autolog.py#test_lgb_autolog_ends_auto_created_run`.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -990,3 +992,43 @@ def test_autolog_does_not_throw_when_infer_signature_fails():
     model_conf = get_model_conf(run.info.artifact_uri)
     assert "signature" not in model_conf.to_dict()
     assert "saved_input_example_info" not in model_conf.to_dict()
+
+
+def test_autolog_configuration_options():
```
dbczumar, line 997: We should also adopt @harupy 's suggestion from https://github.com/mlflow/mlflow/pull/3449/files#r494000413 within this sklearn tests

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -423,3 +429,42 @@ def test_lgb_autolog_continues_logging_even_if_signature_inference_fails(tmpdir)
     assert data is not None
     assert "run_id" in data
     assert "signature" not in data
+
+
+@pytest.mark.large
+def test_lgb_autolog_configuration_options(bst_params):
```
dbczumar, line 435: We should also adopt @harupy 's suggestion from https://github.com/mlflow/mlflow/pull/3449/files#r494000413 in these lightgbm tests

tests/lightgbm/test_lightgbm_autolog.py
```diff
@@ -423,3 +429,42 @@ def test_lgb_autolog_continues_logging_even_if_signature_inference_fails(tmpdir)
...
+    iris = datasets.load_iris()
+    X = pd.DataFrame(iris.data[:, :2], columns=iris.feature_names[:2])
+    y = iris.target
+
+    with mlflow.start_run() as run:
+        mlflow.lightgbm.autolog(log_input_example=False, log_model_signature=False)
+        dataset = lgb.Dataset(X, y)
+        lgb.train(bst_params, dataset)
+    model_conf = get_model_conf(run.info.artifact_uri)
+    assert "saved_input_example_info" not in model_conf.to_dict()
```
dbczumar, line 445: Nit: I think this can be simplified as `model_conf.saved_input_example_info is None`. This is the approach we take in other test suites.

dbczumar (commented):
@andrewnitu This is awesome! Left a few small comments. Can you also elaborate on the issue you encountered with overwriting patches when `autolog()` is called multiple times & how you ultimately resolved it?

tests/xgboost/test_xgboost_autolog.py
andrewnitu: thats awesome!

mlflow/lightgbm.py
andrewnitu: great point, now they are separate toggles so it makes no sense if the input example disappears when signature inference fails lol

andrewnitu: I will follow this up in another PR

tests/config_helper.py
andrewnitu: hmm fair enough. I was thinking more along the lines of avoiding to break the test cases, but I guess for those that it did break I can pass the values that the test is expecting.

tests/lightgbm/test_lightgbm_autolog.py
andrewnitu: Personally I think this way is slightly cleaner, because it makes it clear that you don't expect that field to be there at all. Versus the one above either `saved_input_example_info` could not exist, or it could have the value actually be none.

What do you think?

tests/xgboost/test_xgboost_autolog.py
andrewnitu: thanks Haru, this makes it very clean! I love it :D

andrewnitu:
actually I don't think we should merge this in the current state... there are big problems with the error messaging, for example we would tell the user "input_example collection failed" even when they disabled the input example collection... I am working on a follow-up to fix this.

andrewnitu:
okay i fixed that issue, I pulled out the logic for the input example and signature as well as the associated logging, so that we can have tests for that behaviour.

unfortunately I had to also pull in the PR where I changed lightgbm autolog implementation to patch the Dataset object to store the input example, since the new util I made expects a _InputExampleInfo object.

mlflow/utils/input_example_and_signature_utils.py
```diff
@@ -0,0 +1,76 @@
...
+class _InputExampleInfo:
+    def __init__(self, input_example=None, error_msg=None):
+        self.input_example = input_example
+        self.error_msg = error_msg
+
+
+def handle_input_example_and_signature(
+    input_example_info, log_input_example, log_model_signature, model_predict_fn
+):
+    if input_example_info is None:
```
andrewnitu, line 19: the autoformatting here :(

mlflow/utils/input_example_and_signature_utils.py
```diff
@@ -0,0 +1,76 @@
+from mlflow.models import infer_signature
```
dbczumar, line 1: I like the idea of trying to DRY out the parsing of input examples / signatures; however, I think this implementation is pretty specific to the XGBoost / LightGBM use cases with the use of `_InputExampleInfo`. I also found the outputs of the functions, which are tuples of input examples, signatures, and warning messages to be somewhat difficult to reason about.

Perhaps we could instead structure this as a single `get_input_example_and_signature()` high-order function with two function arguments - `get_signature()` and `get_input_example()` - and two conf arguments `warn_input_example`, `warn_signature`.

This function could then call `get_signature` / `get_input_example` and log the appropriate warning messages.

Finally, can we place this logic in `mlflow.utils.autologging_utils` instead of defining a new module?

dbczumar (changes requested):
@andrewnitu Awesome! Left a couple comments / suggestions that should hopefully simplify things a bit

mlflow/sklearn/__init__.py
```diff
@@ -806,8 +810,8 @@ def _log_posttraining_metadata(estimator, *args, **kwargs):
                     :INPUT_EXAMPLE_SAMPLE_ROWS
                 ]
 
-                model_output = estimator.predict(input_example)
-                signature = infer_signature(input_example, model_output)
+                if log_model_signature:
+                    signature = infer_signature(input_example, estimator.predict(input_example))
             except Exception as e:  # pylint: disable=broad-except
                 input_example = None
                 msg = "Failed to infer an input example and model signature: " + str(e)
```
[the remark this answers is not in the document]
andrewnitu: We can get this for free by using the new function :)

mlflow/utils/input_example_and_signature_utils.py
andrewnitu: I think we can reuse this in sklearn by just creating an object `_InputExampleInfo(input_example, None)`

I'm not sure what you mean by the function arguments, will these get_signature and get_input_example be defined in xgboost, lightgbm, then simply passed to this util?

We could move the logging inside the method, but I did it this way to avoid having functions with side effects. What do you think?

Yes, I will move it to the `autologging_utils` module

dbczumar: Yep! We'd have each module define a function for fetching the input example and the signature (with the signature taking the input example as an argument). This gives the modules complete flexibility over how to compute the signature. Right now, we assume that signature is always going to be computed using the `model.predict()` followed by `infer_signature()` pattern.

I'd also rather not force all autologging modules to adopt InputExampleInfo when it's only going to be required for a subset of cases (XGBoost, LightGBM).

Finally, if we're concerned about side effects, we can always have the function take a `logger` argument and document that it's used to log warnings when input example / signature inference fails. Effectively, I think the goal of this utility should be to extract out common error handling / warning logic, not to extract out logic for computing the input example / signature.

andrewnitu: I honestly think we need to keep this InputExampleInfo, since when we collect an input example we must record the error message since get_signature needs to print this message as part of the error given to the user (e.g. `unable to infer model signature due to an error encountered when sampling a few rows of data: NoneType has no attribute slice()`)

Thus I propose the library, whether xgboost, sklearn, or lightgbm, must construct an InputExampleInfo because the error is needed at a later stage. This is passed to the helper fn, where the helper fn will check it for existence/errors and if correct then will pass it to get_signature where the signature will be computed from the InputExampleInfo. Then the helper fn will also log the messages for the user.

mlflow/utils/autologging_utils.py
```diff
@@ -113,3 +118,65 @@ def wrap_patch(destination, name, patch, settings=None):
...
+                              succeeds, whether to return the input example to the user. We collect
+                              it even if this parameter is False because it is needed for inferring
+                              the model signature.
+    :param log_model_signature: whether to infer and return the model signature.
+    :param logger: the logger instance used to log warnings to the user during input example
+                   collection and model signature inference.
+
+    :return: A tuple of input_example and signature. Either or both could be None based on the
+             values of log_input_example and log_model_signature.
+    """
```
dbczumar, line 153: I know we discussed this offline, but can we insert some logic to short-circuit & return None, None if both of these conditions are false? Just in case sampling that input data is expensive for some reason.

mlflow/lightgbm.py
```diff
@@ -292,10 +293,40 @@ def autolog():
         - inferred signature of the inputs and outputs of the model.
 
     Note that the `scikit-learn API`_ is not supported.
+
+    :param log_input_example: whether to log a sample of the training data as an example for future
```
dbczumar, line 297: Nit: Instead of "whether X", can we say: "If `True`, logs a sample of the training data. If `False`, a sample is not logged.` Just to be explicit. Can we make this same tweak for the signature param here, as well as `sklearn.py` and `xgboost.py`?

mlflow/utils/autologging_utils.py
```diff
@@ -113,3 +118,65 @@ def wrap_patch(destination, name, patch, settings=None):
...
     gorilla.apply(patch)
+
+
+class _InputExampleInfo:
+    def __init__(self, input_example=None, error_msg=None):
+        self.input_example = input_example
+        self.error_msg = error_msg
+
+
+def handle_input_example_and_signature(
```
dbczumar, line 129: Nit: Is there a better / more descriptive verb we can use than `handle`? `resolve`?

mlflow/utils/autologging_utils.py
```diff
@@ -7,6 +7,11 @@
 
 
 INPUT_EXAMPLE_SAMPLE_ROWS = 5
+FAILED_INPUT_EXAMPLE_PREFIX_TEXT = "Failed to gather input example: "
+FAILED_MODEL_SIGNATURE_PREFIX_TEXT = "Failed to infer model signature: "
+ENSURE_AUTOLOGGING_ENABLED_TEXT = (
```
dbczumar, line 12: Nit: This is unused. Did we intend to swap it into `xgboost.py` and `lightgbm.py`?

mlflow/utils/autologging_utils.py
```diff
@@ -7,6 +7,11 @@
 
 
 INPUT_EXAMPLE_SAMPLE_ROWS = 5
+FAILED_INPUT_EXAMPLE_PREFIX_TEXT = "Failed to gather input example: "
```
dbczumar, line 10: Afaik, this is only defined as a module-level variable for testing purposes later on, where it's used to validate that warning messages appear as expected. Directly checking against this value in the test case leaves us to uncaught regressions: e.g., "FAILED_INPUT_EXAMPLE_PREFIX_TEXT" becomes `Failed to gather signature` due to some unintended line swapping, and we don't catch it because we're just checking that the content of `FAILED_INPUT_EXAMPLE_PREFIX_TEXT` appears in the text.

Accordingly, can we drop these constants and test against the warning messages using inlined strings?

mlflow/xgboost.py
```diff
@@ -294,16 +298,13 @@ def autolog(importance_types=["weight"]):  # pylint: disable=W0102
     Note that the `scikit-learn API`_ is not supported.
 
     :param importance_types: importance types to log.
-
+    :param log_input_example: whether to log a sample of the training data as an example for future
```
dbczumar, line 301: Can we clarify that the example is logged as part of the model? (i.e., as opposed to a tag / parameter)

dbczumar (commented):
@andrewnitu This is really close - great work! Left a few small comments

mlflow/utils/autologging_utils.py
andrewnitu: we already have this: the get_input_example only fires `if log_input_example or log_model_signature:`

mlflow/utils/autologging_utils.py
andrewnitu: i like resolve :)

mlflow/utils/autologging_utils.py
dbczumar: Doh!

mlflow/sklearn/__init__.py
```diff
@@ -794,24 +805,29 @@ def _log_posttraining_metadata(estimator, *args, **kwargs):
...
+            fit_arg_names = _get_arg_names(estimator.fit)
+            X_var_name, y_var_name = fit_arg_names[:2]
+            input_example = _get_Xy(args, kwargs, X_var_name, y_var_name)[0][
+                :INPUT_EXAMPLE_SAMPLE_ROWS
+            ]
+            return input_example
+
+        def infer_model_signature(input_example):
+            if not hasattr(estimator, "predict"):
+                raise Exception("the model used has no predict function")
```
dbczumar, line 820:
```suggestion
                raise Exception("the trained model does not specify a `predict` function, which is required in order to infer the signature")
```

mlflow/sklearn/__init__.py
```diff
@@ -794,24 +805,29 @@ def _log_posttraining_metadata(estimator, *args, **kwargs):
...
+
+            return infer_signature(input_example, estimator.predict(input_example))
+
+        input_example, signature = resolve_input_example_and_signature(
+            get_input_example,
+            infer_model_signature,
+            log_input_example,
+            log_model_signature and hasattr(estimator, "predict"),
```
dbczumar, line 828: We should remove this line, otherwise signature collection fails silently.

dbczumar (approved):
LGTM once https://github.com/mlflow/mlflow/pull/3449/files#r498969404 and https://github.com/mlflow/mlflow/pull/3449/files#r498973038 are addressed. Great work, @andrewnitu !

mlflow/sklearn/__init__.py
andrewnitu: oops sorry i thought i did this :(

dbczumar:
Given that all of our CI tests have passed, with the exception of R tests (which are independent of this PR and have been addressed via https://github.com/mlflow/mlflow/pull/3477), I'm going to proceed to merge this PR. I've also done manual testing with scikit-learn LogisticRegression and PCA models under varying combinations of `log_model_signature = True/False` and `log_input_example = True / False` with sliceable and non-sliceable inputs to ensure that expected outputs and warning messages are produced across input cases.

PR #3451, opened 2020-09-23
Performance improvement in Spark UDFs that return an array

jgc128:

This PR improves performance in Spark UDFs that return an array. Pandas `.iterrows()` is notoriously slow and this fix brings quite a big improvement, especially on big datasets.

## How is this patch tested?

Existing spark unit tests are passing.

Performance comparison on a regular Pandas DataFrame with 100,000 rows and 2 columns:
image

Performance comparison of `mlflow.pyfunc.spark_udf` with a simple model that returns two columns on a local Spark 2.4.0 instance on the same data:

| Number of rows           | Time Original   | Time Modified  |
|----------------|------------|-----------|
| 10,000 | 0.333 sec  | 0.166 sec |
| 100,000 | 1.945 sec  | 0.205 sec |
| 1,000,000 | 18.018 sec | 0.652 sec |

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -690,7 +690,7 @@ def spark_udf(spark, model_uri, result_type="double"):
             result = result.applymap(str)
 
         if type(result_type) == ArrayType:
-            return pandas.Series([row[1].values for row in result.iterrows()])
+            return pandas.Series(result.values.tolist())
         else:
             return result[result.columns[0]]
 
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -690,7 +690,7 @@ def spark_udf(spark, model_uri, result_type="double"):
             result = result.applymap(str)
 
         if type(result_type) == ArrayType:
-            return pandas.Series(result.values.tolist())
+            return pandas.Series(result.to_numpy().tolist())
         else:
             return result[result.columns[0]]
 


mlflow/pyfunc/__init__.py
```diff
@@ -690,7 +690,7 @@ def predict(*args):
             result = result.applymap(str)
 
         if type(result_type) == ArrayType:
-            return pandas.Series([row[1].values for row in result.iterrows()])
+            return pandas.Series(result.to_numpy().tolist())
```
harupy, line 693: @jgc128 Can we remove `.tolist()`?

harupy:
@jgc128 Thanks for the PR! Left a comment.

mlflow/pyfunc/__init__.py
jgc128: We can't - if we remove `.tolist()`, `pd.Series` would throw an exception that the input data must be 1-dimensional. 
The goal here is to create a series where each element is not a single number but an array. It's not idea but that's a limitation of Spark's pandas_udf API, as it should return a series and not a dataframe.

harupy: Makes sense. We create a pandas sereis like below:

```python
pandas.Series([[1,], [2,]])
```

harupy: Confriemd the following code throws:

```
pandas.Series(np.array([[1,], [2,], ...]))
```

jgc128: Yeah, if it's an array, pd.Series would check for dimensions

harupy (approved):
LGTM!

harupy:
@jgc128 Merged!

PR #3456, opened 2020-09-24
Remove pycodestyle & improve 'black' error message

smurching:

Removes `pycodestyle` lint check (superseded by `black`) and improves the error message when `black` lint fails to explain how to rerun `black` to autoformat Python code

## How is this patch tested?
Tested manually by messing up formatting of `mlflow/version.py` and running `./lint.sh`, which now produces:

```
(mlflow-python3) ➜  mlflow git:(fix-black-error-msg) ✗ ./lint.sh   
would reformat /Users/sid.murching/code/mlflow/setup.py
Oh no! 💥 💔 💥
1 file would be reformatted, 416 files would be left unchanged.
Python lint failed. Run 'black --line-length=100 --exclude=mlflow/protos .' from your checkout of MLflow (note the trailing '.' at the end of the command, corresponding to the current directory) to autoformat Python code
```

In the case where there are no linter errors, the behavior is the same as before (this should also be validated by this PR passing lint in Github actions, hopefully :P)

Components 

Interface 

Language 

Integrations

### DIFF
--- a/dev/run-black-linter.sh
+++ b/dev/run-black-linter.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+# Exclude proto files because they are auto-generated
+black --check --line-length=100 --exclude=mlflow/protos .
+exit_code="$?"
+if [[ "$exit_code" != "0" ]]; then
+  echo "Python lint failed. Run 'black --line-length=100 --exclude=mlflow/protos .' from your"\
+    "checkout of MLflow (note the trailing '.' at the end of the command, corresponding to the"\
+    "current directory) to autoformat Python code"
+  exit $exit_code
+fi;
--- a/lint.sh
+++ b/lint.sh
@@ -26,12 +26,8 @@ exclude_dirs=(
   "mlflow/temporary_db_migrations_for_pre_1_users"
 )
 
-# Exclude proto files because they are auto-generated
-black --check --line-length=100 --exclude=mlflow/protos .
-
-exclude=$(join "," "${exclude_dirs[@]}")
-include=$(join " " "${include_dirs[@]}")
-pycodestyle --max-line-length=100 --ignore=E203,W503 --exclude=$exclude -- $include
+# Check Python code for style issues using black
+./dev/run-black-linter.sh
 
 # pylint's `--ignore` option filters files based on their base names, not paths.
 # see: http://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options


dev/run-black-linter.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# Exclude proto files because they are auto-generated
+black --check --line-length=100 --exclude=mlflow/protos .
+exit_code="$?"
+if [[ "$exit_code" != "0" ]]; then
+  echo "Python lint failed. Run 'black --line-length=100 --exclude=mlflow/protos .' from your"\
+    "checkout of MLflow (note the trailing '.' at the end of the command, corresponding to the"\
+    "current directory) to autoformat Python code"
```
apurva-koti, line 9: Should we include this command in another script, like `./autoformat.sh` for ease of use? I assume this would need to be run essentially every iteration of the devloop, and it seems a bit of a pain to be told that your code is badly formatted, to run another command to fix it, but not be told what was actually wrong with it.

apurva-koti: Alternately, we could provide an `--autoformat` option to `lint.sh` itself

tomasatdatabricks: +1 to --autoformat option to lint.sh!

harupy: or we can use `pyproject.toml` and put black options in there so `black .`  becomes equivalent `black <options specified in pyproject.toml> .`

https://black.readthedocs.io/en/stable/pyproject_toml.html

smurching: Both of these ideas sound good! Adding an option to `./lint.sh` seems preferable since users are already running it, as Apurva mentioned. Maybe an option like `--fix-black` since after running `./lint.sh --fix-black` other linters (e.g. pylint) can still fail with errors that need to be fixed manually

harupy: Another advantage of `pyproject.toml` is that an editor like `vscode` can detect `pyproject.toml` and apply auto-formatting with options specified in it.

apurva-koti:
@smurching , any word on this? There seems to be at least one case in which pycodestyle and black give conflicting messages cc @andrewnitu so it'd be good to get these in

andrewnitu:
following up on this, I think it's a pretty annoying workflow to have to run `lint.sh` multiple times to expose all the linting errors. especially if an OSS author is only relying on the github actions to verify the lint results.

previously, an error in pycodestyle would block pylint from running. now, even though we removed pycodestyle, we replaced it with black so an error in `black --check` would block pylint.

I think we should run them all upfront, and expose all the errors at the end. This would speed up the workflow and also prevent false confidence (corey and I merged a PR that only had pycodestyle errors knowing that we would be removing pycodestyle, but unbeknownst to us there were many more pytest errors that were never exposed.)

PR #3470, opened 2020-09-30
[ML-12267] Search using the IN operator for `runId` in search model versions

ankit-db:

This PR adds support for the "IN" operator in a search query when querying the `run_id` parameter in search model versions. Community contribution for support for the other operators if desired would be awesome!

To accomplish this, we have to relax some of the current parsing rules we have, which assume that a `Comparison` is the only kind of valid operator to parse out of a statement. Unfortunately, sqlparse doesn't correctly parse `IN` as a `Comparison` so we have to manually construct the Comparison. Moreover, we currently don't support multiple search expressions at the same time.

## How is this patch tested?

Additional tests for the lists

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/search_utils.py
+++ b/mlflow/utils/search_utils.py
@@ -2,10 +2,11 @@ import base64
 import json
 import operator
 import re
+import ast
 import shlex
 
 import sqlparse
-from sqlparse.sql import Identifier, Token, Comparison, Statement
+from sqlparse.sql import Identifier, Token, Comparison, Statement, Parenthesis, TokenList
 from sqlparse.tokens import Token as TokenType
 
 from mlflow.entities import RunInfo
@@ -29,6 +30,7 @@ class SearchUtils(object):
     VALID_REGISTERED_MODEL_SEARCH_COMPARATORS = CASE_INSENSITIVE_STRING_COMPARISON_OPERATORS.union(
         {"="}
     )
+    VALID_MODEL_VERSIONS_SEARCH_COMPARATORS = set(["=", "IN"])
     VALID_SEARCH_ATTRIBUTE_KEYS = set(RunInfo.get_searchable_attributes())
     VALID_ORDER_BY_ATTRIBUTE_KEYS = set(RunInfo.get_orderable_attributes())
     _METRIC_IDENTIFIER = "metric"
@@ -48,6 +50,7 @@ class SearchUtils(object):
         + list(_ALTERNATE_ATTRIBUTE_IDENTIFIERS)
     )
     STRING_VALUE_TYPES = set([TokenType.Literal.String.Single])
+    DELIMITER_VALUE_TYPES = set([TokenType.Punctuation])
     NUMERIC_VALUE_TYPES = set([TokenType.Literal.Number.Integer, TokenType.Literal.Number.Float])
     # Registered Models Constants
     ORDER_BY_KEY_TIMESTAMP = "timestamp"
@@ -227,7 +230,7 @@ class SearchUtils(object):
         return comp
 
     @classmethod
-    def _invalid_statement_token(cls, token):
+    def _invalid_statement_token_search_runs(cls, token):
         if isinstance(token, Comparison):
             return False
         elif token.is_whitespace:
@@ -238,9 +241,24 @@ class SearchUtils(object):
             return True
 
     @classmethod
+    def _invalid_statement_token_search_model_registry(cls, token):
+        if isinstance(token, Comparison):
+            return False
+        elif isinstance(token, Identifier):
+            return False
+        elif isinstance(token, Parenthesis):
+            return False
+        elif token.is_whitespace:
+            return False
+        elif token.match(ttype=TokenType.Keyword, values=["AND", "IN"]):
+            return False
+        else:
+            return True
+
+    @classmethod
     def _process_statement(cls, statement):
         # check validity
-        invalids = list(filter(cls._invalid_statement_token, statement.tokens))
+        invalids = list(filter(cls._invalid_statement_token_search_runs, statement.tokens))
         if len(invalids) > 0:
             invalid_clauses = ", ".join("'%s'" % token for token in invalids)
             raise MlflowException(
@@ -570,16 +588,25 @@ class SearchUtils(object):
                 error_code=INVALID_PARAMETER_VALUE,
             )
         value_token = stripped_comparison[2]
-        if value_token.ttype not in cls.STRING_VALUE_TYPES:
+        if not isinstance(value_token, Parenthesis) and value_token.ttype not in cls.STRING_VALUE_TYPES:
             raise MlflowException(
                 "Expected a quoted string value for attributes. "
-                "Got value {value}".format(value=value_token.value),
+                "Got value {value} with type {type}".format(value=value_token.value, type=type(value_token)),
                 error_code=INVALID_PARAMETER_VALUE,
             )
+        elif isinstance(value_token, Parenthesis):
+            if len(value_token._groupable_tokens) == 0:
+                raise MlflowException("While parsing a list in the query, expected a non-empty list of string values, but got empty list", error_code=INVALID_PARAMETER_VALUE)
+            elif not all(map(lambda token: token.ttype in cls.STRING_VALUE_TYPES.union(cls.DELIMITER_VALUE_TYPES), value_token._groupable_tokens[0].tokens)):
+                raise MlflowException("While parsing a list in the query, expected string value or punctuation, but got different type in list: {value_token}".format(value_token=value_token), error_code=INVALID_PARAMETER_VALUE)
+            value = ast.literal_eval(value_token.value)
+        else:
+            value = cls._strip_quotes(value_token.value, expect_quoted_value=True)
+
         comp = {
             "key": key,
             "comparator": stripped_comparison[1].value,
-            "value": cls._strip_quotes(value_token.value, expect_quoted_value=True),
+            "value": value,
         }
         return comp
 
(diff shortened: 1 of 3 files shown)


ankit-db:
R tests seem broken, but otherwise, I think we're fine in terms of tests here

mlflow/utils/search_utils.py
```diff
@@ -607,16 +655,47 @@ def _parse_filter_for_model_registry(cls, filter_string, valid_search_keys):
...
+                    )
+                if (
+                    isinstance(statement.tokens[idx], Identifier)
+                    or statement.tokens[idx].match(ttype=TokenType.Keyword, values=["IN"])
+                    or isinstance(statement.tokens[idx], Parenthesis)
+                ):
+                    comparison_subtokens.append(statement.tokens[idx])
+                else:
+                    break
+                idx += 1
```
sueann, line 691: would the following be equivalent?
```
            for token in statement.tokens:
                if isinstance(token, Comparison):
                    raise MlflowException(
                        "Search filter '%s' contains multiple expressions. "
                        "%s " % (filter_string, expected),
                    )

                if (
                    isinstance(token, Identifier)
                    or token.match(ttype=TokenType.Keyword, values=["IN"])
                    or isinstance(token, Parenthesis)
                ):
                    comparison_subtokens.append(token)
                elif not token.is_whitespace:
                    break
```

mlflow/utils/search_utils.py
```diff
@@ -607,16 +655,47 @@ def _parse_filter_for_model_registry(cls, filter_string, valid_search_keys):
...
         if len(invalids) > 0:
             invalid_clauses = ", ".join("'%s'" % token for token in invalids)
             raise MlflowException(
                 "Invalid clause(s) in filter string: %s. " "%s" % (invalid_clauses, expected),
                 error_code=INVALID_PARAMETER_VALUE,
             )
+        # process non-Comparison tokens
+        if len(statement.tokens) == 1 and isinstance(statement.tokens[0], Comparison):
+            token_list = statement.tokens
+        elif len(statement.tokens) > 1:
```
sueann, line 670: what happens if `len(statement.tokens) == 1 and not instance(statement.tokens[0], Comparison)`?

sueann: also is it possible for `len(statement.tokens) == 0` to occur?

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -719,6 +719,40 @@ def search_versions(filter_string):
...
+        assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+        assert "expected string value or punctuation" in exception_context.exception.message
+
+        # search using the IN operator with empty lists should return exceptions
+        with self.assertRaises(MlflowException) as exception_context:
+            search_versions("run_id IN ()")
+        assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+        assert "expected a non-empty list of string values" in exception_context.exception.message
+
+        # search using the IN operator is not allowed with other additional filters
```
sueann, line 746: this is just for the IN query, right? we do allow multiple conditions if the conditions are not IN queries? if so, why did we make this choice here?

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -719,6 +719,40 @@ def search_versions(filter_string):
...
+
+        # search using the IN operator with bad lists should return exceptions
+        with self.assertRaises(MlflowException) as exception_context:
+            search_versions("run_id IN (1,2,3)")
+        assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+        assert "expected string value or punctuation" in exception_context.exception.message
+
+        # search using the IN operator with empty lists should return exceptions
+        with self.assertRaises(MlflowException) as exception_context:
+            search_versions("run_id IN ()")
```
sueann, line 742: do we need to worry about malformed queries here like `run_id IN (` ?

sueann (commented):
I did a pass but could benefit from going over some of the code together.

mlflow/utils/search_utils.py
```diff
@@ -237,10 +240,25 @@ def _invalid_statement_token(cls, token):
         else:
             return True
 
+    @classmethod
+    def _invalid_statement_token_search_model_registry(cls, token):
+        if isinstance(token, Comparison):
+            return False
+        elif isinstance(token, Identifier):
```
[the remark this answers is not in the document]
ankit-db: Sqlparse docs don't help anyone much - will drop time on your calendar

tests/store/model_registry/test_sqlalchemy_store.py
ankit-db: Added a test for it, but it does seem to perform as expected (throws MlflowException with "Invalid clause")

mlflow/utils/search_utils.py
ankit-db: @the first comment - great point - I missed that, and I think that may have been an existing bug too, so will add a case for it (though tbh, I'm not sure if it's very likely that kind of query can be constructed.

@the second comment - I don't think it's possible but I've added an error message to signal that

tests/store/model_registry/test_sqlalchemy_store.py
ankit-db: Per https://github.com/mlflow/mlflow/blob/master/mlflow/store/model_registry/sqlalchemy_store.py#L765, I believe we don't allow multiple conditions in search model versions filters both in OSS and within Databricks, so this test was to make sure that we didn't accidentally mess up the error messaging while adding support for the IN operator

mlflow/utils/search_utils.py
ankit-db: It would - thanks so much for the much cleaner refactor

mlflow/utils/search_utils.py
```diff
@@ -558,6 +584,48 @@ def paginate(cls, runs, page_token, max_results):
...
 
+    @classmethod
+    def _is_valid_identifier_list(cls, value_token):
+        if len(value_token._groupable_tokens) == 0:
+            raise MlflowException(
+                "While parsing a list in the query,"
+                " expected a non-empty list of string values, but got empty list",
+                error_code=INVALID_PARAMETER_VALUE,
+            )
+        elif not isinstance(value_token._groupable_tokens[0], IdentifierList):
```
ankit-db, line 595: New from after our conversation, since we need to ensure that the first element is actually be a list (might not in the case of a list that looks like `(,)`)

mlflow/utils/search_utils.py
```diff
@@ -570,19 +638,89 @@ def _get_comparison_for_model_registry(cls, comparison, valid_search_keys):
...
+                    )
+                elif cls._is_list_component_token(token):
+                    comparison_subtokens.append(token)
+                elif not token.is_whitespace:
+                    break
+            # if we have fewer than 3, that means we have an incomplete statement.
+            if len(comparison_subtokens) == 3:
+                token_list = [Comparison(TokenList(comparison_subtokens))]
+            else:
+                raise MlflowException(
```
ankit-db, line 718: Also new from after our conversation: for ease of reasoning, moved this failure case up into this logic (previously, we would let this case pass until the next stage where the comparison would get rejected. That made it hard to understand what happened in this case)

mlflow/utils/search_utils.py
```diff
@@ -576,6 +584,48 @@ def paginate(cls, runs, page_token, max_results):
     VALID_SEARCH_KEYS_FOR_MODEL_VERSIONS = set(["name", "run_id", "source_path"])
     VALID_SEARCH_KEYS_FOR_REGISTERED_MODELS = set(["name"])
 
+    @classmethod
+    def _is_valid_identifier_list(cls, value_token):
```
sueann, line 588: maybe `_check_valid_identifier_list` since it doesn't need to return anything (it just throws)

mlflow/utils/search_utils.py
```diff
@@ -576,6 +584,48 @@ def paginate(cls, runs, page_token, max_results):
...
+            )
+        ):
+            raise MlflowException(
+                "While parsing a list in the query, expected string value "
+                "or punctuation, but got different type in list: {value_token}".format(
+                    value_token=value_token
+                ),
+                error_code=INVALID_PARAMETER_VALUE,
+            )
+        return value_token
```
sueann, line 615: don't need to return this?

mlflow/utils/search_utils.py
```diff
@@ -576,6 +584,48 @@ def paginate(cls, runs, page_token, max_results):
...
+
+    @classmethod
+    def _parse_list_from_sql_token(cls, token):
+        try:
+            value = ast.literal_eval(token.value)
```
sueann, line 620: nit: just return here?

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -743,6 +743,27 @@ def search_versions(filter_string):
         assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
         assert "expected a non-empty list of string values" in exception_context.exception.message
 
+        # search using an ill-formed IN operator correctly throws exception
+        with self.assertRaises(MlflowException) as exception_context:
```
sueann, line 747: could have used @ parametrize but no need to change now.

sueann (approved):
just a few nits - o/w lgtm!

tests/store/model_registry/test_sqlalchemy_store.py
ankit-db: Ah right - yeah will keep that in mind for next time as well

ankit-db:
Going ahead with the merge, because the R test failure appears to be one that's fixed by https://github.com/mlflow/mlflow/pull/3477

PR #3477, opened 2020-09-30
R test fix

juntai-zheng:

This PR fixes the R test failures. This was caused due to a mismatch between the h2o cluster persisted by the test (3.30.1.2) and the version being run client-side (3.30.1.3). This PR also sets the conda environment versioning of h2o to be equal instead of greater than equal to the version used to save the model in R.

In order to fix the cluster versioning mismatch, I followed the instructions at https://docs.h2o.ai/h2o/latest-stable/h2o-docs/downloading.html#install-in-r to reinstall h2o every time in the tests before calling `h2o.init()`. This creates a cluster with the latest versioning so old clusters don't get used for testing.

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/R/mlflow/R/model-h2o.R
+++ b/mlflow/R/mlflow/R/model-h2o.R
@@ -31,7 +31,7 @@ mlflow_save_model.H2OModel <- function(model,
   )
   yaml::write_yaml(settings, file.path(model_data_path, "h2o.yaml"))
 
-  needed because R and python packages may not have the same patch number
+  # needed because R and python packages may not have the same patch number
   version <- remove_patch_version(
     as.character(utils::packageVersion("h2o"))
   )
--- a/mlflow/R/mlflow/R/model-h2o.R
+++ b/mlflow/R/mlflow/R/model-h2o.R
@@ -37,7 +37,7 @@ mlflow_save_model.H2OModel <- function(model,
   )
 
   conda_env <- create_default_conda_env_if_absent(
-    path, conda_env, default_pip_deps = list("mlflow", paste0("h2o>=", as.character(utils::packageVersion("h2o"))))
+    path, conda_env, default_pip_deps = list("mlflow", paste0("h2o==", as.character(utils::packageVersion("h2o"))))
   )
 
   h2o_conf <- list(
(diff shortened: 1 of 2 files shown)


mlflow/R/mlflow/tests/testthat/test-model-h2o.R
```diff
@@ -7,6 +7,15 @@ predictors <- setdiff(colnames(iris), prediction)
...
 test <- iris[idx[101:nrow(iris)], ]
 
+# Installing most recent h2o package, see https://docs.h2o.ai/h2o/latest-stable/h2o-docs/downloading.html#install-in-r
+if ("package:h2o" %in% search()) { detach("package:h2o", unload=TRUE) }
+if ("h2o" %in% rownames(installed.packages())) { remove.packages("h2o") }
+pkgs <- c("RCurl","jsonlite")
+for (pkg in pkgs) {
+  if (! (pkg %in% rownames(installed.packages()))) { install.packages(pkg) }
+}
+install.packages("h2o", type="source", repos=(c("http://h2o-release.s3.amazonaws.com/h2o/latest_stable_R")))
```
apurva-koti, line 17: Do we know this link to be stable/reliable?

mlflow/R/mlflow/R/model-h2o.R
```diff
@@ -31,13 +31,8 @@ mlflow_save_model.H2OModel <- function(model,
   )
   yaml::write_yaml(settings, file.path(model_data_path, "h2o.yaml"))
 
-  # needed because R and python packages may not have the same patch number
```
juntai-zheng, line 34: Removed this because setting the versions equal requires the whole version number, but `remove_patch_version` only returned `3.30`.

mlflow/R/mlflow/tests/testthat/test-model-h2o.R
juntai-zheng: I believe it's more reliable than CRAN, since CRAN might have update delays compared to the source or pypi (as seen when their CRAN download page was taken down for a few days when they updated their package).

apurva-koti: Cool, we could probably just revisit this if this link causes any trouble then.

apurva-koti (approved):
LGTM!

Note: we should keep an eye on this to see if anything breaks, or if the link rots.

Issue #3483, opened 2020-10-01
[FR] Enable using azure-identity for azure blob for AAD


thomasfrederikhoeck:
## Willingness to contribute

- [ ] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

Allow users to authenticate to Azure Blob storage with AAD instead of storage key

## Motivation
- Moving all access management to deployer of MLflow and not user
- Limiting what access is given to storage account
- Won't have to make sure users roll their storage key as it is renewed.

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interfaces
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Languages 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

## Details

[...]

PR #3484, opened 2020-10-01
Added the option to use AAD for auth for store

thomasfrederikhoeck:

This change allows for using a Azure blob storage where AAD is used as authentication as described in #3483 

The main idea is that it allows for using a service principal that only has access to the blob container. It futhermore
allows for controlling access to the Artifact Store through AAD as user/developers/model-builders can login with Azure CLI locally through the Azure Activate Directory login flow. This means that the normal Azure login flow can be used for model-builders to get access to the Artifact store.

## How is this patch tested?

The current unit-test has been extended, and I have tried connecting to a AAD enabled blob container.

Azure Activate Directory authentication when using Azure blob as storage for artifact store.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/docs/source/tracking.rst
+++ b/docs/source/tracking.rst
@@ -699,12 +699,25 @@ Azure Blob Storage
 
 To store artifacts in Azure Blob Storage, specify a URI of the form
 ``wasbs://<container>@<storage-account>.blob.core.windows.net/<path>``.
-MLflow expects Azure Storage access credentials in the
-``AZURE_STORAGE_CONNECTION_STRING`` or ``AZURE_STORAGE_ACCESS_KEY`` environment variables (preferring
-a connection string if one is set), so you must set one of these variables on both your client
-application and your MLflow tracking server. Finally, you must run ``pip install azure-storage-blob``
-separately (on both your client and the server) to access Azure Blob Storage; MLflow does not declare
-a dependency on this package by default.
+MLflow can access Azure Storage by using an access key or by using Azure Active Directory (AAD).
+
+When using an access key MLflow expects ``AZURE_STORAGE_CONNECTION_STRING`` or ``AZURE_STORAGE_ACCESS_KEY``
+environment variables, so you must set one of these variables on both your client application and your MLflow
+tracking server.
+
+If you use a blob container with AAD enabled you have two options when authenticating. The first is using a service
+principal where MLflow expects the environment variables ``AZURE_TENANT_ID``, ``AZURE_CLIENT_ID`` and
+``AZURE_CLIENT_SECRET``. The second option is to use the Azure CLI with the command ``az login``. To enable this option
+you must set the environment variable ``AZURE_STORAGE_CLI_LOGIN`` equal to ``true``. Finally, you must run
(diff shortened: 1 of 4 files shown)


thomasfrederikhoeck:
Looks like the R fail is fixed in #3477

thomasfrederikhoeck:
> Looks like the R fail is fixed in #3477 or #3472

Solved now.

thomasfrederikhoeck:
@dbczumar will you review this? Sorry for tagging you if that is not how you go about it on the MLflow repo, but I couldn't find anything in the contribution guide on how to show that your PR as ready for review 😄

thomasfrederikhoeck:
Hi @javierluraschi - do you want to review this? 😄

thomasfrederikhoeck:
@dmatrix I have been struggling with finding someone to review this? Do you want to do this?

thomasfrederikhoeck:
@sueann I have been struggling with finding someone to review this? Do you want to do this? :-)

bastidas:
I tried out `python -m pip install git+https://github.com/thomasfrederikhoeck/mlflow.git@feat_store_add --upgrade` and it works. I want to avoid the need to have passwords in a connection string in our git repo; this PR does that.

**I would like this to be merged with master.**

alanjcastonguay:
I do wonder whether `azure.identity.DefaultAzureCredential` should be considered instead of `azure.identity.AzureCliCredential` directly, to support other methods automagically per https://docs.microsoft.com/en-us/azure/developer/python/azure-sdk-authenticate?tabs=cmd. But the current implementation in commit 4feaf5bc16f3c0015fb7aa4b150c8f39bc15b60b does work - I helped @bastidas validate above. It looks good to me. @dbczumar could you take a look?

thomasfrederikhoeck:
@alanjcastonguay I think DefaultAzureCredential can work as long as long as Interactive is disabled (this is the default) which can give silent blocks. It might be a bit harder writing specific error in the style of the ones used for the Env Var login approach. This is why I went for the CLI add.

alanjcastonguay:
@dbczumar @harupy What else is needed here?

gdippolito:
@thomasfrederikhoeck I think interactive is disabled by default as per constrcutor -> https://docs.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python#constructor 

I think it would be valuable to use `DefaultAzureCredential` so that user can login via managed identities if they have mlflow deployed in Azure. This would be a much more elegant and secure option instead of using an environment variable containing sensible data.  

I have created a similar PR (https://github.com/mlflow/mlflow/pull/4510) where I have given the ability to use azure-identity if necessary (avoid breaking changes) when an environment variable is set. I also believe environment variable like `AZURE_STORAGE_ACCESS_KEY` should be namespaced to be clear what the intention of that environment variable is and to avoid any conflict with other clients that will use the same environment variable.

thomasfrederikhoeck:
@gdippolito I agree that using `DefaultAzureCredential ` is better. Fine with another PR - to be honest I had written this PR off as I was struggling to get someone from the mlflow team to review it. :-D

PR #3486, opened 2020-10-02
Unpin and deduplicate test requirements where possible

apurva-koti:

We are hoping to test the latest version of MLflow's dependencies to identify regressions. This involves:
- Removing the version pins from `dev/small-requirements.txt` and `dev/large-requirements.txt`
- Removing dependencies in `dev/small-requirements.txt` and `dev/large-requirements.txt`, if those dependencies are already specified in `setup.py`. 

## How is this patch tested?

Re-run tests and ensure there is no behavior change.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/dev/extra-ml-requirements.txt
+++ b/dev/extra-ml-requirements.txt
@@ -3,8 +3,6 @@
 ##
 # Required by mlflow.azureml
 azureml-sdk==1.2.0; python_version >= "3.0"
-# Required by mlflow.pyfunc
-cloudpickle
 # Required by mlflow.keras
 keras==2.3.1
 # Required by mlflow.sklearn
--- a/dev/extra-ml-requirements.txt
+++ b/dev/extra-ml-requirements.txt
@@ -6,31 +6,31 @@ azureml-sdk==1.2.0; python_version >= "3.0"
 # Required by mlflow.keras
 keras==2.3.1
 # Required by mlflow.sklearn
-scikit-learn
+scikit-learn==0.23.2
 # Required by mlflow.gluon
-mxnet
+mxnet==1.5.0
 # Required by mlflow.fastai
-fastai
+fastai==1.0.60
 # Required by mlflow.spacy
-spacy
+spacy==2.2.3
 # Required by mlflow.tensorflow
-tensorflow
+tensorflow==1.15.2
 # Required by mlflow.pytorch
-torch
-torchvision
+torch==1.4.0
+torchvision==0.5.0
 # Required by mlflow.xgboost
-xgboost
+xgboost>=0.82
 # Required by mlflow.lightgbm
-lightgbm
+lightgbm==2.3.0
 # Required by mlflow.h2o
-h2o
+h2o==3.22.1.4
 # Required by mlflow.onnx
-onnx
-onnxmltools
-onnxruntime
+onnx==1.4.1;
+onnxmltools==1.4.0;
+onnxruntime==0.3.0;
 # Required by mlflow.mleap, and in order to save SparkML models in
 # mleap format via ``mlflow.spark.log_model``, ``mlflow.spark.save_model``
-mleap
+mleap==0.16.0
 # Required by mlflow.spark
-pyspark
+pyspark==2.4.0
 
--- a/dev/extra-ml-requirements.txt
+++ b/dev/extra-ml-requirements.txt
@@ -16,7 +16,7 @@ spacy==2.2.3
 # Required by mlflow.tensorflow
 tensorflow==1.15.2
 # Required by mlflow.pytorch
-torch==1.4.0
+torch==1.6.0
 torchvision==0.5.0
 # Required by mlflow.xgboost
 xgboost>=0.82
--- a/dev/extra-ml-requirements.txt
+++ b/dev/extra-ml-requirements.txt
@@ -4,35 +4,35 @@
 # Required by mlflow.azureml
 azureml-sdk==1.2.0; python_version >= "3.0"
 # Required by mlflow.pyfunc
-cloudpickle==0.8.0
+cloudpickle
 # Required by mlflow.keras
 keras==2.3.1
 # Required by mlflow.sklearn
-scikit-learn==0.23.2
+scikit-learn
 # Required by mlflow.gluon
-mxnet==1.5.0
+mxnet
 # Required by mlflow.fastai
-fastai==1.0.60
+fastai
 # Required by mlflow.spacy
-spacy==2.2.3
+spacy
 # Required by mlflow.tensorflow
-tensorflow==1.15.2
+tensorflow
 # Required by mlflow.pytorch
-torch==1.4.0
-torchvision==0.5.0
+torch
+torchvision
 # Required by mlflow.xgboost
-xgboost>=0.82
+xgboost
 # Required by mlflow.lightgbm
-lightgbm==2.3.0
+lightgbm
 # Required by mlflow.h2o
-h2o==3.22.1.4
+h2o
 # Required by mlflow.onnx
-onnx==1.4.1;
-onnxmltools==1.4.0;
-onnxruntime==0.3.0;
+onnx
+onnxmltools
+onnxruntime
 # Required by mlflow.mleap, and in order to save SparkML models in
 # mleap format via ``mlflow.spark.log_model``, ``mlflow.spark.save_model``
-mleap==0.16.0
+mleap
 # Required by mlflow.spark
-pyspark==2.4.0
+pyspark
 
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -63,7 +63,7 @@ def package_files(directory):
             "scikit-learn==0.20; python_version < '3.5'",
             "boto3>=1.7.12",
             "mleap>=0.16.0",
-            "azure-storage-blob>=12.0",
+            "azure-storage-blob==12.3.0",
```
smurching, line 66:
```suggestion
            "azure-storage-blob>=12.0",
```

apurva-koti:
Caveats:

- `azure-storage-blob==12.3.0` otherwise `cannot import BlobPrefix`.
- `boto3==1.9.84`, otherwise sagemaker token issue
- Unpinning pytest fails a test that uses the `matches` keyword. This has been changed to `match`, test updated.

harupy:
https://github.com/mlflow/mlflow/pull/3486/checks?check_run_id=1244264323#step:4:1290

It looks like `test_sklearn_model_export.py` uses autolog-enabled sklearn model classes. I think this is because we run `test_sklearn_autolog.py` and `test_sklearn_model_export.py` in the same pytest session.

dev/small-requirements.txt
```diff
@@ -2,25 +2,25 @@
...
 ## Test-only dependencies
-attrdict==2.0.0
+attrdict
 pytest==3.2.1
 pytest-cov==2.6.0
 pytest-localserver==0.5.0
-scipy==1.2.1
-moto==1.3.7
+scipy
+moto
```
dbczumar, line 24: `moto` and other test-only dependencies don't need to be unpinned.

dev/small-requirements.txt
```diff
@@ -2,25 +2,25 @@
...
 # Required to run the MLflow server against SQL-backed storage
-sqlalchemy==1.3.0
+sqlalchemy
 # Required by the mlflow.projects module, when running projects against
 # a remote Kubernetes cluster
-kubernetes==9.0.0
+kubernetes
 ## Test-only dependencies
-attrdict==2.0.0
+attrdict
```
dbczumar, line 19: I would rather we remove all of these from `small-requirements` and instead add the comments about specific packages to `setup.py`.

dev/large-requirements.txt
```diff
@@ -1,5 +1,5 @@
 ## Large test reqs
 ## Test-only dependencies
-matplotlib==3.2.1
-tf2onnx==1.5.4
-pyarrow==0.17.0
+matplotlib
+tf2onnx
+pyarrow
```
dbczumar, line 5: `pyarrow` isn't actually a test-only dependency. it's used here: https://github.com/mlflow/mlflow/blob/c9856e7855c9d5340cc0c7b0ea67aba5d29ba075/mlflow/store/artifact/hdfs_artifact_repo.py#L167.

dev/extra-ml-requirements.txt
```diff
@@ -4,35 +4,35 @@
 # Required by mlflow.azureml
 azureml-sdk==1.2.0; python_version >= "3.0"
 # Required by mlflow.pyfunc
-cloudpickle==0.8.0
+cloudpickle
```
dbczumar, line 7: This isn't actually an extra ML requirement (I should've caught this while reviewing Sid's PR). It's included in MLflow by default via `setup.py`. We should remove it from this file.

dev/extra-ml-requirements.txt
```diff
@@ -4,35 +4,35 @@
 # Required by mlflow.azureml
```
dbczumar, line 4: If it's too hard to get tests to pass after unpinning libraries in this file, perhaps we can merge the unpinning of MLflow's core / non-ML-extra requirements and then follow up with a PR to unpin extra ML requirements.

dev/large-requirements.txt
apurva-koti: Strange, it's not listed in `setup.py`. I've added it there.

dbczumar: It should be listed in the `extras` section, rather than `install_requires`, since HDFS artifact repository is not critical to the operation of MLflow in most cases.

setup.py
```diff
@@ -47,6 +47,7 @@ def package_files(directory):
         "protobuf>=3.6.0",
         "gitpython>=2.1.0",
         "pyyaml",
+        "pyarrow",
```
dbczumar, line 50: This should be moved to `extras` - see https://github.com/mlflow/mlflow/pull/3486#discussion_r504420027

dev/small-requirements.txt
```diff
@@ -1,26 +1,17 @@
...
-botocore==1.12.84  # pinned for moto
-boto3==1.9.84      # pinned for moto
-# Required to log artifacts and models to HDFS artifact locations
-pyarrow==0.17.0
 # Required to log artifacts to SFTP artifact locations
-pysftp==0.2.9
+pysftp
 # Required to run the MLflow server against SQL-backed storage
-sqlalchemy==1.3.0
+sqlalchemy
```
dbczumar, line 5: This is a core dependency that's already specified in `setup.py` and should be removed from this file

dev/small-requirements.txt
```diff
@@ -1,26 +1,17 @@
 ## Small test reqs
-# Required
-azure-storage-blob==12.3.0
-# Required to log artifacts and models to GCS artifact locations
-google-cloud-storage==1.14.0
-# Required to log artifacts and models to AWS S3 artifact locations
-botocore==1.12.84  # pinned for moto
-boto3==1.9.84      # pinned for moto
-# Required to log artifacts and models to HDFS artifact locations
-pyarrow==0.17.0
```
dbczumar, line 10: @apurva-koti looks like we removed these comments without re-introducing them in `setup.py`. Can we make sure that these comments are still captured?

tests/store/artifact/test_azure_blob_artifact_repo.py
```diff
@@ -3,8 +3,11 @@
 import pytest
 from unittest import mock
 
+try:
+    from azure.storage.blob import BlobPrefix, BlobProperties
```
dbczumar, line 7: Assuming we're leveraging the latest version of `azure-storage-blob`, I don't think it's 100% necessary to try/catch here.

dev/small-requirements.txt
```diff
@@ -1,26 +1,17 @@
...
 # Required to log artifacts to SFTP artifact locations
-pysftp==0.2.9
+pysftp
 # Required to run the MLflow server against SQL-backed storage
-sqlalchemy==1.3.0
+sqlalchemy
 # Required by the mlflow.projects module, when running projects against
 # a remote Kubernetes cluster
-kubernetes==9.0.0
+kubernetes
```
dbczumar, line 8: Can we add `kubernetes` and `pysftp` to the `extras` section of `setup.py` and remove them from this file?

setup.py
apurva-koti: Ah, missed that - thanks

tests/store/artifact/test_azure_blob_artifact_repo.py
apurva-koti: Good point, will take out

dbczumar (approved):
LGTM once https://github.com/mlflow/mlflow/pull/3486/files#r504392236 has been addressed!

setup.py
```diff
@@ -52,18 +53,28 @@ def package_files(directory):
...
         "gorilla",
         "prometheus-flask-exporter",
     ],
     extras_require={
         "extras": [
             "scikit-learn; python_version >= '3.5'",
-            "boto3>=1.7.12",
-            "mleap>=0.16.0",
-            "azure-storage-blob>=12.0",
+            "scikit-learn==0.20; python_version < '3.5'",
```
harupy, line 64: Can we remove this line because we don't support python < 3.5 anymore?

apurva-koti: Yep! Thanks.

setup.py
```diff
@@ -52,18 +53,27 @@ def package_files(directory):
...
         "gorilla",
         "prometheus-flask-exporter",
     ],
     extras_require={
         "extras": [
-            "scikit-learn; python_version >= '3.5'",
-            "boto3>=1.7.12",
-            "mleap>=0.16.0",
-            "azure-storage-blob>=12.0",
+            "scikit-learn",
```
harupy, line 63: Can we remove `scikit-learn` here because `dev/extra-ml-requirements.txt` already contains it?

dbczumar: I don't think we should remove `scikit-learn`. It's a very common dependency for many of our examples (e.g., on mlflow.org and in the MLflow repo). We also assume that `[extras]` includes scikit-learn in our installation instructions for the MLflow Tutorial: https://mlflow.org/docs/latest/tutorials-and-examples/tutorial.html#what-you-ll-need

harupy: That makes sense 👍

PR #3495, opened 2020-10-05
Document extra ML requirements for MLflow

smurching:

* Separates out ML flavor dependencies (and documents which modules require which dependencies) into their own dev/extra-ml-requirements.txt file, and uses this file to run tests in the OSS MLflow build (so that we test that MLflow works with a documented set of extra ML dependencies)
* Adds an `EXTRA_DEPENDENCIES.rst` file to the root of the MLflow repository that points to extra-ml-requirements.txt, discusses the distinction between core / extra MLflow dependencies, and indicates that ML dependencies must be installed separately.
* Updates the MLflow quickstart installation instructions to guide users towards EXTRA_DEPENDENCIES.rst for more information on extra MLflow dependencies
* Fixed issue where we accidentally installed large (instead of small) dependencies for small Python tests, and moved some tests from the small Python tests -> flavor tests as a result (tests/models, test/utils/test_model_utils.py)

## How is this patch tested?

Existing tests

Components 

Interface 

Language 

Integrations


EXTRA_DEPENDENCIES.rst
```diff
@@ -0,0 +1,19 @@
+=========================
```
smurching, line 1: Rather than combining a description of extra dependencies + installation instructions in a single INSTALLATION.rst file linked-to by the quickstart, I kept the existing installation instructions inline in the quickstart for visibility (and because they have some nice properties like code tabs for installing Python vs R)

Therefore I ended up calling this file EXTRA_DEPENDENCIES.rst & focusing only on describing steps for installing extra dependencies

.github/workflows/master.yml
```diff
@@ -106,6 +106,7 @@ jobs:
     - name: Install dependencies
       env:
         INSTALL_LARGE_PYTHON_DEPS: true
+        INSTALL_SMALL_PYTHON_DEPS: true
```
apurva-koti, line 109: Why do we need to install smalls here as well?

dbczumar (approved):
LGTM with one suggestion: Can we update https://github.com/mlflow/mlflow/blob/master/docs/source/quickstart.rst to refer to this `extra_dependencies` file, rather than `large-requirements.txt` ?

Thanks @smurching !

.github/workflows/master.yml
smurching: It's just the way the install-common-deps.sh helper script is written, we independently flag-on installation of small/large dependencies. In the future, maybe we should refactor the helper script so that you declaratively specify what job type you're running (small or large python tests, java tests, etc) and it installs the appropriate dependencies - this would make it easier to e.g. repro the steps of a github actions run on your local machine.

PR #3498, opened 2020-10-06
Add universal mlflow.autolog() function!

andrewnitu:

- Add universal autolog function that enables autologging for the libraries that have been or will be imported.
- Modify spark autologging to not require an active SparkContext when it is initialized -- instead, patch the function to wait for one to be created. This way, an exception will not be thrown when invoking mlflow.autolog() (since pyspark is often imported in our own mlflow code, hence activating autolog for it)

## How is this patch tested?

- Manual testing
- Unit tests
    - verify the autolog function of each library is invoked after calling universal autolog and importing the library
    - verify spark autologging does not throw anymore when a SparkContext is missing
    - verify spark autologging attaches correctly to a SparkContext created AFTER spark autologging was initialized

Add a universal autolog function mlflow.autolog(), allowing you to make one autolog call and automatically activate all MLflow's supported autologging integrations.

Components 

Interface 

Language 

Integrations


mlflow/tracking/fluent.py
```diff
@@ -1020,3 +1036,26 @@ def _get_experiment_id():
...
         or (is_in_databricks_notebook() and get_notebook_id())
     ) or deprecated_default_exp_id
+
+
+def universal_autolog(
+    log_input_example=False, log_model_signature=True
+):  # pylint: disable=unused-argument
+    # getargvalues isnt actually deprecated
+    # https://docs.python.org/3/library/inspect.html#inspect.getargvalues
+    arg_info = inspect.getargvalues(inspect.currentframe())  # pylint: disable=deprecated-method
```
dbczumar, line 1046: According to the docs for `inspect.currentframe()`, this may return `None` in certain environments (see https://docs.python.org/3/library/inspect.html#inspect.currentframe). Could we instead check explicitly for the presence of `log_input_example` and `log_model_signature` in the function signature inspected during `setup_autologging`?

mlflow/tracking/fluent.py
```diff
@@ -1020,3 +1036,26 @@ def _get_experiment_id():
...
+    arg_values = {k: v for k, v in arg_info.locals.items() if k in arg_info.args}
+
+    def setup_autologging(module):
+        integration_module_obj = getattr(mlflow, AUTOLOG_INTEGRATIONS[module.__name__])
+        autolog_fn = getattr(integration_module_obj, "autolog")
+        needed_params = list(inspect.signature(autolog_fn).parameters.keys())
```
dbczumar, line 1052: According to the docs (https://docs.python.org/3/library/inspect.html#inspect.signature), this function may fail with a `ValueError` for certain objects / environments. Can we handle the ValueError case?

mlflow/tracking/fluent.py
```diff
@@ -1020,3 +1036,26 @@ def _get_experiment_id():
         or _get_experiment_id_from_env()
         or (is_in_databricks_notebook() and get_notebook_id())
     ) or deprecated_default_exp_id
+
+
+def universal_autolog(
```
dbczumar, line 1041: In accordance with the comment in `mlflow/__init__.py`, can we rename this to `autolog()` for consistency with other modules & for code discoverability?

mlflow/tracking/fluent.py
```diff
@@ -1020,3 +1036,26 @@ def _get_experiment_id():
...
+    log_input_example=False, log_model_signature=True
+):  # pylint: disable=unused-argument
```
dbczumar, line 1043: Do we plan to add docs to this method as part of a follow-up PR? (I know there's a JIRA ticket for this, but wanted to double check)

tests/tracking/test_fluent.py
```diff
@@ -681,3 +682,42 @@ def test_delete_tag():
...
+        autolog_fn_spy = mocker.spy(
+            getattr(mlflow, autolog_integrations[integration_name]), "autolog"
+        )
+        autolog_fn_spy.__signature__ = inspect.Signature(params)
+
+        setattr(getattr(mlflow, autolog_integrations[integration_name]), "autolog", autolog_fn_spy)
+        autolog_fn = getattr(getattr(mlflow, autolog_integrations[integration_name]), "autolog")
+        autolog_fn.assert_not_called()
+
+    mlflow.autolog(log_input_example=True, log_model_signature=True)
```
dbczumar, line 713: I think we should be performing `autolog_fn.assert_not_called()` after `mlflow.autolog()` is invoked; this should hold for all libraries except PySpark (for which, as discussed in a comment above, we want to invoke autologging as soon as `mlflow.autolog()` is called).

Otherwise, it's possible that `mlflow.autolog()` calls all of these autologging functions directly and our test doesn't catch it.

tests/tracking/test_fluent.py
```diff
@@ -681,3 +682,42 @@ def test_delete_tag():
     with pytest.raises(MlflowException):
         mlflow.delete_tag("b")
     mlflow.end_run()
+
+
+@pytest.mark.large
+def test_universal_autolog_calls_specific_autologs_correctly(mocker):
```
dbczumar, line 688: I think it would be cleaner to refactor this test case so that it's parameterized by the module name, MLflow autologging function, and supported configs via `@pytest.mark.parametrize`.

dbczumar (changes requested):
@andrewnitu Great progress! Left some comments - let me know if you have questions!

mlflow/tracking/fluent.py
```diff
@@ -1020,3 +1036,26 @@ def _get_experiment_id():
...
+    arg_info = inspect.getargvalues(inspect.currentframe())  # pylint: disable=deprecated-method
+    arg_values = {k: v for k, v in arg_info.locals.items() if k in arg_info.args}
+
+    def setup_autologging(module):
+        integration_module_obj = getattr(mlflow, AUTOLOG_INTEGRATIONS[module.__name__])
+        autolog_fn = getattr(integration_module_obj, "autolog")
+        needed_params = list(inspect.signature(autolog_fn).parameters.keys())
+        filtered = {k: v for k, v in arg_values.items() if k in needed_params}
+
+        autolog_fn(**filtered)
```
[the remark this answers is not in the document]
andrewnitu: yeah, this is a big oversight, not sure how I missed this. I think I was deep into the spark autologging stuff then never wrapped up the remaining work on the universal autolog :(

mlflow/tracking/fluent.py
```diff
@@ -31,6 +34,19 @@
...
+# eg: mxnet.gluon is the actual library, mlflow.gluon is our integration code for it
+AUTOLOG_INTEGRATIONS = {
+    "tensorflow": "tensorflow",
+    "keras": "keras",
+    "mxnet.gluon": "gluon",
+    "xgboost": "xgboost",
+    "lightgbm": "lightgbm",
+    "pyspark": "spark",
+    "sklearn": "sklearn",
+    "fastai": "fastai",
```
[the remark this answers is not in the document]
andrewnitu: makes sense!

mlflow/tracking/fluent.py
```diff
@@ -31,6 +34,19 @@
...
 
+# Mapping of library module name to mlflow module name
+# eg: mxnet.gluon is the actual library, mlflow.gluon is our integration code for it
+AUTOLOG_INTEGRATIONS = {
+    "tensorflow": "tensorflow",
+    "keras": "keras",
+    "mxnet.gluon": "gluon",
+    "xgboost": "xgboost",
+    "lightgbm": "lightgbm",
+    "pyspark": "spark",
```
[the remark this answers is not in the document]
andrewnitu: makes sense, but just curious how the user can interact with it without importing PySpark? Are they given a reference to the session and then they can just run `createDataFrame` on it, for example?

mlflow/tracking/fluent.py
andrewnitu: yup! I have a branch with some docs so far, just needs to be rebased on these new changes

tests/tracking/test_fluent.py
andrewnitu: if we take pyspark as a separate case as described, then yeah we can put `assert_not_called` back as part of this test. Previously, I had to remove it specifically because of pyspark

mlflow/tracking/fluent.py
andrewnitu: this seems to occur mostly for functions defined in C and imported into python using Cython.

Thus, it should definitely suffice to just catch the ValueError and call with no arguments in that case.

mlflow/tracking/fluent.py
andrewnitu: actually i don't think we need this.. we can just check locals() for any variables that each specific autolog integration might accept

mlflow/_spark_autologging.py
```diff
@@ -89,57 +92,67 @@ def _set_run_tag(run_id, path, version, data_format):
...
+            "and ensure you have the mlflow-spark JAR attached to your Spark "
+            "session as described in "
+            "http://mlflow.org/docs/latest/tracking.html#automatic-logging-from-spark-experimental. "
+            "Exception:\n%s" % e
+        )
+
+    # Register context provider for Spark autologging
+    from mlflow.tracking.context.registry import _run_context_provider_registry
+
+    _run_context_provider_registry.register(SparkAutologgingContext)
```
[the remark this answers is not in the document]
andrewnitu: in the case where `_get_current_listener() is None`, we should not trigger this `_run_context_provider_registry.register(SparkAutologgingContext)` according to the existing logic. Leaving this code inside `_listen_for_spark_activity` allows us to maintain this logic, otherwise if we move it outside we would need to check for `get_current_listener()` a second time

tests/tracking/test_fluent.py
```diff
@@ -681,3 +683,98 @@ def test_delete_tag():
...
+            del sys.modules[integration_name]
+        except Exception:
+            pass
+        try:
+            del wrapt.importer._post_import_hooks[integration_name]
+        except Exception:
+            pass
+
+
+@pytest.mark.large
```
dbczumar, line 729: For clarity / debuggability, can we instead parametrize this test by `library_to_mlflow_module`?  This will run a separate test case for each library / MLflow module pair.

tests/tracking/test_fluent.py
```diff
@@ -681,3 +683,98 @@ def test_delete_tag():
...
+        ) as autolog_mock:
+            autolog_mock.side_effect = Exception("asdf")
+            mlflow.autolog()
+            importlib.__import__(integration_name)
+
+    clear_imports_and_import_hooks()
+
+
+@pytest.mark.large
+def test_universal_autolog_calls_specific_autologs_correctly(mocker, library_to_mlflow_module):
```
dbczumar: Ditto here - can we parametrize by `library_to_mlflow_module`? This will run a separate test case for each library / MLflow module pair.

tests/tracking/test_fluent.py
```diff
@@ -681,3 +683,98 @@ def test_delete_tag():
...
+            del wrapt.importer._post_import_hooks[integration_name]
+        except Exception:
+            pass
+
+
+@pytest.mark.large
+def test_universal_autolog_does_not_throw_if_specific_autolog_throws(
+    mocker, library_to_mlflow_module,
+):
+    clear_imports_and_import_hooks()
```
dbczumar: Instead of adding this before and after each test, can we define it as a fixture and auto-use it before each test case? see https://docs.pytest.org/en/stable/fixture.html#autouse-fixtures-xunit-setup-on-steroids

tests/tracking/test_fluent.py
```diff
@@ -681,3 +683,98 @@ def test_delete_tag():
...
+
+    for integration_name in library_to_mlflow_module.keys():
+        with mock.patch(
+            "mlflow." + library_to_mlflow_module[integration_name] + ".autolog"
+        ) as autolog_mock:
+            autolog_mock.side_effect = Exception("asdf")
```
dbczumar, line 737: Can we assert that the mock was called?

tests/tracking/test_fluent.py
```diff
@@ -681,3 +683,105 @@ def test_delete_tag():
...
+        "pyspark": "spark",
+        "sklearn": "sklearn",
+        "xgboost": "xgboost",
+        "lightgbm": "lightgbm",
+        "mxnet.gluon": "gluon",
+    }
+
+
+@pytest.fixture
+def reset_global_states(library_to_mlflow_module):
```
dbczumar, line 703: Can we auto-use this fixture so that we don't have to pass it into each test case? See https://docs.pytest.org/en/stable/fixture.html#autouse-fixtures-xunit-setup-on-steroids

mlflow/tracking/fluent.py
```diff
@@ -1021,3 +1026,50 @@ def _get_experiment_id():
...
+    # the post-import hook also retroactively activates for previously-imported libraries.
+    for module in LIBRARY_TO_AUTOLOG_FN.keys():
+        register_post_import_hook(setup_autologging, module)
+
+    # for pyspark, we activate autologging immediately, without waiting for a module import.
+    # this is because on Databricks a SparkSession already exists and the user can directly
+    #   interact with it, and this activity should be logged.
+    try:
+        spark.autolog()
+        _logger.info("Autologging successfully enabled for spark.")
```
dbczumar, line 1073: There's one more case I hadn't thought about: If PySpark isn't installed, this throws an ImportError. If PySpark is installed, I think we should call `spark.autolog()` immediately; this will handle the Databricks case. Otherwise, we should register an import hook.

dev/small-requirements.txt
```diff
@@ -8,3 +8,20 @@ pytest-localserver==0.5.0
...
+# Required by mlflow.fastai
+fastai==1.0.60
+# Required by mlflow.tensorflow
+tensorflow==1.15.2
+# Required by mlflow.xgboost
+xgboost>=0.82
+# Required by mlflow.lightgbm
+lightgbm==2.3.0
+# Required by mlflow.spark
+pyspark==3.0.0
```
dbczumar, line 27: Our small tests shouldn't depend on heavyweight ML libraries. Can we instead decorate autologging test cases as `@pytest.mark.large` so they're executed as part of our large suites?

andrewnitu: i think the tests are failing because import tensorflow is at the top of the file, so when running the small tests in the file tensorflow is not installed.

it seems to fix this we would either have to extract out the autologging test cases, or mark all the tests in the file as large (which I dont think we want)

dev/large-requirements.txt
```diff
@@ -1,3 +1,20 @@
 ## Large test reqs
 matplotlib
 tf2onnx
+
+# Required by mlflow.keras
+keras==2.3.1
```
dbczumar, line 6: It's a bit confusing since we've refactored our dependency management for tests, but all of these currently reside in https://github.com/mlflow/mlflow/blob/master/dev/extra-ml-requirements.txt and are installed before large tests are executed. We can remove them from this file.

andrewnitu: interesting, the tests were complaining before that tensorflow did not exist before i added it to the small-requirements file.

also, it seems like that file has the wrong version of spark for the tests, I think we need 3.0.0

dbczumar (approved):
LGTM! Thanks @andrewnitu !

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -0,0 +1,150 @@
...
+    tensorflow: "tensorflow",
+    keras: "keras",
+    fastai: "fastai",
+    sklearn: "sklearn",
+    xgboost: "xgboost",
+    lightgbm: "lightgbm",
+    mxnet.gluon: "gluon",
+}
+
+library_to_mlflow_module = {**library_to_mlflow_module_without_pyspark, "pyspark": "spark"}
```
harupy, line 27: Just curious why `pyspark` here is a string, not a module.

andrewnitu: i'll take a look at this and fix in a follow-up PR today. thanks for the comment, we missed this

Issue #3494, opened 2020-10-05
[FR] tracking event callbacks for external observers in R interface of mlflow


yitao-li:
## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

It would be great if the R interface of `mlflow` can provide optional generic callbacks for informing an external observer of tracking events such as setting active run ID, setting active experiment ID, logging metric, etc. If there is no external observer present, then no additional callback will be run.

## Motivation
- What is the use case for this feature?
IDE extensions can act as external observers receiving those callbacks. Users will be able to see current and past experiments and metric associated with them from an external IDE, which will be helpful during development.

- Why is this use case valuable to support for MLflow users in general?
It will be useful for users who run non-trivial number of experiments to have a hierarchical view of runs and experiments in an integrated development environment via an IDE extension.
  
- Why is this use case valuable to support for your project(s) or organization?
Having an integrated view of ongoing and past MLFlow experiments would greatly improve user experience for IDE users.

- Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient)
I don't think there is any existing construct in the R code base of MLflow that can be customized to optionally keep an external observer up-to-date with tracking events such as `set_active_run_id` or `set_active_experiment_id`, so, a small amount of code change is needed for this to happen.

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [x] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interfaces
[...]

PR #3501, opened 2020-10-07
implement optional tracking event callbacks for external observers

yitao-li:

closes #3501 

Optional callbacks related to the proposal in https://github.com/mlflow/mlflow/issues/3494

## How is this patch tested?

Existing unit tests will suffice

Components 

Interface 

Language 

Integrations


javierluraschi:
Test failures are probably unrelated, but let me rerun them to see if we get a clean pass for:

```
 AssertionError: assert 'Invalid filter' in "Invalid clause(s) in filter string: 'IN'. Expected search filter with single comparison operator. e.g. name='myModelName'"
```

javierluraschi (approved):

javierluraschi:
We try to mark R PRs with the label `language/r`, added.

javierluraschi:
@yitao-li can you try rebasing from master? I don't think is possible for this change to break the Python tests, but we want to have tests passing before we merge.

mlflow/R/mlflow/R/tracking-observer.R
```diff
@@ -0,0 +1,13 @@
+# If one or more external observer(s) are present, then inform them of the
+# event. Otherwise do nothing.
+mlflow_register_tracking_event <- function(event_name, data) {
+  observers <- getOption("MLflowObservers")
+  if (length(observers) > 0) {
+    lapply(
```
smurching, line 6: Should we catch exceptions thrown by observers/log warnings? It seems like exceptions thrown by observers shouldn't fail user code

mlflow/R/mlflow/R/tracking-globals.R
```diff
@@ -24,6 +31,8 @@ mlflow_get_active_experiment_id <- function() {
 #' @export
 mlflow_set_tracking_uri <- function(uri) {
   .globals$tracking_uri <- uri
+  mlflow_register_tracking_event("tracking_uri", list(uri = uri))
```
smurching, line 34: Should this be named e.g. "set_tracking_uri" for parity with events like "log_metric"?

mlflow/R/mlflow/R/tracking-globals.R
```diff
@@ -1,5 +1,9 @@
+#' @include tracking-observer.R
+NULL
+
 mlflow_set_active_run_id <- function(run_id) {
```
smurching, line 4: Out of curiosity, why do we register the callback on `set_active_run_id` rather than on `mlflow_start_run`?

smurching (commented):
@yitao-li thanks for the PR! Could we add some tests e.g. under a file called test-tracking-observer.R under https://github.com/mlflow/mlflow/tree/c50b04b53d7d7fcd2b78391fbe95a064205697be/mlflow/R/mlflow/tests/testthat for this new functionality, and document it so that it appears in the R API docs?

See https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst#writing-docs for guidelines on writing docs - our R doc generation process is a bit wonky, you might be able to add a section describing this MLflowObservers option and its usage in [this file](https://github.com/mlflow/mlflow/blob/HEAD/mlflow/R/mlflow/document.R#L42)

I also just want to run this overall proposal by other committers - it seems reasonable & useful to have such callbacks, but just want to make sure there aren't any cons I'm missing. The only downsides I can immediately think of are that it may make it more costly (from an engineering time perspective) to add new fluent APIs if we need to call a corresponding callback.

The interface of passing callback data as a list is also interesting - I'm not an R expert so I'm not sure how user-friendly it is to receive callback information as a list :) but one definite advantage seems to be that we can add new arguments to fluent APIs without breaking existing callbacks (because the new argument will just be a new key-value pair in the dict). 

If you could write up (e.g. in google docs or a github gist) a brief writeup describing the new interfaces in this PR & their pros/cons, along with alternatives considered, it'd also help expedite the review process - thanks!

yitao-li:
@smurching Thanks for the feedback. I'll be more than happy to create R documentation for the callbacks.

Question about documentation: I see the content of R-api.rst is based on what's in mlflow/R/mlflow/Reference_Manual_mlflow.md, but Reference_Manual_mlflow.md is in .gitignore (!!) but there is no Rmd file in the repo that can generate Reference_Manual_mlflow.md.
Can we please update the exact steps needed for creating R documentation (other than "make rdocs")?

yitao-li:
@smurching The writeup for this is in https://gist.github.com/yitao-li/376bf4e4a1ad62f92311679e9567bd1d
but as previously mentioned I would like some clarification on how exactly should documentation be created for mlflow the R package -- currently it's quite unclear

yitao-li:
Tests are passing now.

mlflow/R/mlflow/R/tracking-globals.R
javierluraschi: My personal preference is to prefix absolutely everything, we've done that in packages like `sparklyr` (as in `spark_`, `ml_`, etc), `tensorflow` (as in `tf_`, etc), and more recently `torch` (as in `torch_`, etc.). The reasoning, not having a prefix in R packages is expensive... Once you define something outside of a package, you need to figure out how that might potentially interact with other packages.

For instance, for `log_metric`, we might have to consider how that interacts with other model logging methods from the R community. So far, I don't think there are any package clashes, but one day this could change. Would be nice for instance if one could `log_metric` without using mlflow, then connect mlflow, and the code won't break.

So net-net, I would encourage we keep prefixes in most functions, unless we feel strongly a function is meant to be top-level and deserves the extra work of potentially interfacing with other packages.

mlflow/R/mlflow/R/tracking-observer.R
javierluraschi: I'm 50/50 on this one, since I'm not completely the users of this function should be considered users. Closer description would be developers, which can probably handle just find the exception.

PR #3502, opened 2020-10-08
Pin sqlparse to 0.3.1

harupy:

`TestSqlAlchemyStoreSqlite.test_search_model_versions` fails in the latest version ([0.4.0](https://pypi.org/project/sqlparse/0.4.0/)) of `sqlparse`:

https://github.com/mlflow/mlflow/runs/1223251903#step:4:1610

## How is this patch tested?

(Details)

Components 

Interface 

Language 

Integrations

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
         "docker>=4.0.0",
         "entrypoints",
         # Pin sqlparse for: https://github.com/mlflow/mlflow/issues/3433
-        "sqlparse>=0.3.1",
+        "sqlparse==0.3.1",
         "sqlalchemy<=1.3.13",
         "gorilla",
         "prometheus-flask-exporter",


smurching (approved):
@harupy thanks for the fix -

I was originally going to say: 
> Since dependencies in setup.py affect the end user's Python environment, could we instead be less opinionated about requiring exactly sqlparse==0.3.1 and instead require sqlparse<0.4.0, sqlparse>=0.3.1 for now?

But then realized that `sqlparse<0.4.0, sqlparse>=0.3.1` is equivalent to `sqlparse==0.3.1`. LGTM to unblock the build, would be nice to unpin + fix the offending test in a follow-up (it seems we're asserting on the exact contents of an error message when searching registered models, and the new error message generated when using sqlparse 0.4.1 also looks reasonable, so we may be able to just update the test)

harupy:
@smurching Thanks for merging this PR.

https://github.com/andialbrecht/sqlparse/pull/567: seems to have caused the error.

`IN` is a keyword operator in `sqlparse==0.3.1`, but a comparison operator in `sqlparse==0.4.0`.

https://github.com/mlflow/mlflow/blob/f11a1831806779162b4f4370c8ee71f4c2ccc579/mlflow/utils/search_utils.py#L252-L264

With the current `_invalid_statement_token_search_model_registry` implementation, `IN` is treated as an invalid operator in `sqlparse==0.4.0`.

I'll open a follow-up PR to address this and unpin the sqlparse version.

PR #3513, opened 2020-10-12
Add `mlflow.shap.log_explanation`

harupy:

Implemnet `mlflow.shap.log_explanation`.

## How is this patch tested?

Unit tests

MLflow now supports logging model explanations generated by SHAP.

Components 

Interface 

Language 

Integrations


mlflow/shap.py
```diff
@@ -32,3 +38,75 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+        :caption: Output
+
+        ['shap/base_values.npy', 'shap/shap_values.npy', 'shap/summary_bar_plot.png']
+
+    """
+    import matplotlib.pyplot as plt
+    import shap
+
+    artifact_path = "shap" if artifact_path is None else artifact_path
+    explainer = shap.KernelExplainer(predict_function, shap.kmeans(features, 100))
```
harupy, line 95: If we didnt' sample `features` here with `shpa.kmeans`, it would take really long time to compute shap values (even with a relatively small dataset).

harupy: shap raises a warning if the number of samples is larger than 100:

https://github.com/slundberg/shap/blob/master/shap/explainers/_kernel.py#L76-L79

mlflow/shap.py
```diff
@@ -32,3 +38,75 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+    shap_values = explainer.shap_values(features)
+
+    shap.summary_plot(shap_values, features, plot_type="bar", show=False)
+    fig = plt.gcf()
+    _log_matplotlib_figure(fig, _SUMMARY_BAR_PLOT_FILE_NAME, artifact_path)
+    plt.close(fig)
```
harupy, line 101: `shap.summary_plot` returns nothing so we need to use `plt.gcf` to get the active figure object.

tests/shap/test_shap.py
```diff
@@ -40,3 +83,53 @@ def test_log_matplotlib_figure():
...
+
+    artifact_path_expected = "shap" if artifact_path is None else artifact_path
+    artifacts = set(yield_artifacts(run.info.run_id))
+
+    assert explanation_path == os.path.join(run.info.artifact_uri, artifact_path_expected)
+    assert artifacts == {
+        os.path.join(artifact_path_expected, "base_values.npy"),
+        os.path.join(artifact_path_expected, "shap_values.npy"),
+        os.path.join(artifact_path_expected, "summary_bar_plot.png"),
+    }
```
harupy, line 104: Used a set here to ignore the order of files.

mlflow/shap.py
```diff
@@ -32,3 +38,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+
+    :param predict_function: A function to compute the output of a model.
+    :param features: A matrix of features to compute SHAP values with.
+    :param artifact_path: A run-relative artifact path the explanation is saved to. If unspecified,
+                          defaults to "shap".
+
+    :return: A URI of the logged explanation.
+
+    .. code-block:: python
+        :caption: Example
```
harupy, line 59: I'll add this example code as `mlflow/examples/shap/explain.py` in a follow-up PR.

tests/shap/test_shap.py
```diff
@@ -40,3 +84,63 @@ def test_log_matplotlib_figure():
...
+
+    shap_values = np.load(os.path.join(explanation_path, "shap_values.npy"))
+    base_values = np.load(os.path.join(explanation_path, "base_values.npy"))
+
+    explainer = shap.KernelExplainer(model.predict_proba, shap.kmeans(X, 100))
+    shap_values_expected = explainer.shap_values(X)
+    base_values_expected = np.array(explainer.expected_value)
+
+    np.testing.assert_array_equal(shap_values, shap_values_expected)
+    np.testing.assert_array_equal(base_values, base_values_expected)
```
harupy, line 146: I'm investigating whether or not `KernelExplainer` is really deterministic (I thought it isn't)

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+@experimental
+def log_explanation(predict_function, features, artifact_path=None):
+    """
+    Log the following items generated by SHAP.
+
+        - base values
+        - shap values
+        - summary bar plot
+
+    :param predict_function: A function to compute the output of a model.
```
smurching, line 53: Could we clarify the exact expected input arguments/return type of `predict_function`?

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+    :param features: A matrix of features to compute SHAP values with.
```
smurching, line 54:
```suggestion
    :param features: A matrix of features to compute SHAP values with. The provided features should have shape  
     (nsamples, nfeatures), and can be either a numpy array, pandas DataFrame, or list of lists.
```

Is this correct? Basically the idea is to clarify the expected shape/allowed types of `features`, I tried to infer based on https://shap.readthedocs.io/en/latest/generated/shap.summary_plot.html#shap.summary_plot

Could we clarify

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+    .. code-block:: text
+        :caption: Output
+
+        ['shap/base_values.npy', 'shap/shap_values.npy', 'shap/summary_bar_plot.png']
+
+    """
+    import matplotlib.pyplot as plt
+    import shap
+
+    artifact_path = "shap" if artifact_path is None else artifact_path
```
smurching, line 96:
```suggestion
    artifact_path = "model_explanations_shap" if artifact_path is None else artifact_path
```

Could we call this something more descriptive like "model_explanations_shap"?

mlflow/shap.py
smurching: Could we refactor 100 here into a constant & add a comment in the code explaining the choice of 100? Basically just the reasoning you mentioned above

mlflow/shap.py
smurching: Noob question, since I'm not a matplotlib expert: do we want to get the current figure object, or instead create a new one? It seems it'd be bad if `mlflow.shap.log_explanation` interfered with any active figure objects the user's created thus far

tests/shap/test_shap.py
```diff
@@ -40,3 +94,86 @@ def test_log_matplotlib_figure():
...
+    }
+
+    shap_values = np.load(os.path.join(explanation_path, "shap_values.npy"))
+    base_values = np.load(os.path.join(explanation_path, "base_values.npy"))
+    np.testing.assert_array_equal(shap_values, classifier.shap_values)
+    np.testing.assert_array_equal(base_values, classifier.base_values)
+
+
+@pytest.mark.large
+def test_log_explanation_with_numpy_array(regressor):
```
smurching, line 156: Could we also test an example with a list-of-lists, assuming https://github.com/mlflow/mlflow/pull/3513/files#r504452812 is correct?

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+    """
+    Log the following items generated by SHAP.
+
+        - base values
+        - shap values
+        - summary bar plot
+
+    :param predict_function: A function to compute the output of a model.
+    :param features: A matrix of features to compute SHAP values with.
+    :param artifact_path: A run-relative artifact path the explanation is saved to. If unspecified,
```
smurching, line 55:
```suggestion
    :param artifact_path: The run-relative artifact path to which the explanation is saved. If unspecified,
```

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
     """
     with _log_artifact_contextmanager(out_file, artifact_path) as tmp_path:
         fig.savefig(tmp_path)
+
+
+@experimental
+def log_explanation(predict_function, features, artifact_path=None):
+    """
+    Log the following items generated by SHAP.
```
smurching, line 47:
```suggestion
    Given a ``predict_function`` capable of computing ML model output on the provided ``features``, computes 
    and logs explanations of an ML model's output. In particular, logs the following items generated by SHAP:
```

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+
+        - base values
+        - shap values
+        - summary bar plot
```
smurching, line 51:
```suggestion
        - summary bar plot (shows the average impact of each feature on model output)
```

Could we add brief descriptions like this for each of the logged artifacts to help the users understand them?

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+
+        # print out logged items
+        client = mlflow.tracking.MlflowClient()
+        artifacts = [x.path for x in client.list_artifacts(run.info.run_id, "shap")]
+        print(artifacts)
+
+    .. code-block:: text
+        :caption: Output
+
+        ['shap/base_values.npy', 'shap/shap_values.npy', 'shap/summary_bar_plot.png']
```
smurching, line 90: Nice! In the longer example/follow-up PR, we might want to show how to load back and use the logged base/shap values

mlflow/shap.py
```diff
@@ -32,3 +39,70 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+        - base values
+        - shap values
+        - summary bar plot
+
+    :param predict_function: A function to compute the output of a model.
+    :param features: A matrix of features to compute SHAP values with.
+    :param artifact_path: A run-relative artifact path the explanation is saved to. If unspecified,
+                          defaults to "shap".
+
+    :return: A URI of the logged explanation.
```
smurching, line 58:
```suggestion
    :return: Artifact URI of the logged explanation
```

mlflow/shap.py
smurching: +1, could we also include a screenshot of the generated artifacts in the docs? I think you already figured out how to do this in https://github.com/harupy/mlflow/tree/shap-demo :)

smurching (commented):
A few small comments but largely looks good, thanks @harupy!

mlflow/shap.py
harupy: https://github.com/slundberg/shap/blob/0fe999687f74776cf171e57f0cee4e220e3db4a6/shap/plots/_beeswarm.py#L521

`shap.summary_plot` creates a new figure so it doesn't interfere with active figures.

tests/shap/test_shap.py
harupy: `KernelExplainer` only accepts:

- numpy.array
- pandas.DataFrame
- shap.common.DenseData
- scipy.sparse matrix

https://github.com/slundberg/shap/blob/master/shap/explainers/_kernel.py#L36

I tried to run the test with a list but got this assertion error:

```
AssertionError: Unknown type passed as data object: <class 'list'>
```

mlflow/shap.py
harupy: Should we provide a function to load back the logged values (e.g. `load_explanation`)?

mlflow/shap.py
harupy: Got it :) I'll take a new sceenshot (because I have renamed the bar plot) and embed it in the docstring.

mlflow/shap.py
```diff
@@ -32,3 +41,101 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+        :caption: Output
+
+        ['shap/base_values.npy', 'shap/shap_values.npy', 'shap/summary_bar_plot.png']
+
+    """
+    import matplotlib.pyplot as plt
+    import shap
+
+    artifact_path = _DEFAULT_ARTIFACT_PATH if artifact_path is None else artifact_path
+    background_data = shap.kmeans(features, _BACKGROUND_DATA_NUM_ROWS)
```
harupy, line 129: shap uses a fixed random seed for kmeans.

https://github.com/slundberg/shap/blob/master/shap/utils/_legacy.py#L39

mlflow/shap.py
smurching: Could make sense yeah! I guess the alternative actually is not to use the explanation URI, but rather to remember the run ID & run-relative artifact path of the explanation, and call [download_artifacts](https://www.mlflow.org/docs/latest/python_api/mlflow.tracking.html#mlflow.tracking.MlflowClient.download_artifacts) with those parameters - it'd help to write out the workflow without `load_explanation`, and then we can consider adding an experimental `load_explanation` API

tests/shap/test_shap.py
```diff
@@ -40,3 +94,100 @@ def test_log_matplotlib_figure():
...
+    """
+    Verifies that `log_explanation` does not fail even when `features` has less records than
+    `_MAXIMUM_BACKGROUND_DATA_SIZE`.
+    """
+    model = regressor.model
+    X = regressor.X
+
+    with mlflow.start_run():
+        num_rows = mlflow.shap._MAXIMUM_BACKGROUND_DATA_SIZE - 1
+        mlflow.shap.log_explanation(model.predict, X.iloc[:num_rows])
```
smurching, line 193: Could we also assert on the artifacts logged in this case?

mlflow/shap.py
```diff
@@ -32,3 +41,101 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+    """
+    Given a ``predict_function`` capable of computing ML model output on the provided ``features``,
+    computes and logs explanations of an ML model's output. Explanations are logged as a directory
+    of artifacts containing the following items generated by SHAP:
+
+        - Base values
+        - SHAP values (computed using `shap.KerneExplainer`_)
+        - Summary bar plot (shows the average impact of each feature on model output)
+
+    .. _shap.KerneExplainer: https://github.com/slundberg/shap#kernelexplainer
```
smurching, line 57:
```suggestion
    .. _shap.KernelExplainer: https://github.com/slundberg/shap#kernelexplainer
```

mlflow/shap.py
```diff
@@ -32,3 +41,101 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+
+@experimental
+def log_explanation(predict_function, features, artifact_path=None):
+    """
+    Given a ``predict_function`` capable of computing ML model output on the provided ``features``,
+    computes and logs explanations of an ML model's output. Explanations are logged as a directory
+    of artifacts containing the following items generated by SHAP:
+
+        - Base values
+        - SHAP values (computed using `shap.KerneExplainer`_)
```
smurching, line 54:
```suggestion
        - SHAP values (computed using `shap.KernelExplainer`_)
```

smurching (commented):
Could we add a test that calls `mlflow.shap.log_explanation` without an active run? We can use a try-finally block to ensure the run is ended before the next test, e.g:
```
try:
  mlflow.shap.log_explanation(...)
finally:
  mlflow.end_run()
  assert ... # perform test assertions for artifacts
```

We might also want to add a test on a binary-classification dataset, although.I think the only thing that we might expect to differ in the case of binary classification is the shape of the logged shap values, which is controlled by SHAP, but would still be good to have a test verifying that that code path works (the test itself can serve as a lower-level guarantee/documentation on how to use `log_explanation` with a binary classifier), let me know what you think

Otherwise LGTM!

mlflow/shap.py
harupy: I'll update the example using `download_artifacts` and later add `load_explanation` (if necessary) 👍

mlflow/shap.py
```diff
@@ -32,3 +41,124 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
     """
     with _log_artifact_contextmanager(out_file, artifact_path) as tmp_path:
         fig.savefig(tmp_path)
+
+
+@experimental
+def log_explanation(predict_function, features, artifact_path=None):
+    r"""
```
harupy, line 48: Added `r` here to suppress the following sphinx error:

```
/home/circleci/project/mlflow/shap.py:docstring of mlflow.shap.log_explanation:82: WARNING: Explicit markup ends without a blank line; unexpected unindent.
```

https://app.circleci.com/pipelines/github/mlflow/mlflow/1652/workflows/c94ca5f6-9715-4840-9a32-46be18f5ba20/jobs/1654

mlflow/shap.py
```diff
@@ -82,15 +81,31 @@ def predict_function(X) -> pred:
         have shape (# samples, # features), and can be either of the array-like object
         types listed above.
 
+        .. note::
+            Background data for `shap.KernelExplainer`_ is generated by subsumpling ``features``
```
smurching, line 85:
```suggestion
            Background data for `shap.KernelExplainer`_ is generated by subsampling ``features``
```

mlflow/shap.py
```diff
@@ -32,3 +41,142 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+    computes and logs explanations of an ML model's output. Explanations are logged as a directory
+    of artifacts containing the following items generated by `SHAP`_ (SHapley Additive
+    exPlanations).
+
+        - Base values
+        - SHAP values (computed using `shap.KernelExplainer`_)
+        - Summary bar plot (shows the average impact of each feature on model output)
+
+    :param predict_function:
+        A function to compute the output of a model (e.g. ``predict`` method of
```
smurching, line 59:
```suggestion
        A function to compute the output of a model (e.g. ``predict_proba`` method of scikit-learn classifiers
```

Tiny nit: could we update this example to specify that users need to use ``predict_proba`` for sklearn classifiers? Just because I think that's a bit subtler than the regressor case, where the only choice is to pass ``predict``.

mlflow/shap.py
```diff
@@ -32,3 +41,142 @@ def _log_matplotlib_figure(fig, out_file, artifact_path=None):
...
+
+        .. code-block:: python
+
+            def predict_function(X) -> pred:
+                ...
+
+        - ``X``: An array-like object whose shape should be (# samples,
+          # features).
+        - ``pred``: An array-like object whose shape should be (# samples) for
+          a regressor or (# classes, # samples) for a classifier.
```
smurching, line 70:
```suggestion
          a regressor or (# classes, # samples) for a classifier. For a classifier, the values in ``pred`` should
          correspond to the predicted probability of each class.
```

smurching (approved):
Two more comments, LGTM once they're addressed, thanks @harupy!

mlflow/shap.py
harupy: Makes sense. I'll udpate the example.

mlflow/shap.py
harupy: thanks for the catch 😅

PR #3542, opened 2020-10-17
Add examples for shap integration

harupy:
Signed-off-by: harupy <17039389+harupy@users.noreply.github.com>

## What changes are proposed in this pull request?

- A follow-up for #3513 
- Add an example script for the shap integration.

## How is this patch tested?

`test_examples.py`

### Is this a user-facing change?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [x] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [ ] `rn/feature` - A new user-facing feature worth mentioning in the release notes
[...]

PR #3085, opened 2020-07-11
support xgboost for R

lorenzwalthert:
## What changes are proposed in this pull request?

This is a re-implementation of my previous attempt #2367. I hope @javierluraschi can review this since he already had a look at the previous implementation. Support saving and loading xgboost models from R. The models should be 100% compatible with the Python xgboost models, i.e. after saving with R or Python, they should be loadable with the other language too. The implementation closely follows the one for Keras models in the R package and the xgboost models from the Python package.

## How is this patch tested?

Unit tests, namely three types:

- save from R, predict from R.
- save from R, predict via CLI.
- save from R, load in R with python through reticulate.

The last two test types are important to ensure the model formats in `model/` are compatible between Python and R.

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

The model flavour xgboost is now also available for R. This means that xgboost models saved from R can be loaded in Python and vice versa. You can hence also deploy xgboost models trained in R as python_function on Microsoft Azure ML or Amazon AWS Sagemaker.

### What component(s) does this PR affect?

- [ ] UI
- [ ] CLI
- [ ] API
- [ ] REST-API
- [ ] Examples
- [ ] Docs
- [ ] Tracking
- [ ] Projects
- [ ] Artifacts
- [x] Models
- [ ] Scoring
- [ ] Serving
- [x] R
- [ ] Java
- [ ] Python

### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [ ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [x] `rn/feature` - A new user-facing feature worth mentioning in the release notes
- [ ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes
- [ ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes

PR #3544, opened 2020-10-19
[R] Fix broken h2o archive link

harupy:

The h2o archive link (`https://cran.r-project.org/src/contrib/h2o_3.30.1.3.tar.gz`) is broken. This PR fixes it.

https://github.com/mlflow/mlflow/pull/3542/checks?check_run_id=1273179681#step:9:141

Screen Shot 2020-10-19 at 13 46 22

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/R/mlflow/tests/testthat/test-model-h2o.R
+++ b/mlflow/R/mlflow/tests/testthat/test-model-h2o.R
@@ -20,8 +20,6 @@ for (pkg in pkgs) {
 # Pin h2o to prevent version-mismatch between python and R
 install.packages("https://cran.r-project.org/src/contrib/h2o_3.30.1.3.tar.gz", repos=NULL, type="source")
 
-h2o::h2o.init()
-
 model <- h2o::h2o.randomForest(
   x = predictors, y = prediction, training_frame = h2o::as.h2o(train)
 )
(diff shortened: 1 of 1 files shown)


smurching (approved):
LGTM to unbreak the build - maybe @yitao-li or @lorenzwalthert can comment on the version-pinning approach we're using here & whether it's standard/stable in R?

yitao-li:
Hey @smurching @harupy I feel like I might be missing something there:

in my original implementation of H2O model mismatch in H2O patch version was ignored (see https://github.com/mlflow/mlflow/pull/3416#discussion_r507949778), and that logic was removed in https://github.com/mlflow/mlflow/pull/3477/files (and '>=' replaced with '==' as well for version requirement) -- I think a potential problem with that approach is `utils::packageVersion("h2o")` returns the version of the H2O R package, but that version is not necessarily in sync with the H2O Python package all the time, so, requiring an exact match might not work some of the times

lorenzwalthert:
All I can say is that for XGBoost as implemented in #3085, we write the XGBoost version to the conda env file with the patch version striped and hope it exists for R and Python - but I don't think we test this anywhere. I don't know if you also write h20 versions to conda eng files. For testing, we use the latest version of both R and Python and it does not seem as this ever caused trouble. 

Unrelated: My experience with XGBoost was anyways that the conda environment was not used, which I think is a bug.

lorenzwalthert:
Also, @smurching I think once a package is archived from CRAN using this link is save. But this means you can't ever pin a current version (because these don't live under /archived). More elegant is to use `remotes::install_version(version = ...)` and pin a version there if this is possible with the other install options you want to supply.

PR #3552, opened 2020-10-20
(Issue #3224) Allow keras models to be saved with SavedModel format

### LINKED ISSUE #3224: [BUG] Cannot save Keras/TF_2.x model as SavedModel format using mlflow.keras.log_model with kwargs { "save_format": "tf" }

**Please fill in this bug report template to ensure a timely and thorough response.**

### Willingness to contribute

- [ x] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.

### System information

MLflow Version: 1.10.0
Keras version: 2.4.0
TensorFlow version: 2.3.0
Python Version: 3.7.6
Operating System: Darwin - 19.4.0

### Describe the problem

Cannot save TensorFlow 2.3.0  Keras model as TensorFlow SaveModel format  using mlflow.keras.log_model since it does not honor kwargs { "save_format": "tf" }.  The saved model is still H5.  The  suffix is [hard-coded to h5](https://github.com/mlflow/mlflow/blob/master/mlflow/keras.py#L207) and not dependent on the value of save_format.

TensorFlow SaveModel format is now the default for Keras TensorFlow 2.

### Code to reproduce issue
```
import mlflow
import mlflow.keras

def baseline_model():
    model = Sequential()
    model.add(Dense(2, input_dim=2, kernel_initializer="normal", activation="relu"))
    model.compile(loss="mean_squared_error", optimizer="adam")
    return model
model = baseline_model()
    mlflow.keras.log_model(model, "tensorflow-model", **{ "save_format": "tf" })
```

### Other info / logs

`{ "save_format": "tf" }`  is correctly passed to `keras_model.save`  at https://github.com/mlflow/mlflow/blob/master/mlflow/keras.py#L208 but output is H5.

```
        with tempfile.NamedTemporaryFile(suffix='.h5') as f:
            keras_model.save(f.name, **kwargs)
            f.flush()  # force flush the data
            shutil.copyfile(src=f.name, dst=model_path)
```

Components 

###  Related Issues
- https://github.com/mlflow/mlflow/issues/3246 - [FR] Serialization format for TensorFlow 2.x (Keras) model should be Save Model per TF 2.0 doc

skylarbpayne:
Allow keras models to be saved in SavedModel format (via `save_format="tf"`).

fix issue described in #3224 

closes #3224 

## How is this patch tested?
Additional tests added. We also tested with an internal use case and verified this let us save keras models with preprocessing layers in SavedModel format.

## Release Notes
The default save format in keras is `SavedModel`, but mlflow has set the default to HDF5. If you would like to use `SavedModel`, you can pass `save_format="tf"` to `log_model` which will be propagated to keras.

By default, mlflow.keras will save models in HDF5 format. Prior to this change HDF5 was the only supported format, but now users can use `save_format="tf"` to use `SavedModel` format.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -37,7 +37,7 @@ FLAVOR_NAME = "keras"
 _CUSTOM_OBJECTS_SAVE_PATH = "custom_objects.cloudpickle"
 _KERAS_MODULE_SPEC_PATH = "keras_module.txt"
 # File name to which keras model is saved
-_MODEL_SAVE_PATH = "model.h5"
+_MODEL_SAVE_PATH = "model"
 # Conda env subpath when saving/loading model
 _CONDA_ENV_SUBPATH = "conda.yaml"
 
@@ -390,6 +390,14 @@ def _load_model(model_path, keras_module, **kwargs):
             pickled_custom_objects = cloudpickle.load(in_f)
             pickled_custom_objects.update(custom_objects)
             custom_objects = pickled_custom_objects
+    
+    # TODO: consider removing this fallback when versions of mlflow which save models with h5 prefix are deprecated
+    # To ensure backwards compatibility, if the _MODEL_SAVE_PATH does not exist, we will look for a file with the
+    # "h5" file extension.
+    if not os.path.exists(model_path):
+        # TODO: log a warning?
+        model_path = f"{model_path}.h5"
+    
     from distutils.version import StrictVersion
 
     if StrictVersion(keras_module.__version__.split("-")[0]) >= StrictVersion("2.2.3"):
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -16,7 +16,6 @@ import shutil
 import pandas as pd
 
 from distutils.version import LooseVersion
-from distutils.version import StrictVersion
 from mlflow import pyfunc
 from mlflow.models import Model
 from mlflow.models.model import MLMODEL_FILE_NAME
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -456,7 +456,7 @@ def _load_pyfunc(path):
         import keras
 
         keras_module = keras
-    
+
     # By default, we assume the save_format is h5 for backwards compatibility
     save_format = "h5"
     save_format_path = os.path.join(path, _KERAS_SAVE_FORMAT_PATH)
@@ -475,7 +475,9 @@ def _load_pyfunc(path):
             with graph.as_default():
                 with sess.as_default():  # pylint:disable=not-context-manager
                     K.set_learning_phase(0)
-                    m = _load_model(path, keras_module=keras_module, save_format=save_format, compile=False)
+                    m = _load_model(
+                        path, keras_module=keras_module, save_format=save_format, compile=False
+                    )
                     return _KerasModelWrapper(m, graph, sess)
         else:
             K.set_learning_phase(0)
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -389,7 +389,7 @@ def _save_custom_objects(path, custom_objects):
         cloudpickle.dump(custom_objects, out_f)
 
 
-def _load_model(model_path, keras_module, save_format: str, **kwargs):
+def _load_model(model_path, keras_module, save_format, **kwargs):
     keras_models = importlib.import_module(keras_module.__name__ + ".models")
     custom_objects = kwargs.pop("custom_objects", {})
     custom_objects_path = None
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -16,6 +16,7 @@ import shutil
 import pandas as pd
 
 from distutils.version import LooseVersion
+from distutils.version import StrictVersion
 from mlflow import pyfunc
 from mlflow.models import Model
 from mlflow.models.model import MLMODEL_FILE_NAME
@@ -217,16 +218,20 @@ def save_model(
         f.write(keras_module.__name__)
 
     # By default, Keras uses the SavedModel format -- specified by "tf"
-    # Storing this in the flavor conf allows proper loading of the model later
-    save_format = kwargs.get("save_format", "tf")
+    # However, we choose to align with prior default of mlflow, HDF5
+    save_format = kwargs.get("save_format", "h5")
 
     # save keras save_format to path/data/save_format.txt
     with open(os.path.join(data_path, _KERAS_SAVE_FORMAT_PATH), "w") as f:
         f.write(save_format)
 
-    # save keras model to path/data/model.h5
+    # save keras model
+    # To maintain prior behavior, when the format is HDF5, we save
+    # with the h5 file extension. Otherwise, model_path is a directory
+    # where the saved_model.pb will be stored (for SavedModel format)
+    file_extension = ".h5" if save_format == "h5" else ""
     model_subpath = os.path.join(data_subpath, _MODEL_SAVE_PATH)
-    model_path = os.path.join(path, model_subpath)
+    model_path = os.path.join(path, model_subpath) + file_extension
     if path.startswith("/dbfs/"):
         # The Databricks Filesystem uses a FUSE implementation that does not support
         # random writes. It causes an error.
@@ -401,9 +406,9 @@ def _load_model(model_path, keras_module, save_format: str, **kwargs):
             pickled_custom_objects.update(custom_objects)
             custom_objects = pickled_custom_objects
 
-    # TODO: remove this fallback after older versions of mlflow which save with h5 suffix
-    # are deprecated
-    if save_format == "h5" and not os.path.exists(model_path):
+    # If the save_format is HDF5, then we save with h5 file
+    # extension to align with prior behavior of mlflow logging
+    if save_format == "h5":
         model_path = model_path + ".h5"
 
     from distutils.version import StrictVersion
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -228,11 +228,15 @@ def save_model(
     else:
         keras_model.save(model_path, **kwargs)
 
+    # By default, Keras uses the SavedModel format -- specified by "tf"
+    # Storing this in the flavor conf allows proper loading of the model later
+    save_format = kwargs.get("save_format", "tf")
     # update flavor info to mlflow_model
     mlflow_model.add_flavor(
         FLAVOR_NAME,
         keras_module=keras_module.__name__,
         keras_version=keras_module.__version__,
+        save_format=save_format,
         data=data_subpath,
     )
 
@@ -375,7 +379,7 @@ def _save_custom_objects(path, custom_objects):
         cloudpickle.dump(custom_objects, out_f)
 
 
-def _load_model(model_path, keras_module, **kwargs):
+def _load_model(model_path, keras_module, save_format: str, **kwargs):
     keras_models = importlib.import_module(keras_module.__name__ + ".models")
     custom_objects = kwargs.pop("custom_objects", {})
     custom_objects_path = None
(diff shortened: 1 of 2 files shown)


tests/keras/test_keras_model_export.py
```diff
@@ -506,3 +507,38 @@ def test_sagemaker_docker_model_scoring_with_default_conda_env(model, model_path
...
+    mlflow.keras.save_model(keras_model=keras_model, path=model_path, save_format="tf")
+    _, args, kwargs = keras_model.save.mock_calls[0]
+    # Ensure that save_format propagated through
+    assert kwargs["save_format"] == "tf"
+    # Ensure that the saved model does not have h5 extension
+    assert not args[0].endswith(".h5")
+
+
+@pytest.mark.large
+def test_save_and_load_model_with_tf_save_format(tf_keras_model, model_path):
```
juntai-zheng, line 528: Could we add tests for loading back via `pyfunc` here as well?

juntai-zheng (commented):
Thanks for your contribution! Looks generally good, just left one comment for adding code-coverage testing.

mlflow/keras.py
```diff
@@ -375,7 +389,7 @@ def _save_custom_objects(path, custom_objects):
         cloudpickle.dump(custom_objects, out_f)
 
 
-def _load_model(model_path, keras_module, **kwargs):
+def _load_model(model_path, keras_module, save_format: str, **kwargs):
```
harupy, line 392:
```suggestion
def _load_model(model_path, keras_module, save_format, **kwargs):
```

Only type-annotating save_format looks a bit weird to me. Should we remove it?

tests/keras/test_keras_model_export.py
```diff
@@ -506,3 +507,38 @@ def test_sagemaker_docker_model_scoring_with_default_conda_env(model, model_path
...
+    """Ensures that keras models saved with save_format="tf" can be loaded."""
+    mlflow.keras.save_model(keras_model=tf_keras_model, path=model_path, save_format="tf")
+    model_conf_path = os.path.join(model_path, "MLmodel")
+    model_conf = Model.load(model_conf_path)
+    flavor_conf = model_conf.flavors.get(mlflow.keras.FLAVOR_NAME, None)
+    assert flavor_conf is not None
+    assert flavor_conf.get("save_format") == "tf"
+    assert not os.path.exists(
+        os.path.join(model_path, "data", "model.h5")
+    ), "TF model was saved with HDF5 format; expected SavedModel"
```
harupy, line 538: These assert error messages are actually nice because they clarify what is being asserted :)

tests/keras/test_keras_model_export.py
```diff
@@ -171,6 +171,7 @@ def save(self, path, **kwargs):
     class FakeKerasModule(object):
         __name__ = "some.test.keras.module"
         __version__ = "42.42.42"
+        __package__ = ""
```
harupy, line 174: Is this attribute used somewhere?

mlflow/keras.py
skylarbpayne: Sure!

tests/keras/test_keras_model_export.py
skylarbpayne: Good catch! Left over from previous diff. Removed!

tests/keras/test_keras_model_export.py
skylarbpayne: Great suggestion! I added some more parameters to the `test_model_save_load` test so it covers all combinations (no param, "h5", and "tf").

This uncovered a subtle bug that I've now fixed! For clarity:

Keras will compile the model if there is an optimizer on loading, unless we explicitly request it to _not_ compile. But during prediction, compilation is forced if optimizer state is saved. In the case that the optimizer state was not saved (and thus we can't compile on load), then Keras will log a warning if we set `compile=True`. As such, the fix I applied was to call load_model with `compile=True` if `SavedModel` format was used.

tests/keras/test_keras_model_export.py
```diff
@@ -210,21 +210,30 @@ def _import_module(name, **kwargs):
...
-    mlflow.keras.save_model(keras_model, model_path)
+    kwargs = {"save_format": save_format} if save_format else {}
+    mlflow.keras.save_model(keras_model, model_path, **kwargs)
     # Loading Keras model
     model_loaded = mlflow.keras.load_model(model_path)
-    assert type(keras_model) == type(model_loaded)
+    # When saving as SavedModel, we actually convert the model
+    # to a slightly different format, so we cannot assume it is
+    # exactly the same.
+    if save_format is not "tf":
```
harupy, line 235:
```suggestion
    if save_format != "tf":
```

should we use != here?

tests/keras/test_keras_model_export.py
```diff
@@ -210,21 +210,30 @@ def _import_module(name, **kwargs):
             assert x == b
 
 
-@pytest.mark.skip("Unskip this test once https://github.com/h5py/h5py/issues/1732 is fixed")
-@pytest.mark.parametrize("build_model", [model, tf_keras_model])
+@pytest.mark.parametrize(
+    "build_model,save_format",
+    # Unskip these parameters once https://github.com/h5py/h5py/issues/1732 is fixed
```
harupy, line 215:
```suggestion
    # TODO: Unskip these parameters once https://github.com/h5py/h5py/issues/1732 is fixed
```

This is a nit but can you add `TODO` here to make it easier to notice this later?

skylarbpayne:
@harupy -- it appears the test is now failing; but it looks like it's still "equal" within what I would expect from the underlying type (`float32`) -- I'll change the comparison to compare with a tolerance; LMK if there are any concerns.

harupy:
Screen Shot 2020-11-09 at 13 57 23

skylarbpayne:
> `expected` and `model.predict(x)` are almost equal. Agree with comparing with the tolerance. Is this related to the `SavedModel` format?

I believe so, but candidly, I'm not 100% sure. The reason I suspect this is the case is:

1. The _types_ are not exactly the same when loading in `SavedModel` format -- i.e. if you assert strict equality before/after save/load, it will fail.
2. This discrepancy only occurs with SavedModel
3. There is some post processing that occurs to [trim potential incompatibilities](https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/saved_model/save.py#L907) when saving in SavedModel format

However, I'm not 100% sure that this is the cause; in either case, asserting without tolerance can make the test fail due to subtle changes, so it's probably a good idea either way

tests/keras/test_keras_model_export.py
```diff
@@ -210,25 +210,34 @@ def _import_module(name, **kwargs):
...
+    # When saving as SavedModel, we actually convert the model
+    # to a slightly different format, so we cannot assume it is
+    # exactly the same.
+    if save_format != "tf":
+        assert type(keras_model) == type(model_loaded)
+    assert expected == pytest.approx(model_loaded.predict(x))
     # Loading pyfunc model
     pyfunc_loaded = mlflow.pyfunc.load_model(model_path)
-    assert all(pyfunc_loaded.predict(x).values == expected)
+    assert expected == pytest.approx(pyfunc_loaded.predict(x).values)
```
harupy, line 240: Can we use `numpy.testing.assert_array_almost_equal` instead?

skylarbpayne: Good suggestion! Looking at the docstring on that, it suggests using something else for consistent handling; so I'm trying `np.testing.assert_allclose`. I used that setting relative tolerance to `1e-5`. Any stricter comparison and it will fail; if I'm honest, I expected slightly more tolerance (e.g. ~1e-6). But I'm not sure I have a good argument either way.

Do you think relative tolerance of `1e-5` is reasonable here?

harupy: Looks reasonable :)

harupy:
@skylarbpayne Thanks for the update. Could you add a test like below to ensure that we can still load models logged by `mlflow<=1.11.0` where the `save_format` field doesn't exist?

```python
@pytest.mark.large
def test_xxx(tf_keras_model, model_path):
    mlflow.keras.save_model(keras_model=tf_keras_model, path=model_path, save_format="h5")
    model_conf_path = os.path.join(model_path, "MLmodel")
    model_conf = Model.load(model_conf_path)
    flavor_conf = model_conf.flavors.get(mlflow.keras.FLAVOR_NAME)
    assert flavor_conf is not None
    del flavor_conf["save_format"]
    model_conf.save(model_conf_path)

    model_loaded = mlflow.keras.load_model(model_path)
    assert tf_keras_model.to_json() == model_loaded.to_json()
```

skylarbpayne:
> @skylarbpayne Thanks for the update. Could you add a test like below to ensure that we can still load models logged by `mlflow<=1.11.0` where the `save_format` field doesn't exist?
> 
> ```python
> @pytest.mark.large
> def test_xxx(tf_keras_model, model_path):
>     mlflow.keras.save_model(keras_model=tf_keras_model, path=model_path, save_format="h5")
>     model_conf_path = os.path.join(model_path, "MLmodel")
>     model_conf = Model.load(model_conf_path)
>     flavor_conf = model_conf.flavors.get(mlflow.keras.FLAVOR_NAME)
>     assert flavor_conf is not None
>     del flavor_conf["save_format"]
>     model_conf.save(model_conf_path)
> 
>     model_loaded = mlflow.keras.load_model(model_path)
>     assert tf_keras_model.to_json() == model_loaded.to_json()
> ```

Good catch! Added!

harupy:
https://github.com/mlflow/mlflow/pull/3552/checks?check_run_id=1378317395#step:4:100

```
>       np.testing.assert_allclose(actual_scoring_response, expected, rtol=3e-6)
E       AssertionError: 
E       Not equal to tolerance rtol=3e-06, atol=0
E       
E       Mismatched elements: 1 / 150 (0.667%)
E       Max absolute difference: 1.1920929e-07
E       Max relative difference: 3.9720207e-06
E        x: array([[ 1.651972],
E              [ 1.536899],
E              [ 1.5181  ],...
E        y: array([[ 1.651972],
E              [ 1.536899],
E              [ 1.5181  ],...
```

Is `rtol=3e-6` too strict?

skylarbpayne:
> @skylarbpayne The tensorflow check failed.
> 
> https://github.com/mlflow/mlflow/pull/3552/checks?check_run_id=1378317395#step:4:100
> 
> ```
> >       np.testing.assert_allclose(actual_scoring_response, expected, rtol=3e-6)
> E       AssertionError: 
> E       Not equal to tolerance rtol=3e-06, atol=0
> E       
> E       Mismatched elements: 1 / 150 (0.667%)
> E       Max absolute difference: 1.1920929e-07
> E       Max relative difference: 3.9720207e-06
> E        x: array([[ 1.651972],
> E              [ 1.536899],
> E              [ 1.5181  ],...
> E        y: array([[ 1.651972],
> E              [ 1.536899],
> E              [ 1.5181  ],...
> ```
> 
> Is `rtol=3e-6` too strict?

Oops. That was from a stale commit where I was testing. Yes. The most strict I could find that worked was `4e-6`; I opted to just use `1e-5`, though

tests/keras/test_keras_model_export.py
```diff
@@ -508,3 +516,53 @@ def test_sagemaker_docker_model_scoring_with_default_conda_env(model, model_path
...
+        os.path.join(model_path, "data", "model")
+    ), "Expected directory containing saved_model.pb"
+
+    model_loaded = mlflow.keras.load_model(model_path)
+    assert tf_keras_model.to_json() == model_loaded.to_json()
+
+
+@pytest.mark.large
+def test_load_without_save_format(tf_keras_model, model_path):
+    """Ensures that keras models with save_format can still be loaded."""
```
harupy, line 558:
```suggestion
    """Ensures that keras models without save_format can still be loaded."""
```
nit

skylarbpayne: done

harupy (approved):
LGTM. Thanks for the contribution 👍

PR #3557, opened 2020-10-20
Logging/Loading TorchScripted model

shrinath-suresh:

Current `mlflow.pytorch` library doesnt not have provision for logging/saving the torchscripted model. The new inclusion facilitates logging/saving of the torchscripted model into mlflow.

## How is this patch tested?

Tested with MNIST handwritten digits and IrisClassification example. Added necessary unit tests for the same.

Users can now log and load torchscripted model using `mlflow.pytorch` library.

Components 

Interface 

Language 
- [x] `language/python`: Python APIs and clients

Integrations


mlflow/pytorch/__init__.py
```diff
@@ -325,7 +325,11 @@ def save_model(
         f.write(pickle_module.__name__)
     # Save pytorch model
     model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
-    torch.save(pytorch_model, model_path, pickle_module=pickle_module, **kwargs)
+    if isinstance(pytorch_model, torch.jit.ScriptModule):
+        model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
+        torch.jit.ScriptModule.save(pytorch_model, model_path)
```
harupy, line 330: Is there any difference between `torch.jit.ScriptModule.save` and `torch.jit.save`?

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
+# pylint: disable=W0221
+# pylint: disable=W0613
+# pylint: disable=W0223
```
harupy, line 3: why ignore these pyling rules?

examples/pytorch/torchscript/IrisClassification/README.md
```diff
@@ -0,0 +1,40 @@
+## MNIST example with MLFlow
```
harupy, line 1:
```suggestion
## MNIST example with MLflow
```
nit

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
...
+    scripted_model = torch.jit.script(model)  # scripting the model
+    scripted_model = train_model(scripted_model, args.epochs, X_train, y_train)
+    test_model(scripted_model, X_test, y_test)
+
+    mlflow.tracking.set_tracking_uri(args.tracking_uri)
+    mlflow.start_run()
+    mlflow.pytorch.log_model(scripted_model, "model")  # logging scripted model
+    uri_path = mlflow.get_artifact_uri()
+    mlflow.pytorch.load_model(uri_path + "/model")  # loading scripted model
+    mlflow.end_run()
```
harupy, line 104: Could we use `with mlflow.start_run()` instead of `mlflow.start_run()` and `mlflow.end_run()`?

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def test_model(model, X_test, y_test):
+    with torch.no_grad():
+        predict_out = model.forward(X_test)
+        _, predict_y = torch.max(predict_out, 1)
+
+        print(
+            "prediction accuracy", float(accuracy_score(y_test.cpu(), predict_y.cpu())),
+        )
```
harupy, line 76: Do we need to call `model.eval` before the prediction?

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
...
+
+    return X_train, X_test, y_train, y_test
+
+
+def train_model(model, epochs, X_train, y_train):
+    criterion = nn.CrossEntropyLoss()
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.01)
+
+    for epoch in range(epochs):
+        out = model.forward(X_train)
```
harupy, line 57: Which one is more common, `model.forward(X_train)` vs. `model(X_train)`?

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
...
+
+    parser.add_argument(
+        "--tracking-uri", type=str, default="http://localhost:5000/", help="mlflow tracking uri"
+    )
+    parser.add_argument(
+        "--epochs", type=int, default=100, help="number of epochs to run (default: 50)"
+    )
+
+    args = parser.parse_args()
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
harupy, line 90:
```suggestion
```

because `device` is already defined at line 32.

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
...
+        print(
+            "prediction accuracy", float(accuracy_score(y_test.cpu(), predict_y.cpu())),
+        )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Iris Classification Torchscripted model")
+
+    parser.add_argument(
+        "--tracking-uri", type=str, default="http://localhost:5000/", help="mlflow tracking uri"
```
harupy, line 83: If we set `tracking_uri` to `http://localhost:5000`, we'd need to launch an MLflow server before running this example, which I thinks is a bit cumbersome.

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+    parser.add_argument(
+        "--epochs", type=int, default=100, help="number of epochs to run (default: 50)"
```
harupy, line 86:
```suggestion
        "--epochs", type=int, default=100, help="number of epochs to run (default: 100)"
```

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,104 @@
...
+    X_train, X_test, y_train, y_test = prepare_data()
+    scripted_model = torch.jit.script(model)  # scripting the model
+    scripted_model = train_model(scripted_model, args.epochs, X_train, y_train)
+    test_model(scripted_model, X_test, y_test)
+
+    mlflow.tracking.set_tracking_uri(args.tracking_uri)
+    mlflow.start_run()
+    mlflow.pytorch.log_model(scripted_model, "model")  # logging scripted model
+    uri_path = mlflow.get_artifact_uri()
+    mlflow.pytorch.load_model(uri_path + "/model")  # loading scripted model
```
harupy, line 103:
```suggestion
    model_path = mlflow.get_artifact_uri("model")
    mlflow.pytorch.load_model(model_path)  # loading scripted model
```

should work.

examples/pytorch/torchscript/MNIST/README.md
```diff
@@ -0,0 +1,40 @@
+## MNIST example with MLFlow
```
harupy, line 1:
```suggestion
## MNIST example with MLFlow
```

ditto

examples/pytorch/torchscript/MNIST/mnist_torchscript.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python
+# coding: utf-8
```
harupy, line 2:
```suggestion
```

examples/pytorch/torchscript/IrisClassification/iris_classification.py
harupy: Could you make predictions using the loaded model?

harupy:
Thanks for the PR! Left a few comments. Other things we need to consider are:

1. Does the current `_PyTorchWrapper` implementation work when `pytorch_model` is a scripted model?

https://github.com/mlflow/mlflow/blob/5e22def3e476070012cf2e6ef7d750eb0bec0e60/mlflow/pytorch/__init__.py#L468-L494

2. How can we load back a model in pytorch < 1.6.0 where`torch.load` doesn't automatically detect a model type (eager or scripted)?

```
try:
    # load the model as an eager model.
    return torch.load(...)
except:
    # If fails, assume the model is a scripted model
    return torch.jit.load(...)
```

shrinath-suresh:
> Thanks for the PR! Left a few comments. Other things we need to consider are:
> 
> 1. Does the current `_PyTorchWrapper` implementation work when `pytorch_model` is a scripted model?
> 
> https://github.com/mlflow/mlflow/blob/5e22def3e476070012cf2e6ef7d750eb0bec0e60/mlflow/pytorch/__init__.py#L468-L494
> 
> 1. How can we load back a model via `mlflow.pytorch.load_model` in pytorch < 1.6.0 where`torch.load` doesn't automatically detect a model type (eager or scripted)? One apporach could be:
> 
> ```python
> if torch_version >= (1, 6, 0):
>     model = torch.load(path)
> else:
>     try:
>         # load the model as an eager model.
>         model = torch.load(...)
>     except:
>         # If fails, assume the model is a scripted model
>         model = torch.jit.load(...)
> ```

Thanks for the review and comments. I am working on addressing the review comments. I will check on the above mentioned items and get back to you.

examples/pytorch/torchscript/IrisClassification/iris_classification.py
shrinath-suresh: `W0223: Method '_forward_unimplemented' is abstract in class 'Module' but is not overridden (abstract-method)` - Eventhough the forward method is implemented in IrisClassifier class, pylint throws the error. Hence suppressed W0223. 

i guess W0221 and W0613 are not need for this example. Mostly on pytorch lightning examples, batch_idx would be added as part of training step and since its not used the pylint error is thrown. We cannot remove the batch_idx argument as it will result in non matching abstract method signature. 

But for this example W0221 and W0613  can be removed i believe, i will add a new commit for this change.

examples/pytorch/torchscript/IrisClassification/iris_classification.py
shrinath-suresh: All the examples expects user to launch mlflow ui before running the example. this argument just facilitates user to provide a tracking uri.. For example if mlflow server is running in Amazon Ec2, user can simply specify an argument `--tracking-uri http:<EC2-URL>:5000`.

shrinath-suresh:
> Thanks for the PR! Left a few comments. Other things we need to consider are:
> 
> 1. Does the current `_PyTorchWrapper` implementation work when `pytorch_model` is a scripted model?
> 
> https://github.com/mlflow/mlflow/blob/5e22def3e476070012cf2e6ef7d750eb0bec0e60/mlflow/pytorch/__init__.py#L468-L494
> 
> 1. How can we load back a model via `mlflow.pytorch.load_model` in pytorch < 1.6.0 where`torch.load` doesn't automatically detect a model type (eager or scripted)? One apporach could be:
> 
> ```python
> if torch_version >= (1, 6, 0):
>     model = torch.load(path)
> else:
>     try:
>         # load the model as an eager model.
>         model = torch.load(...)
>     except:
>         # If fails, assume the model is a scripted model
>         model = torch.jit.load(...)
> ```

@harupy  I updated the [IrisClassification](https://github.com/mlflow/mlflow/pull/3557/commits/fa1676e3156b8e5adfb957e00230691b5cb179e5) example to use `_PyTorchWrapper` class for prediction. It is working as expected. I am working on the next item now.

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,113 @@
...
+    mlflow.tracking.set_tracking_uri(args.tracking_uri)
+    with mlflow.start_run() as run:
+        mlflow.pytorch.log_model(scripted_model, "model")  # logging scripted model
+        model_path = mlflow.get_artifact_uri("model")
+        loaded_pytorch_model = mlflow.pytorch.load_model(model_path)  # loading scripted model
+        model.eval()
+        with torch.no_grad():
+            test_datapoint = pd.DataFrame([[4.4000, 3.0000, 1.3000, 0.2000]])
+            actual = "setosa"
+            pywrap = _PyTorchWrapper(loaded_pytorch_model)
```
harupy, line 109: `_PyTorchWrapper` is not meant to be used like this. You can use `mlfow.pyfunc.load_model` instead which returns a model wrapped by `_PyTorchWrapper`.

examples/pytorch/torchscript/MNIST/mnist_torchscript.py
```diff
@@ -0,0 +1,194 @@
+# pylint: disable=W0223
+
+from __future__ import print_function
```
harupy, line 3:
```suggestion
from __future__ import print_function
```

because we no longer support python 2

examples/pytorch/torchscript/IrisClassification/iris_classification.py
shrinath-suresh: Okie. Loading the scripted model and making prediction is already part of Unit Test - https://github.com/mlflow/mlflow/blob/fa1676e3156b8e5adfb957e00230691b5cb179e5/tests/pytorch/test_pytorch_model_export.py#L307 . Shall i revert to the old code where we make the prediction or you want me to update the code and make predictions based on pyfunc model ?

shrinath-suresh:
> > Thanks for the PR! Left a few comments. Other things we need to consider are:
> > 
> > 1. Does the current `_PyTorchWrapper` implementation work when `pytorch_model` is a scripted model?
> > 
> > https://github.com/mlflow/mlflow/blob/5e22def3e476070012cf2e6ef7d750eb0bec0e60/mlflow/pytorch/__init__.py#L468-L494
> > 
> > 
> > 1. How can we load back a model via `mlflow.pytorch.load_model` in pytorch < 1.6.0 where`torch.load` doesn't automatically detect a model type (eager or scripted)? One apporach could be:
> > 
> > ```python
> > if torch_version >= (1, 6, 0):
> >     model = torch.load(path)
> > else:
> >     try:
> >         # load the model as an eager model.
> >         model = torch.load(...)
> >     except:
> >         # If fails, assume the model is a scripted model
> >         model = torch.jit.load(...)
> > ```
> 
> @harupy I updated the [IrisClassification](https://github.com/mlflow/mlflow/pull/3557/commits/fa1676e3156b8e5adfb957e00230691b5cb179e5) example to use `_PyTorchWrapper` class for prediction. It is working as expected. I am working on the next item now.

Tested the torchscripted IrisExample in all 3 versions. In torch 1.5 and 1.6 am able to load the model directly using `torch.load`. Only in 1.4, we need `torch.jit.load()` . As suggested, we can add a version check for this.

examples/pytorch/torchscript/IrisClassification/iris_classification.py
harupy: I was expecting something like:

```python
loaded_pytorch_model = mlflow.pytorch.load_model(model_path)
preds = loaded_pytorch_model(X)

print(preds)
```

shrinath-suresh: Thats how i had it earlier - https://github.com/mlflow/mlflow/pull/3557/commits/62690812d7dca3b7566936681373a3cc7a82b281 . will revert this commit.

shrinath-suresh:
> > > Thanks for the PR! Left a few comments. Other things we need to consider are:
> > > 
> > > 1. Does the current `_PyTorchWrapper` implementation work when `pytorch_model` is a scripted model?
> > > 
> > > https://github.com/mlflow/mlflow/blob/5e22def3e476070012cf2e6ef7d750eb0bec0e60/mlflow/pytorch/__init__.py#L468-L494
> > > 
> > > 
> > > 1. How can we load back a model via `mlflow.pytorch.load_model` in pytorch < 1.6.0 where`torch.load` doesn't automatically detect a model type (eager or scripted)? One apporach could be:
> > > 
> > > ```python
> > > if torch_version >= (1, 6, 0):
> > >     model = torch.load(path)
> > > else:
> > >     try:
> > >         # load the model as an eager model.
> > >         model = torch.load(...)
> > >     except:
> > >         # If fails, assume the model is a scripted model
> > >         model = torch.jit.load(...)
> > > ```
> > 
> > 
> > @harupy I updated the [IrisClassification](https://github.com/mlflow/mlflow/pull/3557/commits/fa1676e3156b8e5adfb957e00230691b5cb179e5) example to use `_PyTorchWrapper` class for prediction. It is working as expected. I am working on the next item now.
> 
> Tested the torchscripted IrisExample in all 3 versions. In torch 1.5 and 1.6 am able to load the model directly using `torch.load`. Only in 1.4, we need `torch.jit.load` . As suggested, we can add a version check for this.

@harupy Since, we need to perform `>=` comparison for version, i thought of using `packaging.version` utility. Looks like `packaging` pip package is not used in mlflow repo. Is it okay to add it in `small-requirements.txt` or we should perform string parse and do the comparison. Your thoughts ?

harupy:
I think we can use `distutils.version.LooseVersion`.

harupy:
Does the current `_PytorchWrapper` implementation works for both eager and scripted model?

shrinath-suresh:
> Does the current `_PytorchWrapper` implementation work for both eager and scripted model?

Yes , the unit test covers both eager and scripted implementation. Anyways , i will cross check once again and update it here.

examples/pytorch/torchscript/IrisClassification/conda.yaml
```diff
@@ -0,0 +1,14 @@
...
+dependencies:
+- python=3.8.2
+- pytorch=1.6.0
+- torchvision=0.7.0
+- pip
+- pip:
+  - sklearn
+  - cloudpickle==1.6.0
+  - boto3
+name: mlflow-env
```
harupy, line 14: Can you remove the `name` field here? I think it's unncessary.

mlflow/pytorch/__init__.py
```diff
@@ -15,6 +15,7 @@
 import cloudpickle
 import numpy as np
 import pandas as pd
+from packaging.version import Version
```
harupy, line 18: Can you remove packaging here as we discussed?

shrinath-suresh: yeah sure  i will replace it with LooseVersion and upload the latest code today.

mlflow/pytorch/__init__.py
```diff
@@ -363,11 +368,17 @@ def _load_model(path, **kwargs):
...
+    pytorch_version = torch.__version__
 
     if os.path.isdir(path):
         # `path` is a directory containing a serialized PyTorch model and a text file containing
         # information about the pickle module that should be used by PyTorch to load it
         model_path = os.path.join(path, "model.pth")
+
+        local_path = os.path.dirname(path)
+        pytorch_conf = _get_flavor_configuration(model_path=local_path, flavor_name=FLAVOR_NAME)
+        pytorch_version = pytorch_conf["pytorch_version"]
```
harupy, line 380: Why do we need to use the pytorch version used to saves the model?

mlflow/pytorch/__init__.py
```diff
@@ -395,7 +406,16 @@ def _load_model(path, **kwargs):
...
         model_path = path
 
-    return torch.load(model_path, **kwargs)
+    if Version(pytorch_version) >= Version('1.5.0'):
+        return torch.load(model_path, **kwargs)
+    else:
+        try:
+            # load the model as an eager model.
+            return torch.load(model_path, **kwargs)
+        except Exception: # pylint: disable=W0703
```
harupy, line 415: Can you replace `W0703` to `catching-too-general-exception`?

shrinath-suresh: sure

mlflow/pytorch/__init__.py
shrinath-suresh: https://github.com/chauhang/mlflow/blob/1eec412236ae8f06c2dabd2dde28175ed6bb06a1/tests/pytorch/test_pytorch_model_export.py#L438 - This test specifically alters the pytorch version and without this condition, pytorch version would be set to 1.6 and  loading the model using `torch.load`  results in exception.

mlflow/pytorch/__init__.py
shrinath-suresh: Updated the code.

mlflow/pytorch/__init__.py
harupy: I think this issue can be fixed like this:
: https://github.com/harupy/mlflow/commit/cba164e1989e7857170a26752538a31bd94fd7c5

shrinath-suresh: Thanks . Let me check.

mlflow/pytorch/__init__.py
```diff
@@ -363,6 +368,7 @@ def _load_model(path, **kwargs):
     :param kwargs: Additional kwargs to pass to the PyTorch ``torch.load`` function.
     """
     import torch
+    pytorch_version = torch.__version__
```
harupy, line 371: Do we need to define this variable? Can we directly pass `torch.__version__` to `LooseVersion`

shrinath-suresh: Done

examples/pytorch/torchscript/IrisClassification/conda.yaml
```diff
@@ -0,0 +1,13 @@
...
+- pytorch
+dependencies:
+- python=3.8.2
+- pytorch=1.6.0
+- torchvision=0.7.0
+- pip
+- pip:
+  - sklearn
+  - cloudpickle==1.6.0
+  - boto3
```
harupy, line 13: nit: linebreak

mlflow/pytorch/__init__.py
```diff
@@ -325,7 +326,11 @@ def save_model(
         f.write(pickle_module.__name__)
     # Save pytorch model
     model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
-    torch.save(pytorch_model, model_path, pickle_module=pickle_module, **kwargs)
+    if isinstance(pytorch_model, torch.jit.ScriptModule):
+        model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
```
harupy, line 330: Can we remove `model_path` at line 330 because it's the same as the one at line 328?

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,107 @@
...
+
+        print(
+            "\nprediction accuracy", float(accuracy_score(y_test.cpu(), predict_y.cpu())),
+        )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Iris Classification Torchscripted model")
+
+    parser.add_argument(
```
smurching, line 80: Similar to the discussion in https://github.com/mlflow/mlflow/pull/3601#discussion_r513088940, in order to follow MLflow best practices could we remove tracking URI as an argument to the script & also remove it from the MLproject file?

examples/pytorch/torchscript/MNIST/MLproject
```diff
@@ -0,0 +1,20 @@
...
+
+conda_env: conda.yaml
+
+entry_points:
+  main:
+    parameters:
+      epochs: {type: int, default: 5}
+      batch_size: {type: int, default: 64}
+      learning_rate: {type: float, default: 1e-3}
+      tracking_uri: {type: str, default: 'http://localhost:5000'}
```
smurching, line 11:
```suggestion
```

examples/pytorch/torchscript/IrisClassification/MLproject
```diff
@@ -0,0 +1,15 @@
+name: iris-classification
+
+conda_env: conda.yaml
+
+entry_points:
+  main:
+    parameters:
+      epochs: {type: int, default: 50}
+      tracking_uri: {type: str, default: 'http://localhost:5000'}
```
smurching, line 9:
```suggestion
```

examples/pytorch/torchscript/IrisClassification/iris_classification.py
```diff
@@ -0,0 +1,107 @@
...
+    model = IrisClassifier()
+    model = model.to(device)
+    X_train, X_test, y_train, y_test, target_names = prepare_data()
+    scripted_model = torch.jit.script(model)  # scripting the model
+    scripted_model = train_model(scripted_model, args.epochs, X_train, y_train)
+    test_model(scripted_model, X_test, y_test)
+
+    mlflow.tracking.set_tracking_uri(args.tracking_uri)
+    with mlflow.start_run() as run:
+        mlflow.pytorch.log_model(scripted_model, "model")  # logging scripted model
```
smurching, line 98:
```suggestion
        mlflow.pytorch.log_model(scripted_model)  # logging scripted model
```

Nit: I think the default model path is "model", so no need to specify the model path here :)

mlflow/pytorch/__init__.py
```diff
@@ -325,7 +326,10 @@ def save_model(
         f.write(pickle_module.__name__)
     # Save pytorch model
     model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
-    torch.save(pytorch_model, model_path, pickle_module=pickle_module, **kwargs)
+    if isinstance(pytorch_model, torch.jit.ScriptModule):
```
smurching, line 329: Could we update the docstrings of `log_model` and `load_model` to mention that you can now also log a TorchScript model (i.e. the output of `torch.jit.trace` or `torch.jit.script`)?

smurching: It'd also be good to update the documentation of ``mlflow.pytorch.load_model`` to clarify that we load back either an eager model (subclass of `torch.nn.Module`) or a TorchScripted model, depending on whether an eager or TorchScripted model was originally logged.

smurching (commented):
Small comments otherwise looks good!

harupy:
@shrinath-suresh Looks like the conlicts block GitHub Actions.

examples/pytorch/torchscript/IrisClassification/iris_classification.py
harupy: @smurching @shrinath-suresh 

`artifact_path` is not an optional argument.

https://github.com/mlflow/mlflow/blob/master/mlflow/pytorch/__init__.py#L69

examples/pytorch/torchscript/IrisClassification/iris_classification.py
shrinath-suresh: removed tracking uri argument and updated readme files with necessary details

examples/pytorch/torchscript/MNIST/MLproject
shrinath-suresh: removed tracking uri argument and updated readme files with necessary details

examples/pytorch/torchscript/IrisClassification/MLproject
shrinath-suresh: removed tracking uri argument and updated readme files with necessary details

examples/pytorch/torchscript/IrisClassification/iris_classification.py
shrinath-suresh: @harupy Thanks harupy. You are right. its not an optional argument.

mlflow/pytorch/__init__.py
shrinath-suresh: Updated the doc strings by mentioning the support of scripted model. In the example section, added the code to log and save model as well

shrinath-suresh: We will raise a separate PR for updating documentation of `mlflow.pytorch`

smurching (approved):
@shrinath-suresh would you be able to reintroduce the `scripted_model` fixture to fix the tests?

I've pushed also pushed a small edit to your docstring - LGTM once the tests are fixed!

mlflow/pytorch/__init__.py
```diff
@@ -391,6 +392,35 @@ def save_model(
...
+            saved_extra_files_dir_subpath = "artifacts"
+            for extra_file in extra_files:
+                _download_artifact_from_uri(
+                    artifact_uri=extra_file, output_path=tmp_extra_files_dir.path()
+                )
+            shutil.move(
+                tmp_extra_files_dir.path(), posixpath.join(path, saved_extra_files_dir_subpath)
+            )
+
+    torch.save(pytorch_model, model_path, pickle_module=pickle_module, **kwargs)
```
harupy: test

harupy: @shrinath-suresh Can you remove `torch.save` here?

shrinath-suresh:
@smurching Thank you very much. I have fixed the tests.

PR #3573, opened 2020-10-21
Final part of code snippets for fluent and MlflowClient APIs

dmatrix:

This is the final part of all the code snippets for the fluent and MlflowClient APIs

This is the larger effort to provide example code snippets for MLflow APIs

- [x ] `rn/documentation` - A user-facing documentation change worth mentioning in the release notes


mlflow/tracking/client.py
```diff
@@ -142,6 +142,57 @@ def get_metric_history(self, run_id, key):
...
+            experiment_id = "0"
+            run = client.create_run(experiment_id)
+            print("run_id: {}".format(run.info.run_id))
+            print("--")
+
+            # Log couple of metrics, update their initial value, and fetch each
+            # logged metrics' history.
+            for k, v in [("m1", 1.5), ("m2", 2.5)]:
+                client.log_metric(run.info.run_id, k, v)
+                client.log_metric(run.info.run_id, k, v+1)
```
harupy, line 171:
```suggestion
                client.log_metric(run.info.run_id, k, v + 1)
```

mlflow/tracking/client.py
```diff
@@ -142,6 +142,57 @@ def get_metric_history(self, run_id, key):
         :param key: Metric name within the run
 
         :return: A list of :py:class:`mlflow.entities.Metric` entities if logged, else empty list
+
+        .. code-block:: python
+            :caption: Example
+
+            from mlflow.tracking import MlflowClient
+
+            def print_metric_info(history):
```
harupy, line 151:
```suggestion
            def print_metric_history(metric_history):
```

mlflow/tracking/client.py
```diff
@@ -1092,6 +1420,42 @@ def list_registered_models(
...
+
+            import mlflow
+            from mlflow.tracking import MlflowClient
+
+            def print_model_info(models):
+            for m in models:
+                print("--")
+                print("name: {}".format(m.name))
+                print("tags: {}".format(m.tags))
+                print("description: {}".format(m.description))
```
harupy, line 1435:
```suggestion
            def print_model_info(models):
                for m in models:
                    print("--")
                    print("name: {}".format(m.name))
                    print("tags: {}".format(m.tags))
                    print("description: {}".format(m.description))
```

mlflow/tracking/client.py
```diff
@@ -1216,6 +1702,46 @@ def delete_registered_model_tag(self, name, key):
...
+                for rm in r_models:
+                    print("name: {}".format(rm.name))
+                    print("tags: {}".format(rm.tags))
+
+            mlflow.set_tracking_uri("sqlite:///mlruns.db")
+            client = MlflowClient()
+
+            # Register a couple of models with respective names and tags
+            for name, tags in [("name1", {"t1": "t1"}),("name2", {"t2": "t2"})]:
+            client.create_registered_model(name, tags)
```
harupy, line 1723:
```suggestion
            for name, tags in [("name1", {"t1": "t1"}),("name2", {"t2": "t2"})]:
                client.create_registered_model(name, tags)
```

mlflow/tracking/client.py
```diff
@@ -1316,6 +1882,50 @@ def update_model_version(self, name, version, description=None):
...
+            rfr = RandomForestRegressor(**params)
+
+            # Log MLflow entities
+            with mlflow.start_run() as run:
+                mlflow.log_params(params)
+                mlflow.sklearn.log_model(rfr, artifact_path="sklearn-model")
+
+            # Create a new version of the rfr model under the registered model name
+            model_uri = "runs:/{}/sklearn-model".format(run.info.run_id)
+            mv = client.create_model_version(name, model_uri, run.info.run_id)
```
harupy, line 1910: `client` is not defined in this code snippet.

harupy: I think `create_registered_model` needs to be called before `create_model_version`.

mlflow/tracking/client.py
```diff
@@ -1421,6 +2229,39 @@ def search_model_versions(self, filter_string):
...
+
+            # Register model name in the model registry
+            client = MlflowClient()
+            client.create_registered_model(name)
+
+            # Create a new version of the rfr model under the registered model name
+            # fetch valid stages
+            model_uri = "runs:/{}/models/sklearn-model".format(run.info.run_id)
+            mv = client.create_model_version(name, model_uri, run.info.run_id)
+            stages = client.get_model_version_stages(name, int(mv.version))
```
harupy, line 2258: Would this line raise an error if we removed `int` and passed `mv.version` as a string to `get_model_version_stages `?

mlflow/tracking/client.py
dmatrix: done!

mlflow/tracking/client.py
dmatrix: ah, joys of cut-n-paste in the block code...:)

mlflow/tracking/client.py
dmatrix: ditto.

mlflow/tracking/client.py
dmatrix: Yeah, missed from the cut-n-paste from [tests](https://github.com/dmatrix/mlflow-tests/tree/master/py/mlflow/apis/client)
```
# Register model name in the model registry
client = MlflowClient()
client.create_registered_model(name)
```

mlflow/tracking/client.py
dmatrix: Yes, it expects an int.

dmatrix: Let me try it.

mlflow/tracking/client.py
```diff
@@ -1017,6 +1143,66 @@ def search_runs(
...
+            from mlflow.tracking import MlflowClient
+            from mlflow.entities import ViewType
+
+            def print_run_info(runs):
+                for r in runs:
+                    print("run_id: {}".format(r.info.run_id))
+                    print("lifecycle_stage: {}".format(r.info.lifecycle_stage))
+                    print("metrics: {}".format(r.data.metrics))
+
+                    # Exclude mlflow tags
```
harupy, line 1160:
```suggestion
                    # Exclude mlflow system tags
```

to be more precise?

mlflow/tracking/client.py
dmatrix: It seems like get_model_version(), update_model_version(), and transition_model_version_stage() can take mv.version, with explict cast. Will change it.

mlflow/tracking/client.py
dmatrix: Agreed!

mlflow/tracking/client.py
```diff
@@ -142,6 +142,57 @@ def get_metric_history(self, run_id, key):
...
+            experiment_id = "0"
+            run = client.create_run(experiment_id)
+            print("run_id: {}".format(run.info.run_id))
+            print("--")
+
+            # Log couple of metrics, update their initial value, and fetch each
+            # logged metrics' history.
+            for k, v in [("m1", 1.5), ("m2", 2.5)]:
+                client.log_metric(run.info.run_id, k, v)
+                client.log_metric(run.info.run_id, k, v + 1)
```
harupy, line 171:
```suggestion
                client.log_metric(run.info.run_id, k, v, step=0)
                client.log_metric(run.info.run_id, k, v + 1, step=1)
```

because some users might think "why overwrite the same metric here"?

dmatrix: Do we need to specify step=0 since it defaults to 0 if the step is omitted in the argument? So these two statements should achieve the same result: 

```
client.log_metric(run.info.run_id, k, v)
client.log_metric(run.info.run_id, k, v + 1, step=1)
```
Is there a specific reason that I'm missing why you have `step=0`?

harupy: sorry it's just my preference...

harupy: > Maybe we should explain a bit more what step=<value> does in the parameter description. What do you think?

I agree, we could do that as a separate (or follow-up) PR.

dmatrix: Cool, will fix it, and yes we can make that as a separate PR. There quite a few places where the parameters' semantics are missing, and a beginner may miss its merits.

harupy: Thanks!

mlflow/tracking/client.py
```diff
@@ -142,6 +142,57 @@ def get_metric_history(self, run_id, key):
...
+            :caption: Example
+
+            from mlflow.tracking import MlflowClient
+
+            def print_metric_info(history):
+                for m in history:
+                    print("name: {}".format(m.key))
+                    print("value: {}".format(m.value))
+                    print("timestamp: {}".format(m.timestamp))
+                    print("--")
```
harupy, line 156:
```suggestion
                    print("name: {}".format(m.key))
                    print("value: {}".format(m.value))
                    print("timestamp: {}".format(m.timestamp))
                    print("value: {}".format(m.step))
                    print("--")
```

to make sure the specifed step is logged?

dmatrix: Sure, why not.

mlflow/tracking/client.py
```diff
@@ -1076,6 +1366,49 @@ def delete_registered_model(self, name):
...
+                    print("name: {}".format(rm.name))
+                    print("tags: {}".format(rm.tags))
+                    print("description: {}".format(rm.description))
+
+            mlflow.set_tracking_uri("sqlite:///mlruns.db")
+            client = MlflowClient()
+
+            # Register a couple of models with respective names, tags, and descriptions
+            for name, tags, desc in [("name1", {"t1": "t1"}, 'description1'),
+                        ("name2", {"t2": "t2"}, 'description2')]:
```
harupy, line 1388:
```suggestion
            for name, tags, desc in [("name1", {"t1": "t1"}, 'description1'),
                                     ("name2", {"t2": "t2"}, 'description2')]:
```

harupy: nit

mlflow/tracking/client.py
```diff
@@ -1349,15 +2019,130 @@ def delete_model_version(self, name, version):
...
+                    print("run_id: {}".format(m.run_id))
+                    print("current_stage: {}".format(m.current_stage))
+
+            mlflow.set_tracking_uri("sqlite:///mlruns.db")
+
+            # Create two runs and log MLflow entities
+            with mlflow.start_run() as run1:
+                params = {"n_estimators": 3, "random_state": 42}
+                rfr = RandomForestRegressor(**params)
+                mlflow.log_params(params)
```
harupy, line 2043:
```suggestion
                rfr = RandomForestRegressor(**params).fit([[0, 1]], [2])
                mlflow.log_params(params)
```

Can we call `fit` using toy data to prevent users from wondering "why fit isn't called?" + make the example more realistic?

harupy: In this example, it doesn't matter whether the model is fitted or not though.

dmatrix: We use this `rfr = RandomForestRegressor(**params)` model in many API calls here. Arguably, if we call `rfr`.fit()` here, we might as well make it ubiquitous elsewhere, for the same argument can be used there too, especially with "create_model_* etc. WDYT?

harupy: Let's just leave `rfr` unfitted in this PR.

harupy: And update them all later in a separate PR.

dmatrix: I like the idea of having the `fit` call in other places too. Let's go with your suggestion of filling another PR where we retrofit this `fit` call.

harupy (approved):
LGTM :)

PR #3575, opened 2020-10-22
Fix bug related with bad alignment of history entries when "initial_epoch" is different from zero.

garciparedes:

Usually, the fit process of `tensorflow`/`keras` models has been performed without modifying the `initial_epoch` parameter (zero by default). But, in some cases that parameter should be set (in cases such as model retraining,fine-tuning, etc.). 

The `kerar.callbacks.History` is reset before starting any fitting process so the old entries are flushed and the index of the new ones does not match with the corresponding epoch. To fix that, the `History` object has a `epoch` attribute that acts as a mapper from the current indexes to the corresponding epochs. Then, `mlflow` should perform that reverse-mapping process when logging the corresponding model epoch. This merge request includes the needed changes to do that. 

## How is this patch tested?

I think I don't have enough project knowledge to properly write a test to validate the behavior of this changes (help needed). Nevertheless, the applied changes don't break any existing test. 

Components 

Interface 

Language 

Integrations


harupy:
@garciparedes Thanks for the PR. Do you have an example where the current keras autolog produces an incorrect result?

garciparedes:
Hi @harupy, Yes! I can provide an example based on the example provided on the `mlflow.keras.autolog()` function:
```python
import mlflow
import mlflow.keras
# Build, compile, enable autologging, and train your model
keras_model = ...
keras_model.compile(optimizer="rmsprop", loss="mse", metrics=["accuracy"])
# autolog your metrics, parameters, and model
mlflow.keras.autolog()
results = keras_model.fit(
    x_train, y_train, epochs=20, initial_epoch=10, batch_size=128, validation_data=(x_val, y_val)
)
``` 
This example raises an `IndexError` because `mlflow` is trying to log the 10th epoch, which corresponds to the 0th position in the `keras.callbacks.History` object, reversely mapped on `history.epoch` that is [10] in that case. 

Then,  applying `history.index(10)` the resulting index is the desired one (0th) and the error is fixed. Note that when `initial_epoch` is not set or is set to `0` then the `history.epoch` "mapping" provides one-to-one values so everything continues working fine.

garciparedes:
Hi! Is there any news about this PR?

garciparedes:
Hi again! Is there any news since my last comment? I think this issue should be fixed so if there is anything I can do to help you with it, let me know please!

harupy:
@garciparedes Sorry for the delayed response. Can you write a test to ensure the proposed change fixes the issue?

garciparedes:
Hi @harupy! Yes, when I write it I'll let you know. Thanks! :slightly_smiling_face:

garciparedes:
Hi @harupy, I've already included the corresponding related tests. If you have any question about the changes feel free to ask me about it.

Offtopic: After writing the tests I realized that the bug I detected only happens when the `initial_epoch` is not `0` and also some `EarlyStopping` callback is used. :slightly_smiling_face:

harupy:
@garciparedes Thanks for adding the tests!

> Offtopic: After writing the tests I realized that the bug I detected only happens when the initial_epoch is not 0 and also some EarlyStopping callback is used. 🙂

Could you elaborate on this?

garciparedes:
> Could you elaborate on this?
Yes! As you can see on [this](https://github.com/mlflow/mlflow/pull/3575/files#diff-d129df3008150ad7a86a33f51c8758e6b03c9467022b91d20902903afef0e9f0R697) line, if there are not any `EarlyStopping` callback the current logic doesn't allow to reach the lines which threw the `IndexError`. As I said before, the error threw while trying to access the corresponding metrics without taking  into account the offset generated by the `initial_epoch` variable.

garciparedes:
As I think the link doesn't redirects properly to the lines I want to refer I'll include it here:
```python
if callback_attrs is None:
    return return
```
That is, if the callback is not `EarlyStopping`-like, which the `callback_attrs` is set to `None` according to the `_get_early_stop_callback_attrs` function, so the `_get_early_stop_callback_attrs` returns and the `history.history[key][...]` is not accessed.

harupy (approved):
@garciparedes Thanks for the explanation. The proposed changes LGTM :)

garciparedes:
Fine! Thanks for accepting my changes @harupy! :wink:

PR #3582, opened 2020-10-23
Sklearn autologging: Fix behavior when a child and its parent are both patched

### LINKED ISSUE #3574: sklearn autolog and TfidfVectorizer do not work together
When running this minimal example with mlflow autolog:

```
from sklearn.feature_extraction.text import TfidfVectorizer
import mlflow

mlflow.sklearn.autolog()

X_train = ['blah blah blah blah', 'ho ho ho ho']
tfidf_count = TfidfVectorizer()
tfidf_count.fit(X_train)
tfidf_count.transform(X_train).toarray()
```

I get a strange error that I cannot understand about `RecursionError: maximum recursion depth exceeded`. However if I remove the autolog statement and run this:

```
from sklearn.feature_extraction.text import TfidfVectorizer
import mlflow

# mlflow.sklearn.autolog()

X_train = ['blah blah blah blah', 'ho ho ho ho']
tfidf_count = TfidfVectorizer()
tfidf_count.fit(X_train)
tfidf_count.transform(X_train).toarray()
```

everything works just fine.

I am on scikit-learn==0.23.2, mlflow==1.11.0.

dbczumar:

Fixes https://github.com/mlflow/mlflow/issues/3574 by addressing the following two issues:

1. In gorilla 0.3.0, applying a patch that stores a hit to a parent class (e.g., `CountVectorizer`) forces `get_original_attribute` of its children to refer to the parent class attribute, even if we patch the child later on. I.e.

This ordering is okay
```
gorilla.Patch(<TfidfVectorizer>)
gorilla.Patch(<CountVectorizer>)
```

but this ordering is not

```
gorilla.Patch(<CountVectorizer>)
gorilla.Patch(<TfidfVectorizer>)
```

This appears to be caused by the following line that prevents the original attribute from being overwritten once it's set: https://github.com/christophercrouzet/gorilla/blob/0c895b34311b6546bde35854e24862c001279add/gorilla.py#L280. As a result, setting the attribute on the parent, which sets the original attribute on the child if the child does not already define it, means that the child is always stuck with the parent's original attribute. This seems to be fixed in the `master` branch of gorilla but has not been released. Either we can pull in the copy from `master` or pull in a copy from 0.3.0 that removes this problematic `hasattr` check.

2. In `fit_mlflow` and `fit_predict`, we perform `get_original_attribute(self, func_name)`. In the case where we're in a super class function (e.g., `CountVectorizer.fit_transform()`), `self` still refers to the subclass (e.g., `TfidfVectorizer`). As a result, calling `get_original_attribute` in the patched super class method fetches invokes the patched subclass method, which creates the infinite recursion problem observed in the issue description. I've filed https://github.com/mlflow/mlflow/pull/3582/ to address this problem by fetching the original function attribute by reference to a class rather than by reference to `self`. This PR still needs docs & tests, and it needs to incorporate a fix for (1) before it's mergeable.

## How is this patch tested?

Unit tests

## Release Notes

Fixes an infinite recursion bug in scikit-learn autologging caused by invoking a `fit()` method on a model class that includes a call to `super.fit()` as part of its implementation (see https://github.com/mlflow/mlflow/issues/3574).

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -725,6 +725,11 @@ def autolog(log_input_example=False, log_model_signature=True):
         )
 
     def fit_mlflow(self, clazz, func_name, *args, **kwargs):
+        """
+        Autologging function that performs model training by executing the training method
+        referred to be `func_name` on the instance of `clazz` referred to by `self` & records
+        MLflow parameters, metrics, tags, and artifacts to a corresponding MLflow Run.
+        """
         should_start_run = mlflow.active_run() is None
         if should_start_run:
             try_mlflow_log(mlflow.start_run)
@@ -893,8 +898,14 @@ def autolog(log_input_example=False, log_model_signature=True):
 
     def patched_fit(self, clazz, func_name, *args, **kwargs):
         """
-        To be applied to a sklearn model class that defines a `fit` method and
-        inherits from `BaseEstimator` (thereby defining the `get_params()` method)
+        Autologging patch function to be applied to a sklearn model class that defines a `fit`
+        method and inherits from `BaseEstimator` (thereby defining the `get_params()` method)
+
+        :param clazz: The scikit-learn model class to which this patch function is being applied for
+                      autologging (e.g., `sklearn.linear_model.LogisticRegression`)
+        :param func_name: The function name on the specified `clazz` that this patch is overriding
+                          for autologging (e.g., specify "fit" in order to indicate that
+                          `sklearn.linear_model.LogisticRegression.fit()` is being patched)
         """
         with _SklearnTrainingSession(clazz=clazz, allow_children=False) as t:
             if t.should_log():
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -8,7 +8,6 @@ Python (native) `pickle <https://scikit-learn.org/stable/modules/model_persisten
 :py:mod:`mlflow.pyfunc`
     Produced for use by generic pyfunc-based deployment tools and batch inference.
 """
-import gorilla
 import os
 import logging
 import pickle
@@ -26,6 +25,7 @@ from mlflow.models.utils import ModelInputExample, _save_example
 from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE, INTERNAL_ERROR
 from mlflow.protos.databricks_pb2 import RESOURCE_ALREADY_EXISTS
 from mlflow.tracking.artifact_utils import _download_artifact_from_uri
+from mlflow.utils import gorilla
 from mlflow.utils.annotations import experimental
 from mlflow.utils.environment import _mlflow_conda_env
 from mlflow.utils.model_utils import _get_flavor_configuration
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -957,7 +957,13 @@ def autolog(log_input_example=False, log_model_signature=True):
     for class_def in estimators_to_patch:
         for func_name in ["fit", "fit_transform", "fit_predict"]:
             if hasattr(class_def, func_name):
-                original = getattr(class_def, func_name)
+                # To enable patching of `fit_*` functions on model classes whose parents
+                # have already patched `fit_*`, we overwrite preexisting patches. To do this
+                # safely, we must first check for the presence of an existing original function
+                # that was overwritten by a previous patch; this ensures that we don't attempt
+                # to call an already-patched function during training
+                original = gorilla.get_original_attribute(class_def, func_name) or getattr(class_def, func_name)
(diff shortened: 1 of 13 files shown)


harupy:
@dbczumar Is this a fix for #3574 ?

mlflow/utils/gorilla.py
```diff
@@ -0,0 +1,898 @@
...
+            prev_patch = getattr(patch.destination, _ACTIVE_PATCH, None)
+            if not hasattr(patch.destination, original_name) or (
+                prev_patch
+                and prev_patch.destination != patch.destination
+                and issubclass(patch.destination, prev_patch.destination)
+            ):
+                setattr(patch.destination, original_name, target)
+
+    setattr(patch.destination, patch.name, patch.obj)
+    setattr(patch.destination, _ACTIVE_PATCH, patch)
```
dbczumar, line 331: @harupy These are the critical portions of the change to address problem (1) from https://github.com/mlflow/mlflow/issues/3574#issuecomment-714994583.

We now retain a reference to the previous patch that was applied to a class or one of its super classes, and, if  the patch was applied to a super class, we overwrite it with the patch that's being applied to this class directly.

mlflow/utils/gorilla.py
```diff
@@ -0,0 +1,898 @@
+#                     __ __ __
```
dbczumar, line 1: We inline gorilla to make the change described here: https://github.com/mlflow/mlflow/pull/3582/files#r512233266

mlflow/sklearn/__init__.py
```diff
@@ -724,16 +724,21 @@ def fetch_logged_data(run_id):
...
+        MLflow parameters, metrics, tags, and artifacts to a corresponding MLflow Run.
+        """
         should_start_run = mlflow.active_run() is None
         if should_start_run:
             try_mlflow_log(mlflow.start_run)
 
         _log_pretraining_metadata(self, *args, **kwargs)
 
-        original_fit = gorilla.get_original_attribute(self, func_name)
+        original_fit = gorilla.get_original_attribute(clazz, func_name)
```
dbczumar, line 739: This (together line 914) is the critical change to address problem (2) from https://github.com/mlflow/mlflow/issues/3574#issuecomment-714994583. 

Assume that the old line was being invoked from a patched `fit` call on a class called `Parent` and that `self` refers to an instance of class `Child` which inherits from `Parent`. Calling `get_original_attribute()` on `self` would yield the `Child.fit()` function, rather than `Parent.fit()`, leading to recursive behavior.

Now, instead of fetching the attribute from `self`, we fetch it from `clazz` (which would refer to `Parent` in the above example).

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1033,3 +1034,47 @@ def test_autolog_does_not_capture_runs_for_preprocessing_or_imputation_estimator
...
+    class ChildMod(ParentMod):
+        def fit(self, X, y):
+            super().fit(X, y)
+            self.prediction = self.prediction + 1
+
+    og_all_estimators = mlflow.sklearn.utils._all_estimators()
+    new_all_estimators = og_all_estimators + [("ParentMod", ParentMod), ("ChildMod", ChildMod)]
+
+    with mock.patch("mlflow.sklearn.utils._all_estimators", return_value=new_all_estimators):
+        mlflow.sklearn.autolog()
```
harupy, line 1072: nice trick :)

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1033,3 +1034,47 @@ def test_autolog_does_not_capture_runs_for_preprocessing_or_imputation_estimator
...
+    # expected constant prediction value, guarding against regressions of
+    # https://github.com/mlflow/mlflow/issues/3574 where ChildMod.fit() would either infinitely
+    # recurse or yield the incorrect prediction result set by ParentMod.fit()
+
+    class ParentMod(sklearn.base.BaseEstimator):
+        def get_params(self, deep=False):
+            return {}
+
+        def fit(self, X, y):
+            self.prediction = 7
```
harupy, line 1058:
```suggestion
            self.prediction = np.array([7])
```
@dbczumar 

This might be a nit but should we return a numpy array here so `infer_signature` doesn't the warning below?

```
Failed to infer model signature: Expected one of (pandas.DataFrame, numpy array,
dictionary of (name -> numpy.ndarray), pyspark.sql.DataFrame) but got '<class 'int'>'
```

harupy: The test actually passes whatever `predict` returns due to `try-catch` though.

dbczumar: Done! Was definitely worth addressing

harupy (approved):
LGTM 👍

PR #3597, opened 2020-10-26
mlflow deployments - Adding predict cli option and adding config argument for delete deployment

shrinath-suresh:

```
Commands:
  create     Deploy the model at ``model_uri`` to the specified target.
  delete     Delete the deployment with name given at `--name` from the...
  get        Print a detailed description of the deployment with name given...
  help       Display additional help for a specific deployment target, e.g.
  list       List the names of all model deployments in the specified...
  run-local  Deploy the model locally.
  update     Update the deployment with ID `deployment_id` in the specified...

```

Adding predict method to the above list, so that cli users can directly invoke `predict` method from command line. 

And current implementation of `delete_deployment` takes only name as an argument. Adding config parameter as an additional argument to it. 
Reason: When multiple version of the same model are registered into the system and user wants to delete a specific version of it, current implementation doesnt support it, as we have only name argument. 

https://pytorch.org/serve/management_api.html#unregister-a-model - In torchserve , there can be multiple models with same name and different version numbers. ex: `noop v1`, `noop v2`. And to specifically delete a version, we need an extra parameter. 

## How is this patch tested?

Tested with examples and added necessary Unit Tests.

Mlflow users can perform `predict` using mlflow deployments cli and users can delete a specific version of a deployed model using config parameter in delete deployment. 

Components 

Interface 

Language 
- [x] `language/r`: Python APIs and clients

Integrations


mlflow/deployments/base.py
```diff
@@ -135,12 +135,14 @@ def update_deployment(self, name, model_uri=None, flavor=None, config=None):
 
     @abc.abstractmethod
     @experimental
-    def delete_deployment(self, name):
+    def delete_deployment(self, name, config):
```
smurching, line 138: Thanks for the PR @shrinath-suresh - one of the principles of the deployment plugin is that a `name` should uniquely identify a deployment, which makes it easy to create, update, get, and delete a deployment using the originally specified name.

It feels a bit heavyweight to add a `config` parameter to the `delete_deployment` API (and presumably we'd also want it for `update_deployment` and `get_deployment` operations) to support specifying a single version in the TorchServe case; I also worry it'd be confusing to users that they can't specify a version via `config` when calling `create_deployment, as TorchServe seems to auto-assign the version of a new model when you hit the [model-registration endpoint](https://pytorch.org/serve/management_api.html#register-a-model)


Maybe for the TorchServe plugin, the `name` of a deployment could be either the original deployment name, or a concatenation of deployment name & version, like `model_name/version` if you want to update/get/delete a single version? That way we don't need to add a `config` parameter, and this behavior seems to align with the [TorchServe REST API](https://pytorch.org/serve/management_api.html#describe-model)

mlflow/deployments/cli.py
```diff
@@ -63,6 +65,20 @@ def _user_args_to_dict(user_list):
     "list of supported config options.",
 )
 
+parse_input = click.option(
```
smurching, line 68: Could we copy the API & parameter descriptions of `mlflow models predict`? There the options look like:

```
  -i, --input-path TEXT    CSV containing pandas DataFrame to predict against.
  -o, --output-path TEXT   File to output results to as json file. If not
                           provided, output to stdout.
```

mlflow/deployments/cli.py
```diff
@@ -220,3 +237,24 @@ def run_local(flavor, model_uri, target, name, config):
...
     interface.run_local(target, name, model_uri, flavor, config_dict)
+
+
+@commands.command("predict")
+@parse_custom_arguments
+@deployment_name
+@target_details
+@parse_input
+@parse_output
+def predict(target, name, input_path, output_path, config):
```
smurching, line 248:
```suggestion
def predict(target, name, input_path, output_path):
```

Can we remove the `config` argument as described in https://github.com/mlflow/mlflow/pull/3597/files#r513748823? Assuming it's only needed to specify a specific model version to score.

mlflow/deployments/cli.py
```diff
@@ -220,3 +237,24 @@ def run_local(flavor, model_uri, target, name, config):
...
+    """
+    Predict the results for the deployed model for the given input(s)
+    """
+    config_dict = _user_args_to_dict(config)
+    df = pd.read_json(input_path)
+    client = interface.get_deploy_client(target)
+    result = client.predict(name, df, config_dict)
+    click.echo("\n")
```
smurching, line 256: A few comments:
* Could we check here whether an output path is specified, and if so not `click.echo` the output?
* Also, when we print the output, could we just print the raw output, rather than wrapping it in newlines and a "RESULT IS" header? 
* We should make sure to print the output in JSON format

See the implementation of `mlflow models predict` [here](https://github.com/mlflow/mlflow/blob/00c7d0bdbb1d1f082abdb4107d2960ab26681381/mlflow/pyfunc/scoring_server/__init__.py#L234-L250) for reference

smurching (commented):
Thanks @shrinath-suresh! Leaving some comments

mlflow/deployments/base.py
chauhang: @smurching Version number has to be specified in TorchServe when deleting models (called [unregister](https://github.com/pytorch/serve/blob/master/docs/management_api.md#unregister-a-model) on TS side). It will be better to just have an optional version parameter to handle this.

smurching: @chauhang makes sense that we'd like the ability to specify a version number for TorchServe use cases, and agree that an explicit version parameter is nicer than passing a name like "myModel/3"

However, the version number seems like a TorchServe-specific concept (e.g. other deployment tools like Sagemaker & AzureML don't have a notion of an model version assigned at creation time), so it'd be nice to find a solution that doesn't change the underlying deployment plugin API, rather than add parameters (like `version=None`) that are meaningful only for certain deploy plugins and ignored by others.

My mental model is that when you [deploy to TorchServe the first time](https://pytorch.org/serve/management_api.html#register-a-model), you're actually creating two entities - the model itself, and v1 of the model. TorchServe then allows for most CRUD & inference operations against both the model and a specfic version of the model (you can [scale workers](https://pytorch.org/serve/management_api.html#scale-workers), [describe](https://pytorch.org/serve/management_api.html#describe-model), or [predict against](https://pytorch.org/serve/inference_api.html#predictions-api) a model using either the model name, or a model name + version). It seems [deletion](https://pytorch.org/serve/management_api.html#unregister-a-model) is an exception, in that there's no way to delete the model itself, only specific versions of the model (why is that the case btw?)

What about the following:
1) When you call `model_info = torchserve_deploy_client.create_deployment("my_model_name")`, the returned `model_info` dict contains a `version_name` key equal to `"<model_name>/<version_number>"`, e.g. "my_model_name/1"
2) Users can subsequently write code like the following to perform CRUD operations on a specific version
   ```
    version_name = model_info["version_name"]
    torchserve_deploy_client.get_deployment(version_name)
    torchserve_deploy_client.predict(version_name, pandas_df)
    ```
3) Or use the CLI:
   ```
    mlflow deployments get -t torchserve -n <version_name>
    mlflow deployments delete -t torchserve -n <version_name>
    ```
4) Deleting a model without specifying a version deletes all model-versions associated with the model.

Another approach would be to add an optional version parameter on the `delete_deployment` method of the TorchServe deploy client itself, but not as part of the core plugin API - this could also work, but wouldn't allow for e.g.deleting an individual model version via the CLI

shrinath-suresh:
> Thanks @shrinath-suresh! Leaving some comments

@smurching Thanks for the review comments. I will update the new version by tomorrow my time.

mlflow/deployments/base.py
shrinath-suresh: Thanks for the detailed description. I read through your comments and went through the current deployment plugin code. Following are my observations if we have separate parameters for `name` and `version`. 

**1. create_deployment**

**Current Behaviour:**

Accepted config keys

1. `version` [optional]- if version is None, `max(version) + 1` will be set . For fresh deployments, version number is set to `1.0`
2. `model_file` [mandatory] - path to model file
3. `handler_file` [mandatory] - path to handler file
4. `extra_files` [optional] - path to extra files 
5. `requirements` [optional] - path to requirements.txt

Based on the above mentioned parameters, mar file will be generated and deployed using management api. 

returns - `name` and `flavour` on successful deployment. 

**New Requirement:**
Add `version` to the return list so that user can use it in CRUD operation. 

config key is already available in `create_deployment`. Works both for python and cli users. 


**2. update_deployment**

**Current Behaviour:**

Sample Input: `config = {"min_worker" : 3}`. 
Plugin constructs the api as - `http://localhost:8081/models/noop?min_worker=3`

Its upto the user to pass any key:value mapping to config parameter in update - depending on the requirement. 

**New Requirement:** 

Add support for setting default version. 

sample url for setting default api - `http://localhost:8081/models/noop/2.0/set-default`.

Use the existing config variable. If user passes, `{'default_version' : True}` - append `set-default` to query string.

config key is already available in `update_deployment`. Works both for python and cli users. 


**3. list_deployment**

Lists all the model registered in torchserve and supports pagination as well. No change in behaviour for python and cli users. 

**4. get_deployment**

**Current Behaviour:**

Given a deployment name, this method returns all the versions of the model.

For example: `mnist_test` - name and if user had registered 3 different versions of model. 

This method derives all 3 version and returns it as a list. 

**New Requirement:** 

Adding support for version based get_deployment.

As per current approach - `curl http://localhost:8081/models/noop/all` is called to derive all different versions of `noop` model.

`curl http://localhost:8081/models/noop/2.0` is supported in torchserve- if we need to query a particular version, we need to add a `version` parameter to this method. 

Important point to note: by adding version number, only the python plugin can perform the delete deployment based on version. cli still invokes the delete without version number. 

Hence, python users can query based on version and for cli users all the versions of same model would be returned. 


**5. delete_deployment**

**Current Behaviour:**

Takes `name` and `config` as input. 

version number has to be specified in the config parameter . ex: `config["version"] = "3.0"`

if config is none - By default it attempts to delete the first version of model ("1.0")

**New Requirement:** 

change `config` param to `version`. 

if version number is provided, delete deployment should delete the corresponding version

if version number is None, find the max version of the deployment and delete.

Python users can delete based on the version number and cli users can delete last deployed model (max_version). 


**6. predict**

**Current Behaviour:**

Takes `df`, `name` and `config` as input.

if the version number is provided in config, prediction is made based on the model version else prediction is done based on version "1.0".


**New requirement:**

change `config` parameter to `version` in deployment plugin

For python users prediction can be made based on version number and for cli user prediction can be made based on version 1 or default version (to be decided)

mlflow/deployments/cli.py
```diff
@@ -63,6 +68,16 @@ def _user_args_to_dict(user_list):
...
 )
 
+parse_input = click.option(
+    "--input-path", "-I", required=True, help="Path to input json file for prediction",
+)
+
+parse_output = click.option(
+    "--output-path",
+    "-O",
+    help="File to output results to as json file. If not" "provided, output to stdout.",
```
smurching, line 78:
```suggestion
    help="File to output results to as a JSON file. If not provided, prints output to stdout.",
```

tests/deployments/test_cli.py
```diff
@@ -75,6 +76,19 @@ def test_get():
     assert "key2: val2" in res.stdout
 
 
+def test_predict():
```
smurching, line 79: Could we use the pytest `tmpdir` fixture here instead? e.g. something like
```
def test_predict(tmpdir):
    temp_input_file_path = tmpdir.join("input.json").strpath
    with open("temp_input_file_path", "w+t") as temp_input_file:
        temp_input_file.write('{"data": [5000]}')
        temp_input_file.seek(0)
        runner = CliRunner()
        res = runner.invoke(
            cli.predict, ["--target", f_target, "--name", f_name, "--input-path", temp_input_file]
        )
        assert "1" in res.stdout
```

I think the current test will also happen to delete `temp_input_file` if the test fails & Python process test exits before calling `temp_input_file.close()`, because the OS tries to close open files when the Python process exits, but would be nice to use the built-in pytest utility for temporary files, which guarantees that tempfiles are removed at the end of the test session

smurching (commented):
One comment otherwise LGTM!

tests/deployments/test_cli.py
shrinath-suresh: Done. I have changed it to use `tmpdir`

mlflow/deployments/cli.py
shrinath-suresh: Done. Copied the parameters and descriptions from `mlflow models`

mlflow/deployments/cli.py
shrinath-suresh: Done. Addressed the above comments.

shrinath-suresh:
@smurching I have adddressed all the review comments and the python tests are passed. Failure in `r build` and `tensorflow` build looks like an upstream issue. I have pulled the latest change from upstream as well. Still facing the issue. Can you please help?

harupy:
#3616 should fix the r, tensorflow, and flavors checks.

harupy:
@shrinath-suresh #3616 has been merged.

shrinath-suresh:
> @shrinath-suresh #3616 has been merged.

Thanks I have pulled the change from master. Waiting for build to pass.

shrinath-suresh:
> > @shrinath-suresh #3616 has been merged.
> 
> Thanks I have pulled the change from master. Waiting for build to pass.

@harupy Once again the `r` build failed

smurching (approved):
LGTM

shrinath-suresh:
@smurching @harupy Thank you very much. Can we please merge this PR, i dont have write access to this repo.

PR #3601, opened 2020-10-27
Autolog with PyTorch using lightning

shrinath-suresh:

Adding autolog feature for pytorch lightning. This feature logs the parameters, metrics, artifacts and model into mlflow by just adding a `autolog()` function call into the pytorch lightning code.

Following are the features added as part of autolog. 

1. Default parameters related to optimizer and trainer. 
2. Metrics such as loss and accuracy are logged
3. Model Summary is saved as an artifact
4. Entire model is stored into mlflow using `mlflow.pytorch.log_model`
5. Additional support to Early Stopping parameters and Metrics
6. Model checkpoints - Storing best model

## How is this patch tested?

Added Mnist Hadwritten digit recognition and Unit test for autolog

Pytorch lightning users can log metrics, artifacts and model into mlflow easily by adding an `autolog` function call. 

Components 

Interface 

Language 
- [x] `language/r`: Python APIs and clients

Integrations


examples/pytorch/BertNewsClassification/MLproject
```diff
@@ -0,0 +1,22 @@
...
+
+entry_points:
+  main:
+    parameters:
+      epochs: {type: int, default: 5}
+      gpus: {type: int, default: 0}
+      batch_size: {type: int, default: 64}
+      num_workers: {type: int, default: 3}
+      learning_rate: {type: float, default: 1e-3}
+      tracking_uri: {type: str, default: 'http://localhost:5000'}
```
smurching, line 13: Thanks @shrinath-suresh for the PR! Could we split out the examples into a separate PR? Also - I don't think we want to encourage passing the tracking URI as a parameter / defining a default value for the tracking URI within an MLproject file. The reason is that MLflow has its own default tracking URI, so by adding another tracking URI to the project we end up either duplicating the MLflow default or deviating from it (which may be confusing to users). We should instead encourage users to specify a custom tracking URI by setting the `MLFLOW_TRACKING_URI` environment variable

examples/pytorch/BertNewsClassification/README.md
```diff
@@ -0,0 +1,75 @@
+## Bert Sentiment Analysis example with MLFlow
+
+In this example, we have used a pretrained Bert model to classify the news reports into any one of the 4 different categories("world", "Sports", "Business", "Sci/Tech").
+The autolog code uses Pytorch Lightning's MLFlowLogger to log metrics.
+The code is trained using pytorch lightning loop and the autolog function call in the main - `autolog()`
+is responsible for logging the params, metrics, model summary and the model.
+ 
+
+### Code related to MLflow:
+* [`mlflow.pytorch.pytorch_autolog`]
```
smurching, line 10:
```suggestion
* [`mlflow.pytorch.pl_autolog`]
```

Are we sure about `mlflow.pytorch.pytorch_autolog` as the API name? If we were to follow the pattern for other frameworks, we'd do `mlflow.pytorch_lightning.autolog`

examples/pytorch/BertNewsClassification/README.md
```diff
@@ -0,0 +1,75 @@
...
+This is an experimental api that logs ML model artifacts and metrics.
+The metrics are logged during training of the model.
+
+## Setup
+
+Mlflow tracking UI is set to `http://localhost:5000` with experiment name as `Default` in the script - `bert_classification.py`
```
smurching, line 16: Note that one advantage of not having tracking URI as a parameter to the project / just relying on the MLflow default URI is that users don't need to launch a tracking server at http://localhost:5000/ via  ``mlflow server`` in order to run the example script

shrinath-suresh:
@smurching Moved examples to https://github.com/mlflow/mlflow/pull/3608

examples/pytorch/BertNewsClassification/README.md
shrinath-suresh: `mlflow.pytorch_lightning.autolog` looks better. I will rename it :-)

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+
+        def on_train_end(self, trainer, pl_module):
+            """
+            Logs the model checkpoint into mlflow - models folder on the training end
+
+            :param trainer: pytorch lightning model instance
+            :param pl_module: pytorch lightning base module
+            """
+
+            mlflow.pytorch.log_model(pytorch_model=trainer.model, artifact_path="models")
```
harupy, line 104:
```suggestion
            mlflow.pytorch.log_model(pytorch_model=trainer.model, artifact_path="model")
```

We usually use "model" when logging a model in auto-logging.

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+                    self._early_stop_check(callback)
+
+        def on_train_start(self, trainer, pl_module):
+            """
+            Logs Optimizer related metrics when the train begins
+
+            :param trainer: pytorch lightning model instance
+            :param pl_module: pytorch lightning base module
+            """
+            mlflow.set_tag(key="Mode", value="training")
```
harupy, line 65: `try_mlflow_log` should be used here.

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+                for optimizer in trainer.optimizers:
+                    try_mlflow_log(mlflow.log_param, "optimizer_name", type(optimizer).__name__)
+
+                    if hasattr(optimizer, "defaults"):
+                        defaults_dict = optimizer.defaults
+                        if "lr" in defaults_dict:
+                            try_mlflow_log(mlflow.log_param, "learning_rate", defaults_dict["lr"])
+
+                        if "eps" in defaults_dict:
+                            try_mlflow_log(mlflow.log_param, "epsilon", defaults_dict["eps"])
```
harupy, line 83: What does `epsilon` represents?

examples/pytorch/BertNewsClassification/README.md
smurching: @shrinath-suresh thanks! Actually sorry for the flip-flop on this - since Pytorch lightning autologging does in fact call `mlflow.pytorch.log_model` (and therefore generates the `pytorch` flavor / saves models that can be loaded back via `mlflow.pytorch.load_model`), we're thinking we should actually just call the API `mlflow.pytorch.autolog`, and mention in the docstring that only pytorch-lightning is supported for now

smurching: Then the user workflow would be:

```
import mlflow.pytorch
mlflow.pytorch.autolog()

# Define model using pytorch lightning
class LitAutoEncoder(pl.LightningModule):
   ...

model = LitAutoEncoder(...)
# Fit model etc
...

# Later, load back model using mlflow.pytorch.load_model
autologged_pl_model = mlflow.pytorch.load_mode("/path/to/logged/model")
```

harupy: @smurching Should we log a model as a scripted model so we can serve it using the torchserve plugin?

https://pytorch-lightning.readthedocs.io/en/latest/production_inference.html#exporting-to-torchscript

dev/extra-ml-requirements.txt
```diff
@@ -18,6 +18,8 @@ tensorflow==1.15.2
 # Required by mlflow.pytorch
 torch==1.6.0
 torchvision==0.5.0
+# Required by mlflow.pytorch
+pytorch_lightning==1.0.2
```
harupy, line 22: Can we unpin `pytorch_lightning `?

mlflow/pytorch/pytorch_autolog.py
harupy, line 83: Why not log all parameters stored in `defaults_dict`?

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -0,0 +1,168 @@
...
+@pytest.fixture
+def pytorch_model_initialized():
+    trainer = pl.Trainer(max_epochs=NUM_EPOCHS)
+    model = IrisClassification()
+    trainer.fit(model)
+    return trainer
+
+
+def test_pytorch_autolog_persists_manually_created_run(pytorch_model_initialized):
+    _ = pytorch_model_initialized
```
harupy, line 65:
```suggestion
    _ = pytorch_model_initialized
```

should we remove this line?

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -0,0 +1,168 @@
...
+    autolog()
+    with mlflow.start_run() as manual_run:
+        assert mlflow.active_run()
```
harupy, line 68:
```suggestion
        assert mlflow.active_run() is not None
```

to clarify what we're asserting here?

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -0,0 +1,168 @@
...
+        assert mlflow.active_run().info.run_id == manual_run.info.run_id
+
+
+def test_pytorch_autolog_ends_auto_created_run(pytorch_model_initialized):
+    _ = pytorch_model_initialized
+    autolog()
+    assert mlflow.active_run() is None
```
harupy, line 75:
```suggestion
    autolog()
    # insert code to train a model
    assert mlflow.active_run() is None
```

What we need to verify in this test is the patched`pl.Trainer.fit` terminates the auto-created run.

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+
+        def __init__(self):
+            self.early_stopping = False
+
+        def on_epoch_end(self, trainer, pl_module):
+            """
+            Log loss and other metrics values after each epoch
+
+            :param trainer: pytorch lightning model instance
```
harupy, line 45:
```suggestion
            :param trainer: pytorch lightning trainer instance
```
?

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+            :param pl_module: pytorch lightning base module
+            """
+            if (pl_module.current_epoch + 1) % every_n_iter == 0:
+                for key, value in trainer.callback_metrics.items():
+                    try_mlflow_log(
+                        mlflow.log_metric, key, float(value + 1), step=pl_module.current_epoch
```
harupy, line 51: Does `value` represent a metric value here?

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+                    self._log_early_stop_params(callback)
+
+            if hasattr(trainer, "optimizers"):
+                for optimizer in trainer.optimizers:
+                    try_mlflow_log(mlflow.log_param, "optimizer_name", type(optimizer).__name__)
+
+                    if hasattr(optimizer, "defaults"):
+                        defaults_dict = optimizer.defaults
+                        if "lr" in defaults_dict:
+                            try_mlflow_log(mlflow.log_param, "learning_rate", defaults_dict["lr"])
```
harupy, line 80: Do we need to prefix `learning_rate` with the optimizer name here?

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+            """
+            Logs all early stopping metrics
+
+            :param early_stop_callback: Early stopping callback object
+            """
+            if early_stop_callback.stopped_epoch != 0:
+
+                if hasattr(early_stop_callback, "stopped_epoch"):
+                    try_mlflow_log(
+                        mlflow.log_metric, "Stopped_Epoch", early_stop_callback.stopped_epoch
```
harupy, line 157: Why capitalize the metrice key here?

mlflow/pytorch/pytorch_autolog.py
```diff
@@ -0,0 +1,197 @@
...
+    @gorilla.patch(pl.Trainer)
+    def fit(self, *args, **kwargs):
+        """
+        Patching trainer.fit method to add autolog class into callback
+        """
+        original = gorilla.get_original_attribute(pl.Trainer, "fit")
+        return _run_and_log_function(self, original, args, kwargs)
+
+    settings = gorilla.Settings(allow_hit=True, store_hit=True)
+    gorilla.apply(gorilla.Patch(pl.Trainer, "fit", fit, settings=settings))
```
harupy, line 197:
```suggestion
    wrap_patch(pl.Trainer, "fit", fit)
```
Can you use the `wrap_patch` function here to prevesr the original arributes (e.g. `__doc__`, `__name__`)?

harupy: How wrap_patch works

https://github.com/mlflow/mlflow/blob/44b152a87e07fce49af3e3a262f7b3caf266bed8/mlflow/utils/autologging_utils.py#L103-L111

examples/pytorch/BertNewsClassification/README.md
chauhang: @harupy Scripting is optional path and not all models will be torchscripted. We support both eager mode and torchscripted mode so it will be upto the user which option then want to go with based on their specific use case and needs. TorchScripting is recommended for deploying in production, however not all models are scriptable and some hardwares require conversion of models to ONNX/TVM etc so we should not be forcing users down a path.

examples/pytorch/BertNewsClassification/README.md
chauhang: @smurching The biggest benefit of MLflow esp in a team environment is for people to be able share their models / experiment runs. So using a common tracking server that has been launched separately is going to be an important scenario. For user's that are just testing on their local machines it is ok to launch separately each time. If MLFLOW_TRACKING_URI environment variable is the common way people use MLflow shared tracking server link, then we can make this change. Current document pages do not clearly describe this part

dev/extra-ml-requirements.txt
chauhang: @harupy It needs to be ver 1.0.2 and above. The PTL apis have changed and there were break issues with the prior versions. The current PTL api is expected to be more stable going forward.

mlflow/pytorch/pytorch_autolog.py
chauhang: @harupy The default names were cryptic. So changed to full name for better readability and ease of use during metrics comparison in the MLflow UI

dev/extra-ml-requirements.txt
harupy:
```suggestion
pytorch_lightning>=1.0.2
```

@chauhang Thanks! If so, should we un-max-pin the version?

mlflow/pytorch/pytorch_autolog.py
harupy: btw do we really need to log what phase the model is currently in?

@smurching What do you think?

examples/pytorch/BertNewsClassification/MLproject
shrinath-suresh: done

examples/pytorch/BertNewsClassification/README.md
shrinath-suresh: will address this in https://github.com/mlflow/mlflow/pull/3608

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: i changed it to `model`

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: > I think `mlflow.set_tag` also should be wrapped by `try_mlflow_log`.

updated the code by using `try_mlflow_log`

dev/extra-ml-requirements.txt
shrinath-suresh: > @chauhang Thanks! If so, should we un-max-pin the version?

updated it to `>=`

tests/pytorch/test_pytorch_autolog.py
shrinath-suresh: pylint throws up the unused argument warning without this line. I removed the line and added module level disable for it.

tests/pytorch/test_pytorch_autolog.py
shrinath-suresh: done

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: Done. Updated the doc string.

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: removed the increment.

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: updated the key

tests/pytorch/test_pytorch_autolog.py
shrinath-suresh: used the existing fixture. Updated the tests

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: For example: By running iris example following are the default params 

`{'lr': 0.02, 'betas': (0.9, 0.999), 'eps': 1e-08, 'weight_decay': 0, 'amsgrad': False}` . The keys are too short and hard to understand.

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: sure i will take a look into this. In the current behaviour, the patched `fit` method  is responsible for starting and ending the mlflow run using `_run_and_log_function`. if we are going to use wrap_patch how do we pass the arguments for `_run_and_log_function` (since wrap_patch doesnt return anything)?

examples/pytorch/BertNewsClassification/README.md
shrinath-suresh: > @shrinath-suresh thanks! Actually sorry for the flip-flop on this - since Pytorch lightning autologging does in fact call `mlflow.pytorch.log_model` (and therefore generates the `pytorch` flavor / saves models that can be loaded back via `mlflow.pytorch.load_model`), we're thinking we should actually just call the API `mlflow.pytorch.autolog`, and mention in the docstring that only pytorch-lightning is supported for now

@smurching @harupy 
Tried moving autolog method to root directory as `pytorch.py`. However, we already have a folder named `pytorch` under root directory and all our log/save model codes are residing in it. When i try to do `mlflow.pytorch.autolog()` it looks into `pytorch` folder and fails. 

In case of keras and fastai, the folders are not present. The entire autolog code is placed as `keras.py` and `fastai.py` so that it can be directly imported as `mlflow.keras.autolog`. 

Technically, the import should be mlflow.pytorch.<FILENAME>.autolog. Any thoughts ?

smurching: @shrinath-suresh good question, could we do something similar to what we do for Spark autologging ([link](https://github.com/mlflow/mlflow/blob/64c93c81167b8a132bf00535fc23f3420c9caf7f/mlflow/spark.py#L630-L688))? Namely, define the actual autologging implementation in an internal `_autolog` method under a module like `mlflow.pytorch._pytorch_autolog`, then import and expose it as `def autolog` in `mlflow/pytorch/__init__.py`?

mlflow/pytorch/pytorch_autolog.py
harupy: > how do we pass the arguments for _run_and_log_function?

`_run_and_log_function` is called with all the required parameters in `fit`.

examples/pytorch/BertNewsClassification/README.md
shrinath-suresh: @smurching  Added the wrapper method in `__init__.py` and i am able to use it as `mlflow.pytorch.autolog()`. Thanks for the suggestion and link.

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: @harupy Thank you very much. i have updated the code to use `wrap_patch`

mlflow/pytorch/__init__.py
```diff
@@ -571,3 +571,15 @@ def predict(self, data, device="cpu"):
...
+
+
+def autolog(log_every_n_iter=1):
+    """
+    Wrapper for `mlflow.pytorch._pytorch_autolog.autolog` method
+
+    :param log_every_n_iter: parameter to log metrics once in `n` iteration. By default, metrics
+                       are logged after every iteration.
+    """
+    from mlflow.pytorch._pytorch_autolog import _autolog
```
shrinath-suresh, line 583: Followed same convention as in spark library.

mlflow/pytorch/__init__.py
```diff
@@ -571,3 +571,16 @@ def predict(self, data, device="cpu"):
             predicted = pd.DataFrame(preds.numpy())
             predicted.index = data.index
             return predicted
+
+
+def autolog(log_every_n_iter=1):
+    """
+    Wrapper for `mlflow.pytorch._pytorch_autolog.autolog` method.
+    Only PyTorch lightning is supported for this autolog feature.
```
shrinath-suresh, line 579: @smurching Mentioned about the support here. Please check if this docstring looks fine or we need more detailed description on this.

mlflow/pytorch/pytorch_autolog.py
shrinath-suresh: Added `weight_decay` and `betas` parameters

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -0,0 +1,206 @@
...
+                        if "eps" in defaults_dict:
+                            try_mlflow_log(mlflow.log_param, "epsilon", defaults_dict["eps"])
+
+                        if "betas" in defaults_dict:
+                            try_mlflow_log(mlflow.log_param, "betas", defaults_dict["betas"])
+
+                        if "weight_decay" in defaults_dict:
+                            try_mlflow_log(
+                                mlflow.log_param, "weight_decay", defaults_dict["weight_decay"]
+                            )
```
harupy, line 93: @shrinath-suresh cc @smurching 

Not sure if we need to address this in this PR, but we should consider how to handle the case where multiplte optimizers are given.

mlflow/pytorch/pytorch_autolog.py
harupy: > The keys are too short and hard to understand.

makes sense

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -0,0 +1,158 @@
...
+    artifacts = client.list_artifacts(run.info.run_id)
+    artifacts = map(lambda x: x.path, artifacts)
+    assert "model_summary.txt" in artifacts
+
+
+# pylint: disable=unused-argument
+def test_pytorch_autolog_persists_manually_created_run(pytorch_model):
+    with mlflow.start_run() as manual_run:
+        assert mlflow.active_run() is not None
+        assert mlflow.active_run().info.run_id == manual_run.info.run_id
```
harupy, line 61: This test should look like:

```python
    mlflow.pytorch.autolog()

        # insert code to train a model

        assert mlflow.active_run()
        assert mlflow.active_run().info.run_id == run.info.
```

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -0,0 +1,158 @@
...
+def test_pytorch_autolog_logs_expected_data(pytorch_model):
+    _, run = pytorch_model
+    data = run.data
+
+    # Checking if metrics are logged
+    assert "loss" in data.metrics
+    assert "val_loss" in data.metrics
+
+    # Testing unwanted parameters are not logged
+    assert "callbacks" not in data.params
```
harupy, line 44: This is a nit, but just because `assert "callbacks" not in data.params` passes doesn't necessarily mean `data.params` doesn't contain unwanted parameters (it might contains a parameter named `"foo"`).

tests/pytorch/test_pytorch_autolog.py
shrinath-suresh: makes sense. Ideally, the run should start first and then the autolog method should be called. will change it.

shrinath-suresh: The intention of the test case is to check if autolog persiste manually created run. I feel the code should be like this 

```

        mlflow.pytorch.autolog()
        # insert code to train a model

        assert mlflow.active_run()
        assert mlflow.active_run().info.run_id == run.info.run_id

```

User creates a run and then calls autolog method. Autolog should check for active run and since the run is already active, it should persist the same run (should  not create a new one). Please correct me if my understanding is wrong.

mlflow/pytorch/__init__.py
```diff
@@ -571,3 +571,16 @@ def predict(self, data, device="cpu"):
...
+
+    :param log_every_n_iter: parameter to log metrics once in `n` iteration. By default, metrics
```
smurching, line 581: Should we rename this to `log_every_n_epoch` for clarity? I know @andrewnitu was in the progress of reworking some of our per-epoch logging parameters, I'll check with him Monday for his recommendation on how this param should be named / behave

mlflow/pytorch/__init__.py
smurching: Thanks @shrinath-suresh! For this PR, how about:

```suggestion
    Automatically log metrics, params, and models from `PyTorch Lightning <https://pytorch-lightning.readthedocs.io/en/latest>`_ model training.
    **Note**: Autologging is only supported for PyTorch Lightning models, i.e. models that subclass `pytorch_lightning.LightningModule <https://pytorch-lightning.readthedocs.io/en/latest/lightning_module.html>`_. In particular, autologging support for vanilla Pytorch models that only subclass `torch.nn.Module <https://pytorch.org/docs/stable/generated/torch.nn.Module.html>`_ is not yet available.
```

Either in this PR or in a follow-up, we should enhance this docstring to describe when autologging is invoked & what exactly gets autologged. See the [sklearn autologging docstring](https://github.com/mlflow/mlflow/blob/37496c3161a9f7561b4c7ca3054ae6bdb3154523/mlflow/sklearn/__init__.py#L537-L703) for an example. We should also make sure to sure to mention Pytorch autologging in the "automatic logging" section within the MLflow Tracking API docs: https://github.com/mlflow/mlflow/blob/master/docs/source/tracking.rst#id70

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -0,0 +1,208 @@
...
+from pytorch_lightning.utilities import rank_zero_only
+from mlflow.utils.autologging_utils import try_mlflow_log, wrap_patch
+
+logging.basicConfig(level=logging.ERROR)
+
+every_n_iter = 1
+
+
+# autolog module uses `try_mlflow_log` - mlflow utility to log param/metrics/artifacts into mlflow
+# Eventhough the same can be achieved using
```
smurching, line 18: @shrinath-suresh thanks for documenting this - should we file a feature request against https://github.com/PyTorchLightning/pytorch-lightning/issues / consider whether we can contribute the improvements in your callback back to Pytorch Lightning's MlflowLogger? It seems long-term it'd be ideal if Pytorch Lightning's MlflowLogger could be used for autologging

smurching: And then I think we could replace this comment with something like

```suggestion
# TODO: Replace __MlflowPLCallback with Pytorch Lightning's built-in MlflowLogger once the issue in <link-to-issue> have been addressed
```

mlflow/pytorch/_pytorch_autolog.py
smurching: @harupy yeah good catch, currently it seems possible to log e.g. the learning rate from one optimizer and the epsilon value from another optimizer, depending on which optimizers contain which fields. @shrinath-suresh would it make sense to suffix the parameter & metric names with the optimizer name in the case of more than one optimizer? e.g. something like `learning_rate_adam_optimizer`?

harupy: or prefix with a class name (e.g. `Adam.learning_rate`)?

Screen Shot 2020-11-02 at 12 29 21

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -0,0 +1,208 @@
...
+            """
+
+            mlflow.pytorch.log_model(pytorch_model=trainer.model, artifact_path="model")
+
+            if self.early_stopping and trainer.checkpoint_callback.best_model_path:
+                try_mlflow_log(
+                    mlflow.log_artifact,
+                    local_path=trainer.checkpoint_callback.best_model_path,
+                    artifact_path="restored_model_checkpoint",
+                )
```
harupy, line 121: Should we log the checkpoint in the `"model"` directory?

mlflow/pytorch/_pytorch_autolog.py
harupy: But this approach doesn't work when the users uses two Adam optimizers (not sure how realistic it is)...

shrinath-suresh:
@karthik-77 is working on adding detailed documentation for autolog - https://github.com/chauhang/mlflow/pull/31 . We can track documentation changes in this PR and bring it to upstream once the reviews are completed.

mlflow/pytorch/_pytorch_autolog.py
shrinath-suresh: checkpoints are saved as `.ckpt`. I guess it would be confusing to have both `pytorch` and `pytorch_lightning` model under same folder `model`.

chauhang: @harupy This will get confusing for user's. The Checkpoints reside in their own separate folder

harupy: Makes. Let's just leave this as it is.

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -0,0 +1,221 @@
...
+import pytorch_lightning as pl
+import shutil
+import tempfile
+from pytorch_lightning.core.memory import ModelSummary
+from pytorch_lightning.utilities import rank_zero_only
+from mlflow.utils.autologging_utils import try_mlflow_log, wrap_patch
+
+logging.basicConfig(level=logging.ERROR)
+
+every_n_epoch = 1
```
harupy, line 14: I found the tensorflow does the same thing (defining a constant and use it with `global` in the autolog function). @smurching Do you happen to know why we need to do this?

https://github.com/mlflow/mlflow/blob/master/mlflow/tensorflow.py#L50

harupy: ah I got it. `_LOG_EVERY_N_STEPS` is used in `_log_event`.

https://github.com/mlflow/mlflow/blob/master/mlflow/tensorflow.py#L601

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -0,0 +1,221 @@
...
+# autolog module uses `try_mlflow_log` - mlflow utility to log param/metrics/artifacts into mlflow
+# MlflowLogger(Pytorch Lightning's inbuild class),
+# following are the downsides in using MlflowLogger.
+# 1. MlflowLogger doesn't provide a mechanism to store an entire model into mlflow.
+#    Only model checkpoint is saved.
+# 2. For storing the model into mlflow `mlflow.pytorch` library is used
+# and the library expects `mlflow` object to be instantiated.
+# In case of MlflowLogger, Run management is completely controlled by the class and
+# hence mlflow object needs to be reinstantiated by setting
+# tracking uri, experiment_id and run_id which may lead to a race condition.
```
smurching, line 26: @shrinath-suresh could we remove this comment in favor of filing an issue against PytorchLightning & linking that? I worry that we may lose track of these details / not take action otherwise :)

mlflow/pytorch/_pytorch_autolog.py
smurching: @harupy including the class name in the metric/param name SGTM. I'd actually suggested suffixing instead of prefixing because the type of metric (learning rate, weight decay, etc) seemed more important than the optimizer it comes from, and we truncate metric names e.g. in the experiment runs table UI, as in this screenshot:

image

That is, it'd be nice to see a truncated metric name like "learning_rate..." instead of "Adam.lear..."

mlflow/pytorch/__init__.py
```diff
@@ -571,3 +571,26 @@ def predict(self, data, device="cpu"):
             predicted = pd.DataFrame(preds.numpy())
             predicted.index = data.index
             return predicted
+
+
+def autolog(log_every_n_epoch=1):
```
smurching, line 576: Can we mark this API as experimental initially to give ourselves flexibility to change it in the future based on user feedback? You should be able to do so by adding (to the top of the file):

```
from mlflow.utils.annotations import experimental
```

And then annotating this method with `@experimental`

mlflow/pytorch/_pytorch_autolog.py
smurching: From offline discussion -  if we want to accelerate getting this PR in, we could simplify the metric/param logging logic to only log metrics/params of the first optimizer, then in a follow-up add support for multiple optimizers - I'm guessing it's not common to have multiple optimizers in general

smurching (commented):
LGTM after we:
* Address the issue around param logging for multiple optimizers (at least initially I think we can simplify things by just logging params of the first optimizer, and later add support for multiple optimizers)
* Mark the new API as experimental

mlflow/pytorch/_pytorch_autolog.py
shrinath-suresh: @smurching  https://github.com/mlflow/mlflow/pull/3601/commits/f3f118b45123d9daa1125fed8c88b711f384a5fa - we have already addressed this as part of this commit. We loop through the optimizers and save the parameters as `{param}_{optimizer_name}`. For Ex: `learning_rate_adam_optimizer`

mlflow/pytorch/__init__.py
shrinath-suresh: i have marked it as experimental.

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -29,7 +30,20 @@
 
 
 @rank_zero_only
+@experimental
 def _autolog(log_every_n_epoch=1):
+    """
+    Enable automatic logging from pytorch to MLflow.
```
shrinath-suresh, line 36: used the doc string from [fastai autolog](https://github.com/mlflow/mlflow/blob/89fb38ec122a1c489488ab1fb0d35fe70c2be649/mlflow/fastai.py#L282) method with minor changes.

shrinath-suresh:
> LGTM after we:
> 
> * Address the issue around param logging for multiple optimizers (at least initially I think we can simplify things by just logging params of the first optimizer, and later add support for multiple optimizers)
> * Mark the new API as experimental

Thanks. We have addressed the above mentioned items. Please review and let us know your comments.

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -0,0 +1,160 @@
...
+
+@pytest.fixture
+def pytorch_model_with_callback(patience):
+    mlflow.pytorch.autolog()
+    model = IrisClassification()
+    early_stopping = EarlyStopping(monitor="val_loss", mode="min", patience=patience, verbose=True)
+
+    checkpoint_callback = ModelCheckpoint(
+        filepath=os.getcwd(), save_top_k=1, verbose=True, monitor="val_loss", mode="min", prefix=""
+    )
```
harupy, line 78: Can we save the checkpoints in a temporary directory?

harupy: and can we add `lightning_logs` (generated during the tests) to `.gitignore`?

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -0,0 +1,235 @@
...
+
+            for callback in trainer.callbacks:
+                if isinstance(callback, pl.callbacks.early_stopping.EarlyStopping):
+                    self.early_stopping = True
+                    self._log_early_stop_params(callback)
+
+            if hasattr(trainer, "optimizers"):
+                for optimizer in trainer.optimizers:
+                    try_mlflow_log(mlflow.log_param, "optimizer_name", type(optimizer).__name__)
+                    optimizer_name = type(optimizer).__name__.lower() + "_optimizer"
```
harupy, line 93: The current approach doesn't work when we use two Adam optimizers. We should address that in follow-ups.

tests/pytorch/test_pytorch_autolog.py
shrinath-suresh: @harupy I will create a follow up PR for autolog and example and i will address all the review comments in the follow up PRs.

javierluraschi:
Removed the `language/r` label since no R changes are being proposed here.

mlflow/pytorch/__init__.py
chauhang: @smurching This is being handled in the Doc PR separately

smurching (approved):
LGTM to get this in incrementally to allow building examples on top of it - per offline discussion, in a follow-up can we:

* Either (a) only add the optimizer name suffix to parameter names (as is done [here](https://github.com/mlflow/mlflow/pull/3601/files#diff-106da39ab042940a984680f627203b29b44a7206059bcd57d5e2ed31816ecbcbR90-R121)) if there's >1 optimizer, or (b) simplify the logic to only log params of the first optimizer. Unconditionally adding the optimizer name is probably a bit of overkill / will complicate comparing params across runs in the common case where you just have a single optimizer per model
* Address @harupy’s comments around putting lightning_logs in .gitignore and save model checkpoints to a temporary directory during tests

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -0,0 +1,235 @@
...
+            """
+
+            mlflow.pytorch.log_model(pytorch_model=trainer.model, artifact_path="model")
+
+            if self.early_stopping and trainer.checkpoint_callback.best_model_path:
+                try_mlflow_log(
+                    mlflow.log_artifact,
+                    local_path=trainer.checkpoint_callback.best_model_path,
+                    artifact_path="restored_model_checkpoint",
+                )
```
harupy, line 148: Hi @shrinath-suresh,

When training is early stopped, does `trainer` restore the best weights?

harupy: Looks like it doesn't:

https://github.com/PyTorchLightning/pytorch-lightning/blob/4018237c309b7d9d6978da73132003615341e04a/pytorch_lightning/trainer/training_loop.py#L591

SeanNaren: Not entirely sure what the question is, when does the trainer restore the model's weights?

harupy: @SeanNaren Thanks for the reply.

```md
Early Stopping patience: 3

| Epoch | Validation loss | comment                |
| :---- | :-------------- | ---------------------- |
| 1     | 9               |                        |
| 2     | 8               |                        |
| 3     | 7               |                        | <- best
| 4     | 7.1             | loss starts increasing |
| 5     | 7.2             |                        |
| 6     | 7.3             | training early-stops   | <- stopped
| 7     | -               |                        |
| 8     | -               |                        |
| 9     | -               |                        |
```

harupy: Keras' early-stopping has this featurs (`restore_best_weights`):
https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping

harupy: Found a related discussion: https://github.com/PyTorchLightning/pytorch-lightning/issues/1395

penghanyuan: Hi, @harupy , it seems that until now mlflow still doesn't restore the best weights when training is early stopped, are there any updates on this feature?

FlorentMeyer:
Hello,

Could someone please tell me why this `__MlflowPLCallback` isn't actually used when using a `pytorch_lightning.Trainer` with a 
`pytorch_lightning.loggers.MLFlowLogger`? I had a hard time understanding that `__MlflowPLCallback` is **not** integrated into `pytorch_lightning`. 

Lightning code [here](https://github.com/Lightning-AI/lightning/blob/95831286b390fd90337bd97e2f624c67f5fba818/src/lightning/pytorch/trainer/connectors/checkpoint_connector.py#L402) and [here](https://github.com/Lightning-AI/lightning/blob/7c80fe6990dfa8c8ce77e74f425239123cfdd50b/src/lightning/pytorch/loggers/mlflow.py#L363) causes checkpoints to be uploaded as state dicts instead of whole `LightningModule`s, therefore it is not possible to use `torch.load()` directly on such checkpoints (not saying I would need this feature, just that it was what I expected `mlflow` to do as it does do it in `mlflow.pytorch.load_model()`)

Thanks in advance ;)

PR #3607, opened 2020-10-28
Log torchserve artifacts under `torchserve_artifacts` and log their locations in MLmodel file

harupy:

- A follow-up for #3436 
- Log torchserve artifacts under `torchserve_artifacts`.
- Log `requirements_file` and `extra_files` in `MLmodel`.

Screen Shot 2020-10-28 at 14 50 33

cc @shrinath-suresh @chauhang

## How is this patch tested?

unit tests

Components 

Interface 

Language 

Integrations


tests/pytorch/test_pytorch_model_export.py
```diff
@@ -817,55 +817,73 @@ def test_sagemaker_docker_model_scoring_with_sequential_model_and_default_conda_
...
+            requirements_file_path = loaded_requirements_file["path"]
+            requirements_file_path = os.path.join(model_path, requirements_file_path)
+            with open(requirements_file_path) as fp:
+                assert fp.read() == content_expected
 
 
 @pytest.mark.large
-def test_requirement_file_save_model(create_requirement_file, sequential_model):
-    requirement_file_path = create_requirement_file
-    with mlflow.start_run(), TempDir(remove_on_exit=True) as tmp:
```
harupy, line 855: Removed unnecessary `mlflow.start_run()`.

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -892,51 +910,75 @@ def test_log_model_invalid_requirement_file_type(sequential_model):
...
-    extra_file_name = "extra1.txt"
-    fp = tmpdir.join(extra_file_name)
-    test_string = "pytest_extra_file"
-    fp.write(test_string)
-    return fp
+    fp1 = tmpdir.join("extra1.txt")
+    fp2 = tmpdir.join("extra2.txt")
+    fp1.write("1")
+    fp2.write("2")
+    return [fp1.strpath, fp2.strpath], ["1", "2"]
```
harupy, line 917: Fixed `create_extra_files` to return multiple files

mlflow/pytorch/__init__.py
```diff
@@ -37,6 +37,9 @@
 
 _SERIALIZED_TORCH_MODEL_FILE_NAME = "model.pth"
 _PICKLE_MODULE_INFO_FILE_NAME = "pickle_module_info.txt"
+_TORCHSERVE_ARTIFACTS_KEY = "torchserve_artifacts"
+_TORCHSERVE_EXTRA_FILES_KEY = "extra_files"
+_TORCHSERVE_REQUIREMENTS_FILE_KEY = "requirements_file"
```
harupy, line 42: Expose these constant variables for the torchserve plugin

harupy, line 42: @shrinath-suresh The torchserve plugin can use these variables to extract required fields from the MLmodel file.

smurching:
@shrinath-suresh would you be able to help review this? It's a follow-up to your PR #3436 adding support for logging additional artifacts with Pytorch models, the primary goal is to expose torchserve artifact metadata (e.g. the location of `requirements.txt` and `extra_files` as part of the MLmodel file so that the TorchServe plugin can look for them under an explicitly specified location

mlflow/pytorch/__init__.py
```diff
@@ -377,31 +380,48 @@ def save_model(
...
                 _download_artifact_from_uri(
                     artifact_uri=extra_file, output_path=tmp_extra_files_dir.path()
                 )
+                rel_path = posixpath.join(
+                    _TORCHSERVE_ARTIFACTS_KEY,
+                    _TORCHSERVE_EXTRA_FILES_KEY,
+                    os.path.basename(extra_file),
+                )
+                torchserve_artifacts_config[_TORCHSERVE_EXTRA_FILES_KEY].append(
+                    {"path": rel_path, "uri": extra_file}
```
smurching, line 420: @harupy what's the intended use case for the URI here? I could imagine one convenience would be to be able to reference e.g. S3 URIs from the TorchServe plugin (for large `extra_files` like embeddings) rather than downloading them + reuploading them as an artifact for the current run. However, since we currently download the files locally anyways & include them alongside the model, maybe we could just expose the relative artifact path for now in `path`

Given that we always download the extra_files

smurching (commented):
@harupy few comments but in general this looks good! Regarding the change to move `extra_files` and `requirements.txt` to be logged under a directory called `torchserve_artifacts`, maybe we can confirm with @chauhang and @shrinath-suresh that TorchServe is the main expected consumer of both `extra_files` and `requirements.txt` (in which case it makes sense to store them under a TorchServe-specific subdirectory)

mlflow/pytorch/__init__.py
harupy: `mllfow.pyfunc.log_model` uses this format when saving the `artifacts` argument in `MLmodel`. I'm not sure what the original intention of this field is, but one benefit I came up with is we can tell where the file came from.

```
# Which foo.txt is this?
path: foo.txt
```

```
# foo.txt in my_bucket
path: foo.txt
uri: s3://my_bucket/foo.txt
```

harupy: URI might contain confidential information though

harupy:
Screen Shot 2020-11-03 at 8 41 39

mlflow/pytorch/__init__.py
```diff
@@ -377,31 +379,36 @@ def save_model(
...
 
         with TempDir() as tmp_requirements_dir:
-            saved_requirements_dir_subpath = "requirements"
             _download_artifact_from_uri(
                 artifact_uri=requirements_file, output_path=tmp_requirements_dir.path()
             )
-            shutil.move(
-                tmp_requirements_dir.path(), posixpath.join(path, saved_requirements_dir_subpath)
-            )
+            rel_path = os.path.basename(requirements_file)
```
harupy, line 392: `os.path.basename` works for URI.

harupy:
```python
>>> os.path.basename("/user/foo.txt")
'foo.txt'
>>> os.path.basename("file:///user/foo.txt")
'foo.txt'
>>> os.path.basename("aaa:bbb///user/foo.txt")
'foo.txt'
```

smurching (approved):
LGTM

PR #3619, opened 2020-11-02
Introduce time-based batch metrics logging and change XGBoost to use it

andrewnitu:

- Introduce a new utility for batch logging metrics: it can be used like a normal metrics logger but it automatically keeps them in memory until the log condition is met, and then does a batch log once the condition is met. The current run condition is when the accumulated training time reaches 10x the time of the accumulated batch logging time (thus logging will introduce a 10% overhead to the training).
- Rewrite XGBoost metrics logging to use this new utility. This is intended as a reference implementation for other ML libraries.

## How is this patch tested?

- Unit tests
- Manual tests

XGBoost autologging will now incrementally log iterations as they are trained, instead of all at the end of the training.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/entities/metric.py
+++ b/mlflow/entities/metric.py
@@ -13,6 +13,14 @@ class Metric(_MLflowObject):
         self._timestamp = timestamp
         self._step = step
 
+    def __eq__(self, other):
+        return (
+            self._key == other._key
+            and self._value == other._value
+            and self._timestamp == other._timestamp
+            and self._step == other._step
+        )
+
     @property
     def key(self):
         """String key corresponding to the metric name."""
--- a/mlflow/entities/metric.py
+++ b/mlflow/entities/metric.py
@@ -13,14 +13,6 @@ class Metric(_MLflowObject):
         self._timestamp = timestamp
         self._step = step
 
-    def __eq__(self, other):
-        return (
-            self._key == other._key
-            and self._value == other._value
-            and self._timestamp == other._timestamp
-            and self._step == other._step
-        )
-
     @property
     def key(self):
         """String key corresponding to the metric name."""
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -195,13 +195,21 @@ def resolve_input_example_and_signature(
     return input_example if log_input_example else None, model_signature
 
 
+# wrapper functions to be able to mock this easily in the tests
+def time_wrapper_for_log():
+    return time.time()
+def time_wrapper_for_current():
+    return time.time()
+def time_wrapper_for_timestamp():
+    return time.time()
+
 # we pass the batch_metrics_handler through, such that the callback can access it
 def _timed_log_batch(batch_metrics_handler, run_id, metrics):
-    start = time.time()
+    start = time_wrapper_for_log()
     metrics_slices = [metrics[i * MAX_METRICS_PER_BATCH:(i + 1) * MAX_METRICS_PER_BATCH] for i in range((len(metrics) + MAX_METRICS_PER_BATCH - 1) // MAX_METRICS_PER_BATCH )]  
     for metrics_slice in metrics_slices:
         MlflowClient().log_batch(run_id=run_id, metrics=metrics_slice)
-    end = time.time()
+    end = time_wrapper_for_log()
     batch_metrics_handler.total_log_batch_time += end - start
     batch_metrics_handler.num_log_batch += 1
 
@@ -246,7 +254,7 @@ class BatchMetricsHandler: # BatchMetricsLogger maybe?
 
     # metrics is a dict representing the set of metrics collected during one iteration
     def record_metrics(self, metrics, step):
-        current_timestamp = time.time()
+        current_timestamp = time_wrapper_for_current()
         if self.previous_training_timestamp is None:
             self.previous_training_timestamp = current_timestamp
             return
@@ -256,9 +264,9 @@ class BatchMetricsHandler: # BatchMetricsLogger maybe?
         self.total_training_time += training_time
 
         if step in self.data:
-            self.data[step].append([int(time.time() * 1000), metrics])
+            self.data[step].append([int(time_wrapper_for_timestamp() * 1000), metrics])
         else:
-            self.data[step] = [[int(time.time() * 1000), metrics]]
+            self.data[step] = [[int(time_wrapper_for_timestamp() * 1000), metrics]]
 
         # print(self.current_step)
         # print("total_training_time " + str(self.total_training_time))
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -231,8 +231,6 @@ class BatchMetricsHandler:  # BatchMetricsLogger maybe?
         self.previous_training_timestamp = None
 
     def _purge(self):
-        print("purging")
-
         run_id = mlflow.tracking.fluent._get_or_start_run().info.run_id
         final_metrics = []
 
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -266,11 +266,9 @@ class BatchMetricsHandler:  # BatchMetricsLogger maybe?
     def record_metrics(self, metrics, step):
         current_timestamp = time_wrapper_for_current()
         if self.previous_training_timestamp is None:
-            self.data[step] = [[int(time_wrapper_for_timestamp() * 1000), metrics]]
             self.previous_training_timestamp = current_timestamp
-            return
 
-        training_time = current_timestamp - self.previous_training_timestamp  # + 0.01
+        training_time = current_timestamp - self.previous_training_timestamp
 
         self.total_training_time += training_time
 
@@ -278,12 +276,6 @@ class BatchMetricsHandler:  # BatchMetricsLogger maybe?
             self.data[step].append([int(time_wrapper_for_timestamp() * 1000), metrics])
         else:
             self.data[step] = [[int(time_wrapper_for_timestamp() * 1000), metrics]]
-
-        # print(self.current_step)
-        # print("total_training_time " + str(self.total_training_time))
-        # print("total_log_batch_time " + str(self.total_log_batch_time))
-        # print("num_log_batch " + str(self.num_log_batch))
-        # print("previous_training_timestamp " + str(self.previous_training_timestamp))
 
         if self._should_purge(current_timestamp):
             self.total_training_time = 0
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -3,12 +3,12 @@ import functools
 import warnings
 import time
 import contextlib
-import concurrent
 
 import mlflow
 from mlflow.utils import gorilla
 from mlflow.entities import Metric
 from mlflow.tracking.client import MlflowClient
+from mlflow.utils.validation import MAX_METRICS_PER_BATCH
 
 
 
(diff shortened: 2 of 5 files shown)


mlflow/entities/metric.py
```diff
@@ -13,6 +13,14 @@ def __init__(self, key, value, timestamp, step):
         self._timestamp = timestamp
         self._step = step
 
+    def __eq__(self, other):
```
andrewnitu, line 16: is it okay to add this? it is for deep compare of matrix objects in the tests

andrewnitu: nvm

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+    return time.time()
+
+
+# we pass the batch_metrics_handler through, such that the callback can access it
+def _timed_log_batch(batch_metrics_handler, run_id, metrics):
+    start = time_wrapper_for_log()
+    metrics_slices = [
+        metrics[i * MAX_METRICS_PER_BATCH : (i + 1) * MAX_METRICS_PER_BATCH]
+        for i in range((len(metrics) + MAX_METRICS_PER_BATCH - 1) // MAX_METRICS_PER_BATCH)
+    ]
```
dbczumar, line 216: I think using the `step` parameter for `range()` will simplify things here. e.g:

```
def chunks(lst, n):
    """Yield successive n-sized chunks from lst."""
    for i in range(0, len(lst), n):
        yield lst[i:i + n]
```

(Credits to https://stackoverflow.com/a/312464/11952869)

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+# wrapper functions to be able to mock this easily in the tests
+def time_wrapper_for_log():
+    return time.time()
+
+
+def time_wrapper_for_current():
+    return time.time()
+
+
+def time_wrapper_for_timestamp():
```
dbczumar, line 206: Instead of wrapping `time.time()` for mocking, can we just manipulate the `total_log_batch_time` and `total_training_time` properties of `BatchMetricsHandler` in our test cases?

If we're concerned about the measurement of `total_training_time` and `total_log_batch_time`, we can always construct another test case that performs sleeps to simulate training / logging and then verifies that `total_training_time` / `total_log_batch_time` exceed expected thresholds.

mlflow/xgboost.py
```diff
@@ -340,6 +341,9 @@ def record_eval_results(eval_results):
             """
 
             def callback(env):
+                batch_metrics_handler.record_metrics(
```
dbczumar, line 344: Can we add `batch_metrics_handler` as an argument to `record_eval_results` and thread it through to this `callback` to ensure that we're not accidentally referencing some state left over from a previous `BatchMetricsHandler`, for example?

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+
+        if self._should_purge():
+            self.total_training_time = 0
+            self._purge()
+
+        self.previous_training_timestamp = current_timestamp
+
+
+@contextlib.contextmanager
+def with_batch_metrics_handler():
```
dbczumar, line 286: Can we add docs here?

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+        MlflowClient().log_batch(run_id=run_id, metrics=metrics_slice)
+    end = time_wrapper_for_log()
+    batch_metrics_handler.total_log_batch_time += end - start
+    batch_metrics_handler.num_log_batch += 1
+
+
+class BatchMetricsHandler:  # BatchMetricsLogger maybe?
+    def __init__(self):
+        # data is an array of tuples of the form (timestamp, metrics at timestamp)
+        self.data = {}
```
dbczumar, line 227: Seems like a dictionary to me! (Though I think it should be a list of `Metric` objects - see comment below)

dbczumar (changes requested):
@andrewnitu Awesome! Left a few comments

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+        current_timestamp = time_wrapper_for_current()
+        if self.previous_training_timestamp is None:
+            self.previous_training_timestamp = current_timestamp
+
+        training_time = current_timestamp - self.previous_training_timestamp
+
+        self.total_training_time += training_time
+
+        if step in self.data:
+            self.data[step].append([int(time_wrapper_for_timestamp() * 1000), metrics])
```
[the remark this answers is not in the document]
dbczumar: On second thought, I'm not sure we need these timer mocks. See https://github.com/mlflow/mlflow/pull/3619#discussion_r516864583

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+                for key, value in metrics_at_timestamp.items():
+                    final_metrics.append(Metric(key, value, timestamp, step))
+
+        _timed_log_batch(self, run_id=run_id, metrics=final_metrics)
+
+        self.data = {}
+
+    def _should_purge(self):
+        if self.num_log_batch == 0:
+            return True
```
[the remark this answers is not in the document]
andrewnitu: divide by 0 if we dont catch it here :)

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+            self.previous_training_timestamp = current_timestamp
+
+        training_time = current_timestamp - self.previous_training_timestamp
+
+        self.total_training_time += training_time
+
+        if step in self.data:
+            self.data[step].append([int(time_wrapper_for_timestamp() * 1000), metrics])
+        else:
+            self.data[step] = [[int(time_wrapper_for_timestamp() * 1000), metrics]]
```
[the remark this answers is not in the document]
andrewnitu: yeah i don't see why not. Seems kinda wasteful to group them by step then ungroup then again as i'm doing now

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,98 @@ def resolve_input_example_and_signature(
...
+    def __init__(self):
+        # data is an array of tuples of the form (timestamp, metrics at timestamp)
+        self.data = {}
+        self.total_training_time = 0
+        self.total_log_batch_time = 0
+        self.num_log_batch = 0
+        self.previous_training_timestamp = None
+
+    def _purge(self):
+        run_id = mlflow.tracking.fluent._get_or_start_run().info.run_id
```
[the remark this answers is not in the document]
andrewnitu: sure, that makes sense. I just needed the run_id and that was my first thought to grab it :D

mlflow/utils/autologging_utils.py
andrewnitu: i tried to test this but for some reason at least in pytest sleep doesnt increase the system clock, even though the test is obviously taking longer (so the sleep is running)

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,90 @@ def resolve_input_example_and_signature(
...
+        training_time = current_timestamp - self.previous_training_timestamp
+
+        self.total_training_time += training_time
+
+        for key, value in metrics.items():
+            self.data.append(Metric(key, value, current_timestamp, step))
+
+        if self._should_purge():
+            self._purge()
+            self.total_training_time = 0
```
dbczumar, line 260: Why do we reset the total training time here?

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,83 @@ def resolve_input_example_and_signature(
...
+            for i in range(0, len(self.data), MAX_METRICS_PER_BATCH)
+        ]
+        for metrics_slice in metrics_slices:
+            MlflowClient().log_batch(run_id=self.run_id, metrics=metrics_slice)
+        end = time.time()
+        self.total_log_batch_time += end - start
+
+    def _should_purge(self):
+        if self.total_log_batch_time == 0:  # we don't yet have data on how long logging takes
+            return True
```
dbczumar, line 224: *Now* I don't think we need this :)

tests/utils/test_autologging_utils.py
```diff
@@ -263,3 +268,100 @@ def modifies(_):
 
     assert x["data"] == 0
     logger.warning.assert_not_called()
+
+
+def test_batch_metrics_logger_logs_all_metrics(start_run):  # pylint: disable=unused-argument
+    with mock.patch.object(MlflowClient, "log_batch") as log_batch_mock:
```
dbczumar, line 274: Do we need to mock `log_batch`? Ideally, it would be nice to test that the metrics are actually logged by leaving this unmocked and querying the run data after the iterative calls to `record_metrics()` complete.

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,80 @@ def resolve_input_example_and_signature(
...
+        current_timestamp = time.time()
+        if self.previous_training_timestamp is None:
+            self.previous_training_timestamp = current_timestamp
+
+        training_time = current_timestamp - self.previous_training_timestamp
+
+        self.total_training_time += training_time
+
+        for key, value in metrics.items():
+            self.data.append(Metric(key, value, current_timestamp, step))
```
dbczumar, line 247: the metric timestamp must be an integer value with millisecond resolution; e.g., this should be the following (based on what we do in the fluent API - https://github.com/mlflow/mlflow/blob/055978cdd425ac7c328069d80230f5bc50265df8/mlflow/tracking/fluent.py#L443):

```suggestion
            self.data.append(Metric(key, value, int(current_timestamp * 1000), step))
```

I found this by running our XGBoost example, where I encountered warning logs from the file store about operating on timestamp content in float form, rather than integer form:

```
/Users/czumar/mlflow/mlflow/xgboost.py:412: DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()
  all_arg_names = inspect.getargspec(original)[0]  # pylint: disable=W1505
[0]     train-mlogloss:0.74723
[1]     train-mlogloss:0.54060
[2]     train-mlogloss:0.40276
[3]     train-mlogloss:0.30789
[4]     train-mlogloss:0.24052
[5]     train-mlogloss:0.19087
[6]     train-mlogloss:0.15471
[7]     train-mlogloss:0.12807
[8]     train-mlogloss:0.10722
[9]     train-mlogloss:0.09053
/Users/czumar/mlflow/mlflow/xgboost.py:387: UserWarning: Logging to MLflow failed: invalid literal for int() with base 10: '1604679208.8766131'
  try_mlflow_log(mlflow.log_artifact, filepath)
/Users/czumar/mlflow/mlflow/xgboost.py:465: UserWarning: Logging to MLflow failed: invalid literal for int() with base 10: '1604679208.8766131'
  try_mlflow_log(mlflow.log_artifact, filepath)
/Users/czumar/mlflow/mlflow/xgboost.py:501: UserWarning: Logging to MLflow failed: invalid literal for int() with base 10: '1604679208.8766131'
  input_example=input_example,
```

^ Interestingly, the error is only encountered when subsequent artifact logging operations are called, at which point the file store reads the logged metric files (not sure why it needs to do this, but it does)

dbczumar: Can we make sure to add a test case for this, ensuring that timestamps have millisecond resolution and are integer values?

mlflow/utils/autologging_utils.py
```diff
@@ -187,3 +192,80 @@ def resolve_input_example_and_signature(
...
+        self.data = []
+
+    def _timed_log_batch(self):
+        start = time.time()
+        metrics_slices = [
+            self.data[i : i + MAX_METRICS_PER_BATCH]
+            for i in range(0, len(self.data), MAX_METRICS_PER_BATCH)
+        ]
+        for metrics_slice in metrics_slices:
+            MlflowClient().log_batch(run_id=self.run_id, metrics=metrics_slice)
```
dbczumar, line 218: Can we wrap this in a call to `try_mlflow_log()` to ensure that failures don't prevent future metrics from being logged? 

We should document this behavior and, if possible, we should add a test case for it too.

dbczumar (commented):
Almost there! Found a few more small issues

tests/utils/test_autologging_utils.py
andrewnitu: do we want to do this for every test? or is it sufficient to do it for one test and know that it works end-to-end, then save time by keeping the mock for the rest of the tests

andrewnitu: actually it seems this really only applies to one of the tests.. the rest care about the intermediate state AS the BatchMetricsLogger is logging, not the final outcome, so it makes more sense to use mocks there

mlflow/utils/autologging_utils.py
```diff
@@ -261,6 +264,9 @@ def batch_metrics_logger(run_id):
     that logging imposes no more than a 10% overhead on the training, where the training is
     measured by adding up the time elapsed between consecutive calls to record_metrics.
 
+    If logging a batch fails, a log will be emitted and subsequent metrics will continue to
```
dbczumar, line 267:
```suggestion
    If logging a batch fails, a warning will be emitted and subsequent metrics will continue to
```

dbczumar (approved):
LGTM with one tiny comment - thanks @andrewnitu !

PR #3630, opened 2020-11-03
fix issue #1997 with sparkling water and s3

delcacho:
Signed-off-by: Carlos Del Cacho Vicente <delcacho@gmail.com>

Issue #1997 with the fix identified in pull request #2882, whereby H2O models are not correctly saved to S3 when using Sparkling Water. Basically download_model should be used instead of save_model, if available. I added a check to ensure the function exists before invoking it, irrespective of H2O version.

## How is this patch tested?

sparkling-water-3.30.0.3-1-2.4
pysparkling

No

integrations/h2o

Bug fix.

### DIFF
--- a/mlflow/h2o.py
+++ b/mlflow/h2o.py
@@ -221,11 +221,12 @@ def _load_model(path, init=False):
 
     model_path = os.path.join(path, params["model_file"])
     if "upload_model" in dir(h2o):
-       model = h2o.upload_model(model_path)
+        model = h2o.upload_model(model_path)
     else:
-       model = h2o.load_model(model_path)
+        model = h2o.load_model(model_path)
 
     return model
+
 
 class _H2OModelWrapper:
     def __init__(self, h2o_model):
--- a/mlflow/h2o.py
+++ b/mlflow/h2o.py
@@ -108,7 +108,7 @@ def save_model(
         _save_example(mlflow_model, input_example, path)
 
     # Save h2o-model
-    if "download_model" in dir(h2o):
+    if hasattr(h2o,"download_model"):
         h2o_save_location = h2o.download_model(model=h2o_model, path=model_data_path)
     else:
         h2o_save_location = h2o.save_model(model=h2o_model, path=model_data_path, force=True)
@@ -220,7 +220,7 @@ def _load_model(path, init=False):
         h2o.no_progress()
 
     model_path = os.path.join(path, params["model_file"])
-    if "upload_model" in dir(h2o):
+    if hasattr(h2o,"upload_model"):
         model = h2o.upload_model(model_path)
     else:
         model = h2o.load_model(model_path)
--- a/mlflow/h2o.py
+++ b/mlflow/h2o.py
@@ -108,7 +108,10 @@ def save_model(
         _save_example(mlflow_model, input_example, path)
 
     # Save h2o-model
-    h2o_save_location = h2o.save_model(model=h2o_model, path=model_data_path, force=True)
+    if "download_model" in dir(h2o):
+        h2o_save_location = h2o.download_model(model=h2o_model, path=model_data_path)
+    else:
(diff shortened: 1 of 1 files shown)


mlflow/h2o.py
```diff
@@ -108,7 +108,10 @@ def save_model(
         _save_example(mlflow_model, input_example, path)
 
     # Save h2o-model
-    h2o_save_location = h2o.save_model(model=h2o_model, path=model_data_path, force=True)
+    if "download_model" in dir(h2o):
```
tomasatdatabricks, line 111: Why the if statement? Is there an h2o version that does not have download_model? I think we should just call download_model without the if statement.

mlflow/h2o.py
```diff
@@ -215,7 +218,14 @@ def _load_model(path, init=False):
     if init:
         h2o.init(**(params["init"] if "init" in params else {}))
         h2o.no_progress()
-    return h2o.load_model(os.path.join(path, params["model_file"]))
+
+    model_path = os.path.join(path, params["model_file"])
+    if "upload_model" in dir(h2o):
```
tomasatdatabricks, line 223: Same as above, can we drop the if statement and just call upload_model?

tomasatdatabricks (commented):
Thanks for the fix! I left 2 comments.

delcacho:
I would keep that for older versions of h2o, they are a recent addition.
The other functions work just fine when working with one machine, the
problem appears in distributed training with Sparkling Water.

tomasatdatabricks:
I see, I did not realize this was a recent addition. Makes sense in that case, but can you use hasattr(h2o, 'download_model') instead of the dir? 

In addition, I am worried about the silent failure as described in the linked issue - i.e. it appears you successfully logged the model but the files are missing. Is there a way we can check? E.g. can we check for existence of the model fail and raise exception if the file is not there?

delcacho:
Ok, changing to hasattr instead. As for the silent failure, it will always fail if using Sparkling Water and calling save_model, because it saves the model in the wrong location (worker instead of driver, where the Mlflow model package is built). I don't know if there is a way to check if h2o is being run over Sparkling Water, I am going to check just in case.

delcacho:
Best path I can think of is check if "pysparkling" in dir(), and throw an MlflowException if download_model is not available. Will that suffice for the silent error?

tomasatdatabricks:
Thanks @delcacho! 

As for the check - I was thinking we can 
a) try if h2o.connection().ip is localhost - but I am not sure if this is bulletproof test since there can be multiple workers, I don't 
remember the logic for which worker would actually save the model) 

or 

b) call save_model and check if any model file was saved.  

What do you think? I am also ok merging it without the check since your PR strictly improves the existing behavior, but it may be nice to have the check in place.

delcacho:
I am not sure my sanity check is a good one either. I would have to carefully look at the internals of Sparkling Water.

I would suggest that we remove it and simply add a comment somewhere in the documentation that says that Mlflow is compatible with Sparkling Water versions 3.30.0.1 and higher.

tomasatdatabricks:
Right, I agree. 

I don't think this is specific to sparkling water btw - it would be the case with any H2O multi-node cluster. 

How about we print a warning in case the download method is missing, saying something like - if your h2o cluster is remote, the model might not be saved correctly, please upgrade your h2o version?

delcacho:
Ok, should be good to go now. Thanks for the comments.

tomasatdatabricks (approved):
lgmt 

thanks @delcacho!

tomasatdatabricks:
Approved, will merge after the tests are done, thanks for the contribution!

delcacho:
You're welcome. That Pytorch test is failing again. I think it is unrelated to the PR.

tomasatdatabricks:
Yeah I don't think it's related to this PR either. I will rerun the test and merge after.

PR #3644, opened 2020-11-06
[SkinnyClient] Remove most dependencies except for pandas, and numpy

eddiestudies:

Initial work for creating a skinny client version of MLflow with fewer dependencies

## How is this patch tested?

The build script added verifies the changes

## Release Notes
No pypi facing changes yet

Components 

Interface 

Language 

Integrations


setup.py
```diff
@@ -25,38 +29,64 @@ def package_files(directory):
...
+    "pandas",
+    "numpy",
+    "sqlparse",
+]
+
+_is_mlflow_skinny = bool(os.environ.get(_MLFLOW_SKINNY_ENV_VAR))
+logging.debug("{} env var is set: {}".format(_MLFLOW_SKINNY_ENV_VAR, _is_mlflow_skinny))
+
+if not _is_mlflow_skinny:
+    requirements.extend(
```
dbczumar, line 55: This call to `extend()` contains a lot of duplicate elements that are already in `requirements`. We should either dedupe these or, perhaps preferably, maintain separate lists of requirements.

dev/run-python-skinny-tests.sh
```diff
@@ -0,0 +1,13 @@
...
+set -x
+# Set err=1 if any commands exit with non-zero status as described in
+# https://stackoverflow.com/a/42219754
+err=0
+trap 'err=1' ERR
+export MLFLOW_HOME=$(pwd)
+export MLFLOW_SKINNY='true'
+
+pytest --verbose tests/test_skinny.py
+pytest --verbose tests/tracking/service/
```
dbczumar, line 11: Guessing this is just a start, but it's worth noting that we'll also want test coverage for project execution & model registry interactions.

dbczumar (commented):
@eedeleon This is a great start - left a few small comments.

One decision I think we should raise / make before merging is whether or not it makes sense to pull some of the non-skinny functionality into separate libraries that all populate the `mlflow` namespace. For example, we could define:

- `mlflow-sqlalchemy`: Plugin code for SQL-based backend storage
- `mlflow-filestore`: Plugin code for file-based backend
- `mlflow-server`: MLflow JS UI & backend server CLI (could also pull in the sqlalchemy and file store plugins for convenience)
- `mlflow-client`: The current MLflow library minus all of the above (i.e. excluding file store, sqlalchemy store, and server).

`mlflow` would continue to serve as the union of these libraries. This has a few benefits:

- Third-party library dependency management becomes clearer; each library has its own `setup.py` file with a collection of requirements. This may be easier to reason about than the list of skinny vs non-skinny.

- Users can more easily pull in the file store / sqlalchemy store / server alongside the skinny client without requiring all of MLflow's other platform libraries.

It also has a few drawbacks:

- The MLflow API surface becomes fragmented across libraries (though perhaps this is easier to reason about than attempting to use an API and encountering import error(s))

- Documenting a fragmented API surface may be challenging as well (the current `mlflow.org` docs are not set up for this kind of separation). Note that we'll eventually need to identify a strategy for documenting skinny vs non-skinny anyway.

- This separation likely complicates the build procedure & code structure / layout

Would be great to get some thoughts from @akshaya-a @mparkhe @smurching here as well. (Perhaps some API segmentation is desired but not in the exact fashion outlined above - let me know!)

setup.py
```diff
@@ -25,38 +29,64 @@ def package_files(directory):
...
+    "protobuf>=3.6.0",
+    "gitpython>=2.1.0",
+    "pyyaml",
+    "querystring_parser",
+    "simplejson",
+    "entrypoints",
+    "gorilla",
+    "pandas",
+    "numpy",
+    "sqlparse",
```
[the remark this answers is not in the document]
eedeleon: I need to check on this. I remember search runs needed this. And I could not get the RestStore tests to run easily because the server code requires extra dependencies, so the FileStore was useful for testing at a minimum. We may be able to find a way of creating the server in a separate conda env on a specified port for builds

setup.py
eedeleon: I will dedup

setup.py
```diff
@@ -25,38 +29,64 @@ def package_files(directory):
...
+            "gorilla",
+            "prometheus-flask-exporter",
+        ]
+    )
+
 setup(
     name="mlflow",
     version=version,
     packages=find_packages(exclude=["tests", "tests.*"]),
     package_data={"mlflow": js_files + models_container_server_files + alembic_files},
```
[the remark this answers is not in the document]
eedeleon: I can take it out for now. We can figure out bringing it back later

setup.py
```diff
@@ -25,38 +29,64 @@ def package_files(directory):
     "../mlflow/temporary_db_migrations_for_pre_1_users/alembic.ini",
 ]
 
+requirements = [
+    "click>=7.0",
+    "cloudpickle",
```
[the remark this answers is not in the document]
eedeleon: Given that cloudpickle is used for pyfunc, I think it should be included in this first pass. fastai, keras, pytorch, and sklearn flavors all need it as well.

Also, all flavors are imported on import mlflow which means all flavors would have to be edited to lazy load now.

dev/run-python-skinny-tests.sh
eedeleon: Yeah this is to unblock tests.

I think the target is plugin + RestStore tests.

For the above comment about nested packages, I think that is out of scope for this initial set of proposed changes

eedeleon: If you have a solution of isolated server tests so we can split the envs: skinny env sdk + server env for e2e tests that would help for the next set of contributions around RestStore and plugins

eedeleon: I am familiarizing myself with the test infra around the server but my current options are quite involved.

I am hoping for a simpler solution

eedeleon:
Moved to https://github.com/mlflow/mlflow/pull/3687 due to git pull --rebase showing 40 new unrelated files in the diff

PR #3649, opened 2020-11-06
Java Client changes to incorporate PreSigned Url support.

arjundc-db:

In https://github.com/mlflow/mlflow/pull/3643, a new artifact repository was added to support downloading model artifacts from Databricks hosted registry using presigned urls.
This PR contains changes to ensure the MLflow Java client is compliant with these changes.

## How is this patch tested?

Unit tests were added.

This is part 2 of #3643, where a new artifact repository is added to support downloading model version artifacts stored in Databricks using presigned urls.

Components 
- [ x] `area/artifacts`: Artifact stores and artifact logging
- [ x] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry

Interface 

Language 
- [ x] `language/java`: Java APIs and clients

Integrations


mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -748,11 +748,12 @@ public String getModelVersionDownloadUri(String modelName, String version) {
    * @return A local file or directory ({@link java.io.File}) containing model artifacts
    */
   public File downloadModelVersion(String modelName, String version) {
-    String downloadUri = getModelVersionDownloadUri(modelName, version);
-
+    String scheme = "models";
```
andychow-db, line 751: This should be a private member variable rather than instantiated within the function.

mlflow/store/artifact/databricks_models_artifact_repo.py
```diff
@@ -0,0 +1,127 @@
...
+from mlflow.utils.file_utils import download_file_using_http_uri
+from mlflow.utils.rest_utils import http_request
+from mlflow.utils.uri import get_databricks_profile_uri_from_artifact_uri
+from mlflow.store.artifact.utils.model_utils import (
+    get_model_name_and_version,
+    is_using_databricks_registry,
+)
+
+_logger = logging.getLogger(__name__)
+_DOWNLOAD_CHUNK_SIZE = 100000000
```
andychow-db, line 19: How did you specify this number? Shouldn't it be 2 powered value at least?

andychow-db (commented):
Is this supposed to be both the python and Java changes? The title and description imply no python changes.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -748,11 +748,12 @@ public String getModelVersionDownloadUri(String modelName, String version) {
...
+    String path = modelName + "/" + version;
+    URIBuilder downloadUriBuilder = new URIBuilder().setScheme(scheme).setPath(path);
     CliBasedArtifactRepository repository = new CliBasedArtifactRepository(null, null,
             hostCredsProvider);
-    return repository.downloadArtifactFromUri(downloadUri);
+    return repository.downloadArtifactFromUri(downloadUriBuilder.toString());
```
sueann, line 756: can we comment that this won't work if the model version download uri points to a single file not a directory due to the way many ArtifactRepository's download_artifacts handles empty subpaths?

mlflow/store/artifact/databricks_models_artifact_repo.py
arjundc-db: At the time the `DatabricksArtifactRepository` was added, we could upload to Azure in chucks of 100 MB. So we just used the same chunk size to download these artifacts.
Also we use response.iter_content to download a file as downloading large files could result in OOM errors. So 100 MB seemed decent to use. Can we changed if you feel strongly about this.

arjundc-db:
@andychow-db :  There are only two files with the changes, whose path is mlflow/java/client/src. The remaining files are a result of #3643 over which this PR is built. It will collapse once #3643 is merged.

mlflow/store/artifact/utils/model_utils.py
```diff
@@ -0,0 +1,57 @@
+import urllib.parse
```
sueann, line 1: hm merge conflict? i think this is now `mlflow/store/artifact/utils/models.py` now.

sueann (approved):
lgtm - don't forget to remove the py changes before merging!

also, in the release notes section, can you mark it as `rn/feature` and in the release notes description mention that this is part 2 of  #3643 (so the release notes can be grouped together into one feature)?

PR #3653, opened 2020-11-09
Do not disable existing loggers when init alembic

arthury1n:

In the [env.py](https://github.com/mlflow/mlflow/blob/master/mlflow/store/db_migrations/env.py#L14) that Mlflow is providing to alembic library, the disable_existing_loggers parameter is set to True(by default). This causes any non-root loggers be disabled thereafter but the sqlalchemy.engine and alembic loggers.

And it has affected us in our CI because we will use 'sqlite://mlflow.db' as the backend db which will be empty when CI job started. And amlebic client will be init then to upgrade this db. Having all loggers disabled in the middle of our tests breaks the pytest fixture 'caplog' as it won't be able to capture any logs, and therefore break the unit tests which used 'caplog' fixture.

As an example below, the first test will succeed and the second will fail.
```
import logging
import mlflow

def test1():
    import mlflow
    c = mlflow.tracking.MlflowClient(tracking_uri="sqlite:///mlflow.db")
    assert c

def test2(caplog):
    logger = logging.getLogger('application')
    logger.warning('test')
    assert len(caplog.records) == 1
```
This PR is to add disable_existing_loggers=False in the [env.py](https://github.com/mlflow/mlflow/blob/master/mlflow/store/db_migrations/env.py#L14). So there will be no impact to any non-root loggers which is created in user's project.

## How is this patch tested?

(Details)

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/store/db_migrations/env.py
+++ b/mlflow/store/db_migrations/env.py
@@ -11,7 +11,7 @@ config = context.config
 
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
-fileConfig(config.config_file_name)
+fileConfig(config.config_file_name, disable_existing_loggers=False)
 
 # add your model's MetaData object here
 # for 'autogenerate' support


harupy (approved):
@arthury1n Thanks for the PR! LGTM :)

PR #3678, opened 2020-11-12
Add `log_text`

harupy:

Add a new tracking API `log_text` which logs a text as an artifact.

```
mlflow.log_text("a", "b.txt")
```

## How is this patch tested?

Unit test

Add a new fluent API `mlflow.log_text` which logs in-memory text as an artifact.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -4,6 +4,9 @@ and model versions. This is a lower level API than the :py:mod:`mlflow.tracking.
 and is exposed in the :py:mod:`mlflow.tracking` module.
 """
 import logging
+import os
+import posixpath
+import tempfile
 
 from mlflow.entities import ViewType
 from mlflow.entities.model_registry.model_version_stages import ALL_STAGES
@@ -865,6 +868,49 @@ class MlflowClient(object):
             is_dir: False
         """
         self._tracking_client.log_artifact(run_id, local_path, artifact_path)
+
+    import contextlib
+
+    @contextlib.contextmanager
+    def _log_artifact_helper(self, run_id, artifact_file):
+        """
+        Yields a temporary path (absolute), and then call `log_artifact` against that path.
+        """
+        norm_path = posixpath.normpath(artifact_file)
+        filename = posixpath.basename(norm_path)
+        artifact_path = posixpath.dirname(norm_path)
+        artifact_path = None if artifact_path == "" else artifact_path
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            tmp_path = os.path.join(tmp_dir, filename)
+            yield tmp_path
+            self.log_artifact(run_id, tmp_path, artifact_path)
+
+    def log_text(self, run_id, text, artifact_file):
+        """
+        Log a text as an artifact.
+
+        :param run_id: String ID of the run.
+        :param text: Text to log.
+        :param artifact_file: Path in ``artifact_uri``.
+
+        ..code-block:: python
+            :caption: Example
+
+            from mlflow.tracking import MlflowClient
+
+            client = MlflowClient()
+            run = client.create_run(experiment_id="0")
+
+            client.log_text(run.info.run_id, "text", "file.txt")
+            client.log_text(run.info.run_id, "text", "dir/file.txt")
+
+            client.log_text(run.info.run_id, "<h1>header</h1>", "index.html")
+            client.log_text(run.info.run_id, "1,2,3", "data.csv")
+        """
+        with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
+            with open(tmp_path, "w") as f:
+                f.write(text)
 
     def log_artifacts(self, run_id, local_dir, artifact_path=None):
         """
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -945,7 +945,7 @@ class MlflowClient(object):
             client = MlflowClient()
             run = client.create_run(experiment_id="0")
 
-            # Log a text in the artifact root
+            # Log text to a file under the run's root artifact directory
             client.log_text(run.info.run_id, "text1", "file1.txt")
 
             # Log a text in a subdirectory
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -931,7 +931,7 @@ class MlflowClient(object):
 
     def log_text(self, run_id, text, artifact_file):
         """
-        Log a text as an artifact.
+        Log text as an artifact.
 
         :param run_id: String ID of the run.
         :param text: Text to log.
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -917,7 +917,7 @@ class MlflowClient(object):
     @contextlib.contextmanager
     def _log_artifact_helper(self, run_id, artifact_file):
         """
-        Yields a temporary path (absolute), and then calls `log_artifact` against that path.
+        Yields a temporary path to store a file, and then calls `log_artifact` against that path.
         """
         norm_path = posixpath.normpath(artifact_file)
         filename = posixpath.basename(norm_path)
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -935,7 +935,8 @@ class MlflowClient(object):
 
         :param run_id: String ID of the run.
         :param text: String containing text to log.
-        :param artifact_file: The run-relative artifact file path to which the text is saved.
+        :param artifact_file: The run-relative artifact file path in posixpath format to which
+                              the text is saved (e.g. "dir/file.txt").
 
         .. code-block:: python
             :caption: Example
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -906,7 +906,7 @@ class MlflowClient(object):
             client.log_text(run.info.run_id, "text", "dir/file.txt")
 
             client.log_text(run.info.run_id, "<h1>header</h1>", "index.html")
-            client.log_text(run.info.run_id, "1,2,3", "data.csv")
+            client.log_text(run.info.run_id, "print('mlflow')", "script.py")
         """
         with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
             with open(tmp_path, "w") as f:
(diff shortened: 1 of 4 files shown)


mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+            run = client.create_run(experiment_id="0")
+
+            # Log a text in the artifact root
+            client.log_text(run.info.run_id, "text1", "file1.txt")
+
+            # Log a text in a subdirectory
+            client.log_text(run.info.run_id, "text2", "dir/file2.txt")
+
+            # Log an HTML text
+            client.log_text(run.info.run_id, "<h1>header</h1>", "index.html")
```
harupy, line 955: Added this example to indicate that this API accepts any text file formats (e.g. .html, .csv, .py, .sh)

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
         self._tracking_client.log_artifacts(run_id, local_dir, artifact_path)
 
+    @contextlib.contextmanager
+    def _log_artifact_helper(self, run_id, artifact_file):
+        """
+        Yields a temporary path to store a file, and then calls `log_artifact` against that path.
+        """
+        norm_path = posixpath.normpath(artifact_file)
+        filename = posixpath.basename(norm_path)
+        artifact_path = posixpath.dirname(norm_path)
```
harupy, line 924: Used `normpath` here to handle a relative path like `"./a.txt"` or `"a/../b.txt"`.

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+            yield tmp_path
+            self.log_artifact(run_id, tmp_path, artifact_path)
+
+    def log_text(self, run_id, text, artifact_file):
+        """
+        Log a text as an artifact.
+
+        :param run_id: String ID of the run.
+        :param text: Text to log.
+        :param artifact_file: The run-relative artifact file path to which the text is saved.
```
harupy, line 938: Should we allow a windows path (e.g. `"foo\\bar.txt"`) as `artifact_file `?

harupy: `mlflow.log_artifact` doesn't accept a windows-style path:

```python
import mlflow


    mlflow.log_artifact("README.rst", "foo\\bar.txt")

    files = [a.path for a in client.list_artifacts(run.info.run_id)]
    print(files)
```

output:

```
['foo\\bar.txt']
```

mlflow/tracking/fluent.py
```diff
@@ -557,6 +557,32 @@ def log_artifacts(local_dir, artifact_path=None):
     MlflowClient().log_artifacts(run_id, local_dir, artifact_path)
 
 
+def log_text(text, artifact_file):
```
harupy, line 560: Should we mark this API as experimental?

mlflow/tracking/client.py
smurching: I don't think we need to support windows-style paths (technically the `artifact_file` is like a suffix to append to the run's artifact URI), but we should just make it clear in the `artifact_file` docstring that the specified path is expected to be in posixpath format

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+    def _log_artifact_helper(self, run_id, artifact_file):
+        """
+        Yields a temporary path to store a file, and then calls `log_artifact` against that path.
+        """
+        norm_path = posixpath.normpath(artifact_file)
+        filename = posixpath.basename(norm_path)
+        artifact_path = posixpath.dirname(norm_path)
+        artifact_path = None if artifact_path == "" else artifact_path
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
smurching, line 927: Seems like we should clean up this directory by calling `cleanup` after logging the artifact? https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

mlflow/tracking/client.py
harupy: Makes sense. I'll update the `artifact_file` description.

mlflow/tracking/fluent.py
```diff
@@ -557,6 +557,32 @@ def log_artifacts(local_dir, artifact_path=None):
...
+    """
+    Log a text as an artifact.
+
+    :param text: Text to log.
```
smurching, line 564:
```suggestion
    :param text: String text to log.
```

mlflow/tracking/fluent.py
```diff
@@ -557,6 +557,32 @@ def log_artifacts(local_dir, artifact_path=None):
     MlflowClient().log_artifacts(run_id, local_dir, artifact_path)
 
 
+def log_text(text, artifact_file):
+    """
+    Log a text as an artifact.
```
smurching, line 562:
```suggestion
    Log text as an artifact.
```

mlflow/tracking/fluent.py
smurching: I think it's fine for this API to be stable; I don't imagine we'll want to rename or remove the `text` or `artifact_file` parameters

mlflow/tracking/client.py
smurching: Could you add a test for logging to such paths?

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+            tmp_path = os.path.join(tmp_dir, filename)
+            yield tmp_path
+            self.log_artifact(run_id, tmp_path, artifact_path)
+
+    def log_text(self, run_id, text, artifact_file):
+        """
+        Log a text as an artifact.
```
smurching, line 934:
```suggestion
        Log text as an artifact.
```

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+
+        :param run_id: String ID of the run.
+        :param text: Text to log.
```
smurching, line 937:
```suggestion
        :param text: String containing text to log.
```

mlflow/tracking/client.py
harupy: Found this comment stackoverflow:

https://stackoverflow.com/questions/13379742/right-way-to-clean-up-a-temporary-folder-in-python-class#comment114370145_19657028

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+
+        .. code-block:: python
+            :caption: Example
+
+            from mlflow.tracking import MlflowClient
+
+            client = MlflowClient()
+            run = client.create_run(experiment_id="0")
+
+            # Log a text in the artifact root
```
smurching, line 948:
```suggestion
            # Log text to a file under the run's root artifact directory
```

mlflow/tracking/client.py
smurching: Oops, I missed

> On completion of the context or destruction of the temporary directory object the newly created temporary directory and all its contents are removed from the filesystem.

from the Python API doc...

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+            client.log_text(run.info.run_id, "text1", "file1.txt")
+
+            # Log a text in a subdirectory
```
smurching, line 951:
```suggestion
            # Log text in a subdirectory of the run's root artifact directory
```

mlflow/tracking/client.py
```diff
@@ -910,6 +914,50 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
...
+            client.log_text(run.info.run_id, "text2", "dir/file2.txt")
+
+            # Log an HTML text
```
smurching, line 954:
```suggestion
            # Log HTML text
```

mlflow/tracking/client.py
harupy: done!

mlflow/tracking/fluent.py
harupy: Makes sense.

mlflow/tracking/client.py
harupy: Done in https://github.com/mlflow/mlflow/pull/3678/commits/b72d1c774d1dea7ebb0b227392b52ec4e586e169

mlflow/tracking/client.py
```diff
@@ -910,6 +914,55 @@ def log_artifacts(self, run_id, local_dir, artifact_path=None):
         """
         self._tracking_client.log_artifacts(run_id, local_dir, artifact_path)
 
+    @contextlib.contextmanager
```
smurching, line 917: BTW, what do you think about pulling these artifact logging utils into a private module, e.g. under `mlflow.artifacts._logging_utils` and actually defining `log_text` there as a static function, and then exposing it via the `MlflowClient` here via a single-line call to `mlflow.artifacts._logging_utils.log_text`? The advantage I think is that we can keep this file a bit more concise, especially as we add logging logic for more artifact types like images etc that may be a bit more complex

smurching: Happy to have the refactor be in a follow-up PR or in this one, lmk what you think

smurching (approved):
LGTM, had one comment around moving artifact-logging utils to an internal module, but happy to address that in a follow up!

mlflow/tracking/client.py
smurching: To summarize offline discussion, this actually seems fine - it's hard to define `log_text` itself as a static function due to the dependency on the MlflowClient's `log_artifact`function.

From offline discussion:


> If anything in the future we could consider having utils for logging an in memory object to a pre-specified file path, & that could eventually become the pluggable interface for logging custom object types. You implement "log object type X to a specified file" and then MLflow knows how to use that method to log the object to a tempfile & call log_artifact on it. But I don't think we need to do that just yet

PR #3682, opened 2020-11-13
Introduce utilities for autologging error tolerance / safety

dbczumar:

**NOTE**: This PR is the first in a 2-part series. It introduces developer APIs for writing fault tolerant autologging integrations with corresponding unit tests. https://github.com/mlflow/mlflow/pull/3815 then applies these APIs to the scikit-learn, Keras, and XGBoost autologging integrations and introduces integration tests (*pending*).

**How the new APIs should be used**

To provide error tolerance, developers of an autologging integration should:

1. Decorate autologging integrations with the `@autologging_integration` decorator, which ensures that common configuration settings (e.g., `disable=True/False`) are respected across all integrations. This will also be used for supported version documentation / management.

2. Define patch functions using the `safe_patch` utility, which handles exceptions occurring in the preamble / postamble of a patched ML routine and ensures that the original function is called in the event that an exception arises in the preamble. It is also responsible for cleaning up any dangling MLflow runs left open by an exception during a patch routine.

For reference, patched autologging functions take the following form:

```
def patched_function(original, *args, **kwargs):
    // Preamble
    // Do some MLflow parameter logging, for example

    // Original / underlying function invocation
    original(*args, **kwargs)

    // Postamble
    // Log post-training metrics, models, for example
```

Developers can either pass a function object or a `PatchFunction` class definition to `safe_patch`. `PatchFunction` provides two methods: `_patch_implementation`, which defines the patched function logic, and `_on_exception`, which defines cleanup logic that is invoked when an uncaught exception in `_patch_implementation` causes the patch code to terminate prematurely.

A common use case for `PatchFunction` is to create and tear down an active MLflow run during patched code execution. Because this use case applies to most of our autologging integrations, we provide a `with_managed_run` method that augments a function object or `PatchFunction` instance to provide run creation & cleanup capabilities.

3. Within patch code, "decorate" all extra arguments passed to original / underlying ML routines with new exception safety decorators: `error_safe_function` (a function decorator used to wrap functions in a try/catch block) and `ErrorSafeClass`, a metaclass used to wrap all methods defined on a class (i.e. an ML training callback) in  a try/catch block.

**Further description of internal utilities:**

- `mlflow.utils.autologging_utils.autologging_integration`: A decorator used to store configuration arguments to `autolog()` functions in global state for later access by patch functions. This enables us to ensure that all autologging integrations respect config parameters such as `disable=True` without requiring additional logic in each integration module.

- `mlflow.utils.autologging_utils.safe_patch`: A replacement for `wrap_patch` providing exception handling / disablement capabilities. `safe_patch` can be configured to operate in "test mode" via environment variable (`MLFLOW_AUTOLOGGING_TESTING` for now); test mode bypasses error handling logic to ensure that our patch code doesn't raise exceptions during testing, and it also performs argument validation on calls to underlying / original functions. We propose to enable test mode before running our autologging test suites.

- `mlflow.utils.autologging_utils.PatchFunction`: An optional class representation for a patch function that can be used to define cleanup logic (via `_on_exception`) in the event of unexpected exceptions causing the primary patch routine (`_invoke_patch`) to terminate prematurely.

-  `mlflow.utils.autologging_utils.with_managed_run`: A convenience function that can be applied to an existing patch function object or `PatchFunction` class definition that provides logic for creating and tearing down an autologging MLflow run when patched code executes (if a run does not already exist).

- `mlflow.utils.autologging_utils.error_safe_function`: A decorator used to add a broad try/catch statement to a particular function. Extra function arguments passed to original / underlying ML training methods (e.g., XGBoost callbacks inserted by autologging integrations) are expected to possess this decorator (this is validated by `safe_patch` in test mode).

- `mlflow.utils.autologging_utils.ErrorSafeClass`: A metaclass that wraps each method defined on a class in a broad try/catch statement. Extra class arguments passed to original / underlying ML training methods (e.g., Keras callbacks inserted by autologging integrations) are expected to have this metaclass.

As an example reference, these utilities have been applied to autologging integrations for scikit-learn, XGBoost, and Keras.

## How is this patch tested?

Unit tests have been introduced to `tests/autologging/test_autologging_utils.py` and `tests/autologging/test_autologging_safety_unit.py`.

Components 

Interface 

Language 

Integrations


mlflow/utils/autologging_utils.py
```diff
@@ -275,3 +282,201 @@ def batch_metrics_logger(run_id):
...
+        if type(autologging_kwarg) == list:
+            user_kwarg = user_kwarg + ([None] * (len(autologging_kwarg) - len(user_kwarg)))
+            for a, u in zip(autologging_kwarg, user_kwarg):
+                _validate(a, u)
+        elif type(autologging_kwarg) == dict:
+            assert set(user_kwarg.keys()).issubset(set(autologging_kwarg.keys()))
+            for key in autologging_kwarg.keys():
+                _validate(autologging_kwarg[key], user_kwarg.get(key))
+        else:
+            assert autologging_kwarg is user_kwarg or autologging_kwarg == user_kwarg
```
dbczumar, line 479: There are several assumptions encoded in this logic that may need to be revisited (they should at least be documented):
- Patch code doesn't make copies of objects that can't be compared via object equality
- Patch code never removes any user-supplied arguments / keyword arguments
- If patch code modifies user-supplied arguments / keyword arguments, modifications must be additive (e.g., appending an element to a list or a new key/value pair to a dictionary)

mlflow/utils/autologging_utils.py
```diff
@@ -275,3 +282,202 @@ def batch_metrics_logger(run_id):
...
+_ATTRIBUTE_EXCEPTION_SAFE = "exception_safe"
+
+
+def exception_safe_function(function):
+    """
+    Wraps the specified function with broad exception handling to guard
+    against unexpected errors during autologging.
+    """
+    if _is_testing():
+        setattr(function, _ATTRIBUTE_EXCEPTION_SAFE, True)
```
dbczumar, line 389: Leverages function attributes from https://www.python.org/dev/peps/pep-0232/ for argument validation purposes

mlflow/utils/autologging_utils.py
```diff
@@ -119,6 +125,7 @@ def wrap_patch(destination, name, patch, settings=None):
 
     original = getattr(destination, name)
     wrapped = functools.wraps(original)(patch)
+    wrapped.__signature__ = inspect.signature(original)
```
dbczumar, line 128: `functools.wraps()` alone doesn't preserve the original signature of the function. For example, invoking `inspect.getargspec()` on a wrapped function yields the wrapper's signature, rather than the underlying function's signature. This may cause problems for frameworks that perform argument inspection; this concern has been raised by stakeholders.

mlflow/utils/autologging_utils.py
```diff
@@ -16,6 +18,10 @@
     "please ensure that autologging is enabled before constructing the dataset."
 )
 
+AUTOLOGGING_INTEGRATIONS = {}
```
aarondav, line 21:
```suggestion
# Dict mapping integration name to its config.
AUTOLOGGING_INTEGRATIONS = {}
```

mlflow/utils/autologging_utils.py
```diff
@@ -275,3 +282,202 @@ def batch_metrics_logger(run_id):
...
+                    failed_during_original = True
+                    raise
+
+            return wrapped_original(*og_args, **og_kwargs)
+
+        original = gorilla.get_original_attribute(destination, function_name)
+        call_original = functools.wraps(original)(call_original)
+        call_original.__signature__ = inspect.signature(original)
+
+        config = AUTOLOGGING_INTEGRATIONS[autologging_integration]
```
aarondav, line 352: What are the set of configurations on autolog? How many of them can we handle in a common manner, or is 'disable' the only one?

dbczumar: Currently, the common configurations are: `log_models`, `log_input_examples`, and `log_model_signatures`, and we intend to introduce `disable`. Of these configurations, `disable` seems like the easiest one to handle in a common manner. The other ones are trickier; we could provide utilities on top of `log_model` that respect these confs & have implementers use them, though this might be brittle.

mlflow/utils/autologging_utils.py
```diff
@@ -275,3 +282,202 @@ def batch_metrics_logger(run_id):
...
+                                    patch.
+    :param destination: The Python class on which the patch function is being defined.
+    :param function_name: The name of the function to patch on the specified `destination` class.
+    :param function: The patch function to apply. The first argument to this function should be
+                     reserved for an `original` method argument representing the underlying /
+                     original function. Subsequent arguments should be identical to those of the
+                     original function being patched.
+    """
+
+    def patched_train(*args, **kwargs):
```
dbczumar, line 323: `patched_train` is a misnomer here, since not all patched methods corresponding to training routines

mlflow/utils/autologging_utils.py
```diff
@@ -275,3 +282,202 @@ def batch_metrics_logger(run_id):
...
+        preexisting_run = mlflow.active_run()
+        original_result = None
+        called_original = False
+        failed_during_original = False
+
+        def call_original(*og_args, **og_kwargs):
```
dbczumar, line 329: As documented, user-defined patch functions are expected to accept an `original` method argument that can be used to call the original function. Here, we supply the original function with some additional error handling logic that can be used to determine whether or not an exception was raised by a call to the original function or, instead, by the patch code preamble / postamble.

mlflow/utils/autologging_utils.py
```diff
@@ -275,3 +282,202 @@ def batch_metrics_logger(run_id):
...
+            return original(*args, **kwargs)
+
+        try:
+            return function(call_original, *args, **kwargs)
+        except Exception as e:
+            if _is_testing():
+                raise
+
+            if not preexisting_run and mlflow.active_run():
+                try_mlflow_log(mlflow.end_run, RunStatus.to_string(RunStatus.FAILED))
```
dbczumar, line 363: Here, we attempt to clean up any dangling runs. As pointed out by @aarondav , it might be better to structure this as part of the patch API via an `on_exception` method defined on a patch class; this way, it's very clear to developers that teardown code is necessary & allows them to augment this teardown code with additional logic.

mlflow/utils/autologging_utils.py
```diff
@@ -119,6 +126,12 @@ def wrap_patch(destination, name, patch, settings=None):
 
     original = getattr(destination, name)
     wrapped = functools.wraps(original)(patch)
+    try:
+        wrapped.__signature__ = inspect.signature(original)
```
dbczumar, line 130: Unfortunately, this can fail (as it does when attempting to run it on `tensorflow.estimator.Estimator.exported_savedmodel` due to a 'hiddenattribute' concept in TensorFlow)

mlflow/tensorflow.py
```diff
@@ -975,41 +985,52 @@ def _log_early_stop_callback_metrics(callback, history):
                     last_epoch = len(history.history[metric_key])
                     try_mlflow_log(mlflow.log_metrics, restored_metrics, step=last_epoch)
 
-    def fit(self, *args, **kwargs):
-        with _manage_active_run():
-            original = gorilla.get_original_attribute(tensorflow.keras.Model, "fit")
 
-            unlogged_params = ["self", "x", "y", "callbacks", "validation_data", "verbose"]
+    class FitPatch(PatchFunction):
```
dbczumar, line 989: This is an incomplete integration with TensorFlow since the remaining functions use `wrap_patch` instead of the new `safe_patch`; it's intended to demonstrate the utility of `PatchFunction`'s `on_exception` class for performing cleanup operations (e.g., tensorboard directory removal) in the event of an unexpected exception in patch code.

cc @aarondav

mlflow/keras.py
```diff
@@ -759,26 +762,24 @@ def _run_and_log_function(self, original, args, kwargs, unlogged_params, callbac
 
         return history
 
-    def fit(self, *args, **kwargs):
-        original = gorilla.get_original_attribute(keras.Model, "fit")
+    def fit(original, self, *args, **kwargs):
```
aarondav, line 765: Consider having this sort of logic extracted out to the autolog wrapper:

```
        if not mlflow.active_run():
            auto_end_run = True
        else:
            auto_end_run = False
```

mlflow/keras.py
```diff
@@ -759,26 +762,24 @@ def _run_and_log_function(self, original, args, kwargs, unlogged_params, callbac
...
         To avoid unintentional creation of nested MLflow runs caused by a patched
         `fit_generator()` method calling a patched `fit()` method, we only patch
         `fit_generator()` in Keras < 2.4.0.
         """
-        original = gorilla.get_original_attribute(keras.Model, "fit_generator")
         unlogged_params = ["self", "generator", "callbacks", "validation_data", "verbose"]
         return _run_and_log_function(self, original, args, kwargs, unlogged_params, 4)
 
-    wrap_patch(keras.Model, "fit", fit)
+    safe_patch(FLAVOR_NAME, keras.Model, "fit", fit)
```
aarondav, line 779: May want to update these patterns: `try_mlflow_log(_log_early_stop_callback_params, callback)`

mlflow/sklearn/__init__.py
```diff
@@ -749,7 +752,7 @@ def fetch_logged_data(run_id):
             stacklevel=2,
         )
 
-    def fit_mlflow(self, clazz, func_name, *args, **kwargs):
+    def fit_mlflow(original, self, *args, **kwargs):
```
aarondav, line 755: Seems like we can replace the start_run/end_run logic with a `manage_run` kind of boolean on `safe_patch` for `patch_fit` safely, and improve the error logic.

mlflow/utils/autologging_utils.py
```diff
@@ -260,3 +292,377 @@ def batch_metrics_logger(run_id):
...
+    """
+
+    AUTOLOGGING_INTEGRATIONS[name] = {}
+
+    def wrapper(_autolog):
+        def autolog(*args, **kwargs):
+            AUTOLOGGING_INTEGRATIONS[name] = kwargs
+            _autolog(**kwargs)
+
+        wrapped_autolog = functools.wraps(_autolog)(autolog)
```
aarondav, line 313: Call update_wrapper_extended

mlflow/keras.py
dbczumar: Done!

mlflow/keras.py
```diff
@@ -754,31 +751,26 @@ def _run_and_log_function(self, original, args, kwargs, unlogged_params, callbac
...
         To avoid unintentional creation of nested MLflow runs caused by a patched
         `fit_generator()` method calling a patched `fit()` method, we only patch
         `fit_generator()` in Keras < 2.4.0.
         """
-        original = gorilla.get_original_attribute(keras.Model, "fit_generator")
         unlogged_params = ["self", "generator", "callbacks", "validation_data", "verbose"]
         return _run_and_log_function(self, original, args, kwargs, unlogged_params, 4)
 
-    wrap_patch(keras.Model, "fit", fit)
+    safe_patch(FLAVOR_NAME, keras.Model, "fit", fit, manage_run=True)
```
dbczumar, line 770: TODO (@dbczumar): Use `try_mlflow_log` in place of certain try/except statements in Keras autologging

mlflow/utils/autologging_utils.py
```diff
@@ -260,3 +292,409 @@ def batch_metrics_logger(run_id):
...
+        if type(inp) == list:
+            for item in inp:
+                _validate_new_input(item)
+        elif callable(inp):
+            assert getattr(inp, _ATTRIBUTE_EXCEPTION_SAFE, False), (
+                "New function argument '{}' passed to original function is not exception-safe."
+                " Please decorate the function with `exception_safe_function`.".format(inp)
+            )
+        elif inspect.isclass(type(inp)):
+            assert type(inp.__class__) == ExceptionSafeClass, (
```
harupy, line 649:
```suggestion
            assert inp.__class__ == ExceptionSafeClass, (
```

Is `type` necessary here?

```python
>>> class Foo:
        pass
>>> f = Foo()

>>> f.__class__
<class '__main__.Foo'>

>>> type(f.__class__)
<class 'type'>
```

harupy: Just never mind. `ExceptionSafeClass` is used as a meta class so we need `type`

mlflow/utils/autologging_utils.py
```diff
@@ -260,3 +292,409 @@ def batch_metrics_logger(run_id):
...
+    patch functions to broadly obey certain configurations (e.g., disable=True) without
+    requiring specific logic to be present in each autologging integration.
+    """
+
+    AUTOLOGGING_INTEGRATIONS[name] = {}
+
+    def wrapper(_autolog):
+        def autolog(**kwargs):
+            AUTOLOGGING_INTEGRATIONS[name] = kwargs
+            _autolog(**kwargs)
```
harupy, line 311:
```suggestion
    def wrapper(_autolog):
        def autolog(*args, **kwargs):
            AUTOLOGGING_INTEGRATIONS[name] = kwargs
            _autolog(*args, **kwargs)
```

do we need `*args` here?

harupy: or should we disallow passing positional arguments to `mlflow.<flavor name>.autolog` using [the keyword-only-arguments syntax](https://www.python.org/dev/peps/pep-3102/#:~:text=Keyword%2Donly%20arguments%20are%20not,therefore%20'required%20keyword'%20arguments.)?

dbczumar: I've disallowed passing positional arguments by introducing validation to `autologging_integration`. Arguments must either specify a default value or must be of type `KEYWORD_ONLY` (as determined by `inspect.signature`).

mlflow/utils/autologging_utils.py
```diff
@@ -260,3 +292,458 @@ def batch_metrics_logger(run_id):
...
+        return wrapped_autolog
+
+    return wrapper
+
+
+def get_autologging_config(flavor_name, config_key, default_value=None):
+    """
+    Returns a desired config value for a specified autologging integration.
+    Returns `None` if specified `flavor_name` has no recorded configs.
+    If `config_key` is not set on the config object, default vlaue is returned.
```
harupy, line 350:
```suggestion
    If `config_key` is not set on the config object, default value is returned.
```

mlflow/utils/autologging_utils.py
```diff
@@ -260,3 +292,458 @@ def batch_metrics_logger(run_id):
...
+
+        - Exceptions thrown from other parts of the patched implementation (`patch_function`)
+          are caught and logged as warnings.
+
+
+    :param autologging_integration: The name of the autologging integration associated with the
+                                    patch.
+    :param destination: The Python class on which the patch is being defined.
+    :param function_name: The name of the function to patch on the specified `destination` class.
+    :param function: The patched function code to apply. This is either a `PatchFunction` class
```
harupy, line 572:
```suggestion
    :param patch_function: The patched function code to apply. This is either a `PatchFunction` class
```

nit

tests/autologging/test_autologging_utils.py
```diff
@@ -396,3 +394,97 @@ def test_batch_metrics_logger_continues_if_log_batch_fails(
...
+        pass
+
+    # Try to use a falsy value that isn't "false"
+    def fn_bad_disable_conf_2(disable=0):
+        pass
+
+    for fn in [fn_missing_disable_conf, fn_bad_disable_conf_1, fn_bad_disable_conf_2]:
+        with pytest.raises(Exception) as exc:
+            autologging_integration("test")(fn)
+        assert "must specify a 'disable' argument" in str(exc)
```
harupy, line 434:
```suggestion
        with pytest.raises(Exception, match="must specify a 'disable' argument") as exc:
            autologging_integration("test")(fn)
```

Shoule we use `match` argument here?

tests/autologging/test_autologging_utils.py
```diff
@@ -396,3 +394,97 @@ def test_batch_metrics_logger_continues_if_log_batch_fails(
...
+        assert "Positional arguments are not allowed" in str(exc)
+
+    # Failure to apply the @autologging_integration decorator should not create a
+    # placeholder for configuration state
+    assert "test" not in AUTOLOGGING_INTEGRATIONS
+
+
+def test_get_autologging_config_returns_configured_values_or_defaults_as_expected():
+
+    assert get_autologging_config("nonexistent_integration", "foo") == None
```
harupy, line 450:
```suggestion
    assert get_autologging_config("nonexistent_integration", "foo") is None
```

nit

mlflow/utils/autologging_utils.py
dbczumar: Fixed! Thanks!

mlflow/utils/autologging_utils.py
dbczumar: Fixed! Thanks :)

tests/autologging/test_autologging_utils.py
dbczumar: Absolutely! I've replaced all instances of `... in str(exc)` with match; thanks so much for the suggestion!

tests/autologging/test_autologging_utils.py
dbczumar: Fixed!

dbczumar (commented):
@harupy Addressed some of your comments! I'm sure you'll have more :)

mlflow/utils/autologging_utils.py
dbczumar: @harupy Based on our offline discussion regarding backwards compatibility for supplying positional arguments to `autolog()` functions, I've introduced support for `args`. I've also added test coverage to ensure that args are reflected in the conf and are forwarded by the `autologging_integration` wrapper to the developer-specified `autolog()` function.

tests/statsmodels/test_statsmodels_model_export.py
```diff
@@ -1,11 +1,11 @@
 import pytest
 import numpy as np
 import pandas as pd
-import mock
 import os
 import yaml
 import json
 import pandas.testing
+from unittest import mock
```
dbczumar, line 8: looks like previous failures (e.g., https://github.com/mlflow/mlflow/runs/1535763935) were attributable to using `mock` via `import mock` instead of `from unittest import mock` in `test_autologging_safety_unit.py`. 
There's a documented incompatibility between `@mock.patch` when accessed via `import mock` and pytest described here: https://stackoverflow.com/questions/39169563/pytest-works-with-old-mock-but-not-unittest-mock.

For safety going forward, I've converted the only other instance of `import mock` in our statsmodels flavor tests to `from unittest import mock`.

mlflow/utils/autologging_utils.py
```diff
@@ -260,3 +292,448 @@ def batch_metrics_logger(run_id):
...
+            _validate_new_input(autologging_call_input)
+            return
+
+        assert type(autologging_call_input) == type(
+            user_call_input
+        ), "Type of input to original function '{}' does not match expected type '{}'".format(
+            type(autologging_call_input), type(user_call_input)
+        )
+
+        if type(autologging_call_input) == list:
```
harupy, line 711:
```suggestion
        if type(autologging_call_input) == tuple:
```

Should we use `tuple` here?

```python
def foo(*args):
    print(type(args))

foo("a", "b")  # -> <class 'tuple'>
```

tests/autologging/test_autologging_utils.py
```diff
@@ -396,3 +394,99 @@ def test_batch_metrics_logger_continues_if_log_batch_fails(
...
+    assert get_autologging_config("nonexistent_integration", "foo") is None
+
+    @autologging_integration("test_integration_for_config")
+    def autolog(foo="bar", t=7, disable=False):
+        pass
+
+    # Before `autolog()` has been invoked, config values should not be available
+    assert get_autologging_config("test_integration_for_config", "foo") is None
+    assert get_autologging_config("test_integration_for_config", "disable") is None
+    assert get_autologging_config("test_integration_for_config", "t", 10) is 10
```
harupy, line 461:
```suggestion
    assert get_autologging_config("test_integration_for_config", "t", 10) == 10
```

tests/autologging/test_autologging_utils.py
```diff
@@ -396,3 +394,99 @@ def test_batch_metrics_logger_continues_if_log_batch_fails(
...
+
+    autolog()
+
+    assert get_autologging_config("test_integration_for_config", "foo") == "bar"
+    assert get_autologging_config("test_integration_for_config", "disable") is False
+    assert get_autologging_config("test_integration_for_config", "t", 10) is 7
```
harupy, line 467:
```suggestion
    assert get_autologging_config("test_integration_for_config", "t", 10) == 7
```

mlflow/utils/autologging_utils.py
```diff
@@ -66,10 +74,33 @@ def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W01
...
+             & signature as `wrapped`.
+    """
+    updated_wrapper = functools.update_wrapper(wrapper, wrapped)
+    # Assign the signature of the `wrapped` function to the updated wrapper function.
+    # Certain frameworks may disallow signature inspection, causing `inspect.signature()` to throw.
+    # One such example is the `tensorflow.estimator.Estimator.export_savedmodel()` function
+    try:
+        updated_wrapper.__signature__ = inspect.signature(wrapped)
+    except Exception:  # pylint: disable=broad-except
+        _logger.debug("Failed to restore original signature for wrapper around %s", wrapped)
```
harupy, line 93:
```suggestion
        _logger.warning("Failed to restore original signature for wrapper around %s", wrapped)
```

Should we use `warning` here?

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -0,0 +1,827 @@
...
+
+    with mock.patch("mlflow.utils.autologging_utils._logger.warning") as logger_mock:
+        throwing_function()
+        assert logger_mock.call_count == 1
+        message, formatting_arg = logger_mock.call_args[0]
+        assert "unexpected error during autologging" in message
+        assert formatting_arg == exc_to_throw
+
+
+def test_exception_safe_function_exhibits_expected_behavior_in_test_mode(test_mode_on,):
```
harupy, line 431:
```suggestion
@pytest.mark.usefixtures(test_mode_on.__name__)
def test_exception_safe_function_exhibits_expected_behavior_in_test_mode():
```

Can we use `pytest.mark.usefixtures` here to avoid adding an unused argument?

dbczumar: Absolutely! Thanks for teaching me about this. I've swapped in `usefixtures` in place of unused fixture arguments

mlflow/utils/autologging_utils.py
dbczumar: Because we expect this to happen in some cases (e.g., when patching `tf.estimator.Estimator.export_savedmodel()`), I think it's best not to emit a warning to users here. The debug log should help us for debugging purposes later on.

mlflow/utils/autologging_utils.py
dbczumar: Yes! this was an oversight on my part. I've added tuple support and updated the associated tests to test against tuple input types! Thanks so much!

tests/autologging/test_autologging_utils.py
dbczumar: Done!

tests/autologging/test_autologging_utils.py
dbczumar: Done!

dbczumar (commented):
@harupy Addressed your latest round of comments!

harupy (approved):
LGTM!

dbczumar:
Thanks for the review @harupy ! Merging!

PR #3687, opened 2020-11-13
[MLflowSkinny] Base changes for MLflow Skinny Client

eddiestudies:

Adding base changes for MLflow to run with all but numpy, pandas, and sqlparse dependencies removed. 

## How is this patch tested?

Runs a subset of existing tests to unblock parallel development streams on mlflow skinny client

Components 

Interface 

Language 

Integrations


tests/resources/mlflow-test-plugin/setup.py
```diff
@@ -8,7 +8,7 @@
     packages=find_packages(),
     # Require MLflow as a dependency of the plugin, so that plugin users can simply install
     # the plugin & then immediately use it with MLflow
-    install_requires=["mlflow"],
+    install_requires=[],  # Remove mlflow dependency, it can cause tests to run against pypi
```
dbczumar, line 11: @harupy @smurching Didn't this come up in the context of another ticket? Can you comment on how to best resolve it?

smurching: It did yeah - we were confused why e.g. `gorilla` was still being installed in our test environment despite the fact that we'd removed it as a dependency from `setup.py`, e.g. in https://github.com/mlflow/mlflow/pull/3666/checks?check_run_id=1382567094#step:3:308. It turns out in our dependency installation scripts for Github actions, we install this test plugin and then install MLflow from source - because we install the test plugin first, we end up installing `mlflow` from pypi as a transitive dependency of the plugin, and then overwriting that installation with MLflow from source.

The change here avoids that problem by removing `mlflow` as a dependency from the test plugin. This seems ok as we ought to only use the test plugin in our CI - in practice though, we reference the test plugin in our [docs on writing plugins](https://mlflow.org/docs/latest/plugins.html#defining-a-plugin). 

To get around this we could either:
a) Define another example plugin to reference in our docs & add tests verifying that it works (seems like more work)
b) See if we can do something like `install_requires=[os.path.join(".", os.pardir, os.pardir, os.pardir)]` to depend on MLflow installed from source (reduces the value of the test plugin as an example)
c) Just update our install-common-deps.sh CI script to [install MLflow from source](https://github.com/mlflow/mlflow/blob/eda5f435c65d8e7ceab4d4ba82e7e18b2e2cdef3/dev/install-common-deps.sh#L51) before we [install other Python dependencies](https://github.com/mlflow/mlflow/blob/eda5f435c65d8e7ceab4d4ba82e7e18b2e2cdef3/dev/install-common-deps.sh#L33-L49), including the test plugin

smurching: IMO option a) is maybe the best actual approach, but I'd be fine with option c) too; the risk is that one of our test requirements has a strict version dependency on `mlflow` and ends up overwriting our source installation of `mlflow`, but this seems unlikely / we could potentially test for this

mlflow/deployments/cli.py
```diff
@@ -237,6 +236,8 @@ def run_local(flavor, model_uri, target, name, config):
 
 
 def predictions_to_json(raw_predictions, output):
+    from mlflow.pyfunc.scoring_server import _get_jsonable_obj
```
dbczumar, line 239: This is kind of a bummer, as it seems to imply that our deployments CLI will never support the `predict()` function unless the complete MLflow platform is installed. I think the hope was that users could install deployment plugins alongside the MLflow skinny client and continue to leverage them without also having to fetch the full platform.

Can we factor out `get_jsonable_obj` into a utility that's independent of `mlflow.pyfunc.scoring_server` and introduce some skinny client test coverage for the use of deployment plugins? E.g. I think we can run this: https://github.com/mlflow/mlflow/blob/master/tests/test_cli.py

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -5,9 +5,6 @@
 import requests
```
dbczumar, line 5: The skinny client will eventually need to support `databricks_artifact_repo` for compatibility with Databricks. Leaving this here as a note for myself to remove the dependency on `azure.storage.blob.BlobClient` for this store. We've done this already for a different compatibility purpose on Databricks, so it should be relatively quick to apply this change.

tests/resources/mlflow-test-plugin/setup.py
dbczumar: As discussed offline with @smurching , I think the concern here is that depending on `mlflow` (even a source-installed copy) will bring in dependencies that aren't included by the skinny client, degrading the utility of our skinny client tests.

As a result, I think it's best for us to adopt option (a) here. @eedeleon Is this something you can do as part of this PR? If not, I can pitch in.

dbczumar (changes requested):
@eedeleon Sorry for the delay here; I left a few comments. This is close!

tests/resources/mlflow-test-plugin/setup.py
eedeleon: I think we could implement a like b.

requirements = [] if MLFLOW_TEST_ENV in os.environ else ["mlflow"]

eedeleon: We can add a test that verifies MLFLOW_TEST_ENV(we can use any env var) is set as well

dbczumar: That `os.environ` code will show up in our example reference too, though, which we probably don't want.

dbczumar: > We can add a test that verifies MLFLOW_TEST_ENV(we can use any env var) is set as well

Agreed!

tests/test_skinny.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+import os
+
+
+@pytest.mark.skipif(
+    "MLFLOW_SKINNY" not in os.environ, reason="This test is only valid for the skinny client"
+)
+def test_fails_sqlalchemy_import():
```
dbczumar, line 8: Are there any other beefy libraries that we should assert aren't importable here?

eedeleon: flask

mlflow/deployments/cli.py
eedeleon: Refactoring it is possible but it won't allow us to run the test_cli.py given it's dependence on flask.

__init__.py also depends on flask and would cause an error.

This would require more changes than move get_jsonable_obj to have a validated test.

Where would we want to move get_jsonable_obj?

dbczumar: Sounds like we can test everything in `test_cli` except for `test_run_local`, right? In that case, perhaps we can move `test_run_local` into its own test suite?

I think `proto_json_utils` seems like a pretty reasonable place to define the helper function.

mlflow/azureml/__init__.py
```diff
@@ -611,7 +611,12 @@ def _create_mlflow_wheel(mlflow_dir, out_dir):
 
 from azureml.core.model import Model
 from mlflow.pyfunc import load_model
-from mlflow.pyfunc.scoring_server import parse_json_input, _get_jsonable_obj
+from mlflow.pyfunc.scoring_server import parse_json_input
+
+try:
+    from mlflow.utils.proto_json_utils import _get_jsonable_obj
+except ImportError:  # Back Compat for https://github.com/mlflow/mlflow/pull/3687
+    from mlflow.pyfunc.scoring_server import _get_jsonable_obj
```
dbczumar, line 619: Is the concern here that we might deploy an MLflow model to AzureML and install an older version of MLflow when creating the serving container Docker image? The deployment procedure seems to ensure that we're pulling in the latest MLflow version (https://github.com/mlflow/mlflow/blob/aaed8043dd9d206a8090a480493118be9ba4240e/mlflow/azureml/__init__.py#L554-L561), so I think it's okay to remove the try/except here.

Let me know if I'm missing something here.

setup.py
```diff
@@ -25,38 +29,59 @@ def package_files(directory):
...
+    "pandas",
+    "python-dateutil",
+    "pyyaml",
+    "protobuf>=3.6.0",
+    "requests>=2.17.3",
+    "six>=1.10.0",
+    "simplejson",
+]
+
+_is_mlflow_skinny = bool(os.environ.get(_MLFLOW_SKINNY_ENV_VAR))
```
dbczumar, line 54: Nit: Can we move this down to line 78-ish (right above `setup()`). The var naming + comments look awesome - thanks so much for the investment there!

tests/deployments/test_cli.py
```diff
@@ -91,13 +91,3 @@ def test_target_help():
     runner = CliRunner()
     res = runner.invoke(cli.target_help, ["--target", f_target])
     assert "Target help is called" in res.stdout
-
-
-def test_run_local():
```
dbczumar, line 96: Sorry, I think I got this wrong. This `run_local` doesn't seem like it actually starts a flask server. Can we revert this move?

dev/run-python-skinny-tests.sh
```diff
@@ -0,0 +1,16 @@
...
+err=0
+trap 'err=1' ERR
+export MLFLOW_SKINNY='true'
+
+pytest --verbose tests/test_skinny.py
+python -m pip install sqlalchemy alembic sqlparse
+pytest --verbose tests/tracking/test_client.py
+pytest --verbose tests/tracking/test_tracking.py
+pytest --verbose tests/projects/test_projects.py
+pytest --verbose tests/deployments/test_cli.py
```
dbczumar, line 14: Can we add `test_deployments` as well? Looks like that one doesn't have any non-skinny dependencies either.

setup.py
```diff
@@ -25,38 +29,59 @@ def package_files(directory):
...
+    # Required to run the MLflow server against SQL-backed storage
+    "sqlalchemy<=1.3.13",
+    "waitress; platform_system == 'Windows'",
+]
+
 setup(
     name="mlflow",
     version=version,
     packages=find_packages(exclude=["tests", "tests.*"]),
     package_data={"mlflow": js_files + models_container_server_files + alembic_files},
```
dbczumar, line 83: I don't think we should embed the UI JS, model container server files, or alembic files with the skinny client. Can we condition adding these files on `_is_mlflow_skinny` and omit them if skinny is True?

dbczumar: Hmm...looks like these files are included independently of `package_data`. I think `find_packages` picks them up. Let's follow up here in a future PR.

dbczumar (changes requested):
@eedeleon This is really close. I left a few more comments.

The most significant thing I noticed while trying this out is that attempting to run the MLflow CLI fails with an alembic import error:

```
import mlflow.cli
ModuleNotFoundError: No module named 'alembic'
```

Ideally, we should provide test coverage for the MLflow CLI. I noticed that `test_cli` aggregates test cases for `mlflow server`, `mlflow ui`, `mlflow gc`, and `mlflow run`. Perhaps we could split this file up: 1. add the `mlflow server` / `mlflow ui` /`mlflow gc` tests to a new `test_cli` suite in the `tests/server` subdirectory & 2. add the `mlflow run` tests to `test_projects_cli`.

Otherwise, I've confirmed that the following workflows are supported by the skinny client on Databricks:

- Tracking operations (logging / loading / searching params, metrics, tags + logging / loading artifacts on AWS - Azure requires removal of the Azure Storage Blob dependency as a follow-up PR)

- Model registration, search, artifact loading, transitions

- Execution of GitHub projects within a Databricks notebook & against a Databricks workspace remotely

- After installing scikit-learn, sklearn autologging + model log / load operations succeed

dev/run-python-skinny-tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
dbczumar, line 1: Ditto here - can we add a small README at the top of this file explaining the purpose of skinny client testing?

dbczumar (approved):
@eedeleon LGTM once a few small comments have been addressed! Thanks so much for this contribution!

Here's a TODO list of follow-up tasks I've put together; ones necessary prior to the initial skinny client release are tagged `[Required for 1.14]`:

- [Required for 1.14] Figure out how to distribute the client. E.g., add to MLflow release CLI scaffolding maintained by Databricks

- [Required for 1.14] Document the skinny client on `mlflow.org` / in the GitHub repository (what does the skinny client do and what does it not do)

- [Required for 1.14] Drop azure-storage-blob requirement for artifact logging on Azure Databricks

- Drop pandas and numpy requirements

- Test restructure: https://github.com/mlflow/mlflow/pull/3687/files#r528014570

dev/run-python-skinny-tests.sh
```diff
@@ -0,0 +1,27 @@
...
+# A different example client/store with a minimal dependency footprint could also work for this purpose.
+
+set -x
+# Set err=1 if any commands exit with non-zero status as described in
+# https://stackoverflow.com/a/42219754
+err=0
+trap 'err=1' ERR
+export MLFLOW_SKINNY='true'
+
+python -m pip uninstall sqlalchemy alembic sqlparse -y
```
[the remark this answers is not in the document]
eedeleon: sqlalchemy was recently added to small-requirements.txt.

I can create a skinny-requirements.txt and use that to avoid further issues but it seemed like overkill.

Any suggestions?

dev/install-common-deps.sh
```diff
@@ -48,8 +55,10 @@ if [[ "$INSTALL_LARGE_PYTHON_DEPS" == "true" ]]; then
   ls -lha $(find $CONDA_DIR/envs/test-environment/ -path "*bin/spark-*")
 fi
 
-pip install .[extras]
-export MLFLOW_HOME=$(pwd)
+# Skinny Client purposefully removes the sqlalchemy dependency
+if [[ "$MLFLOW_SKINNY" == "true" ]]; then
+  pip uninstall sqlalchemy -y
+fi
```
dbczumar, line 61: @eedeleon How do we know that installing MLflow Skinny didn't pull in `sqlalchemy` independently of the Python small dependencies? We could change the ordering here so that the MLflow client is installed after the small dependencies & removal of `sqlalchemy`, but I'd rather create a `skinny-requirements` file rather than rely on subtle ordering properties / environment-variable-based branching logic in `install-common-deps`.

eedeleon: I made the change. There is some overlap between small-requirements and large requirements, but I agree it is worth it since this approach is definitely more explicit and less likely to leak bugs.

PR #3700, opened 2020-11-16
Fix R client for windows

lorenzwalthert:
Because `fs::path_abs()` retuns `D:/x/y/y`, whereas on unix, it returns `/User/x/y/y`. So on windows, paste("file://", fs::path_abs(...)) returns an invalid URI scheme and lacks a slash. 
In Python:
```python 
import mlflow; print(mlflow.get_tracking_uri())
#> file:///D:/a/mlflow-test-3206/mlflow-test-3206/mlruns
```

In R:
```r
# first fixing the path to the executable
Sys.setenv(MLFLOW_BIN = "C:\\Miniconda\\envs\\r-mlflow-1.12.0\\Scripts\\mlflow.exe")
mlflow:::mlflow_get_tracking_uri()
#> file://D:/a/mlflow-test-3206/mlflow-test-3206/mlruns
```

Log: https://github.com/lorenzwalthert/mlflow-test-3206/runs/1405857066?check_suite_focus=true#step:9:11

https://en.wikipedia.org/wiki/File_URI_scheme

### DIFF
--- a/mlflow/R/mlflow/R/tracking-client.R
+++ b/mlflow/R/mlflow/R/tracking-client.R
@@ -4,6 +4,7 @@ new_mlflow_client <- function(tracking_uri) {
 }
 
 new_mlflow_uri <- function(raw_uri) {
+  # fix file:///X:/D case
   # Special case 'databricks'
   if (identical(raw_uri, "databricks")) {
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
Ups wrong target branch. Will come back later...

PR #3705, opened 2020-11-17
Saving and Loading pytorch model as state dict

shrinath-suresh:

The current implementation of `mlflow.pytorch` only supports for saving the entire model into mlflow. Adding support for saving and loading the model using state dict.

Instead of storing the entire model into mlflow, when the model state dicts are saved, the size of the model is reduced to a greater extent - which would be helpful during the deployment of the model.

https://github.com/mlflow/mlflow/pull/3408 - Please read through the discussion points on the PR . It would be helpful for the future use cases as mentioned above.

Implementation Details:

Adding two new methods to mlflow.pytorch - `load_state_dict` and `save_state_dict` for loading and saving the pytorch models. And also added a key `state_dict` under `pytorch:flavor`. By default(for entire model) the key will be set to `false` . Only when the model is saved/logged as state dict, the key would be set to `true`.

Sample screenshot given below

image

## How is this patch tested?

Tested by saving/loading the model as both state dict and entire version. Working on the Unit tests. 

Components 

Interface 

Language 

Integrations


mlflow/pytorch/__init__.py
```diff
@@ -571,6 +621,8 @@ def load_model(model_uri, **kwargs):
             torch.__version__,
         )
     torch_model_artifacts_path = os.path.join(local_model_path, pytorch_conf["model_data"])
+    if load_state_dict:
+        return torch_model_artifacts_path
```
harupy, line 625: @shrinath-suresh Thanks for the PR! With the current approach, we can't reconstruct the model when we serve it with `mlflow models serve`. I'm wondering if we should log a model class (as a python script?) along with a model.

shrinath-suresh: Logging model class would work only for the smaller models. In case of complex models, there will be multiple nested python scripts. It would be better to leave it to the user for loading the  model class.

harupy: Can we serialize a pytorch model **class**?

shrinath-suresh: Input to the [mlflow.pytorch.load_state_dict](https://github.com/chauhang/mlflow/blob/749c99f0158dd57a48432d5e8ed0ad498944d8cf/mlflow/pytorch/__init__.py#L539) is the **object** of the model class (not the model **class** name)
 
By keeping model object as input, user has more control over creating the object from the model class (For example: constructor arguments).

harupy: Ah we need to log constructor arguments to instantiate the model properly from state_di.

```
# To reconstruct this model from state dict, we need to log `a` and `b` along with `MyModel`.
model = MyModel(a=1, b=2)
```

shrinath-suresh: @harupy We anticipated the same and that is the reason we have set the argument for `load_state_dict` as  **the model object (not the model class).** Its upto the user to create the model object and call `load_state_dict` method

harupy: Can we fix `load_state_dict` to just return a state dict because the name `load_state_dict` sounds like it returns a state dict? The workflow would look like:

```python
model = Model(...)
model.load_state_dict(mlflow.pytorch.load_state_dict(...))
```

mlflow/pytorch/__init__.py
```diff
@@ -79,6 +79,7 @@ def log_model(
     await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
     requirements_file=None,
     extra_files=None,
+    save_as_state_dict=False,
```
harupy, line 82: Can we avoid introducing this boolean parameter and provide a separate method (e.g. `log_state_dict`) isntead?

mlflow/pytorch/__init__.py
```diff
@@ -516,7 +536,37 @@ def _load_model(path, **kwargs):
...
+
+    model_path = load_model(model_uri, load_state_dict=True)
+    model_path = os.path.join(model_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
+
+    state_dict = torch.load(model_path)
+    model_class_obj.load_state_dict(state_dict)
+    return model_class_obj
+
+
+def load_model(model_uri, load_state_dict=False, **kwargs):
```
harupy, line 569: Can we avoid introducing `load_state_dict` here and just provide `load_state_dict` instead?

mlflow/pytorch/__init__.py
shrinath-suresh: Thanks for the comments. i have updated the code to return state dict.

mlflow/pytorch/__init__.py
shrinath-suresh: created new method `log_state_dict`

mlflow/pytorch/__init__.py
shrinath-suresh: remove the boolean parameter. Refactored the existing `load_model` method to re-use most of the code in `load_state_dict`

mlflow/pytorch/__init__.py
```diff
@@ -240,6 +240,194 @@ def forward(self, x):
     )
 
 
+def log_state_dict(
```
shrinath-suresh, line 243: @harupy @chauhang 

`log_model` and `log_state_dict` uses `Model.log` to log the model inside mlflow. I have retained all the parameters which were there in `log_model` since there is no much code change. It would help us in some of the scenarios like, logging extra artifacts or registering the model when user logs the model as state dict. 

If you feel, we can reduce the argument and keep only few, please let me know.

mlflow/pytorch/__init__.py
```diff
@@ -536,6 +755,44 @@ def load_model(model_uri, **kwargs):
...
+        run_id = "96771d893a5e46159d9f3b49bf9013e2"
+        pytorch_model = mlflow.pytorch.load_model("runs:/" + run_id + "/" + model_path_dir)
+        y_pred = pytorch_model(x_new_data)
+    """
+
+    torch_model_artifacts_path = _get_model_artifact_path(model_uri)
+    return _load_model(path=torch_model_artifacts_path, **kwargs)
+
+
+def _get_model_artifact_path(model_uri):
```
shrinath-suresh, line 775: I have refactored the load_model into two parts, so that we dont duplicate the code in `load_state_dict`

mlflow/pytorch/__init__.py
```diff
@@ -251,6 +439,7 @@ def save_model(
     input_example: ModelInputExample = None,
     requirements_file=None,
     extra_files=None,
+    save_as_state_dict=False,
```
shrinath-suresh, line 442: The code in save state dict is too clumsy to reuse. If we need to get rid of this boolean parameter and have an independent method as `save_stat_dict`, we have to split the `save_model` into three parts.
  **Where the condition checks are done and model path is derived**
https://github.com/mlflow/mlflow/blob/284742591ae3a7cd05ea03159ffa9d9ebe905163/mlflow/pytorch/__init__.py#L365-L401 

**Actual model saving part using torch.save**
https://github.com/mlflow/mlflow/blob/284742591ae3a7cd05ea03159ffa9d9ebe905163/mlflow/pytorch/__init__.py#L402-L405 



**post tasks like saving extra files, updating conda.yaml etc.**
https://github.com/mlflow/mlflow/blob/284742591ae3a7cd05ea03159ffa9d9ebe905163/mlflow/pytorch/__init__.py#L407-L467

shrinath-suresh: @harupy @chauhang Please let me know if we need to split the method and re-use the code or we can have the boolean variable action as above.

harupy:
@shrinath-suresh Thanks for the updates! btw I have a question about this. 

> Instead of storing the entire model into mlflow, when the model state dicts are saved, the size of the model is reduced to a greater extent - which would be helpful during the deployment of the model.

I wrote a simple script to verify this behavior. 


```python
import os
import torch
from torchvision import models
import shutil
import subprocess


SAVE_DIR = "foo"

if os.path.exists(SAVE_DIR):
    shutil.rmtree(SAVE_DIR)

os.makedirs(SAVE_DIR)


model = models.resnet50(pretrained=True)
torch.save(model, f"{SAVE_DIR}/model.pt")
torch.save(model.state_dict(), f"{SAVE_DIR}/state_dict.pt")

print(subprocess.check_output(["ls", "-lh", SAVE_DIR]).decode("utf-8"))
```

output

```
total 400624
-rw-r--r--  1 harutakakawamura  staff    98M Nov 29 21:56 model.pt
-rw-r--r--  1 harutakakawamura  staff    98M Nov 29 21:56 state_dict.pt
                                         ^^^
                                         almost equal
```

The difference between `torch.save(model.state_dict(), ...)` and `torch.save(model, ...)` is very small. Am I missing something?

harupy:
WIP

shrinath-suresh:
> @shrinath-suresh Thanks for the updates! btw I have a question about whta you mentioned about saved model size.
> 
> > Instead of storing the entire model into mlflow, when the model state dicts are saved, the size of the model is reduced to a greater extent - which would be helpful during the deployment of the model.
> 
> I wrote a simple script to verify this behavior.
> 
> ```python
> import os
> import torch
> from torchvision import models
> import shutil
> import subprocess
> 
> 
> SAVE_DIR = "foo"
> 
> if os.path.exists(SAVE_DIR):
>     shutil.rmtree(SAVE_DIR)
> 
> os.makedirs(SAVE_DIR)
> 
> 
> model = models.resnet50(pretrained=True)
> torch.save(model, f"{SAVE_DIR}/model.pt")
> torch.save(model.state_dict(), f"{SAVE_DIR}/state_dict.pt")
> 
> print(subprocess.check_output(["ls", "-lh", SAVE_DIR]).decode("utf-8"))
> ```
> 
> output
> 
> ```
> total 400624
> -rw-r--r--  1 harutakakawamura  staff    98M Nov 29 21:56 model.pt
> -rw-r--r--  1 harutakakawamura  staff    98M Nov 29 21:56 state_dict.pt
>                                          ^^^
>                                          almost equal
> ```
> 
> The difference between `torch.save(model.state_dict(), ...)` and `torch.save(model, ...)` is very small. Am I missing something?

My observation is from [MNIST](https://github.com/mlflow/mlflow/blob/master/examples/pytorch/MNIST/example2/mnist_autolog_example2.py) example. I ran 10 epochs and here is the result of full model and state dict

```
-rw-rw-r--  1 ubuntu ubuntu 100M Nov 30 12:37 full_model.pth
-rw-rw-r--  1 ubuntu ubuntu 534K Nov 30 12:37 state_dict.pth
```

harupy:
@shrinath-suresh I think this is because we save `trainer.model` (which is a `pl.LightningModule` object) in the MNIST example.

shrinath-suresh:
> @shrinath-suresh This is probably because `mlflow.pytorch.autolog` saves `trainer.model` (which is a `pl.LightningModule` object). `pl.LightningModule` has many attributes that `torch.nn.Module` doesn't have. These attributes increase the saved model size.



> @shrinath-suresh This is probably because `mlflow.pytorch.autolog` saves `trainer.model` (which is a `pl.LightningModule` object). `pl.LightningModule` has many attributes that `torch.nn.Module` doesn't have. These attributes increase the saved model size.


You are right. Same mnist example with pytorch shows same size for both state dict and entire model. We can take this discussion separate, as this PR has no dependency `mlflow.pytorch.autolog`.

mlflow/pytorch/__init__.py
```diff
@@ -437,10 +557,21 @@ class LinearNNModel(torch.nn.Module):
...
+    if isinstance(pytorch_model, OrderedDict):
+        save_state_dict(
+            state_dict=pytorch_model,
+            path=path,
+            mlflow_model=mlflow_model,
+            pickle_module=pickle_module,
+            **kwargs,
+        )
+        return
+
```
harupy, line 574: Is this for allowing `save_model` to log `state_dict`?

shrinath-suresh: yes @harupy

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +285,106 @@ def gen_data():
     )
 
 
+def log_state_dict(state_dict, artifact_path, pickle_module=None, **kwargs):
+    """
+    Log a PyTorch model as an MLflow artifact for the current run.
+
```
harupy, line 291:
```suggestion
```

Can we remove these blank lines?

shrinath-suresh: sure . will remove it

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +285,106 @@ def gen_data():
...
+    model_data_subpath = "data"
+    model_data_path = os.path.join(path, model_data_subpath)
+    os.makedirs(model_data_path)
+
+    pickle_module_path = os.path.join(model_data_path, _PICKLE_MODULE_INFO_FILE_NAME)
+    with open(pickle_module_path, "w") as f:
+        f.write(pickle_module.__name__)
+
+    model_path = os.path.join(model_data_path, _SERIALIZED_TORCH_MODEL_FILE_NAME)
```
harupy, line 384: A uesr might log a state_dict that represents a checkpoint for inference and/or resuming training ([this use case](https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-a-general-checkpoint-for-inference-and-or-resuming-training)). In this case `_SERIALIZED_TORCH_MODEL_FILE_NAME` (= `"model.pth"`) doesn't seem to be the right name because it's not a model.

harupy: Maybe `state_dict.pth` is better?

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +285,106 @@ def gen_data():
     )
 
 
+def log_state_dict(state_dict, artifact_path, pickle_module=None, **kwargs):
```
harupy, line 288: Can you mark these APIs as `experimental` so we can modify them later if necessary?

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +285,106 @@ def gen_data():
...
+    """
+    Log a PyTorch model as an MLflow artifact for the current run.
+
+    :param state_dict: PyTorch model state dict to be saved.
```
harupy, line 292: [Question about TorchServe] I think there are two types of state dicts.

1. state_dict that represents a model:

```python
model.state_dict()
```

2. state_dict that represents a checkpoint (or might something else):

```python
{
   "model": model.state_dict(),
   "optim": optimiser.state_dict(),
   ...
}
```

Can we serve the checkpoint state dict with TorchServe?

mlflow/pytorch/__init__.py
shrinath-suresh: renamed it to `state_dict.pth`

mlflow/pytorch/__init__.py
shrinath-suresh: marked methods as experimental

mlflow/pytorch/__init__.py
shrinath-suresh: We need to save it as `model.state_dict()`. confirmed it with @gchauhang .

mlflow/pytorch/__init__.py
```diff
@@ -530,6 +635,7 @@ class LinearNNModel(torch.nn.Module):
         model_data=model_data_subpath,
         pytorch_version=torch.__version__,
         **torchserve_artifacts_config,
+        state_dict=False,
```
harupy, line 638: Can we remove this field?

shrinath-suresh: Yes. its no longer needed. will remove it right away

shrinath-suresh:
looking into the linting errors

mlflow/pytorch/__init__.py
```diff
@@ -591,6 +696,34 @@ def _load_model(path, **kwargs):
             return torch.jit.load(model_path)
 
 
+@experimental
+def load_state_dict(model_uri):
+    """
+    Load a PyTorch model state dict from a local file or a run.
```
harupy, line 702: Can we simply say `Load a state dict ...` because the state dict might not be a model?

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -738,6 +773,39 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
     )
 
 
+@pytest.mark.large
+def test_load_model_succeeds_when_data_is_model_file_instead_of_directory_saved_as_state_dict(
+    module_scoped_subclassed_model, model_path, data
```
harupy, line 778: Can we simply use `sequential_model` here?

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -694,6 +694,41 @@ def test_load_pyfunc_succeeds_when_data_is_model_file_instead_of_directory(
...
+):
+
+    mlflow.pytorch.save_state_dict(
+        state_dict=module_scoped_subclassed_model.state_dict(), path=model_path
+    )
+
+    model_class = ModuleScopedSubclassedModel()
+    model_state_dict = mlflow.pytorch.load_state_dict(model_path)
+    model_class.load_state_dict(model_state_dict)
+    model_class.eval()
```
harupy, line 709: Can we check the equality of the state dics before/after logging?

```python
state_dict = model.state_dict()
# log
loaded_state_dict = mlflow.pytorch.load_state_dict(...)

compare_state_dict(state_dict, loaded_state_dict)
```

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +287,108 @@ def gen_data():
...
+        save_state_dict(
+            state_dict=state_dict, path=local_path, pickle_module=pickle_module, **kwargs
+        )
+        mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
+
+
+@experimental
+def save_state_dict(state_dict, path, pickle_module=None, **kwargs):
+    """
+    Save the model as state dict to a path on the local file system
```
harupy, line 368: Should we avoid using `model` in `xxx_state_dict` APIs because `model` indicates these APIs expect a model state dict but they actually don't?

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +287,108 @@ def gen_data():
...
+            scripted_pytorch_model = torch.jit.script(model)
+            mlflow.pytorch.log_state_dict(scripted_pytorch_model, "models")
+    """
+    pickle_module = pickle_module or mlflow_pytorch_pickle_module
+
+    with TempDir() as tmp:
+        local_path = tmp.path("model")
+        save_state_dict(
+            state_dict=state_dict, path=local_path, pickle_module=pickle_module, **kwargs
+        )
```
harupy, line 361: I'm wondering if we need to `pickle_module` when logging a state dict. I think the reason that `torch.save` provides this argment is to allow users to use clouldpickle or dill which mitigates issues that pickle has when saving the entire model object. I'm investigating if the same issue still persists when saving a state_dict.

harupy: cc @smurching

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -738,6 +758,40 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+@pytest.mark.parametrize("scripted_model", [True, False])
+def test_load_model_succeeds_when_data_is_model_file_instead_of_directory_saved_as_state_dict(
+    sequential_model, model_path, data
+):
+    """
+    This test verifies that PyTorch models saved in older versions of MLflow are loaded successfully
+    by ``mlflow.pytorch.load_model``. The ``data`` path associated with these older models is
+    serialized PyTorch model file, as opposed to the current format: a directory containing a
+    serialized model file and pickle module information.
+    """
```
harupy, line 771: Is this comment related to this test?

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -738,6 +758,40 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+    artifact_path = "pytorch_model"
+    with mlflow.start_run():
+        mlflow.pytorch.log_state_dict(
+            artifact_path=artifact_path, state_dict=sequential_model.state_dict()
+        )
+        model_path = _download_artifact_from_uri(
+            "runs:/{run_id}/{artifact_path}".format(
+                run_id=mlflow.active_run().info.run_id, artifact_path=artifact_path
+            )
+        )
```
harupy, line 781: This part can be simplified to:

```suggestion
        model_path = mlflow.get_artifact_uri(artifact_path)
```

mlflow/pytorch/__init__.py
shrinath-suresh: should we need to change the `load_state_dict` method as well - https://github.com/chauhang/mlflow/blob/6f85512eb34aa1886a5a2ab142d33477325672d8/mlflow/pytorch/__init__.py#L700 . The argument name is `model_uri`. It makes more sense to me to keep it as model uri.

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -738,6 +758,40 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+    """
+    This test verifies that PyTorch models saved in older versions of MLflow are loaded successfully
+    by ``mlflow.pytorch.load_model``. The ``data`` path associated with these older models is
+    serialized PyTorch model file, as opposed to the current format: a directory containing a
+    serialized model file and pickle module information.
+    """
+    artifact_path = "pytorch_model"
+    with mlflow.start_run():
+        mlflow.pytorch.log_state_dict(
+            artifact_path=artifact_path, state_dict=sequential_model.state_dict()
```
harupy, line 775:
```suggestion
            state_dict=sequential_model.state_dict(), artifact_path=artifact_path
```

This is a nit but can we swap the arguments here?

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: removed the comment.

mlflow/pytorch/__init__.py
harupy: > It makes more sense to me to keep it as model uri.

Is this because it's rare for users to save a state_dict that doesn't contain a model?

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: Okies. Swapped the arguments.

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: done. Used `get_artifact_uri`

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -738,6 +758,30 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
     )
 
 
+@pytest.mark.large
+@pytest.mark.parametrize("scripted_model", [True, False])
+def test_load_model_succeeds_when_data_is_model_file_instead_of_directory_saved_as_state_dict(
```
harupy, line 763: Can we also rename the function name since it doesn't match what it actually tests?

mlflow/pytorch/__init__.py
shrinath-suresh: I have removed the `pickle_module` argument. if needed we can add it as part of follow-up PR.

mlflow/pytorch/__init__.py
shrinath-suresh: I have rephrased the save and log state dict methods (by removing `model`). In `load_state_dict`, the input to the method is an URI  and not the state dict itself. Hence, it makes more sense to me to keep it as such. If you feel otherwise, please let me know.

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: Done. simplified the names.

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: added comparison using [set intersection](https://github.com/chauhang/mlflow/blob/9272ec992a959ea0c671b29aaed328ed0eb8ab70/tests/pytorch/test_pytorch_model_export.py#L708)

harupy:
```python
    assert (
        len(set(model_state_dict.items()).intersection(set(sequential_model.state_dict().items())))
        == 0
    )
```

Could you explain how this asserts two dictionaries are equal?

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +287,95 @@ def gen_data():
     )
 
 
+@experimental
+def log_state_dict(state_dict, artifact_path, **kwargs):
+    """
+    Log a state dict as an MLflow artifact for the current run.
+
+    :param state_dict: PyTorch model state dict to be saved.
+
```
harupy, line 296: Can you remove these blank lines between `:param`?

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: Compared the set interesection of two dictionary items. If the items are same, the length of the set would be zero . otherwise non-zero. @harupy Feel free to suggest any better approach to compare two `OrderedDict`

harupy:
```python
a = {"a": 1}
b = {"a": 1}

assert len(set(a.items()).intersection(set(b.items()))) == 0
```

throws:

```
AssertionError
```

mlflow/pytorch/__init__.py
```diff
@@ -32,11 +32,13 @@
 from mlflow.utils.environment import _mlflow_conda_env
 from mlflow.utils.file_utils import _copy_file_or_tree, TempDir
 from mlflow.utils.model_utils import _get_flavor_configuration
+from mlflow.utils.annotations import experimental  # pylint: disable=reimported
```
harupy, line 35: Why do we need `# pylint: disable=reimported` here?

mlflow/pytorch/__init__.py
```diff
@@ -591,6 +680,33 @@ def _load_model(path, **kwargs):
...
+                    - ``relative/path/to/local/model``
+                    - ``s3://my_bucket/path/to/model``
+                    - ``runs:/<mlflow_run_id>/run-relative/path/to/model``
+                    - ``models:/<model_name>/<model_version>``
+                    - ``models:/<model_name>/<stage>``
+
+                    For more information about supported URI schemes, see
+                    `Referencing Artifacts <https://www.mlflow.org/docs/latest/concepts.html#
+                    artifact-locations>`_.
+    :return: A pytorch model state dict of type OrderedDictionary
```
harupy, line 700: Is the return tyep alwasy `OrderedDictionary`?

```python
torch.save({"model": model.state_dict(), "other_item": ...})
type(state_dict = torch.load(model_path))  # -> ???
```

mlflow/pytorch/__init__.py
```diff
@@ -285,6 +287,92 @@ def gen_data():
     )
 
 
+@experimental
+def log_state_dict(state_dict, artifact_path, **kwargs):
+    """
+    Log a state dict as an MLflow artifact for the current run.
+
+    :param state_dict: PyTorch model state dict to be saved.
```
harupy, line 295:
```suggestion
    :param state_dict: PyTorch state dict to be saved.
```

Can we remove `model` here?

mlflow/pytorch/__init__.py
```diff
@@ -591,6 +680,33 @@ def _load_model(path, **kwargs):
...
+                    - ``models:/<model_name>/<stage>``
+
+                    For more information about supported URI schemes, see
+                    `Referencing Artifacts <https://www.mlflow.org/docs/latest/concepts.html#
+                    artifact-locations>`_.
+    :return: A pytorch model state dict of type OrderedDictionary
+    """
+    import torch
+
+    model_path = _get_model_artifact_path(model_uri)
```
harupy, line 704: Can we just use `_download_artifact_from_uri` here instead of introducing `_get_model_artifact_path`?

```python
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -738,6 +760,33 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+@pytest.mark.parametrize("scripted_model", [True, False])
+def test_log_state_dict(sequential_model, model_path, data):
+    artifact_path = "pytorch_model"
+    with mlflow.start_run():
+        mlflow.pytorch.log_state_dict(
+            state_dict=sequential_model.state_dict(), artifact_path=artifact_path
+        )
+        model_path = mlflow.get_artifact_uri(artifact_path)
+
+    model_class = nn.Sequential(nn.Linear(4, 3), nn.ReLU(), nn.Linear(3, 1),)
```
harupy, line 773: Can we define a util function like below and and reuse it in the `sequential_model` fixture and here?

```python
def get_sequential_model():
    return nn.Sequential(nn.Linear(4, 3), nn.ReLU(), nn.Linear(3, 1),)
```

harupy: and `nn.Sequential(nn.Linear(4, 3), nn.ReLU(), nn.Linear(3, 1),)` is not a class but a model. Can you we rename the varible name `model_class`?

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -738,6 +760,33 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+    model_class.eval()
+
+    assert (
+        len(set(model_state_dict.items()).intersection(set(sequential_model.state_dict().items())))
+        == 0
+    )
+
+    np.testing.assert_array_almost_equal(
+        pd.DataFrame(_predict(model=model_class, data=data)),
+        pd.DataFrame(_predict(model=sequential_model, data=data)),
```
harupy, line 785: I don't think we need to use `pd.DataFrame`.

mlflow/pytorch/__init__.py
shrinath-suresh: without this disable, pylint is throwing a reimported error (eventhough experimental is not imported before)

mlflow/pytorch/__init__.py
shrinath-suresh: Yes. When i save the state dict using `torch.save` and load the state dict back, the type is `OrderedDictionary`

For example:
```
>>> 
>>> import torch
>>> 
>>> 
>>> type(torch.load("iris.pt"))
<class 'collections.OrderedDict'>
>>> 
```

mlflow/pytorch/__init__.py
shrinath-suresh: done

mlflow/pytorch/__init__.py
shrinath-suresh: May be in future when we add MLModel file for state dict implementation, we may once again need `_get_model_artifact_path`. I just followed the same convention as in load_model.

Given the fact that we don't have MLModel file at present, we can replace it with`_download_artifact_from_uri`.

shrinath-suresh: I have replaced and uploaded a newer verision

mlflow/pytorch/__init__.py
harupy: Can you merge the master branch and check if the same error occurs?

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: created `get_sequential_model` and renamed `model_class` to `model`

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: I have removed it.

mlflow/pytorch/__init__.py
harupy: How did you save `iris.pt`?

```python
# torch.save(model.state_dict(), ...)  # not this
torch.save({"model": model.sttate_dict(), ...})
```

mlflow/pytorch/__init__.py
shrinath-suresh: sure will do

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: I have added a method to compare the state dict. The problem here is , the values inside the dictionaries might be tensors. We cannot simply do `x[k] == y[k]` in such case. Added a condition for the same. Please review and lemme know if it is okay or any new approach can be followed to compare state dicts.

mlflow/pytorch/__init__.py
shrinath-suresh: Got it.. In this case, it would be of type `dict`. Removed `Ordered Dictionary` from docstring

mlflow/pytorch/__init__.py
shrinath-suresh: merged master branch and the issue got fixed.

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -769,6 +790,41 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
     )
 
 
+def compare_state_dicts(state_dict1, state_dict2):
+    for key, _ in state_dict1.items():
```
harupy, line 794:
```suggestion
    for key, _ in state_dict1:
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -769,6 +790,41 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
     )
 
 
+def compare_state_dicts(state_dict1, state_dict2):
```
harupy, line 793: Can we rename this function to `state_dict_equal` to clarify this function checks the equality of state dicts? The word `compare` sounds a bit ambiguous.

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -769,6 +790,41 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+    for key, _ in state_dict1.items():
+        if key not in state_dict2:
+            return False
+        value1 = state_dict1[key]
+        value2 = state_dict2[key]
+        if isinstance(value1, torch.Tensor) and isinstance(value2, torch.Tensor):
+            if not torch.all(value1.eq(value2)):
```
harupy, line 800: Can we use `torch.equal` (which I think is easier to understand) here?

https://pytorch.org/docs/stable/generated/torch.equal.html

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -769,6 +790,41 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+        mlflow.pytorch.log_state_dict(
+            state_dict=sequential_model.state_dict(), artifact_path=artifact_path
+        )
+        model_path = mlflow.get_artifact_uri(artifact_path)
+
+    model = get_sequential_model()
+    model_state_dict = mlflow.pytorch.load_state_dict(model_path)
+    model.load_state_dict(model_state_dict)
+    model.eval()
+    assert compare_state_dicts(model_state_dict, sequential_model.state_dict()) is True
```
harupy, line 821:
```suggestion
    assert compare_state_dicts(model_state_dict, sequential_model.state_dict())
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -725,6 +729,23 @@ def test_load_pyfunc_succeeds_when_data_is_model_file_instead_of_directory(
...
+@pytest.mark.parametrize("scripted_model", [True, False])
+def test_save_state_dict(sequential_model, model_path, data):
+
+    mlflow.pytorch.save_state_dict(state_dict=sequential_model.state_dict(), path=model_path)
+
+    model = get_sequential_model()
+    model_state_dict = mlflow.pytorch.load_state_dict(model_path)
+    model.load_state_dict(model_state_dict)
+    model.eval()
+    assert compare_state_dicts(model_state_dict, sequential_model.state_dict()) is True
```
harupy, line 742:
```suggestion
    assert compare_state_dicts(model_state_dict, sequential_model.state_dict())
```

mlflow/pytorch/__init__.py
harupy: Do we still need `_get_model_artifact_path`?

mlflow/pytorch/__init__.py
```diff
@@ -439,6 +527,7 @@ class LinearNNModel(torch.nn.Module):
         predict X: 12.0, y_pred: 23.96
         predict X: 30.0, y_pred: 60.13
     """
+
```
harupy, line 530: Can we revert this unrelated change?

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -769,6 +790,41 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
     )
 
 
+def state_dict_equal(state_dict1, state_dict2):
+    for key in state_dict1.items():
```
harupy, line 794:
```suggestion
    for key in state_dict1:
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -725,6 +729,23 @@ def test_load_pyfunc_succeeds_when_data_is_model_file_instead_of_directory(
...
+@pytest.mark.parametrize("scripted_model", [True, False])
+def test_save_state_dict(sequential_model, model_path, data):
+
+    mlflow.pytorch.save_state_dict(state_dict=sequential_model.state_dict(), path=model_path)
+
+    model = get_sequential_model()
+    model_state_dict = mlflow.pytorch.load_state_dict(model_path)
+    model.load_state_dict(model_state_dict)
+    model.eval()
+    assert state_dict_equal(model_state_dict, sequential_model.state_dict())
```
harupy, line 742:
```suggestion
    state_dict = sequential_model.state_dict()
    mlflow.pytorch.save_state_dict(state_dict=state_dict, path=model_path)

    model = get_sequential_model()
    loaded_state_dict = mlflow.pytorch.load_state_dict(model_path)
    model.load_state_dict(loaded_state_dict)
    model.eval()
    assert state_dict_equal(loaded_state_dict, state_dict)
```

shrinath-suresh:
@harupy Locally, i am getting setup error. I am yet to run the code in my local machine. I will test the code and keep you posted.

mlflow/pytorch/__init__.py
```diff
@@ -593,6 +681,33 @@ def _load_model(path, **kwargs):
...
+    Load a state dict from a local file or a run.
+
+    :param model_uri: The location, in URI format, of the MLflow model, for example:
+
+                    - ``/Users/me/path/to/local/model``
+                    - ``relative/path/to/local/model``
+                    - ``s3://my_bucket/path/to/model``
+                    - ``runs:/<mlflow_run_id>/run-relative/path/to/model``
+                    - ``models:/<model_name>/<model_version>``
+                    - ``models:/<model_name>/<stage>``
```
harupy, line 696: Can we fix the parameter description here because `load_state_dict` doesn't load an MLflow model?

mlflow/pytorch/__init__.py
```diff
@@ -287,6 +288,93 @@ def gen_data():
...
+        save_state_dict(state_dict=state_dict, path=local_path, **kwargs)
+        mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
+
+
+@experimental
+def save_state_dict(state_dict, path, **kwargs):
+    """
+    Save state dict to a path on the local file system
+
+    :param state_dict: state dict of the pytorch model. Ex: model.state_dict()
```
harupy, line 361: Can we simply say `PyTorch state dict to be saved` here?

mlflow/pytorch/__init__.py
```diff
@@ -287,6 +288,93 @@ def gen_data():
...
+    Log a state dict as an MLflow artifact for the current run.
+
+    :param state_dict: PyTorch state dict to be saved.
+    :param artifact_path: Run-relative artifact path.
+    :param kwargs: kwargs to pass to ``torch.save`` method.
+
+    .. code-block:: python
+        :caption: Example
+
+        import torch
```
harupy, line 303: I think the example here is an overkill and difficult to understand.

```python
with mlflow.start_run():
    state_dict = model.state_dict()
    mlflow.pytorch.log_state_dict(state_dict, artifact_path="state_dict")
```

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -769,6 +790,41 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
...
+def state_dict_equal(state_dict1, state_dict2):
+    for key in state_dict1.items():
+        if key not in state_dict2:
+            return False
+        value1 = state_dict1[key]
+        value2 = state_dict2[key]
+        if isinstance(value1, torch.Tensor) and isinstance(value2, torch.Tensor):
+            if not torch.equal(value1, value2):
+                return False
+        elif state_dict1[key] != state_dict2[key]:
```
harupy, line 802: Can we reuse `value1` and `value2` here?

mlflow/pytorch/__init__.py
```diff
@@ -287,6 +288,93 @@ def gen_data():
...
+    :param state_dict: state dict of the pytorch model. Ex: model.state_dict()
+    :param path: Local path where the state_dict is to be saved.
+    :param kwargs: kwargs to pass to ``torch.save`` method.
+    """
+
+    import torch
+
+    os.makedirs(path)
+
+    model_data_subpath = "data"
```
harupy, line 370: Do we need this `data` directory?

shrinath-suresh: https://github.com/mlflow/mlflow/blob/9991104e157a30ae50df295d792291254ed21b96/mlflow/pytorch/__init__.py#L464 - In save model, the actual model is stored under `data` folder. I feel it would be better to follow the same convention. In future, when state dict is considered as model, the folder structure would be uniform.

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: done . i have uploaded the newer version.

mlflow/pytorch/__init__.py
shrinath-suresh: I used the same example as in log/save model. I have updated the code given above.

mlflow/pytorch/__init__.py
shrinath-suresh: done. rephrased the doc string.

mlflow/pytorch/__init__.py
shrinath-suresh: removed `models:/`. Since we doenload the model using `_download_artifacts_from_uri`, the first 4 elements should still work. Correct me if am wrong.

tests/pytorch/test_pytorch_model_export.py
shrinath-suresh: I guess it makes sense to keep `get_sequential_model` before loading the state dict. First we define the model and then load it.

mlflow/pytorch/__init__.py
harupy: If we use the same folder structure, it's a bit confusing isn't it?

```python
# This folder structure looks like an MLflow PyTorch model, but actually is NOT.
- model
  - data
    - state_dict.pt

# How the folder structure would look like if we supported logging a state_dict as an MLflow model
- model
  - MLmodel
  - conda.yml 
  - data
    - state_dict.pt
```

mlflow/pytorch/__init__.py
harupy: @shrinath-suresh Can we revert the changes on `load_model`?

harupy:
@shrinath-suresh Thank you for patiently addressing my comments 👍

mlflow/pytorch/__init__.py
shrinath-suresh: Can you please let me know your recommendation ?

should we save it like below ?

```
- model
  - state_dict.pt
```

shrinath-suresh: In future, when the state dict implementation is considered as model, We don't have to alter the structure again. And we can reuse the code from `load_model` if the same structure is followed. But it is just my opinion, if you feel otherwise, we can change it.

harupy: I think we'll update `log_model`.

shrinath-suresh: Okie. Should we remove the `data` folder then ?

harupy: Yes, and another thing we need to consider is which one the following usages makes more sense for users.


```python
# specify a directory
log_state_dict(state_dict, "dir1")
log_state_dict(state_dict, "dir2")

load_state_dict("dir1")
load_state_dict("dir2")

- dir1
  - state_dict.pt
- dir2
  - state_dict.pt

# ------------- vs. -------------

# specify a file path
log_state_dict(state_dict, "dir/state_dict1.pt")
log_state_dict(state_dict, "dir/state_dict2.pt")

load_state_dict("dir/state_dict1.pt")
load_state_dict("dir/state_dict2.pt")

- dir
  - state_dict1.pt
  - state_dict2.pt
```

shrinath-suresh: I have removed `data` folder .

shrinath-suresh: The seconds approach would be too confusing for users from pytorch side. When user wants to log the model, they would need to give directory as input and while using state dict they need to provide filename as input. 

The first approach(the one we follownow ) is uniform and would avoid too many confusions for the user.

mlflow/pytorch/__init__.py
```diff
@@ -287,6 +288,46 @@ def gen_data():
...
+    .. code-block:: python
+        :caption: Example
+
+        with mlflow.start_run():
+            state_dict = model.state_dict()
+            mlflow.pytorch.log_state_dict(state_dict, artifact_path="state_dict")
+    """
+
+    with TempDir() as tmp:
+        local_path = tmp.path("model")
```
harupy, line 309: I think this line can be removed.

mlflow/pytorch/__init__.py
```diff
@@ -287,6 +288,46 @@ def gen_data():
...
+        save_state_dict(state_dict=state_dict, path=local_path, **kwargs)
+        mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
```
harupy, line 311:
```suggestion
        save_state_dict(state_dict=state_dict, path=tmp.path(), **kwargs)
        mlflow.log_artifacts(local_path, artifact_path)
```

The code here can be simplified like this.

mlflow/pytorch/__init__.py
```diff
@@ -593,6 +634,31 @@ def _load_model(path, **kwargs):
             return torch.jit.load(model_path)
 
 
+@experimental
+def load_state_dict(model_uri):
+    """
+    Load a state dict from a local file or a run.
+
+    :param model_uri: The location, in URI format, of the MLflow model, for example:
```
harupy, line 642: Can we remove `MLflow model` here?

mlflow/pytorch/__init__.py
shrinath-suresh: I have replaced `mlflow.tracking.fluent.log_artifacts` with `mlflow.log_artifacts`. `local_path` variable is assigned as it is used twice, once in `save_state_dict` and other in `mlflow.log_artifacts`. So, can we keep the variable ?

harupy: Yes. I just wanted to say we don't need `.path("model")`.

harupy: Yes, we can keep it.

```python
        local_path = tmp.path()
        save_state_dict(state_dict=state_dict, path=local_path, **kwargs)
        mlflow.log_artifacts(local_path, artifact_path)
```

harupy:
@shrinath-suresh Could you revert changes related to `_get_model_artifact_path`?

mlflow/pytorch/__init__.py
harupy, line 309:
```suggestion
        local_path = tmp.path()
```

mlflow/pytorch/__init__.py
```diff
@@ -593,6 +634,31 @@ def _load_model(path, **kwargs):
...
+                    `Referencing Artifacts <https://www.mlflow.org/docs/latest/concepts.html#
+                    artifact-locations>`_.
+    :return: A pytorch state dict
+    """
+    import torch
+
+    local_model_path = _download_artifact_from_uri(artifact_uri=model_uri)
+    model_path = os.path.join(local_model_path, _TORCH_STATE_DICT_FILE_NAME)
+    state_dict = torch.load(model_path)
+    return state_dict
```
harupy, line 659: Can we avoid using `"model"` in this function to avoid confusion?

harupy, line 659: Can we replace "model" with "state_dict" in this function to avoid confusion and clarify what this function expects is not the URI of an MLflow model?

harupy:
@shrinath-suresh Could you revert changes related to `_get_model_artifact_path`?

mlflow/pytorch/__init__.py
```diff
@@ -593,6 +634,31 @@ def _load_model(path, **kwargs):
             return torch.jit.load(model_path)
 
 
+@experimental
+def load_state_dict(model_uri):
```
harupy, line 638: Can we add `**kwargs` here and propagate it to `torch.load`?

harupy:
@shrinath-suresh I have pushed some commits to clean up the code :)

shrinath-suresh:
@harupy Thank you very much. The changes LGTM. Is there any other comment you have on this PR ? if not can we merge the PR?

harupy (approved):
@shrinath-suresh LGTM! Thanks for all the hard work 👍

mlflow/pytorch/__init__.py
```diff
@@ -725,6 +726,102 @@ def predict(self, data, device="cpu"):
...
 
+@experimental
+def log_state_dict(state_dict, artifact_path, **kwargs):
+    """
+    Log a state_dict as an MLflow artifact for the current run.
+
+    :param state_dict: state_dict to be saved.
+    :param artifact_path: Run-relative artifact path.
+    :param kwargs: kwargs to pass to ``torch.save``.
+
```
harupy, line 737:
```suggestion
.. warning:
    We currently don't support logging a state_dict as an MLflow model.
```

@smurching Shoule we mention that we don't supprt logging a state_dict as an MLflow model to avoid confusing users?

PR #3707, opened 2020-11-17
Add `log_figure`

harupy:

Add `log_figure`  which logs an in-memory figure object as an artifact.

## How is this patch tested?

Unit tests

Add `log_figure` which logs a in-memory figure object as an artifact.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -1052,16 +1052,16 @@ class MlflowClient(object):
             client.log_figure(run.info.run_id, fig, "figure.html")
         """
         try:
+            import matplotlib
             import plotly
-            from matplotlib import pyplot as plt
-            from plotly import graph_objects as go
         except ImportError as ie:
-            print("Failed to import required libraries: {}".format(ie))
+            _logger.error("Failed to import required libraries: {}".format(ie))
+            return
 
         with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
-            if isinstance(figure, plt.Figure):
+            if isinstance(figure, matplotlib.figure.Figure):
                 figure.savefig(tmp_path)
-            elif isinstance(figure, go.Figure):
+            elif isinstance(figure, plotly.graph_objects.Figure):
                 plotly.offline.plot(
                     figure, filename=tmp_path, include_plotlyjs="cdn", auto_open=False
                 )
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -1054,12 +1054,12 @@ class MlflowClient(object):
             client.log_figure(run.info.run_id, fig, "figure.html")
         """
 
-        def is_matplotlib_figure(fig):
+        def _is_matplotlib_figure(fig):
             import matplotlib
 
             return isinstance(fig, matplotlib.figure.Figure)
 
-        def is_plotly_figure(fig):
+        def _is_plotly_figure(fig):
             import plotly
 
             return isinstance(fig, plotly.graph_objects.Figure)
@@ -1068,9 +1068,9 @@ class MlflowClient(object):
             # `is_matplotlib_figure` is executed only when `matplotlib` is found in `sys.modules`.
             # This allows logging a `plotly` figure in an environment where `matplotlib` is not
             # installed.
-            if "matplotlib" in sys.modules and is_matplotlib_figure(figure):
+            if "matplotlib" in sys.modules and _is_matplotlib_figure(figure):
                 figure.savefig(tmp_path)
-            elif "plotly" in sys.modules and is_plotly_figure(figure):
+            elif "plotly" in sys.modules and _is_plotly_figure(figure):
                 figure.write_html(tmp_path, include_plotlyjs="cdn", auto_open=False)
             else:
                 raise TypeError("Unsupported figure object type: '{}'".format(type(figure)))
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -8,6 +8,7 @@ import logging
 import json
 import os
 import posixpath
+import re
 import tempfile
 import yaml
 
@@ -1052,20 +1053,23 @@ class MlflowClient(object):
             run = client.create_run(experiment_id="0")
             client.log_figure(run.info.run_id, fig, "figure.html")
         """
-        MATPLOTLIB_FIGURE_CLASS_NAME = "matplotlib.figure.Figure"
-        PLOTLY_FIGURE_CLASS_NAME = "plotly.graph_objs._figure.Figure"
+        MATPLOTLIB_FIGURE_REGEXP = r"^matplotlib\..*Figure$"
+        PLOTLY_FIGURE_REGEXP = r"^plotly\..*Figure$"
 
-        qual_class_name = "{}.{}".format(figure.__class__.__module__, figure.__class__.__name__)
-        if qual_class_name not in [MATPLOTLIB_FIGURE_CLASS_NAME, PLOTLY_FIGURE_CLASS_NAME]:
+        qual_class_name = figure.__class__.__module__ + "." + figure.__class__.__name__
+        if not any(
+            re.search(regexp, qual_class_name)
+            for regexp in [MATPLOTLIB_FIGURE_REGEXP, PLOTLY_FIGURE_REGEXP]
+        ):
             raise TypeError("Unsupported figure object type: '{}'".format(type(figure)))
 
         with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
             # If we used `isinstance(figure, matplotlib.figure.Figure)` here, a user would need to
             # install matplotlib to log a plotly figure. To avoid that, detect the object type
             # with class name matching.
-            if qual_class_name == MATPLOTLIB_FIGURE_CLASS_NAME:
+            if re.search(MATPLOTLIB_FIGURE_REGEXP, qual_class_name):
                 figure.savefig(tmp_path)
-            elif qual_class_name == PLOTLY_FIGURE_CLASS_NAME:
+            elif re.search(PLOTLY_FIGURE_REGEXP, qual_class_name):
                 import plotly
 
                 plotly.offline.plot(
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -1010,6 +1010,7 @@ class MlflowClient(object):
                 else:
                     json.dump(dictionary, f, indent=2)
 
+    @experimental
     def log_figure(self, run_id, figure, artifact_file):
         """
         Log a figure as an artifact. The following figure objects are supported:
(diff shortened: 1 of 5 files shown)


mlflow/tracking/client.py
```diff
@@ -1010,6 +1010,62 @@ def log_dict(self, run_id, dictionary, artifact_file):
...
+
+            run = client.create_run(experiment_id="0")
+            client.log_figure(run.info.run_id, fig, "figure.html")
+        """
+        try:
+            import plotly
+            from matplotlib import pyplot as plt
+            from plotly import graph_objects as go
+        except ImportError as ie:
+            print("Failed to import required libraries: {}".format(ie))
```
harupy, line 1059: I'll update this line.

mlflow/tracking/client.py
```diff
@@ -1010,6 +1010,62 @@ def log_dict(self, run_id, dictionary, artifact_file):
...
+
+            fig = go.Figure(go.Scatter(x=[0, 1], y=[2, 3]))
+
+            run = client.create_run(experiment_id="0")
+            client.log_figure(run.info.run_id, fig, "figure.html")
+        """
+        try:
+            import plotly
+            from matplotlib import pyplot as plt
+            from plotly import graph_objects as go
```
harupy, line 1057: Should we include `matplotlib` and `plotly` in MLflow's dependencies...?

mlflow/tracking/client.py
eedeleon: Is there a list of log_* calls that are going to be added in an Issue somewhere? Can you link me to it just to understand what is coming? It looks like a lot of mlflow-extend is graduating into MLflow :) ?

harupy: Hi @eedeleon, we're planning to add one more PR which is `log_image`.

harupy: `log_image` takes an object that represents an image (such as `numpy.ndarray`), serializes it, and log it as an artifact.

mlflow/tracking/client.py
```diff
@@ -1010,6 +1010,67 @@ def log_dict(self, run_id, dictionary, artifact_file):
...
+            # install matplotlib to log a plotly figure. To avoid that, detect the object type
+            # with class name matching.
+            if qual_class_name == MATPLOTLIB_FIGURE_CLASS_NAME:
+                figure.savefig(tmp_path)
+            elif qual_class_name == PLOTLY_FIGURE_CLASS_NAME:
+                import plotly
+
+                plotly.offline.plot(
+                    figure, filename=tmp_path, include_plotlyjs="cdn", auto_open=False
+                )
```
harupy, line 1072: There might be a better approach...

tests/tracking/test_tracking.py
```diff
@@ -538,6 +539,53 @@ def test_log_dict(subdir, extension):
             assert loaded == dictionary
 
 
+@pytest.mark.skipif(
+    importlib.util.find_spec("matplotlib") is None, reason="matplotlib is not installed"
+)
```
harupy, line 544: Should we use `@pytest.mark.large` instead?

mlflow/tracking/client.py
smurching: Instead of requiring  `plotly` to log a matplotlib figure and `matplotlib` to log a `plotly` figure (because we import both here), could we just assume e.g. a matplotlib figure and wrap the `savefig` call in a `try ... except AttributeError`, falling back to calling `plotly.offline.plot`? 

Or maybe alternatively we could perform a check like `if "plotly" in sys.modules` and only check whether the object is a `plotly` figure if `plotly` has already been imported, etc

mlflow/tracking/client.py
```diff
@@ -1010,6 +1010,62 @@ def log_dict(self, run_id, dictionary, artifact_file):
...
+        except ImportError as ie:
+            print("Failed to import required libraries: {}".format(ie))
+
+        with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
+            if isinstance(figure, plt.Figure):
+                figure.savefig(tmp_path)
+            elif isinstance(figure, go.Figure):
+                plotly.offline.plot(
+                    figure, filename=tmp_path, include_plotlyjs="cdn", auto_open=False
+                )
```
smurching, line 1073: We probably want to throw here if the object isn't a supported figure. Also one question, if we raise an exception on an unsupported figure type here, will we just end up logging an empty artifact file? Not sure how `self._log_artifact_helper` will behave in that case

harupy: In that case, the process would stop at the line where the exception is thrown.

```python
import contextlib

@contextlib.contextmanager
def foo():
    yield
    print("yes")


with foo():
    raise Exception
```

output:

```
  File "a.py", line 11, in <module>
    raise Exception
Exception
```

harupy: We could validat the object **before** `self._log_artifact_helper`.

mlflow/tracking/client.py
harupy: The try-catch approach might result in a confusing error message:

```python
fig = ""  # user thinks `fig` is a matplotlib figure but actually is not.

try:
    fig.savefig("foo.png")
except AttributeError:
    plotly.offline.plot(fig, filename="foo.html", include_plotlyjs="cdn", auto_open=False)
```

This code results in:

```
  File "a.py", line 5, in <module>
    fig.savefig("foo.png")
AttributeError: 'str' object has no attribute 'savefig'

During handling of the above exception, another exception occurred:

  File "a.py", line 7, in <module>
    plotly.offline.plot(fig, filename="foo.html", include_plotlyjs="cdn", auto_open=False)
  File "/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.6/site-packages/plotly/offline/offline.py", line 573, in plot
    figure = tools.return_figure_from_figure_or_data(figure_or_data, validate)
  File "/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.6/site-packages/plotly/tools.py", line 545, in return_figure_from_figure_or_data
    "The `figure_or_data` positional "
_plotly_utils.exceptions.PlotlyError: The `figure_or_data` positional argument must be `dict`-like, `list`-like, or an instance of plotly.graph_objs.Figure
```

mlflow/tracking/client.py
pogil: Can we write to a memory buffer like io.BytesIO instead of a local file, then use pass that buffer to the remote artifact logging request?

E.g. pass in the buffer to `figure.savefig` or `figure.write_html`

I realize you're probably asking whether there's a better way to inspect the type of figure and right way to write the plot rather than use of a temporary file... but we may be able to avoid use of temp files 🤷

harupy: @pogil If `log_artifact` supported a memory buffer, we could do that but it seems `log_artifact` doesn't.

mlflow/tracking/client.py
harupy: This is what neptune.ai does:

https://neptune-contrib.readthedocs.io/_modules/neptunecontrib/api/chart.html#log_chart

tests/tracking/test_tracking.py
smurching: Seems fine to mark the test as large IMO, rather than risk skipping it if we fail to install matplotlib before running tests

mlflow/tracking/client.py
```diff
@@ -1010,6 +1011,67 @@ def log_dict(self, run_id, dictionary, artifact_file):
                 else:
                     json.dump(dictionary, f, indent=2)
 
+    @experimental
+    def log_figure(self, run_id, figure, artifact_file):
+        """
+        Log a figure as an artifact. The following figure objects are supported:
+
+        - `matplotlib.figure.Figure`_
```
smurching, line 1019: With the current regex approach ("Option 1"): 
(+) We'll be robust to `matplotlib` or `plotly` moving around the module namespace of their `Figure` classes. In general, such API changes seem unlikely given that `plotly` and `matplotlib` both provide stable APIs.
(-) We won't detect objects that subclass matplotlib.figure.Figure - I'm not sure how common subclassing is but this seems like a potentially significant con
(-) We may have false positives with e.g. `Figure` classes in internal plotly/matplotlib modules, but it seems very unlikely that users would supply such internal classes, nor should they expect doing so to work.
(-) The code is a bit complex, but not prohibitively complex IMO

As an alternative  ("Option 2") we could check whether `plotly` / `matplotlib` are in `sys.modules`, and if so run `import plotly` in order to perform an `isinstance` check on the passed-in `figure`. In that case:
(+) The code is clearer than in the regex case - we declaratively perform a type-check instead of indirectly doing it via regex
(+) We support subclasses of `plotly` / `matplotlib` Figure
(-) The API may break against e.g. new plotly versions with this option, but not with option 1, in the case where plotly moves the module containing `Figure`.

Given these tradeoffs I think option 2 is preferable (primarily because it supports subclasses of `Figure`). If subclassing `Figure` is super uncommon then option 2 becomes less preferable, but is still better than option 1 IMO for code clarity. Lmk what you think / if I'm missing something!

harupy: sounds good. I'll make the change.

mlflow/tracking/client.py
```diff
@@ -1053,24 +1053,27 @@ def log_figure(self, run_id, figure, artifact_file):
...
-        MATPLOTLIB_FIGURE_REGEXP = r"^matplotlib\..*Figure$"
-        PLOTLY_FIGURE_REGEXP = r"^plotly\..*Figure$"
 
-        qual_class_name = figure.__class__.__module__ + "." + figure.__class__.__name__
-        if not any(
-            re.search(regexp, qual_class_name)
-            for regexp in [MATPLOTLIB_FIGURE_REGEXP, PLOTLY_FIGURE_REGEXP]
-        ):
-            raise TypeError("Unsupported figure object type: '{}'".format(type(figure)))
+        def is_matplotlib_figure(fig):
```
smurching, line 1057: Can we make this and `is_plotly_figure` private methods? e.g. `_is_plotly_figure`?

smurching (commented):
One comment otherwise LGTM!

mlflow/tracking/client.py
pogil: Should we add that support for `log_artifact`?

smurching (approved):
LGTM, @harupy would you mind posting the result of offline discussion on https://github.com/mlflow/mlflow/pull/3707#discussion_r526531172 in a reply to that thread on the PR? Otherwise looks good!

mlflow/tracking/client.py
harupy: Offline discussion on 2020/11/23:

Updating `log_artifact` to support a memory buffer requires the following changes:

- The first argument of `log_artifact` is currently named as `local_path` which doesn't sound like it accepts a memory buffer. We'd need to rename it to something like `local_path_or_memory_buffer`.
- We'd need to add a new argument (e.g. `artiafact_file`) to specify the the artifact filename.

PR #3723, opened 2020-11-19
Extend use of BatchMetricsLogger to all autologging integrations

mohamad-arabi:

Integrate the use of `BatchMetricsLogger ` for all model flavors, such that metrics are batch logged throughout training phase.

## How is this patch tested?

Added test cases for each relevant ML flavor.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/fastai.py
+++ b/mlflow/fastai.py
@@ -30,7 +30,12 @@ from mlflow.utils import gorilla
 from mlflow.utils.environment import _mlflow_conda_env
 from mlflow.utils.model_utils import _get_flavor_configuration
 from mlflow.utils.annotations import experimental
-from mlflow.utils.autologging_utils import try_mlflow_log, log_fn_args_as_params, wrap_patch
+from mlflow.utils.autologging_utils import (
+    try_mlflow_log,
+    log_fn_args_as_params,
+    wrap_patch,
+    batch_metrics_logger,
+)
 from mlflow.tracking._model_registry import DEFAULT_AWAIT_MAX_SLEEP_SECONDS
 
 
@@ -299,7 +304,7 @@ def autolog():
         """
 
         def __init__(
-            self, learner,
+            self, learner, metrics_logger
         ):
             super().__init__(learner)
             self.learner = learner
@@ -307,6 +312,7 @@ def autolog():
             self.metrics_names = ["train_loss", "valid_loss"] + [
                 o.__name__ for o in learner.metrics
             ]
+            self.metrics_logger = metrics_logger
 
         def on_epoch_end(self, **kwargs):
             """
@@ -318,7 +324,7 @@ def autolog():
             metrics = [kwargs["smooth_loss"]] + kwargs["last_metrics"]
             metrics = map(float, metrics)
             metrics = dict(zip(self.metrics_names, metrics))
-            try_mlflow_log(mlflow.log_metrics, metrics, step=epoch)
+            self.metrics_logger.record_metrics(metrics, epoch)
 
         def on_train_begin(self, **kwargs):
             info = layers_info(self.learner)
@@ -395,25 +401,27 @@ def autolog():
 
         callbacks = [cb(self) for cb in self.callback_fns] + (self.callbacks or [])
 
-        # Checking if the 'callback' argument of the function is set
-        if len(args) > callback_arg_index:
-            tmp_list = list(args)
-            callbacks += list(args[callback_arg_index])
-            tmp_list[callback_arg_index] += [__MLflowFastaiCallback(self)]
-            args = tuple(tmp_list)
-        elif "callbacks" in kwargs:
-            callbacks += list(kwargs["callbacks"])
-            kwargs["callbacks"] += [__MLflowFastaiCallback(self)]
-        else:
-            kwargs["callbacks"] = [__MLflowFastaiCallback(self)]
+        run_id = mlflow.active_run().info.run_id
+        with batch_metrics_logger(run_id) as metrics_logger:
+            # Checking if the 'callback' argument of the function is set
+            if len(args) > callback_arg_index:
+                tmp_list = list(args)
+                callbacks += list(args[callback_arg_index])
+                tmp_list[callback_arg_index] += [__MLflowFastaiCallback(self, metrics_logger)]
+                args = tuple(tmp_list)
+            elif "callbacks" in kwargs:
+                callbacks += list(kwargs["callbacks"])
+                kwargs["callbacks"] += [__MLflowFastaiCallback(self, metrics_logger)]
+            else:
+                kwargs["callbacks"] = [__MLflowFastaiCallback(self, metrics_logger)]
 
-        early_stop_callback = _find_callback_of_type(EarlyStoppingCallback, callbacks)
-        one_cycle_callback = _find_callback_of_type(OneCycleScheduler, callbacks)
+            early_stop_callback = _find_callback_of_type(EarlyStoppingCallback, callbacks)
+            one_cycle_callback = _find_callback_of_type(OneCycleScheduler, callbacks)
 
-        _log_early_stop_callback_params(early_stop_callback)
-        _log_one_cycle_callback_params(one_cycle_callback)
+            _log_early_stop_callback_params(early_stop_callback)
+            _log_one_cycle_callback_params(one_cycle_callback)
 
-        result = original(self, *args, **kwargs)
+            result = original(self, *args, **kwargs)
 
         if auto_end_run:
             try_mlflow_log(mlflow.end_run)
--- a/mlflow/fastai.py
+++ b/mlflow/fastai.py
@@ -300,7 +300,7 @@ def autolog(log_models=True):
     from fastai.callbacks.hooks import model_summary, layers_info
     from fastai.callbacks import EarlyStoppingCallback, OneCycleScheduler
 
-    def getFastaiCallback(metrics_logger):
+    def getFastaiCallback(metrics_logger, learner):
         class __MLflowFastaiCallback(LearnerCallback):
             """
             Callback for auto-logging metrics and parameters.
@@ -357,7 +357,7 @@ def autolog(log_models=True):
                 if log_models:
                     try_mlflow_log(log_model, self.learner, artifact_path="model")
 
-        return __MLflowFastaiCallback
+        return __MLflowFastaiCallback(learner)
 
     def _find_callback_of_type(callback_type, callbacks):
         for callback in callbacks:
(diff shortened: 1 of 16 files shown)


mlflow/tensorflow.py
```diff
@@ -651,10 +656,6 @@ class __MLflowTfKerasCallback(Callback):
         Records model structural information as params after training finishes.
         """
 
-        def __init__(self, log_models):
-            super().__init__()
```
mohamad-arabi, line 655: Note: Since `log_models` is not added to `self`  in `__init__` anymore, removing the entire method so that super's init is invoked automatically without running into `Useless super delegation in method` linting error.

mlflow/tensorflow.py
```diff
@@ -719,10 +720,6 @@ class __MLflowTfKeras2Callback(Callback):
         Records model structural information as params when training starts.
         """
 
-        def __init__(self, log_models):
```
mohamad-arabi, line 722: Note: Since `log_models` is not added to `self`  in `__init__` anymore, removing the entire method so that super's init is invoked automatically without running into `Useless super delegation in method` linting error.

tests/xgboost/test_xgboost_autolog.py
```diff
@@ -238,6 +274,53 @@ def test_xgb_autolog_logs_metrics_with_early_stopping(bst_params, dtrain):
             assert metric_history == evals_result[eval_name][metric_name] + [best_metrics]
 
 
+@pytest.mark.large
+def test_xgb_autolog_batch_metrics_logger_logs_metrics_with_early_stopping(bst_params, dtrain):
```
dbczumar, line 278: During some of our offline discussion, I think we suggested `spy`ing on the `BatchMetricsLogger` so that metrics are logged to the run as well. That way, we can compare `patched_metrics_data` to `run.data.metrics` and ensure that the same keys are present. This, coupled with 1-2 concrete value checks (e.g., `"best_iteration" in metric_dict`), should be sufficient.

Can we update our test cases to follow this paradigm? For setting up spies, the `wraps` keyword argument on `patch()` should be helpful: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock (`wraps` is also defined on `patch()` but only seems to be documented here)

tests/xgboost/test_xgboost_autolog.py
```diff
@@ -238,6 +274,53 @@ def test_xgb_autolog_logs_metrics_with_early_stopping(bst_params, dtrain):
...
+    for eval_name in [e[1] for e in evals]:
+        for metric_name in params["eval_metric"]:
+            metric_key = "{}-{}".format(eval_name, metric_name)
+            metric_history = [x[1] for x in patched_metrics_data if x[0] == metric_key]
+
+            assert metric_key in metric_dict
+            assert len(metric_history) == 20 + 1
+
+            best_metrics = evals_result[eval_name][metric_name][model.best_iteration]
+            assert metric_history == evals_result[eval_name][metric_name] + [best_metrics]
```
dbczumar, line 321: On second thought, I don't think we need to check the metric history this rigorously. Presumably, we do this in other test cases in the suite. Our goal here is just to make sure that `batch_metrics_logger` is being used; we can be reasonably confident that this is the case by comparing metric names in `patched_metrics_data`  entries with `run.data.metrics`.

dbczumar (changes requested):
@mohamad-arabi Implementation looks great - left a couple comments around testing!

tests/xgboost/test_xgboost_autolog.py
dbczumar: EDIT: My suggestion to use `patch()` is actually misguided. Instead, we should update the mocked method to call the original method. We can do this by grabbing a reference to the original method before we apply the patch.

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -49,152 +49,151 @@ def _autolog(log_every_n_epoch=1, log_models=True):
...
+            def on_test_end(self, trainer, pl_module):
+                """
+                Logs accuracy and other relevant metrics on the testing end
+
+                :param trainer: pytorch lightning trainer instance
+                :param pl_module: pytorch lightning base module
+                """
+                try_mlflow_log(mlflow.set_tag, "Mode", "testing")
+                for key, value in trainer.callback_metrics.items():
+                    try_mlflow_log(mlflow.log_metric, key, float(value))
```
dbczumar, line 145: Should we use `record_metrics` here? (Probably not too significant since this is presumably called at the end of training)

mlflow/fastai.py
```diff
@@ -295,63 +300,64 @@ def autolog(log_models=True):
...
+                        f.write(summary)
+                    try_mlflow_log(mlflow.log_artifact, local_path=summary_file)
+                finally:
+                    shutil.rmtree(tempdir)
+
+            def on_train_end(self, **kwargs):
+                if log_models:
+                    try_mlflow_log(log_model, self.learner, artifact_path="model")
+
+        return __MLflowFastaiCallback
```
dbczumar, line 360: Nit: Can we have these `get*Callback` methods across the various modules construct an instance of the callback class, or is there a case where we need to return the class definition instead?

tests/fastai/test_fastai_autolog.py
```diff
@@ -262,3 +264,34 @@ def test_fastai_autolog_non_early_stop_callback_does_not_log(fastai_random_data_
...
+            patched_metrics_data.extend(metrics.items())
+            original(self, metrics, step)
+
+        record_metrics_mock.side_effect = record_metrics_side_effect
+        _, run = fastai_random_data_run_with_callback(
+            iris_data(), fit_variant, manual_run, callback, patience
+        )
+
+    patched_metrics_data = dict(patched_metrics_data)
+    original_metrics = run.data.metrics
```
dbczumar, line 294: Can we add a small check to each of these test cases across the various modules?:

- Pick *one* concrete metric name (e.g., `train_loss`) and assert that the metric name is in `run.data.metrics`.

This assertion will guard against the case that our mocking logic somehow resulted in metrics being swallowed or inadvertently mutated. Looks like we do the this with Keras and TF Keras for  `restored_epoch` and `stopped_epoch`, which is awesome; let's add it to the other ones!

tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -285,6 +287,47 @@ def test_tf_keras_autolog_early_stop_logs(tf_keras_random_data_run_with_callback
     assert history.history["loss"][restored_epoch] == metric_history[-1].value
 
 
+@pytest.mark.large
+@pytest.mark.parametrize("restore_weights", [True])
+@pytest.mark.parametrize("callback", ["early"])
+@pytest.mark.parametrize("patience", [0, 1, 5])
+def test_tf_keras_autolog_batch_metrics_logger_logs_expected_metrics(
```
dbczumar, line 294: Is there a non-tf-keras test case that we should add? Vanilla TensorFlow workflows are also important here.

dbczumar (commented):
@mohamad-arabi This is very close! Left a few small comments.

mlflow/pytorch/_pytorch_autolog.py
mohamad-arabi: No, because after extensive  reading of pytorch docs, from my understanding: testing stage which invokes `on_test_end ` is not part of the training stage whether it be `fit` or `train`. Therefore the batch metrics logger context would have been torn down, and recording metrics using it would not make sense. I'm sure you would know better, if I can guarantee that `on_test_end()` will be called before the context is teared down then definitely we should use `record_metrics()` instead of `log_metric `.

mlflow/fastai.py
mohamad-arabi: Yes we can, will update in commit

mohamad-arabi: Only pytorch we need to return class, because we check `isinstance(callbacks, __MLflowPLCallback)`.

tests/tensorflow_autolog/test_tensorflow2_autolog.py
mohamad-arabi: No. We only add batch metrics logger to tf.keras `fit` and `fit_generator`.  We don't use the logger for Vanilla TensorFlow. For `tf.estimator` we use TensorBoard to log metrics, and  that already has a custom batching solution implemented `_metric_queue`. Correct me if I am wrong

dbczumar: Right! We had this discussion previously. Thanks for the reminder!

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -49,152 +49,157 @@ def _autolog(log_every_n_epoch=1, log_models=True):
...
+
+            def on_train_end(self, trainer, pl_module):
+                """
+                Logs the model checkpoint into mlflow - models folder on the training end
+
+                :param trainer: pytorch lightning trainer instance
+                :param pl_module: pytorch lightning base module
+                """
+                # manually flushing any remaining metrics from training.
+                metrics_logger._flush()
```
dbczumar, line 130: Can we make this a public API on `BatchMetricsLogger` (i.e. drop the `_`) since we have legitimate use cases for calling it outside of the module where it's defined?

mohamad-arabi: @dbczumar  Now that I made it a public API do you think I need to add a unti test for `flush()` in `test_autologging_utils.py`?

mlflow/utils/autologging_utils.py
```diff
@@ -156,7 +156,16 @@ def resolve_input_example_and_signature(
...
 
 class BatchMetricsLogger:
-    def __init__(self, run_id):
+    """
+    The BatchMetricsLogger will log metrics in batch against an mlflow run.
+    If run_id is passed to to constructor then all recording and logging will
+    happen against that run_id.
+    If no run_id is passed into constructor, then it is assumed that there is
+    an active mlflow run when there is a need to log_batch metrics. The id of that active run
+    will be retrieved from mlflow.
```
dbczumar, line 165:
```suggestion
    If no run_id is passed into constructor, then the run ID will be fetched
    from `mlflow.active_run()` each time `record_metrics()` is called; in this
    case, callers must ensure that an active run is present before invoking
    `record_metrics()`.
```

dbczumar (approved):
LGTM once https://github.com/mlflow/mlflow/pull/3723/files/1655615b27d4e4721555b102bfc7133c62f9e67d..bcbda87059117e443d1880caace0709977ce87ac#r534889780 is addressed & tests pass!

PR #3728, opened 2020-11-23
Add `log_image`

harupy:

Adds `log_image` that logs an in-memory image object as an artifact.

## How is this patch tested?

Unit tests

Adds `log_image` that logs an in-memory image object as an artifact.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -1148,7 +1148,7 @@ class MlflowClient(object):
             return isinstance(image, np.ndarray)
 
         def _normalize_to_uint8(x):
-            # Based on: https://github.com/matplotlib/matplotlib/blob/06567e021f21be046b6d6dcf00380c1cb9adaf3c/lib/matplotlib/image.py#L684
+            # Based on: https://github.com/matplotlib/matplotlib/blob/06567e021f21be046b6d6dcf00380c1cb9adaf3c/lib/matplotlib/image.py#L684  # noqa
 
             is_int = np.issubdtype(x.dtype, np.integer)
             low = 0
@@ -1179,7 +1179,7 @@ class MlflowClient(object):
                         "Please install it via: pip install Pillow"
                     ) from exc
 
-                # Ref.: https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html#numpy-dtype-kind
+                # Ref.: https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html#numpy-dtype-kind  # noqa
                 valid_data_types = {
                     "b": "bool",
                     "i": "signed integer",
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -1104,8 +1104,8 @@ class MlflowClient(object):
 
             - shape (H: height, W: width):
 
-                - H x W (Gray)
-                - H x W x 1 (Gray)
+                - H x W (Grayscale)
+                - H x W x 1 (Grayscale)
                 - H x W x 3 (RGB)
                 - H x W x 4 (RGBA)
 
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -1127,24 +1127,8 @@ class MlflowClient(object):
 
             return isinstance(image, np.ndarray)
 
-        def _to_rgba(img):
-            # grayscale (2D)
-            if image.ndim == 2:
-                alpha = np.ones_like(img, dtype=np.uint8) * 255
-                return np.stack((img, img, img, alpha), axis=-1)
-
-            # grayscale (3D)
-            elif image.ndim == 3 and image.shape[2] == 1:
-                alpha = np.ones_like(img, dtype=np.uint8) * 255
-                return np.concatenate((img, img, img, alpha), axis=-1)
-
-            # RGB
-            elif img.shape[2] == 3:
-                alpha = np.ones_like(img[:, :, [0]], dtype=np.uint8) * 255
-                return np.concatenate((img, alpha), axis=-1)
-
-            # RGBA
-            return img
+        def _gray_to_rgb(img):
+            return np.stack((img if img.ndim == 2 else img[:, :, 0],) * 3, axis=-1)
 
         with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
             if "PIL" in sys.modules and _is_pillow_image(image):
@@ -1180,7 +1164,11 @@ class MlflowClient(object):
                         )
                     )
 
-                Image.fromarray(_to_rgba(image)).save(tmp_path)
+                # Convert grayscale to RGB
+                if (image.ndim == 2) or (image.ndim == 3 and image.shape[2] == 1):
+                    image = _gray_to_rgb(image)
+
+                Image.fromarray(image).save(tmp_path)
 
             else:
                 raise TypeError("Unsupported image object type: '{}'".format(type(image)))
--- a/mlflow/tracking/client.py
+++ b/mlflow/tracking/client.py
@@ -1155,7 +1155,9 @@ class MlflowClient(object):
             high = 255 if is_int else 1
             if x.min() < low or x.max() > high:
                 _logger.warning(
-                    "Clipping array (dtype: '{}') to [{}, {}]".format(x.dtype, low, high)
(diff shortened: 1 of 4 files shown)


mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,116 @@ def _is_plotly_figure(fig):
...
                 raise TypeError("Unsupported figure object type: '{}'".format(type(figure)))
 
+    @experimental
+    def log_image(self, run_id, image, artifact_file):
+        """
+        Log an image as an artifact. The following image objects are supported:
+
+        - `numpy.ndarray`_
+        - `PIL.Image.Image`_
+
```
harupy, line 1085: I'll update the docstring.

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,116 @@ def _is_plotly_figure(fig):
...
+            return img
+
+        with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
+            if "PIL" in sys.modules and _is_pillow_image(image):
+                image.save(tmp_path)
+            elif "numpy" in sys.modules and _is_numpy_array(image):
+                import numpy as np
+
+                high = 255 if np.issubdtype(image.dtype, np.integer) else 1.0
+                image = np.clip(image, 0, high)
```
harupy, line 1156: Clip out-of-bounds values.

harupy: I'll insert a warning here.

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,116 @@ def _is_plotly_figure(fig):
...
+
+                if image.dtype.kind == "f":
+                    image *= 255
```
harupy, line 1159: If the array data type is float, multiply 255.

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,116 @@ def _is_plotly_figure(fig):
...
+
+                image = image.astype(np.uint8)
```
harupy, line 1161: Conver the array to `np.uint8` so `Image.fromarray` accepts it.

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,116 @@ def _is_plotly_figure(fig):
...
+                return np.concatenate((img, alpha), axis=-1)
+
+            # RGBA
+            return img
+
+        with self._log_artifact_helper(run_id, artifact_file) as tmp_path:
+            if "PIL" in sys.modules and _is_pillow_image(image):
+                image.save(tmp_path)
+            elif "numpy" in sys.modules and _is_numpy_array(image):
+                import numpy as np
```
harupy, line 1153: The same logic as `log_figure`.

harupy:
A script to make sure `log_image` works fine.

```python
import os

from skimage import data
import numpy as np

import mlflow

    # Log images
    mlflow.log_image(data.astronaut(), "RGB_uint.png")
    mlflow.log_image(data.astronaut().astype(np.int16), "RGB_int.png")
    mlflow.log_image(data.astronaut().astype(np.int16) + 127, "RGB_int_clipped.png")
    mlflow.log_image(data.astronaut() / 255, "RGB_float.png")
    mlflow.log_image(data.astronaut() / 255 - 0.5, "RGB_float_clipped.png")
    mlflow.log_image(data.logo(), "RGBA_uint.png")
    mlflow.log_image(data.binary_blobs(), "bool.png")
    mlflow.log_image(data.grass(), "gray_2d.png")
    mlflow.log_image(data.grass()[..., np.newaxis], "gray_3d.png")
    mlflow.log_image(np.array(["a"], dtype=object), "gray_3d.png")

    # Print out logged images
    artifact_uri = mlflow.get_artifact_uri().replace("file://", "")

    for f in client.list_artifacts(run.info.run_id):
        print(os.path.join(artifact_uri, f.path))
```

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,128 @@ def _is_plotly_figure(fig):
...
+
+            return isinstance(image, np.ndarray)
+
+        def _normalize_to_uint8(x):
+            # Based on: https://github.com/matplotlib/matplotlib/blob/06567e021f21be046b6d6dcf00380c1cb9adaf3c/lib/matplotlib/image.py#L684
+
+            is_int = np.issubdtype(x.dtype, np.integer)
+            low = 0
+            high = 255 if is_int else 1
+            if x.min() < low or x.max() > high:
```
smurching, line 1136: From offline discussion - if we don't clip values, the cast to uint8 below will throw. It seems acceptable to clip & log a warning

harupy: >  if we don't clip values, the cast to uint8 below will throw

Not throw, but overflow:

```
>>> np.array([300]).astype(np.uint8)
array([44], dtype=uint8)
```

harupy: If a user turned off warnings, he wouldn't notice that the array was clipped.

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,128 @@ def _is_plotly_figure(fig):
...
+                _logger.warning(
+                    "Clipping array (dtype: '{}') to [{}..{}]".format(x.dtype, low, high)
+                )
```
harupy, line 1139: I started thinking we should raise an exception instead of a warning because if a user turned off warnings (with `warnings.filterwarnings("ignore")`), he wouldn't notice that the array was clipped.

harupy: I often see people turn off warnings **on a notebook** because they are distracting.

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,148 @@ def _is_plotly_figure(fig):
...
+                .. warning::
+
+                    - Out-of-range integer values will be **clipped** to [0, 255].
+                    - Out-of-range float values will be **clipped** to [0.0, 1.0].
+
+            - shape (H: height, W: width):
+
+                - H x W (Gray)
+                - H x W x 1 (Gray)
+                - H x W x 3 (RGB)
```
smurching, line 1109:
```suggestion
                - H x W x 3 (an RGB channel order is assumed)
```

mlflow/tracking/client.py
```diff
@@ -1075,6 +1075,148 @@ def _is_plotly_figure(fig):
...
+                - H x W x 4 (RGBA)
```
smurching, line 1110:
```suggestion
                - H x W x 4 (an RGBA channel order is assumed)
```

tests/tracking/test_tracking.py
```diff
@@ -590,6 +590,114 @@ def test_log_figure_raises_error_for_unsupported_figure_object_type():
         mlflow.log_figure("not_figure", "figure.png")
 
 
+@pytest.mark.large
+@pytest.mark.parametrize("subdir", [None, ".", "dir", "dir1/dir2", "dir/.."])
```
smurching, line 594: Could we also add tests for some of the `if` branches in the `log_image` implementation, to ensure we have thorough [branch coverage](https://en.wikipedia.org/wiki/Code_coverage#Basic_coverage_criteria)? For example, can we add a test for the clipping behavior?

smurching: Actually, I think ^the other branches are covered pretty well, besides the clipping behavior

tests/tracking/test_tracking.py
```diff
@@ -590,6 +590,114 @@ def test_log_figure_raises_error_for_unsupported_figure_object_type():
...
+
+    image = np.random.randint(0, 256, size=(100, 100, 3), dtype=np.uint8)
+
+    with mlflow.start_run():
+        mlflow.log_image(image, artifact_file)
+
+        artifact_path = None if subdir is None else posixpath.normpath(subdir)
+        artifact_uri = mlflow.get_artifact_uri(artifact_path)
+        run_artifact_dir = local_file_uri_to_path(artifact_uri)
+        assert os.listdir(run_artifact_dir) == [filename]
```
smurching, line 609: Could we also enhance these test to load back the image and verify that the contents match the original? I wonder if Pillow makes it easy to load the image back and compare to the original etc

smurching (commented):
Some small comments on docs/testing but otherwise looks great!

smurching (approved):
LGTM

PR #3731, opened 2020-11-24
Cross version tests for MLflow Models / autologging integrations

harupy:

Add a github config to run the cross version tests for MLflow Models / autologging integrations.

## How is this patch tested?

`.github/workflows/cross-versions-tests.yml`

Components 

Interface 

Language 

Integrations


.github/workflows/sweep.yml
```diff
@@ -0,0 +1,70 @@
+name: ML
+
+on:
+  schedule:
+    # Run this workflow daily at 7:00 UTC
+    - cron: "0 7 * * *"
+  pull_request:
```
harupy, line 7: Enabling manual triggering might be useful in the case where we want to quickly run this workflow.

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

ml-package-versions-master.yml
```diff
@@ -0,0 +1,43 @@
+sklearn:
```
harupy, line 1: Dummy file to make sure the detect-diff logic works as expected.

ml-package-versions-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    maximum: "0.23.2"
+    run: |
+      pytest tests/sklearn/test_sklearn_model_export.py --large
+
+  autologging:
+    package: "scikit-learn"
+    minimum: "0.22.2"
+    maximum: "0.23.2"
+    unsupported: ["0.22.2"]
+    requirements: ["matplotlib"]
```
harupy, line 14: Additional requirements to run tests.

.github/workflows/cross_version_tests.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          python-version: "3.6"
+      - name: Install dependencies
+        run: |
+          pip install PyYAML
+      - id: set-matrix
+        run: |
+          ARGS=$([[ "${{ github.event_name }}" = "pull_request" ]] && echo "--diff-only" || echo "")
+          MATRIX_JSON=$(python dev/generate_matrix.py $ARGS)
+          echo "::set-output name=matrix::$MATRIX_JSON"
```
harupy, line 28: On pull requests, ignore unchanged flavors or configurations

dev/generate_matrix.py
```diff
@@ -0,0 +1,209 @@
...
+    args = parse_args()
+
+    dummy_git_diff = """
+mlflow/pytorch/__init__.py
+mlflow/xgboost.py
+"""
+
+    diff_files = dummy_git_diff.strip().split("\n")
+
+    config_pr = read_yaml("ml-package-versions-pr.yml")
```
dbczumar, line 149: I was under the impression that the `pull_request` cases will check out the associated PR branch and fetch any updates to the `ml-package-versions` yaml that are made by the PR. Perhaps that's not how things work?

ml-package-versions-master.yml
```diff
@@ -0,0 +1,34 @@
...
+    minimum: "0.19.2"
+    maximum: "0.23.2"
+    run: |
+      pytest tests/sklearn/test_sklearn_model_export.py --large
+
+  autologging:
+    package: "scikit-learn"
+    minimum: "0.22.2"
+    maximum: "0.23.2"
+    unsupported: ["0.22.2"]
```
dbczumar, line 13: Note that `0.22.2` isn't actually unsupported (this was used to showcase an example of `unsupported` in the design doc).

dev/generate_matrix.py
```diff
@@ -0,0 +1,209 @@
...
+                            "version": key,
+                            "install": cfg["install"],
+                            "requirements": " ".join(cfg.get("requirements", [])),
+                            "run": cfg["run"],
+                        }
+                    )
+                    continue
+
+                versions = get_latest_micro_versions(
+                    cfg["package"], cfg["minimum"], cfg["maximum"], cfg.get("unsupported")
```
dbczumar, line 187: For *testing* purposes, we should test all the way up until the latest available version. We shouldn't limit ourselves to the configured maximum version in the YAML file. The goal of the YAML file is to document the latest version of each ML library that we tested MLflow against before releasing it. Accordingly, we need to test against the latest possible version in order to confidently update our `maximum` version documentation. Let me know if this makes sense.

ml-package-versions-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    run: |
+      pytest tests/sklearn/test_sklearn_autolog.py --large
+
+  dev:
+    package: "scikit-learn"
+    install: |
+      pip install git+https://github.com/scikit-learn/scikit-learn.git@master
+    requirements: ["matplotlib"]
+    run: |
+      pytest tests/sklearn --large
```
dbczumar, line 24: Ideally, we should run the same autologging / models test cases against the dev version. This particular `run` statement makes that happen, but it might be better for consistency / correctness going forward if we specify package information at the top of the config & let the `models` and `autologging` sections implicitly reference it.

For example:

```
sklearn:
  package_info:
      pip_release: "scikit-learn"
      pip_dev: "git+https://github.com/scikit-learn/scikit-learn.git@master"

  models:
    minimum: "0.19.2"
    maximum: "0.23.2"
    run: |
      pytest tests/sklearn/test_sklearn_model_export.py --large

  autologging:
    minimum: "0.22.2"
    maximum: "0.23.2"
    unsupported: ["0.22.2"]
    requirements: ["matplotlib"]
```

ml-package-versions-pr.yml
```diff
@@ -0,0 +1,42 @@
...
+    maximum: "0.23.2"
+    unsupported: ["0.22.2"]
+    requirements: ["matplotlib"]
+    run: |
+      pytest tests/sklearn/test_sklearn_autolog.py --large
+
+  dev:
+    package: "scikit-learn"
+    install: |
+      pip install git+https://github.com/scikit-learn/scikit-learn.git@master
```
dbczumar, line 21: Awesome that we're testing dev / nightly versions. Can we find nightlies and configure them for other libraries as well? For example: https://pypi.org/project/tf-nightly/ for TensorFlow

dbczumar (commented):
@harupy This is awesome! Left a few comments - super excited to get this merged!

.github/workflows/build-wheel.yml
```diff
@@ -1,100 +0,0 @@
-name: Build wheel
```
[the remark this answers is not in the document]
harupy: I just temporarily removed them to avoid triggering too many github actions. I'll revert them once we make sure everything works fine.

dev/generate_matrix.py
```diff
@@ -0,0 +1,209 @@
...
+
+import yaml
+
+
+def read_yaml(path):
+    with open(path) as f:
+        return yaml.load(f, Loader=yaml.SafeLoader)
+
+
+def get_released_versionos(package_name):
```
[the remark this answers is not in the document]
harupy: Thanks for the catch. Fixed!

dev/generate_matrix.py
harupy: Makes senes. I'll fix it.

ml-package-versions-pr.yml
harupy: Makes sense. I'll update the config file structure and `generate_matrix.py`.

dev/generate_matrix.py
harupy: I was thinking the followin approach.

1. Read `ml-package-versions.yml` on the PR branch
2. Read `ml-package-versions.yml`on the merge target branch (e.g. master)
3. Compare them to detect the difference.
4. Generate a matrix from the diffrence.

dev/generate_matrix.py
```diff
@@ -0,0 +1,209 @@
...
+
+def is_valid_version(ver):
+    """
+    Returns True if the given version string only contains numbers and ".", otherwise False.
+
+    Example
+    -------
+    >>> is_valid_version("1.1.0")
+    True
+    >>> is_valid_version("1.1.0rc")
```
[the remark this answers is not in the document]
harupy: Makes sense. I'll fix the logic.

ml-package-versions-pr.yml
harupy: - LightGBM: can be installed from master (requires build)
- XGBoost: can be istalled from master (requires build)
- pytorch: nightly version is available
- sklearn: can be installed from github repo.

dev/generate_matrix.py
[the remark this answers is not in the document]
harupy: PEP440 says:

> A version identifier that consists solely of a release segment and optionally an epoch identifier is termed a "final release". The release segment consists of one or more non-negative integer values, separated by dots:

```
```

dev/generate_matrix.py
dbczumar: I'd imagine that we should still run the matrix against all supported minor versions on PRs though, right? If the `ml-package-versions.yml` in the PR branch matches the one `master`, does the lack of a diff imply that we're not going to run any cross-version tests?

ml-package-versions-master.yml
harupy: Removed :)

harupy:
### Example 1:

```diff
# ml-package-versions.yml

sklearn:
  ...

  models:
-    minimum: "0.19.2"
+    minimum: "0.20.2"
    maximum: "0.23.2"
    run: |
      pytest tests/sklearn/test_models.py

  autologging:
    ...
```

Detects changes in `sklearn.models` -> run only `sklearn.models` tests.

```
// matrix woud look like

[
  // run only `models` tests
  {
    "package": "scikit-learn",
    "version": "0.20.2",
    "run": "pytest tests/sklearn/test_models.py"
  },
  {
    "package": "scikit-learn",
    "version": "0.21.2",
    "run": "pytest tests/sklearn/test_models.py"
  },
  ...,
  {
    "package": "scikit-learn",
    "version": "0.23.2",
    "run": "pytest tests/sklearn/test_models.py"
  }
]
```

---

### Example 2:

```diff
# ml-package-versions.yml

+ fastai:
+   ...
+
+   models:
+     minimum: "0.19.2"
+     maximum: "0.23.2"
+     run: |
+       pytest ...
+
+   autologging:
+     ...
```

Detects a new flavor -> run `fastai`. The matrix would look like:

```
// matrix woud look like

[
  {
    "package": "fastai",
    "version": "0.20.2",
    "run": "pytest tests/fastai/test_models.py"
  },
  ...,
  {
    "package": "fastai",
    "version": "0.23.2",
    "run": "pytest tests/fast/test_autologging.py"
  }
]
```

---

### Example 3:

```diff
# mlflow/sklearn/__init__.py

+ print("foo")
```

Detects changes in `mlflow/sklearn/__init__.py` -> run both `sklearn.models` and `sklearn.autologging`. The matrix would look like:

```
// matrix woud look like

[
  // run both `models` and `autologging` tests
  {
    "package": "scikit-learn",
    "version": "0.20.2",
    "run": "pytest tests/sklearn/test_models.py"
  },
  ...,
  {
    "package": "scikit-learn",
    "version": "0.23.2",
    "run": "pytest tests/sklearn/test_autolog.py"
  },
]
```

ml-package-versions.yml
```diff
@@ -0,0 +1,100 @@
+sklearn:
+  package_info:
+    pip_release: "scikit-learn"
+    pip_dev: |
+      pip install git+https://github.com/scikit-learn/scikit-learn.git@master
+
+  models:
+    minimum: "0.23.2"
```
harupy, line 8: Temporarily use this value to avoid creating too many github actions jobs.

ml-package-versions.yml
```diff
@@ -0,0 +1,100 @@
...
+    minimum: "1.2.1"
+    maximum: "1.2.1"
+    requirements: ["scikit-learn", "matplotlib"]
+    run: |
+      pytest tests/xgboost/test_xgboost_autolog.py --large
+
+lightgbm:
+  package_info:
+    pip_release: "lightgbm"
+    pip_dev: |
```
harupy, line 83: Should we rename this filed (e.g. `install_dev`)?

dbczumar: If we require more complex instructions than `pip install {package_name}` where `{package_name}` is a PyPI-hosted package name or a GitHub URL, then I'd propose we do `install_dev`. We could then change `pip_release` to `package_name`.

ml-package-versions.yml
```diff
@@ -0,0 +1,119 @@
...
+      pytest --verbose tests/tensorflow_autolog/test_tensorflow2_autolog.py --large
+
+keras:
+  package_info:
+    pip_release: "keras"
+
+  models:
+    minimum: "2.4.3"
+    maximum: "2.4.3"
+    requirements: ["tensorflow", "packaging", "scikit-learn"]
```
harupy, line 65: Filed a PR to remove `packaging` and use `distutils` instead: https://github.com/mlflow/mlflow/pull/3747

ml-package-versions.yml
```diff
@@ -0,0 +1,119 @@
...
+    pip_dev: |
+      git clone --recursive https://github.com/microsoft/LightGBM.git
+      cd LightGBM/python-package
+      pip install wheel
+      python setup.py install
+
+  models:
+    minimum: "3.1.0"
+    maximum: "3.1.0"
+    requirements: ["scikit-learn"]
```
harupy, line 110: I noticed that `sklearn` is used to **just** load the iris dataset. Should we create a utility function that returns the iris dataset to avoid installing `sklearn`?

```python
# mlflow/tests/utils.py
def get_iris(...)
    data = [[1, 2, 3, 4], ...]
    labels = [1, ...]
    X = ...
    y = ...
    return X, y

# -------------------

# test_xxx_autolog.py
from mlfow.tests.utils import get_iris

def test_yyy():
    iris = get_iris(...)
```

dbczumar: That would be wonderful! Let's cut out unnecessary dependencies where possible

harupy: I'll address this as a folllow-up :)

harupy: Filed a (draft) PR: https://github.com/mlflow/mlflow/pull/3748

harupy:
The xgboost autologging test fails with:

https://github.com/mlflow/mlflow/pull/3731/checks?check_run_id=1471767133#step:8:220

```
E           AssertionError: assert 'train-merror' in {'train-mlogloss': 0.244426, 'valid-mlogloss': 0.244426}
E            +  where {'train-mlogloss': 0.244426, 'valid-mlogloss': 0.244426} = <RunData: metrics={'train-mlogloss': 0.244426, 'valid-mlogloss': 0.244426}, params={'early_stopping_rounds': 'None',\n ...',\n 'mlflow.source.name': '/usr/share/miniconda/bin/pytest',\n 'mlflow.source.type': 'LOCAL',\n 'mlflow.user': 'runner'}>.metrics
```

This PR seems to be the cause: https://github.com/dmlc/xgboost/pull/6183

ml-package-versions.yml
harupy: @dbczumar btw how do we determine the minimum version to support?

ml-package-versions.yml
```diff
@@ -0,0 +1,131 @@
...
+  tf1:
+    minimum: "1.15.4"
+    maximum: "1.15.4"
+    requirements: ["keras==2.2.5"]
+    includes_dev: False
+    run: |
+      pytest tests/tensorflow/test_tensorflow_model_export.py --large
+      pytest tests/tensorflow_autolog/test_tensorflow_autolog.py --large
+      pytest tests/keras --large
+      pytest tests/keras_autolog --large
```
harupy, line 67: Do we need to include `tensorflow 1.x` in this config file?

dbczumar: Let's include it for now, since we haven't deprecated support for TF 1.x yet.

ml-package-versions.yml
dbczumar: @harupy Good question! For scikit-learn autologging, that's fortunately well-defined. For other integrations, I'd recommend that we go back and take a look at when we introduced them. For integrations that are >= 6 months old, it seems reasonable to use the library minor version at the time of introduction. For newer integrations (i.e. PyTorch lightning, perhaps @smurching has input)

harupy:
https://github.com/mlflow/mlflow/pull/3731/checks?check_run_id=1472774489#step:8:172

```
ERROR: Could not find a version that satisfies the requirement tensorflow==2.5.0-dev20201129...
```

When we install `tf-nightly`, `conda.yml` logged by `mlflow.{tensorflow,keras}` ends up like this.

```
# conda.yml

...

   - tensorflow==2.5.0-dev20201129  # this version doesn't exist
```

ml-package-versions.yml
```diff
@@ -0,0 +1,162 @@
...
+    run: |
+      pytest --verbose tests/tensorflow_autolog/test_tensorflow2_autolog.py --large
+
+  tf1:
+    minimum: "1.15.4"
+    maximum: "1.15.4"
+    # Avoid fetching versions up until the latest
+    until_latest: False
+    # Avoid running tests on `tf-nightly`
+    include_dev: False
```
dbczumar, line 78: Instead of nesting `tf1` under `tensorflow` and introducing additional YAML fields (`include_latest` and `include_dev`), can we instead use two top-level YAML entries: one for `tensorflow1.x` and the other for `tensorflow2.x`? 

To solve the `until_latest` challenge, I think it makes sense to adopt the following convention:

```
For maximum version x.a.b, we'll test up to x.y.z, where x.y.z is the latest minor version that shares the same major version as the configured maximum.
```

This way, for lightgbm, which spans multiple major versions, we'll test from 2.x to 3.x. For TensorFlow 1, which uses a single major version (`1`), we'll only test 1.x.

Regarding `include_dev`, I think we can drop this and only run dev tests if `package_info` specifies `pip_dev`. That way, the `tensorflow1.x` tests can just omit `pip_dev`.

ml-package-versions.yml
```diff
@@ -0,0 +1,162 @@
...
+    requirements:
+      "== 2.3.1": ["tensorflow==2.2.1"]
+      "> 2.3.1": ["tensorflow"]
+    run: |
+      pytest --verbose tests/keras_autolog/test_keras_autolog.py --large
+
+xgboost:
+  package_info:
+    pip_release: "xgboost"
+    pip_dev: |
```
dbczumar, line 123: `pip_dev` is a misnomer here, since the installation instructions don't involve `pip`. Can we call this field `install_dev` instead? That way, we have two fields: `pip_release` and `install_dev`.

harupy: makes sense. I'll fix this

harupy: Done!

ml-package-versions.yml
harupy: Nice approach. I made the change!

dev/set_matrix.py
```diff
@@ -0,0 +1,412 @@
...
+                or (cfg != config_ref[flavor][key])
+            ):
+                print("Processing {}.{}".format(flavor, key))
+
+                # released versions
+                all_versions = get_released_versions(package_info["pip_release"])
+                versions = filter_versions(
+                    all_versions, cfg["minimum"], cfg["maximum"], cfg.get("unsupported"),
+                )
+                versions = select_latest_micro_versions(versions)
```
harupy, line 369: Should we **always** include teh latest version in each flavor in the matrxi to replace the current flavor tests (`run-python-flavor-tests.sh` and `run-python-tf-tests.sh`)?

ml-package-versions.yml
```diff
@@ -0,0 +1,177 @@
...
+      "> 2.3.1": ["tensorflow", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
+    run: |
+      pytest tests/keras/test_keras_model_export.py --large
+
+  autologging:
+    minimum: "2.3.1"
+    maximum: "2.4.3"
+    requirements:
+      "== 2.3.1": ["tensorflow==2.2.1"]
+      "> 2.3.1": ["tensorflow"]
```
harupy, line 131: Note that `> 2.3.1` includes the dev version.

tests/keras/test_keras_model_export.py
```diff
@@ -242,6 +242,7 @@ def test_model_save_load(build_model, save_format, model_path, data):
         model_uri=os.path.abspath(model_path),
         data=pd.DataFrame(x),
         content_type=pyfunc_scoring_server.CONTENT_TYPE_JSON_SPLIT_ORIENTED,
+        extra_args=["--no-conda"],
```
harupy, line 245: Workaround for `tf-nightly`. If we didn't specify this, `pyfunc_serve_and_score_model` would try to install `tensorflow-dev2020xxxx` which doesn't exist on PyPI and throw.

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+def get_changed_flavors(changed_files, flavors):
+    """
+    Detects changed flavors from a list of changed files.
+
+    Examples
+    --------
+    >>> flavors = ["pytorch", "xgboost"]
+    >>> get_changed_flavors(["mlflow/pytorch/__init__.py", "mlflow/xgboost.py"], flavors)
```
harupy, line 165: this is out of the scope of this PR but should we unify the file structure for flavors?

Current:

```
# both file and folder/__init__.py exist

xgboost.py
pytorch/__init__.py
keras.py
sklearn/__init__.py
```

Proposed:

```
# unify them to folder/__init__.py

xgboost/__init__.py
pytorch/__init__.py
keras/__init__.py
sklearn/__init__.py
```

dbczumar: Sounds good too me as a follow-up change!

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+                        }
+                    )
+
+    matrix = {"job_name": job_names, "include": includes}
+    print(divider("Result"))
+    # specify `indent` to prettify the output
+    print(json.dumps(matrix, indent=2))
+
+    if "GITHUB_ACTIONS" in os.environ:
+        # this line prints out nothing in the console on github actions
```
dbczumar, line 422: Why does it print nothing? Do we need to print nothing?

dev/set_matrix.py
dbczumar: Is there a limit on how many CI jobs can run concurrently? Will the aggregate test time be decreased? Assuming there isn't an issue with having a slightly larger CI job count, I like this idea. It simplifies the steps taken by each job and should make debugging easier.

Could we do this as a follow-up PR?

dev/set_matrix.py
harupy: Screen Shot 2020-12-02 at 17 02 01

This is how it looks like

harupy: It seems GitHub automatically hide `::set-output:xxx...`

dev/set_matrix.py
harupy: Yes, there is a limit. It's 20: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#usage-limits

ml-package-versions.yml
```diff
@@ -0,0 +1,179 @@
...
+
+  models:
+    minimum: "1.4.0"
+    maximum: "1.7.0"
+    requirements:
+      # https://github.com/pytorch/vision#installation
+      "== 1.4.0": ["torchvision==0.5.0", "scikit-learn"]
+      "== 1.5.0": ["torchvision==0.6.0", "scikit-learn"]
+      "== 1.5.1": ["torchvision==0.6.1", "scikit-learn"]
+      "== 1.6.0": ["torchvision==0.7.0", "scikit-learn"]
```
dbczumar, line 35: This seems to unfortunately imply that we're going to have to add `== 1.7.0` with a different version of `torchvision`. Would it instead be sufficient to unpin `torchvision` and trust the PyPI resolver to find an appropriate version of `torchvision` given the pinned version of `torch`?

ml-package-versions.yml
dbczumar: I don't fully understand why we need to use the dev version here. From what I can tell, recent versions of Keras depend on TensorFlow, so presumably installing the correct version of Keras should pull in a compatible version of TensorFlow. For that reason, I think it's probably sufficient to just specify an unpinned `tensorflow` dependency. In fact, that only reason to specify `tensorflow` at all would be to handle the case where Keras drops TensorFlow as a dependency in the future, which seems somewhat unlikely.

Let me know if this reasoning makes sense.

ml-package-versions.yml
```diff
@@ -0,0 +1,179 @@
...
+    maximum: "2.4.3"
+    requirements:
+      "== 2.3.1":
+        ["tensorflow==2.2.1", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
+      "> 2.3.1": ["tensorflow", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
+    run: |
+      pytest tests/keras/test_keras_model_export.py --large
+
+  autologging:
+    minimum: "2.3.1"
```
dbczumar, line 127: I think we support Keras 2.0.0 and above. Are we sure this is the correct minimum version?

ml-package-versions.yml
harupy: You're right. If we run `pip install 'torch==1.4.0' torchvision`, pip should automatically pick an appropriate version. I'll fix it.

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+                            "install": install,
+                            "run": run,
+                        }
+                    )
+
+                # development version
+                if "install_dev" in package_info:
+                    job_name = " / ".join([flavor, "dev", key])
+                    job_names.append(job_name)
+                    requirements = process_requirements(cfg.get("requirements"), "dev")
```
dbczumar, line 404: Is it necessary to install dev versions of test requirements as well? The dev release of library X is unlikely to depend on behavior in the dev release of library Y, since the authors of X don't know when Y will be released. It's more likely that library X's behavior will interact with a currently-installable version of Y.

Let me know if I'm missing something, but I think we can drop this "dev" handling.

dbczumar: Edit: Misinterpreted this. I think this is doing exactly what I described.

ml-package-versions.yml
dbczumar: Edit: Misinterpreted this, doesn't seem like we're actually pulling in the dev version of TensorFlow. Can you clarify what you mean here @harupy ?

harupy: Sorry for the cofusion. I just wanted to tell `"> x.x.x": [package1, package2]` means that package1 and package2 will be included in the dev test.

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+    dev_numeric = "9999.9999.9999"
+
+    if version == "dev":
+        version = dev_numeric
+
+    if isinstance(requirements, dict):
+        for ver_spec, packages in requirements.items():
+            op_and_ver_pairs = map(get_operator_and_version, ver_spec.split(","))
+            match_all = all(
+                op(LooseVersion(version), LooseVersion(dev_numeric if ver == "dev" else ver))
```
dbczumar, line 283: Is it ever the case that `ver == 'dev'`? I don't currently see the use of "dev" as a version in the `requirements` section of any entries in `ml-package-versions.yml`. Do we need this?

ml-package-versions.yml
dbczumar: Ah, I see. `> 2.3.1` includes the dev version of *keras*, not *TensorFlow*. Makes sense!

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+    if requirements is None:
+        return []
+
+    if isinstance(requirements, list):
+        return requirements
+
+    dev_numeric = "9999.9999.9999"
+
+    if version == "dev":
+        version = dev_numeric
```
dbczumar, line 277: Can we add a comment explaining why we're using `9999.9999.9999` and provide an example of how this resolves correct requirements for a dev library version? I.e. maybe we can explain why this works for keras-tf2.x.

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+
+def get_released_versions(package_name):
+    """
+    Fetches the released versions & datetimes of the specified Python package.
+    """
+    url = "https://pypi.python.org/pypi/{}/json".format(package_name)
+    data = json.load(urllib.request.urlopen(url))
+
+    versions = {
+        ver: files[0]["upload_time"] for ver, files in data["releases"].items() if len(files) > 0
```
dbczumar, line 49: Is it ever possible that "upload_time" is missing from a release? Can we add a comment explaining why we're selecting the first element of `files`?

dev/set_matrix.py
harupy: For example,

```
xgboost:
  ...

  models:
    ...
    requirements:
      "> 1.2.0, != dev": ["XXX"]  # we need XXX in xgboost > 1.2.0 but don't need it in the dev version

```

tests/keras/test_keras_model_export.py
dbczumar: Sounds good!

dbczumar (commented):
@harupy This PR is great - just about ready to merge! For folks who may have to debug / modify this workflow, can we add a README to the workflows directory along with comments in the relevant files that direct users to the README for more context? It would be great if the README explains the role of each file and the different execution modes (i.e. `set_matrix.py` can be invoked on a PR, in which case it creates a test matrix for files affected by the PR using the `--diff-only` and `--changed-files` flags OR it can be invoked on a branch, in which case it creates a test matrix for all ML library integrations on the branch).

Any other details that you think might be noteworthy or clarifications that you think would help users understand the workflow would be great to include in the README as well.

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--diff-only",
+        action="store_true",
+        help="If specified, ignore unchanged flavors or configs.",
+    )
+    # make `--ref-config` and `--diff-files` configurable to make it easier to test this script.
+    parser.add_argument(
+        "--ref-config",
+        default="https://raw.githubusercontent.com/mlflow/mlflow/master/{}".format(CONFIG_PATH),
```
dbczumar, line 319: Nit: It might be slightly easier to interpret this if it the naming conventions were `--ref-versions-yaml` and `VERSIONS_YAML_PATH` instead.

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+        help=(
+            "URL or local file path of the reference config. "
```
dbczumar, line 321: Can we explain that the reference config is compared with the config on the branch where `set_matrix` is running in order to identify updates?

harupy: Sure, I'll update the description of this parameter.

dev/set_matrix.py
dbczumar: I see! Can we update the comment to indicate that this is used to set the output for the Github Actions job but doesn't actually print anything to the console?

dev/set_matrix.py
dbczumar: Got it. One other potential complicating factor is that `run-python-flavor-tests` includes test coverage for things like AzureML & our top-level `mlflow.autolog()` API (which requires that many ML libraries be installed at once). Perhaps we can revisit this some time in the future.

dev/set_matrix.py
```diff
@@ -0,0 +1,427 @@
...
+    # assuming that the top-level keys in `ml-package-versions.yml` have the format:
+    # <flavor name>(-<suffix>) (e.g. sklearn, tensorflow-1.x, keras-tf1.x)
+    flavors = set(x.split("-")[0] for x in config.keys())
+    changed_flavors = get_changed_flavors(args.changed_files, flavors)
+
+    job_names = []
+    includes = []
+    should_include_all_items = not args.diff_only
+
+    for flavor, cfgs in config.items():
```
dbczumar, line 361: What happens if `config` from a PR branch is out-of-date relative to `config-ref`? Seems like we skip things that were added in `config_ref` in that case. To simplify the logic here, can we use the file that has the most recent commit (the config on `master` or the config on the PR branch) as the source of truth?

dbczumar: Actually, shouldn't the GitHub actions workflow be executed on a branch that merges the PR with `master`? In that case, isn't `config` already up to date with all of the changes from the PR along with the most recent content in `master`? If so, I don't see why we need to do this. I know we talked about this before, but the logic is still a bit lost on me.

dev/set_matrix.py
dbczumar: Got it. Is there a case where the requirements for our MLflow tests would be different for the dev version specifically? The library requirements for the dev version are automatically installed when we run the `install_dev` command, so I think this is only relevant to MLflow test requirements.

dev/set_matrix.py
dbczumar: Edit: I see. We provide `config_ref` so that we can test  *additive* changes to the config in the PR branch without retesting the entire config. Can we add an inline comment explaining this?

dev/set_matrix.py
harupy: Done :)

dev/set_matrix.py
harupy: sure!

ml-package-versions.yml
harupy: Fixed!

dev/set_matrix.py
harupy: updated!

dev/set_matrix.py
harupy: Added a commen in the `get_released_versions` :)

ml-package-versions.yml
harupy: Thanks for the catch. Fixed it and found new errors!

harupy: In keras 2.16.0, `is_plain_keras` doesn't seem to work.

https://github.com/mlflow/mlflow/blob/7d185a8f777864bd5d7e66735108e457b1a80852/mlflow/keras.py#L168-L175

https://github.com/mlflow/mlflow/pull/3731/checks?check_run_id=1487102127#step:7:2043

harupy: In keras 2.0.9,

```
ValueError: The first layer in a Sequential model must get an `input_shape` or `batch_input_shape` argument.
```

https://github.com/mlflow/mlflow/pull/3731/checks?check_run_id=1487102127#step:7:114

```
UnboundLocalError: local variable 'arrays' referenced before assignment
```

https://github.com/mlflow/mlflow/pull/3731/checks?check_run_id=1487102103#step:7:1671

dev/set_matrix.py
harupy: Done!

tests/keras/test_keras_model_export.py
```diff
@@ -86,14 +95,14 @@ def tf_keras_model(data):
     model = TfSequential()
     model.add(TfDense(3, input_dim=4))
     model.add(TfDense(1))
-    model.compile(loss="mean_squared_error", optimizer=TfSGD())
-    model.fit(x, y)
+    model.compile(loss="mean_squared_error", optimizer=TfSGD(learning_rate=0.001))
+    model.fit(x.values, y.values)
```
harupy, line 99: Fix for keras 2.0.9. Passing a dataframe to `fix` causes an error.

ml-package-versions.yml
```diff
@@ -0,0 +1,151 @@
...
+    run: |
+      pytest tests/keras/test_keras_model_export.py --large
+
+  autologging:
+    minimum: "2.2.5"
+    maximum: "2.4.3"
+    requirements:
+      "< 2.3.1": ["tensorflow==1.15.4", "h5py<3.0"]
+      "== 2.3.1": ["tensorflow==2.2.1"]
+      "> 2.3.1, != dev": ["tensorflow"]
```
harupy, line 103: Exclude `tensorflow` in the dev test because `tf-nightly` will be installaed

harupy:
@dbczumar Can I work on README in a separate PR?

ml-package-versions.yml
```diff
@@ -0,0 +1,151 @@
...
+      pip install git+https://github.com/scikit-learn/scikit-learn.git@master
+
+  models:
+    minimum: "0.20.4"
+    maximum: "0.23.2"
+    run: |
+      pytest tests/sklearn/test_sklearn_model_export.py --large
+
+  autologging:
+    minimum: "0.20.4"
```
dbczumar, line 14: Looks like we use `0.20.3` as the minimum version in our current codebase: https://github.com/mlflow/mlflow/blob/2c77d25a2b699053e7b22253c17b63dfd5b5b6da/mlflow/sklearn/utils.py#L27. Do we need 0.20.4 instead?

dev/set_matrix.py
```diff
@@ -0,0 +1,487 @@
...
+    # `changed_flavors`.
+    should_include_all_items = (args.ref_versions_yaml is None) and (args.changed_files is None)
+
+    job_names = []
+    includes = []
+
+    for flavor, cfgs in config.items():
+        package_info = cfgs.pop("package_info")
+
+        should_include_all_items_in_this_flavor = (
```
dbczumar, line 431: Can we add a comment describing this boolean variable? It seems like it determines whether or not we need to run the full matrix of tests for the flavor or just the latest version, right?

dev/set_matrix.py
```diff
@@ -0,0 +1,487 @@
...
+            should_include_all_items
+            or any(flavor.startswith(x) for x in changed_flavors)
+            or (flavor not in config_ref)
+            or (package_info != config_ref[flavor]["package_info"])
+        )
+
+        for key, cfg in cfgs.items():
+            if (
```
dbczumar, line 439: Can we add a comment explaining these conditions?

ml-package-versions.yml
```diff
@@ -0,0 +1,151 @@
...
+      # https://github.com/keras-team/keras#multi-backend-keras-and-tfkeras
+      pip install tf-nightly 'keras>=2.4.0'
+
+  models:
+    minimum: "2.0.0"
+    maximum: "2.4.3"
+    requirements:
+      "< 2.3.1": ["tensorflow==1.15.4", "h5py<3.0", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
+      "== 2.3.1": ["tensorflow==2.2.1", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
+      "> 2.3.1": ["tensorflow", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
```
dbczumar, line 93: Do we need to exclude `dev` here as well? https://github.com/mlflow/mlflow/pull/3731/files#r535010817

ml-package-versions.yml
```diff
@@ -0,0 +1,151 @@
...
+    run: |
+      pytest tests/keras/test_keras_model_export.py --large
+
+  autologging:
+    minimum: "2.2.5"
+    maximum: "2.4.3"
+    requirements:
+      "< 2.3.1": ["tensorflow==1.15.4", "h5py<3.0"]
+      "== 2.3.1": ["tensorflow==2.2.1"]
```
dbczumar, line 102: Was there a particular reason that Keras `2.3.1` depends on exactly TF `2.2.1`? (Sorry if we already discussed this)

ml-package-versions.yml
```diff
@@ -0,0 +1,151 @@
...
+      # In keras >= 2.4.0, `keras` simply redirects to `tensorflow.keras`:
+      # https://github.com/keras-team/keras#multi-backend-keras-and-tfkeras
+      pip install tf-nightly 'keras>=2.4.0'
+
+  models:
+    minimum: "2.0.0"
+    maximum: "2.4.3"
+    requirements:
+      "< 2.3.1": ["tensorflow==1.15.4", "h5py<3.0", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
+      "== 2.3.1": ["tensorflow==2.2.1", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
```
dbczumar, line 92: Was there a particular reason that Keras `2.3.1` depends on exactly TF `2.2.1`? (Sorry if we already discussed this). If so, can we add an inline comment explaining why?

ml-package-versions.yml
harupy: Added a comment on why `tf 2.2.1` is required for `keras 2.3.1` 👍

ml-package-versions.yml
harupy: Thanks for the catch. Addressed this :)

dev/set_matrix.py
```diff
@@ -0,0 +1,489 @@
...
+        flavor = flavor_key.split("-")[0]
+        package_info = cfgs.pop("package_info")
+
+        for key, cfg in cfgs.items():
+            # Released versions
+            versions = get_released_versions(package_info["pip_release"])
+            versions = filter_versions(
+                versions, cfg["minimum"], cfg["maximum"], cfg.get("unsupported"),
+            )
+            versions = select_latest_micro_versions(versions)
```
dbczumar, line 419: @harupy We also want to explicitly include the minimum supported version in our testing procedure. For example, `scikit-learn` tests should run against `0.20.3` *and* `0.20.4`.

dev/set_matrix.py
```diff
@@ -0,0 +1,489 @@
...
+                return yaml.load(f, Loader=yaml.SafeLoader)
+    except Exception as e:
+        if if_error is not None:
+            print("Failed to read '{}' due to: `{}`".format(location, e))
+            return if_error
+
+        raise
+
+
+@functools.lru_cache(maxsize=None)
```
dbczumar, line 81: What's the benefit of the lru cache here?

dbczumar (approved):
@harupy This is amazing! LGTM once https://github.com/mlflow/mlflow/pull/3731/files#r537840310 is addressed.

dev/set_matrix.py
harupy: It's just for a speedup. If we didn't cache, we'd query `https://pypi.python.org/pypi/{package_name}/json` four times (("models" + "autologging") x (`config` + `config_ref`)).

harupy: I'll remove this because the benefit is too subtle.

PR #3735, opened 2020-11-25
Unpin TF / Keras and fix bugs in autologging integrations for newer versions

dbczumar:

This PR fixes a number of issues in MLflow's TensorFlow & Keras autologging integrations and unpins these libraries for testing purposes in order to catch regressions in the future.

## How is this patch tested?

MLflow autologging tests

## Release Notes

Bug fixes for autologging compatibility with the latest versions of TensorFlow and Keras

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/fastai.py
+++ b/mlflow/fastai.py
@@ -405,7 +405,7 @@ def autolog(log_models=True):
             callbacks += list(args[callback_arg_index])
             tmp_list[callback_arg_index] += [__MLflowFastaiCallback(self)]
             args = tuple(tmp_list)
-        elif "callbacks" in kwargs:
+        elif kwargs.get("callbacks"):
             callbacks += list(kwargs["callbacks"])
             kwargs["callbacks"] += [__MLflowFastaiCallback(self)]
         else:
--- a/mlflow/fastai.py
+++ b/mlflow/fastai.py
@@ -395,7 +395,7 @@ def autolog(log_models=True):
         else:
             auto_end_run = False
 
-        log_fn_args_as_params(original, [self] + list(args), kwargs, unlogged_params)
+        log_fn_args_as_params(original, list(args), kwargs, unlogged_params)
 
         callbacks = [cb(self) for cb in self.callback_fns] + (self.callbacks or [])
 
--- a/mlflow/keras.py
+++ b/mlflow/keras.py
@@ -728,7 +728,7 @@ def autolog(log_models=True):
             early_stop_callback = _early_stop_check(tmp_list[callback_arg_index])
             tmp_list[callback_arg_index] += [__MLflowKerasCallback()]
             args = tuple(tmp_list)
-        elif "callbacks" in kwargs:
+        elif kwargs.get("callbacks"):
             early_stop_callback = _early_stop_check(kwargs["callbacks"])
             kwargs["callbacks"] += [__MLflowKerasCallback()]
         else:
@@ -751,9 +751,20 @@ def autolog(log_models=True):
         return _run_and_log_function(self, original, args, kwargs, unlogged_params, 5)
 
     def fit_generator(self, *args, **kwargs):
+        """
+        NOTE: `fit_generator()` is deprecated in Keras >= 2.4.0 and simply wraps `fit()`.
+        To avoid unintentional creation of nested MLflow runs caused by a patched
+        `fit_generator()` method calling a patched `fit()` method, we only patch
+        `fit_generator()` in Keras < 2.4.0.
+        """
         original = gorilla.get_original_attribute(keras.Model, "fit_generator")
         unlogged_params = ["self", "generator", "callbacks", "validation_data", "verbose"]
         return _run_and_log_function(self, original, args, kwargs, unlogged_params, 4)
 
     wrap_patch(keras.Model, "fit", fit)
-    wrap_patch(keras.Model, "fit_generator", fit_generator)
+    # `fit_generator()` is deprecated in Keras >= 2.4.0 and simply wraps `fit()`.
+    # To avoid unintentional creation of nested MLflow runs caused by a patched
+    # `fit_generator()` method calling a patched `fit()` method, we only patch
+    # `fit_generator()` in Keras < 2.4.0.
+    if LooseVersion(keras.__version__) < LooseVersion("2.4.0"):
+        wrap_patch(keras.Model, "fit_generator", fit_generator)
--- a/mlflow/tensorflow.py
+++ b/mlflow/tensorflow.py
@@ -1017,10 +1017,10 @@ def autolog(every_n_iter=100, log_models=True):
 
     def fit_generator(self, *args, **kwargs):
         """
-        NOTE: `fit_generator()` is deprecated in TF > 2.0.0 and simply wraps `fit()`.
+        NOTE: `fit_generator()` is deprecated in TF >= 2.1.0 and simply wraps `fit()`.
         To avoid unintentional creation of nested MLflow runs caused by a patched
         `fit_generator()` method calling a patched `fit()` method, we only patch
-        `fit_generator()` in TF <= 2.0.0.
+        `fit_generator()` in TF < 2.1.0.
         """
         with _manage_active_run():
             original = gorilla.get_original_attribute(tensorflow.keras.Model, "fit_generator")
@@ -1068,11 +1068,11 @@ def autolog(every_n_iter=100, log_models=True):
         (tensorflow.estimator.Estimator, "export_savedmodel", export_savedmodel),
         (FileWriter, "add_summary", add_summary),
     ]
-    if LooseVersion(tensorflow.__version__) <= LooseVersion("2.0.0"):
-        # `fit_generator()` is deprecated in TF > 2.0.0 and simply wraps `fit()`.
+    if LooseVersion(tensorflow.__version__) < LooseVersion("2.1.0"):
+        # `fit_generator()` is deprecated in TF >= 2.1.0 and simply wraps `fit()`.
         # To avoid unintentional creation of nested MLflow runs caused by a patched
         # `fit_generator()` method calling a patched `fit()` method, we only patch
-        # `fit_generator()` in TF <= 2.0.0
+        # `fit_generator()` in TF < 2.1.0
         patches.append((tensorflow.keras.Model, "fit_generator", fit_generator))
 
     for p in patches:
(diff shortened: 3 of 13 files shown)


mlflow/keras.py
```diff
@@ -728,7 +728,7 @@ def _run_and_log_function(self, original, args, kwargs, unlogged_params, callbac
             early_stop_callback = _early_stop_check(tmp_list[callback_arg_index])
             tmp_list[callback_arg_index] += [__MLflowKerasCallback()]
             args = tuple(tmp_list)
-        elif "callbacks" in kwargs:
+        elif kwargs.get("callbacks"):
```
dbczumar, line 731: `kwargs["callbacks"]` may still be `None`, as this is the default value for the argument. For example, `fit_generator()` calls `fit()` with `callbacks=None` in Keras >= 2.4.0. Many of our other autologging integrations address this case, but this one does not.

tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -221,7 +221,12 @@ def tf_keras_random_data_run_with_callback(
             restore_best_weights=restore_weights,
         )
     else:
-        callback = tf.keras.callbacks.ProgbarLogger(count_mode="samples")
+
+        class CustomCallback(tf.keras.callbacks.Callback):
```
dbczumar, line 225: As per https://github.com/tensorflow/tensorflow/issues/38618, manually constructing `ProgbarLogger` produces an infinity-related arithmetic error on recent versions of Keras. Accordingly, we swap in a simpler, valid callback here instead.

tests/keras_autolog/test_keras_autolog.py
```diff
@@ -186,18 +186,19 @@ def keras_random_data_run_with_callback(
...
             restore_best_weights=restore_weights,
         )
     else:
-        if fit_variant == "fit_generator":
-            count_mode = "steps"
-        else:
-            count_mode = "samples"
-        callback = keras.callbacks.callbacks.ProgbarLogger(count_mode=count_mode)
+
+        class CustomCallback(keras.callbacks.Callback):
```
dbczumar, line 197: As per https://github.com/tensorflow/tensorflow/issues/38618, manually constructing `ProgbarLogger` produces an infinity-related arithmetic error on recent versions of Keras. Accordingly, we swap in a simpler, valid callback here instead.

tests/keras_autolog/test_keras_autolog.py
```diff
@@ -186,18 +186,19 @@ def keras_random_data_run_with_callback(
     model = create_model()
     if callback == "early":
         # min_delta is set as such to guarantee early stopping
-        callback = keras.callbacks.callbacks.EarlyStopping(
+        callback = keras.callbacks.EarlyStopping(
```
dbczumar, line 189: `keras.callbacks.EarlyStopping` is valid for the versions of Keras we support, whereas `keras.callbacks.callbacks.EarlyStopping` was removed / renamed in 2.4.0.

mlflow/utils/autologging_utils.py
```diff
@@ -70,39 +36,27 @@ def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W01
...
-        for name in [name for name in defaults.keys() if name in unlogged]:
-            del defaults[name]
-        try_mlflow_log(mlflow.log_params, defaults)
-
-    # Logging the arguments passed by the user
-    args_dict = dict(
-        (param_name, param_val)
-        for param_name, param_val in zip(all_param_names, args)
-        if param_name not in unlogged
+    param_spec = inspect.signature(fn).parameters
```
dbczumar, line 39: According to https://docs.python.org/3/library/inspect.html#inspect.getfullargspec, `inspect.signature()` is the recommended utility for method signature inspection in Python 3. `getfullargspec` is kept for compatibility with Python 2 code, which is no longer a concern for us. Further `inspect.getfullargspec` does not return accurate signature information for the latest versions of TensorFlow and Keras.

Accordingly, we drop in `inspect.signature()` in place of the previous `getfullargspec` and clean up the logic a bit.

mlflow/tensorflow.py
```diff
@@ -1058,11 +1064,16 @@ def add_summary(self, *args, **kwargs):
         (EventFileWriterV2, "add_event", add_event),
         (tensorflow.estimator.Estimator, "train", train),
         (tensorflow.keras.Model, "fit", fit),
-        (tensorflow.keras.Model, "fit_generator", fit_generator),
         (tensorflow.estimator.Estimator, "export_saved_model", export_saved_model),
         (tensorflow.estimator.Estimator, "export_savedmodel", export_savedmodel),
         (FileWriter, "add_summary", add_summary),
     ]
+    if LooseVersion(tensorflow.__version__) <= LooseVersion("2.0.0"):
+        # `fit_generator()` is deprecated in TF > 2.0.0 and simply wraps `fit()`.
```
dbczumar, line 1072: Ideally, in the future, we will make use of a utility similar to scikit-learn's `_SklearnTrainingSession` to guard against nested runs in other ML autologging integrations. For now, this appears to be a quick & perfectly valid fix.

mlflow/keras.py
```diff
@@ -751,9 +751,20 @@ def fit(self, *args, **kwargs):
...
+        `fit_generator()` method calling a patched `fit()` method, we only patch
+        `fit_generator()` in Keras < 2.4.0.
+        """
         original = gorilla.get_original_attribute(keras.Model, "fit_generator")
         unlogged_params = ["self", "generator", "callbacks", "validation_data", "verbose"]
         return _run_and_log_function(self, original, args, kwargs, unlogged_params, 4)
 
     wrap_patch(keras.Model, "fit", fit)
-    wrap_patch(keras.Model, "fit_generator", fit_generator)
+    # `fit_generator()` is deprecated in Keras >= 2.4.0 and simply wraps `fit()`.
```
dbczumar, line 765: Ideally, in the future, we will make use of a utility similar to scikit-learn's `_SklearnTrainingSession` to guard against nested runs in other ML autologging integrations. For now, this appears to be a quick & perfectly valid fix.

dbczumar, line 765: Interestingly, Keras >= 2.4.0 pulls in its `Model` class definition from `tensorflow.keras` with a dependency on `tensorflow >= 2.2.0`. This explains why `fit_generator()` is deprecated in Keras 2.4.0, since `tensorflow` deprecated it in TensorFlow 2.1.0.

mlflow/fastai.py
```diff
@@ -405,7 +405,7 @@ def _run_and_log_function(self, original, args, kwargs, unlogged_params, callbac
             callbacks += list(args[callback_arg_index])
             tmp_list[callback_arg_index] += [__MLflowFastaiCallback(self)]
             args = tuple(tmp_list)
-        elif "callbacks" in kwargs:
+        elif kwargs.get("callbacks"):
```
dbczumar, line 408: Fixing this logical error since I caught it in https://github.com/mlflow/mlflow/pull/3735/files#r530146856 as well

mlflow/tensorflow.py
```diff
@@ -994,7 +994,7 @@ def fit(self, *args, **kwargs):
                 early_stop_callback = _early_stop_check(tmp_list[5])
                 tmp_list[5], log_dir = _setup_callbacks(tmp_list[5], log_models)
                 args = tuple(tmp_list)
-            elif "callbacks" in kwargs:
+            elif kwargs.get("callbacks"):
```
dbczumar, line 997: `kwargs["callbacks"]` may still be `None`, as this is the default value for the argument. For example, `fit_generator()` calls `fit()` with `callbacks=None` in TensorFlow >= 2.1.0. Many of our other autologging integrations address this case, but this one does not.

harupy (approved):
LGTM once the lint errors and `ImportError` for `get_unspecified_default_args` are fixed :)

harupy:
This error:

https://github.com/mlflow/mlflow/pull/3735/checks?check_run_id=1452107744#step:4:1531

seems related to this change:

> Metrics and losses are now reported under the exact name specified by the user (e.g. if you pass metrics=['acc'], your metric will be reported under the string "acc", not "accuracy", and inversely metrics=['accuracy'] will be reported under the string "accuracy".

https://github.com/keras-team/keras/releases/tag/2.3.0

mlflow/utils/autologging_utils.py
```diff
@@ -70,39 +36,27 @@ def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W01
...
+    # Fetch the parameter names for specified positional arguments from the function
+    # signature & create a mapping from positional argument name to specified value
+    params_to_log = {
+        param_info.name: param_val
+        for param_info, param_val in zip(list(param_spec.values())[len(args) :], args)
+    }
```
harupy, line 45:
```suggestion
    params_to_log = {
        param_info.name: param_val
        for param_info, param_val in zip(list(param_spec.values()), args)
    }
```

Should `[len(args) :]` here be removed?

dbczumar: I don't think so, unless I'm missing something. `param_spec.values()` contains all of the methods parameters in order: every positional argument and keyword argument. To find parameter names for only the user-specified `args`, we should slice the first `len(args)` arguments, which are guaranteed to be positional arguments (or positional-or-keyword type arguments). Let me know if this makes sense.

dbczumar: Ah, this should be `:len(args)` though.

dbczumar: Fixed and added a test case: `test_tf_keras_autolog_names_positional_parameters_correctly`

mlflow/utils/autologging_utils.py
```diff
@@ -70,39 +36,31 @@ def log_fn_args_as_params(fn, args, kwargs, unlogged=[]):  # pylint: disable=W01
...
-
-    # Logging the arguments passed by the user
-    args_dict = dict(
-        (param_name, param_val)
-        for param_name, param_val in zip(all_param_names, args)
-        if param_name not in unlogged
+    param_spec = inspect.signature(fn).parameters
+    # Filter out `self` from the signature under the assumption that it is not specified
+    # as an explicit parameter in the call to `fn`
+    relevant_params = [param for param in param_spec.values() if param.name != "self"]
```
dbczumar, line 42: Not a huge fan of this change (i.e. what if `self` is actually the first element of `args`?). Will revisit this tomorrow.

dbczumar: Seems acceptable - callers can filter out `self` via the patched function signature. I've documented the expectation that `self` is not a part of the `args` passed to this function.

PR #3753, opened 2020-11-30
PyTorch model snippets 

dmatrix:

This is our continued effort to add demonstrative code snippets for the MLflow Model flavor APIs

Components 
- [ x] `area/docs`: MLflow documentation pages

- [ x] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section


mlflow/pytorch/__init__.py
```diff
@@ -176,51 +252,89 @@ def log_model(
...
-            mlflow.pytorch.log_model(scripted_pytorch_model, "models")
+            mlflow.pytorch.log_model(scripted_pytorch_model, "scripted_models_pth")
+
+        # fetch the logged model artifacts
+        print("--")
+        print("run_id: {}".format(run.info.run_id))
+        for artifact_path in ["models_pth/data", "scripted_models_pth/data"]:
+            artifacts = [f.path for f in MlflowClient().list_artifacts(run.info.run_id,
+                        artifact_path)]
+            print("artifacts: {}".format(artifacts))
```
harupy, line 318: Maybe just showing these lines should be enough?

harupy: We could use `...` too.

```python
class MyModel(str):
    ...


model = MyModel()

# train the model
...

with mlflow.start_run():
    mlflow.pytorch.log_model(model, "model")
```

dmatrix: Thanks for the input. For all but `log_model()` and `autolog_log()`,  I have elided the code. However, for the other two methods, I think, it's good to have a full example that they can try.

mlflow/pytorch/__init__.py
```diff
@@ -46,8 +46,47 @@
...
+        ...
+
+        # Log PyTorch model
+        with mlflow.start_run() as run:
+            mlflow.pytorch.log_model(model, "models_pth")
+
+        # Fetch the associated conda environment
+        print("--")
+        print("run_id: {}".format(run.info.run_id))
+        env = mlflow.pytorch.get_default_conda_env()
```
harupy, line 77: I think just showing `get_default_conda_env` should be enough.

mlflow/pytorch/__init__.py
```diff
@@ -343,24 +418,59 @@ def save_model(
...
-        pytorch_model_path = ...
-        # train our model
-        for epoch in range(500):
-            y_pred = pytorch_model(x_data)
-            ...
-        # Save the model
+
+        # Class defined here
+        class LinearNNModel(torch.nn.Module):
+        ...
```
harupy, line 429:
```suggestion
        class LinearNNModel(torch.nn.Module):
             ...
```

mlflow/pytorch/__init__.py
```diff
@@ -176,51 +215,87 @@ def log_model(
...
+            print("artifacts: {}".format(artifacts))
+
+    .. code-block:: text
+        :caption: Output
+
+        Epoch: 50.0/250, loss: 0.0003
+        Epoch: 100.0/250, loss: 0.0003
+        Epoch: 150.0/250, loss: 0.0002
+        Epoch: 200.0/250, loss: 0.0002
+        Epoch: 250.0/250, loss: 0.0002
```
harupy, line 288:
```suggestion
```

I don't think we need to print out the epochs because they are not important information.

mlflow/pytorch/__init__.py
```diff
@@ -343,24 +418,59 @@ def save_model(
...
-            # Saving scripted model
+            # Convert to a scripted model and save it
             scripted_pytorch_model = torch.jit.script(model)
-            mlflow.pytorch.save_model(scripted_pytorch_model, pytorch_model_path)
+            mlflow.pytorch.save_model(scripted_pytorch_model, "scripted_models_pth")
+
+        # Load each saved model for inference
+        print("--")
+        for model_path in ["models_pth", "scripted_models_pth"]:
+            model_uri = "{}/{}".format(os.getcwd(), model_path)
```
harupy, line 449: Is this line really correct? `"{}/{}".format(os.getcwd(), model_path)` doesn't seem to be the right path.

mlflow/pytorch/__init__.py
```diff
@@ -539,14 +649,44 @@ def load_model(model_uri, **kwargs):
...
+
+        # Training loop
+        ...
+
+        # Log the model
+        with mlflow.start_run() as run:
+            mlflow.pytorch.log_model(model, "models_pth")
+
+        # Inference after loading the logged model
+        print("--")
```
harupy, line 675:
```suggestion
```

mlflow/pytorch/__init__.py
```diff
@@ -539,14 +649,44 @@ def load_model(model_uri, **kwargs):
     .. code-block:: python
         :caption: Example
 
+        import warnings
+        import os
```
harupy, line 653:
```suggestion
```

Can we remove these because they don't seem to  be used?

harupy:
Merging the current master branch should fix the issue in the CircleCI build :)

mlflow/pytorch/__init__.py
dmatrix: I can remove it from the output, but it makes sense to keep it in the code; it's a good indication that your model is merging as the loss is decreasing.

mlflow/pytorch/__init__.py
dmatrix: Yes, it [seems to work.](https://github.com/dmatrix/mlflow-tests/blob/master/py/mlflow/models/pytorch/save_model.py) The saving model saves it under the local path.

mlflow/pytorch/__init__.py
dmatrix: As cut-n-paste leftover over from [here](https://github.com/dmatrix/mlflow-tests/blob/master/py/mlflow/models/pytorch/log_model.py). :)

mlflow/pytorch/__init__.py
harupy: If this was a tutorial about training a pytorch model, it would make sense to keep the epoch and loss.

dmatrix: I suppose one can always add it if one wants to. Will remove it.

mlflow/pytorch/__init__.py
```diff
@@ -46,8 +46,43 @@
...
+        # Initialize our model, criterion and optimizer
+        model = LinearNNModel()
+        ...
+
+        # Training loop
+        ...
+
+        # Log PyTorch model
+        with mlflow.start_run() as run:
+            mlflow.pytorch.log_model(model, "models_pth")
```
harupy, line 72:
```suggestion
        import mlflow.pytorch
```

Can we remove unrelated code?

mlflow/pytorch/__init__.py
harupy: I think this is because the artifact directory points to the current working directory.

mlflow/pytorch/__init__.py
```diff
@@ -539,14 +630,39 @@ def load_model(model_uri, **kwargs):
...
+        ...
+
+        # Log the model
+        with mlflow.start_run() as run:
+            mlflow.pytorch.log_model(model, "models_pth")
+
+        # Inference after loading the logged model
+        model_uri = "runs:/{}/models_pth".format(run.info.run_id)
+        loaded_model = mlflow.pytorch.load_model(model_uri)
+        for hv in [4.0, 6.0, 30.0]:
```
harupy, line 655: Can we pass `[4.0, 6.0, 30.0]` to the model as a tensor?

```
model(xxx_yyy([4.0, 6.0, 30.0]))
```

mlflow/pytorch/__init__.py
dmatrix: Is there anything wrong with it If the current artifact directory points to cwd and it's where you wish to save the model?

mlflow/pytorch/__init__.py
dmatrix: You're minimalist, aren't you? :-) 

Why do you feel that this code is unrelated? This provides some context, so If they want to try a full example, then can
use look at the full example above. At the least, we should have:

        # Class defined here
        class LinearNNModel(torch.nn.Module):
            ...
        # Initialize our model, criterion and optimizer
        ...

        # Training loop
        ...

mlflow/pytorch/__init__.py
dmatrix: You can do something like this since the network expects the input as shape (n, 1) tensor.

```
X = torch.tensor([4, 6, 30]).view(-1,1).float()
y_pred = loaded_model(X)
print(X)
print(y_pred)
```

But I prefer to simplify and send individual tensor so we can print the X, --> y on the same line. Yes, for larger models
you want your input to be the whole batch, but for here I just made it simple: inferring one tensor at a time.

mlflow/pytorch/__init__.py
harupy: If the current artifact directory didn't point to cwd, it would fail. I think it's safe to use `mlflow.get_artifact_uri`.

harupy: or we could use `runs:{}/{}` format.

mlflow/pytorch/__init__.py
harupy: If we had:

```
        # Class defined here
        class LinearNNModel(torch.nn.Module):
            ...
        # Initialize our model, criterion and optimizer
        ...

        # Training loop
        ...
```

it would look like the class definition and training loop affect the result of `get_deafult_conda_env`.

mlflow/pytorch/__init__.py
```diff
@@ -343,24 +401,55 @@ def save_model(
...
-            mlflow.pytorch.save_model(scripted_pytorch_model, pytorch_model_path)
+            mlflow.pytorch.save_model(scripted_pytorch_model, "scripted_models_pth")
+
+        # Load each saved model for inference
+        for model_path in ["models_pth", "scripted_models_pth"]:
+            model_uri = "{}/{}".format(os.getcwd(), model_path)
+            loaded_model = mlflow.pytorch.load_model(model_uri)
+            print("model: {}".format(model_path))
+            for hv in [6.0, 8.0, 12.0, 30.0]:
+                hour_var = torch.Tensor([[hv]])
```
harupy, line 434: Can we rename `hour_var` to `X`?

mlflow/pytorch/__init__.py
```diff
@@ -343,24 +401,55 @@ def save_model(
...
+                y_pred = loaded_model(hour_var)
+                print("predict X:{}, y_pred: {:.2f}".format(hv, y_pred.data.item()))
```
harupy, line 436:
```suggestion
                print("predict X: {}, y_pred: {:.2f}".format(hv, y_pred.data.item()))
```

nit

mlflow/pytorch/__init__.py
dmatrix: Ah, I see what you mean! 

(The reason I had it is, because if you had just trained and logged a PyTorch model in the elided code above, then you can fetch its conda environment with this API code.)

I can remove it.

mlflow/pytorch/__init__.py
dmatrix: Yes, indeed!

mlflow/pytorch/__init__.py
```diff
@@ -343,24 +388,55 @@ def save_model(
...
+            mlflow.pytorch.save_model(model, "models_pth")
 
-            # Saving scripted model
+            # Convert to a scripted model and save it
             scripted_pytorch_model = torch.jit.script(model)
-            mlflow.pytorch.save_model(scripted_pytorch_model, pytorch_model_path)
+            mlflow.pytorch.save_model(scripted_pytorch_model, "scripted_models_pth")
+
+        # Load each saved model for inference
+        for model_path in ["models_pth", "scripted_models_pth"]:
```
harupy, line 416: Why do we suffix directorie with `_pth`?

mlflow/pytorch/__init__.py
harupy: sorry I misunderstood this is `log_model`.

mlflow/pytorch/__init__.py
dmatrix: Good observation and point! No need to adopt the PyTorch suffix for model directories. We use generic "models," so better to stick with "models" and scripted_models" respectively.

mlflow/pytorch/__init__.py
```diff
@@ -46,8 +46,30 @@
...
              :func:`save_model()` and :func:`log_model()`.
+
+    .. code-block:: python
+        :caption: Example
+
+        import mlflow.pytorch
+
+        # Log PyTorch model
+        with mlflow.start_run() as run:
+            mlflow.pytorch.log_model(model, "models_pth")
```
harupy, line 59: Should we remove `_pth` here?

mlflow/pytorch/__init__.py
```diff
@@ -539,14 +614,38 @@ def load_model(model_uri, **kwargs):
...
+            X = torch.Tensor([[x]])
+            y_pred = loaded_model(X)
+            print("predict X: {}, y_pred: {:.2f}".format(x, y_pred.data.item()))
+
+    .. code-block:: text
+        :caption: Output
+
+        predict X:4.0, y_pred: 7.57
+        predict X:6.0, y_pred: 11.64
+        predict X:30.0, y_pred: 60.48
```
harupy, line 648:
```suggestion
        predict X: 30.0, y_pred: 60.48
```
nit

harupy:
LGTM once https://github.com/mlflow/mlflow/pull/3753#discussion_r538973300 and https://github.com/mlflow/mlflow/pull/3753#discussion_r538973418 have been addressed.

mlflow/pytorch/__init__.py
dmatrix: Removed other instances and usage of `_pth` suffix 👍

harupy (approved):

harupy:
https://github.com/mlflow/mlflow/pull/3802 fixes failures in the R and flavors checks.

PR #3793, opened 2020-12-08
test cross-version-tests

harupy:


ml-package-versions.yml
```diff
@@ -152,3 +153,14 @@ lightgbm:
...
+unknown:
+  package_info:
+    pip_release: "lightgbm"
+
+  models:
+    minimum: "2.3.1"
+    maximum: "2.3.1"
+    requirements: ["scikit-learn"]
+    run: |
+      pytest tests/lightgbm/test_lightgbm_model_export.py --large
```
harupy, line 166: new flavor

mlflow/xgboost.py
```diff
@@ -524,3 +524,4 @@ def infer_model_signature(input_example):
 
     wrap_patch(xgboost, "train", train)
     wrap_patch(xgboost.DMatrix, "__init__", __init__)
+
```
harupy, line 527: flavor file change

ml-package-versions.yml
```diff
@@ -93,6 +94,7 @@ keras:
       "> 2.3.1": ["tensorflow", "scikit-learn", "pyspark==2.4.0", "pyarrow"]
       "== dev": ["scikit-learn", "pyspark==2.4.0", "pyarrow"]
     run: |
+      # just a comment (should not re-run this item)
```
harupy, line 97: This should not trigger return the (keras, models) tests

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -249,3 +249,4 @@ def test_pytorch_test_metrics_logged(pytorch_model_tests):
     data = run.data
     assert "test_loss" in data.metrics
     assert "test_acc" in data.metrics
+
```
harupy, line 252: Changes in test files should trigger the corresponding flavor tests.

harupy:
@dbczumar Seems to be working as we expected.

PR #3798, opened 2020-12-08
Update schema enforcement

tomasatdatabricks:
Integers in python can not encode missing values (there is no NaN alternative). Therefore, integer data with missing values is typically represented as float64. As a result, datatype of integer columns can vary across data samples. This can lead to schema enforcement errors at runtime. For example, the following will raise:
Infer schema on dataset with integer column and no missing values
Attempt to score data that has the value for the integer column missing (null)
The opposite situation (infer with missing values, score on data with no missing values) will raise as well.

- Update documentation, highlight this issue and suggest using double instead of int/long as a workaround
- Improve error messages whenever possible
- Allow conversion from integers (int32) to double.

## How is this patch tested?
updated schema enforcement unit tests.

## Release Notes
N/A

Components 

Interface 

Language 

Integrations


mlflow/pyfunc/__init__.py
```diff
@@ -310,7 +311,7 @@ def _enforce_type(name, values: pandas.Series, t: DataType):
                 "Failed to convert column {0} from type {1} to {2}.".format(name, values.dtype, t)
             )
 
-    if values.dtype in (t.to_pandas(), t.to_numpy()):
+    if t.to_pandas() == values.dtype or t.to_numpy() == values.dtype:
```
tomasatdatabricks, line 314: NB: comparison of pandas and numpy data type fails when numpy data type is on the left hand side of the comparison operator. It works, however, if pandas type is on the left hand side. That is because pandas is aware of numpy.

mlflow/pyfunc/__init__.py
```diff
@@ -321,17 +322,46 @@ def _enforce_type(name, values: pandas.Series, t: DataType):
         return values
 
     numpy_type = t.to_numpy()
-    is_compatible_type = values.dtype.kind == numpy_type.kind
-    is_upcast = values.dtype.itemsize <= numpy_type.itemsize
-    if is_compatible_type and is_upcast:
+    if values.dtype.kind == numpy_type.kind:
+        is_upcast = values.dtype.itemsize <= numpy_type.itemsize
+    elif values.dtype.kind == "u" and numpy_type.kind == "i":
```
tomasatdatabricks, line 327: Allow cast of unsigned int to a (larger) signed int type. For example, uint8 can be safely converted to int32.

mlflow/pyfunc/__init__.py
```diff
@@ -321,17 +322,47 @@ def _enforce_type(name, values: pandas.Series, t: DataType):
...
         return values.astype(numpy_type, errors="raise")
     else:
         # NB: conversion between incompatible types (e.g. floats -> ints or
         # double -> float) are not allowed. While supported by pandas and numpy,
         # these conversions alter the values significantly.
+        def all_ints(xs):
+            for x in xs:
+                if not pandas.isnull(x) and int(x) != x:
+                    return False
+            return True
```
dbczumar, line 345:
```suggestion
        return all([pandas.isnull(x) or int(x) == x])
```

mlflow/pyfunc/__init__.py
```diff
@@ -321,17 +322,47 @@ def _enforce_type(name, values: pandas.Series, t: DataType):
...
-    is_compatible_type = values.dtype.kind == numpy_type.kind
-    is_upcast = values.dtype.itemsize <= numpy_type.itemsize
-    if is_compatible_type and is_upcast:
+    if values.dtype.kind == numpy_type.kind:
+        is_upcast = values.dtype.itemsize <= numpy_type.itemsize
+    elif values.dtype.kind == "u" and numpy_type.kind == "i":
+        is_upcast = values.dtype.itemsize < numpy_type.itemsize
+    elif values.dtype.kind in ("i", "u") and numpy_type == np.float64:
+        # allow (u)int => double conversion
+        is_upcast = values.dtype.itemsize <= 6
```
dbczumar, line 331: Why is `6` the threshold here? I know we need < 8. Is 7 bad (does a 7 byte dtype even exist?). Maybe an explanatory comment would help.

docs/source/models.rst
```diff
@@ -146,8 +146,22 @@ names, matching is done by position (i.e. MLflow will only check the number of c
...
+Integer data with missing values is typically represented as floats in Python. Therefore, data
+types of integer columns in Python can vary depending on the data sample. This type variance can
+cause schema enforcement errors at runtime since integer and float are not compatible types. For
+example, if your training data did not have any missing values for integer column c, its type will
+be integer. However, when you attempt to score a sample of the data that does include a missing
+value in column c, its type will be float. If your model signature specified c to have integer type,
+MLflow will raise an error since it can not convert float to int. Note that MLflow uses python to
+serve models and to deploy models to Spark, so this can affect most model deployments. The best way
+to avoid this problem is to declare integer columns as doubles (float64) whenever there can be
+missing values.
```
dbczumar, line 164: Can we also say that it's best to pass in examples with missing values when using `infer_signature` to ensure that the correct data type is captured?

tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
```diff
@@ -182,34 +182,55 @@ def predict(pdf):
...
+
+    # 9. long -> double raises
     pdf["d"] = pdf["d"].astype(np.int64)
     with pytest.raises(MlflowException) as ex:
         pyfunc_model.predict(pdf)
     assert "Incompatible input types" in str(ex)
     pdf["d"] = pdf["d"].astype(np.float64)
 
-    # 6. ints -> floats raises
+    # 10. any float -> any int raises
```
dbczumar, line 233: For completeness, do we need to test `double -> long` and `double -> float`?

dbczumar (commented):
@tomasatdatabricks Left a few comments. Once these are addressed in addition to adding the warning we discussed offline to `infer_signature`, this LGTM! Thanks!

mlflow/pyfunc/__init__.py
tomasatdatabricks: Double has 52 bits of mantissa, so the size has to be < 52. 6 is the smallest multiple of 8 smaller than 52. I guess we don't have to be that specific and just say integers or smaller.

mlflow/pyfunc/__init__.py
tomasatdatabricks: that makes sense!

tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
tomasatdatabricks: We might as well

docs/source/models.rst
tomasatdatabricks: Makes sense. Let me rephrase it.

dbczumar (approved):
LGTM once return & if / else logic in `_infer_schema` is fixed!

PR #3808, opened 2020-12-10
Add support for np array-based tensors for Keras, Pytorch and TF flavors

wentinghu:
Signed-off-by: Wendy Hu <wendy.hu@databricks.com>

## How is this patch tested?

- Added PyFunc tests
- Added tests for each affected flavor

Added support for np array-based input tensors for Keras. PyTorch and TF flavors
- pyfunc keras models now accept pd.DataFrames, dicts, lists, and np.ndarrays
- pyfunc pytorch models now accept pd.DataFrames and np.ndarrays
- TF (v1 and v2) pytorch models now accept pd.DataFrames and dicts

Components 

Interface 

Language 

Integrations


mlflow/keras.py
```diff
@@ -449,15 +451,18 @@ def __init__(self, keras_model, graph, sess):
...
+    def predict(self, data):
         # In TensorFlow < 2.0, we use a graph and session to predict
+        input_data = data
+        if isinstance(data, pd.DataFrame):
+            input_data = data.values
         if self._graph is not None:
             with self._graph.as_default():
                 with self._sess.as_default():
-                    predicted = pd.DataFrame(self.keras_model.predict(dataframe.values))
+                    predicted = pd.DataFrame(self.keras_model.predict(input_data))
```
tomasatdatabricks, line 462: I think that we only want to return DataFrame if the input was DataFrame. If the was one of the numpy array based data types we should return the result as is.

mlflow/pyfunc/__init__.py
```diff
@@ -404,24 +421,29 @@ class PyFuncModel(object):
...
         if not hasattr(model_impl, "predict"):
             raise MlflowException("Model implementation is missing required predict method.")
         if not model_meta:
             raise MlflowException("Model is missing metadata.")
         self._model_meta = model_meta
         self._model_impl = model_impl
+        self._supports_tensor_input = supports_tensor_input
 
-    def predict(self, data: pandas.DataFrame) -> PyFuncOutput:
+    def predict(self, data: Any) -> PyFuncOutput:
```
tomasatdatabricks, line 433: Can we create a Union data type similar to the PyFuncOutput?

mlflow/pyfunc/__init__.py
```diff
@@ -404,24 +421,29 @@ class PyFuncModel(object):
...
         """
         input_schema = self._model_meta.get_input_schema()
         if input_schema is not None:
             data = _enforce_schema(data, input_schema)
-        return self._model_impl.predict(data)
+        if isinstance(data, pandas.DataFrame) or (
+            self._supports_tensor_input and _is_supported_tensor(data)
+        ):
+            return self._model_impl.predict(data)
+        raise MlflowException("Provided input data is not of a supported type")
```
tomasatdatabricks, line 446: Can we be more specific when raising exception? 
1. It would be good to know if the input does not work because the model does not support tensor input or the input is not supported altogether. 
2. We should also include the list of supported data types, something like MLflow python function interface supports the following input data: pandas.DataFrame or numpy array for tabular data, numpy array or a list of numpy array or a dictionary of str -> numpy array for tensors. If you are trying to score other data type that is understood by your model, please load your model via appropriate model flavor, e.g. `mlflow.pytorch.load_model`.

mlflow/pyfunc/__init__.py
```diff
@@ -345,8 +346,12 @@ def _enforce_schema(pdf: pandas.DataFrame, input_schema: Schema):
     For column types, we make sure the types match schema or can be safely converted to match the
     input schema.
     """
-    if isinstance(pdf, list):
-        pdf = pandas.DataFrame(pdf)
+    if isinstance(pdf, list) or isinstance(pdf, np.ndarray) or isinstance(pdf, dict):
+        try:
+            pdf = pandas.DataFrame(pdf)
+        except Exception:
+            message = "The input type provided could not be converted into a pandas DataFrame."
```
tomasatdatabricks, line 353: I think we should include that the model contains model signature that suggests DataFrame input.

tomasatdatabricks: And we should also include error message to provide more details.

tests/keras/test_pyfunc_model_works_with_all_input_types.py
```diff
@@ -0,0 +1,158 @@
...
+    assert type(actual) == pd.DataFrame
+    np.testing.assert_allclose(actual.values, expected, rtol=1e-5)
+
+    # Calling predict with a np array should return a np array
+    actual = model_loaded.predict(x.values)
+    assert type(actual) == np.ndarray
+    np.testing.assert_allclose(actual, expected, rtol=1e-5)
+
+    # Calling predict with a dict should make keras raise an error
+    with pytest.raises(Exception):
```
tomasatdatabricks, line 112: nit: we do not have to test this, this is on Keras not us.

tests/keras/test_pyfunc_model_works_with_all_input_types.py
```diff
@@ -0,0 +1,158 @@
...
+        else {"learning_rate": lr}
+    )
+    model.compile(loss="mean_squared_error", optimizer=SGD(**kwargs))
+    model.fit(x.values, y.values)
+    return model
+
+
+@pytest.fixture(scope="module")
+def multi_tensor_input_model_list(data):
+    x, y = data
```
tomasatdatabricks, line 58: nit:It probably does not matter too much for the test, but could we change the input to be 2 tensors of shape(-1,2) instead of (-1, 4) and pass the first 2 columns of the dataframe as input_a and the second 2 as tensor_b?

That way the model should look the same as in other case and it may be easier to check / reason about it if something goes wrong.

tests/keras/test_pyfunc_model_works_with_all_input_types.py
```diff
@@ -0,0 +1,158 @@
...
+    return model
+
+
+@pytest.fixture(scope="module")
+def multi_tensor_input_model_dict(data):
+    x, y = data
+    input_a = Input(shape=(4,))
+    input_b = Input(shape=(4,))
+    output = Dense(1)(Concatenate()([input_a, input_b]))
+    model = Model(inputs={"a": input_a, "b": input_b}, outputs=output)
```
tomasatdatabricks, line 80: Same thing as above, can we make input_a the first 2 and input_b the last to columns of the dataframe?

tests/keras/test_pyfunc_model_works_with_all_input_types.py
```diff
@@ -0,0 +1,158 @@
...
+    # Loading Keras model via PyFunc
+    model_loaded = mlflow.pyfunc.load_model(model_path)
+
+    # Calling predict with a list should return a np.ndarray output
+    actual = model_loaded.predict(test_input)
+    assert type(actual) == np.ndarray
+    np.testing.assert_allclose(actual, expected, rtol=1e-5)
+
+    # Calling predict with a non-list input should raise an error
+    with pytest.raises(Exception):
```
tomasatdatabricks, line 133: I don't think we need to test this.

tests/keras/test_pyfunc_model_works_with_all_input_types.py
```diff
@@ -0,0 +1,158 @@
...
+    model_loaded = mlflow.pyfunc.load_model(model_path)
+
+    # Calling predict with a dict should return a np.ndarray output
+    actual = model_loaded.predict(test_input)
+    assert type(actual) == np.ndarray
+    np.testing.assert_allclose(actual, expected, rtol=1e-5)
+
+    # Calling predict with a non-dict input should raise an error
+    with pytest.raises(Exception):
+        model_loaded.predict([1, 2, 3, 4])
```
tomasatdatabricks, line 158: same as above, we do not need to test this.

mlflow/pyfunc/__init__.py
```diff
@@ -709,7 +715,7 @@ def save_model(
     artifacts=None,
     signature: ModelSignature = None,
     input_example: ModelInputExample = None,
-    **kwargs
+    **kwargs,
```
harupy, line 718: @wentinghu Can you revert this change because this causes a syntax error in Python 3.5?

mlflow/pyfunc/__init__.py
```diff
@@ -345,8 +347,15 @@ def _enforce_schema(pdf: pandas.DataFrame, input_schema: Schema):
     For column types, we make sure the types match schema or can be safely converted to match the
     input schema.
     """
-    if isinstance(pdf, list):
-        pdf = pandas.DataFrame(pdf)
+    if isinstance(pdf, list) or isinstance(pdf, np.ndarray) or isinstance(pdf, dict):
```
harupy, line 350: This line can be simplified to:

```suggestion
    if isinstance(pdf,(list, np.ndarray, dict)):
```

tomasatdatabricks (commented):
Looks great, I left some comments.

mlflow/tensorflow.py
```diff
@@ -541,19 +550,25 @@ def __init__(self, infer):
...
-            # with the shared name. TensorFlow cannot make eager tensors out of pandas
-            # DataFrames, so we convert the DataFrame to a numpy array here.
-            val = df[df_col_name]
-            if isinstance(val, pandas.DataFrame):
-                val = val.values
-            feed_dict[df_col_name] = tensorflow.constant(val)
+        if isinstance(data, dict):
+            feed_dict = {k: tensorflow.constant(v) for k, v in data.items()}
+        elif isinstance(data, pandas.DataFrame):
+            for df_col_name in list(data):
```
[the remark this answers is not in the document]
wentinghu: Oh I think the purpose of this implementation is explained in the comments below. We need to also check if `data[df_col_name]` is also a dataframe (there are multiple columns with the same name) and convert that to a np array instead. If there's only one column, `data[df_col_name]` will have a `pandas.core.series.Series` type, which is fine to pass into `tensorflow.constant` I guess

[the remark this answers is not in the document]
tomasatdatabricks: Oh, interesting, I did not know! Thanks for clarifying this.
I guess this is another way of passing tensors in a Dataframe.

tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
```diff
@@ -267,6 +267,49 @@ def predict(pdf):
     res = pyfunc_model.predict(pdf)
     assert res.dtypes.to_dict() == expected_types
 
+    # 8. np.ndarrays can be converted to dataframe but have no columns
+    with pytest.raises(MlflowException) as ex:
```
tomasatdatabricks, line 271: Can we also test that his works if the schema has no column names?

tomasatdatabricks (approved):
Looks great! I think we should add one minor test case, but otherwise lgtm.

tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
wentinghu: i think this case is tested [here](https://github.com/mlflow/mlflow/pull/3808/files/eaa1adc24505080fcb35a82c7508a3cb9b1d5966#diff-d2db4c32cb1efe50d5cb6196b5e24a807707c07e5806ac8e60c569d6ca3ce8dfR360) under `test_schema_enforcement_no_col_names`

PR #3819, opened 2020-12-11
support MLeap for R

### LINKED ISSUE #1425: MLeap support for R
Save/load mleap bundles.

yitao-li:

This PR implements support for saving and loading MLeap models in R.

Notice unlike other ML model formats, cross-version model portability is something handled by MLeap internally, which is why the `mleap` R package doesn't really bother to offer an option of loading the saved model using a particular version of Mleap library.

At the moment the version of Mleap java library is still saved as part of the model spec, but it's for informational purpose only and is not checked when loading the model.

Also note, because there is no support for loading `MLeap` model in Python yet, this PR will not have test cases of creating `MLeap` model in R and loading it in Python or vice versa.

## How is this patch tested?

Unit test

This version of MLflow has implemented support for loading and saving MLeap models.

Components 

Interface 

Language 

Integrations

closes #1425


mlflow/R/mlflow/DESCRIPTION
```diff
@@ -40,17 +40,20 @@ Imports:
...
+  dplyr,
   keras,
   lintr,
+  sparklyr,
   testthat (>= 2.0.0),
   stringi (< 1.4.4),
   xgboost,
-  h2o
+  h2o,
+  mleap
```
lorenzwalthert, line 54: seems like description is tidy (before you added mlleap). Suggest to run `usethis::use_tidy_description()`

lorenzwalthert: I think it would be great at some point to add pre-commit.com hooks to this repo, then you could use https://lorenzwalthert.github.io/precommit/articles/available-hooks.html#deps-in-desc for this.

mlflow/R/mlflow/DESCRIPTION
```diff
@@ -40,17 +40,20 @@ Imports:
     xml2,
     yaml,
     zeallot
-Suggests: 
+Suggests:
   covr,
   carrier,
+  dplyr,
```
lorenzwalthert, line 46: I vote against adding a suggested dplyr dependency. The 3 calls you use dplyr for in the tests are in MHO not worth the extra dependency (haven't checked if dplyr was in the recursive dependency tree already but still), in particular these can be achieved easily with base R.

lorenzwalthert:
Nice, was looking forward to this one 😊. I suggest you also add the R flavor to mlleap to the docs: https://mlflow.org/docs/latest/models.html#mleap-mleap. Also, I see there is no MLeap Python model (yet), but maybe there are still some tests we could run (maybe not as extensive as for xgboost, but still. From the docs: 

> You can save Spark models in MLflow format with the mleap flavor by specifying the sample_input argument of the mlflow.spark.save_model() or mlflow.spark.log_model() method (recommended)

So could we try to load such a model in R and make sure it does not fail?

mlflow/R/mlflow/R/model-mleap.R
```diff
@@ -0,0 +1,53 @@
...
+  if (dir.exists(path)) unlink(path, recursive = TRUE)
+  dir.create(path)
+
+  model_filename <- "model.zip"
+  model_path <- file.path(path, model_filename)
+  mleap::ml_write_bundle(model, sample_input = sample_input, path = model_path)
+  version <- mleap::mleap_installed_versions()$mleap %>%
+    purrr::map(~ numeric_version(.x)) %>%
+    purrr::reduce(~ (if (.x > .y) .x else .y)) %>%
+    as.character()
```
lorenzwalthert, line 28: I think using `unlist(mleap::mleap_installed_versions()$mleap) %>% max() %>% as.character()` should also work and is probably more readable but it's also taste I guess.

mlflow/R/mlflow/R/model-mleap.R
```diff
@@ -0,0 +1,53 @@
...
+                                                ...) {
+  if (is.null(sample_input)) {
+    stop("`sample_input` is required by MLeap for data schema inference.")
+  }
+
+  assert_pkg_installed("mleap")
+  if (dir.exists(path)) unlink(path, recursive = TRUE)
+  dir.create(path)
+
+  model_filename <- "model.zip"
```
lorenzwalthert, line 22: Quick check did not give me an answer, so: Is model.zip the file name currently used with MLeap models in mlflow? Also, at some point, it might be good to keep model names for all flavors in a variable in the R mlflow namespace instead of defining them ad-hoc once on saving and once on loading.

mlflow/R/mlflow/DESCRIPTION
lorenzwalthert: I suggested to use pre-commit for MLflow: https://github.com/mlflow/mlflow/issues/3823

mlflow/R/mlflow/R/model-mleap.R
yitao-li: I thought so too but unfortunately no: `unlist()` strips the `numeric_version` from the class of the operand, so the rest of the pipe won't work correctly

mlflow/R/mlflow/R/model-mleap.R
yitao-li: For MLeap currently it just needs to be any file name ending with `.zip`

yitao-li:
> Nice, was looking forward to this one . I suggest you also add the R flavor to mleap to the docs: https://mlflow.org/docs/latest/models.html#mleap-mleap. Also, I see there is no MLeap Python model (yet), but maybe there are still some tests we could run (maybe not as extensive as for xgboost, but still. From the docs:
> 
> > You can save Spark models in MLflow format with the mleap flavor by specifying the sample_input argument of the mlflow.spark.save_model() or mlflow.spark.log_model() method (recommended)
> 
> So could we try to load such a model in R and make sure it does not fail?

Yep the current test case covers loading the model with `mlflow_load_model()` in R

mlflow/R/mlflow/R/model-mleap.R
lorenzwalthert: oh right...

lorenzwalthert:
> Yep the current test case covers loading the model with mlflow_load_model() in R. 

Well, I thought that you could use `mlflow.spark.save_model()`, e.g. as shown in the example section of the [`mlflow.spark.save_model()`](https://mlflow.org/docs/latest/python_api/mlflow.spark.html#mlflow.spark.log_model) to log the model from python and load it with R to ensure compatibility between languages.

yitao-li:
Re: compatibility with MLeap models produced by Python -- The implementation in this PR needs to be tweaked a bit to be able to load  models of MLeap flavor produced by Python, because the ones outputted by `mleap.save_model` in Python have a slightly different directory structure.

Also, a bug fix in https://github.com/mlflow/mlflow/pull/3872 needs to happen first before we can work with those 2 slightly different varieties of Mleap models.

lorenzwalthert:
I see. But I think we should aim for full compatibility of the Python and R models if possible. Because the R runtime might not be available in the deployment context for example.

kevinykuo:
Can we test that we can work with saved models here https://github.com/mlflow/mlflow/tree/master/mlflow/java/scoring/src/test/resources/org/mlflow
in the spirit of, e.g., https://github.com/mlflow/mlflow/blob/8e74cdd108ff324175e38d621b12d9b2aced05bd/mlflow/java/scoring/src/test/java/org/mlflow/MLeapPredictorTest.java ?

yitao-li:
@kevinykuo Yes indeed! Once we have the bug fix in https://github.com/mlflow/mlflow/pull/3872 then I can revisit this PR to add the test case you suggested.

yitao-li:
The ci/circleci: build_doc error saying "configure: error: Java interpreter '/usr/lib/jvm/default-java/jre/bin/java' does not work" seems to be unrelated to this change.

kevinykuo:
@yitao-li sorry getting back to this now. Could we re-trigger the build?

yitao-li:
@kevinykuo I just rebased and re-ran the build. Ci/circleci - build_doc_r is still failing with "configure: error: Java interpreter '/usr/lib/jvm/default-java/jre/bin/java' does not work". Everything else is passing.

yitao-li:
@kevinykuo Hey just getting back to this: it looks like the circleci: build_doc_r error can be fixed with https://github.com/mlflow/mlflow/pull/4024 and is definitely unrelated to this change. I think we should merge https://github.com/mlflow/mlflow/pull/4024 first, and then rebase this PR again and wait until all builds pass.

yitao-li:
Update: all builds are passing now, and we don't need changes iin #4024 to be in a separate PR as it turns out to be actually related to this change (because of `mleap`)

kevinykuo (approved):

harupy:
@kevinykuo @yitao-li 

Looke like the R CI check failed on maser:

https://github.com/mlflow/mlflow/runs/1828958160#step:10:1159

Seems related to this issue:

https://github.com/mlflow/mlflow/pull/4051

kevinykuo:
@harupy yep do you mind reverting this so we don't break other PRs? Laptop out of commission right now... @yitao-li let's investigate in the AM.

harupy:
Thanks for the quick reply. #4060 seems to have fixed the issue. Could you review it?

PR #3815, opened 2020-12-11
Add autologging safety utils to several autologging integrations

dbczumar:
## What changes are proposed in this pull request?

This PR applies the utilities introduced by https://github.com/mlflow/mlflow/pull/3682 to the following autologging integrations:

- scikit-learn
- keras
- XGBoost

It also updates `try_mlflow_log` to respect the autologging test mode introduced via #3682. When test mode is enabled, failures that are usually swallowed by `try_mlflow_log` are raised in order to more effectively catch errors in MLflow's autologging integrations.

**TODO**: Add docs for `disable` parameter to each autologging method!

## How is this patch tested?

- Autologging safety integration tests (`tests/autologging/test_autologging_safety_integration.py`)
- Pre-existing autologging ML library integration tests, which are now executed in test mode by setting the `MLFLOW_AUTOLOGGING_TESTING` environment variable to `"true"`.

### Is this a user-facing change?

- [X] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [X] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
[...]

PR #3838, opened 2020-12-14
Add disable=True/False flag for Spark autologging

mohamad-arabi:

Add a disable boolean flag to Spark autologging, whether to log to mlflow from spark session automatically. This can be helpful if one wants to disable autologging after having enabling it for a spark session.

## How is this patch tested?

Added pytests to safe guard feature.

Components 

Interface 

Language 

Integrations


mlflow/_spark_autologging.py
```diff
@@ -217,6 +220,8 @@ def _notify(self, path, version, data_format):
         Method called by Scala SparkListener to propagate datasource read events to the current
         Python process
         """
+        if autologging_is_disabled("spark"):
```
mohamad-arabi, line 223: @dbczumar  I need access to the flavor name but the problem is if I define `FLAVOR_NAME` variable then https://github.com/mlflow/mlflow/blob/master/tests/test_flavors.py#L30 is picking up `mlflow/_spark_autologging.py` as a flovor as opposed to just `mlflow/spark.py` and fails the test. I am still trying to think of a better way to pass in flavor name to `_spark_autologging.py`

mlflow/_spark_autologging.py
```diff
@@ -248,6 +253,9 @@ def in_context(self):
         return True
 
     def tags(self):
+        # if autologging is disabled, then short circuit `tags()` and return empty dict.
+        if autologging_is_disabled("spark"):
```
mohamad-arabi, line 257: Same as comment above

mlflow/spark.py
```diff
@@ -682,6 +684,8 @@ def autolog():
         # next-created MLflow run if no run is currently active
         with mlflow.start_run() as active_run:
             pandas_df = loaded_df.toPandas()
+
+        :param disable: Whether to enable or disable autologging.
```
dbczumar, line 688: Can we match the param docstring of `mlflow.sklearn.autolog()` (and others) introduced by #3815 ?

mlflow/spark.py
```diff
@@ -627,7 +628,8 @@ def predict(self, pandas_df):
 
 
 @experimental
-def autolog():
+@autologging_integration(FLAVOR_NAME)
+def autolog(disable=False):  # pylint: disable=unused-argument
     """
     Enables automatic logging of Spark datasource paths, versions (if applicable), and formats
```
dbczumar, line 634: For consistency with docstring updates in #3815, can we update this to say:

```suggestion
    Enables (or disables) and configures logging of Spark datasource paths, versions (if applicable), and formats
```

tests/spark_autologging/test_spark_disable_autologging.py
```diff
@@ -0,0 +1,113 @@
...
+    with mlflow.start_run():
+        run_id = mlflow.active_run().info.run_id
+        df.filter("number1 > 0").collect()
+        time.sleep(1)
+    run = mlflow.get_run(run_id)
+    _assert_spark_data_logged(run=run, path=file_path, data_format=data_format)
+
+
+@pytest.mark.large
+def test_enabling_autologging_does_not_throw_when_spark_hasnt_been_started(spark_session):
```
dbczumar, line 111: @mohamad-arabi Can we address https://github.com/dbczumar/mlflow/pull/20#discussion_r541764153 as well?

dbczumar: Nevermind, I see that's done elsewhere :)

mlflow/_spark_autologging.py
dbczumar: @mohamad-arabi Nice catch! Can we update https://github.com/mlflow/mlflow/blob/master/tests/test_flavors.py#L30 to only pick up public modules (e.g., modules that aren't prefixed with `_`?). That should solve our problem of importing `FLAVOR_NAME` in `_spark_autologging`.

mlflow/_spark_autologging.py
dbczumar: Suggested a fix: https://github.com/mlflow/mlflow/pull/3838/files#r542753198

mlflow/_spark_autologging.py
mohamad-arabi: Not really @dbczumar , as you can see in line 31, it expects one of the files to be `__init__.py`.  will check with Haru, if we can filter out files starting with `_` and remove the if case for `__init__.py`. Not sure why it is there

https://github.com/mlflow/mlflow/blob/master/tests/test_flavors.py#L31

dbczumar: As discussed offline, we'll keep `__init__` as a special case and otherwise omit any private modules beginning with the `_` character.

dbczumar (approved):
LGTM! I tried this out manually & confirmed that the enable / disable flag is respected when loading data & then creating several runs, as well as when creating a run and loading data within it. Thanks @mohamad-arabi !

dbczumar:
Remaining test failures are unrelated to this PR. Merging!

PR #3859, opened 2020-12-17
Add autologging safety utils to statsmodels & Fast.ai

dbczumar:

This PR introduces autologging safety utilities to fastai & statsmodels. It also makes some fixes / simplifications / improvements to statsmodels:

- `mlflow.statsmodels.autolog` now imports the required `statsmodels.api` module, which is required for identifying the set of statsmodels classes to pass. This import failure would have been caught by `tests/tracking/fluent/test_fluent_autolog.py` if we propagated exceptions in test mode. Accordingly, this PR augments `tests/tracking/fluent/test_fluent_autolog.py` to observe test mode and propagate exceptions, ensuring that these issues are caught going forward.

- `statsmodels.autolog` now accepts the `log_models` argument

- statsmodels autologging now records parameters *before* model training (rather than after)

- complex alternative patching logic (`apply_gorilla_patch`) from https://github.com/mlflow/mlflow/blob/6dee5cb250460e8dc7accb487e54df8c95921e0e/mlflow/statsmodels.py#L349 has been removed, as the case where a patch is applied to a parent class followed by a child class is already handled by `mlflow.utils.gorilla`: https://github.com/mlflow/mlflow/blob/6dee5cb250460e8dc7accb487e54df8c95921e0e/mlflow/utils/gorilla.py#L322-L328

- Logic to prevent patches from being applied multiple times has been removed. Patching is idempotent, and we should allow users to retry patching if an unexpected failure occurs the first time. This also simplifies patch safety testing.

## How is this patch tested?

- Existing fastai & statsmodels flavor tests
- Addition of `fastai` & `statsmodels` to `tests/autologging/test_autologging_safety_integration.py`
- Expanded test coverage in `tests/tracking/fluent/test_fluent_autolog.py` (we now respect the test mode flag and raise exceptions when autologging enablement fails in test mode)

Note: The test failure in `test_universal_autolog_calls_specific_autologs_correctly` will be addressed once autologging safety utilities are added to `mlflow.tensorflow.autolog()` via https://github.com/mlflow/mlflow/pull/3861.

## Release Notes

Bug fixes: `statsmodels.autolog()` now imports the required `statsmodels.api` module and logs parameters prior to training, rather than after.

Components 

Interface 

Language 

Integrations


mlflow/statsmodels.py
```diff
@@ -346,39 +362,6 @@ def overrides(klass, function_name):
         except (IndexError, AttributeError):
             return False
 
-    def apply_gorilla_patch(patch, force_backup=True):
```
dbczumar, line 349: The logic where a patch is applied to a parent class followed by a child class is already handled by `mlflow.utils.gorilla`: https://github.com/mlflow/mlflow/blob/6dee5cb250460e8dc7accb487e54df8c95921e0e/mlflow/utils/gorilla.py#L322-L328

mlflow/statsmodels.py
```diff
@@ -473,61 +452,36 @@ def results_to_dict(results):
...
-        in corner cases where `self` still refers to the subclass but the method we want to invoke
-        (in the context of the subclass) belongs to a superclass
-
-        :param original_method: the original function object that will be replaced by this function
-        :return: the new fit function, from which we will be doing a call to the original fit
-                 method at some point
-        """
-
-        def fit(self, *args, **kwargs):
+    def wrapper_fit(original, self, *args, **kwargs):
```
dbczumar, line 455: I don't think there's a need for the patch implementation to call the `original_method` of a superclass from a subclass. As discussed in https://github.com/mlflow/mlflow/pull/3859/files#r544936250, `safe_patch` will handle parent/child patch ordering by way of `mlflow.utils.gorilla` and will deliver the correct `original` function as the first parameter to the patch implementation (`original`).

@harupy Please let me know if I'm missing anything here.

mlflow/tracking/fluent.py
```diff
@@ -1308,20 +1309,27 @@ def print_auto_logged_info(r):
...
         try:
-            needed_params = list(inspect.signature(autolog_fn).parameters.keys())
-            filtered = {k: v for k, v in locals_copy if k in needed_params}
-        except ValueError:
-            filtered = {}
+            autolog_fn = LIBRARY_TO_AUTOLOG_FN[module.__name__]
+            try:
+                needed_params = list(inspect.signature(autolog_fn).parameters.keys())
+                filtered = {k: v for k, v in locals_copy if k in needed_params}
+            except ValueError:
```
dbczumar, line 1317: Restructuring this for safety. Previously, if the dictionary lookup failed or the `try` block threw a non-ValueError exception, `mlflow.autolog()` would break.

mlflow/tracking/fluent.py
```diff
@@ -1308,20 +1309,27 @@ def print_auto_logged_info(r):
...
+                filtered = {}
 
-        try:
             autolog_fn(**filtered)
             _logger.info("Autologging successfully enabled for %s.", module.__name__)
         except Exception as e:  # pylint: disable=broad-except
-            _logger.warning(
-                "Exception raised while enabling autologging for %s: %s", module.__name__, str(e)
-            )
+            if _is_testing():
```
dbczumar, line 1323: Allows `mlflow.autolog()` to respect test mode, enabling us to catch errors raised by `mlflow.autolog()`.

tests/autologging/fixtures.py
```diff
@@ -1,20 +1,33 @@
...
+from mlflow.utils.autologging_utils import _is_testing, _AUTOLOGGING_TEST_MODE_ENV_VAR
 
 
 @pytest.fixture
 def test_mode_off():
-    with mock.patch("mlflow.utils.autologging_utils._is_testing") as testing_mock:
-        testing_mock.return_value = False
-        assert not autologging_utils._is_testing()
+    try:
+        prev_env_var_value = os.environ.pop(_AUTOLOGGING_TEST_MODE_ENV_VAR, None)
```
dbczumar, line 11: I've refactored these fixtures to move away from mocking `autologging_utils._is_testing`, which is only effective if a module calls `autologging_utils._is_testing` without re-importing the `_is_testing` function. If we rely purely on environment variables (which are unit tested in the autologging safety suite), mocking challenges are avoided.

tests/autologging/test_autologging_safety_integration.py
```diff
@@ -38,31 +40,29 @@ def disable_autologging_at_test_end():
         integration.autolog(disable=True)
 
 
-def test_autologging_integrations_expose_configs_and_support_disablement():
-    for integration in AUTOLOGGING_INTEGRATIONS_TO_TEST:
-        integration.autolog(disable=False)
+@pytest.mark.parametrize("integration", AUTOLOGGING_INTEGRATIONS_TO_TEST.keys())
```
dbczumar, line 43: Refactoring for debugability (one test case per integration)

mlflow/statsmodels.py
```diff
@@ -473,61 +452,36 @@ def results_to_dict(results):
 
         return results_dict
 
-    def wrapper_fit(original_method):
```
dbczumar, line 476: I don't think there's a need for the patch implementation to call the `original_method` of a superclass from a subclass. As discussed in https://github.com/mlflow/mlflow/pull/3859/files#r544936250, `safe_patch` will handle parent/child patch ordering by way of `mlflow.utils.gorilla` and will deliver the correct `original` function as the first parameter to the patch implementation (`original`).

@harupy Please let me know if I'm missing anything here.

mlflow/statsmodels.py
```diff
@@ -473,61 +452,36 @@ def results_to_dict(results):
...
-                    try_mlflow_log(mlflow.end_run)
+            return model
 
-        return fit
+        finally:
+            # Clean the shared flag for future calls in case it had been set here ...
+            if should_autolog:
+                AutologHelpers.should_autolog = True
 
-    if not AutologHelpers.patching_accomplished:
```
dbczumar, line 531: As discussed in the PR description, patching is idempotent. Further, if an user-addressable patching issue occurs during the first call to `mlflow.autolog()`, users should be able to try patching again.

FWIW, because `statsmodels.api` was not imported by default previously, calling `mlflow.statsmodels.autolog()` without importing this module would fail at `patch_class_tree(...)`, but `patching_accomplished` would already be `True`. This prevented users from fixing the import error and trying again.

mlflow/gluon.py
```diff
@@ -321,8 +321,8 @@ def autolog(log_models=True, disable=False):  # pylint: disable=unused-argument
 
     :param log_models: If ``True``, trained models are logged as MLflow model artifacts.
                        If ``False``, trained models are not logged.
-    :param disable: If ``True``, disables all supported autologging integrations. If ``False``,
-                    enables all supported autologging integrations.
+    :param disable: If ``True``, disables the MXNet Gluon autologging integration. If ``False``,
```
dbczumar, line 324: Using this PR as an opportunity to fix a slew of doc errors introduced by me. Namely, the `disable` parameter for framework-specific autologging integrations was incorrectly documented to state that it would disable *all* autologging integrations.

mlflow/statsmodels.py
harupy: We fixed this child-parent issue in https://github.com/mlflow/mlflow/pull/3582, right?

mlflow/tracking/fluent.py
harupy: Should we use `KeyError` here instead of `ValueError`?

```
>>> {}["foo"]
KeyError: 'foo'
```

harupy: wait

harupy: which line causes `ValueError`?

mlflow/statsmodels.py
dbczumar: Yep! We did!

mlflow/tracking/fluent.py
dbczumar: `inspect.signature` raises the `ValueError` (https://docs.python.org/3/library/inspect.html#inspect.signature). Though, according to the docs, it may also throw `TypeError`. To avoid headaches with exception typing later on, I've gone ahead and just changed this to a broad except.

harupy (approved):
LGTM!

Issue #3733, opened 2020-11-24
[FR] Adding NextPageToken support to ListExperiments API


wamartin-aml:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

(In a few sentences, provide a clear, high-level description of the feature request)
It would be helpful to support pagination for experiment listing. A list request would come back with a subset of results and a continuation token. That token could then be passed in with a next request to get the subsequent subset of results and a second token, etc.

## Motivation
- What is the use case for this feature? For workspaces with many experiments, it can be too much to get a list of all experiments at once, but a user may still want a way to go through all experiments. Pagination is the standard way to make that possible.
- Why is this use case valuable to support for MLflow users in general? Introducing pagination gives users more levers directly, but also enables downstream scenarios like UIs built on MLflow APIs to improve their performance.
- Why is this use case valuable to support for your project(s) or organization? Some of our key scenarios are failing because pagination is currently unsupported for ListExperiments from the MLflow side.
- Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient) If we need to search through every experiment in a workspace right now, there's no way to do that other than to get the full list of experiments all at once.

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
[...]

PR #3881, opened 2020-12-19
list experiments page token

wamartin-aml:

This PR adds page_token support to the ListExperiments API. If the user passes in the `max_results` parameter, they will receive a `PagedList` of up to `max_results` experiments, as well as a continuation token if there are any more experiments left. If the user does not pass the `max_results` parameter, the API will work exactly as it does right now, and return a list of every experiment.

Addresses https://github.com/mlflow/mlflow/issues/3733

## How is this patch tested?

Increased test coverage in test_sqlalchemy_store.py. Still lacking coverage in test_rest_store.py. I wanted to get this pull request out to start people looking at it.

Added support for pagination on the ListExperiments API. If `max_results` is passed in, the call will be paginated, while without it, all experiments will be returned.

Components 

Interface 

Language 

Integrations


wamartin-aml:
@smurching, a couple things that I know are missing here at the moment:
* generated proto files - I'll need some help with getting that going
* testing for the rest store

Beyond that, grateful for any feedback and discussion about whether this even the direction MLflow wants to move in.

mlflow/protos/service.proto
```diff
@@ -682,9 +682,18 @@ message ListExperiments {
   // If unspecified, return only active experiments.
   optional ViewType view_type = 1;
 
+  // Maximum number of registered models desired. Max threshold is 50K.
+  optional int64 max_results = 2 [default = 50000];
```
akshaya-a, line 686: how did we arrive at 50k? also nit copy paste in the comment

is this max_results (eg top which only makes sense combined with order_by) or page_size?

akshaya-a:
Have we also considered an alternative approach of adding a new SearchExperiments API (similar to SearchRuns, SearchModels) that can start paginated?

wamartin-aml:
@smurching and I talked about that in the issue I raised. The people I chatted with from Azure ML preferred smaller surface area, but they said it was pretty much a toss-up.

mlflow/protos/service.proto
wamartin-aml: 50k was the analogous limit for registered models but there's absolutely no reason it needs to stay there.

`max_results` is page size - `max_results` is used the same way elsewhere in the in user-facing APIs so I'm sure how much of a choice we have

mlflow/server/handlers.py
```diff
@@ -504,9 +504,13 @@ def _get_metric_history():
 @catch_mlflow_exception
 def _list_experiments():
     request_message = _get_request_message(ListExperiments())
-    experiment_entities = _get_tracking_store().list_experiments(request_message.view_type)
```
eedeleon, line 507: for abstract methods being changed, it is better to pass kwargs.

This will break old stores otherwise since they will likely not pass args through but may pass kwargs through since it is pythonic

mlflow/store/tracking/rest_store.py
```diff
@@ -54,16 +55,28 @@ def _call_endpoint(self, api, json_body):
...
-            for experiment_proto in response_proto.experiments
-        ]
+        if max_results is not None:
+            return PagedList([
+                Experiment.from_proto(experiment_proto)
+                for experiment_proto in response_proto.experiments
+            ], response_proto.next_page_token)
+        else:
+            return [
+                Experiment.from_proto(experiment_proto)
```
eedeleon, line 77: IT might be simpler to flatten the conditional

protos = Experiment.from_proto(experiment_proto)
                for experiment_proto in response_proto.experiments
            ]
return PagedList(protos, , response_proto.next_page_token) if max_results is not None else protos

mlflow/tracking/client.py
```diff
@@ -352,7 +357,7 @@ def print_experiment_info(experiments):
             - experiment_id: 1, name: Experiment 1, lifecycle_stage: active
             - experiment_id: 2, name: Experiment 2, lifecycle_stage: deleted
         """
-        return self._tracking_client.list_experiments(view_type)
+        return self._tracking_client.list_experiments(view_type, max_results, page_token)
```
eedeleon, line 360: same kwarg comment

mlflow/protos/service.proto
smurching: I guess a related design decision is whether we paginate within the `MlflowClient.list_experiments` API, or have it be a thin wrapper over the REST API that returns a PagedList.

I guess for parity with existing APIs we should do the latter (which is what this PR currently does), but seems like if we implemented pagination within `MlflowClient.list_experiments` itself, we could have a separate default for in the REST API and in the client, e.g. have `MlflowClient.list_experiments(max_results=10000)`  but impose a stricter default at the REST API level

mlflow/tracking/_tracking_service/client.py
```diff
@@ -106,12 +106,21 @@ def list_run_infos(
             experiment_id, run_view_type, max_results, order_by, page_token
         )
 
-    def list_experiments(self, view_type=None):
+    def list_experiments(
+        self, view_type=None, max_results=SEARCH_MAX_RESULTS_DEFAULT, page_token=None
```
smurching, line 110: BTW note that the value of `SEARCH_MAX_RESULTS_DEFAULT` here is 1000 - I'd like to do a bit of data science to understand existing client usage before settling on this value, just to make sure I understand what impact this might have on existing users (of course from a server-implementor perspective a lower default is better from a load perspective)

smurching (commented):
Thanks @wamartin-aml! Leaving a few comments

mlflow/store/tracking/file_store.py
```diff
@@ -221,7 +221,7 @@ def _get_active_experiments(self, full_path=False):
     def _get_deleted_experiments(self, full_path=False):
         return list_subdirs(self.trash_folder, full_path)
 
-    def list_experiments(self, view_type=ViewType.ACTIVE_ONLY):
+    def list_experiments(self, view_type=ViewType.ACTIVE_ONLY, max_results=None, page_token=None):
```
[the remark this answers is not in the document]
wamartin-aml: Interestingly, in the three or four other pagination PRs that have gone in, file_store has remained untouched - that's why I hadn't added anything along those lines (a. because I wasn't sure that we wanted it since we didn't for anything else, b. because I didn't have a template to work from making me less confident). I've added pagination here using `_search_runs` as a model, but it sounds like file_store in general needs some discussion.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -251,12 +252,79 @@ def _list_experiments(
...
+            else:
+                queried_experiments = (
+                    session.query(SqlExperiment).options(*query_options).filter(*conditions).all()
+                )
+
+            experiments = [exp.to_mlflow_entity() for exp in queried_experiments]
+        if max_results is not None:
+            return PagedList(experiments[:max_results], compute_next_token(len(experiments)))
+        else:
+            return experiments
```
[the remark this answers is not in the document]
wamartin-aml: Easy enough!

mlflow/tracking/_tracking_service/client.py
wamartin-aml: I'm certainly not set on anything in terms of these limits, I just defaulted to whatever I could find elsewhere. Whenever you have time to poke around and settle on something, let me know!

mlflow/protos/service.proto
wamartin-aml: I've removed the 50k limit from protos. When we have consensus on good limits, I'll make sure they're implemented through the other pieces of the PR.

mlflow/server/handlers.py
wamartin-aml: For this PR, I believe I've made the necessary changes

mlflow/protos/service.proto
wamartin-aml: @smurching it looks like removing the default totally leads to problems in the Java client, at least - without something supplied in the protos, it defaults to 0 and pulls no experiments at all. I think having some absurdly high limit like 50k may be necessary, where a more thoughtful limit is then applied in other layers.

wamartin-aml: yeah, just confirmed that if I add the default back the Java and R test suites are passing again.

mlflow/store/tracking/file_store.py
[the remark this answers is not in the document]
smurching: Ah @wamartin-aml yeah sorry for the confusion - the other APIs we've most-recently added pagination to are model registry APIs, which don't have file_store support (i.e. in order to use the model registry you need to use a SQL-backed store). In this case, we do need to add it for FileStore since ``list_experiments`` is supported in the FileStore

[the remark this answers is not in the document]
smurching: Thanks for doing so!

mlflow/protos/service.proto
smurching: @wamartin-aml could we set the default here to 10,000? Just because it feels a little strange to be able to specify a higher `max_results` by passing nothing at all (50,000) than if you explicitly pass `max_results` (we limit to 10,000)

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -251,12 +252,79 @@ def _list_experiments(
...
+        """
+        :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                            passed, all experiments will be returned.
+        :param page_token: Token specifying the next page of results. It should be obtained from
+                            a ``list_experiments`` call.
+        :param eager: If ``True``, eagerly loads each experiments's tags. If ``False``, these tags
+                      are not eagerly loaded and will be loaded if/when their corresponding
+                      object properties are accessed from a resulting ``SqlExperiment`` object.
+        """
+        stages = LifecycleStage.view_type_to_stages(view_type)
```
smurching, line 273: Could we add validation to make sure that `max_results` is below our limit (e.g. 10,000)? For reference see [the helper method for validating params](https://github.com/mlflow/mlflow/blob/ed2b8d7359e053c1b3ded54c3a2c0c064ccf347b/mlflow/utils/validation.py#L151) and how we use it (we basically apply it in each of our built-in store implementations)

smurching: And you can add tests in [this file](https://github.com/mlflow/mlflow/blob/ed2b8d7359e053c1b3ded54c3a2c0c064ccf347b/tests/utils/test_validation.py#L65-L71)

smurching (commented):
@wamartin-aml this basically LGTM, thanks for all the updates! I just had a few comments about validating the protobuf limit on `max_results` in our built-in stores. I'm otherwise happy to merge this & update the max results limit as needed in my own follow-up PR (so that I'm not blocking you), thanks again for the PR!

smurching (approved):
LGTM thanks @wamartin-aml!

harupy:
The `Cross version tests / test (pytorch-lightning / dev / autologging) (pull_request)` failure is unrelated to this PR.

Here's a fix for this error: https://github.com/mlflow/mlflow/pull/4302/

tests/tracking/test_tracking.py
```diff
@@ -134,6 +134,67 @@ def _assert_exps(ids_to_lifecycle_stage, view_type_arg):
...
+    while result.token:
+        result = client.list_experiments(max_results=max_results, page_token=result.token)
+        assert len(result) <= max_results
+        returned_experiments.extend(result)
+    assert result.token is None
+    returned_exp_id_set = set([exp.experiment_id for exp in returned_experiments])
+    assert set(experiments) - returned_exp_id_set == set()
+
+
+def test_list_experiments_paginated_returns_in_correct_order():
```
harupy, line 156: @smurching @wamartin-aml 

I found that this test fails with `FileStore`.

Code to reproduce:

```suggestion
@pytest.mark.notrackingurimock
def test_list_experiments_paginated_returns_in_correct_order(tmpdir):
    mlflow.set_tracking_uri(tmpdir.strpath)
```

Is this what we intend?

mlflow/protos/service.proto
```diff
@@ -682,9 +682,21 @@ message ListExperiments {
   // If unspecified, return only active experiments.
   optional ViewType view_type = 1;
 
+  // Maximum number of experiments desired. If specified, max_results can be at most 10,000. 
+  // Servers may impose a default max_results of their choice if unspecified by the client. 
+  // Callers of this endpoint are encouraged to pass max_results explicitly and leverage page_token 
+  // to iterate through experiments.
+  optional int64 max_results = 2 [default = 10000];
```
dbczumar, line 689: Hi @wamartin-aml. After some offline design discussions, the largest `max_results` value that can be accommodated by Databricks is currently 1,000 (rather than 10,000). AML is free to set a higher threshold. If possible, can we update the default value and documentation here to reflect 1,000 as the default and the maximum threshold that all backends must support?

cc @harupy

mlflow/protos/service.proto
```diff
@@ -682,9 +682,21 @@ message ListExperiments {
...
+
+  // Pagination token to go to the next page based on a previous query.
+  optional string page_token = 3;
+
   message Response {
     // All experiments.
     repeated Experiment experiments = 1;
+
+    // Pagination token to request next page of models for the same query.
```
dbczumar, line 698:
```suggestion
    // Pagination token to request next page of experiments for the same query.
```

harupy:
@wamartin-aml @smurching @dbczumar I've pushed some commits to address Corey's comments and fix tests.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -251,12 +253,72 @@ def _list_experiments(
...
+            query_options = self._get_eager_experiment_query_options() if eager else []
+            if max_results is not None:
+                offset = SearchUtils.parse_start_offset_from_page_token(page_token)
+                queried_experiments = (
+                    session.query(SqlExperiment)
+                    .options(*query_options)
+                    .filter(*conditions)
+                    .offset(offset)
+                    .limit(max_results_for_query)
+                    .all()
```
harupy, line 304: Do we need to sort experiments (by experiment name and id) here (note there is creation_time column in open-source MLflow)?

wamartin-aml: Thanks for making all of these changes @harupy. I don't see any guarantee currently about experiments being sorted, so I don't think that's a requirement for this feature. It might be a useful follow-up item though.

harupy: Agree that this is a useful follow-up item. May we could sort experiments by `experiment_id` in decsending order so newer experiments are fetched early.

harupy:
@wamartin-aml We're planning to provide `mlflow.list_experiments` (which handles the pagination token and return all or a specified number of experiments) as a workaround for the `list_experiments` behavior change. What do you think?

```python
# Assumption: we're using the REST store

experiments = mflow.tracking.MlflowClient().list_experiments()
# `experiments` may not contain all experiments.

# Workaround
all_experiments = mlflow.list_experiments()  # handles the page token and fetches all experiments
all_experiments = mlflow.list_experiments(max_results=5000)  # you can specify `max_results`
```

wamartin-aml:
@harupy I'm set with everything mentioned here - I think we'll be good to go once you've done the DCO signoff for your commits!

mlflow/protos/service.proto
```diff
@@ -682,9 +682,21 @@ message ListExperiments {
   // If unspecified, return only active experiments.
   optional ViewType view_type = 1;
 
+  // Maximum number of experiments desired. If specified, max_results can be at most 1000.
+  // Servers may impose a default max_results of their choice if unspecified by the client.
```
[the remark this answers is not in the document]
harupy: @jinzhang21 I've updated the comment on `max_results`. Could you take another pass?

jinzhang21 (approved):
LGTM on the API change.

mlflow/store/tracking/file_store.py
```diff
@@ -221,13 +222,18 @@ def _get_active_experiments(self, full_path=False):
...
```
jinzhang21, line 225: Per Slack discussion, if we remove max_results default value in the proto field, the API should always provide a default value instead of None. For example:

def list_experiments(self, view_type=ViewType.ACTIVE_ONLY, max_results=1000, page_token=None):
  ...

ditto for the other API list_experiments(...) below.

harupy:
@smurching @jinzhang21

The following tables summarizes what `client.list_experiments()` returns for each backend in the current implementation:

| Backend type      | What does `client.list_experiments()` return (note `max_results` is unspecified)? |
| :---------------- | :-------------------------------------------------------------------------------- |
| SQL               | Return all experiments                                                            |
| File              | Return all experiments                                                            |
| REST              | Backend can determine what to return                                              |
| Databricks REST\* | Backend can determine what to return                                              |

mlflow/store/tracking/file_store.py
harupy: Added the defaulut values!

harupy:
@smurching @jinzhang21

The table below describes what `client.list_experiments` returns for each backend in **the current implementation**:

| Backend type      | When `max_results` is unspecified    | When `max_results` is specified      |
| :---------------- | :----------------------------------- | :----------------------------------- |
| SQL               | All experiments                      | Paginated results                    |
| File              | All experiments                      | Paginated results                    |
| REST              | Backend determines what to return | Backend determines what to return |
| Databricks REST\* | Backend determines what to return | Backend determines what to return |

```python
import mlflow

client = mlflow.tracking.MlflowClient(tracking_uri="...")
client.list_experiments()  # `max_results` is unspecified
client.list_experiments(max_results=100)  # `max_results` is specified
```

Is this correct or not?

mlflow/store/tracking/rest_store.py
```diff
@@ -291,6 +311,9 @@ def record_logged_model(self, run_id, mlflow_model):
         self._call_endpoint(LogModel, req_body)
 
 
+_LIST_EXPERIMENTS_MAX_RESULTS_DEFAULT_IN_DATABRICKS = 1000
```
jinzhang21, line 314: Unsure about the convention to put DB specific constant. Perhaps move it to the top of the file for better readability?

Also what's the reason to choose a separate parameter instead of using SEARCH_MAX_RESULTS_DEFAULT?

mlflow/store/tracking/rest_store.py
```diff
@@ -312,8 +335,47 @@ def get_experiment_by_name(self, experiment_name):
...
+        #
+        # improt mlflow
+        # client = mlflow.tracking.MlflowClient(tracking_uri="https://...")
+        # client.list_experiments()
+        # ^ This calls `client._tracking_client.store.list_experiments(..., max_results=None)`
+        #
+        # To guarantee that the `max_results` field is always present in the request sent by
+        # MLflow client >= 1.18.0, set the default value if `max_results` is None.
+        # This allows MLflow backend in Databricks to infer the client version by examining
+        # the presence of the `max_results` field and determine what to return.
```
jinzhang21, line 372: Setting max_results in the default value to some number already effectively enforce a limit. What's the downside to allow user to explicitly specify None here? I assume the user would want an infinite limit effectively. If the backend just respect that and return min(Infinite, BackendInternalLimit), it should still work, no?

mlflow/utils/validation.py
```diff
@@ -33,6 +33,7 @@
 MAX_ENTITY_KEY_LENGTH = 250
 MAX_MODEL_REGISTRY_TAG_KEY_LENGTH = 250
 MAX_MODEL_REGISTRY_TAG_VALUE_LENGTH = 5000
+MAX_EXPERIMENTS_LISTED_PER_PAGE = 1000
```
jinzhang21, line 36: This value probably should be larger than 1k. The 50k 'threshold' for search seems make sense.

mlflow/utils/validation.py
```diff
@@ -148,6 +149,30 @@ def _validate_model_version_tag(key, value):
     _validate_length_limit("Model version value", MAX_MODEL_REGISTRY_TAG_VALUE_LENGTH, value)
 
 
+def _validate_experiment_pagination(max_results):
```
jinzhang21, line 152: change the fn name to reflect what it's actually checking:
_validate_experiment_pagination() -> _validate_max_results()
Or if you want to validate more stuff later, _validate_list_experiments_input() is a good choice.

tests/utils/test_validation.py
```diff
@@ -195,6 +196,15 @@ def test_validate_experiment_name():
             _validate_experiment_name(invalid_name)
 
 
+def test_validate_experiment_pagination():
```
jinzhang21, line 199: Prefer to not test internal functions, but focus on testing public APIs. For example, in this case, prefer to actually call mlflow.list_experiments() with out of range max_results as an argument and assert failure from there. I see it's already tested to some extent in test_list_experiments_paginated_errors().

jinzhang21 (commented):
Some minor comments. It's almost there. Looks great overall. Thanks, Haru!

mlflow/utils/validation.py
harupy: Makes sense. I'll fix it.

harupy: Makes sense. Renamed this function.

mlflow/store/tracking/rest_store.py
harupy: If we allow `None` here, that means the new client also sends a request to the MLflow service backend in Databricks and can't distinguish the new and old clients.

tests/utils/test_validation.py
harupy: Makes sense. I'll replace `_validate_experiment_pagination` with `mlflow.list_experiments()`.

mlflow/store/tracking/rest_store.py
```diff
@@ -54,16 +56,34 @@ def _call_endpoint(self, api, json_body):
         response_proto = api.Response()
         return call_endpoint(self.get_host_creds(), endpoint, method, json_body, response_proto)
 
-    def list_experiments(self, view_type=ViewType.ACTIVE_ONLY):
+    def list_experiments(
+        self,
+        view_type=ViewType.ACTIVE_ONLY,
+        max_results=SEARCH_MAX_RESULTS_DEFAULT,
+        page_token=None,
+    ):
```
WeichenXu123, line 64: Shall we also create feature flag for this ? (if feature flag off, run as old non-pagination behavior) ?

harupy: If a user passes `None`, he can get all experiments.

mlflow/store/tracking/rest_store.py
harupy: > Also what's the reason to choose a separate parameter instead of using SEACH_MAX_RESULTS_DEFAULT?

To avoid the case where someones updates `SEARCH_MAX_RESULTS_DEFAULT` without knowing it's also used in `DatabricksRestStore`

mlflow/store/tracking/file_store.py
```diff
@@ -24,7 +24,11 @@
 from mlflow.exceptions import MlflowException, MissingConfigException
 import mlflow.protos.databricks_pb2 as databricks_pb2
 from mlflow.protos.databricks_pb2 import INTERNAL_ERROR, RESOURCE_DOES_NOT_EXIST
-from mlflow.store.tracking import DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH, SEARCH_MAX_RESULTS_THRESHOLD
+from mlflow.store.tracking import (
+    DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH,
+    SEARCH_MAX_RESULTS_DEFAULT,
+    SEARCH_MAX_RESULTS_THRESHOLD,
```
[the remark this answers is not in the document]
harupy: It's acutally used in `_search_runs` :)

mlflow/store/tracking/abstract_store.py
```diff
@@ -22,12 +22,18 @@ def __init__(self):
         pass
 
     @abstractmethod
-    def list_experiments(self, view_type=ViewType.ACTIVE_ONLY):
+    def list_experiments(self, view_type=ViewType.ACTIVE_ONLY, max_results=None, page_token=None):
         """
 
         :param view_type: Qualify requested type of experiments.
+        :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                            passed, all experiments will be returned.
```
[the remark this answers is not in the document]
harupy: Applied your suggestion!

mlflow/store/tracking/rest_store.py
harupy: How do we controll the feature flag?

mlflow/store/tracking/rest_store.py
jinzhang21: I think the point is to have a consistent behavior when observing certain input, irrespective of where it's coming from. The contract established here could be,
1) max_results cannot be 0 or negative. An error is returned;
2) max_results can be None / unset. The server will attempt to return all experiments. However, it could apply its internal limit if there are too many to return.
3) max_results is positive. In this case, the server will respect the input value. However, it could apply its internal limit if the requested value is too large.

The contract does not really care about where the request is sent from. Specifically on 2), it could be from an old client, which doesn't know max_results (hence unset), or from a new client, which knows how to use max_results but deliberately unsets the field to request maximum amount allowed by the server. Either way, the server only needs to check the field, and proceeds with 2) if max_results field does not exist, as listed in the contract.

Does it make sense?

mlflow/store/tracking/rest_store.py
WeichenXu123: withdraw this

mlflow/store/tracking/rest_store.py
```diff
@@ -54,16 +55,28 @@ def _call_endpoint(self, api, json_body):
...
+        self, view_type=ViewType.ACTIVE_ONLY, max_results=None, page_token=None,
+    ):
         """
-        :return: a list of all known Experiment objects
+        :param view_type: Qualify requested type of experiments.
+        :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                            passed, the server will pick a maximum number of results to return.
+        :param page_token: Token specifying the next page of results. It should be obtained from
+                            a ``list_experiments`` call.
+        :return: a PagedList of all known Experiment objects
```
dbczumar, line 67: Awesome! @harupy can we also updated the return type docstring for `search_runs` to reflect that a PagedList is returned? The current return type is technically inaccurate: https://github.com/mlflow/mlflow/blob/1c43176cefb5531fbb243975b9c8c5bfb9775e66/mlflow/tracking/_tracking_service/client.py#L378-L380

mlflow/tracking/_tracking_service/client.py
```diff
@@ -109,12 +109,20 @@ def list_run_infos(
...
+    def list_experiments(self, view_type=None, max_results=None, page_token=None):
         """
-        :return: List of :py:class:`mlflow.entities.Experiment`
+        :param max_results: If passed, specifies the maximum number of experiments desired.
+                            If not passed, all experiments will be returned for the File and
+                            SQLAlchemy backends. For the REST backend, the server will determine
+                            an appropriate number of experiments to return.
+        :param page_token: Token specifying the next page of results. It should be obtained from
+                            a ``list_experiments`` call.
+        :return: PagedList of :py:class:`mlflow.entities.Experiment`
```
dbczumar, line 120: Can we update the return type docstring for `search_runs()` as well? https://github.com/mlflow/mlflow/pull/3881/files#r651172964

mlflow/protos/service.proto
```diff
@@ -682,9 +682,21 @@ message ListExperiments {
   // If unspecified, return only active experiments.
   optional ViewType view_type = 1;
 
+  // Maximum number of experiments desired.
+  // Servers may select a desired default value and threshold for `max_results`.
+  // Callers of this endpoint are encouraged to pass max_results explicitly and
+  // leverage page_token to iterate through experiments.
```
dbczumar, line 688: @harupy @wamartin-aml Can we indicate that the max threshold is guaranteed to be at least 1000 for all servers?

dbczumar, line 688: Can we update the `SearchRuns` documentation for the `max_results` field to read similarly? Currently, it just says `Maximum number of runs desired. Max threshold is 50000`, which may be inaccurate for certain backends (which may support even more results per page).

mlflow/utils/validation.py
dbczumar: I think we should remove this client-side validation for the page length. It's up to the backend whether or not a particular `max_results` value is supported. I don't see a good reason to perform that validation here (we don't perform client-side validation for search).

I've also proposed that we update the proto docstring for the `max_results` field in SearchRuns to indicate that all servers are guaranteed to support a `max_results` threshold of at least 50,000 but may support more (see https://github.com/mlflow/mlflow/pull/3881/files#r651181532).

dbczumar (changes requested):
@harupy @wamartin-aml Generally looks great! I left a few small comments, with https://github.com/mlflow/mlflow/pull/3881/files#r651182740 (a +1 to @jinzhang21 's comment about the ListExperiments page size for SQL / File Stores) being the most significant.

Edit: I also left some comments on the fluent API - https://github.com/mlflow/mlflow/pull/3881/files#r651188477 is the most significant.

mlflow/tracking/fluent.py
```diff
@@ -815,6 +816,26 @@ def get_experiment_by_name(name: str) -> Optional[Experiment]:
     return MlflowClient().get_experiment_by_name(name)
 
 
+def list_experiments(
+    view_type: Optional[str] = None, max_results: Optional[int] = None,
+) -> PagedList[Experiment]:
```
dbczumar, line 821: @harupy @wamartin-aml Can we change the return type to `List[Experiment]`? Callers of the fluent API don't need to reason about pagination. Additionally, it seems that `_paginate` returns a vanilla list type rather than a `PagedList` (see https://github.com/mlflow/mlflow/blob/1c43176cefb5531fbb243975b9c8c5bfb9775e66/mlflow/tracking/fluent.py#L1213), so this return type hint seems incorrect.

mlflow/tracking/fluent.py
```diff
@@ -815,6 +816,26 @@ def get_experiment_by_name(name: str) -> Optional[Experiment]:
     return MlflowClient().get_experiment_by_name(name)
 
 
+def list_experiments(
+    view_type: Optional[str] = None, max_results: Optional[int] = None,
```
dbczumar, line 820: Can we explicitly set `view_type` to `ACTIVE_ONLY` by default? `None` is harder for users to reason about, and most users probably only care about active experiments.

mlflow/tracking/fluent.py
harupy: Thanks for the catch!

mlflow/tracking/fluent.py
harupy: Got it, I'll set `view_type` to `ACTIVE_ONLY`.

mlflow/tracking/_tracking_service/client.py
harupy: @dbczumar Created a follow-up PR [here](https://github.com/mlflow/mlflow/pull/4455)

mlflow/protos/service.proto
harupy: Got it, I'll address this in https://github.com/mlflow/mlflow/pull/4455

harupy: Updated the comment to indicate that!

mlflow/tracking/fluent.py
```diff
@@ -815,6 +815,26 @@ def get_experiment_by_name(name: str) -> Optional[Experiment]:
...
 
+def list_experiments(
+    view_type: int = ViewType.ACTIVE_ONLY, max_results: Optional[int] = None,
+) -> List[Experiment]:
+    """
+        :param view_type: Qualify requested type of experiments.
+        :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                            passed, all experiments will be returned.
+        :param page_token: Token specifying the next page of results. It should be obtained from
+                            a ``list_experiments`` call.
```
dbczumar, line 826: This isn't part of the API definition. Can we remove it?

dbczumar (commented):
@harupy Looks great! Happy to LGTM once https://github.com/mlflow/mlflow/pull/3881/files#r652195522 is addressed!

mlflow/tracking/fluent.py
harupy: Thanks for the catch! Will remove it.

dbczumar (approved):
LGTM! Thanks @wamartin-aml @harupy !

PR #3883, opened 2020-12-21
Autologging: Add context to MLflow warnings raised by patch implementations

dbczumar:

MLflow routines called by autologging patch code may issue warnings via the `warnings.warn` API. In many cases, the user cannot remediate the cause of these warnings because they result from the autologging patch implementation, rather than a user-facing API call.

Accordingly, this PR introduces a context manager that is designed to augment MLflow warnings issued during
autologging patch code execution, explaining the warnings were raised as a result of MLflow's autologging implementation. MLflow warnings are also redirected from `sys.stderr` to an MLflow logger with level WARNING. Warnings issued by code outside of MLflow are not modified. When the context manager exits, the original output behavior for MLflow warnings
is restored.

On Databricks, this helps reroute potentially confusing warning messages that may arise during autologging; e.g., warnings about integer columns during schema inference.

**Example - run the following workflow:**

```
import mlflow
mlflow.autolog()

from sklearn import datasets
iris = datasets.load_iris()

from scipy.stats import uniform

logistic = LogisticRegression(solver="saga", tol=1e-2, max_iter=200, random_state=0)
train_out = logistic.fit(iris.data, iris.target)
```

**Before**:
```
2020/12/21 01:22:43 INFO mlflow.tracking.fluent: Autologging successfully enabled for sklearn.
/Users/czumar/mlflow/mlflow/models/signature.py:124: UserWarning: Hint: Inferred schema contains integer column(s).
 Integer columns in Python cannot represent missing values. If your input data contains missing values at inference time, it
 will be encoded as floats and will cause a schema enforcement error. The best way to avoid this problem is to infer the
 model schema based on a realistic data sample (training dataset) that includes missing values. Alternatively, you can
 declare integer columns as doubles (float64) whenever these columns may have missing values. See `Handling Integers
 With Missing Values <https://www.mlflow.org/docs/latest/models.html#handling-integers-with-missing-values>`_ for more
 details.
 outputs = _infer_schema(model_output) if model_output is not None else None
```

**After**:
```
2020/12/21 01:23:12 INFO mlflow.tracking.fluent: Autologging successfully enabled for sklearn.
2020/12/21 01:23:13 WARNING mlflow.utils.autologging_utils: MLflow issued a warning during sklearn autologging:
 "/Users/czumar/mlflow/mlflow/models/signature.py:124: UserWarning: Hint: Inferred schema contains integer column(s).
 Integer columns in Python cannot represent missing values. If your input data contains missing values at inference time, it
 will be encoded as floats and will cause a schema enforcement error. The best way to avoid this problem is to infer the
 model schema based on a realistic data sample (training dataset) that includes missing values. Alternatively, you can
 declare integer columns as doubles (float64) whenever these columns may have missing values. See `Handling Integers
 With Missing Values <https://www.mlflow.org/docs/latest/models.html#handling-integers-with-missing-values>`_ for more
 details."
```

## How is this patch tested?

New unit tests for warning behavior, existing unit + integration tests for correctness

## Release Notes

Improve clarity of warning messages raised during autologging

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -387,8 +387,6 @@ def _is_testing():
         - Disables exception handling for patched function logic, ensuring that patch code
           executes without errors during testing
     """
-    import os
-
     return os.environ.get(_AUTOLOGGING_TEST_MODE_ENV_VAR, "false") == "true"
 
 
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -1066,6 +1066,8 @@ def _augment_mlflow_warnings(autologging_integration):
     def autologging_showwarning(message, category, filename, lineno, *args, **kwargs):
         mlflow_root_path = Path(os.path.dirname(mlflow.__file__)).resolve()
         warning_source_path = Path(filename).resolve()
+        # If the warning's source file is contained within the MLflow package's base
+        # directory, it is an MLflow warning and should be emitted via `logger.warning`
         if mlflow_root_path in warning_source_path.parents:
             _logger.warning(
                 "MLflow issued a warning during %s autologging:" ' "%s:%d: %s: %s"',
@@ -1076,7 +1078,7 @@ def _augment_mlflow_warnings(autologging_integration):
                 message,
             )
         else:
-            original_showwarning(message, category, filename, *args, **kwargs)
+            original_showwarning(message, category, filename, lineno, *args, **kwargs)
 
     try:
         # NB: Reassigning `warnings.showwarning` is the standard / recommended approach for
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -919,7 +919,7 @@ def safe_patch(
 
         with _augment_mlflow_warnings(
             autologging_integration
-        ), _AutologgingSessionManager.start_session(autologging_integration):
+        ), _AutologgingSessionManager.start_session(autologging_integration) as session:
             try:
 
                 def call_original(*og_args, **og_kwargs):
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -691,7 +691,7 @@ def safe_patch(
         # The active MLflow run (if any) associated with patch code execution
         patch_function_run_for_testing = None
 
-        with _redirect_mlflow_warnings(
+        with _augment_mlflow_warnings(
             autologging_integration
         ), _AutologgingSessionManager.start_session(autologging_integration):
             try:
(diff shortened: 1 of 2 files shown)


mlflow/utils/autologging_utils.py
```diff
@@ -1041,6 +1043,52 @@ def call_original(*og_args, **og_kwargs):
...
+    :param autologging_integration: The name of the active autologging integration for which
+                                    MLflow warnings are to be augmented during patch code
+                                    execution.
+    """
+    original_showwarning = warnings.showwarning
+
+    def autologging_showwarning(message, category, filename, lineno, *args, **kwargs):
+        mlflow_root_path = Path(os.path.dirname(mlflow.__file__)).resolve()
+        warning_source_path = Path(filename).resolve()
+        if mlflow_root_path in warning_source_path.parents:
```
apurva-koti, line 1069: Is this a way of checking if the file from where the warning originated is a descendant of the mlflow directory?

dbczumar: Yessir!

mlflow/utils/autologging_utils.py
```diff
@@ -1041,6 +1043,52 @@ def call_original(*og_args, **og_kwargs):
...
+            _logger.warning(
+                "MLflow issued a warning during %s autologging:" ' "%s:%d: %s: %s"',
+                autologging_integration,
+                filename,
+                lineno,
+                category.__name__,
+                message,
+            )
+        else:
+            original_showwarning(message, category, filename, *args, **kwargs)
```
dbczumar, line 1079: Whoops! Missing lineno here...

mlflow/utils/autologging_utils.py
dbczumar: Will add an explanatory comment!

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -734,6 +735,66 @@ def patch_impl(original, *args, **kwargs):
...
+
+    with pytest.warns(UserWarning) as user_warnings_outside_patch:
+        warnings.warn_explicit(**mlflow_warning_kwargs)
+        warnings.warn_explicit(**external_warning_kwargs)
+
+    # Verify that MLflow warnings and external warnings are emitted as normal outside
+    # of autologging patch execution
+    assert set([warning.message.args[0] for warning in user_warnings_outside_patch]) == set(
+        ["Mock MLflow warning", "Mock external warning"]
+    )
```
apurva-koti, line 795: Nice :)

mlflow/utils/autologging_utils.py
apurva-koti: Whoops! Missed that myself 😆

mlflow/utils/autologging_utils.py
apurva-koti: Just curious, is there any other way of doing this? Never seen this approach before. (Don't really have an issue with it myself though)

dbczumar (commented):
@apurva-koti Addressed your comment, fixed the `lineno` issue, and improved test coverage for various warning attributes (lineno, file name, message)

apurva-koti (approved):
Nice!! 🍔

PR #3894, opened 2020-12-22
Add support for tf serving input to REST API

wentinghu:
Signed-off-by: Wendy Hu <wendy.hu@databricks.com>

Add support for tf serving input to REST API

## How is this patch tested?

Added tests to scoring server to ensure tf-serving formatted JSON inputs work as expected

Added support for tf serving input to REST API

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyfunc/scoring_server/__init__.py
+++ b/mlflow/pyfunc/scoring_server/__init__.py
@@ -60,6 +60,40 @@ CONTENT_TYPES = [
 _logger = logging.getLogger(__name__)
 
 
+def infer_and_parse_json_input(json_input, schema: Schema = None):
+    """
+    :param json_input: A JSON-formatted string representation of TF serving input or a Pandas
+                       DataFrame, or a stream containing such a string representation.
+    :param schema: Optional schema specification to be used during parsing.
+    """
+    try:
+        decoded_input = json.loads(json_input)
+    except json.decoder.JSONDecodeError:
+        _handle_serving_error(
+            error_message=(
+                "Failed to parse input from JSON. Ensure that input is a valid JSON"
+                " formatted string."
+            ),
+            error_code=MALFORMED_REQUEST,
+        )
+
+    if isinstance(decoded_input, list):
+        return parse_json_input(json_input=json_input, orient="records", schema=schema)
+    elif isinstance(decoded_input, dict):
+        if "instances" in decoded_input or "inputs" in decoded_input:
+            return parse_tf_serving_input(decoded_input)
+        else:
+            return parse_json_input(json_input=json_input, orient="split", schema=schema)
+    else:
+        _handle_serving_error(
+            error_message=(
+                "Failed to parse input from JSON. Ensure that input is a valid JSON"
+                " list or dictionary."
+            ),
+            error_code=MALFORMED_REQUEST,
+        )
+
+
 def parse_json_input(json_input, orient="split", schema: Schema = None):
     """
     :param json_input: A JSON-formatted string representation of a Pandas DataFrame, or a stream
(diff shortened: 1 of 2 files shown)


tests/pyfunc/test_scoring_server.py
```diff
@@ -346,6 +422,70 @@ def test_parse_with_schema(pandas_df_with_all_types):
...
+    tfserving_input = {
+        "instances": [
+            {"a": "s1", "b": 1},
+            {"a": "s2", "b": 2, "c": [4, 5, 6]},
+            {"a": "s3", "b": 3, "c": [7, 8, 9]},
+        ]
+    }
+    with pytest.raises(MlflowException) as ex:
+        pyfunc_scoring_server.parse_tf_serving_input(tfserving_input)
+    assert "The length of values for each input/column name are not the same" in str(ex)
```
harupy, line 449:
```suggestion
    with pytest.raises(MlflowException, match="The length of values for each input/column name are not the same"):
        pyfunc_scoring_server.parse_tf_serving_input(tfserving_input)
```

Can we use a `match` argument ([ref](https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions)) here? I know we're using a bare pytest-raises (= `pytest.raises` without `match`) in many tests. I'm planning to fix this.

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -184,7 +241,23 @@ def transformation():  # pylint: disable=unused-variable
             data = flask.request.data.decode("utf-8")
             csv_input = StringIO(data)
             data = parse_csv_input(csv_input=csv_input)
-        elif flask.request.content_type in [CONTENT_TYPE_JSON, CONTENT_TYPE_JSON_SPLIT_ORIENTED]:
+        elif flask.request.content_type == CONTENT_TYPE_JSON:
+            json_str = flask.request.data.decode("utf-8")
+            try:
+                inp_dict = json.loads(json_str)
```
tomasatdatabricks, line 247: the input does not have to be a dict, it can also be a list in case the data is in orient='records'

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -184,7 +241,23 @@ def transformation():  # pylint: disable=unused-variable
...
+                    error_message=(
+                        "Failed to parse input from JSON. Ensure that input is a valid JSON"
+                        " formatted string."
+                    ),
+                    error_code=MALFORMED_REQUEST,
+                )
+            if "instances" in inp_dict or "inputs" in inp_dict:
+                data = parse_tf_serving_input(inp_dict)
+            else:
+                data = parse_json_input(json_input=json_str, orient="split", schema=input_schema)
```
tomasatdatabricks, line 259: we also want to infer the correct orient value in this case based on the input type, list -> orient='records', dict with at least 'values' attribute -> orient='split'.

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -127,6 +127,63 @@ def parse_split_oriented_json_input_to_numpy(json_input):
...
 
 
+def parse_tf_serving_input(inp_dict):
+    """
+    :param inp_dict: A dict deserialized from a JSON string formatted as described in TF's
+                     serving API doc
+                     (https://www.tensorflow.org/tfx/serving/api_rest#request_format_2)
+    """
+    # pylint: disable=broad-except
+    if "instances" in inp_dict and "inputs" in inp_dict:
```
tomasatdatabricks, line 137: I think we should check that the input contains exactly one of ("instances", "inputs") and nothing else. In particular, we should look for signature_name and return an error informing users that signature_name is not supported.

tests/pyfunc/test_scoring_server.py
```diff
@@ -346,6 +422,71 @@ def test_parse_with_schema(pandas_df_with_all_types):
     assert all(df["bad_boolean"] == [True, False, True])
 
 
+def test_parse_tf_serving_input():
```
tomasatdatabricks, line 425: Can we also add a test that the json parser detects and correctly parses pandas dataframe with the 2 supported orient encodings ('records' and 'split')?

tomasatdatabricks (commented):
Looks good, left some comments.

tomasatdatabricks (approved):
lgtm!

PR #3946, opened 2021-01-05
Created request_header_provider and implemented databricks_request_header

jimmyxu-db:

Adding a new plugin interface: the RequestHeaderProvider. This plugin will create a dict of request headers to attach to outgoing requests. The actual attaching to requests will happen in a follow up PR.

Implemented the DatabricksRequestHeaderProvider, which creates request headers for databricks environments.

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/request_header/databricks_request_header_provider.py
+++ b/mlflow/tracking/request_header/databricks_request_header_provider.py
@@ -9,7 +9,6 @@ class DatabricksRequestHeaderProvider(RequestHeaderProvider):
     def request_headers(self):
         request_headers = {}
         if databricks_utils.is_in_databricks_notebook():
-            print("hello")
             request_headers["notebook_id"] = databricks_utils.get_notebook_id()
         if databricks_utils.is_in_databricks_job():
             request_headers["job_id"] = databricks_utils.get_job_id()
--- a/mlflow/tracking/request_header/databricks_request_header_provider.py
+++ b/mlflow/tracking/request_header/databricks_request_header_provider.py
@@ -4,7 +4,11 @@ from mlflow.utils import databricks_utils
 
 class DatabricksRequestHeaderProvider(RequestHeaderProvider):
     def in_context(self):
-        return databricks_utils.is_in_cluster()
+        return (
+            databricks_utils.is_in_cluster()
+            or databricks_utils.is_in_databricks_notebook()
+            or databricks_utils.is_in_databricks_job()
+        )
 
     def request_headers(self):
         request_headers = {}
--- a/mlflow/tracking/request_header/databricks_request_header_provider.py
+++ b/mlflow/tracking/request_header/databricks_request_header_provider.py
@@ -1,0 +1,21 @@
+from mlflow.tracking.request_header.abstract_request_header_provider import RequestHeaderProvider
+from mlflow.utils import databricks_utils
+
+
+class DatabricksRequestHeaderProvider(RequestHeaderProvider):
+    def in_context(self):
+        return databricks_utils.is_in_cluster()
+
+    def request_headers(self):
+        request_headers = {}
+        if databricks_utils.is_in_databricks_notebook():
+            print("hello")
+            request_headers["notebook_id"] = databricks_utils.get_notebook_id()
+        if databricks_utils.is_in_databricks_job():
+            request_headers["job_id"] = databricks_utils.get_job_id()
+            request_headers["job_run_id"] = databricks_utils.get_job_run_id()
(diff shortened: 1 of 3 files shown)


mlflow/tracking/request_header/databricks_request_header_provider.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class DatabricksRequestHeaderProvider(RequestHeaderProvider):
+    def in_context(self):
+        return databricks_utils.is_in_cluster()
+
+    def request_headers(self):
+        request_headers = {}
+        if databricks_utils.is_in_databricks_notebook():
+            print("hello")
```
mohamad-arabi, line 12: Do you want to remove this?

mlflow/tracking/request_header/databricks_request_header_provider.py
```diff
@@ -0,0 +1,21 @@
...
+            request_headers["notebook_id"] = databricks_utils.get_notebook_id()
+        if databricks_utils.is_in_databricks_job():
```
mohamad-arabi, line 14: Are these if statements mutually exclusive? should these be `if` or `else if`. Can `is_in_databricks_notebook ` and `is_in_databricks_job ` both be true?

mlflow/tracking/request_header/databricks_request_header_provider.py
```diff
@@ -0,0 +1,21 @@
+from mlflow.tracking.request_header.abstract_request_header_provider import RequestHeaderProvider
+from mlflow.utils import databricks_utils
+
+
+class DatabricksRequestHeaderProvider(RequestHeaderProvider):
+    def in_context(self):
+        return databricks_utils.is_in_cluster()
```
dbczumar, line 7: Is the cluster ID guaranteed to be present in all contexts? I'm wondering if instead we should return `is_in_cluster() or is_in_databricks_notebook() or is_in_databricks_job()` to make sure that we still use a subset of headers even if the cluster ID can't be fetched

tests/tracking/request_header/test_databricks_request_header_provider.py
```diff
@@ -0,0 +1,68 @@
...
+from mlflow.tracking.request_header.databricks_request_header_provider import (
+    DatabricksRequestHeaderProvider,
+)
+
+
+def test_databricks_cluster_run_context_in_context():
+    with mock.patch("mlflow.utils.databricks_utils.is_in_cluster") as in_cluster_mock:
+        assert DatabricksRequestHeaderProvider().in_context() == in_cluster_mock.return_value
+
+
```
dbczumar, line 13: Can we add a couple of tests for the case where we're in multiple contexts, ensuring that all relevant information is captured: one for the case where we have cluster ID + notebook info available and one for the case where we have cluster ID + job info available?

mlflow/tracking/request_header/databricks_request_header_provider.py
dbczumar: While it's probably not the case that notebook id / job id can be present together, the MLflow client shouldn't consider it an error case if they are. Notebook id can be present with cluster id, and job id can also be present with cluster id.

Accordingly, I think it's best to capture all information that's available, preferring the repeated `if` structure over `if, elif`.

mlflow/tracking/request_header/databricks_request_header_provider.py
jimmyxu-db: Good idea!

mlflow/tracking/request_header/databricks_request_header_provider.py
jimmyxu-db: i think it makes sense to do separate ifs to return what we have

tests/tracking/request_header/test_databricks_request_header_provider.py
jimmyxu-db: Do you mean to update the `in_context()` to be an OR of the 3 environments and then update this test accordingly?

jimmyxu-db: Don't the existing tests add whatever info we have available?

dbczumar: They don’t verify that we add cluster id and notebook info if both are available. I.e. I’m pretty sure that all tests would pass if we used the kind of if / elif logic that mohamad asked about above.

dbczumar (approved):
LGTM! Thanks @jimmyxu-db !

Issue #3953, opened 2021-01-07
[FR] mlflow gc working for s3 artifact-store


sklingel:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [x] Yes. I can contribute this feature independently. (I can try)
- [x] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary
Fill the function 'delete_artifacts'  in 'mlflow/store/s3_artifact_repo.py', instead of throwing an MLFlowExpection

## Motivation
Enable mlflow gc to work with s3 aritfact deletion

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [x] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interfaces
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Languages 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
[...]

PR #3958, opened 2021-01-08
mlflow gc - delete_artifacts() for s3

sklingel:

Implements #3953. S3 mlflow gc artifact deletion. 

## How is this patch tested?
unit-test with s3 mock

Enables ```mlflow gc``` to work with s3 artifacts

Components 

Interface 

Language 

Integrations

### DIFF
--- a/tests/store/artifact/test_s3_artifact_repo.py
+++ b/tests/store/artifact/test_s3_artifact_repo.py
@@ -218,3 +218,32 @@ def test_get_s3_file_upload_extra_args_invalid_json():
 
     with pytest.raises(ValueError):
         S3ArtifactRepository.get_s3_file_upload_extra_args()
+
+
+def test_delete_artifacts(s3_artifact_root, tmpdir):
+    subdir_path = str(tmpdir.mkdir("subdir"))
+    nested_path = os.path.join(subdir_path, "nested")
+    os.makedirs(nested_path)
+    path_a = os.path.join(subdir_path, "a.txt")
+    path_b = os.path.join(subdir_path, "b.tar.gz")
+    path_c = os.path.join(nested_path, "c.csv")
+
+    with open(path_a, "w") as f:
+        f.write("A")
+    with tarfile.open(path_b, "w:gz") as f:
+        f.add(path_a)
+    with open(path_c, "w") as f:
+        f.write("col1,col2\n1,3\n2,4\n")
+
+    repo = get_artifact_repository(posixpath.join(s3_artifact_root, "some/path"))
+    repo.log_artifacts(subdir_path)
+
+    # confirm that artifacts are present
+    artifact_file_names = [obj.path for obj in repo.list_artifacts()]
+    assert "a.txt" in artifact_file_names
+    assert "b.tar.gz" in artifact_file_names
+    assert "nested" in artifact_file_names
+
+    repo.delete_artifacts()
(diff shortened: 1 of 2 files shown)


sklingel:
Would it also make sense to update the documentation, mentioning that gc can also delete s3 artifacts?

tests/store/artifact/test_s3_artifact_repo.py
```diff
@@ -218,3 +218,31 @@ def test_get_s3_file_upload_extra_args_invalid_json():
...
+        f.add(path_a)
+    with open(path_c, "w") as f:
+        f.write("col1,col2\n1,3\n2,4\n")
+
+    repo = get_artifact_repository(posixpath.join(s3_artifact_root, "some/path"))
+    repo.log_artifacts(subdir_path)
+
+    repo.delete_artifacts()
+
+    tmpdir_objects = repo.list_artifacts()
```
dbczumar, line 243: Nit: Can we confirm that `list_artifacts()` *does* contain `path_a`, `path_b`, etc before we call `delete`?

dbczumar, line 243: I think the artifacts list should be empty after deletion, right?

Also, the `path` attributes of artifact info returned by `list_artifacts` refer to run-relative artifact paths (e.g., `"a.txt"` rather than `path_a`. (See https://mlflow.org/docs/latest/rest-api.html#fileinfo)

sklingel: Good point, I will insert an assert before

sklingel: > I think the artifacts list should be empty after deletion, right?
Again, right :). Just checking that the list is empty should be enough, instead of 3 asserts. 

I will provide an update soon.

sklingel:
@dbczumar, I have included your suggestions in the newest version. Thanks for the feedback

dbczumar (approved):
LGTM! Thanks so much @sklingel !

tests/store/artifact/test_s3_artifact_repo.py
```diff
@@ -218,3 +218,32 @@ def test_get_s3_file_upload_extra_args_invalid_json():
...
+
+    # confirm that artifacts are present
+    artifact_file_names = [obj.path for obj in repo.list_artifacts()]
+    assert "a.txt" in artifact_file_names
+    assert "b.tar.gz" in artifact_file_names
+    assert "nested" in artifact_file_names
+
+    repo.delete_artifacts()
+    tmpdir_objects = repo.list_artifacts()
+    assert not tmpdir_objects
```
harupy, line 249: Hi @sklingel :)

This is a nit but can we fix this line as follows?

```python
assert tmpdir_objects == []
```

because any falsy value (0, "", None, (), False) can pass this assertion.

dbczumar: I don’t think that’s necessary. We have separate test coverage for the return type / behavior of list artifacts. We also use falsey object comparison of this form throughout other test suites.

harupy: Ok makes senes :)

@sklingel Sorry for the confusion 😅

bushaev:
is this documented ? I don't see how to specify artifact store [here in the documentation for `mlflow gc`](https://www.mlflow.org/docs/latest/cli.html)

PR #3964, opened 2021-01-11
[MINOR FIX] Unpin pip and try new dependency resolver introduced in 20.3

harupy:

Unpin pip and try new dependency resolver introduced in 20.3

https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020 says:

> The most significant changes to the resolver are:
> - It will reduce inconsistency: it will no longer install a combination of packages that is mutually inconsistent. In older versions of pip, it is possible for pip to install a package which does not satisfy the declared requirements of another installed package. For example, in pip 20.0, pip install "six<1.12" "virtualenv==20.0.2" does the wrong thing, “successfully” installing six==1.11, even though virtualenv==20.0.2 requires six>=1.12.0,<2 (defined here). The new resolver, instead, outright rejects installing anything if it gets that input.
> - It will be stricter - if you ask pip to install two packages with incompatible requirements, it will refuse (rather than installing a broken combination, like it did in previous versions).

## How is this patch tested?

The existing package installation process doesn't break

Components 

Interface 

Language 

Integrations


dev/install-common-deps.sh
```diff
@@ -28,7 +28,8 @@ conda create -q -n test-environment python=3.6
 source activate test-environment
 
 python --version
-pip install --upgrade pip==19.3.1
+pip install --upgrade pip
+pip --version
```
harupy, line 32: Note that this change won't affect the cross version tests because we don't use `dev/install-common-deps.sh` in `cross-version-tests.yml`

harupy, line 32: Confirmed the pip version has been updated:

```
++ pip --version
pip 21.0.1 from /usr/share/miniconda/envs/test-environment/lib/python3.6/site-packages/pip (python 3.6)
```

https://github.com/mlflow/mlflow/runs/1860085508#step:4:245

harupy: And all the CI checks have passed :)

dbczumar (approved):
LGTM!

harupy:
@dbczumar Thanks!

dev/install-common-deps.sh
WeichenXu123, line 32: Nit:
Shall we use 
```
echo "Upgrade pip to version: `pip --version`"
```
to make it more clear ?

WeichenXu123 (approved):

WeichenXu123:
Should we add flag to suppress warning like:
```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.
We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.
```

harupy:
I don't think we should because it makes it difficult to detect warnings (which may turn into erros in the future)

dev/install-common-deps.sh
harupy: @WeichenXu123 Thanks for the comment. I think `pip install --upgrade pip` prints out similar message. What do you think?

```
++ pip install --upgrade pip
Requirement already satisfied: pip in /usr/share/miniconda/envs/test-environment/lib/python3.6/site-packages (20.3.3)
Collecting pip
  Downloading pip-21.0.1-py3-none-any.whl (1.5 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.3.3
    Uninstalling pip-20.3.3:
      Successfully uninstalled pip-20.3.3
Successfully installed pip-21.0.1
++ pip --version
pip 21.0.1 from /usr/share/miniconda/envs/test-environment/lib/python3.6/site-packages/pip (python 3.6)
```

WeichenXu123: then it's fine

PR #3972, opened 2021-01-12
[SkinnyClient] Remove pandas and numpy dependency

eddiestudies:

Remove pandas and numpy dependencies from the MLflow Skinny Client

## How is this patch tested?

New tests are added for the search_runs behavior. Existing tests cover the expected scenarios.

Components 

Interface 

Language 

Integrations


mlflow/models/__init__.py
```diff
@@ -22,7 +22,15 @@
...
 
-__all__ = ["Model", "ModelSignature", "ModelInputExample", "infer_signature", "FlavorBackend"]
+__all__ = ["Model", "FlavorBackend"]
+
+
+try:
+    from .signature import ModelSignature, infer_signature
+    from .utils import ModelInputExample
+except ImportError
+    pass
```
eedeleon, line 33: add logging here

eedeleon:
Missing tests for search runs as_pandas and the results

dbczumar (commented):
@eedeleon Upon seeing the full impact of this approach on the stronger typing system associated with model signatures, I'm wondering if we might be able to circumvent a lot of the import challenges while retaining support for most of the Python 3 type hints by try/catching the definition of `MlflowInferableDataset` in `mlflow.signature` after all (https://github.com/mlflow/mlflow/blob/a0b1b59656cae4cf8acc6076a8dedb8f46c4002e/mlflow/models/signature.py#L15-L22).

I realize that you brought this up before. Reexamining the layout of the model signatures type system, I think the definition of `MlflowInferableDataset` is one of the only things we need to try/catch right now in order to allow us to continue importing `ModelSignature` / `ModelInputExample` in the various MLflow Models integrations. If pandas / numpy / pyspark aren't installed, then we can fall back to assigning `MlflowInferableDataset` to `Any`. We may also have to tuck pandas / numpy imports into some class definitions (e.g., in https://github.com/mlflow/mlflow/blob/master/mlflow/types/schema.py).

Can you take a look at this approach and let me know what you think? Sorry for the inconvenience here!

mlflow/pyfunc/__init__.py
```diff
@@ -366,11 +373,13 @@ def all_ints(xs):
...
         raise MlflowException(
             "Incompatible input types for column {0}. "
-            "Can not safely convert {1} to {2}.{3}".format(name, values.dtype, numpy_type, hint)
+            "Can not safely convert {1} to {2}.{3}".format(
+                name, values.dtype, numpy_type, hint
+            )
         )
 
 
-def _enforce_schema(pdf: PyFuncInput, input_schema: Schema):
```
dbczumar, line 373: @tomasatdatabricks Are `PyFuncInput` and `PyfuncOutput` part of the public MLflow API - i.e. do we expect users to import and interact with these directly?

If not, IMO, I think the `PyFuncInput` type hint is acceptable to drop in order to support a pandas/numpy-free skinny client since it's only used in an internal function. However, in accordance with https://github.com/mlflow/mlflow/pull/3972#pullrequestreview-568001973, it seems important to keep user-facing type hints around (e.g. `ModelSignature`, `Schema`, and `ModelInputExample` when used in user-facing MLflow Models functions).

dbczumar: Doh, I missed an instance. This is a public APi.

mlflow/pyfunc/__init__.py
```diff
@@ -237,8 +237,6 @@
 PY_VERSION = "python_version"
 
 _logger = logging.getLogger(__name__)
-PyFuncInput = Union[pandas.DataFrame, np.ndarray, List[Any], Dict[str, Any]]
-PyFuncOutput = Union[pandas.DataFrame, pandas.Series, np.ndarray, list]
```
dbczumar, line 241: These types help define the input / output interfaces for pyfunc models. Can we try/catch the definition for these in a similar fashion to the suggestion made in https://github.com/mlflow/mlflow/pull/3972#pullrequestreview-568001973 (aka your original suggestion from about a week ago)?

mlflow/pyfunc/__init__.py
dbczumar: Left a comment in https://github.com/mlflow/mlflow/pull/3972/files#r557198770 indicating that these should be preserved as they're public-facing type hints.

mlflow/pyfunc/__init__.py
eedeleon: Makes sense. It felt odd removing this from so many places.

This should reduce the # of lines in the PR too. I really wish Union were more malleable than it is. Union.create(list_of_possible_types)  would really simplify the logic but is not a thing.

mlflow/models/__init__.py
```diff
@@ -22,7 +22,19 @@
 
 from .model import Model
 from .flavor_backend import FlavorBackend
```
dbczumar, line 24: Do we need `Model` to be importable in the absence of `ModelSignature` / `infer_signature`? Given that pandas and numpy are requirements for the MLflow Models component, it's not clear that we need to make this distinction at this point. 

I see that we're importing `mlflow.models.cli` in `mlflow/cli.py`, which is presumably why this change was made (let me know if there are other places where we're importing from `mlflow.models` within tracking code). Instead, could we make `mlflow.models.cli` a conditional import in `mlflow.cli` (as you've done for `mlflow.azureml.cli` and `mlflow.sagemaker.cli`)?

mlflow/cli.py
```diff
@@ -455,12 +454,26 @@ def gc(backend_store_uri, run_ids):
 
 cli.add_command(mlflow.models.cli.commands)
```
dbczumar, line 455: Re: https://github.com/mlflow/mlflow/pull/3972/files#r560013572 - can we make `mlflow.models.cli` a conditional import here as well?

mlflow/__init__.py
```diff
@@ -46,22 +46,54 @@
...
+try:
+    # pylint: disable=unused-import
+    import pandas  # noqa: E402
+    import numpy  # noqa: E402
+except ImportError as e:
+    warnings.warn(
+        "Model flavors could not be loaded due to missing numpy " "or pandas dependencies. %s." % e
+    )
+else:
+    _supports_flavors = True
```
dbczumar, line 59: Unused boolean

mlflow/__init__.py
```diff
@@ -46,22 +46,54 @@
...
-import mlflow.shap as shap  # noqa: E402
+_model_flavors_supported = []
+try:
+    # pylint: disable=unused-import
+    import pandas  # noqa: E402
+    import numpy  # noqa: E402
+except ImportError as e:
+    warnings.warn(
+        "Model flavors could not be loaded due to missing numpy " "or pandas dependencies. %s." % e
+    )
```
dbczumar, line 57: I don't think we need to emit a separate warning message here. Users of the skinny client shouldn't expect the MLflow Models component to work out of the box (given the documentation that we'll provide), and users of the full client shouldn't encounter this warning.

Accordingly, I think it's acceptable to try/catch `ImportError` for all of the MLflow Models imports. We have test coverage ensuring that the full client successfully imports all of MLflow's flavor modules: https://github.com/mlflow/mlflow/blob/032ec8841dad13492a6eb3c9acdf78803ed90f76/tests/test_flavors.py#L36.

mlflow/cli.py
```diff
@@ -455,12 +454,26 @@ def gc(backend_store_uri, run_ids):
...
 
+try:
+    # pylint: disable=unused-import
+    import pandas  # noqa: E402
+    import numpy  # noqa: E402
+except ImportError as e:
+    warnings.warn(
+        "Built in deployment plugins could not be loaded due to missing numpy "
+        "or pandas dependencies. %s." % e
+    )
```
dbczumar, line 470: In a similar vein to https://github.com/mlflow/mlflow/pull/3972/files#r560019940, I don't think we need to emit a warning here as long as we document what is / isn't supported by the skinny client. I think a warning here will create more noise than benefit / clarity for skinny client users.

mlflow/tracking/fluent.py
```diff
@@ -955,6 +940,7 @@ def search_runs(
     run_view_type=ViewType.ACTIVE_ONLY,
     max_results=SEARCH_MAX_RESULTS_PANDAS,
     order_by=None,
+    as_pandas=True,
```
dbczumar, line 943: Nit: I think it would be clearer / more pythonic / more extensible to use `format="pandas"` and `format="list"` rather than `as_pandas=True/False`. Let me know what you think here - sorry for the hassle!

mlflow/utils/proto_json_utils.py
```diff
@@ -83,8 +80,8 @@ def default(self, o):  # pylint: disable=E0202
 
 
 def _dataframe_from_json(
-    path_or_str, schema: Schema = None, pandas_orient: str = "split", precise_float=False
-) -> pd.DataFrame:
+    path_or_str, schema=None, pandas_orient: str = "split", precise_float=False
+):  # -> pd.DataFrame
```
dbczumar, line 84: Nit: I think we can just drop the `pd.DataFrame` comment since this is an internal utility & the documentation / method name both indicate that a dataframe is returned.

setup.py
```diff
@@ -41,10 +41,9 @@ def package_files(directory):
     "databricks-cli>=0.8.7",
     "entrypoints",
     "gitpython>=2.1.0",
-    "numpy",
-    "pandas",
     "pyyaml",
     "protobuf>=3.6.0",
+    "pytz",
```
dbczumar, line 46: Is this library necessary? I don't see it imported elsewhere in the PR code and I was under the impression that the only datetime functionality used by Pandas is leveraged in `search_runs`, which we're working around by providing an alternative list-based output format.

tests/test_search_runs.py
```diff
@@ -0,0 +1,43 @@
...
+import numpy as np
+import pandas as pd
+import shutil
+import tempfile
+import textwrap
+
+from mlflow import experiments
+
+
+def test_csv_generation():
```
dbczumar, line 14: This test case description doesn't really match the suite / module name (`test_search_runs`), since it's not really testing `search_runs` behavior. Can we move this back to `test_runs` and do a `skipif skinny` ?

mlflow/__init__.py
dbczumar: Can we add an inline comment explaining why we're conditionally importing these modules? I.e. the skinny client doesn't support them

mlflow/cli.py
dbczumar: Re: https://github.com/mlflow/mlflow/pull/3972/files#r560029814 Can we add an inline comment explaining why the conditional import logic is necessary (i.e because the skinny client doesn't support these deployment CLIs due to dependencies on pandas / numpy / MLflow Models, which aren't included in the skinny client)?

tests/tracking/test_tracking.py
```diff
@@ -867,56 +867,58 @@ def test_search_runs():
...
-    def verify_runs(runs, expected_set):
-        assert set([r.info.run_id for r in runs]) == set([logged_runs[r] for r in expected_set])
-
     experiment_id = MlflowClient().get_experiment_by_name("exp-for-search").experiment_id
 
     # 2 runs in this experiment
     assert len(MlflowClient().list_run_infos(experiment_id, ViewType.ACTIVE_ONLY)) == 2
 
+    def run_and_verify(experiment_ids, query, check, **kwargs):
+        runs = MlflowClient().search_runs(experiment_ids, query, **kwargs)
```
dbczumar, line 876: Out of curiosity, how are these tests passing for the skinny client? Is it because installing alembic / sqlalchemy / sqlparse is also pulling in pandas? https://github.com/mlflow/mlflow/blob/032ec8841dad13492a6eb3c9acdf78803ed90f76/dev/run-python-skinny-tests.sh#L17

tests/tracking/test_tracking.py
```diff
@@ -867,56 +867,58 @@ def test_search_runs():
...
+
+        assert set([r.info.run_id for r in runs]) == set([logged_runs[r] for r in check])
+
+        # verify as_pandas=False returns the same as mlflow_client.search_runs
+        fluent_search_runs_run_ids = [
+            run.info.run_id
+            for run in mlflow.search_runs(experiment_ids, query, as_pandas=False, **kwargs)
+        ]
+        assert fluent_search_runs_run_ids == [run.info.run_id for run in runs]
```
dbczumar, line 885: I think fluent search runs behavior is best tested in https://github.com/mlflow/mlflow/blob/master/tests/tracking/fluent/test_fluent.py. Can we add a test case there for the new list output format of `search_runs` and add `tests/tracking/fluent/test_fluent.py` to https://github.com/mlflow/mlflow/blob/master/dev/run-python-skinny-tests.sh, skipping any test cases that we don't expect to pass on the skinny client (e.g., the tests for `search_runs` with *pandas* output format)?

dev/skinny-requirements.txt
```diff
@@ -1,5 +1,3 @@
-## Small test reqs
```
dbczumar, line 1: Speaking of test requirements, is there a way we can be sure that installing `sqlalchemy` / `sqlparse` / `alembic` in our skinny client test environment (https://github.com/mlflow/mlflow/blob/032ec8841dad13492a6eb3c9acdf78803ed90f76/dev/run-python-skinny-tests.sh#L17) doesn't also install pandas and numpy? It'd be a bit of a bummer if numpy / pandas were somehow pulled in and we ended up missing pd / np imports in skinny client modules (tracking, etc)

mlflow/tracking/fluent.py
eedeleon: format is a builtin funciton in python. I am changing to format but I usually avoid overriding builtins

eedeleon: I thought of output_format but it feels too verbose

setup.py
eedeleon: mlflow/utils/time_utils.py:from pytz import reference

mlflow/tracking/fluent.py
eedeleon: I am not sure conditional outputs is worth making extensible

tests/statsmodels/model_fixtures.py:    data = sm.datasets.scotland.load(as_pandas=False)

We won't be the only library making this distinction either so it might be ok to keep it rather than allow for that extensibility

xgboost, statsmodels, and autogluon also uses as_pandas

alphavantage and iexfinance uses output_format='pandas' probably for the builtin issue

I think given statsmodels and xgboost are also flavors it might be ok to mirror their approach with as_pandas unless we have intentions to add more formats

dev/skinny-requirements.txt
eedeleon: I split test_skinny.py into test_skinny and test_skinny_sql. 

Test_skinny_sql runs before the pip installs and tests_skinny, checks pandas, after the pip installs

setup.py
dbczumar: Ah, cool. This was pulled in transitively via pandas before?

mlflow/__init__.py
```diff
@@ -46,22 +46,52 @@
...
+_model_flavors_supported = []
+try:
+    # pylint: disable=unused-import
+    import pandas  # noqa: E402
+    import numpy  # noqa: E402
+except ImportError as e:
+    # We are conditional loading these commands since the skinny client does
+    # not support them due to the pandas and numpy dependencies of MLflow Models
+    pass
+else:
```
dbczumar, line 58: I'd rather we just try to import the flavor modules directly and catch `ImportError` (rather than try/catching import error on numpy and pandas preemptively). Our test coverage for the complete / core client will verify that all of these modules are successfully imported.

mlflow/cli.py
```diff
@@ -453,14 +450,28 @@ def gc(backend_store_uri, run_ids):
...
 
+try:
+    # pylint: disable=unused-import
+    import pandas  # noqa: E402
+    import numpy  # noqa: E402
+except ImportError as e:
+    # We are conditional loading these commands since the skinny client does
+    # not support them due to the pandas and numpy dependencies of MLflow Models
+    pass
+else:
```
dbczumar, line 467: I'd rather we just try to import the CLI modules directly and catch `ImportError` (rather than try/catching import error on numpy and pandas preemptively). Our test coverage for the azureml / sagemaker / models CLIs should be sufficient to ensure that these CLIs are importable / usable when the full set of core dependencies is specified.

dev/skinny-requirements.txt
dbczumar: Were those changes pushed?

mlflow/tracking/fluent.py
```diff
@@ -955,6 +940,7 @@ def search_runs(
     run_view_type=ViewType.ACTIVE_ONLY,
     max_results=SEARCH_MAX_RESULTS_PANDAS,
     order_by=None,
+    format="pandas",
```
dbczumar, line 943: As lengthy as `output_format` is, I think that works well here (didn't realize that `format` was a builtin). Could we make that minor change?

dbczumar (commented):
@eedeleon I think this is just about ready once the tests are passing and a couple of nitty comments are addressed. Thanks so much!

mlflow/tracking/fluent.py
eedeleon: yup

setup.py
eedeleon: yup

tests/tracking/test_tracking.py
eedeleon: I added tests but avoided the ones with NaN and datetime that wasn't now

eedeleon: Which was 1 test

mlflow/__init__.py
eedeleon: done

eedeleon:
I made the changes. Thank you for the quick reviews! 

I wanted to do a final PR for the release so we can do more docs and other release related changes. It would mostly be setup.py and devops changes.

mlflow/tracking/fluent.py
```diff
@@ -968,11 +954,16 @@ def search_runs(
     :param order_by: List of columns to order by (e.g., "metrics.rmse"). The ``order_by`` column
                      can contain an optional ``DESC`` or ``ASC`` value. The default is ``ASC``.
                      The default ordering is to sort by ``start_time DESC``, then ``run_id``.
+    :param output_format: The output format to be returned. If "pandas" a pandas.DataFrame
+                          and if "list" List[mlflow.entitie.Run].
```
dbczumar, line 958:
```suggestion
                          and if "list" List[py:class:`mlflow.entities.Run`].
```

mlflow/tracking/fluent.py
```diff
@@ -968,11 +954,16 @@ def search_runs(
...
 
-    :return: A pandas.DataFrame of runs, where each metric, parameter, and tag
-        are expanded into their own columns named metrics.*, params.*, and tags.*
-        respectively. For runs that don't have a particular metric, parameter, or tag, their
-        value will be (NumPy) Nan, None, or None respectively.
+    :return: If output_format is "pandas":  pandas.DataFrame of runs, where each metric, parameter,
+             and tag are expanded into their own columns named metrics.*, params.*, and tags.*
+             respectively. For runs that don't have a particular metric, parameter, or tag, their
+             value will be (NumPy) Nan, None, or None respectively.
+             If output_format is "list": List[mlflow.entitie.Run].
```
dbczumar, line 966:
```suggestion
             If output_format is "list": List[py:class:`mlflow.entities.Run`].
```

mlflow/tracking/fluent.py
```diff
@@ -1016,74 +1007,84 @@ def search_runs(
...
+        for key in metrics:
+            data["metrics." + key] = metrics[key]
+        for key in params:
+            data["params." + key] = params[key]
+        for key in tags:
+            data["tags." + key] = tags[key]
+        return pd.DataFrame(data)
+    else:
+        raise ValueError(
+            "Unsupported format: %s. Supported strings are 'pandas' or 'list'" % output_format
```
dbczumar, line 1086:
```suggestion
            "Unsupported output format: %s. Supported string values are 'pandas' or 'list'" % output_format
```

tests/tracking/fluent/test_fluent.py
```diff
@@ -519,26 +520,81 @@ def test_get_run():
         assert run.info.user_id == "my_user_id"
 
 
+def call_search_runs():
+    return search_runs(output_format="list") if "MLFLOW_SKINNY" in os.environ else search_runs()
```
dbczumar, line 524: Ideally, it'd be nice to test both output formats in the full client and skip the pandas output format tests in the skinny client. Perhaps parameterizing search_runs test cases by `output_format` and conditionally defining `output_format` based on `os.environ["MLFLOW_SKINNY"]` would help (via `@pytest.mark.parametrize`).

tests/tracking/fluent/test_fluent.py
```diff
@@ -519,26 +520,81 @@ def test_get_run():
...
+        for run in results:
+            result_data["status"].append(run.info.status)
+            result_data["artifact_uri"].append(run.info.artifact_uri)
+            result_data["experiment_id"].append(run.info.experiment_id)
+            result_data["run_id"].append(run.info.run_id)
+            result_data["start_time"].append(run.info.start_time)
+            result_data["end_time"].append(run.info.end_time)
+
+        assert result_data == data
+    else:
```
dbczumar, line 539: Re: https://github.com/mlflow/mlflow/pull/3972/files#r563478944: this if/else logic depends on the assumption that full client testing always uses `output_format = 'pandas'`, which is a bit hard to infer without comments / documentation. I think it'd be clearer if we tested both output formats in the full client & performed these kinds of branching asserts based on `output_format` rather than `os.environ["MLFLOW_SKINNY"]`.

tests/tracking/test_tracking.py
```diff
@@ -867,56 +867,58 @@ def test_search_runs():
...
+        runs = MlflowClient().search_runs(experiment_ids, query, **kwargs)
+
+        assert set([r.info.run_id for r in runs]) == set([logged_runs[r] for r in check])
+
+        # verify output_fromat="list" returns the same as mlflow_client.search_runs
+        fluent_search_runs_run_ids = [
+            run.info.run_id
+            for run in mlflow.search_runs(experiment_ids, query, output_format="list", **kwargs)
+        ]
+        assert fluent_search_runs_run_ids == [run.info.run_id for run in runs]
```
dbczumar, line 885: I think we can remove this logic now that we're testing `search_runs` behavior in `tests/tracking/fluent/test_fluent.py`. While `search_runs(output_format='list')` produces the same output format as `MlflowClient.search_runs()`, I don't think we need this extra validation logic since we're not explicitly documenting / guaranteeing that these APIs share the same output format.

tests/tracking/test_tracking.py
```diff
@@ -867,56 +867,58 @@ def test_search_runs():
...
 
-    def verify_runs(runs, expected_set):
-        assert set([r.info.run_id for r in runs]) == set([logged_runs[r] for r in expected_set])
-
     experiment_id = MlflowClient().get_experiment_by_name("exp-for-search").experiment_id
 
     # 2 runs in this experiment
     assert len(MlflowClient().list_run_infos(experiment_id, ViewType.ACTIVE_ONLY)) == 2
 
+    def run_and_verify(experiment_ids, query, check, **kwargs):
```
dbczumar, line 875: Nit: 
```suggestion
    def search_and_verify_runs(experiment_ids, query, check, **kwargs):
```

dev/run-python-skinny-tests.sh
```diff
@@ -13,8 +13,9 @@ err=0
 trap 'err=1' ERR
 export MLFLOW_SKINNY='true'
 
-pytest --verbose tests/test_skinny.py
+pytest --verbose tests/test_skinny_sql.py
 python -m pip install sqlalchemy alembic sqlparse
+pytest --verbose tests/test_skinny.py
```
dbczumar, line 18: Nit: Is it possible to provide slightly more descriptive names for these test suites? E.g.:

`test_skinny_client_omits_sql_libs`
`test_skinny_client_omits_data_science_libs`

mlflow/models/model.py
```diff
@@ -31,7 +30,7 @@ def __init__(
         run_id=None,
         utc_time_created=None,
         flavors=None,
-        signature: ModelSignature = None,
+        signature=None,
```
dbczumar, line 33:
```suggestion
        signature=None,  # ModelSignature
```

dbczumar (commented):
@eedeleon LGTM once remaining nit comments are addressed. Overall, impl and tests look great. I've manually confirmed that the following workflows are still supported by the skinny client on Databricks with this change applied (taken from https://github.com/mlflow/mlflow/pull/3687#pullrequestreview-538907153):

Tracking operations (logging / loading / searching params, metrics, tags + logging / loading artifacts on AWS - Azure requires removal of the Azure Storage Blob dependency as a follow-up PR)

Model registration, search, artifact loading, transitions

Execution of GitHub projects within a Databricks notebook & against a Databricks workspace remotely

After installing scikit-learn, numpy, pandas: sklearn autologging + model log / load operations succeed

dbczumar (approved):
LGTM! Thanks @eedeleon

PR #3981, opened 2021-01-14
Temporary workaround for keras EarlyStopping issue

harupy:

A temporary workaround for https://github.com/tensorflow/tensorflow/issues/46429

### tf-nightly 2.5.0.dev20210113 (broken)

```
# training log for `EarlyStopping.patience = 0`

Epoch 1/10
5/5 [==============================] - 0s 871us/step - loss: 5.1400 - acc: 0.3351
Restoring model weights from the end of the best epoch.
Epoch 00001: early stopping

# training should not early stop after the first epoch
# because the loss has improved from inf (automatically set by keras when traning starts) to 5.1400
```

### tf-nightly 2.5.0.dev20210112 (fine)

```
# training log for `EarlyStopping.patience = 0`

Epoch 1/10
5/5 [==============================] - 0s 871us/step - loss: 5.1400 - acc: 0.3351
Epoch 2/10
5/5 [==============================] - 0s 837us/step - loss: 4.3136 - acc: 0.3537
Restoring model weights from the end of the best epoch.
Epoch 00002: early stopping

# training continues after the first epoch and early stops after the second epoch
# because loss improvement (5.14000 - 4.3136) is smaller than specified `min_delta`
# which is 99999999 (we intentionally use this value to enforce early stopping regardless of loss value)
# This is what our unit tests expect
```

## How is this patch tested?

Existing unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/ml-package-versions.yml
+++ b/ml-package-versions.yml
@@ -97,7 +97,7 @@ keras:
       # https://github.com/keras-team/keras#multi-backend-keras-and-tfkeras
       # Unpin `tf-nightly` once the following issue has been fixed:
       # https://github.com/tensorflow/tensorflow/issues/46429
-      pip install 'tf-nightly==2.5.0.dev20210113' 'keras>=2.4.0'
+      pip install 'tf-nightly==2.5.0.dev20210112' 'keras>=2.4.0'
 
   models:
     minimum: "2.2.4"
--- a/ml-package-versions.yml
+++ b/ml-package-versions.yml
@@ -73,6 +73,7 @@ tensorflow-2.x:
   package_info:
     pip_release: "tensorflow"
     install_dev: |
+      # Unpin `tf-nightly` once the following issue is addressed:
       # https://github.com/tensorflow/tensorflow/issues/46429
       pip install 'tf-nightly==2.5.0.dev20210112'
 
@@ -96,7 +97,7 @@ keras:
     install_dev: |
       # In keras >= 2.4.0, `keras` simply redirects to `tensorflow.keras`:
       # https://github.com/keras-team/keras#multi-backend-keras-and-tfkeras
-      # Unpin `tf-nightly` once the following issue has been fixed:
+      # Unpin `tf-nightly` once the following issue is addressed:
       # https://github.com/tensorflow/tensorflow/issues/46429
       pip install 'tf-nightly==2.5.0.dev20210112' 'keras>=2.4.0'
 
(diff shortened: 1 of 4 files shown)


tests/keras_autolog/test_keras_autolog.py
```diff
@@ -205,6 +205,7 @@ def keras_random_data_run_with_callback(
             patience=patience,
             min_delta=99999999,
             restore_best_weights=restore_weights,
+            verbose=1,
```
harupy, line 208: Added this line to trigger the keras tests.

harupy: Can we keep this change to make it easier to debug issues?

harupy:
Is keras early stopping logic broken?

### Code

```python
import tensorflow as tf
import keras
from sklearn.datasets import load_iris
import pandas as pd


def create_model():
    model = keras.Sequential()
    model.add(keras.layers.Dense(16, activation="relu", input_shape=(4,)))
    model.add(keras.layers.Dense(3, activation="softmax"))
    model.compile(
        optimizer=keras.optimizers.Adam(lr=0.001, epsilon=1e-07),
        loss="categorical_crossentropy",
        metrics=["acc"],
    )
    return model


def get_tranining_data():
    X, y = load_iris(return_X_y=True)
    return X, pd.get_dummies(y)


callback = keras.callbacks.EarlyStopping(
    monitor="loss",
    patience=XXX,  # <- 🔥 manually specify patience value
    min_delta=99999999,
    restore_best_weights=True,
    verbose=1,
)


def main():
    model = create_model()
    Xy = get_tranining_data()

    print("tf:", tf.__version__)
    print("keras:", keras.__version__)
    history = model.fit(*Xy, epochs=10, callbacks=[callback], initial_epoch=0)


if __name__ == "__main__":
    main()

```

tests/keras_autolog/test_keras_autolog.py
dbczumar: Sure - makes sense to keep this!

ml-package-versions.yml
```diff
@@ -73,7 +73,8 @@ tensorflow-2.x:
   package_info:
     pip_release: "tensorflow"
     install_dev: |
-      pip install tf-nightly
+      # https://github.com/tensorflow/tensorflow/issues/46429
```
dbczumar, line 76: @harupy Can we file an internal ticket to track unpinning this once https://github.com/tensorflow/tensorflow/issues/46429 is addressed?

dbczumar (approved):
LGTM! Thanks @harupy !

PR #4044, opened 2021-02-02
Drop azure-blob-storage dependency from MLflow client

WeichenXu123:

Drop azure-blob-storage dependency from MLflow client

## How is this patch tested?

UT.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/dev/large-requirements.txt
+++ b/dev/large-requirements.txt
@@ -2,3 +2,4 @@
 matplotlib
 plotly
 tf2onnx
+azure-storage-blob>=12.0.0
--- a/dev/large-requirements.txt
+++ b/dev/large-requirements.txt
@@ -2,4 +2,3 @@
 matplotlib
 plotly
 tf2onnx
-azure-storage-blob>=12.0.0
--- a/dev/lint-requirements.txt
+++ b/dev/lint-requirements.txt
@@ -6,4 +6,3 @@ rstcheck==3.2
 black==19.10b0
 # Pin isort version to prevent pylint from raising an error
 isort<5.0.0
-azure-storage-blob>=12.0.0
--- a/dev/lint-requirements.txt
+++ b/dev/lint-requirements.txt
@@ -6,3 +6,4 @@ rstcheck==3.2
 black==19.10b0
 # Pin isort version to prevent pylint from raising an error
 isort<5.0.0
+azure-storage-blob>=12.0.0
--- a/dev/small-requirements.txt
+++ b/dev/small-requirements.txt
@@ -8,3 +8,4 @@ pytest==3.2.1
 pytest-cov==2.6.0
 pytest-localserver==0.5.0
 moto==1.3.16
+azure-storage-blob>=12.0.0
--- a/mlflow/azure/client.py
+++ b/mlflow/azure/client.py
@@ -27,16 +27,11 @@ def put_block(sas_url, block_id, data, headers):
     headers = deepcopy(headers)
     headers.update(_PUT_BLOCK_HEADERS)
 
-    request_url = _append_query_parameters(
-        sas_url,
-        {
-            "comp": "block",
-            "blockid": block_id,
-        }
-    )
+    request_url = _append_query_parameters(sas_url, {"comp": "block", "blockid": block_id,})
 
     with rest_utils.cloud_storage_http_request(
-            'put', request_url, data, headers=headers) as response:
+        "put", request_url, data, headers=headers
+    ) as response:
         response.raise_for_status()
 
 
@@ -52,16 +47,12 @@ def put_block_list(sas_url, block_list, headers):
                        https://docs.microsoft.com/en-us/rest/api/storageservices/put-block-list.
     :param headers: Headers to include in the Put Block request body.
     """
-    request_url = _append_query_parameters(
-        sas_url,
-        {
-            "comp": "blocklist",
-        }
-    )
+    request_url = _append_query_parameters(sas_url, {"comp": "blocklist",})
     data = _build_block_list_xml(block_list)
 
     with rest_utils.cloud_storage_http_request(
-            'put', request_url, data, headers=headers) as response:
+        "put", request_url, data, headers=headers
+    ) as response:
         response.raise_for_status()
 
 
--- a/mlflow/azure/client.py
+++ b/mlflow/azure/client.py
@@ -1,7 +1,7 @@
 from copy import deepcopy
 import urllib
 
-from mlflow.utils.rest_utils import cloud_storage_http_request
+from mlflow.utils import rest_utils
 
 _PUT_BLOCK_HEADERS = {
     "x-ms-blob-type": "BlockBlob",
@@ -32,7 +32,7 @@ def put_block(sas_url, block_id, data, headers):
         }
     )
 
-    with cloud_storage_http_request('put', request_url, data, headers=headers) as response:
+    with rest_utils.cloud_storage_http_request('put', request_url, data, headers=headers) as response:
         response.raise_for_status()
 
 
@@ -56,7 +56,7 @@ def put_block_list(sas_url, block_list, headers):
     )
     data = _build_block_list_xml(block_list)
 
-    with cloud_storage_http_request('put', request_url, data, headers=headers) as response:
+    with rest_utils.cloud_storage_http_request('put', request_url, data, headers=headers) as response:
         response.raise_for_status()
 
 
--- a/mlflow/azure/client.py
+++ b/mlflow/azure/client.py
@@ -1,0 +1,81 @@
+from copy import deepcopy
+import urllib
+
+from mlflow.utils.rest_utils import cloud_storage_http_request
+
+_PUT_BLOCK_HEADERS = {
+    "x-ms-blob-type": "BlockBlob",
+}
+
+
+def put_block(sas_url, block_id, data, headers):
+    """
+    Performs an Azure `Put Block` operation
+    (https://docs.microsoft.com/en-us/rest/api/storageservices/put-block)
+
+    :param sas_url: A shared access signature URL referring to the Azure Block Blob
+                    to which the specified data should be staged.
+    :param block_id: A base64-encoded string identifying the block.
+    :param data: Data to include in the Put Block request body.
+    :param headers: Additional headers to include in the Put Block request body
+                    (the `x-ms-blob-type` header is always included automatically).
+    """
+    # Copy the headers to avoid mutating the input `headers` dictionary
+    headers = deepcopy(headers)
+    headers.update(_PUT_BLOCK_HEADERS)
+
+    request_url = _append_query_parameters(
+        sas_url,
+        {
+            "comp": "block",
+            "blockid": block_id,
+        }
+    )
+
+    with cloud_storage_http_request('put', request_url, data, headers=headers) as response:
+        response.raise_for_status()
+
+
+def put_block_list(sas_url, block_list, headers):
+    """
+    Performs an Azure `Put Block List` operation
+    (https://docs.microsoft.com/en-us/rest/api/storageservices/put-block-list)
+
+    :param sas_url: A shared access signature URL referring to the Azure Block Blob
+                    to which the specified data should be staged.
+    :param block_list: A list of uncommitted base64-encoded string block IDs to commit. For
+                       more information, see
+                       https://docs.microsoft.com/en-us/rest/api/storageservices/put-block-list.
+    :param headers: Headers to include in the Put Block request body.
+    """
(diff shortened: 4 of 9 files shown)


WeichenXu123:
@dbczumar Could you take an initial review ? thanks!

and are these failure in this CI related to this PR or they are irrelative but flaky ?
https://github.com/mlflow/mlflow/pull/4044/checks?check_run_id=1814513431

mlflow/utils/rest_utils.py
```diff
@@ -169,6 +171,48 @@ def call_endpoint(host_creds, endpoint, method, json_body, response_proto):
...
+    408, # Request Timeout
+    429, # Too Many Requests
+    500, # Internal Server Error
+    502, # Bad Gateway
+    503, # Service Unavailable
+    504, # Gateway Timeout
+]
+
+
+def http_put(*args, **kwargs):
```
dbczumar, line 187: Cool! Can we include & leverage a similar `http_get` method to provide retries for AWS / Azure / GCP uploads & downloads (https://github.com/mlflow/mlflow/blob/ef17afd5766dff33cf2c8d963223eb7be4a61b61/mlflow/store/artifact/databricks_artifact_repo.py#L194-L206, https://github.com/mlflow/mlflow/blob/ef17afd5766dff33cf2c8d963223eb7be4a61b61/mlflow/store/artifact/databricks_artifact_repo.py#L244)?

dbczumar, line 187: Can we further distinguish this method from the preexisting `http_request` function in this module by calling this `cloud_storage_http_put` and adds some docs indicating that this implements retries for an HTTP put operation in accordance with cloud vendor guidelines for storage APIs?

dbczumar: In conjunction with https://github.com/mlflow/mlflow/pull/4044/files#r569026315, it might be good to define a single `cloud_storage_http_request` method that can be used for PUTs and GETs based on a caller-specified `method` argument.

mlflow/azure/client.py
```diff
@@ -0,0 +1,81 @@
+from copy import deepcopy
```
dbczumar, line 1: Can we add a small docstring to this module explaining that it provides utilities for performing Azure Blob Storage operations without requiring the heavyweight `azure-storage-blob` library dependency?

mlflow/azure/client.py
```diff
@@ -0,0 +1,81 @@
...
+from six.moves import urllib
```
harupy, line 2:
```suggestion
import urllib
```

Should we fix the import statement here since we've excluded `six` from MLflow's dependencies?

WeichenXu123:
Create a test notebook on azure shard:
https://adb-7064161269814046.2.staging.azuredatabricks.net/?o=7064161269814046#notebook/4191897345698799/command/4191897345699466

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -36,6 +36,8 @@
 MOCK_SUBDIR = "subdir/path"
 MOCK_SUBDIR_ROOT_URI = posixpath.join(MOCK_RUN_ROOT_URI, MOCK_SUBDIR)
 
+pytestmark = pytest.mark.skip
```
WeichenXu123, line 39: Many tests in this module get stuck... @dbczumar Could you provide some guide/clue what happened ?

mlflow/utils/rest_utils.py
```diff
@@ -169,6 +171,54 @@ def call_endpoint(host_creds, endpoint, method, json_body, response_proto):
...
+    """
+    retry_attempts = kwargs.get('retry_attempts', 5)
+    retry_strategy = Retry(
+        total=None,
+        status=retry_attempts,
+        status_forcelist=TRANSIENT_FAILURE_RESPONSE_CODES,
+        backoff_factor=1,
+    )
+    adapter = HTTPAdapter(max_retries=retry_strategy)
+    http = requests.Session()
```
WeichenXu123, line 211: @dbczumar Shall we also close Session by :
```
with requests.Session() as session:
  ...
```

tests/store/artifact/test_databricks_artifact_repo.py
dbczumar: @WeichenXu123 Following up from our offline discussion, it appears that the tests in this module are stuck because we're attempting to make an HTTP request against an invalid URL, and this request is being erroneously retried many times because we're specifying `total=None` and `connect=None` in our arguments to `urllib3.util.Retry`.

From the docs (https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html), the category of request failure encountered in the tests is a `connect`-related error. By default, `connect=None` when constructing Retry, implying that connect-related errors will be retried infinitely unless `total` is not `None`. Because we construct `Retry` with `total=None`, these requests are retried infinitely and our tests hang.

mlflow/utils/rest_utils.py
```diff
@@ -169,6 +171,54 @@ def call_endpoint(host_creds, endpoint, method, json_body, response_proto):
...
+    :args: Positional arguments to pass to `requests.Session.put/get()`
+    :kwargs: Keyword arguments to pass to `requests.Session.put/get()`
+    """
+    retry_attempts = kwargs.get('retry_attempts', 5)
+    retry_strategy = Retry(
+        total=None,
+        status=retry_attempts,
+        status_forcelist=TRANSIENT_FAILURE_RESPONSE_CODES,
+        backoff_factor=1,
+    )
```
dbczumar, line 209: To address https://github.com/mlflow/mlflow/pull/4044/files#r570841358, I think we can make the following change:

```suggestion
    retry_strategy = Retry(
        total=None,
        connect=0, # Don't retry on connect-related errors raised before a request reaches a remote server
        read=1, # Retry once for errors reading the response from a remote server,
        redirect=3, # Limit the number of redirects to avoid infinite redirect loops
        status=retry_attempts, # Retry a specified number of times for response codes indicating transient failures, 
        other = 0, # Don't retry for other, unclassified errors 
        status_forcelist=TRANSIENT_FAILURE_RESPONSE_CODES,
        backoff_factor=1,
    )
```

dbczumar (commented):
@WeichenXu123 Can we also drop the `azure-storage-blob` dependency from `setup.py` as part of this PR?

mlflow/utils/rest_utils.py
dbczumar: Sounds good to me!

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -142,52 +144,50 @@ def _extract_headers_from_credentials(self, headers):
...
+                # the bytestring for compliance with Azure Blob Storage API requests
+                block_id = base64.b64encode(uuid.uuid4().hex.encode()).decode("utf-8")
                 try:
-                    service.stage_block(block_id, chunk, headers=headers)
-                except ClientAuthenticationError:
-                    _logger.warning(
+                    put_block(credentials.signed_uri, block_id, chunk, headers=headers)
+                except requests.HTTPError as e:
+                    if e.response.status_code in [401, 403]:
+                        eprint(
```
dbczumar, line 166: I don't think `eprint` is defined / imported in this module, so this will probably throw. Can we use `_logger.info` instead?

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -142,52 +144,50 @@ def _extract_headers_from_credentials(self, headers):
...
+                        ).credentials
+                        put_block(credentials.signed_uri, block_id, chunk, headers=headers)
+                    else:
+                        raise e
+                uploading_block_list.append(block_id)
+            try:
+                put_block_list(credentials.signed_uri, uploading_block_list, headers=headers)
+            except requests.HTTPError as e:
+                if e.response.status_code in [401, 403]:
+                    eprint(
```
dbczumar, line 181: I don't think `eprint` is defined / imported in this module, so this will probably throw. Can we use `_logger.info` instead?

dev/large-requirements.txt
```diff
@@ -2,3 +2,4 @@
 matplotlib
 plotly
 tf2onnx
+azure-storage-blob>=12.0.0
```
dbczumar, line 5: @WeichenXu123 To clarify, we've added this as a test requirement because it's required for `azure_blob_artifact_repo.py` right?

dev/small-requirements.txt
```diff
@@ -8,3 +8,4 @@ pytest==3.2.1
 pytest-cov==2.6.0
 pytest-localserver==0.5.0
 moto==1.3.16
+azure-storage-blob>=12.0.0
```
dbczumar, line 11: @WeichenXu123 To clarify, we've added this as a test requirement because it's required for `azure_blob_artifact_repo.py` right?

dev/lint-requirements.txt
```diff
@@ -6,3 +6,4 @@ rstcheck==3.2
 black==19.10b0
 # Pin isort version to prevent pylint from raising an error
 isort<5.0.0
+azure-storage-blob>=12.0.0
```
dbczumar, line 9: I don't think this needs to be included in lint-requirements. Please let me know if the lint tests fail without the presence of this dependency and we can revisit this. Otherwise, let's remove it.

mlflow/azure/client.py
```diff
@@ -0,0 +1,152 @@
...
+                       more information, see
+                       https://docs.microsoft.com/en-us/rest/api/storageservices/put-block-list.
+    :param headers: Headers to include in the Put Block request body.
+    """
+    request_url = _append_query_parameters(sas_url, {"comp": "blocklist"})
+    data = _build_block_list_xml(block_list)
+
+    request_headers = {}
+    for name, value in headers.items():
+        if _is_valid_put_block_header(name):
```
dbczumar, line 61: I've introduced some client-side header name validation for the `put_block_list` and `put_block` operations after encountering an issue introduced by a recent Databricks MLflow change where the `x-ms-blob-type` header is returned as part of the SAS URI credentials from the `GetCredentialsForWrite` operation; the current client logic in this PR passes this `x-ms-blob-type` header to `put_block_list` operation, but `x-ms-blob-type` isn't supported for Put Block List, which causes the request to fail. As a result, uploads on Azure Databricks throw exceptions when this client is run with the latest MLflow service code.

@WeichenXu123 can we add some unit tests for `mlflow/azure/client.py` to make sure that unsupported headers are removed prior to invocation of `cloud_storage_http_request`?

mlflow/azure/client.py
```diff
@@ -0,0 +1,152 @@
...
+    :param block_id: A base64-encoded string identifying the block.
+    :param data: Data to include in the Put Block request body.
+    :param headers: Additional headers to include in the Put Block request body
+                    (the `x-ms-blob-type` header is always included automatically).
+    """
+    request_url = _append_query_parameters(sas_url, {"comp": "block", "blockid": block_id})
+
+    request_headers = deepcopy(_PUT_BLOCK_HEADERS)
+    for name, value in headers.items():
+        if _is_valid_put_block_header(name):
```
dbczumar, line 33: I've introduced some client-side header name validation for the `put_block_list` and `put_block` operations after encountering an issue introduced by a recent Databricks MLflow change where the `x-ms-blob-type` header is returned as part of the SAS URI credentials from the `GetCredentialsForWrite` operation; the current client logic in this PR passes this `x-ms-blob-type` header to `put_block_list` operation (https://github.com/mlflow/mlflow/blob/125747259f3d85ad83ac838db87f1d2b5c715659/mlflow/store/artifact/databricks_artifact_repo.py#L188), but `x-ms-blob-type` isn't supported for Put Block List, which causes the request to fail. As a result, uploads on Azure Databricks throw exceptions when this client is run with the latest MLflow service code.

@WeichenXu123 can you add some unit tests for `mlflow/azure/client.py` to make sure that unsupported headers are removed prior to invocation of `cloud_storage_http_request`?

dev/small-requirements.txt
WeichenXu123: Yes!

dev/lint-requirements.txt
WeichenXu123: No. Required. Because lint will check tests code. When lint checking, it need import required packages, so includes azure storage blob package.

harupy: @WeichenXu123  In the lint check, we install `small-requirements.txt` and `large-requirements.txt` too.

https://github.com/mlflow/mlflow/blob/ebd4e5ed2e12055d5b1d3ff3c740fd838301f616/.github/workflows/master.yml#L17-L27

dbczumar (approved):
LGTM! Thanks @WeichenXu123!

PR #4050, opened 2021-02-03
Don't log autolog success message if disable=True

apurva-koti:

If `mlflow.autolog(disable=True)` is followed by `import supported_module`, we output a log message `"Autologging successfully enabled for supported_module"`.
This is because the log is called after the autologging params are processed, so the message will display regardless of what the params were.

Add a conditional to disable this log if autologging is disabled.

Test script:
```
import mlflow
mlflow.autolog(disable=True)
import tensorflow
```

Before:
image

After:
image

## How is this patch tested?

New unit tests

Components 

Interface 

Language 

Integrations


harupy:
Looks like something is wrong with spark. Tests related to spark are constantly failing. I'm investigating.

tests/tracking/fluent/test_fluent_autolog_logger.py
```diff
@@ -0,0 +1,11 @@
...
+from unittest import mock
+
+
+def test_autolog_success_message_suppressed_when_disabled():
+    with mock.patch("mlflow.tracking.fluent._logger.info") as autolog_logger_mock:
+        mlflow.autolog(disable=True)
+        # pylint: disable=unused-variable
+        import tensorflow
+
+        autolog_logger_mock.assert_not_called()
```
apurva-koti, line 11: Had a lot of trouble adding this to `test_fluent_autolog.py` because `import tensorflow` was already at the top of the file, so the import hooks wouldn't trigger on re-import. Would like some pointers on working around that if any

dbczumar: test_fluent_autolog.py has utilities for this. We should move this test case to that file and use the utilities present there, instead of creating a new test suite

apurva-koti: Helped offline + done

mlflow/tracking/fluent.py
```diff
@@ -1341,7 +1341,8 @@ def setup_autologging(module):
             autolog_fn = LIBRARY_TO_AUTOLOG_FN[module.__name__]
             autologging_params = get_autologging_params(autolog_fn)
             autolog_fn(**autologging_params)
-            _logger.info("Autologging successfully enabled for %s.", module.__name__)
+            if "disable" in autologging_params and not autologging_params["disable"]:
```
dbczumar, line 1344: Nit (also covers the case where `disable` is not present in `autologging_params`):

```suggestion
            if not autologging_params.get("disable", False):
```

apurva-koti: Oh that's better, thanks

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -192,3 +192,10 @@ def log_autolog_called(self, integration, call_args, call_kwargs):
     call = universal_autolog_event_logging_calls[0]
     assert call.integration == "mlflow"
     assert {"disable": True, "exclusive": True}.items() <= call.call_kwargs.items()
+
+
+def test_autolog_success_message_suppressed_when_disabled():
+    with mock.patch("mlflow.tracking.fluent._logger.info") as autolog_logger_mock:
```
dbczumar, line 198: Can we extend this test case to test both the enable & disable path? I'd like to make sure that we're mocking the correct logger here (otherwise, this test case will pass vacuously).

apurva-koti: Yup! I did manually verify that the test fails without the change. But good to make it explicit

dbczumar: Looks gr9!

dbczumar (approved):
LGTM! Can't wait to merge this. Note that we still expect R tests to fail due to an unrelated issue addressed via https://github.com/mlflow/mlflow/pull/4052

Issue #3950, opened 2021-01-06
[FR] Experiment Tracking Decorator to remove boilerplate


Chris-hughes10:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [ ] Yes. I can contribute this feature independently.
- [x] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

With the current tracking API, some boilerplate code is required to create an experiment and start a run. Specifically, it is quite common to see code blocks such as:
```
mlflow.set_tracking_uri(MLFLOW_TRACKING_SERVICE_URI)
experiment_id = mlflow.set_experiment(EXPERIMENT_NAME)
with mlflow.start_run(experiment_id=experiment_id):
    ...
``` 
at the start of a training script.

In an effort to remove this when using mlflow for personal projects, I implemented a decorator that can be added to a training function, which can set some of these parameters using decorator arguments or environment variables. The syntax currently looks like:
```
@mlflow_experiment
def train_model_1():
   ...

@mlflow_experiment(experiment_name='Train Random Forest', autolog=True)
def train_sklearn_model():
    ...

This has worked well for me and has resulted in my training scripts requiring less mlflow specific code - especially when combined with autologging - and eases adoption. I have primarily used this in scripts that are deployed on platforms such as AzureML, Kubeflow and Argo as opposed to in a notebook environment. This approach has received positive feedback from my colleagues at Microsoft, and I believe this could be a useful feature for other members of the community, and as a single function which uses existing components, it would be simple to integrate.

## Motivation
- What is the use case for this feature?
To remove boilerplate code from training scripts when using the tracking API for simple use cases.
- Why is this use case valuable to support for MLflow users in general?
Less code is required to start using mlflow, and it arguably provides a cleaner approach than a context manager.
- Why is this use case valuable to support for your project(s) or organization?
[...]

PR #4065, opened 2021-02-04
Add mlflow_run convenience decorator to fluent API

Chris-hughes10:
Signed-off-by: Chris Hughes <31883449+Chris-hughes10@users.noreply.github.com>

Added a convenience decorator to the fluent API which can be used in place of `mlflow.start_run` by wrapping the decorated function in an `ActiveRun` object. The decorator can be used with or without arguments, where the signature and defaults are consistent with `mlflow.start_run`.

This was discussed in : https://github.com/mlflow/mlflow/issues/3950

## How is this patch tested?
Tests have been added to ` tests/tracking/fluent/test_fluent.py` which test the decorator use with and without arguments.

Added a convenience decorator to the fluent API which can be used in place of `mlflow.start_run` by wrapping the decorated function in an `ActiveRun` object. The decorator can be used with or without arguments, where the signature and defaults are consistent with `mlflow.start_run`.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -104,7 +104,10 @@ def start_run(run_id=None, experiment_id=None, run_name=None, nested=False, tags
     will be logged. The return value can be used as a context manager within a ``with`` block;
     otherwise, you must call ``end_run()`` to terminate the current run.
 
-    If you pass a ``run_id`` or the ``MLFLOW_RUN_ID`` environment variable is set,
+    If you pass a ``run_id`` or the ``
+    
+    
+    _ID`` environment variable is set,
     ``start_run`` attempts to resume a run with the specified run ID and
     other parameters are ignored. ``run_id`` takes precedence over ``MLFLOW_RUN_ID``.
 
@@ -306,10 +309,7 @@ def mlflow_run(
 
         return run_wrapper
 
-    if _func is None:
-        return run_decorator
-    else:
-        return run_decorator(_func)
+    return run_decorator if _func is None else run_decorator(_func)
 
 
 def active_run():
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -5,6 +5,7 @@ MLflow run. This module is exposed to users at the top-level :py:mod:`mlflow` mo
 import os
 
 import atexit
+import functools
 import time
 import logging
 import inspect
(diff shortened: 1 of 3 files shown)


mlflow/tracking/fluent.py
```diff
@@ -266,6 +267,51 @@ def end_run(status=RunStatus.to_string(RunStatus.FINISHED)):
...
+                run_name=run_name,
+                nested=nested,
+                tags=tags,
+            ):
+                value = func(*args, **kwargs)
+            return value
+
+        return run_wrapper
+
+    if _func is None:
```
eedeleon, line 309: This seems like a nice use case for the ternary operator

return run_decorator if _func is None else run_decorator(_func)

mlflow/tracking/fluent.py
```diff
@@ -266,6 +267,51 @@ def end_run(status=RunStatus.to_string(RunStatus.FINISHED)):
 atexit.register(end_run)
 
 
+def mlflow_run(
+    _func=None, *, run_id=None, experiment_id=None, run_name=None, nested=False, tags=None,
```
eedeleon, line 271: I am assuming start_run handles it but would decorator use + function nesting work as intended if nested=True?

That would be pretty promising.

Should we pass kwargs through in case the spec of start_run expands or not?

eedeleon: It may be worth having def mlflow_run have an _mlflow_run_kwargs that is then passsed through to the start_run call.

That way, if start_run expands its spec, users can leverage the newer functionality without a new SDK release

mlflow/tracking/fluent.py
Chris-hughes10: Good shout, I will update that

mlflow/tracking/fluent.py
Chris-hughes10: > I am assuming start_run handles it but would decorator use + function nesting work as intended if nested=True?
> 
> That would be pretty promising.

You are correct, that is handled by start run. So using nested decorators is just equivalent to using nested context managers.

Chris-hughes10: > It may be worth having def mlflow_run have an _mlflow_run_kwargs that is then passsed through to the start_run call.
> 
> That way, if start_run expands its spec, users can leverage the newer functionality without a new SDK release

Here, are you suggesting replacing the keyword arguments with mlflow_run_kwargs or adding it additionally? The reason that I went for the keyword argument option was for ease of code completion and inline docs usage, but this does come with the drawback that the decorator signature would have to be updated to match `start_run`. I am open to debate on this.

stephanecollot:
Hi, any news on this?

Chris-hughes10:
> Hi, any news on this?

The implementation is done and in the PR, the failing test was caused by a different issue outside of this PR. Still waiting for review if there is interest.

Chris-hughes10:
@dbczumar is there still any interest on this?

Issue #4055, opened 2021-02-03
[BUG] infer_signature fails when dataframe contains ExtensionDtype


caleboverman:
### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [x] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: No
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS 11.1
- **MLflow installed from (source or binary)**: 
- **MLflow version (run ``mlflow --version``)**: 1.13.1
- **Python version**: 3.8.5
- **npm version, if running the dev UI**:
- **Exact command to reproduce**: 

### Describe the problem
Pandas supports `ExtensionDtypes` for extending types from numpy. These types allow additional functionality (like `NA` support for bools or ints). However, calling `infer_signature` on a dataframe with these types fails because the columns are not of type `numpy.ndarray`.

### Code to reproduce issue
```
import pandas as pd
from mlflow.models import infer_signature

df = pd.DataFrame(
    {
        "ints": [1, 2, 3],
        "bools": [True, False, True],
        "strings": ["a", "b", "c"]
    }
)

# doing any of the following will cause infer_signature to fail
df["ints"] = df["ints"].astype("Int64")
infer_signature(df)

df["bools"] = df["bools"].astype("boolean")
infer_signature(df)

df["strings"] = df["strings"].astype("string")
infer_signature(df)
```

### Other info / logs
```
TypeError                                 Traceback (most recent call last)
<ipython-input-122-64641ea9310a> in <module>
      1 df["strings"] = df["strings"].astype("string")
----> 2 infer_signature(df)

~/opt/anaconda3/envs/fts/lib/python3.8/site-packages/mlflow/models/signature.py in infer_signature(model_input, model_output)
    121     :return: ModelSignature
    122     """
--> 123     inputs = _infer_schema(model_input)
    124     outputs = _infer_schema(model_output) if model_output is not None else None
    125     return ModelSignature(inputs, outputs)

~/opt/anaconda3/envs/fts/lib/python3.8/site-packages/mlflow/types/utils.py in _infer_schema(data)
[...]

PR #4069, opened 2021-02-05
#4055 allow infer_signature to read pandas ExtensionDtype

caleboverman:

Allows `infer_signature` to handle dataframes containing ExtensionDtype, addresses #4055 

## How is this patch tested?

Modified `tests/models/test_model_input_examples.py` to contain additional datatypes.

`infer_signature` no longer fails when given a pandas dataframe containing an `ExtensionDtype`

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/types/utils.py
+++ b/mlflow/types/utils.py
@@ -139,7 +139,22 @@ def _infer_pandas_column(col: pd.Series) -> DataType:
     if not isinstance(col, pd.Series):
         raise TypeError("Expected pandas.Series, got '{}'.".format(type(col)))
 
-    if pd.api.types.infer_dtype(col) == "bytes":
+    class IsInstanceOrNone(object):
+        def __init__(self, *args):
+            self.classes = args
+            self.seen_instances = 0
+
+        def __call__(self, x):
+            if x is None:
+                return True
+            elif any(map(lambda c: isinstance(x, c), self.classes)):
+                self.seen_instances += 1
+                return True
+            else:
+                return False
+
+    is_binary_test = IsInstanceOrNone(bytes, bytearray)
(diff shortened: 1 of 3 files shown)


caleboverman:
@tomasatdatabricks looks like the previous tests were expecting `infer_signature` to set a binary column type to string in the signature. Not sure if that was done for a reason and if we should set the signature type to binary now.

tomasatdatabricks:
@caleboverman It's actually the other way around, the tests asserts that binary is inferred as binary but it was inferred as string in your case. It's because pd.api.types.infer_dtype(col) returns "mixed" instead of "binary". I used bytearray instead of bytes in the test. I think it should work for both - i.e. both bytearray and bytes should be inferred as binary. Can you maybe do the same check we do fro numpy array objects?

Otherwise the PR looks great, thanks for the contribution!

mlflow/types/utils.py
```diff
@@ -135,6 +135,45 @@ def _infer_numpy_dtype(dtype: np.dtype) -> DataType:
...
+            if x is None:
+                return True
+            elif any(map(lambda c: isinstance(x, c), self.classes)):
+                self.seen_instances += 1
+                return True
+            else:
+                return False
+
+    is_binary_test = IsInstanceOrNone(bytes, bytearray)
+    if all(map(is_binary_test, col)) and is_binary_test.seen_instances > 0:
```
tomasatdatabricks, line 157: since this is the most expensive test, can we put it at the end?

tomasatdatabricks (commented):
Looks good, just one small request, after that lgtm.

tomasatdatabricks (approved):
lgtm! Thanks for the contirbution @caleboverman!

I've decided to merge this as is since the last requested change was a minor and we have changes coming to schema enforcement as a result of adding tensor support.

PR #4079, opened 2021-02-08
Add multi-dimensional array (aka Tensor) support

arjundc-db:

This PR adds support for multi-dimensional arrays (aka Tensor) for pyfunc.predict, model signature and model input example.

## How is this patch tested?

Unit tests have been added, and manual testing was performed.

`ModelSignature` now supports tensors as inputs. Few member functions of `Schema` have been marked as deprecated.
`_Example` now supports tensors as inputs. User's may store a sample tensor inputs as model metadata.

Components 

Interface 

Language 

Integrations


mlflow/types/schema.py
```diff
@@ -116,84 +116,189 @@ def __repr__(self) -> str:
...
+        if self.is_tensor_spec():
+            return [x.type for x in self.data_rep]
+        return [x.type.to_numpy() for x in self.data_rep]
 
     def pandas_types(self) -> List[np.dtype]:
-        """ Convenience shortcut to get the datatypes as pandas types."""
-        return [x.type.to_pandas() for x in self.columns]
+        """ Convenience shortcut to get the datatypes as pandas types. Unsupported by TensorSpec."""
+        if self.is_tensor_spec():
+            raise MlflowException("Datatype conversion is not supported by TensorSpec")
```
tomasatdatabricks, line 259: We should unify the error message with column_types, e.g. `raise MlflowException("TensorSpec only supports numpy types, use numpy_types() instead")`

mlflow/types/schema.py
```diff
@@ -116,84 +116,189 @@ def __repr__(self) -> str:
...
     def as_spark_schema(self):
         """Convert to Spark schema. If this schema is a single unnamed column, it is converted
         directly the corresponding spark data type, otherwise it's returned as a struct (missing
         column names are filled with an integer sequence).
+        Unsupported by TensorSpec.
         """
-        if len(self.columns) == 1 and self.columns[0].name is None:
-            return self.columns[0].type.to_spark()
+        if self.is_tensor_spec():
+            raise MlflowException("Datatype conversion is not supported by TensorSpec")
```
tomasatdatabricks, line 269: Lets say tensor spec can not be converted to spark dataframe.

mlflow/types/utils.py
```diff
@@ -17,45 +17,60 @@ def __init__(self, msg):
...
+    including a variable dimension. Assumes the variable dimension is the first dimension.
+
+    NB: For 1D tensors, the output is (-1, shape).
+
+    :param data: Dataset to infer from.
+    :return: tuple : Shape of the inputted data (including a variable dimension)
+    """
+    if not isinstance(data, np.ndarray):
+        raise TypeError("Expected numpy.ndarray, got '{}'.".format(type(data)))
+    if len(data.shape) == 1:
```
tomasatdatabricks, line 34: hmm I don't think we need this branch. If the input is a 1-d np array we should just return [-1]. The else branch takes care of that.

mlflow/types/utils.py
```diff
@@ -17,45 +17,60 @@ def __init__(self, msg):
...
 
 
+def _get_tensor_shape(data: np.ndarray) -> tuple:
+    """
+    Infer the shape of the inputted data.
+
+    This method creates the shape of the tensor to store in the TensorSpec. Always returns a shape
+    including a variable dimension. Assumes the variable dimension is the first dimension.
+
+    NB: For 1D tensors, the output is (-1, shape).
```
tomasatdatabricks, line 27: I think 1d tensors should just return [-1]

mlflow/types/utils.py
```diff
@@ -17,45 +17,60 @@ def __init__(self, msg):
         )
 
 
+def _get_tensor_shape(data: np.ndarray) -> tuple:
```
tomasatdatabricks, line 20: should we add batch_dimension=0 argument?

mlflow/types/utils.py
```diff
@@ -17,45 +17,60 @@ def __init__(self, msg):
...
+    This method will raise an exception if the user data contains incompatible types or is not
+    passed in one of the supported formats (containers).
 
     The input should be one of these:
       - pandas.DataFrame or pandas.Series
       - dictionary of { name -> numpy.ndarray}
       - numpy.ndarray
       - pyspark.sql.DataFrame
 
-    The element types should be mappable to one of :py:class:`mlflow.models.signature.DataType`.
```
tomasatdatabricks, line 36: Lets leave this message in, just say mappable to one of the DataType for dataframes and one of the numpy types for tensors.

mlflow/types/schema.py
```diff
@@ -116,84 +116,189 @@ def __repr__(self) -> str:
...
+    def from_json_dict(cls, **kwargs):
+        """
+        Deserialize from a json loaded dictionary.
+        The dictionary is expected to contain `type` and `shape` keys.
+        """
+        tensor_type = np.dtype(kwargs["type"])
+        tensor_shape = tuple(kwargs["shape"])
+        return cls(tensor_type, tensor_shape, kwargs["name"] if "name" in kwargs else None)
+
+    def __eq__(self, other) -> bool:
```
tomasatdatabricks, line 180: we should check if other is TensorSpec. just in case someone compares schema with colspec with schema with tensorspec

tests/models/test_signature.py
```diff
@@ -38,6 +38,71 @@ def test_model_signature():
     assert signature5 == signature6
 
 
+def test_model_signature_with_tensorspec():
+    signature1 = ModelSignature(
```
tomasatdatabricks, line 42: Can we also test comparison of DataframeSchema with TensorSchema?

tomasatdatabricks (commented):
looks good, I've left some comments.

mlflow/models/utils.py
```diff
@@ -97,18 +93,27 @@ def _is_scalar(x):
...
-        del self.data["index"]
-        if all(input_example.columns == range(len(input_example.columns))):
-            # No need to write default column index out
-            del self.data["columns"]
-        self.info = {
-            "artifact_path": example_filename,
-            "type": "dataframe",
-            "pandas_orient": "split",
-        }
+        if isinstance(input_example, pd.DataFrame):
```
tomasatdatabricks, line 98: I think we should do it the other way - we should check if the input is tensors (i.e. a numpy array or a dict of numpy arrays) otherwise proceed to store dataframe. Note that we do support e.g. dict of scalars, or a list as valid examples. This is so that uses can easily handcraft examples.

mlflow/models/utils.py
```diff
@@ -97,18 +93,27 @@ def _is_scalar(x):
...
+            if all(input_example.columns == range(len(input_example.columns))):
+                # No need to write default column index out
+                del self.data["columns"]
+            self.info = {
+                "artifact_path": example_filename,
+                "type": "dataframe",
+                "pandas_orient": "split",
+            }
+        else:
+            self.data = {"instances": input_example}
```
tomasatdatabricks, line 111: you probably want to call input_example.tolist().

We also want to handle dictionary of tensors, the easies would be to use the `instances`  columnar representation, so maybe lets use instances for both numpy array and dict of numpy arrays?

mlflow/models/utils.py
```diff
@@ -97,18 +93,27 @@ def _is_scalar(x):
...
+            self.info = {
+                "artifact_path": example_filename,
+                "type": "ndarray",
+                "format": "instances"
```
tomasatdatabricks: should we call the format tf-serving?

mlflow/models/utils.py
```diff
@@ -146,10 +151,25 @@ def _read_example(mlflow_model: Model, path: str):
...
     input_schema = mlflow_model.signature.inputs if mlflow_model.signature is not None else None
     path = os.path.join(path, mlflow_model.saved_input_example_info["artifact_path"])
-    return _dataframe_from_json(path, schema=input_schema, precise_float=True)
+    if (example_type == "ndarray") or (input_schema and input_schema.is_tensor_spec()):
+        return _read_tensor_input_from_json(path)
+    else:
+        return _dataframe_from_json(path, schema=input_schema, precise_float=True)
+
+
+def _read_tensor_input_from_json(path_or_str):
```
tomasatdatabricks, line 166: it's not really path_or_str in this case, it can only be path right?

mlflow/models/utils.py
```diff
@@ -146,10 +151,25 @@ def _read_example(mlflow_model: Model, path: str):
...
+    import mlflow.pyfunc.scoring_server as pyfunc_scoring_server
+    with open(path_or_str, "r") as handle:
+        try:
+            inp_dict = json.load(handle)
+        except Exception:
+            raise MlflowException(
+                "Error when loading file containing input example stored at {0}".format(path_or_str)
+            )
+        return pyfunc_scoring_server.parse_tf_serving_input(inp_dict)
```
tomasatdatabricks: we should not be using scoring_server directly, we should extract the functionality into proto_json_utils.py files, similar to `_dataframe_from_json`

mlflow/types/utils.py
```diff
@@ -17,25 +17,20 @@ def __init__(self, msg):
...
     """
     if not isinstance(data, np.ndarray):
         raise TypeError("Expected numpy.ndarray, got '{}'.".format(type(data)))
-    if len(data.shape) == 1:
-        variable_input_data_shape = [-1, data.shape[0]]
-    else:
-        variable_input_data_shape = list(data.shape)
-        variable_input_data_shape[0] = -1
+    variable_input_data_shape = list(data.shape)
+    variable_input_data_shape[batch_dimension] = -1
```
tomasatdatabricks, line 33: should we give an option for batch dimension to be None signaling there is no batch dimension (i.e. all dimensions would be fixed)?

tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
```diff
@@ -339,49 +339,50 @@ def predict(pdf):
...
-
-    # Can only provide data type that can be converted to dataframe...
-    with pytest.raises(MlflowException) as ex:
-        pyfunc_model.predict(set([1, 2, 3]))
-    assert "Expected input to be DataFrame or list. Found: set" in str(ex)
-
-    # 9. dictionaries of str -> list/nparray work
-    d = {"a": [1.0], "b": [2.0], "c": [3.0]}
-    assert pyfunc_model.predict(d).equals(pd.DataFrame(d))
+# TODO : @wendyhu/@arjundc fix this test in enforcement PR
```
tomasatdatabricks, line 342: Does this fail?

mlflow/types/schema.py
```diff
@@ -116,84 +116,189 @@ def __repr__(self) -> str:
...
 
+class TensorSpec(object):
+    """
+    Specification used to represent a dataset stored as a Tensor.
+    """
+
+    def __init__(
+        self,
+        type: np.dtype,  # pylint: disable=redefined-builtin
+        shape: tuple,
```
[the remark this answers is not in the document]
arjundc-db: Added supported for inputting list. 
I am not aware that we can specify tuple[int] in python.

mlflow/types/utils.py
arjundc-db: This is really cool

tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
arjundc-db: my bad

mlflow/types/schema.py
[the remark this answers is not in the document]
tomasatdatabricks: kk, thanks! I was not sure about tuple[int], I guess it does not work.

mlflow/models/utils.py
```diff
@@ -97,18 +94,31 @@ def _is_scalar(x):
...
-            del self.data["columns"]
-        self.info = {
-            "artifact_path": example_filename,
-            "type": "dataframe",
-            "pandas_orient": "split",
-        }
+        if isinstance(input_example, (np.ndarray, dict)):
+            if isinstance(input_example, np.ndarray):
+                input_example = input_example.tolist()
+                self.data = {"instances": input_example}
```
tomasatdatabricks, line 102: should we use inputs in for both numpy arrays and dictionaries for consistency? I don't think it makes a difference in case of a single tensor, but it may be easier for users to read the json if we always stick to one format.

arjundc-db: This might get a little weird since the `inputs` format always expects a dictionary. https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/scoring_server/__init__.py#L200
Thus for a single tensor, we will have to store it as a dict. Again while reading the json, read the dict and return the single example (otherwise it wont work as-is).

Plus, we have the utility to load the tf-serving format which can be used.

mlflow/models/utils.py
```diff
@@ -146,10 +156,24 @@ def _read_example(mlflow_model: Model, path: str):
...
     example_type = mlflow_model.saved_input_example_info["type"]
-    if example_type != "dataframe":
+    if example_type not in ["dataframe", "ndarray"]:
         raise MlflowException(
             "This version of mlflow can not load example of type {}".format(example_type)
         )
     input_schema = mlflow_model.signature.inputs if mlflow_model.signature is not None else None
     path = os.path.join(path, mlflow_model.saved_input_example_info["artifact_path"])
-    return _dataframe_from_json(path, schema=input_schema, precise_float=True)
+    if (example_type == "ndarray") or (input_schema and input_schema.is_tensor_spec()):
```
tomasatdatabricks, line 165: I think the example type must be "ndarray", otherwhise we can not read it with `_read_tensor_input_from_json`

mlflow/models/utils.py
```diff
@@ -146,10 +156,24 @@ def _read_example(mlflow_model: Model, path: str):
...
+        return _read_tensor_input_from_json(path)
+    else:
+        return _dataframe_from_json(path, schema=input_schema, precise_float=True)
+
+
+def _read_tensor_input_from_json(path):
+    with open(path, "r") as handle:
+        try:
+            inp_dict = json.load(handle)
+        except Exception:
```
tomasatdatabricks, line 175: Not sure if we need to catch&rethrow, it may be better to just throw the underlying error.

mlflow/types/schema.py
```diff
@@ -116,84 +118,191 @@ def __repr__(self) -> str:
...
+                "Creating Schema with a combination of {0} and {1} is not supported. "
+                "Please choose one of {0} or {1}".format(ColSpec.__class__, TensorSpec.__class__)
+            )
+        self._data_rep = dataset
 
     @property
-    def columns(self) -> List[ColSpec]:
-        """The list of columns that defines this schema."""
-        return self._cols
+    def data_rep(self) -> List[ColSpec]:
```
tomasatdatabricks, line 232: the return type should be a List[Union[ColSpec,TensorSpec]] right? Or does TensorSpec inherit from ColSpec?

tomasatdatabricks, line 232: I am not sure if data_rep is descriptive name, what do you think about calling this `inputs` instead?

mlflow/types/schema.py
```diff
@@ -116,84 +118,191 @@ def __repr__(self) -> str:
...
+        """Representation of a dataset that defines this schema."""
+        return self._data_rep
+
+    def is_tensor_spec(self) -> bool:
+        """Return true iff this schema is specified using TensorSpec"""
+        return self.data_rep and isinstance(self.data_rep[0], TensorSpec)
 
     def column_names(self) -> List[Union[str, int]]:
```
tomasatdatabricks, line 240: I think it's a bit confusing that `columns` sometimes refers to dataframe columns and sometimes can be either tensor or dataframe column. I wonder if we should make all the `column_***` methods valid for DataFrame schema only and introduce new ones that work for both, e.g. we call `data_rep` `inputs`, we can introduce input_names. What do you think?

mlflow/types/utils.py
```diff
@@ -17,45 +17,71 @@ def __init__(self, msg):
...
-    This method captures the column names and data types from the user data. The signature
-    represents model input and output as data frames with (optionally) named columns and data
-    type specified as one of types defined in :py:class:`DataType`. This method will raise
-    an exception if the user data contains incompatible types or is not passed in one of the
-    supported formats (containers).
+    Data inputted as a numpy array or a dictionary is represented by :py:class:`TensorSpec`.
+    All other inputted data types are specified by :py:class:`ColSpec`.
+
+    A `TensorSpec` captures the data shape (default variable axis is 0), the data type (numpy.dtype)
+    and an optional name to represent the dataset.
```
tomasatdatabricks, line 58: nit: I think we should say `an optional name for each individual tensor of the dataset` to mimic what we say for ColSpec.

tests/models/test_model_input_examples.py
```diff
@@ -55,40 +65,53 @@ def test_input_examples(pandas_df_with_all_types):
...
+            filename = example.info["artifact_path"]
+            with open(tmp.path(filename), "r") as f:
+                data = json.load(f)
+                assert set(data.keys()) == set(("instances",))
+            parsed_ary = _read_tensor_input_from_json(tmp.path(filename))
+            assert np.array_equal(parsed_ary, input_example)
+
+    # pass multidimensional array as a list
+    example = np.array([[1, 2, 3]])
+    with pytest.raises(TensorsNotSupportedException):
```
tomasatdatabricks, line 113: hopefully this is failing now? :)

tomasatdatabricks (commented):
Left some more comments, but we're getting close!

mlflow/models/utils.py
```diff
@@ -97,18 +94,31 @@ def _is_scalar(x):
...
-            "artifact_path": example_filename,
-            "type": "dataframe",
-            "pandas_orient": "split",
-        }
+        if isinstance(input_example, (np.ndarray, dict)):
+            if isinstance(input_example, np.ndarray):
+                input_example = input_example.tolist()
+                self.data = {"instances": input_example}
+            else:
+                self.data = {"inputs": input_example}
```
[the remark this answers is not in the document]
arjundc-db: Yep, this was done here : https://github.com/mlflow/mlflow/pull/4079/files/c0ee7933e4c62666a41506a3c1f1c8d49d94644d#diff-395a8aeaf3442dfb27e45ec1a038b71c76c7b8ad42744ee67587c8314acfee04R67

mlflow/types/schema.py
arjundc-db: Really hard to think of a simple name that gets the job done, but guess `inputs` comes closest.
Other options : `specification` or `dataset`. Using inputs for now.

wentinghu: for type annotation, this should return `List[SchemaInput]` right?

mlflow/models/utils.py
```diff
@@ -54,61 +56,78 @@ def __init__(self, input_example: ModelInputExample):
...
+                    input_ex = pd.DataFrame(input_ex)
+            elif not isinstance(input_ex, pd.DataFrame):
+                try:
+                    import pyspark.sql.dataframe
+
+                    if isinstance(input_example, pyspark.sql.dataframe.DataFrame):
+                        raise MlflowException(
+                            "Examples can not be provided as Spark Dataframe. "
+                            "Please make sure your example is of a small size and "
+                            "turn it into a pandas DataFrame or a tensor."
```
tomasatdatabricks, line 99: nit: I think we can drop `or a tensor` from the message, spark dataframes naturally map to pandas DataFrames and tensor may be confusing.

mlflow/models/utils.py
```diff
@@ -122,9 +141,13 @@ def _save_example(mlflow_model: Model, input_example: ModelInputExample, path: s
 
     The metadata is a dictionary with the following fields:
       - 'artifact_path': example path relative to the model directory.
-      - 'type': Type of example. Currently the only supported value is 'dataframe'
-      - 'pandas_orient': Determines the json encoding for dataframe examples in terms of pandas
-                         orient convention. Defaults to 'split'.
+      - 'type': Type of example. Currently the supported values are 'dataframe' and 'ndarray'
+      -  This field depends on the type of data being stored
```
tomasatdatabricks, line 145: Nit: Should we say `one of` instead of `This field depends ...`? One of is used in protobuf.

mlflow/models/utils.py
```diff
@@ -146,10 +169,19 @@ def _read_example(mlflow_model: Model, path: str):
...
-    if example_type != "dataframe":
+    if example_type not in ["dataframe", "ndarray"]:
         raise MlflowException(
             "This version of mlflow can not load example of type {}".format(example_type)
         )
     input_schema = mlflow_model.signature.inputs if mlflow_model.signature is not None else None
     path = os.path.join(path, mlflow_model.saved_input_example_info["artifact_path"])
-    return _dataframe_from_json(path, schema=input_schema, precise_float=True)
+    if example_type == "ndarray":
+        return _read_tensor_input_from_json(path)
```
tomasatdatabricks, line 179: We should also enable using tensor schema in json parsing (used in model serving as well!) so we get the data types right, but that's for a follow-up.

mlflow/types/schema.py
```diff
@@ -116,84 +118,189 @@ def __repr__(self) -> str:
...
-    with unique non empty name for every column, or unnamed with implicit integer index defined by
-    their list indices. Combination of named and unnamed columns is not allowed.
+    The dataset represented by a schema can be named, with unique non empty names for every input.
+    In the case of :py:class:`ColSpec`, the dataset columns can be unnamed with implicit integer
+    index defined by their list indices.
+    Combination of named and unnamed data inputs are not allowed.
     """
 
-    def __init__(self, cols: List[ColSpec]):
+    def __init__(self, dataset: List[Union[ColSpec, TensorSpec]]):
```
tomasatdatabricks, line 210: nit: `dataset` -> `inputs` ? I think that is more clear since you access it as `inputs` after.

mlflow/types/schema.py
```diff
@@ -116,84 +118,189 @@ def __repr__(self) -> str:
...
+            or all(map(lambda x: isinstance(x, ColSpec), dataset))
+        ):
+            raise MlflowException(
+                "Creating Schema with a combination of {0} and {1} is not supported. "
+                "Please choose one of {0} or {1}".format(ColSpec.__class__, TensorSpec.__class__)
+            )
+        self._inputs = dataset
 
     @property
-    def columns(self) -> List[ColSpec]:
```
tomasatdatabricks, line 139: I think we need to keep all the `columns*` methods for backwards compatibility. We should mark them deprecated thought and raise exception if the schema is a tensor schema.

mlflow/types/schema.py
```diff
@@ -116,84 +118,189 @@ def __repr__(self) -> str:
...
 
     def to_dict(self) -> List[Dict[str, Any]]:
         """Serialize into a jsonable dictionary."""
-        return [x.to_dict() for x in self.columns]
+        return [x.to_dict() for x in self.inputs]
 
     @classmethod
     def from_json(cls, json_str: str):
         """ Deserialize from a json string."""
-        return cls([ColSpec(**x) for x in json.loads(json_str)])
```
tomasatdatabricks, line 190: The json for tensors is supposed to look like this:
```
inputs: [{
  "name": "image",
  “type”: “tensor”,
  “tensor-spec“: {
    “dtype”: “float32”,
    “shape”: [28, 28, 3]
  }
}],
outputs: [{
  # probabilities:  
  “type”: “tensor”,
  “tensor-spec”: {
    “dtype”: “float32”,
    “shape”: [3] 
   }
}]
```

so we need to change this part a bit.

mlflow/types/schema.py
```diff
@@ -116,84 +119,281 @@ def __repr__(self) -> str:
...
+        try:
+            return cls([ColSpec(**x) for x in json.loads(json_str)])
```
tomasatdatabricks, line 388: I think relying on a TypeError is brittle. Sorry I should have mentioned this last time! Can we have one parsing method that will look at the type field and create either a ColSpec or TensorSpec for each input?

tests/models/test_model_input_examples.py
```diff
@@ -55,40 +65,53 @@ def test_input_examples(pandas_df_with_all_types):
             .all()
         )
 
+    # NB: Choose nparrays columns that can be encoded using proto_json_utils.pyNumpyEncoder
+    new_df = pandas_df_with_all_types[["integer", "long", "float", "double"]]
```
tomasatdatabricks, line 69: saving tensor example does not work with other types? I guess we do not expect the pandas extension types to work, but what about boolean, string and binary?

tests/models/test_model_input_examples.py
```diff
@@ -55,40 +65,53 @@ def test_input_examples(pandas_df_with_all_types):
...
-            _Example({"x": example, "y": example})
+    for col in dict_of_ndarrays:
+        input_example = dict_of_ndarrays[col]
+        with TempDir() as tmp:
+            example = _Example(input_example)
+            example.save(tmp.path())
+            filename = example.info["artifact_path"]
+            with open(tmp.path(filename), "r") as f:
+                data = json.load(f)
+                assert set(data.keys()) == set(("instances",))
```
tomasatdatabricks, line 107: nit: I think we should drop this assert. the important part is it parses back in correctly

tests/models/test_model_input_examples.py
arjundc-db: Other than binary, they work, since binary is converted to base64 encoded strings.

tomasatdatabricks (approved):
lgtm!

arjundc-db:
The example test failure is unrelated to this PR and its fix is tracked here : https://github.com/mlflow/mlflow/pull/4091

PR #4082, opened 2021-02-09
New env var to choose artifcats storage root when tracking on db

### LINKED ISSUE #4077: [BUG] No way to configure artifact store path when tracking on SQLAlchemy
## Willingness to contribute

- [X ] Yes. I can contribute this feature independently.

## Proposal Summary

By default, artifacts are stored to `./mlruns` (hard coded constant: DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH). The idea is to give the possibility to change this behaviour by setting an env var.

The same mechanism exists but not generalized:

```
_TRACKING_DIR_ENV_VAR = "MLFLOW_TRACKING_DIR"

def _default_root_dir():
    return get_env(_TRACKING_DIR_ENV_VAR) or os.path.abspath(DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH)
```

alikefia:

New env var to configure artifacts storage root when tracking to db via SQLAlchemy (#4077).
- When tracking on files, this path is defined via `MLFLOW_TRACKING_URI`
- When tracking to DB, MLFLOW_TRACKING_URI is used to setup the connection to Database and the file storage for artifacts is hard coded to `DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH=./mlruns`

## How is this patch tested?

N/A

Gives the possibility to configure the root directory where artifacts are stored when tracking on DB.

Components 
- [X] `area/artifacts`: Artifact stores and artifact logging

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/_tracking_service/utils.py
+++ b/mlflow/tracking/_tracking_service/utils.py
@@ -12,6 +12,7 @@ from mlflow.utils.file_utils import path_to_local_file_uri
 from mlflow.utils.databricks_utils import get_databricks_host_creds
 
 _TRACKING_URI_ENV_VAR = "MLFLOW_TRACKING_URI"
+_ARTIFACT_ROOT_ENV_VAR = "MLFLOW_ARTIFACT_ROOT"
 
 # Extra environment variables which take precedence for setting the basic/bearer
 # auth on http requests.
@@ -113,7 +114,8 @@ def _get_sqlalchemy_store(store_uri, artifact_uri):
     from mlflow.store.tracking.sqlalchemy_store import SqlAlchemyStore
 
     if artifact_uri is None:
-        artifact_uri = DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
+        artifact_uri = os.environ.get(_ARTIFACT_ROOT_ENV_VAR, None) or \
+            DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
     return SqlAlchemyStore(store_uri, artifact_uri)
 
 
--- a/mlflow/tracking/_tracking_service/utils.py
+++ b/mlflow/tracking/_tracking_service/utils.py
@@ -115,7 +115,7 @@ def _get_sqlalchemy_store(store_uri, artifact_uri):
 
     if artifact_uri is None:
         artifact_uri = (
-            os.environ.get(_ARTIFACT_ROOT_ENV_VAR, None) or DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
+            os.environ.get(_ARTIFACT_ROOT_ENV_VAR, DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH)
         )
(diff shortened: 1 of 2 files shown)


tomzx (commented):
Change makes sense. Can you add some tests in `tests/tracking/_tracking_service/test_utils.py`?

alikefia:
Done.
Picked the fix for list_experiments (SQLAlchemy v1.4) to run the tests locally.

mlflow/tracking/_tracking_service/utils.py
```diff
@@ -113,7 +114,9 @@ def _get_sqlalchemy_store(store_uri, artifact_uri):
     from mlflow.store.tracking.sqlalchemy_store import SqlAlchemyStore
 
     if artifact_uri is None:
-        artifact_uri = DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
+        artifact_uri = (
+            os.environ.get(_ARTIFACT_ROOT_ENV_VAR, None) or DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
```
tomzx, line 118: Is there a reason you preferred this over
```suggestion
            os.environ.get(_ARTIFACT_ROOT_ENV_VAR, DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH)
```
While I understand an empty/unset `MLFLOW_ARTIFACT_ROOT`  would be replaced by the default value, is there another reason?

From the code above (`get_tracking_uri`), it appears that if the `_TRACKING_URI_ENV_VAR` environment variable is set but empty, we would still use its empty value. Thus the two behavior are different which might not be what we want.

tests/tracking/_tracking_service/test_utils.py
```diff
@@ -155,6 +168,31 @@ def test_get_store_sqlalchemy_store_with_artifact_uri(tmp_wkdir, db_type):
     mock_create_engine.assert_called_once_with(uri, pool_pre_ping=True)
 
 
+def test_get_store_sqlalchemy_store_with_artifact_root():
```
tomzx, line 171: Would it be possible to convert the 3 tests (+ this one) that share almost the same code into a parameterized test?

alikefia: No reason indeed, totally agree

mlflow/tracking/_tracking_service/utils.py
alikefia: No reason indeed, agree !

tests/tracking/_tracking_service/test_utils.py
alikefia: Done

shachibista:
Any updates on this?

andreweskeclarke:
Agreed, any updates here? Would love to be able to set the artifacts store like I can set the mlruns.db path. In particular I am trying to add integration tests and this is unnecessarily creating an ./mlruns directory that I can't put in my `tmp_path` location.

andreweskeclarke:
@tomzx (see above comment)

PR #4084, opened 2021-02-09
Support tensor-based model signatures in schema enforcement

wentinghu:

Schema enforcement is performed against models with tensor-based signatures (`TensorSpec`s). The input's names, dtype, and shape is checked against expected values from the model signature.

## How is this patch tested?

Added unit tests

Schema enforcement is performed against models with tensor-based signatures (`TensorSpec`s). The input's names, dtype, and shape is checked against expected values from the model signature.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -472,9 +472,9 @@ def _enforce_schema(pfInput: PyFuncInput, input_schema: Schema):
                 for col_name, tensor_spec in zip(input_schema.input_names(), input_schema.inputs):
                     if not isinstance(pfInput[col_name], np.ndarray):
                         message = (
-                            "This model contains a tensor-based model signature with input names, which"
-                            " suggests a dictionary input mapping input name to a numpy array, but a dict with"
-                            " value type {0} was found."
+                            "This model contains a tensor-based model signature with input names,"
+                            " which suggests a dictionary input mapping input name to a numpy"
+                            " array, but a dict with value type {0} was found."
                         ).format(type(pfInput[col_name]))
                         raise MlflowException(message)
                     new_pfInput[col_name] = _enforce_tensor_spec(pfInput[col_name], tensor_spec)
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -216,7 +216,7 @@ from mlflow.models.utils import _save_example
 from mlflow.pyfunc.model import PythonModel, PythonModelContext  # pylint: disable=unused-import
 from mlflow.pyfunc.model import get_default_conda_env
 from mlflow.tracking.artifact_utils import _download_artifact_from_uri
-from mlflow.types import DataType, Schema
+from mlflow.types import DataType, Schema, TensorSpec
 from mlflow.utils import PYTHON_VERSION, get_major_minor_py_version
 from mlflow.utils.annotations import deprecated
 from mlflow.utils.file_utils import TempDir, _copy_file_or_tree
@@ -283,7 +283,7 @@ def _load_model_env(path):
     return _get_flavor_configuration(model_path=path, flavor_name=FLAVOR_NAME).get(ENV, None)
 
 
-def _enforce_type(name, values: pandas.Series, t: DataType):
+def _enforce_mlflow_type(name, values: pandas.Series, t: DataType):
     """
     Enforce the input column type matches the declared in model input schema.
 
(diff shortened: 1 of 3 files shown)


tomasatdatabricks (commented):
looks good, I've left some comments.

mlflow/pyfunc/__init__.py
```diff
@@ -380,52 +410,95 @@ def _enforce_schema(pdf: PyFuncInput, input_schema: Schema):
...
-        col_names = pdf.columns[: len(input_schema.columns)]
-    col_types = input_schema.column_types()
-    new_pdf = pandas.DataFrame()
-    for i, x in enumerate(col_names):
-        new_pdf[x] = _enforce_type(x, pdf[x], col_types[i])
-    return new_pdf
+
+    if input_schema.is_tensor_spec():
+        if input_schema.has_column_names():
+            if isinstance(pfInput, dict):
```
tomasatdatabricks, line 467: 1) if there is only a single input, we also want to support a numpy array and wrap it in a dictionary with the expected input name. 
2) pandas.DataFrame should be allowed: 
   if there are multiple inputs in the schema, we should treat it as a dict[column -> column.values].

mlflow/pyfunc/__init__.py
```diff
@@ -370,62 +370,149 @@ def all_ints(xs):
...
+        if isinstance(pfInput, (list, np.ndarray, dict)):
+            try:
+                pfInput = pandas.DataFrame(pfInput)
+            except Exception as e:
+                message = (
+                    "This model contains a column-based signature, which suggests a DataFrame"
+                    " input. There was an error casting the input data to a DataFrame:"
+                    " {0}".format(str(e))
+                )
+                raise MlflowException(message)
```
arjundc-db, line 426: Nit : Can make this a single line
`raise MlflowException("....")`

mlflow/pyfunc/__init__.py
```diff
@@ -370,62 +370,149 @@ def all_ints(xs):
...
+        if not isinstance(pfInput, pandas.DataFrame):
+            message = "Expected input to be DataFrame or list. Found: %s" % type(pfInput).__name__
```
arjundc-db, line 428: Same as above ^.

mlflow/pyfunc/__init__.py
```diff
@@ -370,62 +370,149 @@ def all_ints(xs):
...
 
-def _enforce_schema(pdf: PyFuncInput, input_schema: Schema):
+def _enforce_tensor_spec(values: np.ndarray, tensor_spec: TensorSpec):
     """
-    Enforce column names and types match the input schema.
+    Enforce the input tensor shape and type matches the declared in model input schema.
+    """
+    expected_shape = tensor_spec.shape
+    actual_shape = values.shape
+    if len(expected_shape) != len(actual_shape):
```
arjundc-db, line 379: Suppose the stored spec is (-1,28,28,3) and a user passes a single input of (28,28,3), I think we should allow that input.
It will fail the check here.

arjundc-db: cc @tomasatdatabricks

mlflow/pyfunc/__init__.py
arjundc-db: PS : Please do the same for all such cases.

tomasatdatabricks (approved):
lgtm!

mlflow/pyfunc/__init__.py
```diff
@@ -370,62 +370,149 @@ def all_ints(xs):
...
+        missing_cols = [c for c in input_names if c in missing_cols]
+        extra_cols = [c for c in actual_cols if c in extra_cols]
         if missing_cols:
             message = (
-                "Model input is missing columns {0}."
-                " Note that there were extra columns: {1}".format(missing_cols, extra_cols)
+                "Model is missing inputs {0}."
+                " Note that there were extra inputs: {1}".format(missing_cols, extra_cols)
             )
             raise MlflowException(message)
```
arjundc-db, line 455: After this line, I think it would help readability and modularity to branch off based on whether the the schema defines a TensorSpec/ColSpec.
Otherwise it feels the enforcement is getting interleaved. 
Something like:
`new_pyInput = _enforce_tensor_spec(pfInput) if input_schema.is_tensor_spec() else _enforce_col_spec(pfInput).`

arjundc-db (changes requested):
Going through tests, but would be great to get a few of these in.
Looks great though.

mlflow/pyfunc/__init__.py
wentinghu: hm, i think it makes more sense to always expect input to be a batch (i believe tf and other frameworks i've used in the past have this expectation too)

arjundc-db: I think that makes sense. 
What about the other way, the stored spec does not contain a batch (28, 28, 3) and the input contains a batch?

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -871,46 +871,44 @@ def test_autolog_does_not_throw_when_mlflow_logging_fails(func_to_fail):
...
-#     # [[0.171504346208176  ]
-#     #  [0.3434615044164018 ]  <- diff
-#     #  [0.06895096846585136]  <- diff
-#     #  [0.05925789882165455]
-#     #  [0.03424907823290102]]
-#     #
-#     # As a workaround, use `assert_array_almost_equal` instead of `assert_array_equal`
-#     np.testing.assert_array_almost_equal(pyfunc_model.predict(input_example),
-#     model.predict(X[:5]))
+@pytest.mark.parametrize("data_type", [pd.DataFrame, np.array])
```
arjundc-db, line 874: Sweet! 🎉

mlflow/pyfunc/__init__.py
tomasatdatabricks: I think we should always honor the expected shape as the DL frameworks do.

mlflow/pyfunc/__init__.py
```diff
@@ -283,7 +283,7 @@ def _load_model_env(path):
     return _get_flavor_configuration(model_path=path, flavor_name=FLAVOR_NAME).get(ENV, None)
 
 
-def _enforce_type(name, values: pandas.Series, t: DataType):
+def _enforce_mlflow_type(name, values: pandas.Series, t: DataType):
```
arjundc-db, line 286: nit : Maybe call this _enforce_mlflow_datatype.

arjundc-db (approved):
One nit, sorry I missed it.
Looks great, thanks @wentinghu

PR #4085, opened 2021-02-09
Fix for autolog disable only working when imports in specific order

apurva-koti:

A bug in the code for the `setup_autolog` post import hook for libraries meant that the `disable` flag was only respected when the order of:
- importing the library
- `mlflow.autolog()`
- `mlflow.library.autolog(disable=True)`

is in a specific order. In particular, the `setup_autolog` function called `mlflow.library.autolog` as a post-import hook, but with the _default_ values for its params, rather than the ones that may have already been set as stored in `AUTOLOGGING_INTEGRATIONS`.

Summary of manually verified fixes:
1. mlflow.autolog, then library autolog disable, then import library:
   Now fixed
2. mlflow.autolog, then import library, then library autolog disable
   Always worked
3. library autolog disable, then mlflow.autolog, then import library
   Works - verified
4. library autolog disable, then import library, then mlflow.autolog
   Works - verified
5. import library, then mlflow.autolog, then library autolog disable
   Works - verified. Still prints `autologging successfully enabled for library`. Technically correct, but misleading because it’s immediately disabled afterwards. Should fix in a follow-up effort
6. import library, then library autolog disable, then mlflow.autolog
   Works - verified

## How is this patch tested?

(Details)

Fixed a bug where disabling autolog for a library was only respected if the library import, autolog disable, and mlflow.autolog calls happened in a specific order.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -1385,7 +1385,9 @@ def autolog(
     # this is because on Databricks a SparkSession already exists and the user can directly
     #   interact with it, and this activity should be logged.
     try:
-        autologging_params = get_autologging_params("spark", spark.autolog)
+        set_mlflow_autolog_params_in_config("pyspark")
+        autologging_params = get_autologging_params("pyspark", spark.autolog)
+        autologging_params["_mlflow_called"] = True
         spark.autolog(**autologging_params)
     except ImportError as ie:
         # if pyspark isn't installed, a user could potentially install it in the middle
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -1341,7 +1341,8 @@ def autolog(
             autolog_fn = LIBRARY_TO_AUTOLOG_FN[module.__name__]
             autologging_params = get_autologging_params(autolog_fn)
             autolog_fn(**autologging_params)
-            _logger.info("Autologging successfully enabled for %s.", module.__name__)
+            if not autologging_params.get("disable", False):
+                _logger.info("Autologging successfully enabled for %s.", module.__name__)
         except Exception as e:
             if _is_testing():
                 # Raise unexpected exceptions in test mode in order to detect
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -17,7 +17,7 @@ from mlflow.tracking import artifact_utils, _get_store
 from mlflow.tracking.context import registry as context_registry
 from mlflow.store.tracking import SEARCH_MAX_RESULTS_DEFAULT
 from mlflow.utils import env
-from mlflow.utils.autologging_utils import _is_testing, autologging_integration
+from mlflow.utils.autologging_utils import _is_testing, autologging_integration, AUTOLOGGING_INTEGRATIONS
 from mlflow.utils.databricks_utils import is_in_databricks_notebook, get_notebook_id
 from mlflow.utils.import_hooks import register_post_import_hook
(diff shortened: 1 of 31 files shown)


mlflow/tracking/fluent.py
```diff
@@ -1340,6 +1344,7 @@ def setup_autologging(module):
         try:
             autolog_fn = LIBRARY_TO_AUTOLOG_FN[module.__name__]
             autologging_params = get_autologging_params(autolog_fn)
+            autologging_params.update(AUTOLOGGING_INTEGRATIONS[module.__name__])
```
dbczumar, line 1347: @apurva-koti what happens if I do something like:

```
mlflow.autolog()
import sklearn
mlflow.autolog(disable=True)
```

In this case, will the disable= True property be applied for sklearn autologging? It seems like the second configuration of mlflow.autolog will be ignored. This is important for the databricks autologging case.

mlflow/tracking/fluent.py
```diff
@@ -1329,19 +1333,30 @@ def print_auto_logged_info(r):
...
     }
 
-    def get_autologging_params(autolog_fn):
+    def get_autologging_params(integration, autolog_fn):
         try:
             needed_params = list(inspect.signature(autolog_fn).parameters.keys())
-            return {k: v for k, v in locals_copy if k in needed_params}
+            param_spec = inspect.signature(autolog_fn).parameters
+            default_params = {param.name: param.default for param in param_spec.values()}
+            return {k: AutologgingConfigManager.get_config(integration, k, default_params[k]) for k in needed_params}
```
apurva-koti, line 1341: Previously, this function gathered the values for params required by the autolog fn from `locals()`. We no longer want to directly call autolog functions with the values passed into `mlflow.autolog`.
Changed this function to return the correct configs from `ConfigManager` instead.

mlflow/tracking/fluent.py
```diff
@@ -1329,19 +1333,30 @@ def print_auto_logged_info(r):
...
             return {}
 
+    def set_mlflow_autolog_params_in_config(integration):
+        try:
+            needed_params = list(inspect.signature(autolog).parameters.keys())
+            [AutologgingConfigManager.set_mlflow_config(integration, k, v)
+             for k, v in locals_copy if k in needed_params]
+        except Exception:
+            pass
+
```
apurva-koti, line 1352: Following up on above comment, this new function gathers the arguments of `mlflow.autolog` and sets them in the config manager, key-ed by the integration name.

mlflow/tracking/fluent.py
```diff
@@ -1329,19 +1333,30 @@ def print_auto_logged_info(r):
...
     def setup_autologging(module):
         try:
             autolog_fn = LIBRARY_TO_AUTOLOG_FN[module.__name__]
-            autologging_params = get_autologging_params(autolog_fn)
+            autologging_params = get_autologging_params(module.__name__, autolog_fn)
+            autologging_params["_mlflow_called"] = True
             autolog_fn(**autologging_params)
```
apurva-koti, line 1358: Given the situation:
```
mlflow.autolog(disable=False)
import sklearn
mlflow.autolog(disable=True)
```
The first line would set up an import hook which would run `mlflow.sklearn.autolog()` with the current configs at the time.
`mlflow.sklearn.autolog` receives a call (with some arguments), so it ends up updating the per-integration configs in ConfigManager with the configs that existed at the time of line 1.
Of course, the per-integration config for sklearn should be empty because there is no explicit user-called `mlflow.sklearn.autolog()`. This flag exists to differentiate between an implicit `mlflow.autolog` hook-added call and an explicit user-called `mlflow.sklearn.autolog()`.

apurva-koti:
Superseded by #4093

PR #4099, opened 2021-02-12
Improve method to parse tf-serving json

arjundc-db:

This improves parsing the tf-serving json to perform two tasks:
- 'inputs' now interprets (nested) list as a singular tensor
- Casting the inputted data into the type defined in the dataset schema.

## How is this patch tested?
Unit tests added.

Components 
- [ x] `area/scoring`: Local serving, model deployment tools, spark UDFs

Interface 

Language 

Integrations

- [ x] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes

### DIFF
--- a/mlflow/models/utils.py
+++ b/mlflow/models/utils.py
@@ -68,7 +68,7 @@ class _Example(object):
                     result[name] = input_tensor[name].tolist()
                 return {"inputs": result}
             else:
-                return {"instances": input_example.tolist()}
+                return {"inputs": input_example.tolist()}
 
         def _handle_dataframe_input(input_ex):
             if isinstance(input_ex, dict):
@@ -176,12 +176,12 @@ def _read_example(mlflow_model: Model, path: str):
     input_schema = mlflow_model.signature.inputs if mlflow_model.signature is not None else None
     path = os.path.join(path, mlflow_model.saved_input_example_info["artifact_path"])
     if example_type == "ndarray":
-        return _read_tensor_input_from_json(path)
+        return _read_tensor_input_from_json(path, schema=input_schema)
     else:
         return _dataframe_from_json(path, schema=input_schema, precise_float=True)
 
 
-def _read_tensor_input_from_json(path):
+def _read_tensor_input_from_json(path, schema=None):
     with open(path, "r") as handle:
         inp_dict = json.load(handle)
-        return parse_tf_serving_input(inp_dict)
+        return parse_tf_serving_input(inp_dict, schema)
--- a/mlflow/models/utils.py
+++ b/mlflow/models/utils.py
@@ -68,7 +68,7 @@ class _Example(object):
                     result[name] = input_tensor[name].tolist()
                 return {"inputs": result}
             else:
-                return {"inputs": input_example.tolist()}
+                return {"inputs": input_tensor.tolist()}
 
         def _handle_dataframe_input(input_ex):
             if isinstance(input_ex, dict):
--- a/mlflow/utils/proto_json_utils.py
+++ b/mlflow/utils/proto_json_utils.py
@@ -154,18 +154,20 @@ def parse_tf_serving_input(inp_dict, schema=None):
                     input_data = {input_names[0]: input_data}
                 if not isinstance(input_data, dict):
                     raise MlflowException(
-                        "This model contains a tensor-based model signature with input names, which"
-                        " suggests a dictionary input mapping input name to tensor, but an input of"
-                        " type {0} was found.".format(type(input_data))
+                        "Failed to parse input data. This model contains a tensor-based model"
+                        " signature with input names, which suggests a dictionary input mapping"
+                        " input name to tensor, but an input of type {0} was found.".format(
+                            type(input_data)
+                        )
                     )
                 for col_name, tensor_spec in zip(schema.input_names(), schema.inputs):
                     input_data[col_name] = np.array(input_data[col_name], dtype=tensor_spec.type)
             else:
                 if not isinstance(input_data, list):
                     raise MlflowException(
-                        "This model contains an un-named tensor-based model signature which"
-                        " expects a single n-dimensional array as input, however, an input of"
-                        " type {0} was found.".format(type(input_data))
+                        "Failed to parse input data. This model contains an un-named tensor-based"
+                        " model signature which expects a single n-dimensional array as input,"
+                        " however, an input of type {0} was found.".format(type(input_data))
                     )
                 input_data = np.array(input_data, dtype=schema.inputs[0].type)
         else:
--- a/mlflow/utils/proto_json_utils.py
+++ b/mlflow/utils/proto_json_utils.py
@@ -104,7 +104,7 @@ def _dataframe_from_json(
             path_or_str, orient=pandas_orient, dtype=dtypes, precise_float=precise_float
         )
         actual_cols = set(df.columns)
-        for type_, name in zip(schema.column_types(), schema.input_names()):
+        for type_, name in zip(schema.input_types(), schema.input_names()):
             if type_ == DataType.binary and name in actual_cols:
                 df[name] = df[name].map(lambda x: base64.decodebytes(bytes(x, "utf8")))
         return df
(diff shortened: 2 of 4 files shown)


tests/utils/test_proto_json_utils.py
```diff
@@ -104,36 +188,19 @@ def test_parse_tf_serving_input():
...
     ):
-        parse_tf_serving_input(tfserving_input)
-
-    # values for each column are properly converted to a tensor
-    arr = [
-        [[1, 2, 3], [4, 5, 6], [7, 8, 9]],
-        [[3, 2, 1], [6, 5, 4], [9, 8, 7]],
-    ]
-    tfserving_input = {"instances": arr}
-    result = parse_tf_serving_input(tfserving_input)
```
wentinghu, line 115: why was this test case removed?

tests/utils/test_proto_json_utils.py
```diff
@@ -79,22 +80,105 @@ def test_back_compat():
...
     }
+    # Without Schema
     result = parse_tf_serving_input(tfserving_input)
-    assert (result["a"] == np.array(["s1", "s2", "s3"])).all()
-    assert (result["b"] == np.array([1, 2, 3])).all()
-    assert (result["c"] == np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]])).all()
+    expected_result_no_schema = {}
+    expected_result_no_schema["a"] = np.array(["s1", "s2", "s3"])
+    expected_result_no_schema["b"] = np.array([1.1, 2.2, 3.3], dtype="float64")
+    expected_result_no_schema["c"] = np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype="int64")
```
wentinghu, line 102:
```suggestion
    expected_result_no_schema = {
       "a": np.array(["s1", "s2", "s3"]),
       "b": np.array([1.1, 2.2, 3.3], dtype="float64"),
       "c": np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype="int64")
```
nit to keep it consistent with other examples

tests/utils/test_proto_json_utils.py
```diff
@@ -79,22 +80,105 @@ def test_back_compat():
...
+    # With schema
+    schema = Schema(
+        [
+            TensorSpec(np.dtype("object"), [-1], "a"),
+            TensorSpec(np.dtype("float32"), [-1], "b"),
+            TensorSpec(np.dtype("int32"), [-1], "c"),
+        ]
+    )
+    result = parse_tf_serving_input(tfserving_input, schema)
+    expected_result_schema = {}
```
wentinghu, line 114: same nit as above

wentinghu (approved):
lgtm! thanks for fixing :) few small comments

tests/utils/test_proto_json_utils.py
arjundc-db: Umm, not removed, it was moved here : https://github.com/mlflow/mlflow/pull/4099/files/c1544c874d7be8dcbf3b2afabe349db8e251fba6#diff-5bf4b1bc0b3499b52dd643e44054f4e7410d5e549fb7673ce0c45757f74ff11cR142

wentinghu: oh i'm just bad at reading oops

mlflow/utils/proto_json_utils.py
```diff
@@ -189,4 +194,16 @@ def parse_tf_serving_input(inp_dict):
                 "Failed to parse data as TF serving input. The length of values for"
                 " each input/column name are not the same"
             )
+
+    # Attempt casting
+    if schema is not None:
```
tomasatdatabricks, line 200: can we cast right when we create the array - i.e. call np.array(data, dtype=...))? I am worried that casting at the end adds another data copy.

mlflow/utils/proto_json_utils.py
```diff
@@ -104,7 +104,7 @@ def _dataframe_from_json(
             path_or_str, orient=pandas_orient, dtype=dtypes, precise_float=precise_float
         )
         actual_cols = set(df.columns)
-        for type_, name in zip(schema.column_types(), schema.input_names()):
```
tomasatdatabricks, line 107: good catch

tomasatdatabricks (commented):
Looks good. 

One request - I would prefer if we can create the numpy array with the correct type rather than cast it after the fact.

mlflow/utils/proto_json_utils.py
```diff
@@ -136,14 +136,33 @@ def _get_jsonable_obj(data, pandas_orient="records"):
...
+        if schema is not None:
+            if schema.has_input_names():
+                input_names = schema.input_names()
+                if len(input_names) == 1 and isinstance(input_data, (np.ndarray, list)):
+                    # for schemas with a single column, match input with column
+                    input_data = {input_names[0]: input_data}
+                for col_name, tensor_spec in zip(schema.input_names(), schema.inputs):
+                    input_data[col_name] = np.array(input_data[col_name], dtype=tensor_spec.type)
+            else:
+                input_data = np.array(input_data, dtype=schema.inputs[0].type)
```
tomasatdatabricks, line 158: We should check that the input_data is a list.

mlflow/utils/proto_json_utils.py
```diff
@@ -136,14 +136,33 @@ def _get_jsonable_obj(data, pandas_orient="records"):
...
     import numpy as np
 
+    def cast_schema_type(input_data):
+        if schema is not None:
+            if schema.has_input_names():
+                input_names = schema.input_names()
+                if len(input_names) == 1 and isinstance(input_data, (np.ndarray, list)):
+                    # for schemas with a single column, match input with column
+                    input_data = {input_names[0]: input_data}
+                for col_name, tensor_spec in zip(schema.input_names(), schema.inputs):
```
tomasatdatabricks, line 155: we should check that input is a dictionary at this point.

mlflow/models/utils.py
```diff
@@ -68,7 +68,7 @@ def _handle_tensor_input(input_tensor: Union[np.ndarray, dict]):
                     result[name] = input_tensor[name].tolist()
                 return {"inputs": result}
             else:
-                return {"instances": input_example.tolist()}
+                return {"inputs": input_example.tolist()}
```
wentinghu, line 71: oops sorry missed this
```suggestion
                return {"inputs": input_tensor.tolist()}
```

tomasatdatabricks: nice catch Wendy! I guess the both work in this case cause they are the same?

wentinghu: yeah and `input_example` doesn't get modified anywhere in the constructor so it technically doesn't matter, but felt dangerous to leave as is

mlflow/utils/proto_json_utils.py
```diff
@@ -136,14 +136,45 @@ def _get_jsonable_obj(data, pandas_orient="records"):
...
+    def cast_schema_type(input_data):
+        if schema is not None:
+            if schema.has_input_names():
+                input_names = schema.input_names()
+                if len(input_names) == 1 and isinstance(input_data, list):
+                    # for schemas with a single column, match input with column
+                    input_data = {input_names[0]: input_data}
+                if not isinstance(input_data, dict):
+                    raise MlflowException(
+                        "This model contains a tensor-based model signature with input names, which"
```
tomasatdatabricks, line 157: nit: I think the message should contain something like: Failed to parse, the input ins incompatible with the model signature, since in this context, we are parsing the input data rather than scoring the model .

mlflow/utils/proto_json_utils.py
```diff
@@ -136,14 +136,45 @@ def _get_jsonable_obj(data, pandas_orient="records"):
...
+                        " suggests a dictionary input mapping input name to tensor, but an input of"
+                        " type {0} was found.".format(type(input_data))
+                    )
+                for col_name, tensor_spec in zip(schema.input_names(), schema.inputs):
+                    input_data[col_name] = np.array(input_data[col_name], dtype=tensor_spec.type)
+            else:
+                if not isinstance(input_data, list):
+                    raise MlflowException(
+                        "This model contains an un-named tensor-based model signature which"
```
tomasatdatabricks, line 166: Same as above, slightly different phrasing would be more clear.

tomasatdatabricks (approved):
lgtm!

PR #4101, opened 2021-02-13
Fix `enable_test_mode_by_default_for_autologging_integrations`

harupy:
Signed-off-by: harupy <17039389+harupy@users.noreply.github.com>

## What changes are proposed in this pull request?

Currently `enable_test_mode_by_default_for_autologging_integrations ` just creates a generator and doesn't actually enable the autologging testing mode. This PR fixes the issue by using `yield from`.

```python
@pytest.fixture(autouse=True, scope="session")
def enable_test_mode_by_default_for_autologging_integrations():
    test_mode_on()
    # this just creates a generator and doesn't enable the autologging testing mode
```

## How is this patch tested?

(Details)

### Is this a user-facing change?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

<a name="release-note-category"></a>
### How should the PR be classified in the release notes? Choose one:

[...]

PR #3989, opened 2021-01-15
SHAP Explainer Logging Integration

vivekchettiar:
## What changes are proposed in this pull request?

Addition of 'SHAP' model flavor, Ability to log/load SHAP explainers.

## How is this patch tested?

Added to tests\shap to test logging and loading of explainers that try to explain sklearn and pytorch models

## Release Notes

1) Addition of 'SHAP' model flavor
2) Ability to log/load SHAP explainers (With internal use of MLFlow save/load for underlying models of supported flavors)

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [X] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [X] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [X ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

<a name="release-note-category"></a>
### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
[...]

PR #4102, opened 2021-02-14
[Follow-up for #4101] Fix broken autologging example tests

harupy:

#4101 enabled the autologging testing mode. This change has revealed several hidden errors in the example tests. This PR fixes them:

https://github.com/mlflow/mlflow/runs/1896644303?check_suite_focus=true

## How is this patch tested?

Existing unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,4 +1,3 @@
-# Dummy change to trigger example tests. Remove this before merge
 name: Examples
 
 on:
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,3 +1,4 @@
+
 name: Examples
 
 on:
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,4 +1,4 @@
-
+# Dummy change to trigger example tests. Remove this before merge
 name: Examples
 
 on:
--- a/mlflow/fastai.py
+++ b/mlflow/fastai.py
@@ -511,7 +511,6 @@ def autolog(log_models=True, disable=False, exclusive=False):  # pylint: disable
                     try_mlflow_log(mlflow.log_param, "train_bn", self.train_bn)
 
                 summary = model_summary(self.learner)
-                try_mlflow_log(mlflow.set_tag, "model_summary", summary)
 
                 tempdir = tempfile.mkdtemp()
                 try:
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -17,7 +17,6 @@ import mlflow
 from mlflow.entities.run_status import RunStatus
 from mlflow.entities import Metric
 from mlflow.tracking.client import MlflowClient
-from mlflow.tracking.fluent import _RUN_ID_ENV_VAR
 from mlflow.utils import gorilla
 from mlflow.utils.mlflow_tags import MLFLOW_AUTOLOGGING
 from mlflow.utils.validation import MAX_METRICS_PER_BATCH
@@ -805,7 +804,7 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
     """
 
     def create_managed_run():
-        run_id_env_var_exists = _RUN_ID_ENV_VAR in os.environ
+        run_id_env_var_exists = mlflow.tracking._RUN_ID_ENV_VAR in os.environ
         managed_run = mlflow.start_run(tags=tags)
         _logger.info(
             "Created MLflow autologging run with ID '%s', which will track hyperparameters,"
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -17,6 +17,7 @@ import mlflow
 from mlflow.entities.run_status import RunStatus
 from mlflow.entities import Metric
 from mlflow.tracking.client import MlflowClient
+from mlflow.tracking.fluent import _RUN_ID_ENV_VAR
 from mlflow.utils import gorilla
 from mlflow.utils.mlflow_tags import MLFLOW_AUTOLOGGING
 from mlflow.utils.validation import MAX_METRICS_PER_BATCH
@@ -804,6 +805,7 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
     """
 
     def create_managed_run():
+        run_id_env_var_exists = _RUN_ID_ENV_VAR in os.environ
         managed_run = mlflow.start_run(tags=tags)
         _logger.info(
             "Created MLflow autologging run with ID '%s', which will track hyperparameters,"
@@ -812,6 +814,8 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
             managed_run.info.run_id,
             autologging_integration,
         )
+        if run_id_env_var_exists:
+            try_mlflow_log(mlflow.set_tags, tags)
         return managed_run
 
     if inspect.isclass(patch_function):
--- a/mlflow/utils/autologging_utils.py
+++ b/mlflow/utils/autologging_utils.py
@@ -804,8 +804,8 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
     """
 
     def create_managed_run():
-        run_id_env_var_exists = mlflow.tracking._RUN_ID_ENV_VAR in os.environ
-        managed_run = mlflow.start_run(tags=tags)
+        managed_run = mlflow.start_run()
+        try_mlflow_log(mlflow.set_tags, tags)
         _logger.info(
             "Created MLflow autologging run with ID '%s', which will track hyperparameters,"
             " performance metrics, model artifacts, and lineage information for the"
@@ -813,8 +813,6 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
             managed_run.info.run_id,
             autologging_integration,
         )
-        if run_id_env_var_exists:
-            try_mlflow_log(mlflow.set_tags, tags)
         return managed_run
 
     if inspect.isclass(patch_function):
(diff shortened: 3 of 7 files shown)


mlflow/fastai.py
```diff
@@ -511,7 +511,6 @@ def on_train_begin(self, **kwargs):
                     try_mlflow_log(mlflow.log_param, "train_bn", self.train_bn)
 
                 summary = model_summary(self.learner)
-                try_mlflow_log(mlflow.set_tag, "model_summary", summary)
```
harupy, line 514: We curretly log `model_summary` as an artifact and don't need log it as a tag.

mlflow/utils/autologging_utils.py
```diff
@@ -812,6 +813,8 @@ def create_managed_run():
             managed_run.info.run_id,
             autologging_integration,
         )
+        if run_id_env_var_exists:
+            try_mlflow_log(mlflow.set_tags, tags)
```
harupy, line 817: Fix for this error:

https://github.com/mlflow/mlflow/pull/4102/checks?check_run_id=1897305736#step:7:1605

```
AssertionError: Autologging run with id bfae40e28de046c4b17f590730c3f458 failed to set autologging tag with expected value. Expected: 'pytorch', Actual: 'None'
```

The error above occurs when running `pytorch/MNIST/example1` using `mlflow run`. `mlflow run` creates a new run and store its `run_id` in the `MLFLOW_RUN_ID` environment variable before running an entrypoint script. If `MLFLOW_RUN_ID` exists, `mlflow.start_run` just resumes a run with the specified id and **doesn't set tags**.

harupy: Should we modify `mlflow.start_run` instead?

mlflow/utils/autologging_utils.py
```diff
@@ -804,6 +804,7 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
     """
 
     def create_managed_run():
+        run_id_env_var_exists = mlflow.tracking._RUN_ID_ENV_VAR in os.environ
         managed_run = mlflow.start_run(tags=tags)
```
harupy, line 808: Note that `mlflow.start_run` deletes `MLFLOW_RUN_ID` right after extracting a run_id from it:

https://github.com/mlflow/mlflow/blob/4ba469ea5d67ca327c23a6e57d1fc2d2256c6f1c/mlflow/tracking/fluent.py#L178-L180

eedeleon:
Can you requeue the build for https://github.com/mlflow/mlflow/pull/3989 when this is completed? It resolves the test failure

mlflow/utils/autologging_utils.py
dbczumar: @harupy Thanks for catching this. In this case, I think I'd rather we split the `mlflow.start_run(tags=...)` into two separate API calls; i.e.:

```
mlflow.start_run()
mlflow.set_tags(...)
```

This way, we don't have to special case the environment variable and rely on behaviors internal to `mlflow.start_run()`. I'm hesitant to modify the behavior of `mlflow.start_run()` to set the tags on an existing run, since the goal of this API mode is to resume an existing run, rather than mutate its data.

Can we separate the API calls and update the docs in `mlflow.start_run()` to indicate that `tags` are ignored if an existing run is being resumed? https://github.com/mlflow/mlflow/blob/4ba469ea5d67ca327c23a6e57d1fc2d2256c6f1c/mlflow/tracking/fluent.py#L132

harupy:
@eedeleon Yes :)

mlflow/utils/autologging_utils.py
harupy: @dbczumar Thanks for the feedback. I'll make the change :)

.github/workflows/examples.yml
```diff
@@ -1,3 +1,4 @@
+# Dummy change to trigger example tests. Remove this before merge
```
dbczumar, line 1: LGTM once this is removed :)

dbczumar (approved):
LGTM! Thanks @harupy !

harupy:
@dbczumar Thanks for the quick review!

harupy:
hmm the tensorflow check started failing:

https://github.com/mlflow/mlflow/runs/1914795924#step:5:1196

harupy:
One workaround I came up with is replace `run = mlflow.active_run()` with `run = MlflowClient().get_run(mlflow.active_run().info.run_id)`.

dbczumar:
> One workaround I came up with is replace `run = mlflow.active_run()` with `run = MlflowClient().get_run(mlflow.active_run().info.run_id)`.

@harupy If possible, I'd rather drop the tag-based assignment of `_AUTOLOG_RUN_ID` in `tensorflow.py` entirely (i.e. remove the conditional on https://github.com/mlflow/mlflow/blob/c573aaa2983ca5e2135ae6b4b97515a1f37b7487/mlflow/tensorflow.py#L901-L904). Do we need this conditional / tag check?

harupy:
> Do we need this conditional / tag check?

I think we can drop the conditional.

harupy:
`test_tf_estimator_autolog_persists_manually_created_run` failed with:

```
tmpdir = local('/tmp/pytest-of-runner/pytest-0/test_tf_estimator_autolog_pers0')
export = True

    @pytest.mark.large
    @pytest.mark.parametrize("export", [True, False])
    def test_tf_estimator_autolog_persists_manually_created_run(tmpdir, export):
        directory = tmpdir.mkdir("test")
            create_tf_estimator_model(str(directory), export)
>           assert mlflow.active_run()
E           assert None
E            +  where None = <function active_run at 0x7ff7de93c510>()
E            +    where <function active_run at 0x7ff7de93c510> = mlflow.active_run
```

https://github.com/mlflow/mlflow/pull/4102/checks?check_run_id=1916219422#step:8:117


https://github.com/mlflow/mlflow/blob/c573aaa2983ca5e2135ae6b4b97515a1f37b7487/mlflow/tensorflow.py#L959-L962

These lines seems to be the root cause. I remember the reason we insert `if MLFLOW_AUTOLOGGING in active_run.data.tags: ` is to avoid ending a run if it's a user-generated run.

dbczumar:
> `test_tf_estimator_autolog_persists_manually_created_run` failed with:
> 
> ```
> tmpdir = local('/tmp/pytest-of-runner/pytest-0/test_tf_estimator_autolog_pers0')
> export = True
> 
>     @pytest.mark.large
>     @pytest.mark.parametrize("export", [True, False])
>     def test_tf_estimator_autolog_persists_manually_created_run(tmpdir, export):
>         directory = tmpdir.mkdir("test")
>         with mlflow.start_run() as run:
>             create_tf_estimator_model(str(directory), export)
> >           assert mlflow.active_run()
> E           assert None
> E            +  where None = <function active_run at 0x7ff7de93c510>()
> E            +    where <function active_run at 0x7ff7de93c510> = mlflow.active_run
> ```
> 
> https://github.com/mlflow/mlflow/pull/4102/checks?check_run_id=1916219422#step:8:117
> 
> https://github.com/mlflow/mlflow/blob/c573aaa2983ca5e2135ae6b4b97515a1f37b7487/mlflow/tensorflow.py#L959-L962
> 
> These lines seems to be the root cause. I remember the reason we insert `if MLFLOW_AUTOLOGGING in active_run.data.tags: ` is to avoid ending a run if it's a user-generated run.

What if we have `export_saved_model` only log to a run if `_AUTOLOG_RUN_ID` is already defined, at which point we reset `_AUTOLOG_RUN_ID` to `None`? It seems kind of odd to create an MLflow run for a tensorflow export call if the user didn't do any training with that model.

harupy:
> What if we have export_saved_model only log to a run if _AUTOLOG_RUN_ID is already defined, at which point we reset _AUTOLOG_RUN_ID to None? It seems kind of odd to create an MLflow run for a tensorflow export call if the user didn't do any training with that model.

We could do that but it doesn't resolve an issue where `export_saved_model` terminates both auto-generated and user-generated runs. I thought we could remove the conditional `if MLFLOW_AUTOLOGGING in active_run.data.tags` but I was wrong.

```python
    def train(original, self, *args, **kwargs):
        active_run = mlflow.active_run()
        
        # BEFORE: Set `_AUTOLOG_RUN_ID` only when `active_run`
        if MLFLOW_AUTOLOGGING in active_run.data.tags:
            global _AUTOLOG_RUN_ID
             _AUTOLOG_RUN_ID = active_run.info.run_id
        
        # NOW: Set `_AUTOLOG_RUN_ID` regardless of how `active_run` was generated.
        global _AUTOLOG_RUN_ID
        _AUTOLOG_RUN_ID = active_run.info.run_id
```

dbczumar:
> > What if we have export_saved_model only log to a run if _AUTOLOG_RUN_ID is already defined, at which point we reset _AUTOLOG_RUN_ID to None? It seems kind of odd to create an MLflow run for a tensorflow export call if the user didn't do any training with that model.
> 
> We could do that but it doesn't resolve an issue where `export_saved_model` terminates both auto-generated and user-generated runs. I thought we could remove the conditional `if MLFLOW_AUTOLOGGING in active_run.data.tags` but I was wrong.
> 
> ```python
>     def train(original, self, *args, **kwargs):
>         active_run = mlflow.active_run()
>         
>         # BEFORE: Set `_AUTOLOG_RUN_ID` only when `active_run`
>         if MLFLOW_AUTOLOGGING in active_run.data.tags:
>             global _AUTOLOG_RUN_ID
>              _AUTOLOG_RUN_ID = active_run.info.run_id
>         
>         # NOW: Set `_AUTOLOG_RUN_ID` regardless of how `active_run` was generated.
>         global _AUTOLOG_RUN_ID
>         _AUTOLOG_RUN_ID = active_run.info.run_id
>         
>         # NOTE `export_saved_model` terminates an active run if it exists and its run id matches `_AUTOLOG_RUN_ID`
> ```

My proposal is that `export_saved_model` shouldn't terminate a run. We can just use `MlflowClient().log_artifact()` to log an artifact to `_AUTOLOG_RUN_ID`.

dbczumar (approved):
LGTM! Thanks @harupy !

PR #4110, opened 2021-02-17
Support parsing df input with tensor schema and parsing tensor input with df schema. 

tomasatdatabricks:
Signed-off-by: tomasatdatabricks <tomas.nykodym@databricks.com>

As part of adding support for tensor input to MLflow models, we've introduced a tensor schema and tensor JSON input. The schema is used during JSON parsing to cast the input to correct data types. The assumption is that tensor models will be scored with tensor input and dataframe models will be scored with dataframe input. However, this does not always have to be the case as 2d input can be represented  as both tensor or dataframe input. This PR fixes the behavior of model serving when schema type and json type are not the same. 

## How is this patch tested?
Added tests to test_proto_utils.

## Release Notes
n/a

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/proto_json_utils.py
+++ b/mlflow/utils/proto_json_utils.py
@@ -102,7 +102,7 @@ def _dataframe_from_json(
         if schema.is_tensor_spec():
             # The schema can be either:
             #  - a single tensor: attempt to parse all columns with the same dtype
-            #  - a dictionary of tensors: each column gets the type from ____ named tensor
+            #  - a dictionary of tensors: each column gets the type from an equally named tensor
             if len(schema.inputs) == 1:
                 dtypes = schema.numpy_types()[0]
             else:
--- a/mlflow/utils/proto_json_utils.py
+++ b/mlflow/utils/proto_json_utils.py
@@ -171,9 +171,11 @@ def parse_tf_serving_input(inp_dict, schema=None):
                             type(input_data)
                         )
                     )
-                for col_name, col_type in zip(schema.input_names(), schema.numpy_types()):
-                    if col_name in input_data:
-                        input_data[col_name] = np.array(input_data[col_name], dtype=col_type)
+                type_dict = dict(zip(schema.input_names(), schema.numpy_types()))
+                for col_name in input_data.keys():
+                    input_data[col_name] = np.array(
+                        input_data[col_name], dtype=type_dict.get(col_name)
+                    )
             else:
                 if not isinstance(input_data, list):
                     raise MlflowException(
@@ -181,7 +183,7 @@ def parse_tf_serving_input(inp_dict, schema=None):
                         " model signature which expects a single n-dimensional array as input,"
                         " however, an input of type {0} was found.".format(type(input_data))
                     )
-                input_data = np.array(input_data, dtype=schema.inputs[0].type)
+                input_data = np.array(input_data, dtype=schema.numpy_types()[0])
         else:
             if isinstance(input_data, dict):
                 input_data = {k: np.array(v) for k, v in input_data.items()}
--- a/mlflow/utils/proto_json_utils.py
+++ b/mlflow/utils/proto_json_utils.py
@@ -102,7 +102,7 @@ def _dataframe_from_json(
         if schema.is_tensor_spec():
             # The schema can be either:
             #  - a single tensor: attempt to parse all columns with the same dtype
-            #  - a dictionary of tensors: each column gets the type from equaly named tensor
+            #  - a dictionary of tensors: each column gets the type from ____ named tensor
             if len(schema.inputs) == 1:
                 dtypes = schema.numpy_types()[0]
             else:
--- a/mlflow/utils/proto_json_utils.py
+++ b/mlflow/utils/proto_json_utils.py
@@ -99,14 +99,25 @@ def _dataframe_from_json(
     from mlflow.types import DataType
 
     if schema is not None:
-        dtypes = dict(zip(schema.input_names(), schema.pandas_types()))
+        if schema.is_tensor_spec():
+            # The schema can be either:
+            #  - a single tensor: attempt to parse all columns with the same dtype
+            #  - a dictionary of tensors: each column gets the type from equaly named tensor
+            if len(schema.inputs) == 1:
+                dtypes = schema.numpy_types()[0]
+            else:
+                dtypes = dict(zip(schema.input_names(), schema.numpy_types()))
+        else:
+            dtypes = dict(zip(schema.input_names(), schema.pandas_types()))
+
         df = pd.read_json(
             path_or_str, orient=pandas_orient, dtype=dtypes, precise_float=precise_float
         )
-        actual_cols = set(df.columns)
-        for type_, name in zip(schema.input_types(), schema.input_names()):
-            if type_ == DataType.binary and name in actual_cols:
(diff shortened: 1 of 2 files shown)


mlflow/utils/proto_json_utils.py
```diff
@@ -99,14 +99,25 @@ def _dataframe_from_json(
...
+        if schema.is_tensor_spec():
+            # The schema can be either:
+            #  - a single tensor: attempt to parse all columns with the same dtype
+            #  - a dictionary of tensors: each column gets the type from equaly named tensor
+            if len(schema.inputs) == 1:
+                dtypes = schema.numpy_types()[0]
+            else:
+                dtypes = dict(zip(schema.input_names(), schema.numpy_types()))
+        else:
+            dtypes = dict(zip(schema.input_names(), schema.pandas_types()))
```
wentinghu, line 111: should we check that input names isn't none first?

wentinghu (approved):
lgtm, thanks for catching this!

mlflow/utils/proto_json_utils.py
tomasatdatabricks: discussed offline: the input names are actually returned as integers 0..N in case the schema is unnamed.

mlflow/utils/proto_json_utils.py
```diff
@@ -99,14 +99,25 @@ def _dataframe_from_json(
     from mlflow.types import DataType
 
     if schema is not None:
-        dtypes = dict(zip(schema.input_names(), schema.pandas_types()))
+        if schema.is_tensor_spec():
+            # The schema can be either:
+            #  - a single tensor: attempt to parse all columns with the same dtype
+            #  - a dictionary of tensors: each column gets the type from equaly named tensor
```
arjundc-db, line 105: nit Typo:  `an equally` though `a similarly` may be better
```suggestion
            #  - a dictionary of tensors: each column gets the type from ____ named tensor
```

mlflow/utils/proto_json_utils.py
```diff
@@ -99,14 +99,25 @@ def _dataframe_from_json(
...
-        actual_cols = set(df.columns)
-        for type_, name in zip(schema.input_types(), schema.input_names()):
-            if type_ == DataType.binary and name in actual_cols:
-                df[name] = df[name].map(lambda x: base64.decodebytes(bytes(x, "utf8")))
+        if not schema.is_tensor_spec():
+            actual_cols = set(df.columns)
+            for type_, name in zip(schema.input_types(), schema.input_names()):
+                if type_ == DataType.binary and name in actual_cols:
+                    df[name] = df[name].map(lambda x: base64.decodebytes(bytes(x, "utf8")))
         return df
```
arjundc-db, line 121: Should there a corresponding else that converts it to a numpy input, since we know the signature is tensorSpec.

tests/utils/test_proto_json_utils.py
```diff
@@ -106,18 +110,22 @@ def assert_result(result, expected_result):
     # With schema
     schema = Schema(
         [
-            TensorSpec(np.dtype("object"), [-1], "a"),
+            TensorSpec(np.dtype("str"), [-1], "a"),
             TensorSpec(np.dtype("float32"), [-1], "b"),
             TensorSpec(np.dtype("int32"), [-1], "c"),
         ]
     )
+    dfSchema = Schema([ColSpec("string", "a"), ColSpec("float", "b"), ColSpec("integer", "c"),])
```
arjundc-db, line 118: Nit : move this after the comment on line 126.
```suggestion
    dfSchema = Schema([ColSpec("string", "a"), ColSpec("float", "b"), ColSpec("integer", "c")])
```

tests/utils/test_proto_json_utils.py
```diff
@@ -223,3 +235,94 @@ def test_parse_tf_serving_raises_expected_errors():
...
+            ColSpec("boolean", "boolean"),
+            ColSpec("string", "string"),
+            ColSpec("float", "float"),
+            ColSpec("double", "double"),
+            ColSpec("integer", "integer"),
+            ColSpec("long", "long"),
+            ColSpec("binary", "binary"),
+        ]
+    )
+    expected_dtypes = {
```
arjundc-db, line 267: Unused variable.

tests/utils/test_proto_json_utils.py
```diff
@@ -223,3 +235,94 @@ def test_parse_tf_serving_raises_expected_errors():
...
+    # NB: tensor schema does not automatically decode base64 encoded bytes.
+    assert parsed.equals(jsonable_df)
+
+    # Test parse with TesnorSchema with a single tensor
+    tensor_schema = Schema([TensorSpec(np.dtype("float32"), [-1, 3])])
+    source = pd.DataFrame(
+        {
+            "a": np.array([1, 2, 3], dtype=np.float32),
+            "b": np.array([4, 5, 6], dtype=np.float32),
+            "c": np.array([7, 8, 9], dtype=np.float32),
```
arjundc-db, line 315: This test will make more sense if the dtypes of the remaining inputs are not same, to ensure we can check that they were converted based on the singular tensorSpec in the Schema

arjundc-db (changes requested):

mlflow/utils/proto_json_utils.py
arjundc-db: Ignore this one.

tests/utils/test_proto_json_utils.py
```diff
@@ -85,6 +88,7 @@ def assert_result(result, expected_result):
         assert result.keys() == expected_result.keys()
         for key in result:
             assert (result[key] == expected_result[key]).all()
+            assert result[key].dtype == expected_result[key].dtype
```
arjundc-db, line 91: The windows test is failing here because python interprets int as 32 bits be default in windows
https://stackoverflow.com/questions/22513445/python-handles-long-ints-differently-on-windows-and-unix.
Arggg.

mlflow/utils/proto_json_utils.py
```diff
@@ -99,14 +99,25 @@ def _dataframe_from_json(
     from mlflow.types import DataType
 
     if schema is not None:
-        dtypes = dict(zip(schema.input_names(), schema.pandas_types()))
+        if schema.is_tensor_spec():
+            # The schema can be either:
+            #  - a single tensor: attempt to parse all columns with the same dtype
+            #  - a dictionary of tensors: each column gets the type from ____ named tensor
```
arjundc-db, line 105: Sorry I kept it as `____` previously if a better name came to mind.
```suggestion
            #  - a dictionary of tensors: each column gets the type from an equally named tensor
```

tests/utils/test_proto_json_utils.py
tomasatdatabricks: Hmm, I don't understand what do you mean? The types have to be the same since we parse them as a single type. We do make sure they're typecasted to float32, without casting they would've been read as int64. I guess I can make some of the columns floats.

tests/utils/test_proto_json_utils.py
tomasatdatabricks: Hmm thanks, I'll look into that.

tests/utils/test_proto_json_utils.py
arjundc-db: Ohh that makes sense, my bad.

arjundc-db (approved):
LGTM, thank you @tomasatdatabricks

PR #4118, opened 2021-02-19
Avoid duplicate metrics logging in pytorch-lightning 1.2.0

harupy:

In pytorch-lightning 1.2.0 (released yesterday), the `on_epoch_callback` is called twice, once after train epoch, and once after validation epoch: See the PR below for details:

https://github.com/PyTorchLightning/pytorch-lightning/pull/5986 

This change makes our pytorch autologging callback log the same metrics:

https://github.com/mlflow/mlflow/runs/1933446600?check_suite_focus=true#step:8:33

```
E       AssertionError: assert 40 == 20
E        +  where 40 = len([<Metric: key='loss', step=0, timestamp=1613719593042, value=0.9595870971679688>, <Metric: key='loss', step=0, timesta...93481, value=0.7134160995483398>, <Metric: key='loss', step=2, timestamp=1613719593495, value=0.7134160995483398>, ...])
```

This PR fixes this issue by using `on_train_epoch_end` and `on_validation_epoch_end`.

## How is this patch tested?

Existing unit tests + a test ensuring pytorch autologging works a model which doesn't contain `validation_step`

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pytorch/_pytorch_autolog.py
+++ b/mlflow/pytorch/_pytorch_autolog.py
@@ -2,9 +2,10 @@ from distutils.version import LooseVersion
 import logging
 import mlflow.pytorch
 import os
-import pytorch_lightning as pl
 import shutil
 import tempfile
+from distutils.version import LooseVersion
+import pytorch_lightning as pl
 from pytorch_lightning.core.memory import ModelSummary
 from pytorch_lightning.utilities import rank_zero_only
 
@@ -87,6 +88,9 @@ def _create_patch_fit(log_every_n_epoch=1, log_models=True):
 
             def __init__(self):
                 self.early_stopping = False
+                self._is_pytorch_lightning_older_than_1_2_0 = LooseVersion(
+                    pl.__version__
+                ) < LooseVersion("1.2.0")
 
             def _log_metrics(self, trainer, pl_module):
                 if (pl_module.current_epoch + 1) % every_n_epoch == 0:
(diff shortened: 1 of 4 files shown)


tests/pytorch/iris.py
```diff
@@ -58,5 +60,24 @@ def test_step(self, batch, batch_idx):
         self.log("test_acc", self.test_acc.compute())
 
 
+class IrisClassificationWithoutValidation(IrisClassificationBase):
```
harupy, line 63: Defined a new lightning-module class that doesn't override `validation_step` for testing

mlflow/pytorch/_pytorch_autolog.py
```diff
@@ -106,6 +104,46 @@ def on_epoch_end(self, trainer, pl_module):
                     if isinstance(callback, pl.callbacks.early_stopping.EarlyStopping):
                         self._early_stop_check(callback)
 
+            def on_train_epoch_end(self, trainer, pl_module, _):
```
dbczumar, line 107: Rather than checking `_is_pytorch_lightning_older_than_1_2_0` each time one of these callback methods is invoked,  we could conditionally define these callback methods: e.g.,

```
if LooseVersion(pl.__version) >= LooseVersion("1.2.0"):

    def on_train_epoch_end(...)
    def on_validation_epoch_end(...)

else:

    def on_epoch_end(...)
``` 

This may be a bit easier to read / reason about and will remove some boolean checks when executing callback hooks.

harupy: oh I didn't know we could do that. Let me fix that

harupy: @dbczumar Fixed!

dbczumar (approved):
LGTM! Thanks @harupy !

harupy:
@dbczumar Thanks for the review & approval!

etiennedub:
Hi, could I know why you remove the condition on `trainer.running_sanity_check`?
In my case, the validation metric on the first epoch is log 2 times, one time after the validation sanity check and one time at the end of the epoch.
If there any reason to remove this condition?

harupy:
Hi @etiennedub, could you provide code to reproduce the issue?

harupy:
@etiennedub I just confirmed the issue. Let me file a PR.

PR #4119, opened 2021-02-19
Disable autologging for untested library versions

WeichenXu123:

 Disable autologging for untested library versions

## How is this patch tested?

Unit tests and Integration test.

Doc build test snapshot:
Gluon:
image

Pytorch:
image

Tensorflow:
image

Components 

Interface 

Language 

Integrations


.github/workflows/cross-version-tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           fi
 
           if [ "${{ github.event_name }}" = "pull_request" ]; then
-            REF_VERSIONS_YAML="https://raw.githubusercontent.com/mlflow/mlflow/master/ml-package-versions.yml"
+            REF_VERSIONS_YAML="https://raw.githubusercontent.com/mlflow/mlflow/master/mlflow/ml-package-versions.yml"
```
harupy, line 36: Is this a valid file URL?

WeichenXu123: I move ml-package-versions.yml into `mlflow` sub-directory. So that when install package, it will be installed inside the `.../site-package/mlflow/ml-package-versions.yml`
otherwise it will be installed in `.../site-package/ml-package-versions.yml`

mlflow/utils/autologging_utils.py
```diff
@@ -381,7 +415,15 @@ def autologging_is_disabled(flavor_name):
 
     :param flavor_name: An autologging integration flavor name.
     """
-    return get_autologging_config(flavor_name, "disable", True)
+    explicit_disabled = get_autologging_config(flavor_name, "disable", False)
```
WeichenXu123, line 418: @harupy 
I change this line 
`get_autologging_config(flavor_name, "disable", True)`
to
`get_autologging_config(flavor_name, "disable", False)`
is it right ?

harupy: why did you need this change?

WeichenXu123: Because in every package autolog definition the disable argument default value is False, like:
https://github.com/mlflow/mlflow/blob/2d3ac48b01da021d220fb1bf4450c70976f3b5d7/mlflow/sklearn/__init__.py#L536

CC @dbczumar

.github/workflows/cross-version-tests.yml
harupy: ah I missed that.

mlflow/utils/autologging_utils.py
dbczumar: @WeichenXu123 we should keep the previous construction of the line, i.e.:

```
is_autologging_disabled_for_flavor = get_autologging_config(flavor_name, "disable", True)
```

The reason is that we need to capture explicit *and* implicit disablement. For example, in the case of scikit-learn, if the user hasn't called the `mlflow.autolog()` or `mlflow.sklearn.autolog()` APIs, then `autologging_is_disabled("sklearn")` should return `True`. However, in this case, it seems to return `False`:

```
In [1]: import mlflow

In [2]: from mlflow.utils.autologging_utils import autologging_is_disabled

In [3]: autologging_is_disabled("sklearn")
Out[3]: False
```

Can we fix this and add test coverage?

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,36 @@ def batch_metrics_logger(run_id):
...
+        return LooseVersion(v).version[:2]
+
+    return get_minor_version(min_version) <= get_minor_version(pkg_ver) \
+        <= get_minor_version(max_version)
+
+
+def is_autologging_tested(flavor_name):
+    version_file_path = resource_filename(__name__, "../ml-package-versions.yml")
+    with open(version_file_path) as f:
+        version_file_json = yaml.load(f, Loader=yaml.SafeLoader)
```
dbczumar, line 329: can we load this file once at the module level to avoid reading / parsing YAML every time `autologging_is_disabled` is called?

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,36 @@ def batch_metrics_logger(run_id):
...
+    min_version = version_file_json[lib_key]["autologging"]["minimum"]
+    max_version = version_file_json[lib_key]["autologging"]["maximum"]
+
+    pkg_ver = importlib.import_module(module_name).__version__
+
+    def get_minor_version(v):
+        return LooseVersion(v).version[:2]
+
+    return get_minor_version(min_version) <= get_minor_version(pkg_ver) \
+        <= get_minor_version(max_version)
```
dbczumar, line 323: As discussed offline, can we amend this logic to include micro version discrepancies? I.e. `0.23.3` should be considered untested if `0.23.2` was the last tested library version prior to an MLflow release.

mlflow/fastai.py
```diff
@@ -380,7 +380,9 @@ def load_model(model_uri):
 
 @experimental
 @autologging_integration(FLAVOR_NAME)
-def autolog(log_models=True, disable=False, exclusive=False):  # pylint: disable=unused-argument
+def autolog(
+    log_models=True, disable=False, exclusive=False, disable_for_untested_versions=False
```
dbczumar, line 384: As part of the documentation effort here, can we document the min and max tested versions for each autologging integration? We should be able to accomplish this by augmenting the `@autologging_integration` decorator to add / format some content into the docstring (see the definition of `@experimental` in `annotations.py` - https://github.com/mlflow/mlflow/blob/2d3ac48b01da021d220fb1bf4450c70976f3b5d7/mlflow/utils/annotations.py#L5-L17 - for a similar reference).

One thought would be to prepend a notice like the following before the remainder of the docstring:

```
Autologging for fastai has been tested with versions ``1.0.60`` thru ``1.0.61`` and may not work for versions outside of this range. To try autologging with untested versions, specify ``disable_for_untested_versions=True``.
```

dbczumar, line 384: @smurching @harupy @WeichenXu123 I'd like to rename `disable_for_untested_versions` to `disable_for_unsupported_versions` because, in the future, there may be other reasons that a version is excluded other than it being untested - e.g. in the event of a broken / fundamentally incompatible ML library release.

We can clarify via documentation that "unsupported versions" includes ML library versions with known incompatibilities *and* versions that have not been tested prior to the given MLflow release. We can also recommend that users upgrade their MLflow client if possible in order to use autologging with the latest ML library versions.

Thoughts?

mlflow/fastai.py
```diff
@@ -398,6 +400,7 @@ def autolog(log_models=True, disable=False, exclusive=False):  # pylint: disable
     :param exclusive: If ``True``, autologged content is not logged to user-created fluent runs.
                       If ``False``, autologged content is logged to the active fluent run,
                       which may be user-created.
+    :param disable_for_untested_versions: Disable for untested versions
```
dbczumar, line 403: Can we be a bit more specific with the docs here? E.g., "If True, disable autologging for versions of fastai that have not been tested against this version of the MLflow client or are incompatible." (Feel free to modify / refine the phrasing here)

mlflow/fastai.py
dbczumar, line 384: When a user calls `mlflow.autolog()`, `mlflow.fastai.autolog()`, etc with `disable_for_untested_versions=False`, can we still emit a warning message if the user is using an untested / unsupported library version via `_logger.warn`? It would be great if this message also encouraged users to either upgrade MLflow or downgrade the ML library to a documented supported version if they encounter errors during autologging.

The rationale behind this warning message is that users may be unaware of this flag because it defaults to `False`, so a warning message will help users interpret errors / failures.

dbczumar (commented):
@WeichenXu123 This is awesome! Great start! In addition to a few inline comments, I'm leaving a couple notes here as a reminder:

1. Let's make sure to add unit & integration test coverage for disable_for_untested_versions=True/False. For integration testing, i'd recommend picking a single integration (e.g. `sklearn`), mocking the `__version__` attribute, and running example training workflows to ensure that runs are / are *not* created based on the configuration and library version.

2. This is more of a reminder for myself: after merging this change, prior to releasing MLflow 1.15.0, we should create an automated CI job to update `ml-package-versions.yml` with the latest tested versions prior to each MLflow client release.

mlflow/utils/autologging_utils.py
WeichenXu123: OK.

One issue to clarify, in this case, should `0.23.1` regarded as tested or not ? @dbczumar @harupy

mlflow/fastai.py
harupy: +1

mlflow/utils/autologging_utils.py
WeichenXu123: There's already a test cover this case. But, strange, the test is skipped.
https://github.com/mlflow/mlflow/blob/2d3ac48b01da021d220fb1bf4450c70976f3b5d7/tests/autologging/test_autologging_utils.py#L560

It is skipped because of this mark:
https://github.com/mlflow/mlflow/blob/2d3ac48b01da021d220fb1bf4450c70976f3b5d7/tests/autologging/test_autologging_utils.py#L32

CC @harupy Could you help check this ?

mlflow/fastai.py
WeichenXu123: Anther question is when disable_for_untested_versions=True, and meet unsupported version, should we also print message say it is unsupported, or just silently log nothing ?

mlflow/utils/autologging_utils.py
harupy: hmm, looks like files under `tests/autologging` are not executed.

mlflow/utils/autologging_utils.py
dbczumar: we should consider `0.23.1` to be tested, since we would’ve tested it at some point in the past prior to the release of `0.23.2`

WeichenXu123:
@dbczumar 

* I add `statsmodels` section into `ml-package-versions.yml`
* I do not add `disable_for_unsupported_versions` argument for `mlflow.spark.autolog`, because in its code it already forbid unsupported version:
```
    if _get_spark_major_version(spark_context) < 3:
        raise MlflowException("Spark autologging unsupported for Spark versions < 3")
```

mlflow/ml-package-versions.yml
```diff
@@ -246,3 +246,22 @@ spacy:
...
+    requirements: []
+    run: |
+      pytest tests/statsmodels/test_statsmodels_model_export.py --large
+
+  autologging:
+    minimum: "0.11.1"
+    maximum: "0.12.2"
+    requirements: []
+    run: |
+      pytest tests/statsmodels/test_statsmodels_autolog.py --large
```
harupy, line 267: Can we create a separate PR for this change?

mlflow/ml-package-versions.yml
```diff
@@ -246,3 +246,22 @@ spacy:
...
+
+
+statsmodels:
+  package_info:
+    pip_release: "statsmodels"
+
+  models:
+    minimum: "0.11.1"
+    maximum: "0.12.2"
+    requirements: []
```
harupy, line 258: I think `requirements` can be removed

mlflow/ml-package-versions.yml
```diff
@@ -246,3 +246,22 @@ spacy:
     requirements: ["scikit-learn"]
     run: |
       pytest tests/spacy/test_spacy_model_export.py --large
+
+
+statsmodels:
+  package_info:
+    pip_release: "statsmodels"
```
harupy, line 253:
```suggestion
    pip_release: "statsmodels"
    install_dev: |
      pip install git+https://github.com/statsmodels/statsmodels
```

Can we add `install_dev` to run the statsmodels tests using the dev version of statsmodels?

harupy:
https://github.com/PyTorchLightning/pytorch-lightning/pull/6162/ broke `Cross version tests / test (pytorch-lightning / dev / autologging) (pull_request)`.

mlflow/utils/autologging_utils.py
```diff
@@ -353,6 +431,11 @@ def autolog(*args, **kwargs):
         # function, allowing the integration name to be extracted from the function. This is used
         # during the execution of import hooks for `mlflow.autolog()`.
         wrapped_autolog.integration_name = name
+
+        if name in _flavor_to_module_name_and_json_key:
+            wrapped_autolog.__doc__ = (
+                gen_autologging_package_version_requirements_doc(name) + wrapped_autolog.__doc__
+            )
```
dbczumar, line 438: This seems to have a strange interaction with the `@experimental` decorator. When `@experimental` and `@autologging_integration` are stacked on a method, the resulting docs look like:

Screen Shot 2021-02-25 at 7 37 26 PM

Observe that all of the method params / descriptions are now contained in a blue `Note` box, which isn't the desired behavior (only the version information should be shown in the `Note` box - see https://mlflow.org/docs/latest/python_api/mlflow.fastai.html#mlflow.fastai.autolog for reference).

To debug, you can build the MLflow docs locally by navigating to the `mlflow/docs` folder and running `make rsthtml`. This will generate html docs in the `mlflow/docs/build` folder that you can open in your browser.

mlflow/utils/autologging_utils.py
```diff
@@ -353,6 +431,11 @@ def autolog(*args, **kwargs):
         # function, allowing the integration name to be extracted from the function. This is used
         # during the execution of import hooks for `mlflow.autolog()`.
         wrapped_autolog.integration_name = name
+
+        if name in _cross_tested_flavor_to_module_name_and_json_key:
+            wrapped_autolog.__doc__ = (
```
dbczumar, line 436: This seems to have a strange interaction with the `@experimental` decorator. When `@experimental` and `@autologging_integration` are stacked on a method, the resulting docs look like:

Screen Shot 2021-02-25 at 7 37 26 PM

Observe that all of the method params / descriptions are now contained in a blue `Note` box, which isn't the desired behavior (only the version information should be shown in the `Note` box - see https://mlflow.org/docs/latest/python_api/mlflow.fastai.html#mlflow.fastai.autolog for reference).

To debug, you can build the MLflow docs locally by navigating to the `mlflow/docs` folder and running `make rsthtml`. This will generate html docs in the `mlflow/docs/build` folder that you can open in your browser.

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,80 @@ def batch_metrics_logger(run_id):
...
+    def get_module_version(module_name):
+        return importlib.import_module(module_name).__version__
+
+    return _is_autologging_supported(flavor_name, get_module_version)
+
+
+def gen_autologging_package_version_requirements_doc(flavor_name):
+    def _gen_single_requirement(json_key):
+        min_ver, max_ver, pip_release = _get_min_max_version_and_pip_release(json_key)
+        return "{min_ver}<={pip_release}<={max_ver}".format(
```
dbczumar, line 370: Formatting suggestion to render prettier / more readable rst:

```suggestion
        return "``{min_ver}`` <= ``{pip_release}`` <= ``{max_ver}``".format(
```

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,80 @@ def batch_metrics_logger(run_id):
...
+        [
+            _gen_single_requirement(json_key)
+            for _, json_key in _cross_tested_flavor_to_module_name_and_json_key[flavor_name]
+        ]
+    )
+
+    return (
+        ".. Note:: Only supported autologging integration with the following package versions: "
+        + required_pkg_versions
+    )
```
dbczumar:
```suggestion
    return (
        ".. Note:: Autologging is known to be compatible with the following package versions: "
        + required_pkg_versions
        + ". Autologging may not succeed when used with package versions outside of this range.
    )
```

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,80 @@ def batch_metrics_logger(run_id):
...
+        [
+            _gen_single_requirement(json_key)
+            for _, json_key in _cross_tested_flavor_to_module_name_and_json_key[flavor_name]
+        ]
+    )
+
+    return (
+        ".. Note:: Only supported autologging integration with the following package versions: "
+        + required_pkg_versions + "\n"
+    )
```
dbczumar, line 384:
```suggestion
    return (
        ".. Note:: Autologging is known to be compatible with the following package versions: "
        + required_pkg_versions
        + ". Autologging may not succeed when used with package versions outside of this range.
    )
```

mlflow/utils/autologging_utils.py
```diff
@@ -381,7 +464,20 @@ def autologging_is_disabled(flavor_name):
...
+    explicit_disabled = get_autologging_config(flavor_name, "disable", True)
+    if explicit_disabled:
+        return True
+
+    if flavor_name in _cross_tested_flavor_to_module_name_and_json_key and not is_autologging_supported(flavor_name):
+        if get_autologging_config(flavor_name, "disable_for_unsupported_versions", False):
+            return True
+        else:
+            _logger.warning(
+                "You are using an unsupported library version for flavor %s, you can upgrade MLflow version or downgrade relative library version if you encounter errors during autologging.",
```
dbczumar, line 476:
```suggestion
                "You are using an unsupported version of %s. If you encounter errors during autologging, try upgrading / downgrading %s to a supported version, or try upgrading MLflow.",
```

mlflow/fastai.py
```diff
@@ -380,7 +380,9 @@ def load_model(model_uri):
 
 @experimental
 @autologging_integration(FLAVOR_NAME)
-def autolog(log_models=True, disable=False, exclusive=False):  # pylint: disable=unused-argument
+def autolog(
```
dbczumar, line 383: Looks like we're missing `disable_for_unsupported_versions` in `mlflow.autolog()`. Can we add this argument to `mlflow.autolog()` (defined in `mlflow/fluent.py`) & ensure that we have test coverage in https://github.com/mlflow/mlflow/blob/master/tests/tracking/fluent/test_fluent_autolog.py#L123-L137?

mlflow/ml-package-versions.yml
```diff
@@ -262,5 +262,6 @@ statsmodels:
   autologging:
     minimum: "0.11.1"
     maximum: "0.12.2"
+
```
dbczumar, line 265: Nit: Can we revert this newline?

mlflow/utils/autologging_utils.py
dbczumar, line 476: @WeichenXu123 Can we call this when the user calls `mlflow.autolog()`, `mlflow.skearn.autolog()`, etc, rather than calling it every time `autologging_is_disabled` is checked? This gets pretty verbose / redundant if I'm training multiple models / training models multiple times.

For example, the following workflow calls `sklearn.fit()` 3 times and results in 3 "unsupported" log messages:

```
In [1]: import mlflow
   ...: mlflow.autolog()
   ...:
   ...:
   ...: from sklearn import datasets
   ...: from sklearn.linear_model import LogisticRegression
   ...:
   ...: iris = datasets.load_iris()
   ...:
   ...: logistic = LogisticRegression(solver="saga", max_iter=200)
   ...: logistic.fit(iris.data, iris.target)
   ...: logistic.fit(iris.data, iris.target)
   ...: logistic.fit(iris.data, iris.target)
2021/02/25 20:08:32 INFO mlflow.tracking.fluent: Autologging successfully enabled for sklearn.
2021/02/25 20:08:32 WARNING mlflow.utils.autologging_utils: You are using an unsupported library version for flavor sklearn, you can upgrade MLflow version or downgrade relative library version if you encounter errors during autologging.
2021/02/25 20:08:32 INFO mlflow.utils.autologging_utils: Created MLflow autologging run with ID 'b7a55f2013774656855b9a085c991d64', which will track hyperparameters, performance metrics, model artifacts, and lineage information for the current sklearn workflow
2021/02/25 20:08:32 WARNING mlflow.utils.autologging_utils: You are using an unsupported library version for flavor sklearn, you can upgrade MLflow version or downgrade relative library version if you encounter errors during autologging.
2021/02/25 20:08:32 INFO mlflow.utils.autologging_utils: Created MLflow autologging run with ID 'ff10954c52044d00bcf862e1dc0ac50d', which will track hyperparameters, performance metrics, model artifacts, and lineage information for the current sklearn workflow
2021/02/25 20:08:32 WARNING mlflow.utils.autologging_utils: You are using an unsupported library version for flavor sklearn, you can upgrade MLflow version or downgrade relative library version if you encounter errors during autologging.
2021/02/25 20:08:32 INFO mlflow.utils.autologging_utils: Created MLflow autologging run with ID '43dda6d3bcea48bc8b28cb3b968cf3e8', which will track hyperparameters, performance metrics, model artifacts, and lineage information for the current sklearn workflow
```

mlflow/utils/autologging_utils.py
harupy: @WeichenXu123 This seems related to the indent level. I've experineced something similar before.

```
.. note:
   blah

   .. note:
       blah
```

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +644,134 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+
+def test_is_autologging_supported():
+    assert _check_version_in_range("1.0.2", "1.0.1", "1.0.3")
+    assert _check_version_in_range("1.0.1", "1.0.1", "1.0.3")
+    assert _check_version_in_range("1.0.3", "1.0.1", "1.0.3")
+    assert not _check_version_in_range("1.0.0", "1.0.1", "1.0.3")
+    assert not _check_version_in_range("1.0.4", "1.0.1", "1.0.3")
+    assert not _check_version_in_range("0.99.99", "1.0.1", "1.0.3")
+    assert not _check_version_in_range("1.1.0", "1.0.1", "1.0.3")
+    assert _check_version_in_range("1.0.3", "1.0.1", "1.0.3.post1")
```
harupy, line 657: Can we move these tests in a separate test function to make it easier to debug?

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +644,134 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+        def get_model_version_fn(module_name):
+            if module_name in module_name_to_version_map:
+                return module_name_to_version_map[module_name]
+            raise ValueError("Test error: unknown module name: " + module_name)
+
+        return get_model_version_fn
+
+    assert _is_autologging_integration_supported(
+        "fastai", gen_get_module_version_fn({"fastai": "1.0.60"})
+    )
```
harupy, line 669: Can we use parametrization to make it eaisier to debug test failures + make the code DRY?

https://docs.pytest.org/en/stable/parametrize.html#parametrizing-fixtures-and-test-functions

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,101 @@ def batch_metrics_logger(run_id):
...
+    return (
+        "    .. Note:: Autologging is known to be compatible with the following package versions: "
+        + required_pkg_versions
+        + ". Autologging may not succeed when used with package versions outside of this range..\n\n"
+    )
+
+
+def _log_warning_for_unsupported_integration(flavor_name):
+    _logger.warning(
+        "You are using an unsupported version of %s. If you encounter errors during autologging, try upgrading / downgrading %s to a supported version, or try upgrading MLflow.",
```
harupy, line 392: We could wrap a long string like this to avoid a lint error:

```python
a = (
    "longlonglonglonglonglonglonglong "
    "stringstringstringstringstringstring "
)
```

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +644,134 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+    assert not _is_autologging_integration_supported(
+        "fastai", gen_get_module_version_fn({"fastai": "1.0.50"})
+    )
```
harupy, line 672: What happens if we make the following change to `ml-package-versions.yml`?

```diff
fastai-1.x:
  package_info:
    pip_release: "fastai"

  models:
-   minimum: "1.0.60"
+   minimum: "1.0.30"
```

harupy: Can we mock `_version_file_json` and do something like below?

```python
with mock.patch.object("_version_file_json", new_value={"min": "0.24.0", "max": "0.25.0"}):
    with mock.patch("sklearn.__version__", new_value="0.23.0"):
        assert not is_autologging_supported("sklearn")

    ...
```

@dbczumar Any thoughts on this?

mlflow/ml-package-versions.yml
```diff
@@ -203,7 +203,7 @@ fastai-1.x:
     pip_release: "fastai"
 
   models:
-    minimum: "1.0.60"
+    minimum: "1.0.30"
```
harupy, line 206: Sorry I should have clarified. We don't need to change the minimum version for `fastai`.

tests/autologging/test_autologging_utils.py
```diff
@@ -657,8 +657,8 @@ def test_check_version_in_range():
 
 
 @pytest.mark.parametrize("flavor,module_version_dict,expected_result", [
-    ("fastai", {"fastai": "1.0.60"}, True),
-    ("fastai", {"fastai": "1.0.50"}, False),
+    ("fastai", {"fastai": "1.0.30"}, True),
+    ("fastai", {"fastai": "1.0.20"}, False),
```
harupy: With the current approach, we'd need to make update the tests when updating `mlflow/ml-package-versions.yml`.

WeichenXu123: No. minimal version is fixed. so I only test against patch version < min version and in range  cases.

tests/autologging/test_autologging_utils.py
WeichenXu123: Updated. Mock _version_file_json dict.

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,101 @@ def batch_metrics_logger(run_id):
...
+    "lightgbm": [("lightgbm", "lightgbm")],
+    "statsmodels": [("statsmodels", "statsmodels")],
+    "tensorflow": [("tensorflow", "tensorflow")],
+    "xgboost": [("xgboost", "xgboost")],
+    "sklearn": [("sklearn", "sklearn")],
+    "pytorch": [("torch", "pytorch"), ("pytorch_lightning", "pytorch-lightning")],
+}
+
+
+def _get_min_max_version_and_pip_release(json_key):
```
harupy, line 340: Can we rename `json_key` to a more descriptive name?

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +645,143 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+                log_warn_fn.assert_called_once_with("sklearn")
+        with mock.patch(log_warn_fn_name) as log_warn_fn:
+            mlflow.sklearn.autolog(disable_for_unsupported_versions=False)
+            log_warn_fn.assert_called_once_with("sklearn")
+
+        AUTOLOGGING_INTEGRATIONS.clear()
+        with mock.patch(log_warn_fn_name) as log_warn_fn:
+            mlflow.autolog(disable_for_unsupported_versions=True)
+            with pytest.raises(AssertionError):
+                log_warn_fn.assert_any_call("sklearn")
```
harupy, line 784: I think you can use `assert_not_called` here:
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called

WeichenXu123: No. It is possible being called, with other flavor_name.
because the universal autolog function will hook all flavor libraries.

harupy: btw what are you asserting here?

harupy: call_args_list

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +645,143 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+)
+@mock.patch(
+    "mlflow.utils.autologging_utils._module_version_info_dict", _module_version_info_dict_patch
+)
+def test_is_autologging_integration_supported(flavor, module_version_dict, expected_result):
+    with ExitStack() as stack:
+        for module_name in module_version_dict:
+            stack.enter_context(
+                mock.patch(module_name + ".__version__", module_version_dict[module_name])
+            )
```
harupy, line 737: @WeichenXu123 Could you add a comment here to clarify what we're doing here?

mlflow/ml-package-versions.yml
dbczumar: @WeichenXu123 Can we revert this?

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,97 @@ def batch_metrics_logger(run_id):
...
+_cross_tested_flavor_to_module_name_and_module_key = {
+    "fastai": [("fastai", "fastai-1.x")],
+    "gluon": [("mxnet", "gluon")],
+    "keras": [("keras", "keras")],
+    "lightgbm": [("lightgbm", "lightgbm")],
+    "statsmodels": [("statsmodels", "statsmodels")],
+    "tensorflow": [("tensorflow", "tensorflow")],
+    "xgboost": [("xgboost", "xgboost")],
+    "sklearn": [("sklearn", "sklearn")],
+    "pytorch": [("torch", "pytorch"), ("pytorch_lightning", "pytorch-lightning")],
```
dbczumar, line 336: @WeichenXu123 @harupy I think it's sufficient for us to only map `pytorch` to `pytorch-lightning`, since `torch` is a requirement for `pytorch-lightning` and, therefore, installing `pytorch-lightning` will pull in a compatible version of `torch`.

Accordingly, I think we can make this a map from string to tuple, rather than a map from string to a list of tuples. I also think we can drop the `autologging` subsection from the `pytorch` section of `ml-package-versions.yaml`: https://github.com/mlflow/mlflow/blob/ab6fa411a4237c01e19b3b9747179c85f6fd137e/ml-package-versions.yml#L35-L40

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,97 @@ def batch_metrics_logger(run_id):
...
+def _log_warning_for_unsupported_integration(flavor_name):
+    _logger.warning(
+        "You are using an unsupported version of %s. If you encounter errors during autologging, "
+        + "try upgrading / downgrading %s to a supported version, or try upgrading MLflow.",
+        flavor_name,
+        flavor_name,
+    )
+
+
+def check_and_log_warning_for_unsupported_integration(flavor_name):
```
dbczumar, line 394: Nit: I think this should be a module-"private" method since the other usage of it in `fluent.py` seems unnecessary (see comment below).

```suggestion
def _check_and_log_warning_for_unsupported_integration(flavor_name):
```

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,97 @@ def batch_metrics_logger(run_id):
...
+    )
+
+
+def _log_warning_for_unsupported_integration(flavor_name):
+    _logger.warning(
+        "You are using an unsupported version of %s. If you encounter errors during autologging, "
+        + "try upgrading / downgrading %s to a supported version, or try upgrading MLflow.",
+        flavor_name,
+        flavor_name,
+    )
```
dbczumar, line 391: Can we move this directly inside of `check_and_log_warning_for_unsupported_integration` since `_log_warning_for_unsupported_integration` is only used in one place and consists of a single function call? I don't think we need the extra method here.

mlflow/utils/autologging_utils.py
dbczumar, line 394: Can we add some docs for this method?

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,97 @@ def batch_metrics_logger(run_id):
...
+        return "``{min_ver}`` <= ``{pip_release}`` <= ``{max_ver}``".format(
+            min_ver=min_ver, pip_release=pip_release, max_ver=max_ver
+        )
+
+    required_pkg_versions = ",".join(
+        [
+            _gen_single_requirement(module_key)
+            for _, module_key in _cross_tested_flavor_to_module_name_and_module_key[flavor_name]
+        ]
+    )
```
dbczumar, line 375: In accordance with my comment below about us not needing to version check both `torch` and `pytorch-lightning` (`pytorch-lightning` is sufficient), I think we can refactor this so we only return a docstring for a single requirement.

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,97 @@ def batch_metrics_logger(run_id):
...
+
+
+def _get_min_max_version_and_pip_release(module_key):
+    min_version = _module_version_info_dict[module_key]["autologging"]["minimum"]
+    max_version = _module_version_info_dict[module_key]["autologging"]["maximum"]
+    pip_release = _module_version_info_dict[module_key]["package_info"]["pip_release"]
+    return min_version, max_version, pip_release
+
+
+def is_autologging_integration_supported(flavor_name):
```
dbczumar, line 347: Can we add a docstring for this method?

dbczumar, line 347: Nit: since this method is only consumed within `autologging_utils`, let's make it module-"private":

```suggestion
def _is_autologging_integration_supported(flavor_name):
```

mlflow/lightgbm.py
```diff
@@ -325,6 +326,9 @@ def autolog(
     :param exclusive: If ``True``, autologged content is not logged to user-created fluent runs.
                       If ``False``, autologged content is logged to the active fluent run,
                       which may be user-created.
+    :param disable_for_unsupported_versions: If True, disable autologging for versions of lightgbm
```
dbczumar, line 329: Nit: 
```suggestion
    :param disable_for_unsupported_versions: If ``True``, disable autologging for versions of lightgbm
```

mlflow/tracking/fluent.py
```diff
@@ -1392,6 +1397,7 @@ def setup_autologging(module):
             ] = True
             if not autologging_params.get("disable", False):
                 _logger.info("Autologging successfully enabled for %s.", module.__name__)
+            check_and_log_warning_for_unsupported_integration(autolog_fn.integration_name)
```
dbczumar, line 1400: I don't think we need this logic since we're calling `autolog_fn()` immediately above, which should log the same warning message.

For example, the following workflow produces two identical warning messages:

```
import mlflow
mlflow.autolog()

from sklearn import datasets

iris = datasets.load_iris()

logistic = LogisticRegression(solver="saga", max_iter=200, random_state=0)
logistic.fit(iris.data, iris.target)
```

```
2021/03/01 23:27:01 WARNING mlflow.utils.autologging_utils: You are using an unsupported version of sklearn. If you encounter errors during autologging, try upgrading / downgrading sklearn to a supported version, or try upgrading MLflow.
2021/03/01 23:27:02 INFO mlflow.tracking.fluent: Autologging successfully enabled for sklearn.
2021/03/01 23:27:02 WARNING mlflow.utils.autologging_utils: You are using an unsupported version of sklearn. If you encounter errors during autologging, try upgrading / downgrading sklearn to a supported version, or try upgrading MLflow.
2021/03/01 23:27:02 INFO mlflow.utils.autologging_utils: Created MLflow autologging run with ID '6baa8a55e05d47d2be5bc5c234edb27b', which will track hyperparameters, performance metrics, model artifacts, and lineage information for the current sklearn workflow
```

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -218,10 +218,16 @@ def test_autolog_obeys_disabled():
...
+    assert not get_autologging_config("sklearn", "disable")
     mlflow.sklearn.autolog(disable=True)
     assert get_autologging_config("sklearn", "disable")
 
+    mlflow.autolog(disable_for_unsupported_versions=False)
+    mlflow.utils.import_hooks.notify_module_loaded(sklearn)
+    assert not get_autologging_config("sklearn", "disable_for_unsupported_versions")
+    mlflow.sklearn.autolog(disable_for_unsupported_versions=True)
+    assert get_autologging_config("sklearn", "disable_for_unsupported_versions")
+
```
dbczumar, line 230: Cool! Can we also add the `disable_for_unsupported_versions` argument to the `test_universal_autolog_calls_specific_autologs_correctly` test case to ensure that it's propagated for all supported autologging frameworks?

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -218,10 +218,16 @@ def test_autolog_obeys_disabled():
 
     mlflow.autolog(disable=False)
     mlflow.utils.import_hooks.notify_module_loaded(sklearn)
-    assert not get_autologging_config("sklearn", "disable", False)
+    assert not get_autologging_config("sklearn", "disable")
     mlflow.sklearn.autolog(disable=True)
     assert get_autologging_config("sklearn", "disable")
 
+    mlflow.autolog(disable_for_unsupported_versions=False)
```
dbczumar, line 225: Can we also test the case where we explicitly invoke `mlflow.autolog(disable_for_unsupported_versions=True)`?

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +645,139 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+
+
+@mock.patch(
+    "mlflow.utils.autologging_utils._module_version_info_dict", _module_version_info_dict_patch
+)
+def test_disable_for_unsupported_versions_warning_sklearn_integration():
+    log_warn_fn_name = "mlflow.utils.autologging_utils._log_warning_for_unsupported_integration"
+
+    with mock.patch("sklearn.__version__", "0.20.3"):
+        AUTOLOGGING_INTEGRATIONS.clear()
```
dbczumar, line 750: Why do we need to clear `AUTOLOGGING_INTEGRATIONS` each time? I'd expect the warning message to fire even if a preexisting autologging configuration was present. Let me know if I'm missing something here.

mlflow/utils/autologging_utils.py
```diff
@@ -381,7 +483,19 @@ def autologging_is_disabled(flavor_name):
...
+        return True
+
+    if (
+        flavor_name in _cross_tested_flavor_to_module_name_and_module_key
+        and not is_autologging_integration_supported(flavor_name)
+    ):
+        if get_autologging_config(flavor_name, "disable_for_unsupported_versions", False):
+            return True
+        else:
+            return False
```
dbczumar, line 497:
```suggestion
        return get_autologging_config(flavor_name, "disable_for_unsupported_versions", False)
```

dbczumar (commented):
@WeichenXu123 Almost there! Nice work!

mlflow/utils/autologging_utils.py
WeichenXu123: This is for unit test. I patch this method and test whether it is called in expected way.

mlflow/utils/autologging_utils.py
harupy: @dbczumar Thanks for pointing this out. Agree with dropping the autologging from `pytorch`. Let me just open a PR to address this.

mlflow/utils/autologging_utils.py
WeichenXu123: Withdraw above comments. 
We can patch `_logger.warning` and then check input args

tests/autologging/test_autologging_utils.py
WeichenXu123: According to the setup_autologging code, 
if we called  a specific lib autolog first, then call universal autolog, then the latter universal autolog won't change the specific autolog config (this case the specific lib `autolog_fn(**autologging_params)` won't be called so won't fire the warning):
```
            # Only call integration's autolog function with `mlflow.autolog` configs
            # if the integration's autolog function has not already been called by the user.
            # Logic is as follows:
            # - if a previous_config exists, that means either `mlflow.autolog` or
            #   `mlflow.integration.autolog` was called.
            # - if the config contains `CONF_KEY_IS_GLOBALLY_CONFIGURED`, the configuration
            #   was set by `mlflow.autolog`, and so we can safely call `autolog_fn` with
            #   `autologging_params`.
            # - if the config doesn't contain this key, the configuration was set by an
            #   `mlflow.integration.autolog` call, so we should not call `autolog_fn` with
            #   new configs.
```

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +645,130 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+        with mock.patch(log_warn_fn_name) as log_warn_fn:
+            mlflow.autolog(disable_for_unsupported_versions=True)
+            assert all(not is_sklearn_warning_fired(args) for args in log_warn_fn.call_args_list)
+        with mock.patch(log_warn_fn_name) as log_warn_fn:
+            mlflow.autolog(disable_for_unsupported_versions=False)
+            assert any(is_sklearn_warning_fired(args) for args in log_warn_fn.call_args_list)
+
+        with mock.patch(log_warn_fn_name) as log_warn_fn:
+            mlflow.sklearn.autolog(disable_for_unsupported_versions=True)
+            assert len(log_warn_fn.call_args_list) == 0
```
harupy, line 770:
```suggestion
            log_warn_fn.assert_not_called()
```

Can we use `assert_not_called` here?

tests/autologging/test_autologging_utils.py
```diff
@@ -642,3 +645,130 @@ def test_autologging_event_logger_default_impl_warns_for_log_autolog_called_with
...
+        with mock.patch(log_warn_fn_name) as log_warn_fn:
+            mlflow.sklearn.autolog(disable_for_unsupported_versions=False)
+            assert len(log_warn_fn.call_args_list) == 1 and \
+                is_sklearn_warning_fired(log_warn_fn.call_args_list[0])
```
harupy, line 774:
```suggestion
            assert log_warn_fn.call_count == 1 and is_sklearn_warning_fired(log_warn_fn.call_args)
```

- https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args
- https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count

tests/autologging/test_autologging_utils.py
dbczumar: Got it :)! Thanks!

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,84 @@ def batch_metrics_logger(run_id):
...
+def _get_min_max_version_and_pip_release(module_key):
+    min_version = _module_version_info_dict[module_key]["autologging"]["minimum"]
+    max_version = _module_version_info_dict[module_key]["autologging"]["maximum"]
+    pip_release = _module_version_info_dict[module_key]["package_info"]["pip_release"]
+    return min_version, max_version, pip_release
+
+
+def _is_autologging_integration_supported(flavor_name):
+    """
+    check whether the flavor package version is compatible with mlflow,
```
dbczumar, line 349:
```suggestion
    :return: True if the flavor's associated package version is compatible with mlflow, False otherwise.
```

mlflow/utils/autologging_utils.py
```diff
@@ -306,6 +310,84 @@ def batch_metrics_logger(run_id):
...
+    """
+    module_name, module_key = _cross_tested_flavor_to_module_name_and_module_key[flavor_name]
+    actual_version = importlib.import_module(module_name).__version__
+    min_version, max_version, _ = _get_min_max_version_and_pip_release(module_key)
+    return _check_version_in_range(actual_version, min_version, max_version)
+
+
+def gen_autologging_package_version_requirements_doc(flavor_name):
```
dbczumar, line 357: Nit: 

```suggestion
def _gen_autologging_package_version_requirements_doc(flavor_name):
```

Can we add a docstring too?

mlflow/tracking/fluent.py
```diff
@@ -1225,6 +1225,7 @@ def autolog(
     log_models=True,
     disable=False,
     exclusive=False,
+    disable_for_unsupported_versions=False,
```
dbczumar, line 1228: @WeichenXu123 Looks like we still print the `Autologging successfully enabled for ...` message even when `disable_for_unsupported_versions=True` and an unsupported version is used, due to logic present here: https://github.com/mlflow/mlflow/blob/ab6fa411a4237c01e19b3b9747179c85f6fd137e/mlflow/tracking/fluent.py#L1393-L1394. Rather than checking for `autologging_params.get("disable", False)`, on line 1393, can we condition the message based on `autologging_utils.autologging_is_disabled`?

dbczumar (approved):
LGTM! Awesome work @WeichenXu123 !

PR #4153, opened 2021-03-01
Move to new SHAP Serialization API

vivekchettiar:
Signed-off-by: Vivek Chettiar <vivekrchettiar@gmail.com>

Expected Change to the SHAP serialization API, Making changes to use the new API in mlflow.

## How is this patch tested?

Ran exsisting SHAP logging tests

## Release Notes

Internal change to SHAP serialization API Change

There is minor change to the SHAP serialization API call expected to be released soon (0.38.2). 

Details:
To indicate that SHAP shouldn't serialize model itself

**Old API:**
    explainer.model.save = None
    explainer.save()

**New API:**
    explainer.save(model_saver = False)

Since, mlflow makes use of this call it needs to be updated.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/shap.py
+++ b/mlflow/shap.py
@@ -415,13 +415,14 @@ def save_explainer(
 
     underlying_model_flavor = None
     underlying_model_path = None
+    model_saver = None
 
     # saving the underlying model if required
     if serialize_model_using_mlflow:
         underlying_model_flavor = get_underlying_model_flavor(explainer.model)
 
         if underlying_model_flavor != _UNKNOWN_MODEL_FLAVOR:
-            explainer.model.save = None  # this prevents SHAP from serializing the underlying model
+            model_saver = False  # this prevents SHAP from serializing the underlying model
             underlying_model_path = os.path.join(path, _UNDERLYING_MODEL_SUBPATH)
 
         if underlying_model_flavor == mlflow.sklearn.FLAVOR_NAME:
@@ -433,7 +434,7 @@ def save_explainer(
     explainer_data_subpath = "explainer.shap"
     explainer_output_path = os.path.join(path, explainer_data_subpath)
     with open(explainer_output_path, "wb") as explainer_output_file_handle:
-        explainer.save(explainer_output_file_handle)
+        explainer.save(explainer_output_file_handle, model_saver=model_saver)
 
     conda_env_subpath = "conda.yaml"
     if conda_env is None:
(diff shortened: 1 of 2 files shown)


mlflow/shap.py
```diff
@@ -574,7 +575,6 @@ def _load_explainer(explainer_file, model=None):
     import shap
 
     def inject_model_loader(in_file):
-        pickle.load(in_file)  # No-Op to move file pointer forward
```
harupy, line 577: This seems like a breaking change, right? Can we still load the explainer logged by `mlflow <= 1.14.0`?

mlflow/shap.py
```diff
@@ -433,7 +434,7 @@ def save_explainer(
     explainer_data_subpath = "explainer.shap"
     explainer_output_path = os.path.join(path, explainer_data_subpath)
     with open(explainer_output_path, "wb") as explainer_output_file_handle:
-        explainer.save(explainer_output_file_handle)
+        explainer.save(explainer_output_file_handle, model_saver=model_saver)
```
harupy, line 437: When will the new SHAP Serialization API be relased? Next release?

mlflow/shap.py
vivekchettiar: No, that wouldn't be possible. Even SHAP (v0.39.0) will not be able to load the explainers logged by SHAP (v0.38.0). There were changes to the underlying serialization process. Hence this consequently affects mlflow.

mlflow/shap.py
vivekchettiar: It was released yesterday https://github.com/slundberg/shap/releases/tag/v0.39.0

vivekchettiar: This change to the API has back-compat for the API but not for the actual artifacts generated by saving.

mlflow/shap.py
akshaya-a: @vivekchettiar  Are we at least emitting enough version metadata to detect this situation and inform users how they can load their older explainers? What are we doing to protect against another break moving forward?

vivekchettiar: @akshaya-a thanks for bringing this up, part of update to SHAP serialization was to track the version of the serialization code that generated the explainers. When SHAP encounters a saved explainer from a older version, it'll try best-effort deserialization otherwise it'll inform the user about the incompatibility and print our the version that generated the file. Also, we're adding downstream integration tests in SHAP to check if there is a breaking change in the future.

harupy (approved):
LGTM

PR #4173, opened 2021-03-12
Add silent option for autologging integrations

dbczumar:

**Much of the PR diff is docstrings & test cases, indentation**

This PR introduces a `silent` mode for MLflow's `autolog()` APIs, allowing users to disable MLflow warnings and logging statements during autologging setup and training. The implementation of `silent` mode takes care to ensure that warnings and loggings are suppressed in multithreaded environments where multiple autologged training sessions may be executing concurrently (for additional details about why this is necessary, ask @dbczumar).

## How is this patch tested?
- Unit and integration test cases for the new `silent` mode introduced by this PR
- Existing unit / integration tests for autologging to guard against regressions

## Release Notes

Add a `silent` option to `autolog()` APIs, allowing users to disable MLflow warnings and logging statements during autologging setup and training.

Components 

Interface 

Language 

Integrations


mlflow/utils/autologging_utils.py
```diff
@@ -1,1360 +1,33 @@
-import os
```
dbczumar, line 1: For modularity, the previous contents of `autologging_utils` have been moved to `mlflow/utils/autologging/__init__`.   `autologging_utils` now imports select contents from the new `mlflow/utils/autologging/__init__` file to preserve compatibility with other modules that import from `mlfow.utils.autologging_utils`.

I will file a follow-up PR that drops this `autologging_utils` file and changes existing imports to import from `mlflow.utils.autologging` instead of `mlflow.utils.autologging_utils`.

mlflow/utils/autologging/logging_and_warnings.py
```diff
@@ -0,0 +1,265 @@
...
+
+    - If a silent autologging session is run in parallel with a non-silent autologging session,
+      MLflow event logging statements and warnings from the non-silent session may be suppressed.
+
+    - If `warnings.showwarning` is unexpectedly re-assigned during the execution of a user's
+      original / underlying ML code within an autologging session, warnings may not be
+      properly suppressed.
+
+    - If MLflow's autologging implementation code makes calls to multithreaded APIs that raise warnings
+      from a different thread, these warnings may not be properly suppressed.
```
dbczumar, line 206: I'd recommend starting with this comment for context about threading behavior. Please let me know if you have any questions!

tests/autologging/fixtures.py
```diff
@@ -1,8 +1,26 @@
...
+def patch_destination():
+    class PatchObj:
+        def __init__(self):
+            self.fn_call_count = 0
+
+        def fn(self, *args, **kwargs):
+            self.fn_call_count += 1
+            return PATCH_DESTINATION_FN_DEFAULT_RESULT
+
+    return PatchObj()
```
dbczumar, line 23: Moved this out of `tests/autologging/test_autologging_safety_unit.py` for reusability across test modules.

tests/autologging/fixtures.py
```diff
@@ -31,3 +49,30 @@ def test_mode_on():
             os.environ[_AUTOLOGGING_TEST_MODE_ENV_VAR] = prev_env_var_value
         else:
             del os.environ[_AUTOLOGGING_TEST_MODE_ENV_VAR]
+
+
+@pytest.fixture(autouse=True)
+def reset_stderr():
```
dbczumar, line 55: Since some of our tests reassign `stderr` to a test stream to ensure that no content is written during silent mode, this fixture is used to reset `stderr` at the end of each test.

tests/autologging/fixtures.py
```diff
@@ -31,3 +49,30 @@ def test_mode_on():
...
+    prev_stderr = sys.stderr
+    yield
+    sys.stderr = prev_stderr
+
+
+@pytest.fixture(autouse=True)
```
dbczumar, line 61: Since some of our tests disable MLflow event logging to test silent mode, this fixture is used to re-enable MLflow event logging at the end of each test.

mlflow/utils/autologging/__init__.py
```diff
@@ -1023,219 +1048,207 @@ def safe_patch_function(*args, **kwargs):
...
+                            original_has_been_called = True
+
+                            nonlocal original_result
+                            # Show all non-MLflow warnings as normal (i.e. not as event logs) during
+                            # original function execution, even if silent mode is enabled
+                            # (`silent=True`), since these warnings originate from the ML framework
+                            # or one of its dependencies and are likely relevant to the caller
+                            with set_non_mlflow_warnings_behavior_for_current_thread(
+                                disable_warnings=False, reroute_warnings=False,
+                            ):
```
dbczumar, line 1155: Small addition of new logic

mlflow/utils/autologging/__init__.py
```diff
@@ -1023,219 +1048,207 @@ def safe_patch_function(*args, **kwargs):
...
-        original = gorilla.get_original_attribute(destination, function_name)
-
-        # If the autologging integration associated with this patch is disabled,
-        # call the original function and return
-        if autologging_is_disabled(autologging_integration):
-            return original(*args, **kwargs)
-
-        # Whether or not to exclude auto-autologged content from content explicitly logged via
-        # `mlflow.start_run()`
-        exclusive = get_autologging_config(autologging_integration, "exclusive", False)
```
dbczumar, line 1038: This logic and the logic below it was right-intended and wrapped in the `set_mlflow_events_and_warnings_behavior_globally` / `set_non_mlflow_warnings_behavior_for_current_thread` context managers. The logic itself remains unchanged (with the exception of 3 lines here: https://github.com/mlflow/mlflow/pull/4173/files#r600105840)

mlflow/utils/autologging/__init__.py
```diff
@@ -1023,219 +1048,207 @@ def safe_patch_function(*args, **kwargs):
...
+                if original_has_been_called:
+                    return original_result
+                else:
+                    return original(*args, **kwargs)
 
     _wrap_patch(destination, function_name, safe_patch_function)
 
 
-@contextmanager
-def _augment_mlflow_warnings(autologging_integration):
```
dbczumar, line 1192: Replaced with a `reroute_warnings` on option on each of `set_mlflow_events_and_warnings_behavior_globally` / `set_non_mlflow_warnings_behavior_for_current_thread` (introduced in `mlflow/utils/autologging/logging_and_warnings.py`)

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -52,25 +53,12 @@ def turn_test_mode_off_by_default(test_mode_off):
     """
 
 
-@pytest.fixture
```
dbczumar, line 55: Moved to `tests/autologging/fixtures.py` for reuse across multiple modules

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -736,76 +724,6 @@ def patch_impl(original, *args, **kwargs):
     assert original_error.exception == exc_to_raise
 
 
-def test_safe_patch_augments_mlflow_warnings_and_preserves_others(
```
dbczumar, line 739: Removed and replaced with `test_autologging_warnings_are_redirected_as_expected` in `test_autologging_behaviors_unit.py`

tests/autologging/fixtures.py
```diff
@@ -1,10 +1,28 @@
...
 
+import mlflow.utils.logging_utils as logging_utils
 from mlflow.utils.autologging import _is_testing, _AUTOLOGGING_TEST_MODE_ENV_VAR
 
 
+PATCH_DESTINATION_FN_DEFAULT_RESULT = "original_result"
+
+
+@pytest.fixture
+def patch_destination():
```
dbczumar, line 14: Moved from `tests/autologging/test_autologging_safety_unit.py` for reuse across multiple test suites

conftest.py
```diff
@@ -93,8 +93,9 @@ def pytest_ignore_collect(path, config):
             # this test is included here because it imports many big libraries like tf, keras, etc
             "tests/tracking/fluent/test_fluent_autolog.py",
             # cross flavor autologging related tests.
-            "tests/autologging/test_autologging_safety_integration.py",
             "tests/autologging/test_autologging_safety_unit.py",
+            "tests/autologging/test_autologging_behaviors_unit.py",
+            "tests/autologging/test_autologging_behaviors_integration.py",
```
dbczumar, line 98: Add a new `tests/autologging/test_autologging_behaviors_unit.py` test suite and rename `tests/autologging/test_autologging_safety_integration.py` to `tests/autologging/test_autologging_behaviors_integration.py`

mlflow/fastai.py
```diff
@@ -381,7 +381,11 @@ def load_model(model_uri):
 @experimental
 @autologging_integration(FLAVOR_NAME)
 def autolog(
-    log_models=True, disable=False, exclusive=False, disable_for_unsupported_versions=False
+    log_models=True,
+    disable=False,
+    exclusive=False,
+    disable_for_unsupported_versions=False,
+    silent=False,
```
dbczumar, line 388: Add a `silent` param with docstring description to each framework-specific autologging function (e.g. `mlflow.fastai.autolog()`, `mlflow.sklearn.autolog()`). Also add the `silent` parameter to the universal `mlflow.autolog()` function.

mlflow/utils/autologging/logging_and_warnings.py
```diff
@@ -0,0 +1,287 @@
+import os
```
dbczumar, line 1: Introduce utilities for rerouting and silencing MLflow & non-MLflow warnings, providing support for multithreaded contexts to ensure that autologging warnings are not leaked.

mlflow/utils/autologging/__init__.py
```diff
@@ -422,18 +430,35 @@ def autolog(*args, **kwargs):
...
+            with set_mlflow_events_and_warnings_behavior_globally(
+                # MLflow warnings emitted during autologging setup / enablement are likely
+                # actionable and relevant to the user, so they should be emitted as normal
+                # when `silent=False`
+                reroute_warnings=False,
+                disable_event_logs=is_silent_mode,
+                disable_warnings=is_silent_mode,
+            ), set_non_mlflow_warnings_behavior_for_current_thread(
+                reroute_warnings=True, disable_warnings=is_silent_mode,
+            ):
```
dbczumar, line 446: For reference, https://docs.python.org/3/howto/logging.html#when-to-use-logging indicates that `warnings.warn()` should be used for avoidable issues that the client application / caller should address by modifying behaviors, while `logger.warning()` should be used to inform users of unexpected errors.

When enabling MLflow autologging, MLflow warnings may be emitted to inform callers of an issue. These are actionable & should be represented as warnings. non-MLflow warnings emitted during enablement are a byproduct of the MLflow autologging implementation and are therefore not actionable for the user, so we reroute them to `logger.warning()`.

mlflow/utils/autologging/__init__.py
```diff
@@ -1023,219 +1048,207 @@ def safe_patch_function(*args, **kwargs):
...
+            disable_warnings=is_silent_mode,
+        ), set_non_mlflow_warnings_behavior_for_current_thread(
+            # non-MLflow Warnings emitted during the autologging preamble (before the original /
+            # underlying ML function is called) and postamble (after the original / underlying ML
+            # function is called) are likely not actionable and result from the autologging
+            # implementation invoking an API from a dependent library. Accordingly, we reroute these
+            # warnings to the MLflow event logger with level WARNING
+            reroute_warnings=True,
+            disable_warnings=is_silent_mode,
+        ):
```
dbczumar, line 1078: For reference, https://docs.python.org/3/howto/logging.html#when-to-use-logging indicates that `warnings.warn()` should be used for avoidable issues that the client application / caller should address by modifying behaviors, while `logger.warning()` should be used to inform users of unexpected errors.

During an MLflow autologging session, MLflow warnings are not actionable as they are a byproduct of the autologging implementation. non-MLflow warnings emitted during the patch function preamble (i.e. before execution of the original / underlying ML function) and postamble (i.e. after execution of the original / underlying ML function) are also a byproduct of the autologging implementation and are not actionable. Accordingly, we reroute both of these to `logger.warning()`.

tests/autologging/fixtures.py
```diff
@@ -31,3 +49,30 @@ def test_mode_on():
...
+@pytest.fixture(autouse=True)
+def reset_stderr():
+    prev_stderr = sys.stderr
+    yield
+    sys.stderr = prev_stderr
+
+
+@pytest.fixture(autouse=True)
+def reset_logging_enablement():
+    yield
```
jinzhang21, line 63: What does this line do? I don't have context how this function is used.

tests/autologging/fixtures.py
```diff
@@ -31,3 +49,30 @@ def test_mode_on():
...
+    logging_utils.enable_logging()
+
+
+class TestStream:
```
jinzhang21, line 67: This works but have you considered standard options like io.StringIO? Inheriting from io.IOBase and override the class method there could also indicate a more explicit intent.

tests/autologging/test_autologging_behaviors_unit.py
```diff
@@ -0,0 +1,237 @@
...
+    return logging.getLogger(mlflow.__name__)
+
+
+@pytest.fixture
+def autolog_function(patch_destination, logger):
+    def original_impl():
+        # Sleep during the original function implementation to increase the likelihood of
+        # overlapping session stages (i.e. simultaneous preamble / postamble / original function
+        # execution states across autologging sessions) during multithreaded execution
+        time.sleep(0.01)
```
jinzhang21, line 33: Interesting idea. Maybe also state the reason for the choice of 0.01s?

tests/autologging/test_autologging_behaviors_unit.py
```diff
@@ -0,0 +1,237 @@
...
+        original()
+        raise Exception("postamble error")
+
+    @autologging_integration("test_integration")
+    def test_autolog(disable=False, silent=False):
+        safe_patch("test_integration", patch_destination, "fn", patch_impl)
+        raise Exception("enablement error")
+
+    def parallel_fn():
+        # Sleep for a random interval to increase the likelihood of overlapping session stages
```
jinzhang21, line 221: indicate that the sleep value is between 0 and 1s.

jinzhang21 (approved):
Overall LGTM. High level comment: if in the future there's a request on tuning the level of logging to be silenced / spitted-out, a boolean might no longer be sufficient:)

tests/autologging/test_autologging_behaviors_integration.py
```diff
@@ -169,3 +176,65 @@ def mlflow_autolog():
...
+
+    assert all([e.result() is True for e in executions])
+    assert stream.content is None
+    # Verify that `warnings.showwarning` was restored to its original value after training
+    # and that MLflow event logs are enabled
+    assert warnings.showwarning == og_showwarning
+    logger.info("verify that event logs are enabled")
+    assert "verify that event logs are enabled" in stream.content
+
+    mlflow.sklearn.autolog(silent=False, log_input_examples=True)
```
smurching, line 229: Nice, could we also add a test that setting different values of `silent` across different per-framework autologging integrations doesn't interfere with each other? Maybe actually in the unit tests in `tests/autologging/test_autologging_behaviors_unit.py`? e.g. to cover the (probably rare) case where the user calls something like `mlflow.sklearn.autolog(silent=True)` and `mlflow.spark.autolog(silent=False)`?

mlflow/utils/autologging/logging_and_warnings.py
```diff
@@ -0,0 +1,287 @@
...
+import warnings
+from contextlib import contextmanager
+from pathlib import Path
+from threading import RLock, get_ident as get_current_thread_id
```
smurching, line 5: Was wondering about the performance of `get_current_thread_id` since we call it every time a warning is logged & @harupy previously found that other Python introspection methods (e.g. getting the current call stack) were rather slow e.g. took ~1ms - seems like it's super fast:

```
>>> def benchmark(n):
...   s = time.time()
...   for i in range(n):
...     get_current_thread_id()
...   print("Took %s sec" % (time.time() - s))
... 
>>> benchmark(1000)
Took 0.00011706352233886719 sec
>>> benchmark(10000)
Took 0.0009729862213134766 sec
>>> benchmark(100000)
Took 0.008674144744873047 sec
```

In general seems reasonable that logging a warning (writing to some non-memory output stream) would be quite a bit slower than the in-memory checks we do to determine whether to log the warning

smurching (commented):
Had one comment asking about a test, LGTM after it's addressed

tests/autologging/test_autologging_behaviors_integration.py
dbczumar: Absolutely! Added a `test_silent_mode_operates_independently_across_integrations` test case to `test_autologging_behaviors_unit.py`.

dbczumar (commented):
@smurching @jinzhang21 Thanks so much for the comments / feedback! I've addressed them & plan to merge once tests pass.

smurching (approved):
LGTM thanks @dbczumar for the updates!

PR #4174, opened 2021-03-12
[ML-14300] Add a script that updates `ml-package-versions.yml`

harupy:

Adds a script that update `mlflow/ml-package-versions.yml`.

## How is this patch tested?

(Details)

Components 

Interface 

Language 

Integrations

### DIFF
--- a/dev/update_ml_package_versions.py
+++ b/dev/update_ml_package_versions.py
@@ -1,0 +1,148 @@
+"""
+A script to update the maximum package versions in 'mlflow/ml-package-versions.yml'.
+
+# Prerequisites:
+$ pip install packaging pyyaml
+
+# How to run (make sure you're in the repository root):
+$ python dev/update_ml_package_versions.py
+"""
+import argparse
+import json
+from packaging.version import Version
+import re
+import sys
+import urllib.request
+import yaml
+
+
+def read_file(path):
+    with open(path) as f:
+        return f.read()
+
+
+def save_file(src, path):
+    with open(path, "w") as f:
+        f.write(src)
+
+
+def get_package_versions(package_name):
+    url = "https://pypi.python.org/pypi/{}/json".format(package_name)
+    with urllib.request.urlopen(url) as res:
+        data = json.load(res)
+
+    def is_dev_or_pre_release(version_str):
+        v = Version(version_str)
+        return v.is_devrelease or v.is_prerelease
+
+    return [
+        version
+        for version, dist_files in data["releases"].items()
+        if len(dist_files) > 0 and not is_dev_or_pre_release(version)
+    ]
+
+
+def get_latest_version(candidates):
+    return sorted(candidates, key=Version, reverse=True)[0]
+
+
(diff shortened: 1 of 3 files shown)


mlflow/ml-package-versions.yml
```diff
@@ -212,14 +212,14 @@ fastai-1.x:
...
-    maximum: "1.0.61"
+    maximum: "1.0.61" # IGNORE_UPDATE: 2.x is not supported yet
     requirements: ["scikit-learn"]
     run: |
       pytest tests/fastai/test_fastai_model_export.py --large
 
   autologging:
     minimum: "1.0.60"
-    maximum: "1.0.61"
+    maximum: "1.0.61" # IGNORE_UPDATE: 2.x is not supported yet
```
WeichenXu123, line 222: Why not we add the IGNORE_UPDATE config by a field like:

autologging:
    minimum: "1.0.60"
    maximum: "1.0.61"
    ignore_update: "2.x"

the comment way looks trigger when parsing.

dev/update_ml_package_versions.py
```diff
@@ -0,0 +1,148 @@
...
+        key=re.escape(key), category=category
+    )
+    # Matches the following pattern:
+    #
+    # <key>:
+    #   ...
+    #   <category>:
+    #     ...
+    #     maximum: "1.2.3" # IGNORE_UPDATE
+    return re.search(pattern, src, flags=re.DOTALL).group(1) is not None
```
WeichenXu123, line 103: So the logic is, if detecting "# IGNORE_UPDATE xxxx" exists, then return True ? otherwise return False ?

WeichenXu123 (approved):

mlflow/ml-package-versions.yml
harupy: I agree. Adding a new field makes more sense. Let me fix.

dbczumar (changes requested):
@harupy Looks awesome generally! I noticed that this script attempts to update the max tested version beyond the current major version. For example, for `fastai v1`, the max tested version is `1.0.61`, but running this PR attempts to set it to a `2.x` version. Can we fix this before merging?

mlflow/ml-package-versions.yml
```diff
@@ -211,13 +211,16 @@ fastai-1.x:
     pip_release: "fastai"
 
   models:
+    # Avoid updating the maximum version to 2.x which isn't supported yet
+    freeze: True
```
harupy, line 215: @dbczumar Thanks for [this comment](https://github.com/mlflow/mlflow/pull/4174#pullrequestreview-654748537), I  added this field `freeze`. When this field is present, the script doesn't skips updating the maximum version. WDYT?

WeichenXu123: Nit: call the field "frozen" ?

WeichenXu123 (approved):

dbczumar (approved):
LGTM! Thanks @harupy !

harupy:
@dbczumar Thanks!

PR #4194, opened 2021-03-22
TensorFlow autologging: change every_n_iter default to 1, always log last epoch

dbczumar:

Currently, the default behavior for TensorFlow autologging is to only log metrics every 100 epochs. Many TensorFlow training sessions (especially tf.keras sessions) last far fewer than 100 epochs, resulting in valuable metrics being omitted. This behavior is also confusing to users who may not have read the API docs / defaults carefully.

Accordingly, this PR adjusts the default value for `every_n_iter` to `1`. It also ensures that metrics for the last training epoch are always logged, regardless of the epoch number. We are not concerned about excessive metric logging latency because: 1. the default behavior in TensorFlow for estimator models is to record TensorBoard summary metrics every 100 epochs already (i.e. this change is a no-op for the default metric logging behavior w.r.t TensorFlow estimators) 2. we've integrated batching capabilities for metrics logging (`BatchMetricsLogger`) to minimize the impact on training latency.

## How is this patch tested?

TF2 autologging unit / integration tests

## Release Notes

TensorFlow autologging: the default value of `every_n_iter` has been changed to `1`, and metrics will always be logged for the last training iteration.

Components 

Interface 

Language 

Integrations


tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -700,14 +723,6 @@ def test_tf_estimator_autolog_model_can_load_from_artifact(tf_estimator_random_d
     mlflow.tensorflow.load_model("runs:/" + tf_estimator_random_data_run.info.run_id + "/model")
 
 
-@pytest.mark.large
```
dbczumar, line 703: This test is no longer necessary since autologging config updates are implemented & tested via higher-level `autologging_utils` APIs. Removing it. This was failing due to line 708 after the changes to always log metrics for the last epoch were implemented.

mlflow/tensorflow.py
```diff
@@ -619,12 +620,41 @@ def _flush_queue():
...
+            # For thread safety and to avoid modifying a list while iterating over it, we record a
+            # separate list of the items being flushed and remove each one from the metric queue,
+            # rather than clearing the metric queue or reassigning it (clearing / reassigning is
+            # dangerous because we don't block threads from adding to the queue while a flush is
+            # in progress)
+            snapshot = _metric_queue[:]
+            for item in snapshot:
+                _metric_queue.remove(item)
+
+            # NB: Most TensorFlow APIs use one-indexing for epochs, while tf.Keras
```
apurva-koti, line 633: bruh...

mlflow/tensorflow.py
```diff
@@ -619,12 +620,41 @@ def _flush_queue():
...
+            metrics_to_log_immediately = [
+                item for item in snapshot if (item[1].step - 1) % _LOG_EVERY_N_STEPS == 0
+            ]
+
+            metrics_by_run = _assoc_list_to_map(metrics_to_log_immediately)
+            for run_id, metrics in metrics_by_run.items():
+                try_mlflow_log(client.log_batch, run_id, metrics=metrics, params=[], tags=[])
+
+            # If metrics for the last training epoch were not logged due to filtering on
+            # `every_n_iter`, log them now
```
apurva-koti, line 646: I'm a little confused, why exactly is this necessary? If they were not logged due to filtering on `every_n_iter`, isn't that the purpose of that filter?

dbczumar: That's a fair point. I was concerned that users might miss metrics for the final training epoch, which is more likely to produce the best results compared to other randomly selected epochs. However, because we're defaulting to `every_n_iter=1`, it seems acceptable to drop metrics for the last epoch for cases where `every_n_iter != 1` since users are opting into this behavior & likely know what they're doing. I've removed the logic that was introduced to unconditionally log metrics for the last epoch, which simplifies things quite a bit. Thanks!

apurva-koti: Sounds great!

mlflow/tensorflow.py
```diff
@@ -619,12 +619,19 @@ def _flush_queue():
...
+            # rather than clearing the metric queue or reassigning it (clearing / reassigning is
+            # dangerous because we don't block threads from adding to the queue while a flush is
+            # in progress)
+            snapshot = _metric_queue[:]
+            for item in snapshot:
+                _metric_queue.remove(item)
+
+            metrics_by_run = _assoc_list_to_map(snapshot)
+            for run_id, metrics in metrics_by_run.items():
+                try_mlflow_log(client.log_batch, run_id, metrics=metrics, params=[], tags=[])
```
apurva-koti, line 634: I am pretty sure this was the behavior before anyway, but this makes it a little more clear that if `try_mlflow_log` fails to log a metric x, that metric won't be in the queue anymore due to line 630 and so will never be logged.
Not blocking though since this was already the case.

apurva-koti (approved):
Thanks for the changes, looks a lot cleaner :)

mlflow/tensorflow.py
dbczumar: Yep! That's expected / is consistent with past behavior. Thanks for calling this out!

PR #4195, opened 2021-03-22
Bug fix for TensorFlow autologging: do not mutate user-specified callbacks list

dbczumar:

TensorFlow autologging passes new callbacks to the `fit()` / `fit_generator()` function. If preexisting user-defined callbacks already exist, these new callbacks are added to the user-specified ones. Unfortunately, the current implementation mutates the list of user-defined callbacks, rather than creating a copy of the list and modifying it. As a result, when users call `fit()` / `fit_generator()` multiple times, they end up with a list of multiple MLflow-defined tf.keras callbacks for metric logging.

This becomes problematic in hyperopt use cases. In the following example, sklearn's GridSearch API is used with a tf.keras model. Each call to `fit()` in the grid search attaches an MLflow metric logging callback for all preceding `fit()` invocations. Each callback logs metrics to the run corresponding to its initial `fit()` call. As a result, all runs end up with the same metrics (the ones corresponding to the final `fit()` call). Here's some example code:

```
import tensorflow as tf
from tensorflow.keras.wrappers.scikit_learn import KerasClassifier
from tensorflow.keras import layers, regularizers, metrics
from tensorflow.keras.optimizers import Adam
from tensorflow.python.keras.metrics import AUC, Precision, Recall

from sklearn.model_selection import GridSearchCV, RandomizedSearchCV
from sklearn.preprocessing import LabelEncoder, OneHotEncoder

import mlflow 
import mlflow.tensorflow

import warnings
warnings.filterwarnings("ignore")

def load_data():
  import pandas as pd
  X = pd.read_csv("x.csv").drop("Unnamed: 0", axis=1)
  y = pd.read_csv("y.csv")["bad_loan"]
  return X, y

def encode_inputs(X):
  ohe = OneHotEncoder()
  ohe.fit(X)
  X = ohe.transform(X)
  return X

def encode_labels(y): 
  enc = LabelEncoder()
  enc.fit(y)
  return enc.transform(y)

def build_model(optimizer='rmsprop', init='glorot_uniform'):
  model = Sequential()
  model.add(Dense(17, input_dim=24064, kernel_initializer=init, 
                  kernel_regularizer=regularizers.l2(1e-2), activation='relu'))
  model.add(Dense(1, activation='sigmoid'))
  
  model.compile(optimizer=optimizer,
                loss='binary_crossentropy',
                metrics=['accuracy'])
  
  return model

X, y = load_data()
X_enc = encode_inputs(X)
y_enc = encode_labels(y)

mlflow.tensorflow.autolog(every_n_iter=1)

model = KerasClassifier(build_fn=build_model, verbose=2)

# Test with small/few parameters
optimizers = ['rmsprop']
inits= ['glorot_uniform', 'normal']
epochs = [3, 5]  

experiment_log_dir = "foobazbiz"
tensorboard_callback = tf.keras.callbacks.TensorBoard(log_dir=experiment_log_dir)

fit_params = dict(callbacks=[tensorboard_callback])
# fit_params = {} 

param_grid = dict(optimizer=optimizers, epochs=epochs, init=inits)
grid = GridSearchCV(estimator=model, param_grid=param_grid, cv=3)

grid_result = grid.fit(X_enc, y_enc, **fit_params)

mlflow.end_run()
```

(please ask me for datasets if you want to run this yourself)

Before this PR, the results are as follows (note that all runs share the same metrics): 

Screen Shot 2021-03-22 at 11 55 51 AM

After this PR, the results are as follows (note that each run has different metrics, as expected):

Screen Shot 2021-03-22 at 11 57 33 AM

## How is this patch tested?

New unit tests + manual gridsearch test case described immediately above.

## Release Notes

Fix a bug in TensorFlow autologging causing user-specified callback lists passed to `fit()` / `fit_generator()` to be modified.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tensorflow.py
+++ b/mlflow/tensorflow.py
@@ -1061,9 +1061,9 @@ def autolog(
                     args[5] = callbacks
                     args = tuple(args)
                 else:
-                    callbacks = list(kwargs.get("callbacks", []))
                     # Make a shallow copy of the preexisting callbacks and introduce TensorBoard
                     # & tf.keras callbacks if necessary
+                    callbacks = list(kwargs.get("callbacks", []))
                     kwargs["callbacks"], self.log_dir = _setup_callbacks(
                         callbacks, log_models, metrics_logger
                     )
--- a/mlflow/tensorflow.py
+++ b/mlflow/tensorflow.py
@@ -1113,21 +1113,30 @@ def autolog(
             run_id = mlflow.active_run().info.run_id
 
             with batch_metrics_logger(run_id) as metrics_logger:
-                # Checking if the 'callback' argument of fit() is set
+                # Check if the 'callback' argument of fit() is set positionally
                 if len(args) >= 5:
-                    tmp_list = list(args)
-                    tmp_list[4], self.log_dir = _setup_callbacks(
-                        tmp_list[4], log_models, metrics_logger
+                    # Convert the positional training function arguments to a list in order to
+                    # mutate the contents
+                    args = list(args)
+                    # Make a shallow copy of the preexisting callbacks to avoid permanently
+                    # modifying their contents for future training invocations. Introduce
+                    # TensorBoard & tf.keras callbacks if necessary
+                    callbacks = list(args[4])
+                    callbacks, self.log_dir = _setup_callbacks(
+                        callbacks, log_models, metrics_logger
                     )
-                    args = tuple(tmp_list)
-                elif kwargs.get("callbacks"):
-                    kwargs["callbacks"], self.log_dir = _setup_callbacks(
-                        kwargs["callbacks"], log_models, metrics_logger
-                    )
+                    # Replace the callbacks positional entry in the copied arguments and convert
+                    # the arguments back to tuple form for usage in the training function
+                    args[4] = callbacks
+                    args = tuple(args)
                 else:
+                    # Make a shallow copy of the preexisting callbacks and introduce TensorBoard
+                    # & tf.keras callbacks if necessary
+                    callbacks = list(kwargs.get("callbacks") or [])
                     kwargs["callbacks"], self.log_dir = _setup_callbacks(
-                        [], log_models, metrics_logger
+                        callbacks, log_models, metrics_logger
                     )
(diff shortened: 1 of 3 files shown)


mlflow/tensorflow.py
```diff
@@ -1044,25 +1044,33 @@ def _patch_implementation(
...
+                    # Make a shallow copy of the preexisting callbacks and introduce TensorBoard
+                    # & tf.keras callbacks if necessary
                     kwargs["callbacks"], self.log_dir = _setup_callbacks(
-                        [], log_models, metrics_logger
+                        callbacks, log_models, metrics_logger
                     )
 
+                early_stop_callback = _get_early_stop_callback(callbacks)
                 _log_early_stop_callback_params(early_stop_callback)
+
```
dbczumar, line 1073: Add a newline to more clearly delineate early stopping callback logging from the original / underlying ML function invocation

tests/tensorflow_autolog/test_tensorflow2_autolog.py
```diff
@@ -458,6 +458,39 @@ def test_tf_keras_autolog_non_early_stop_callback_no_log(tf_keras_random_data_ru
     assert len(metric_history) == num_of_epochs
 
 
+@pytest.mark.parametrize("fit_variant", ["fit", "fit_generator"])
+def test_tf_keras_autolog_does_not_mutate_original_callbacks_list(
```
apurva-koti, line 462: Just making sure, did this test fail without your change above?

dbczumar: Oh yeah, big time. It still fails against TF 2.0.4 - will need to look into that before merging

dbczumar: Fixed the 2.0.4 issue (needed to apply similar fix logic to our patch for `fit_generator()`, which is only relevant for TF < 2.1.0 (e.g. 2.0.4)

apurva-koti (approved):
Thanks @dbczumar !

PR #4200, opened 2021-03-24
End autologged runs when execution is interrupted (sigint)

dbczumar:

Autologging creates a fluent MLflow run on behalf of a user if a run does not already exist when a training session begins. However, if the training session is interrupted via a SIGINT (for example, if the the user presses `ctrl-c` on a Mac), this automatically-created run is not ended in certain cases. Specifically, if run was created by an autologging integration that makes use of [PatchFunction](https://github.com/mlflow/mlflow/blob/b57f95498ac11d70d3d153ce5e0895a7ab4ab7e9/mlflow/utils/autologging/__init__.py#L582) (`mlflow.tensorflow.autolog()`, for example: https://github.com/mlflow/mlflow/blob/b57f95498ac11d70d3d153ce5e0895a7ab4ab7e9/mlflow/tensorflow.py#L1043).

This PR fixes that issue, ensuring that runs created on behalf of a user during MLflow autologging are terminated in the event of a user interrupt.

## How is this patch tested?

- Unit tests
- Manual test: I ran the following TensorFlow training code in a Databricks notebook cell, hit "stop execution", and confirmed that there was no active run (i.e. `assert not mlflow.active_run()`) in a subsequent cell after execution was stopped. In contrast, an active run was left behind when the example was run using MLflow 1.14.1.

```
import tensorflow as tf
from tensorflow.keras.wrappers.scikit_learn import KerasClassifier
from tensorflow.keras import layers, regularizers, metrics
from tensorflow.keras.optimizers import Adam
from tensorflow.python.keras.metrics import AUC, Precision, Recall

from sklearn.model_selection import GridSearchCV, RandomizedSearchCV
from sklearn.preprocessing import LabelEncoder, OneHotEncoder

import mlflow 
import mlflow.tensorflow

import warnings
warnings.filterwarnings("ignore")

def load_data():
  import pandas as pd
  X = pd.read_csv("x.csv").drop("Unnamed: 0", axis=1)
  y = pd.read_csv("y.csv")["bad_loan"]
  return X, y

def encode_inputs(X):
  ohe = OneHotEncoder()
  ohe.fit(X)
  X = ohe.transform(X)
  return X

def encode_labels(y): 
  enc = LabelEncoder()
  enc.fit(y)
  return enc.transform(y)

def build_model(optimizer='rmsprop', init='glorot_uniform'):
  model = Sequential()
  model.add(Dense(17, input_dim=24064, kernel_initializer=init, 
                  kernel_regularizer=regularizers.l2(1e-2), activation='relu'))
  model.add(Dense(1, activation='sigmoid'))
  
  model.compile(optimizer=optimizer,
                loss='binary_crossentropy',
                metrics=['accuracy'])
  
  return model

X, y = load_data()
X_enc = encode_inputs(X)
y_enc = encode_labels(y)

mlflow.tensorflow.autolog(every_n_iter=1)

model = KerasClassifier(build_fn=build_model, verbose=2)

model.fit(X_enc, y_enc, epochs=100)
```

Please ask @dbczumar for a copy of `x.csv` and `y.csv` if you want to try this particular code snippet yourself. Any tf.keras model that trains for more than 5-10 seconds is sufficient.

## Release Notes

Runs created by autologging are now terminated correctly in the event of sigints / keyboard interrupts.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/autologging/__init__.py
+++ b/mlflow/utils/autologging/__init__.py
@@ -618,7 +618,7 @@ class PatchFunction:
     def __call__(self, original, *args, **kwargs):
         try:
             return self._patch_implementation(original, *args, **kwargs)
-        except Exception as e:
+        except (Exception, KeyboardInterrupt) as e:
             try:
                 self._on_exception(e)
             finally:
@@ -954,7 +954,7 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
 
             try:
                 result = patch_function(original, *args, **kwargs)
-            except:
+            except (Exception, KeyboardInterrupt):
                 if managed_run:
                     try_mlflow_log(mlflow.end_run, RunStatus.to_string(RunStatus.FAILED))
                 raise
--- a/mlflow/utils/autologging/__init__.py
+++ b/mlflow/utils/autologging/__init__.py
@@ -619,6 +619,9 @@ class PatchFunction:
         try:
             return self._patch_implementation(original, *args, **kwargs)
         except (Exception, KeyboardInterrupt) as e:
+            # Handle keyboard interrupts, in addition to standard Python exceptions, to ensure
+            # that runs are terminated if a user prematurely interrupts training execution
+            # (e.g. via sigint / ctrl-c)
             try:
                 self._on_exception(e)
             finally:
@@ -955,6 +958,9 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
             try:
                 result = patch_function(original, *args, **kwargs)
             except (Exception, KeyboardInterrupt):
+                # Handle keyboard interrupts, in addition to standard Python exceptions, to ensure
+                # that runs are terminated if a user prematurely interrupts training execution
+                # (e.g. via sigint / ctrl-c)
                 if managed_run:
                     try_mlflow_log(mlflow.end_run, RunStatus.to_string(RunStatus.FAILED))
                 raise
--- a/mlflow/utils/autologging/__init__.py
+++ b/mlflow/utils/autologging/__init__.py
@@ -604,8 +604,9 @@ class PatchFunction:
     @abstractmethod
     def _on_exception(self, exception):
         """
-        Called when an unhandled exception prematurely terminates the execution
-        of `_patch_implementation`.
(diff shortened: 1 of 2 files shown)


mlflow/utils/autologging/__init__.py
```diff
@@ -954,7 +955,10 @@ def patch_with_managed_run(original, *args, **kwargs):
 
             try:
                 result = patch_function(original, *args, **kwargs)
-            except:
+            except (Exception, KeyboardInterrupt):
+                # In addition to standard Python exceptions, handle keyboard interrupts to ensure
+                # that runs are terminated if a user prematurely interrupts training execution
+                # (e.g. via sigint / ctrl-c)
                 if managed_run:
                     try_mlflow_log(mlflow.end_run, RunStatus.to_string(RunStatus.FAILED))
```
dbczumar, line 963: Ending the run with status `FAILED` in response to a keyboard interrupt is consistent with the behavior of `with mlflow.start_run()` (see https://github.com/mlflow/mlflow/blob/b57f95498ac11d70d3d153ce5e0895a7ab4ab7e9/mlflow/tracking/fluent.py#L100).

While `KILLED` may make more sense, I'd rather maintain consistency and avoid making backwards-incompatible changes.

mlflow/utils/autologging/__init__.py
```diff
@@ -618,7 +619,7 @@ def call(cls, original, *args, **kwargs):
     def __call__(self, original, *args, **kwargs):
         try:
             return self._patch_implementation(original, *args, **kwargs)
-        except Exception as e:
+        except (Exception, KeyboardInterrupt) as e:
```
dbczumar, line 622: Instead of handling `KeyboardInterrupt`, I also considered handling SIGINTS via `signal.signal` (https://docs.python.org/3/library/signal.html#signal.signal). However, performing cleanup logic in `signal.signal` has some scary implications (e.g., it can lead to deadlock if the cleanup operating makes use of `logger` since `logger` uses a non-reetrant lock & a signal handler may be called in the middle of another logging statement).

While `KeyboardInterrupt` isn't guaranteed to fire 100% of the time on a SIGINT, it should be "good enough" for most practical cases.

smurching (approved):
LGTM

PR #4218, opened 2021-03-30
Add mlflow.sklearn.eval_and_log_metrics

alkispoly-db:

Introduce mlflow.sklearn.eval_and_log_metrics() to enable callers to compute and register evaluation-time metrics and artifacts that mirror the autologged training-time metrics and artifacts.

The implementation is based on a refactoring of the auto-logging code, so that future changes to training-time metrics or artifacts will be automatically reflected to log_eval_metrics().

## How is this patch tested?

Unit tested

- mlflow.sklearn.log_eval_metrics() allows the computation and logging of evaluation-time metrics and artifacts that mirror the autologged training-time metrics and artifacts. For instance, this enables logging quality metrics based on golden evaluation datasets. 

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -818,14 +818,6 @@ def autolog(
         :param kwargs: The keyword arguments passed to the scikit-learn training routine.
         """
 
-        # log common metrics and artifacts for estimators (classifier, regressor)
-
-        def get_input_example():
-            # Fetch an input example using the first several rows of the array-like
-            # training data supplied to the training routine (e.g., `fit()`)
-            input_example = X[:INPUT_EXAMPLE_SAMPLE_ROWS]
-            return input_example
-
         def infer_model_signature(input_example):
             if not hasattr(estimator, "predict"):
                 raise Exception(
@@ -836,6 +828,8 @@ def autolog(
             return infer_signature(input_example, estimator.predict(input_example))
 
         (X, y_true, sample_weight) = _get_args_for_metrics(estimator.fit, args, kwargs)
+
+        # log common metrics and artifacts for estimators (classifier, regressor)
         _log_estimator_content(
             estimator=estimator,
             prefix=_TRAINING_PREFIX,
@@ -844,6 +838,12 @@ def autolog(
             y_true=y_true,
             sample_weight=sample_weight,
         )
+
+        def get_input_example():
+            # Fetch an input example using the first several rows of the array-like
+            # training data supplied to the training routine (e.g., `fit()`)
+            input_example = X[:INPUT_EXAMPLE_SAMPLE_ROWS]
+            return input_example
 
         if log_models:
             # Will only resolve `input_example` and `signature` if `log_models` is `True`.
@@ -1020,7 +1020,7 @@ def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_we
 
 
     Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the
-    metrics and artifacts are named 'eval_XXXXX'. Note that training-time metrics are auto-logged
+    metrics and artifacts are named 'val_XXXXX'. Note that training-time metrics are auto-logged
     as 'training_XXXXX'.
 
     The run under which to log the metrics/artifacts is chosen as follows:
@@ -1028,6 +1028,11 @@ def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_we
     - Or, under the same run as the model_uri if the latter is of the form 'runs:/...'
     - Or, under a new run
     The chosen run is in the return value of the method.
+
+    Raises an error if:
+    - prefix is empty
+    - neither model nor model_uri are specified
+    - the model_uri cannot be resolved to a valid model
 
     :param model: The model to be evaluated. Exactly one model_uri or model must be specified.
     :param model_uri: The URI of the model to be evaluated. Exactly one model_uri or model must be
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -1028,8 +1028,8 @@ def eval_and_log_metrics(model, X, y_true, *, prefix, sample_weight=None):
 
     Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the
     metrics and artifacts are named 'val_XXXXX'. Note that training-time metrics are auto-logged
-    as 'training_XXXXX'. Metrics and artifacs are logged under the current run if one exists,
-    otherwise a new run is started and left open.
+    as 'training_XXXXX'. Metrics and artifacts are logged under the currently active run if one exists,
+    otherwise a new run is started and left active.
 
     Raises an error if:
     - prefix is empty
@@ -1039,13 +1039,13 @@ def eval_and_log_metrics(model, X, y_true, *, prefix, sample_weight=None):
     from sklearn.base import BaseEstimator
 
     if prefix is None or prefix == "":
-        raise ValueError("must specify a non-empty prefix")
+        raise ValueError("Must specify a non-empty prefix")
 
     if not isinstance(model, BaseEstimator):
-        raise ValueError("not a sklearn model")
+        raise ValueError("The provided model was not an sklearn estimator. Please ensure the passed-in model is a sklearn estimator subclassing sklearn.base.BaseEstimator")
 
     if not hasattr(model, "predict"):
-        raise ValueError("model does not support predictions")
+        raise ValueError("Model does not support predictions. Please pass a model object defining a predict() method")
 
     active_run = mlflow.active_run()
     run = active_run if active_run is not None else mlflow.start_run()
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -1032,8 +1032,8 @@ def eval_and_log_metrics(model, X, y_true, *, prefix, sample_weight=None):
     otherwise a new run is started and left active.
 
     Raises an error if:
-    - prefix is empty
-    - model is not an sklearn estimator or does not support the 'predict' method
+      - prefix is empty
+      - model is not an sklearn estimator or does not support the 'predict' method
     """
     from mlflow.sklearn.utils import _log_estimator_content
     from sklearn.base import BaseEstimator
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -836,8 +836,14 @@ def autolog(
             return infer_signature(input_example, estimator.predict(input_example))
 
         (X, y_true, sample_weight) = _get_args_for_metrics(estimator.fit, args, kwargs)
-        _log_estimator_content(estimator=estimator, prefix=_TRAINING_PREFIX, run_id=mlflow.active_run().info.run_id,
-                               X=X, y_true=y_true, sample_weight=sample_weight)
+        _log_estimator_content(
+            estimator=estimator,
+            prefix=_TRAINING_PREFIX,
+            run_id=mlflow.active_run().info.run_id,
+            X=X,
+            y_true=y_true,
+            sample_weight=sample_weight,
+        )
 
         if log_models:
             # Will only resolve `input_example` and `signature` if `log_models` is `True`.
@@ -984,8 +990,9 @@ def autolog(
 
 def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_weight=None):
     """
-    Computes and logs evaluation metrics (and artifacts) for the given model and labeled dataset. The metrics/artifacts
-    mirror what is auto-logged when training a model (see mlflow.sklearn.autolog).
+    Computes and logs evaluation metrics (and artifacts) for the given model and labeled dataset.
+    The metrics/artifacts mirror what is auto-logged when training a model
+    (see mlflow.sklearn.autolog).
 
     ** Example **
 
@@ -1012,8 +1019,9 @@ def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_we
             mlflow.log_eval_metrics(model=model, X=X_eval, y_true=y_eval, prefix="val_")
 
 
-    Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the metrics and artifacts
-    are named 'eval_XXXXX'. Note that training-time metrics are auto-logged as 'training_XXXXX'.
+    Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the
+    metrics and artifacts are named 'eval_XXXXX'. Note that training-time metrics are auto-logged
+    as 'training_XXXXX'.
 
     The run under which to log the metrics/artifacts is chosen as follows:
     - Under the active run if one exists
@@ -1022,13 +1030,15 @@ def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_we
     The chosen run is in the return value of the method.
 
     :param model: The model to be evaluated. Exactly one model_uri or model must be specified.
-    :param model_uri: The URI of the model to be evaluated. Exactly one model_uri or model must be specified.
+    :param model_uri: The URI of the model to be evaluated. Exactly one model_uri or model must be
+                    specified.
     :param X: The features for the evaluation dataset.
     :param y_true: The labels for the evaluation dataset.
     :param prefix: Prefix used to name metrics and artifacts.
     :param sample_weight: Per-sample weights to apply in the computation of metrics/artifacts.
-    :return: A tuple (run_id, metrics, artifacts) where run_id is the run under which metrics/artifacts are logged,
-             metrics is a dict of the logged metrics, and artifacts a list of the logged artifact paths.
+    :return: A tuple (run_id, metrics, artifacts) where run_id is the run under which
+             metrics/artifacts are logged, metrics is a dict of the logged metrics, and artifacts a
+             list of the logged artifact paths.
     """
     from mlflow.sklearn.utils import _log_estimator_content
     from mlflow.store.artifact.runs_artifact_repo import RunsArtifactRepository
@@ -1040,7 +1050,7 @@ def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_we
         raise ValueError("only one of `model` or `model_uri` must be specified")
 
     if model_uri is None and model is None:
-        raise ValueError('at least one of `model` or `model_uri` must be specified')
+        raise ValueError("at least one of `model` or `model_uri` must be specified")
 
     if model_uri is not None:
         model = mlflow.sklearn.load_model(model_uri)
@@ -1051,17 +1061,27 @@ def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_we
 
     if mlflow.active_run() is not None:
         run_id = mlflow.active_run().info.run_id
-        (metrics, artifacts) = _log_estimator_content(estimator=model,
-                                                      run_id=run_id, prefix=prefix, X=X, y_true=y_true,
-                                                      sample_weight=sample_weight)
+        (metrics, artifacts) = _log_estimator_content(
+            estimator=model,
+            run_id=run_id,
+            prefix=prefix,
+            X=X,
+            y_true=y_true,
+            sample_weight=sample_weight,
+        )
     else:
         if RunsArtifactRepository.is_runs_uri(model_uri):
             run_id, _ = RunsArtifactRepository.parse_runs_uri(model_uri)
 
         with mlflow.start_run(run_id=run_id) as run:
             run_id = run.info.run_id
-            (metrics, artifacts) = _log_estimator_content(estimator=model,
-                                                          run_id=run_id, prefix=prefix, X=X, y_true=y_true,
-                                                          sample_weight=sample_weight)
+            (metrics, artifacts) = _log_estimator_content(
+                estimator=model,
+                run_id=run_id,
+                prefix=prefix,
+                X=X,
+                y_true=y_true,
+                sample_weight=sample_weight,
+            )
 
     return (run_id, metrics, artifacts)
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -988,9 +988,9 @@ def autolog(
                 )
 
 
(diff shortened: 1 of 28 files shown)


mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,102 @@ def patched_fit(original, self, *args, **kwargs):
...
+
+        # train a model
+        model = LinearRegression()
+        with mlflow.start_run() as run:
+            model.fit(X, y)
+            mlflow.log_eval_metrics(model=model, X=X_eval, y_true=y_eval, prefix="val_")
+
+
+    Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the
+    metrics and artifacts are named 'eval_XXXXX'. Note that training-time metrics are auto-logged
```
mengxr, line 1023: "eval" -> "val" or rename the prefix in L1019.

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,102 @@ def patched_fit(original, self, *args, **kwargs):
...
+    as 'training_XXXXX'.
+
+    The run under which to log the metrics/artifacts is chosen as follows:
+    - Under the active run if one exists
+    - Or, under the same run as the model_uri if the latter is of the form 'runs:/...'
```
mengxr, line 1028: not sure this is what users want. maybe easier to require an active run.

mlflow/sklearn/__init__.py
```diff
@@ -817,31 +817,13 @@ def _log_posttraining_metadata(estimator, *args, **kwargs):
...
 
         def get_input_example():
             # Fetch an input example using the first several rows of the array-like
             # training data supplied to the training routine (e.g., `fit()`)
-            fit_arg_names = _get_arg_names(estimator.fit)
-            X_var_name, y_var_name = fit_arg_names[:2]
-            input_example = _get_Xy(args, kwargs, X_var_name, y_var_name)[0][
-                :INPUT_EXAMPLE_SAMPLE_ROWS
-            ]
+            input_example = X[:INPUT_EXAMPLE_SAMPLE_ROWS]
```
mengxr, line 826: A bit strange to reference a global var that is not yet defined. Move it after L838?

mengxr (changes requested):

mlflow/sklearn/__init__.py
alkispoly-db: Always requiring an active run seems to go against the mlflow APIs (e.g., mlflow.log_metric() will initialize a new run if one is not present). So this leaves the question of what to do when there is no active run and the user specifies a runs:/ URL. We can avoid this for now if we decide to not support model_uri. However, if we support it then I can see a case for re-using the same run to log more metrics/artifacts for different datasets, so that everything is available in the same run that generated the model. The flip side is that it makes the handling of model_uri is inconsistent, and this might indeed be surprising to the users.

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,102 @@ def patched_fit(original, self, *args, **kwargs):
...
+    as 'training_XXXXX'.
+
+    The run under which to log the metrics/artifacts is chosen as follows:
+    - Under the active run if one exists
+    - Or, under the same run as the model_uri if the latter is of the form 'runs:/...'
+    - Or, under a new run
+    The chosen run is in the return value of the method.
+
+    :param model: The model to be evaluated. Exactly one model_uri or model must be specified.
+    :param model_uri: The URI of the model to be evaluated. Exactly one model_uri or model must be
```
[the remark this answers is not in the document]
alkispoly-db: Agreed. Having a model_uri is mostly a convenience for the user and suggested by another reviewer. If this is inconsistent for mlflow in general then we can remove it and add it back later if users ask for it. Wdyt?

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,102 @@ def patched_fit(original, self, *args, **kwargs):
...
+                    specified.
+    :param X: The features for the evaluation dataset.
+    :param y_true: The labels for the evaluation dataset.
+    :param prefix: Prefix used to name metrics and artifacts.
+    :param sample_weight: Per-sample weights to apply in the computation of metrics/artifacts.
+    :return: A tuple (run_id, metrics, artifacts) where run_id is the run under which
```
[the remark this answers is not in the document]
alkispoly-db: If the run is created as part of this method (see the case for creating a new run) then it seems useful to return it so that the caller can access everything that got logged. If we remove run_id from the return then we need to return artifacts + potentially more stuff that we end up logging in the future. Maybe safer to return run_id for now + metrics and omit everything else (e.g., artifacts)?

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1106,3 +1107,326 @@ def fit(self, X, y):
...
+    X_eval = X[:-1, :]
+    y_eval = y_true[:-1]
+    with mlflow.start_run() as run:
+        model = fit_model(model, X, y_true, "fit")
+        eval_run_id, eval_metrics, eval_artifacts = mlflow.sklearn.log_eval_metrics(
+            model=model, X=X_eval, y_true=y_eval, prefix="eval_"
+        )
+    # Check correctness for the returned metrics/artifacts
+    y_pred = model.predict(X_eval)
+    assert eval_metrics == {
```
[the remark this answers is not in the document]
alkispoly-db: Yeah, good point. I basically followed the pattern in other tests which perform the same type of equality checks. I can change the new tests in this PR, or create a separate PR that changes the tests in the whole sklearn suite. Wdyt?

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1106,3 +1107,326 @@ def fit(self, X, y):
...
+    }
+    assert len(eval_artifacts) == 0
+
+    # Check the the logged metrics/artifacts are the same as the returned ones.
+    _, metrics, _, artifacts = get_run_data(run_id)
+    assert eval_metrics == metrics
+    assert eval_artifacts == artifacts
+
+
+def test_log_eval_metrics_throws_with_invalid_args():
```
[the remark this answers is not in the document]
alkispoly-db: I can think of the following cases:
- A non-sklearn model.
- A model_uri that fails to resolve to a valid model.

Did you have other cases in mind?

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,102 @@ def patched_fit(original, self, *args, **kwargs):
                 safe_patch(
                     FLAVOR_NAME, class_def, func_name, patched_fit, manage_run=True,
                 )
+
+
+def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_weight=None):
+    """
+    Computes and logs evaluation metrics (and artifacts) for the given model and labeled dataset.
+    The metrics/artifacts mirror what is auto-logged when training a model
+    (see mlflow.sklearn.autolog).
```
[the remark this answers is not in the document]
alkispoly-db: Added a comment about potential failure modes. I can update it if we decide to remove model_uri.

mlflow/sklearn/__init__.py
alkispoly-db: Done.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1106,3 +1107,326 @@ def fit(self, X, y):
     _, _, tags, _ = get_run_data(run.info.run_id)
     assert {"estimator_name": "ChildMod"}.items() <= tags.items()
     assert model.predict(1) == np.array([8])
+
```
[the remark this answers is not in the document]
alkispoly-db: Added one with classifier, which has the most metrics/artifacts.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1106,3 +1107,326 @@ def fit(self, X, y):
...
+    # disable autologging so that we can check for the sole existence of eval-time metrics
+    mlflow.sklearn.autolog(disable=True)
+
+    # use simple `LinearRegression`, which only implements `fit`.
+    model = sklearn.linear_model.LinearRegression()
+    X, y_true = get_iris()
+    X_eval = X[:-1, :]
+    y_eval = y_true[:-1]
+    with mlflow.start_run() as run:
+        model = fit_model(model, X, y_true, "fit")
```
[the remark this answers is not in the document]
alkispoly-db: Added a "fake" option for such models. PTAL.

mlflow/sklearn/__init__.py
alkispoly-db: My bad. Moved it later.

mlflow/sklearn/__init__.py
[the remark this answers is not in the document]
mengxr: Let's remove it for now.

mlflow/sklearn/__init__.py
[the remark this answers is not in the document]
mengxr: See my above comment about leaving the newly created run open.

mlflow/sklearn/utils.py
```diff
@@ -503,6 +472,54 @@ def _log_specialized_estimator_content(fitted_estimator, run_id, fit_args, fit_k
 
             try_mlflow_log(mlflow_client.log_artifacts, run_id, tmp_dir.path())
 
+    return (metrics, artifact_paths)
+
+
+def _log_estimator_content(estimator, run_id, prefix, X, y_true, sample_weight):
```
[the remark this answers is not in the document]
mengxr: Okay. I couldn't think of a better name ...

tests/sklearn/test_sklearn_autolog.py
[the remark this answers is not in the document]
mengxr: nvm. let's wait for the first failure to fix it.

tests/sklearn/test_sklearn_autolog.py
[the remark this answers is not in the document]
mengxr: Are there sklearn models without `score()`? If not, I think testing a non-sklearn model should be sufficient.

mengxr (changes requested):

mlflow/sklearn/__init__.py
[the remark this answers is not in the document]
alkispoly-db: Done.

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,107 @@ def patched_fit(original, self, *args, **kwargs):
                 safe_patch(
                     FLAVOR_NAME, class_def, func_name, patched_fit, manage_run=True,
                 )
+
+
+def log_eval_metrics(*, model=None, model_uri=None, X, y_true, prefix, sample_weight=None):
```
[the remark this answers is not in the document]
alkispoly-db: Good point. Renamed.

mlflow/sklearn/__init__.py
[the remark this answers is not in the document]
alkispoly-db: Done.

mlflow/sklearn/__init__.py
alkispoly-db: Done.

tests/sklearn/test_sklearn_autolog.py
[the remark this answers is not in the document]
alkispoly-db: The sklearn documentation suggests that all Estimators have a `score` although I found this to be wrong (e.g., some clusterers inherit from BaseEstimator but do not support neither `score` nor `predict`). However, we don't care about the score method since we check for it before recording any metrics. I added a test to exercise this behavior. Instead, we care for `predict` so I added this to the list of requirements for `model`. Let me know if this works.

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,79 @@ def patched_fit(original, self, *args, **kwargs):
...
+    (metrics, artifacts) = _log_estimator_content(
+        estimator=model,
+        run_id=run.info.run_id,
+        prefix=prefix,
+        X=X,
+        y_true=y_true,
+        sample_weight=sample_weight,
+    )
+
+    return (metrics, artifacts)
```
mengxr, line 1064: I still prefer returning only `metrics`. If users need artifact paths, they can lookup the run info. I think that will match what users guess based on the method name. If we return a tuple, most users need to lookup the API doc to understand the return value structure, which is also easy to forget.

mengxr (changes requested):
Only one comment about the return value. cc: @smurching @harupy to make a pass.

mlflow/sklearn/__init__.py
harupy: Is there any reasonable use case for `artifacts` 🤔 ?

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,79 @@ def patched_fit(original, self, *args, **kwargs):
                 safe_patch(
                     FLAVOR_NAME, class_def, func_name, patched_fit, manage_run=True,
                 )
+
+
+def eval_and_log_metrics(*, model, X, y_true, prefix, sample_weight=None):
```
harupy, line 991: Is it a bit inconvenient if `model`, `X` and `y` are keyword-only arguments?

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,79 @@ def patched_fit(original, self, *args, **kwargs):
...
+    Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the
+    metrics and artifacts are named 'val_XXXXX'. Note that training-time metrics are auto-logged
+    as 'training_XXXXX'. Metrics and artifacs are logged under the current run if one exists,
+    otherwise a new run is started and left open.
+
+    Raises an error if:
+    - prefix is empty
+    - model is not an sklearn estimator or does not support the 'predict' method
+
+    :param model: The model to be evaluated.
```
harupy, line 1032: Can we move the arguments block above the example code?

mlflow/sklearn/__init__.py
alkispoly-db: For logging I think there is, in order to maintain symmetry with training auto-logging. For programmatic access, it's a fair question. If there are no precedents in MLflow's usage patterns then we can remove the artifacts return and simplify following mengxr's suggestion. Wdyt?

mlflow/sklearn/__init__.py
alkispoly-db: Personally I find keyword arguments more robust in terms of avoiding mistakes but I am happy to go with the conventions of MLflow. Should we drop them?

mlflow/sklearn/__init__.py
alkispoly-db: Sure thing -- I can address this in the next commit along with the previous comments (once we reach a resolution).

mlflow/sklearn/__init__.py
harupy: One use case (which I think is minor) that I could come up with is:

```python
metrics, artifact_names = mlflow.sklearn.eval_and_log_metrics(...)
artifact_paths = ...

# I want to see the results now on the notebook, not on MLflow UI!

for p in artifact_paths:
    plt.imread(p)
    plt.show()
```

harupy: However, I'd vote for returning only metrics for simplicity.

mlflow/sklearn/__init__.py
harupy: > I am happy to go with the conventions of MLflow.

Currently, there is no specific rule on when to use positional / keyword-only arguments in MLflow, but I'd prefer something similar to sklearn's convention (e.g. [cross_val_score](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html) where we can pass the first three arguments in positional form).

@smurching @mengxr What do you think?

mengxr: I think "model, X, y" order should be fine and they match the actions listed in the method name. The only question is whether we put `prefix` before or after `*`. Slightly prefer putting it after `*` because it is not obvious from the method name:

~~~
eval_and_log_metrics(model, X, y, prefix="val_")
~~~

mlflow/sklearn/__init__.py
alkispoly-db: Done.

mlflow/sklearn/__init__.py
alkispoly-db: I went with mengxr's suggestion in the interest of time. If smurching has a different opinion then I can revert/change back to another scheme.

mlflow/sklearn/__init__.py
alkispoly-db: Done.

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,77 @@ def patched_fit(original, self, *args, **kwargs):
                 safe_patch(
                     FLAVOR_NAME, class_def, func_name, patched_fit, manage_run=True,
                 )
+
+
+def eval_and_log_metrics(model, X, y_true, *, prefix, sample_weight=None):
+    """
+    Computes and logs metrics (and artifacts) for the given model and labeled dataset.
+    The metrics/artifacts mirror what is auto-logged when training a model
+    (see mlflow.sklearn.autolog).
```
smurching, line 995:
```suggestion
    (see mlflow.sklearn.autolog). Metrics and artifacts are logged under the currently active run.
    If no run is active, this method will create a new active run.
```

smurching: Oops, realized we state this later on!

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,77 @@ def patched_fit(original, self, *args, **kwargs):
...
+        # train a model
+        model = LinearRegression()
+        with mlflow.start_run() as run:
+            model.fit(X, y)
+            metrics = mlflow.sklearn.eval_and_log_metrics(model, X_eval, y_eval, prefix="val_")
+
+
+    Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the
+    metrics and artifacts are named 'val_XXXXX'. Note that training-time metrics are auto-logged
+    as 'training_XXXXX'. Metrics and artifacs are logged under the current run if one exists,
```
smurching, line 1031:
```suggestion
    as 'training_XXXXX'. Metrics and artifacts are logged under the current run if one exists,
```

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,77 @@ def patched_fit(original, self, *args, **kwargs):
...
+    otherwise a new run is started and left open.
```
smurching, line 1032:
```suggestion
    otherwise a new run is started and left active.
```

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,77 @@ def patched_fit(original, self, *args, **kwargs):
...
+
+    Raises an error if:
+    - prefix is empty
+    - model is not an sklearn estimator or does not support the 'predict' method
+    """
+    from mlflow.sklearn.utils import _log_estimator_content
+    from sklearn.base import BaseEstimator
+
+    if prefix is None or prefix == "":
+        raise ValueError("must specify a non-empty prefix")
```
smurching, line 1042:
```suggestion
        raise ValueError("Must specify a non-empty prefix")
```

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,77 @@ def patched_fit(original, self, *args, **kwargs):
...
+
+    if not isinstance(model, BaseEstimator):
+        raise ValueError("not a sklearn model")
```
smurching, line 1045:
```suggestion
        raise ValueError("The provided model was not an sklearn estimator. Please ensure the passed-in model is a sklearn estimator subclassing sklearn.base.BaseEstimator")
```

Minor suggestion to make the error message more actionable

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,77 @@ def patched_fit(original, self, *args, **kwargs):
...
+
+    if not hasattr(model, "predict"):
+        raise ValueError("model does not support predictions")
```
smurching, line 1048:
```suggestion
        raise ValueError("Model does not support predictions. Please pass a model object defining a predict(pandas_df) method")
```

smurching (commented):
Just a few minor comments aimed at improving docs/error messages, otherwise LGTM, thanks @alkispoly-db! Great that we're able to reuse logic from the existing autologging implementation :)

mlflow/sklearn/__init__.py
harupy: +1 for making `prefix` keyword-only :)

mlflow/sklearn/__init__.py
harupy: Thanks!

mlflow/sklearn/__init__.py
```diff
@@ -994,3 +986,81 @@ def patched_fit(original, self, *args, **kwargs):
...
+
+
+    Each metric's and artifact's name is prefixed with `prefix`, e.g., in the previous example the
+    metrics and artifacts are named 'val_XXXXX'. Note that training-time metrics are auto-logged
+    as 'training_XXXXX'. Metrics and artifacts are logged under the currently active run if one exists,
+    otherwise a new run is started and left active.
+
+    Raises an error if:
+    - prefix is empty
+    - model is not an sklearn estimator or does not support the 'predict' method
```
harupy, line 1036: Nit: can we indent the list items here?

```suggestion
    Raises an error if:
      - prefix is empty
      - model is not an sklearn estimator or does not support the 'predict' method
```

Screen Shot 2021-04-06 at 10 53 14

https://6278-136202695-gh.circle-artifacts.com/0/docs/build/html/python_api/mlflow.sklearn.html#mlflow.sklearn.eval_and_log_metrics

harupy (approved):
LGTM once the unrelated changes have been removed!

harupy:
Hmm, spark seems flaky:

https://github.com/mlflow/mlflow/pull/4218/checks?check_run_id=2275232513#step:5:229

```
E                   py4j.protocol.Py4JJavaError: An error occurred while calling None.org.apache.spark.api.java.JavaSparkContext.
E                   : java.net.BindException: Cannot assign requested address: Service 'sparkDriver' failed after 16 retries (on a random free port)! Consider explicitly setting the appropriate binding address for the service 'sparkDriver' (for example spark.driver.bindAddress for SparkDriver) to the correct binding address.
```

PR #4225, opened 2021-04-05
Added two special cases to the Model Registry Workflow API

dmatrix:

This PR adds two scenarios to the Model Registry workflow API:
 

- How to register existing saved models in a native format not saved or trained with MLflow
- How to register models with no MLflow built-in model flavor support.

Components 
- [x ] `area/docs`: MLflow documentation pages

- [ x] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section

### DIFF
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -452,7 +452,7 @@ register the loaded model with the Model Registry.
     Model name: SklearnLinearRegression, version 1
     Created version '1' of model 'SklearnLinearRegression'.
 
-Now, using MLflow fluent APIs, we reload the from the Model Registry and score.
+Now, using MLflow fluent APIs, we reload the model from the Model Registry and score.
 
 .. code-block:: py
 
--- a/docs/source/model-registry.rst
+++ b/docs/source/model-registry.rst
@@ -349,3 +349,133 @@ You can either delete specific versions of a registered model or you can delete 
 
     # Delete a registered model along with all its versions
     client.delete_registered_model(name="sk-learn-random-forest-reg-model")
+
+While the above workflow API demonstrates interactions with the Model Registry, there are two
+special cases that require attention. One is when you have existing ML models saved from training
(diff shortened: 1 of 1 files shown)


docs/source/model-registry.rst
```diff
@@ -349,3 +349,256 @@ You can either delete specific versions of a registered model or you can delete
...
+
+.. code-block:: text
+
+    --
+    Successfully registered model 'SklearnLinearRegression'.
+    2021/04/02 16:30:57 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.
+    Model name: SklearnLinearRegression, version 1
+    Created version '1' of model 'SklearnLinearRegression'.
+
+Now, using MLflow fluent APIs, we reload the from the Model Registry and score.
```
sueann, line 456: do we need this section specifically here? seems generic for all registered models. or is it to specifically show that the registered model can be applied to the exact same data set as the originally saved model?

sueann: i guess it does not hurt to have this so there is an end to end workflow example here, just was worried about the length

sueann (commented):
the content generally looks good to me - thanks for doing this @dmatrix ! cc @andreakress for the language and possibly organization.

docs/source/model-registry.rst
dmatrix: Yes, @sueann, I put it in there for that purpose. In this case, I was comparing and verifying that the model saved without using MLflow and then logged and registered with MLflow produces the same predictions.

andreakress:
LGTM, very clearly written. Couple of style suggestions, I might not start a sentence with "And..." and I might not use "wish" but they don't detract from the content at all so up to you.

dmatrix:
Thankyou @andreakress. Noted your style suggestions and incorporated.

sueann (approved):

PR #4236, opened 2021-04-09
Enable spark_udf to use column names from model signature by default

Loquats:
Signed-off-by: Andy Zhang <yue.zhang@databricks.com>

spark_udf returns a spark UDF python function. The user must specify the column names as input:
```
model_udf = mlflow.pyfunc.spark_udf(spark, path, result_type=”string”)
column_names = df.drop('label').schema.names
pred_df = df.withColumn('prediction', model_udf(*column_names))
```

Use [functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps) to create a wrapper for the UDF so that it can be used without specifying column names:
```
model_udf = mlflow.pyfunc.spark_udf(spark, path)
pred_df = df.withColumn('prediction', model_udf())
```
The wrapper fills in column names using the input schema logged in model signatures. If the input schema doesn't exist, revert to old behavior.

## How is this patch tested?
`pytest tests/pyfunc/test_spark.py --large`

Allow the UDF returned by `mlflow.pyfunc.spark_udf` to be called with no column name arguments, in which case use input names from the model signature.

Components 

Interface 

Language 

Integrations


tests/pyfunc/test_spark.py
```diff
@@ -120,6 +121,37 @@ def test_spark_udf(spark, model_path):
                 actual = list(new_df.select("prediction").toPandas()["prediction"])
                 assert expected == actual
 
+def test_spark_udf_autofills_no_arguments(spark, model_path):
```
Loquats, line 124: We may want to merge this test with `test_spark_udf_autofills_column_names_with_schema`

mlflow/pyfunc/__init__.py
```diff
@@ -847,7 +848,23 @@ def predict(*args):
...
+
+    @functools.wraps(udf)
+    def udf_with_default_cols(*args):
+        if len(args) == 0:
+            model = SparkModelCache.get_or_load(archive_path)
+            input_schema = model.metadata.get_input_schema()
+
+            if input_schema and input_schema.has_input_names():
+                input_names = input_schema.input_names()
+                return udf(*input_names)
```
tomasatdatabricks, line 861: I think we should wrap the input_names in a struct for consistency, since that is what we should be recommending for users to do (e.g. in case the model has no signature). It does not really make a difference since we ask for the columns in the schema order and we autofill column names, but just in case there is any difference in behavior.

tests/pyfunc/test_spark.py
```diff
@@ -121,6 +122,39 @@ def test_spark_udf(spark, model_path):
...
+        udf = mlflow.pyfunc.spark_udf(
+            spark, "runs:/{}/model".format(run.info.run_id), result_type=ArrayType(StringType())
+        )
+        good_data = spark.createDataFrame(
+            pd.DataFrame(
+                columns=["a", "b", "c", "d"], data={"a": [1], "b": [2], "c": [3], "d": [4]}
+            )
+        )
+        res = good_data.withColumn("res", udf()).select("res").toPandas()
+        assert res["res"][0] == ["a", "b", "c"]
```
tomasatdatabricks, line 146: Can we also test that if user gives argument list we respect that? E.g. assert that the following raises: 
`res = good_data.withColumn("res", udf(struct("b", "c")))`

tests/pyfunc/test_spark.py
```diff
@@ -121,6 +122,39 @@ def test_spark_udf(spark, model_path):
...
+
+        # this dataframe won't work because it's missing column a
+        bad_data = spark.createDataFrame(
+            pd.DataFrame(
+                columns=["x", "b", "c", "d"], data={"x": [1], "b": [2], "c": [3], "d": [4]}
+            )
+        )
+        with pytest.raises(AnalysisException, match=r"cannot resolve '`a`' given input columns"):
+            bad_data.withColumn("res", udf())
+
```
tomasatdatabricks, line 156: Can we add a test case for what happens if the model has no signature and user calls it with empty argument list?

mlflow/pyfunc/__init__.py
```diff
@@ -847,7 +848,23 @@ def predict(*args):
         else:
             return result[result.columns[0]]
 
-    return pandas_udf(predict, result_type)
+    udf = pandas_udf(predict, result_type)
+
+    @functools.wraps(udf)
+    def udf_with_default_cols(*args):
+        if len(args) == 0:
+            model = SparkModelCache.get_or_load(archive_path)
```
tomasatdatabricks, line 856: Rather than loading the full model, can we read the model yaml file from `local_model_path` (line 780)? 
I think it may be safer to avoid loading the model on the driver.

tomasatdatabricks (commented):
Looks good, I left a few comments.

mlflow/pyfunc/__init__.py
Loquats: I tried this out and haven't gotten it to work yet. Changing that line to `return udf(struct(*input_names))` results in this error:
```
TypeError: Unsupported type in conversion from Arrow: struct<a: int64, b: int64, c: int64>
```
I'm not sure if it's actually a bug, because the same error happens if I add `struct` wrappers to existing unit tests, but leave the code unchanged (master branch). See this example: https://github.com/mlflow/mlflow/pull/4244

Do you know how to avoid this issue when using structs, or should we leave the code as is?

Loquats (commented):
Addressed some comments, and left some notes/questions.

mlflow/pyfunc/__init__.py
```diff
@@ -847,7 +848,23 @@ def predict(*args):
...
+            input_schema = model.metadata.get_input_schema()
+
+            if input_schema and input_schema.has_input_names():
+                input_names = input_schema.input_names()
+                return udf(*input_names)
+            else:
+                return udf(*args)
```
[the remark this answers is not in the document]
Loquats: I guess we should just remove `*args` here since it's an empty array.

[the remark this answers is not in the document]
tomasatdatabricks: I see. Can we catch the missing argument error and display the message then? I am worried that there is no way to tell if you can call the model udf with no arguments and it can be confusing. If we at least can inform the user that this particular model has no schema or no column names that would be helpful.

mlflow/pyfunc/__init__.py
tomasatdatabricks: Oh I see, that's probably because we test with spark < 3. We can leave it as is then.

mlflow/pyfunc/__init__.py
tomasatdatabricks: I think you can load the Model while the tempdir exists (e.g. add the `model = Model.load(local_model_path)` to the with block on line 780) and use it later.

tomasatdatabricks: in fact, it may be useful to attach the model object to the returned function object, e.g. `udf.metadata = model`

tomasatdatabricks:
Oh, I forgot to ask in my review, we also need to update docs (arguably that's the most important part otherwise users will have no clue this is an option). Are you planning on updating docs in a followup? I think it would be better to do it as part of a single PR.

Loquats:
Thanks for the reminder. I'll update docs as part of this PR.

mlflow/pyfunc/__init__.py
```diff
@@ -780,6 +782,7 @@ def spark_udf(spark, model_uri, result_type="double"):
             artifact_uri=model_uri, output_path=local_tmpdir.path()
         )
         archive_path = SparkModelCache.add_local_model(spark, local_model_path)
+        model = Model.load(os.path.join(local_model_path, MLMODEL_FILE_NAME))
```
tomasatdatabricks, line 785: I've realized that we use `model` in the predict function as well. It should not matter but it is confusing. Can you rename this to `model_metadata`?

mlflow/pyfunc/__init__.py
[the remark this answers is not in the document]
Loquats: Sounds good. The missing argument error is actually raised by the `predict` function defined a few lines above. Instead of catching the error and displaying a different message, does it make sense if we just clarify the error message if the model has no schema and no col name arguments (line 840 in the diff) ? Let me know if you actually meant something different.

Loquats:
I added some documentation in the docstring as well as models.rst. Let me know what you think, and if there are other doc pages to update.

mlflow/pyfunc/__init__.py
```diff
@@ -847,7 +862,23 @@ def predict(*args):
...
+    @functools.wraps(udf)
+    def udf_with_default_cols(*args):
+        if len(args) == 0:
+            input_schema = model_metadata.get_input_schema()
+
+            if input_schema and input_schema.has_input_names():
+                input_names = input_schema.input_names()
+                return udf(*input_names)
+            else:
+                return udf()
```
tomasatdatabricks, line 877: can we check if the model has input schema with no column names? and raise an error in that case?

mlflow/pyfunc/__init__.py
```diff
@@ -832,11 +838,20 @@ def predict(*args):
             result = result.select_dtypes(include=(np.number,)).astype(np.float64)
 
         if len(result.columns) == 0:
+            message = (
```
tomasatdatabricks, line 841: did you check what is scikit's behavior when you pass it empty dataframe?

mlflow/pyfunc/__init__.py
tomasatdatabricks: Actually, would it be fair to say that we do not support models with empty argument list that have no signature? And just raise an error whenever there is an empty argument list but no signature (or one with unnamed columns)? 

I think this would lead to much better user experience for the common case. It is possible there is not a single model with no arguments that is being deployed this way.

mlflow/pyfunc/__init__.py
```diff
@@ -832,11 +838,20 @@ def predict(*args):
...
+                "The model did not produce any values compatible with the requested type "
+                "'{}'. ".format(str(elem_type))
+            )
+            if len(args) == 0 and input_schema is None:
+                message += (
+                    "Apply the udf by specifying column name arguments, for example "
```
tomasatdatabricks, line 847: I meant to add this message at the top level. If that is not possible, I would not add it at all because I don't think we can add the message reliably  at the python level - by which I mean some models will raise exceptions, some will produce empty output, it would be difficult to reliably detect when predict fails due to missing columns.

tomasatdatabricks (approved):
lgtm!

Loquats:
All tests pass. Ready to merge.

PR #4242, opened 2021-04-13
Add return type annotations to key MLflow APIs

harupy:

Put type annotations on key MLflow APIs.

#### Which API should be annotated?

- Fluent APIs (e.g. `mlflow.start_run`)
- Client APIs  (e.g. `mlflow.tracking.MlflowClient.get_run`)

## How is this patch tested?

Manual testing

Components 

Interface 

Language 

Integrations

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -345,7 +345,7 @@ def _get_reference_map():
     """
     Gets a dict that maps an invalid reference to a valid one.
     """
-    res = {
+    ref_map = {
         "mlflow.tracking.fluent.ActiveRun": "mlflow.ActiveRun",
         "mlflow.store.entities.paged_list.PagedList": "mlflow.store.entities.PagedList",
     }
@@ -354,7 +354,7 @@ def _get_reference_map():
     for entity_name in mlflow.entities.__all__:
         invalid_ref = "mlflow.entities.{}.{}".format(_camel_to_snake(entity_name), entity_name)
         valid_ref = "mlflow.entities.{}".format(entity_name)
-        res[invalid_ref] = valid_ref
+        ref_map[invalid_ref] = valid_ref
 
     # Model registry entities
     for entity_name in mlflow.entities.model_registry.__all__:
@@ -362,9 +362,9 @@ def _get_reference_map():
             _camel_to_snake(entity_name), entity_name
         )
         valid_ref = "mlflow.entities.model_registry.{}".format(entity_name)
-        res[invalid_ref] = valid_ref
+        ref_map[invalid_ref] = valid_ref
 
-    return res
+    return ref_map
 
 
 reference_map = _get_reference_map()
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -337,10 +337,6 @@ nitpick_ignore = [
 ]
 
 
-def _camel_to_snake(x):
-    return re.sub(r"(?<!^)(?=[A-Z])", "_", x).lower()
-
-
 def _get_reference_map():
     """
     Gets a dict that maps an invalid reference to a valid one.
@@ -352,15 +348,15 @@ def _get_reference_map():
 
     # Tracking entities
     for entity_name in mlflow.entities.__all__:
-        invalid_ref = "mlflow.entities.{}.{}".format(_camel_to_snake(entity_name), entity_name)
+        entity_cls = getattr(mlflow.entities, entity_name)
+        invalid_ref = entity_cls.__module__ + "." + entity_name
         valid_ref = "mlflow.entities.{}".format(entity_name)
         ref_map[invalid_ref] = valid_ref
 
     # Model registry entities
     for entity_name in mlflow.entities.model_registry.__all__:
-        invalid_ref = "mlflow.entities.model_registry.{}.{}".format(
-            _camel_to_snake(entity_name), entity_name
-        )
+        entity_cls = getattr(mlflow.entities.model_registry, entity_name)
+        invalid_ref = entity_cls.__module__ + "." + entity_name
         valid_ref = "mlflow.entities.model_registry.{}".format(entity_name)
         ref_map[invalid_ref] = valid_ref
 
@@ -368,6 +364,8 @@ def _get_reference_map():
 
 
 reference_map = _get_reference_map()
+
+print(reference_map)
 
 
 def resolve_missing_references(app, doctree):
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -338,9 +338,26 @@ nitpick_ignore = [
 
 def _get_reference_map():
     """
-    Gets a dict that maps an invalid reference to a valid one.
+    Sphinx computes references for type annotations using fully-qualified classnames,
+    so references in undocumented modules (even if the referenced object is exposed via
+    a different module from the one it's defined in) are considered invalid by Sphinx.
+
+    Example:
+    ```
+    def start_run(...) -> ActiveRun:
+        # ActiveRun is defined in `mlflow/tracking/fluent.py`
+        ...
+    ```
+
+    For this code, Sphinx tries to create a link for `ActiveRun` using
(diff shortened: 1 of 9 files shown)


mlflow/tracking/fluent.py
```diff
@@ -29,6 +30,13 @@
 from mlflow.utils.validation import _validate_run_id
 from mlflow.utils.annotations import experimental
 
+try:
+    import pandas as pd
+
+    SearchRunsReturnType = Union[pd.DataFrame, List[Run]]
+except ImportError:
+    SearchRunsReturnType = List[Run]
+
```
harupy, line 39: A workaround for the skinny client.

mlflow/tracking/fluent.py
```diff
@@ -861,7 +869,7 @@ def delete_experiment(experiment_id):
     MlflowClient().delete_experiment(experiment_id)
 
 
-def delete_run(run_id):
+def delete_run(run_id) -> None:
```
harupy, line 872: I'm planning to add `-> None` to functions that don't return anything (e.g. `mllfow.set_tags`)

mlflow/tracking/client.py
```diff
@@ -386,7 +389,7 @@ def get_experiment(self, experiment_id):
         """
         return self._tracking_client.get_experiment(experiment_id)
 
-    def get_experiment_by_name(self, name):
+    def get_experiment_by_name(self, name) -> Experiment:
```
smurching, line 392: I think this API might return `None` if no experiment with the provided name exists, would you be able to double check? If so we may want to fix the docstring as well to indicate that behavior.

harupy: Thanks for the catch. Yes, it might return None. I'll wrap `Experiment` with `Optional`.

docs/source/conf.py
```diff
@@ -329,6 +336,56 @@
...
+def _camel_to_snake(x):
+    return re.sub(r"(?<!^)(?=[A-Z])", "_", x).lower()
+
+
+def _get_reference_map():
+    """
+    Gets a dict that maps an invalid reference to a valid one.
+    """
+    ref_map = {
+        "mlflow.tracking.fluent.ActiveRun": "mlflow.ActiveRun",
```
harupy, line 349: Sphinx tries to create a link for `ActiveRun` using its fully qualified class name (`mlflow.tracking.fluent.ActiveRun`), but our doc doesn't contain the reference `mlflow.tracking.fluent.ActiveRun` so it will result in the following warning:

```
/home/circleci/project/mlflow/__init__.py:docstring of mlflow.tracking.fluent.active_run:: WARNING: py:class reference target not found: mlflow.tracking.fluent.ActiveRun
```

As a workaround, map `mlflow.tracking.fluent.ActiveRun` to `mlflow.ActiveRun` (which exists).

harupy:
Offline discussion: Ideally, we should put both argument and return type annotations because Databricks notebooks support both, but we can address them separately. In this PR, just pul return type annotation.

docs/source/conf.py
WeichenXu123: I am thinking for these:
```
mlflow.sklearn.xxx
mflow.pyspark.ml.xx
```
Will it prompot candidate list for possible "xxx" here ?
if no, can we add type hint for attributes like "mlflow.sklearn" ?

docs/source/conf.py
```diff
@@ -329,6 +336,56 @@
...
+        valid_ref = "mlflow.entities.{}".format(entity_name)
+        ref_map[invalid_ref] = valid_ref
+
+    # Model registry entities
+    for entity_name in mlflow.entities.model_registry.__all__:
+        invalid_ref = "mlflow.entities.model_registry.{}.{}".format(
+            _camel_to_snake(entity_name), entity_name
+        )
+        valid_ref = "mlflow.entities.model_registry.{}".format(entity_name)
+        ref_map[invalid_ref] = valid_ref
```
WeichenXu123, line 365: Why we generate the `invalid_ref` by `_camel_to_snake` ? does it always match the rule of "camel --> snake" format ?

I suggest we can generate the original ref by:
```
entity_cls = getattr(mlflow.entities, entity_name)
invalid_ref = entity_cls.__module__ + entity_name
```

harupy: Much better approach! Thanks!

docs/source/conf.py
WeichenXu123: Seems missing some similar cases, such as:

```
mlflow.deployments.BaseDeploymentClient
mflow.models.{Model, ModelSignature, ModelInputExample, infer_signature, FlavorBackend}
mflow.projects.SubmittedRun

mlflow.tracking.MlflowClient
mflow.types.{DataType, ColSpec, Schema, TensorSpec}
```

WeichenXu123: Maybe we can automatically generate the `reference_map`, by searching all `__all__` exports in every sub-module, and for each item in `__all__` , if it is not a function but an Class, then we can add it into `reference_map`, we can generate the original ref by https://github.com/mlflow/mlflow/pull/4242/files#r614140548

harupy: True, but we don't need to address them in this PR. I'll create a follow-up if necessary.

harupy: One challenge is:

```
{"mlflow.tracking.fluent.ActiveRun": "???"}
```

How can we dynamically get `???` ?

mlflow/store/entities/paged_list.py
```diff
@@ -1,4 +1,9 @@
-class PagedList(list):
-    def __init__(self, items, token):
+from typing import TypeVar, List
+
+T = TypeVar("T")
+
+
+class PagedList(List[T]):
+    def __init__(self, items: List[T], token):
```
smurching, line 7: Could we add a short docstring to describe these fields? e.g. something like

"""
Wrapper class around the base Python `List` type. Contains an additional `token`  string attribute
that can be passed to the pagination API that returned this list to fetch additional elements, if any
are available
"""

docs/source/conf.py
```diff
@@ -329,6 +335,54 @@
...
+        invalid_ref = entity_cls.__module__ + "." + entity_name
+        valid_ref = "mlflow.entities.model_registry.{}".format(entity_name)
+        ref_map[invalid_ref] = valid_ref
+
+    return ref_map
+
+
+reference_map = _get_reference_map()
+
+print(reference_map)
```
smurching, line 367:
```suggestion
```

docs/source/conf.py
```diff
@@ -329,6 +335,54 @@
     ("py:class", "mlflow.models.signature.ModelSignature"),
 ]
 
+
+def _get_reference_map():
+    """
+    Gets a dict that maps an invalid reference to a valid one.
```
smurching, line 341: Could we add some context (e.g. basically just your description in https://github.com/mlflow/mlflow/pull/4242/files#r613161818) to why this is necessary in the docstring, for future readers? Namely that Sphinx computes references using fully-qualified classnames, so references in undocumented modules (even if the referenced object is exposed via a different module from the one it's defined in) are considered invalid by Sphinx (you can use the ActiveRun example)

smurching (approved):
Basically LGTM with minor comments

docs/source/conf.py
harupy: thanks for the catch!

docs/source/conf.py
harupy: Thanks for the comment! Added some context!

mlflow/store/entities/paged_list.py
harupy: Done!

harupy:
@smurching Thanks for the review! I'll merge the PR once all the CI checks pass.

docs/source/conf.py
WeichenXu123: > How can we dynamically get < valid_ref > ?

Using module (include the __all__ list) name + item name?
we can search 2 depth level modules

PR #4260, opened 2021-04-19
Parallelize Databricks artifact downloading and uploading

dbczumar:

This PR updates the Databricks artifact repository used for uploading / downloading artifacts on Databricks to make parallel network requests where possible, which is helpful when reading / writing many model types that consist of many small files (otherwise, single-threaded / sequential reading / writing of files doesn't make good use of network bandwidth).

## How is this patch tested?

- Unit tests
- Manual testing on AWS and Azure Databricks to ensure that multi-file uploads / downloads are performed quickly and correctly.

## Release Notes

Improve artifact upload / download performance on Databricks.

Components 

Interface 

Language 

Integrations


arjundc-db (approved):
LGTM.
However, as I have not used concurrency in python, would recommend getting another pair of eyes on this if possible.

mlflow/store/artifact/artifact_repo.py
```diff
@@ -144,6 +159,9 @@ def _download_file(self, remote_file_path, local_path):
         :param remote_file_path: Source path to the remote file, relative to the root
                                  directory of the artifact repository.
         :param local_path: The path to which to save the downloaded file.
+        :return: Either `None` (for synchronous download implementations) or an instance of
```
jinzhang21, line 162: Returning 2 types of objects depending on the implementation is confusing. I suggest not to touch this API. Instead, add a new non-abstract method, which can be used in future for other ArtifactRepo subclasses. Below code might not work as is, please pardon the lack of testing.

def _download_files(self, remote_file_paths, local_path):
    exceptions = []
    for f in concurrent.futures.as_completed([self.thread_pool.submit(
            self._download_file, path, local_path) for path in remote_file_paths]):
        try:
          f.result()
        except Exception as e:
           exceptions.append(e)
      if exceptions:
          ...  # Throw aggregated exception for potentially multiple download errors.

jinzhang21 (commented):
Looks good overall. I have some minor comments on the implementation details.

Ideally, asyncio is my preferred choice since it doesn't really involve multi-threading, but given it's only supported in Python 3.7+. We'd better to wait another 1 - 2 years before moving forward to adapt MLflow with asyncio.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -98,6 +99,11 @@ def __init__(self, artifact_uri):
         self.run_relative_artifact_repo_root_path = (
             "" if run_artifact_root_path == artifact_repo_root_path else run_relative_root_path
         )
+        # Limit the number of threads used for artifact uploads, using at most `8` threads or
+        # 2 * the number of CPU cores available on the system (whichever is smaller)
+        num_cpus = os.cpu_count() or 4
+        num_artifact_workers = min(num_cpus * 2, 8)
+        self.thread_pool = ThreadPoolExecutor(max_workers=num_artifact_workers)
```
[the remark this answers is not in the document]
apurva-koti: Without being very knowledgeable thread pools and their resource overhead, I feel like 8 is a big improvement purely because we currently have no parallelization. Not sure if there's a good reason to overdo it from the get-go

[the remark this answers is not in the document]
dbczumar: I took a look at the average number of artifact files on Databricks, and the p25 - p75 number of files is 3, while the p95 is 4. The p99 is over 2000. `min(8, 2 * cpus)` seems okay given that knowledge.

tests/store/artifact/test_artifact_repo.py
```diff
@@ -1,6 +1,6 @@
 import posixpath
-from unittest import mock
 import pytest
+from unittest import mock
```
dbczumar, line 3: @apurva-koti can we revert this change? Doesn't seem necessary

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -273,7 +280,13 @@ def log_artifacts(self, local_dir, artifact_path=None):
...
             for name in filenames:
                 file_path = os.path.join(dirpath, name)
-                self.log_artifact(file_path, artifact_subdir)
+                upload_future = self.thread_pool.submit(
+                    self.log_artifact, file_path, artifact_subdir
+                )
+                upload_futures.append(upload_future)
+
+        for upload_future in upload_futures:
+            upload_future.result()
```
dbczumar, line 289: Looks great! :)

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -318,7 +331,9 @@ def _download_file(self, remote_file_path, local_path):
             self.run_relative_artifact_repo_root_path, remote_file_path
         )
         read_credentials = self._get_read_credentials(self.run_id, run_relative_remote_file_path)
-        self._download_from_cloud(read_credentials.credentials, local_path)
+        return self.thread_pool.submit(
+            self._download_from_cloud, read_credentials.credentials, local_path,
+        )
```
dbczumar, line 336: @apurva-koti I don't think this will work because the implementation of `ArtifactRepository.download_artifacts()` assumes that this method is synchronous (https://github.com/mlflow/mlflow/blob/df8ba4e440a260c4b78294a735be0bfe1fe94902/mlflow/store/artifact/artifact_repo.py#L90).

As a result, calls such as `mlflow.pyfunc.load_model()` that attempt to load file contents into memory immediately after the invocation of `download_artifacts()` will likely fail, since the artifacts aren't actually present.

Additionally, this approach only provides parallelism on a per-directory level.

Instead, we should override the entire `download_artifacts()` method in `DatabricksArtifactRepository` and provide a custom implementation.

apurva-koti: Ohh yeah that makes sense. Let's override `download_artifacts` to keep all this special parallelism contained within `databricks_artifact_repo`

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -320,5 +339,88 @@ def _download_file(self, remote_file_path, local_path):
...
+        """
+        Parallelized implementation of `download_artifacts` for Databricks.
+        """
+
+        def download_file(fullpath):
+            """
+            Submit a download of `fullpath` to the thread pool and return the
+            resultant Future.
+            """
+            fullpath = fullpath.rstrip("/")
```
jinzhang21, line 352: How about changing this to
full_path = os.path.abspath(full_path)?
It handles more cases than ending '/'.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -320,5 +339,88 @@ def _download_file(self, remote_file_path, local_path):
...
+            dirpath, _ = posixpath.split(fullpath)
```
jinzhang21, line 353: Prefer dir_path = os.path.dirname(full_path)

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -273,7 +286,13 @@ def log_artifacts(self, local_dir, artifact_path=None):
                 artifact_subdir = posixpath.join(artifact_path, rel_path)
             for name in filenames:
                 file_path = os.path.join(dirpath, name)
-                self.log_artifact(file_path, artifact_subdir)
+                upload_future = self.thread_pool.submit(
+                    self.log_artifact, file_path, artifact_subdir
+                )
+                upload_futures.append(upload_future)
+
+        for upload_future in upload_futures:
```
jinzhang21, line 294: The issue of iterating through all futures for result, is that when there's multiple errors, only the first one will be raised. Consider writing a util function to capture exceptions from all futures.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -320,5 +339,88 @@ def _download_file(self, remote_file_path, local_path):
...
 
+    def download_artifacts(self, artifact_path, dst_path=None):
+        """
+        Parallelized implementation of `download_artifacts` for Databricks.
+        """
+
+        def download_file(fullpath):
+            """
+            Submit a download of `fullpath` to the thread pool and return the
+            resultant Future.
```
jinzhang21, line 350: It looks the actual return is a tuple of downloaded local file path and its corresponding future.
In fact if a dict of {file_path: future} is returned here, the return signature will be the same for both download_file() and download_artifact_dir(). The handling of futures can be combined then (see below comment).

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -320,5 +339,88 @@ def _download_file(self, remote_file_path, local_path):
...
+                error_code=INVALID_PARAMETER_VALUE,
+            )
+
+        if self._is_directory(artifact_path):
+            local_dir, futures = download_artifact_dir(artifact_path)
+            # Join futures to ensure that all files in the directory have
+            # been downloaded prior to returning
+            for future in futures:
+                future.result()
+            return local_dir
```
jinzhang21, line 418: This part can be combined with L422 if the returning signature is a dict.

In addition, see comment on L294 for util functions to handle multiple failures.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -320,5 +339,88 @@ def _download_file(self, remote_file_path, local_path):
         read_credentials = self._get_read_credentials(self.run_id, run_relative_remote_file_path)
         self._download_from_cloud(read_credentials.credentials, local_path)
 
+    def download_artifacts(self, artifact_path, dst_path=None):
+        """
+        Parallelized implementation of `download_artifacts` for Databricks.
+        """
+
+        def download_file(fullpath):
```
jinzhang21, line 347: prefer to pass dst_path as an argument explicitly.

fullpath is confusing. Perhaps src_file for better readability. ditto below for dir_path -> src_dir etc.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -320,5 +339,88 @@ def _download_file(self, remote_file_path, local_path):
...
+            if not os.path.exists(local_dir_path):
+                os.makedirs(local_dir_path)
+            return (
+                local_file_path,
+                self.thread_pool.submit(
+                    self._download_file, remote_file_path=fullpath, local_path=local_file_path
+                ),
+            )
+
+        def download_artifact_dir(dir_path):
```
jinzhang21, line 365: prefer to pass dst_path as an argument explicitly.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -320,5 +339,88 @@ def _download_file(self, remote_file_path, local_path):
...
+            download_futures = []
+            local_dir = os.path.join(dst_path, dir_path)
+            dir_content = [  # prevent infinite loop, sometimes the dir is recursively included
+                file_info
+                for file_info in self.list_artifacts(dir_path)
+                if file_info.path != "." and file_info.path != dir_path
+            ]
+            if not dir_content:  # empty dir
+                if not os.path.exists(local_dir):
+                    os.makedirs(local_dir)
```
jinzhang21, line 380: What's the reason to create a local dir in this case?

jinzhang21 (commented):
Thanks for the change, Corey. The logic LGTM. Added a couple of comments, mostly for readability reasons.

mlflow/store/artifact/databricks_artifact_repo.py
dbczumar: I like this idea! In addition to a collection of futures, we also need to maintain the top-level destination directory that should be returned by `download_artifacts()`, hence the first element of the tuple.

To achieve consistent return types and deduplicate the logic, I've modified the signature of each method to return a documented `InflightDownload` named tuple whose first element is the top-level destination directory and whose second element is the mapping you outlined above.

This mapping you suggested has the added benefit of enabling us to provide failure information for each specific path in the event of one or more errors.

mlflow/store/artifact/databricks_artifact_repo.py
dbczumar: Great idea! Addressed! (See above comments).

dbczumar (commented):
@jinzhang21 Thank you for the thorough review! I've addressed your comments and added some test cases. Can you take another pass?

jinzhang21 (approved):
The change looks great!

PR #4263, opened 2021-04-19
Implementation: pyspark ML autologging for pipeline

WeichenXu123:

Implementation: pyspark ML autologging for pipeline

## How is this patch tested?

Unit tests.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -80,14 +80,16 @@ def _get_warning_msg_for_skip_log_model(model):
 
 def _should_log_model(spark_model):
     from pyspark.ml.base import Model
+
     # TODO: Handle PipelineModel/CrossValidatorModel/TrainValidationSplitModel
     class_name = _get_fully_qualified_class_name(spark_model)
     if class_name in _log_model_allowlist:
         if class_name == "pyspark.ml.classification.OneVsRestModel":
             return _should_log_model(spark_model.models[0])
-        elif class_name == 'pyspark.ml.pipeline.PipelineModel':
-            return all(_should_log_model(stage) for stage in spark_model.stages
-                       if isinstance(stage, Model))
+        elif class_name == "pyspark.ml.pipeline.PipelineModel":
+            return all(
+                _should_log_model(stage) for stage in spark_model.stages if isinstance(stage, Model)
+            )
         else:
             return True
     else:
@@ -134,7 +136,7 @@ def _get_instance_param_map(instance):
     is_pipeline = isinstance(instance, Pipeline)
 
     for k, v in param_map.items():
-        if is_pipeline and k == 'stages':
+        if is_pipeline and k == "stages":
             expanded_param_map[k] = _get_pipeline_stage_hierarchy(instance)[instance.uid]
         elif isinstance(v, Params):
             # handle the case param value type inherits `pyspark.ml.param.Params`
@@ -246,13 +248,12 @@ def autolog(
         param_map = _get_instance_param_map(estimator)
         if isinstance(estimator, Pipeline):
             pipeline_hierarchy = _get_pipeline_stage_hierarchy(estimator)
-            try_mlflow_log(mlflow.log_dict, pipeline_hierarchy,
-                           artifact_file='pipeline_hierarchy.json')
+            try_mlflow_log(
+                mlflow.log_dict, pipeline_hierarchy, artifact_file="pipeline_hierarchy.json"
+            )
 
         # Chunk model parameters to avoid hitting the log_batch API limit
-        for chunk in _chunk_dict(
-            param_map, chunk_size=MAX_PARAMS_TAGS_PER_BATCH,
-        ):
+        for chunk in _chunk_dict(param_map, chunk_size=MAX_PARAMS_TAGS_PER_BATCH,):
             truncated = _truncate_dict(chunk, MAX_ENTITY_KEY_LENGTH, MAX_PARAM_VAL_LENGTH)
             try_mlflow_log(mlflow.log_params, truncated)
 
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -132,7 +132,7 @@ def _get_instance_param_map(instance):
     return expanded_param_map
 
 
-def _get_pipeline_stage_hierarchy_and_params(pipeline_stages):
+def _get_pipeline_stage_hierarchy_and_params(pipeline):
     from pyspark.ml import Pipeline
     """
     :param pipeline_stages: stages of a pyspark ML pipeline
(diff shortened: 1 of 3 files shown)


mlflow/pyspark/ml/__init__.py
```diff
@@ -218,9 +240,17 @@ def _log_pretraining_metadata(estimator, params):
         if params and isinstance(params, dict):
             estimator = estimator.copy(params)
 
+        param_map = _get_instance_param_map(estimator)
+        if isinstance(estimator, Pipeline):
+            pipeline_hyerarchy, stages_param_maps = _get_pipeline_stage_hierarchy_and_params(
+                estimator)
+            param_map.update(stages_param_maps)
+            try_mlflow_log(mlflow.log_dict, pipeline_hyerarchy,
+                           artifact_file='pipeline_hyerarchy.json')
```
harupy, line 249: Offline discussion: `pipeline_hyerarchy.json` should look like this:

```json
{
  "Pipeline_db95ddc399a0": [
    "Tokenizer_390f8f5e9b72",
    {
      "Pipeline_db95ddc399a0": [
        "HashingTF_40849cb2ccd3",
        "LogisticRegression_b95ce4c49c0e"
      ]
    }
  ]
}
```

mlflow/pyspark/ml/__init__.py
```diff
@@ -218,9 +244,17 @@ def _log_pretraining_metadata(estimator, params):
...
```
dbczumar, line 253: Nit: 
```suggestion
                           artifact_file='pipeline_hierarchy.json')
```

mlflow/pyspark/ml/__init__.py
```diff
@@ -132,6 +136,27 @@ def _get_instance_param_map(instance):
...
+    for stage in pipeline_stages:
+        if isinstance(stage, Pipeline):
+            hierarchy_elem, stage_param_map = _get_pipeline_stage_hierarchy_and_params(stage)
+        else:
+            hierarchy_elem = stage.uid
+            stage_param_map = {f'{stage.uid}.{k}': v
+                               for k, v in _get_instance_param_map(stage).items()}
+        stage_hierarchy.append(hierarchy_elem)
+        param_map.update(stage_param_map)
+    return {pipeline.uid: stage_hierarchy}, param_map
```
dbczumar, line 157: @WeichenXu123 One thing I noticed about this construction is that we log the parameters for each nested pipeline stage, but the `stages` parameter for a nested pipeline only refers to the top-level pipeline (e.g. `[Tokenizer_112334721b1b, Pipeline_07c6a754c9a8]`).

This makes it a bit difficult to figure out where the stages of `Pipeline_07c6a754c9a8`, such as `HashingTF_59b3cf717aab` and `LogisticRegression_3b5baf5ff098` in the example you linked offline, come from.

Could we consider updating the `stages` parameter so that it expands each pipeline?

E.g. `{"stages": [Tokenizer_112334721b1b, Pipeline_07c6a754c9a8]}` would become `{"stages": [Tokenizer_112334721b1b, [HashingTF_59b3cf717aab, LogisticRegression_3b5baf5ff098]}`

dbczumar: If the pipeline uid is important, we could also include it in the `stages` hierarchy  (e.g. `Pipeline_07c6a754c9a8: [HashingTF_59b3cf717aab, LogisticRegression_3b5baf5ff098]`

mlflow/pyspark/ml/__init__.py
```diff
@@ -79,11 +79,15 @@ def _get_warning_msg_for_skip_log_model(model):
...
 
 def _should_log_model(spark_model):
+    from pyspark.ml.base import Model
     # TODO: Handle PipelineModel/CrossValidatorModel/TrainValidationSplitModel
     class_name = _get_fully_qualified_class_name(spark_model)
     if class_name in _log_model_allowlist:
         if class_name == "pyspark.ml.classification.OneVsRestModel":
             return _should_log_model(spark_model.models[0])
+        elif class_name == 'pyspark.ml.pipeline.PipelineModel':
+            return all(_should_log_model(stage) for stage in spark_model.stages
```
dbczumar, line 89: Are there any non-Model stages (that aren't instances of `pyspark.ml.base.Model`) that we should exclude due to size?

mlflow/pyspark/ml/__init__.py
WeichenXu123: Then it become the same with `pipeline_hyerarchy.json`, so shall we skip log `stages` param ?

mlflow/pyspark/ml/__init__.py
WeichenXu123: No.
The pipelineModel will only include stages of Model or Transformer, if it is Transformer, it will only include some parameters and it will be small size.

mlflow/pyspark/ml/__init__.py
dbczumar: If anything, I think we could omit the hierarchy artifact and keep the stage parameter. Let me know if you disagree!

harupy: Should we keep the hierarchy artifact, the stages hierarchy tends to be long and easily exceeds the maximum param value length (250) and will be truncated?

tests/spark_autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -245,3 +278,36 @@ def test_param_map_captures_wrapped_params(dataset_binomial):
...
+    }
+
+    for estimator in [pipeline, nested_pipeline]:
+        with mlflow.start_run() as run:
+            model = estimator.fit(dataset_text)
+
+        run_id = run.info.run_id
+        run_data = get_run_data(run_id)
+        assert run_data.params == truncate_param_dict(
+            stringify_dict_values(_get_instance_param_map(estimator))
```
dbczumar, line 307: @WeichenXu123 can we add some test coverage for `_get_instance_param_map` to ensure that it's producing a reasonable set of parameters? Otherwise, this test would pass even if the parameters returned by `_get_instance_param_map` were broken / incorrect.

mlflow/pyspark/ml/__init__.py
```diff
@@ -111,27 +132,42 @@ def _get_instance_param_map(instance):
         if instance.isDefined(param)
     }
     expanded_param_map = {}
+
+    is_pipeline = isinstance(instance, Pipeline)
+    is_parameter_search_estimator = isinstance(instance, (CrossValidator, TrainValidationSplit))
+
     for k, v in param_map.items():
```
harupy, line 139:
```suggestion
    for param_name, param_value in param_map.items():
```

Can we rename these variables for better readability?

dbczumar (approved):
LGTM! Thanks @WeichenXu123 ! Tried this out manually with some non-nested, nested, and doubly-nested pipelines. Seems to work like a charm!

harupy (approved):
LGTM! Thanks for all the hard work!

PR #4270, opened 2021-04-20
Implementation: autologging for hyperparameter-tuning

WeichenXu123:

Implementation: autologging for hyperparameter-tuning

## How is this patch tested?

Unit tests.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -430,7 +430,7 @@ def autolog(
             best_param_map = estimator_param_maps[best_index]
             try_mlflow_log(
                 mlflow.log_dict,
-                search_result,
+                best_param_map,
                 artifact_file="best_parameters.json"
             )
 
@@ -440,9 +440,12 @@ def autolog(
                 try_mlflow_log(
                     mlflow.spark.log_model, spark_model, artifact_path="model",
                 )
-                try_mlflow_log(
-                    mlflow.spark.log_model, spark_model.bestModel, artifact_path="best_model",
-                )
+                if _is_parameter_search_model(spark_model):
+                    try_mlflow_log(
+                        mlflow.spark.log_model,
+                        spark_model.bestModel,
+                        artifact_path="best_model",
+                    )
             else:
                 _logger.warning(_get_warning_msg_for_skip_log_model(spark_model))
 
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -454,9 +454,9 @@ def autolog(
           calls ``fit()``, it internally calls ``fit()`` on its child estimators. Autologging
           does NOT perform logging on these constituent ``fit()`` calls.
 
-          A "estimator_info.json" artifact will be logged, the artifact include an item call
-          "hierarchy" which describe the hierarchy of the meta estimators. The hierarchy will
-          expand all nested stage such as nested pipeline stages.
+          A "estimator_info.json" artifact is logged, which includes a `hierarchy` entry
+          describing the hierarchy of the meta estimator. The hierarchy includes expanded
+          entries for all nested stages, such as nested pipeline stages.
 
       **Parameter search**
           In addition to recording the information discussed above, autologging for parameter
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -559,7 +559,9 @@ def autolog(
                 )
                 _logger.warning(msg)
 
-            estimator_param_maps = _get_tuning_param_maps(estimator)
+            estimator_param_maps = _get_tuning_param_maps(
+                estimator, estimator._autologging_metadata.uid2name_map
+            )
 
             metric_key, metrics = _get_param_search_metrics(estimator, spark_model)
             _log_parameter_search_results_as_artifact(
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -408,7 +408,7 @@ def autolog(
                 metrics = spark_model.avgMetrics
 
             search_result = []
-            for i in len(estimator_param_maps):
+            for i in range(len(estimator_param_maps)):
                 search_result.append({
                     'param_maps': estimator_param_maps[i],
                     metric_key: metrics[i]
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -585,7 +585,7 @@ def autolog(
 
             # Log best_param_map as autologging parameters as well
             _log_estimator_params({f'best_{param_name}': param_value
-                                   for param_name, param_value in best_param_map})
+                                   for param_name, param_value in best_param_map.items()})
 
         if log_models:
             if _should_log_model(spark_model):
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -176,7 +176,7 @@ def _gen_instance_metadata_recusively(
         return {'name': instance_name, 'stages': sub_stages}
     elif isinstance(instance, OneVsRest):
         classifier = _gen_instance_metadata_recusively(
-            instance, uid2name_map, cls2count_map, cls2first_uid_map
+            instance, uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
         )
         return {'name': instance_name, 'classifier': classifier}
     elif _is_parameter_search_estimator(instance):
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -1,4 +1,4 @@
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 import logging
 import numpy as np
 import time
@@ -138,64 +138,75 @@ def _should_log_hierarchy(estimator):
 
 
 AutologgingEstimatorMetadata = namedtuple('AutologgingEstimatorMetadata', [
-    'hierarchy', 'uid2name_map', 'param_search_estimators'
+    'hierarchy', 'uid_to_indexed_name_map', 'param_search_estimators'
 ])
 
 
-def _gen_instance_metadata_recursively(
-        instance,
-        uid2name_map,
-        cls2count_map,
-        cls2first_uid_map,
-        param_search_estimators,
+def _traverse_stage(stage):
+    from pyspark.ml import Pipeline
+    from pyspark.ml.classification import OneVsRest
+
+    yield stage
+    if isinstance(stage, Pipeline):
+        for stage in stage.getStages():
+            yield from _traverse_stage(stage)
+    elif isinstance(stage, OneVsRest):
+        yield from _traverse_stage(stage.getClassifier())
+    elif _is_parameter_search_estimator(stage):
+        yield from _traverse_stage(stage.getEstimator())
+        yield from _traverse_stage(stage.getEvaluator())
+
+
+def _get_uid_to_indexed_name_map(estimator):
+    counter = defaultdict(int)
+    uid_to_classname_and_count = {}
+    for child in _traverse_stage(estimator):
+        class_name = child.__class__.__name__
+        counter[class_name] += 1
+        uid_to_classname_and_count[child.uid] = (class_name, counter[class_name])
+    return {
+        uid: f"{class_name}_{count}" if counter[class_name] > 1 else class_name
+        for uid, (class_name, count) in uid_to_classname_and_count.items()
+    }
+
+
+def _gen_stage_hierarchy_recursively(
+        stage,
+        uid_to_indexed_name_map,
 ):
     from pyspark.ml import Pipeline
     from pyspark.ml.classification import OneVsRest
 
-    def gen_instance_name(instance):
-        stage_cls = instance.__class__.__name__
-        stage_uid = instance.uid
+    stage_name = uid_to_indexed_name_map[stage.uid]
 
-        if stage_cls not in cls2count_map:
-            cls2count_map[stage_cls] = 1
-            uid2name_map[stage_uid] = stage_cls
-            cls2first_uid_map[stage_cls] = stage_uid
-        else:
-            cls2count_map[stage_cls] += 1
-            uid2name_map[stage_uid] = f'{stage_cls}_{cls2count_map[stage_cls]}'
-
-        return uid2name_map[stage_uid]
-
-    instance_name = gen_instance_name(instance)
-    if isinstance(instance, Pipeline):
+    if isinstance(stage, Pipeline):
         sub_stages = []
-        for sub_stage in instance.getStages():
-            sub_stages.append(_gen_instance_metadata_recursively(
-                sub_stage, uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
-            ))
-        return {'name': instance_name, 'stages': sub_stages}
-    elif isinstance(instance, OneVsRest):
-        classifier = _gen_instance_metadata_recursively(
-            instance.getClassifier(),
-            uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
-        )
-        return {'name': instance_name, 'classifier': classifier}
-    elif _is_parameter_search_estimator(instance):
-        param_search_estimators.append(instance)
-        evaluator = instance.getEvaluator()
-        tuned_estimator = instance.getEstimator()
-        return {
-            'name': instance_name,
-            'evaluator': _gen_instance_metadata_recursively(
-                evaluator, uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
-            ),
-            'tuned_estimator': _gen_instance_metadata_recursively(
-                tuned_estimator, uid2name_map, cls2count_map, cls2first_uid_map,
-                param_search_estimators
+        for sub_stage in stage.getStages():
+            sub_hierarchy = _gen_stage_hierarchy_recursively(
+                sub_stage, uid_to_indexed_name_map
             )
+            sub_stages.append(sub_hierarchy)
+        return {'name': stage_name, 'stages': sub_stages}
+    elif isinstance(stage, OneVsRest):
+        classifier_hierarchy = _gen_stage_hierarchy_recursively(
+            stage.getClassifier(), uid_to_indexed_name_map
+        )
+        return {'name': stage_name, 'classifier': classifier_hierarchy}
+    elif _is_parameter_search_estimator(stage):
+        evaluator = stage.getEvaluator()
+        tuned_estimator = stage.getEstimator()
+        tuned_estimator_hierarchy = _gen_stage_hierarchy_recursively(
+            tuned_estimator, uid_to_indexed_name_map
+        )
+        return {
+            'name': stage_name,
+            'evaluator': _gen_stage_hierarchy_recursively(
+                evaluator, uid_to_indexed_name_map
+            )[0],
+            'tuned_estimator': tuned_estimator_hierarchy
         }
     else:
-        return instance_name
+        return {'name': stage_name}
 
 
 def _gen_estimator_metadata(estimator):
@@ -204,35 +215,21 @@ def _gen_estimator_metadata(estimator):
     The AutologgingEstimatorMetadata object includes:
      - hierarchy: the hierarchy of the estimator, it will expand
          pipeline/meta estimator/param tuning estimator
-     - uid2name_map: a map of `uid` -> `name`, each nested instance uid will be
+     - uid_to_indexed_name_map: a map of `uid` -> `name`, each nested instance uid will be
          mapped to a fixed name. The naming rule is using `{class_name}` if the
          instance type occurs once, or `{class_name}_{index}` if the instance type occurs
          multiple times. The index order is in line with depth-first traversing.
      - param_search_estimators: a list includes all param search estimators in the
          hierarchy tree.
     """
-    uid2name_map = {}
-    cls2count_map = {}
-    cls2first_uid_map = {}
-    param_search_estimators = []
-
-    hierarchy = _gen_instance_metadata_recursively(
-        estimator, uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
-    )
-    for cls, first_uid in cls2first_uid_map.items():
-        if cls2count_map[cls] > 1:
-            # change the first stage name from "{class_name}" to "{class_name}_1"
-            uid2name_map[first_uid] = f'{cls}_1'
-
-    for i in range(len(param_search_estimators)):
-        name = param_search_estimators[i]
-        if name in cls2count_map and cls2count_map[name] > 1:
-            # change the estimator name from "{class_name}" to "{class_name}_1"
-            param_search_estimators[i] = f'{name}_1'
+    uid_to_indexed_name_map = _get_uid_to_indexed_name_map(estimator)
+    param_search_estimators = [stage for stage in _traverse_stage(estimator)
+                               if _is_parameter_search_estimator(stage)]
+    hierarchy = _gen_stage_hierarchy_recursively(estimator, uid_to_indexed_name_map)
 
     metadata = AutologgingEstimatorMetadata(
         hierarchy=hierarchy,
-        uid2name_map=uid2name_map,
+        uid_to_indexed_name_map=uid_to_indexed_name_map,
         param_search_estimators=param_search_estimators
     )
     estimator._autologging_metadata = metadata
@@ -240,7 +237,7 @@ def _gen_estimator_metadata(estimator):
     return metadata
 
 
-def _get_instance_param_map_recursively(instance, level, uid2name_map):
+def _get_instance_param_map_recursively(instance, level, uid_to_indexed_name_map):
     from pyspark.ml.param import Params
     from pyspark.ml.pipeline import Pipeline
 
@@ -257,22 +254,22 @@ def _get_instance_param_map_recursively(instance, level, uid2name_map):
     if level == 0:
         logged_param_name_prefix = ''
     else:
-        logged_param_name_prefix = uid2name_map[instance.uid] + '.'
+        logged_param_name_prefix = uid_to_indexed_name_map[instance.uid] + '.'
 
     for param_name, param_value in param_map.items():
         logged_param_name = logged_param_name_prefix + param_name
 
         if is_pipeline and param_name == "stages":
             expanded_param_map[logged_param_name] = [
-                uid2name_map[stage.uid] for stage in instance.getStages()
+                uid_to_indexed_name_map[stage.uid] for stage in instance.getStages()
             ]
             for stage in instance.getStages():
                 stage_param_map = _get_instance_param_map_recursively(
-                    stage, level + 1, uid2name_map
+                    stage, level + 1, uid_to_indexed_name_map
                 )
                 expanded_param_map.update(stage_param_map)
         elif is_parameter_search_estimator and param_name == 'estimator':
-            expanded_param_map[logged_param_name] = uid2name_map[param_value.uid]
+            expanded_param_map[logged_param_name] = uid_to_indexed_name_map[param_value.uid]
             # skip log estimator's nested params because they will be logged as JSON artifact,
             # and they will be logged in nested runs as well.
             pass
@@ -283,9 +280,9 @@ def _get_instance_param_map_recursively(instance, level, uid2name_map):
             # handle the case param value type inherits `pyspark.ml.param.Params`
             # e.g. param like
             # `OneVsRest.classifier`/`CrossValidator.evaluator`
-            expanded_param_map[logged_param_name] = uid2name_map[param_value.uid]
+            expanded_param_map[logged_param_name] = uid_to_indexed_name_map[param_value.uid]
             internal_param_map = _get_instance_param_map_recursively(
-                param_value, level + 1, uid2name_map
+                param_value, level + 1, uid_to_indexed_name_map
             )
             expanded_param_map.update(internal_param_map)
         else:
@@ -294,9 +291,9 @@ def _get_instance_param_map_recursively(instance, level, uid2name_map):
     return expanded_param_map
 
 
-def _get_instance_param_map(instance, uid2name_map, ):
+def _get_instance_param_map(instance, uid_to_indexed_name_map):
     return _get_instance_param_map_recursively(
-        instance, level=0, uid2name_map=uid2name_map
+        instance, level=0, uid_to_indexed_name_map=uid_to_indexed_name_map
     )
 
 
@@ -328,7 +325,7 @@ def _create_child_runs_for_parameter_search(parent_estimator, parent_model, pare
         )
 
         params_to_log = _get_instance_param_map(
-            child_estimator, parent_estimator._autologging_metadata.uid2name_map
+            child_estimator, parent_estimator._autologging_metadata.uid_to_indexed_name_map
         )
         param_batches_to_log = _chunk_dict(params_to_log, chunk_size=MAX_PARAMS_TAGS_PER_BATCH)
         metrics_to_log = {metric_key: metrics[i]}
@@ -381,10 +378,10 @@ def _get_warning_msg_for_fit_call_with_a_list_of_params(estimator):
     )
 
 
-def _get_tuning_param_maps(param_search_estimator, uid2name_map):
+def _get_tuning_param_maps(param_search_estimator, uid_to_indexed_name_map):
     tuning_param_maps = []
     for eps in param_search_estimator.getEstimatorParamMaps():
-        tuning_param_maps.append({f'{uid2name_map[k.parent]}.{k.name}': str(v)
+        tuning_param_maps.append({f'{uid_to_indexed_name_map[k.parent]}.{k.name}': str(v)
                                   for k, v in eps.items()})
     return tuning_param_maps
 
@@ -404,6 +401,15 @@ def _get_param_search_metrics(param_search_estimator, param_search_model):
         raise RuntimeError(f'Unknown parameter search model type {type(param_search_model)}.')
 
     return metric_key, metrics
+
+
+def _log_estimator_params(param_map):
+    # Chunk model parameters to avoid hitting the log_batch API limit
+    for chunk in _chunk_dict(
+            param_map, chunk_size=MAX_PARAMS_TAGS_PER_BATCH,
+    ):
+        truncated = _truncate_dict(chunk, MAX_ENTITY_KEY_LENGTH, MAX_PARAM_VAL_LENGTH)
+        try_mlflow_log(mlflow.log_params, truncated)
 
 
 @experimental
@@ -504,23 +510,23 @@ def autolog(
 
         artifact_dict = {}
 
-        param_map = _get_instance_param_map(estimator, autologging_metadata.uid2name_map)
+        param_map = _get_instance_param_map(estimator, autologging_metadata.uid_to_indexed_name_map)
         if _should_log_hierarchy(estimator):
             artifact_dict['hierarchy'] = autologging_metadata.hierarchy
 
         for param_search_estimator in autologging_metadata.param_search_estimators:
             param_search_estimator_name = \
-                f'{autologging_metadata.uid2name_map[param_search_estimator.uid]}'
+                f'{autologging_metadata.uid_to_indexed_name_map[param_search_estimator.uid]}'
             artifact_dict[param_search_estimator_name] = {}
 
             artifact_dict[param_search_estimator_name]['tuning_parameter_map_list'] = \
                 _get_tuning_param_maps(
-                    param_search_estimator, autologging_metadata.uid2name_map
+                    param_search_estimator, autologging_metadata.uid_to_indexed_name_map
                 )
 
             artifact_dict[param_search_estimator_name]['tuned_estimator_parameter_map'] = \
                 _get_instance_param_map_recursively(
-                    param_search_estimator.getEstimator(), 1, autologging_metadata.uid2name_map
+                    param_search_estimator.getEstimator(), 1, autologging_metadata.uid_to_indexed_name_map
                 )
 
         if artifact_dict:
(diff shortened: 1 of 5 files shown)


mlflow/pyspark/ml/__init__.py
```diff
@@ -101,37 +111,115 @@ def _get_estimator_info_tags(estimator):
...
-        if isinstance(v, Params):
+        if is_tuning_estimator and k == 'estimator':
+            expanded_param_map[k] = v.uid
+            # skip log estimator's nested params because they will be
+            # logged in nested runs.
+            # TODO:
+            #  consider the nested case:
+            #  the tuned estimator is a pipeline or a composite estimator like OneVsRest
+            #  then nested estimators uid need logged, e.g. uids of stage in
+            #  pipeline.stages/ova.classifier
```
WeichenXu123, line 143: Discussion:
For this case, shall we log the hierarchy of the estimator ?
e.g. , CrossValidator tune on a pipeline, and pipeline includes many stages,
and the `estimatorParamMaps` include map of `Param --> value`, the Param here may be a param of a stage in the pipeline, and we log the param name as "{Param.parent}.{param_name}`, the `Param.parent` here is the uid of the stage(in the pipeline),

So log a hierarchy (includes all nested stage uids) for the tuned estimator is valuable.

WeichenXu123: e.g. this example code:
https://github.com/apache/spark/blob/361444890e23057d32e737d88e8efa4fafde7ed3/examples/src/main/python/ml/cross_validator.py#L71

For the `lr.regParam` in autologging artifacts `estimatorParamMaps`, we will log an item of `{lr.uid}.regParam`

mlflow/pyspark/ml/__init__.py
```diff
@@ -101,37 +111,115 @@ def _get_estimator_info_tags(estimator):
...
+    # start time of child runs, since we cannot precisely determine when each point
+    # in the parameter search space was explored
+    child_run_start_time = parent_run.info.start_time
+    child_run_end_time = int(time.time() * 1000)
+
+    estimator_param_maps = parent_estimator.getEstimatorParamMaps()
+    tuning_estimator = parent_estimator.getEstimator()
+    metric_key = parent_estimator.getEvaluator().getMetricName()
+    if isinstance(parent_model, CrossValidatorModel):
+        metric_key = 'avg_' + metric_key
```
WeichenXu123, line 176: Note: I add an `avg_` prefix for CrossValidator case.

mlflow/pyspark/ml/__init__.py
```diff
@@ -101,37 +111,115 @@ def _get_estimator_info_tags(estimator):
...
+        metrics = parent_model.avgMetrics
+    else:
+        raise RuntimeError(f'Unknown model type {type(parent_model)}.')
+
+    for i in range(len(estimator_param_maps)):
+        estimator = tuning_estimator.copy(estimator_param_maps[i])
+        tags_to_log = dict(child_tags) if child_tags else {}
+        tags_to_log.update({MLFLOW_PARENT_RUN_ID: parent_run.info.run_id})
+        tags_to_log.update(_get_estimator_info_tags(estimator))
+        # TODO: add child run index into tag ?
```
WeichenXu123, line 188: Shall we log index ?

mlflow/pyspark/ml/__init__.py
```diff
@@ -84,6 +92,8 @@ def _should_log_model(spark_model):
     if class_name in _log_model_allowlist:
         if class_name == "pyspark.ml.classification.OneVsRestModel":
             return _should_log_model(spark_model.models[0])
+        elif _is_parameter_search_model(spark_model):
+            return _should_log_model(spark_model.bestModel)
```
WeichenXu123, line 96: Note: check nested model

mlflow/pyspark/ml/__init__.py
```diff
@@ -218,18 +302,48 @@ def _log_pretraining_metadata(estimator, params):
...
+                )
+            except Exception as e:
+                import traceback
+                msg = (
+                    "Encountered exception during creation of child runs for parameter search."
+                    " Child runs may be missing. Exception: {}".format(traceback.format_exc())
+                )
+                _logger.warning(msg)
+
+        # TODO: log parameter search results (best model parameters) as an artifact ?
```
WeichenXu123, line 345: @harupy Thoughts ?

dbczumar: If it helps, the scikit-learn autologging integration for GridSearch logs the best model parameters to the top-level (parent) run at the end of training. I think it would make sense to do the same here.

mlflow/pyspark/ml/__init__.py
harupy: Is there any useful use case?

mlflow/pyspark/ml/__init__.py
dbczumar: Based on our discussion last week, I think it's okay to omit the parameters for the seed estimator, since we'll log the estimator parameters as part of each child run (as you've indicated in the inline comment above).

It definitely makes sense to log parameters for the optimal estimator after training, as discussed in https://github.com/mlflow/mlflow/pull/4270/files#r617115186, including parameters for constituent stages.

mlflow/pyspark/ml/__init__.py
dbczumar: Is the index helpful / valuable? If not, I don't think we should include it.

mlflow/pyspark/ml/__init__.py
harupy: In sklearn autologging, we log the `cv_results` attribute of a fitted parameter search estimator as an csv file in the parent run.

Screen Shot 2021-04-21 at 11 58 12

https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html

```
python examples/sklearn_autolog/grid_search_cv.py
mlflow ui
```

dbczumar: ^ For sklearn, `cv_results` contains all explored parameters and metrics, which is great to log as an artifact. I think we should do something similar for pyspark *and* log the parameters of the best estimator.

harupy: btw does `CrossValidatorModel` store stats like above?

mlflow/pyspark/ml/__init__.py
```diff
@@ -101,37 +111,115 @@ def _get_estimator_info_tags(estimator):
...
+    if isinstance(parent_model, CrossValidatorModel):
+        metric_key = 'avg_' + metric_key
+        metrics = parent_model.avgMetrics
+    elif isinstance(parent_model, TrainValidationSplitModel):
+        metrics = parent_model.avgMetrics
+    else:
+        raise RuntimeError(f'Unknown model type {type(parent_model)}.')
+
+    for i in range(len(estimator_param_maps)):
+        estimator = tuning_estimator.copy(estimator_param_maps[i])
```
dbczumar, line 184: Just to confirm, is this a fast / low-cost operation?

mlflow/pyspark/ml/__init__.py
harupy: We could create a table like below and log it as a csv file:

```
| param1 | param2 | avgMetric |
| :----- | :----- | :-------- |
| 1      | 0.1    | 0.4       |
| 1      | 0.2    | 0.5       |
| 2      | 0.1    | 0.6       |
| 2      | 0.2    | 0.7       |
```

mlflow/pyspark/ml/__init__.py
WeichenXu123: yes. estimator.copy(new_params) is fast / low-cost operation, it is also used in `CrossValidator.fit` implementation.

WeichenXu123:
Note: I merge https://github.com/mlflow/mlflow/pull/4263 into this PR to prevent generating more conflicts

mlflow/pyspark/ml/__init__.py
WeichenXu123: I already updated PR to add the hierarchy if the estimator is a pipeline.

mlflow/pyspark/ml/__init__.py
WeichenXu123: Currently the param name is `{estimator_uid}.{param_name}`, it include a `.`, seems cannot be used as csv column name. I use json format, does it make sense?

harupy: > seems cannot be used as csv column name

Is this really true?

mlflow/pyspark/ml/__init__.py
dbczumar: Awesome, thanks!

mlflow/pyspark/ml/__init__.py
dbczumar: @WeichenXu123 can we remove the TODO from line 223?

mlflow/pyspark/ml/__init__.py
```diff
@@ -218,18 +346,79 @@ def _log_pretraining_metadata(estimator, params):
...
+            if isinstance(spark_model, CrossValidatorModel):
+                metric_key = 'avg_' + metric_key
+                metrics = spark_model.avgMetrics
+            elif isinstance(spark_model, TrainValidationSplitModel):
+                metrics = spark_model.avgMetrics
+
+            search_result = []
+            for i in range(len(estimator_param_maps)):
+                search_result.append({
+                    'param_maps': estimator_param_maps[i],
```
dbczumar, line 413: Can we call this `parameters` or `params` since it's a single map of parameters (rather than a plural list of maps)?

mlflow/pyspark/ml/__init__.py
```diff
@@ -218,18 +346,79 @@ def _log_pretraining_metadata(estimator, params):
...
+            except Exception as e:
+                import traceback
+                msg = (
+                    "Encountered exception during creation of child runs for parameter search."
+                    " Child runs may be missing. Exception: {}".format(traceback.format_exc())
+                )
+                _logger.warning(msg)
+
+            estimator_param_maps = _get_tuning_param_maps(estimator)
+
```
dbczumar, line 402: @WeichenXu123 Can we also log the best parameters to the parent run after the hyperparameter sweep is complete? We do this for `sklearn` and MLflow + hyperopt, and I think it would be a nice addition here.

As an example, see:

```
import mlflow
mlflow.sklearn.autolog()

from sklearn import svm, datasets
from sklearn.model_selection import GridSearchCV

iris = datasets.load_iris()
parameters = {"kernel": ("linear", "rbf"), "C": [1, 5, 10]}
svc = svm.SVC()
clf = GridSearchCV(svc, parameters, n_jobs=1, return_train_score=True, refit=True)
clf.fit(iris.data, iris.target)
```

which logs the following optimal params to the parent:

```
{'best_C': '5', 'best_kernel': 'rbf'}
```

mlflow/pyspark/ml/__init__.py
dbczumar: (I think `param_maps` refers to the top-level list of parameter maps for all sets of explored parameters; let me know if I'm incorrect here)

mlflow/pyspark/ml/__init__.py
```diff
@@ -218,18 +346,79 @@ def _log_pretraining_metadata(estimator, params):
...
+                    metric_key: metrics[i]
+                })
+            try_mlflow_log(
+                mlflow.log_dict,
+                search_result,
+                artifact_file="search_result.json"
+            )
+
         if log_models:
             if _should_log_model(spark_model):
```
dbczumar, line 423: For sklearn, we log the CV model *and* the best model as separate artifacts. This allows users to more easily register the best model with the MLflow Model Registry. Can we do that here too?

mlflow/pyspark/ml/__init__.py
```diff
@@ -111,27 +152,109 @@ def _get_instance_param_map(instance):
...
+    estimator_param_maps = parent_estimator.getEstimatorParamMaps()
+    tuning_estimator = parent_estimator.getEstimator()
+    metric_key = parent_estimator.getEvaluator().getMetricName()
+    if isinstance(parent_model, CrossValidatorModel):
+        metric_key = 'avg_' + metric_key
+        metrics = parent_model.avgMetrics
+    elif isinstance(parent_model, TrainValidationSplitModel):
+        metrics = parent_model.avgMetrics
+    else:
+        raise RuntimeError(f'Unknown model type {type(parent_model)}.')
```
dbczumar, line 216:
```suggestion
        raise RuntimeError(f'Unknown parameter search model type {type(parent_model)}.')
```

dbczumar (commented):
@WeichenXu123 I left a few inline comments.

Can we add a section to the `mlflow.pyspark.ml.autolog()` API docs about pipelines and hyperparameter search models? E.g. similar to https://github.com/mlflow/mlflow/blob/94581a3fe4debd2deb9a53b53470c89d69291c29/mlflow/sklearn/__init__.py#L549-L558

mlflow/pyspark/ml/__init__.py
WeichenXu123: Now I switch to csv format for searching result artifact, according to @harupy 's suggestion.

mlflow/pyspark/ml/__init__.py
```diff
@@ -167,7 +249,86 @@ def _get_instance_param_map_recursively(instance, level):
...
+
+
+def _log_parameter_search_results_as_artifact(param_maps, metrics, metric_name, run_id):
+    import pandas as pd
+    result_dict = {key: [] for key in param_maps[0].keys()}
+    result_dict[metric_name] = []
+    for i in range(len(param_maps)):
+        for param_name, param_value in param_maps[i].items():
+            result_dict[param_name].append(param_value)
+        result_dict[metric_name].append(metrics[i])
```
harupy, line 325: I think this code can be simplifed to:

```python
from collections import defaultdict

result_dict = defaultdict(list)
result_dict[metric_name] = metrics
for param_map in param_maps:
    for param_name, param_value in param_map.items():
        result_dict[param_name].append(param_value)
```

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,128 @@ def _get_estimator_info_tags(estimator):
...
+     - hierarchy: the hierarchy of the estimator, it will expand
+         pipeline/meta estimator/param tuning estimator
+     - uid2name_map: a map of `uid` -> `name`, each nested instance uid will be
+         mapped to a fixed name. The naming rule is using `{class_name}` if the
+         instance type occurs once, or `{class_name}_{index}` if the instance type occurs
+         multiple times. The index order is in line with depth-first traversing.
+     - param_search_estimators: a list includes all param search estimators in the
+         hierarchy tree.
+    """
+    uid2name_map = {}
```
harupy, line 213: Can we just generate `uid2name_map` first, then generate other matadata using it to simplify the code?

```python
uid2name_map = ...

hierach = func(..., uid2name_map)
...
```

mlflow/pyspark/ml/__init__.py
```diff
@@ -257,29 +500,101 @@ def _log_pretraining_metadata(estimator, params):
...
+
+        for param_search_estimator in autologging_metadata.param_search_estimators:
+            param_search_estimator_name = \
+                f'{autologging_metadata.uid2name_map[param_search_estimator.uid]}'
+            artifact_dict[param_search_estimator_name] = {}
+
+            artifact_dict[param_search_estimator_name]['tuning_parameter_map_list'] = \
+                _get_tuning_param_maps(
+                    param_search_estimator, autologging_metadata.uid2name_map
+                )
```
harupy, line 519: Can we log a param grid instead of a list of param maps?

```python
# grid: Easier to tell how the search space looks like.
{
  "a.b": [0, 1],
  "c.d": [2, 3],
}

# a list of param maps: Harder to how the search space looks like.
[
  {"a.b": 0, "c.d": 2},
  {"a.b": 0, "c.d": 3},
  {"a.b": 1, "c.d": 2},
  {"a.b": 1, "c.d": 3},
]
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -517,10 +519,12 @@ def call_original(*og_args, **og_kwargs):
                         e,
                     )
 
+                    import traceback
                     _logger.warning(
-                        "Encountered unexpected error during %s autologging: %s",
+                        "Encountered unexpected error during %s autologging: %s, traceback: %s",
                         autologging_integration,
                         e,
+                        traceback.format_stack()
```
harupy, line 527: Is this a temporary change to make it easier to debug errors?

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,129 @@ def _get_estimator_info_tags(estimator):
...
+            'evaluator': _gen_instance_metadata_recursively(
+                evaluator, uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
+            ),
+            'tuned_estimator': _gen_instance_metadata_recursively(
+                tuned_estimator, uid2name_map, cls2count_map, cls2first_uid_map,
+                param_search_estimators
+            )
+        }
+    else:
+        return instance_name
```
harupy, line 198:
```suggestion
        return {"name': instance_name}
```

Should we return a dict here to make all items in the hierarchy dict have the same format (`{"name": "xyz", ...}`)

```jsonc
{
  # current
  "hierarchy": {
    "name": "CrossValidator",
    "evaluator": "BinaryClassificationEvaluator",
    "tuned_estimator": {
      "name": "Pipeline",
      "stages": [
        "Tokenizer",
        "HashingTF",
        "LogisticRegression"
      ]
    }
  },
  # proposed?
  "hierarchy": {
    "name": "CrossValidator",
    "evaluator": "BinaryClassificationEvaluator",
    "tuned_estimator": {
      "name": "Pipeline",
      "stages": [
        {"name": "Tokenizer"},
        {"name": "HashingTF"},
        {"name": "LogisticRegression"}
      ]
    }
  }
}

```

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,129 @@ def _get_estimator_info_tags(estimator):
...
+    cls2first_uid_map = {}
+    param_search_estimators = []
+
+    hierarchy = _gen_instance_metadata_recursively(
+        estimator, uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
+    )
+    for cls, first_uid in cls2first_uid_map.items():
+        if cls2count_map[cls] > 1:
+            # change the first stage name from "{class_name}" to "{class_name}_1"
+            uid2name_map[first_uid] = f'{cls}_1'
```
harupy, line 225: Does this logic really work? It looks like `hierarchy` is generated before the first estimator name is suffixed with `_1` (sorry If I'm missing something)?

mlflow/pyspark/ml/__init__.py
harupy: I think we use the `_param_grid` attribute:
https://github.com/apache/spark/blob/0494dc90af48ce7da0625485a4dc6917a244d580/python/pyspark/ml/tuning.py#L111

mlflow/pyspark/ml/__init__.py
WeichenXu123: yes. it didn't update hierarchy. will update the logic here.

mlflow/utils/autologging_utils/safety.py
WeichenXu123: yes, will revert later

mlflow/pyspark/ml/__init__.py
WeichenXu123: but CrossValidator not only support param_grid, but also support tuning on an arbitrary param_map list.

mlflow/pyspark/ml/__init__.py
WeichenXu123: yes. make sense.

mlflow/pyspark/ml/__init__.py
harupy: Ah got it. If so the current logic is fine.

harupy: Can we do:

```python
if isinstance(estimatorParamMaps, ParamGridBuilder):
    # log grid
else:
    # log param map list
```

?

dbczumar: IMO, either format is fine (though the grid format is certainly more compressed)

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,129 @@ def _get_estimator_info_tags(estimator):
...
+])
+
+
+def _gen_instance_metadata_recursively(
+        instance,
+        uid2name_map,
+        cls2count_map,
+        cls2first_uid_map,
+        param_search_estimators,
+):
```
dbczumar, line 151: Can we add some docs to this method including docstrings for the parameters? From reading the code, the meaning of these `cls2...` maps is not so obvious

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,129 @@ def _get_estimator_info_tags(estimator):
...
+        _is_parameter_search_estimator(estimator)
+
+
+AutologgingEstimatorMetadata = namedtuple('AutologgingEstimatorMetadata', [
+    'hierarchy', 'uid2name_map', 'param_search_estimators'
+])
+
+
+def _gen_instance_metadata_recursively(
+        instance,
```
dbczumar, line 146: Nit: This method fetches metadata for a particular estimator stage, so I think it might be clearer to rename it:

```suggestion
def _gen_stage_metadata_recursively(
        stage,
```

mlflow/pyspark/ml/__init__.py
dbczumar, line 151:
```suggestion
def _gen_estimator_metadata_recursively(
        estimator,
        uid2name_map,
        cls2count_map,
        cls2first_uid_map,
        param_search_estimators,
):
```

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,129 @@ def _get_estimator_info_tags(estimator):
...
+         hierarchy tree.
+    """
+    uid2name_map = {}
+    cls2count_map = {}
+    cls2first_uid_map = {}
+    param_search_estimators = []
+
+    hierarchy = _gen_instance_metadata_recursively(
+        estimator, uid2name_map, cls2count_map, cls2first_uid_map, param_search_estimators
+    )
```
dbczumar, line 221: Nit: can we have `_gen_instance_metadata_recursively` return these maps / lists rather than constructing empty  containers and passing them to the method for population? Returning the containers directly is more pythonic.

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,129 @@ def _get_estimator_info_tags(estimator):
...
+    for cls, first_uid in cls2first_uid_map.items():
+        if cls2count_map[cls] > 1:
+            # change the first stage name from "{class_name}" to "{class_name}_1"
+            uid2name_map[first_uid] = f'{cls}_1'
+
+    for i in range(len(param_search_estimators)):
+        name = param_search_estimators[i]
+        if name in cls2count_map and cls2count_map[name] > 1:
+            # change the estimator name from "{class_name}" to "{class_name}_1"
+            param_search_estimators[i] = f'{name}_1'
```
dbczumar, line 231: I'm a bit confused about why we use `f'{name}_1'` instead of `f'{name}_i'`. Am I missing something or is there an error in the code? If it's a bug, can we add a test case? If it isn't a bug, can we add an line comment explaining why this is something we want to do?

mlflow/pyspark/ml/__init__.py
```diff
@@ -257,29 +500,101 @@ def _log_pretraining_metadata(estimator, params):
...
+                best_index = np.argmax(metrics)
+            else:
+                best_index = np.argmin(metrics)
+
+            best_param_map = estimator_param_maps[best_index]
+            try_mlflow_log(
+                mlflow.log_dict,
+                best_param_map,
+                artifact_file="best_parameters.json"
+            )
```
dbczumar, line 584: Can we log these as params on the run as well?

For example, given the following contents of `best_parameters.json`:

```
{
  "LogisticRegression.maxIter": "10",
  "LogisticRegression.regParam": "0.1"
}
```

can we log the parameters: `best_maxIter = 10` and `best_regParam = 0.1` to MLflow tracking?

mlflow/pyspark/ml/__init__.py
dbczumar: Actually, if there are 1000s of explored params, a grid is definitely preferable.

mlflow/pyspark/ml/__init__.py
```diff
@@ -216,6 +443,26 @@ def autolog(
...
 
+    **How does autologging work for meta estimators?**
+          When a meta estimator (e.g. `pyspark.ml.Pipeline`_, `pyspark.ml.tuning.CrossValidator`_)
+           calls ``fit()``, it internally calls ``fit()`` on its child estimators. Autologging
+           does NOT perform logging on these constituent ``fit()`` calls.
+
+      **Pipeline**
+          In addition to recording the information discussed above, autologging for pipeline
+          records "pipeline hierarchy" which describe the structure of pipeline stages and
+          nested pipeline stages.
```
dbczumar, line 454:
```suggestion
          records a "pipeline hierarchy" artifact which describe the structure of pipeline stages
          and nested pipeline stages.
```

mlflow/pyspark/ml/__init__.py
WeichenXu123:
```
if isinstance(estimatorParamMaps, ParamGridBuilder):
    # log grid
else:
    # log param map list
```
this doesn't work. The estimatorParamMaps will always be a list type, (generated by the ParamGridBuilder.build() )

mlflow/pyspark/ml/__init__.py
WeichenXu123: Refactor related logic to make it easier to read.

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,126 @@ def _get_estimator_info_tags(estimator):
...
+    param_search_estimators = [stage for stage in _traverse_stage(estimator)
+                               if _is_parameter_search_estimator(stage)]
+    hierarchy = _gen_stage_hierarchy_recursively(estimator, uid_to_indexed_name_map)
+
+    metadata = AutologgingEstimatorMetadata(
+        hierarchy=hierarchy,
+        uid_to_indexed_name_map=uid_to_indexed_name_map,
+        param_search_estimators=param_search_estimators
+    )
+    estimator._autologging_metadata = metadata
```
harupy, line 235: Why do we need to attach `metadata` to `estimator`?

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,126 @@ def _get_estimator_info_tags(estimator):
...
+        for stage in stage.getStages():
+            yield from _traverse_stage(stage)
+    elif isinstance(stage, OneVsRest):
+        yield from _traverse_stage(stage.getClassifier())
+    elif _is_parameter_search_estimator(stage):
+        yield from _traverse_stage(stage.getEstimator())
+        yield from _traverse_stage(stage.getEvaluator())
+
+
+def _get_uid_to_indexed_name_map(estimator):
```
harupy, line 160: Can we add tests for this function?

mlflow/pyspark/ml/__init__.py
```diff
@@ -179,6 +378,40 @@ def _get_warning_msg_for_fit_call_with_a_list_of_params(estimator):
     )
 
 
+def _get_tuning_param_maps(param_search_estimator, uid_to_indexed_name_map):
+    tuning_param_maps = []
+    for eps in param_search_estimator.getEstimatorParamMaps():
+        tuning_param_maps.append({f'{uid_to_indexed_name_map[k.parent]}.{k.name}': str(v)
+                                  for k, v in eps.items()})
```
harupy, line 385:
```suggestion
        tuning_param_maps.append({f'{uid_to_indexed_name_map[k.parent]}.{k.name}': v
                                  for k, v in eps.items()})
```

Can we remove `str` here?

mlflow/pyspark/ml/__init__.py
```diff
@@ -134,40 +248,124 @@ def _get_instance_param_map_recursively(instance, level):
...
                 expanded_param_map.update(stage_param_map)
-        elif is_parameter_search_estimator and param_name in ["estimator", "estimatorParamMaps"]:
-            # skip log estimator Param and its nested params because they will be
-            # logged in nested runs.
-            # TODO: Log `estimatorParamMaps` as JSON artifacts.
+        elif is_parameter_search_estimator and param_name == 'estimator':
+            expanded_param_map[logged_param_name] = uid_to_indexed_name_map[param_value.uid]
+            # skip log estimator's nested params because they will be logged as JSON artifact,
+            # and they will be logged in nested runs as well.
+            pass
```
harupy, line 274:
```suggestion
```

mlflow/pyspark/ml/__init__.py
WeichenXu123: in pre training and post training logging we need use metadata, so attach it to estimator we can pass less argument

mlflow/pyspark/ml/__init__.py
```diff
@@ -216,6 +448,26 @@ def autolog(
           supported.
           See ``log_models`` param below for details.
 
+    **How does autologging work for meta estimators?**
+          When a meta estimator (e.g. `pyspark.ml.Pipeline`_, `pyspark.ml.tuning.CrossValidator`_)
```
harupy, line 452: To insert a link for `pyspark.ml.Pipeline`, we need to define the link target like this:

https://github.com/mlflow/mlflow/blob/9d9d4b1f1f62de82637e24c1eb1daeec405e6c30/mlflow/sklearn/__init__.py#L565-L580

mlflow/pyspark/ml/__init__.py
WeichenXu123: added in `test_gen_estimator_metadata`

WeichenXu123:
Update tests. Near ready. Only leftover minor doc/lint issues.

mlflow/pyspark/ml/__init__.py
WeichenXu123: I re-check the sklearn side, it log a single column called "params" and that column include K-V values of all params. So I updated to be similar style.

tests/spark_autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -290,45 +327,150 @@ def test_pipeline(dataset_text):
...
+    lr = LinearRegression(solver="l-bfgs", regParam=0.01)
+    lrParamMaps = [
+        {lr.maxIter: 1, lr.standardization: False},
+        {lr.maxIter: 200, lr.standardization: True},
+        {lr.maxIter: 2, lr.standardization: False},
+    ]
+    best_params = {"LinearRegression.maxIter": 200, "LinearRegression.standardization": True}
+    for ParamSearchEstimator in [CrossValidator, TrainValidationSplit]:
+        # Test on rmse (smaller is better) and r2 (larger is better)
+        for metricName in ["rmse", "r2"]:
```
harupy, line 361: Can we use `pytest.mark.parametrize` here?

```python
@pytest.mark.parametrize("metric_name", [...])
@pytest.mark.parametrize("param_search_estimator", [...])
def test_param_search_estimator(spark_session, dataset_regression, metric_name, param_search_estimator):
```

mlflow/pyspark/ml/__init__.py
harupy: Have you checked how `cv_results.csv` looks like?

mlflow/pyspark/ml/__init__.py
harupy: Got it. If so, logging a list of param maps is fine.

mlflow/pyspark/ml/__init__.py
harupy: Screen Shot 2021-04-29 at 8 24 41

> it log a single column called "params"

sklearn autologging creates a parameter named `grid_search`. Are you talking about this parameter?

harupy: Ah I just found `cv_results.csv` has a column named `params`:

Screen Shot 2021-04-29 at 9 18 22

tests/spark_autologging/ml/test_pyspark_ml_autologging.py
WeichenXu123: Good idea

tests/spark_autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -70,7 +82,17 @@ def dataset_text(spark_session):
             (3, "hadoop mapreduce", 0.0),
         ],
         ["id", "text", "label"],
-    )
+    ).cache()
+
+
+@pytest.fixture(scope="module")
+def dataset_regression(spark_session):
+    features_array, label_array = load_boston(return_X_y=True)
```
harupy, line 90: Can we create a regression dataset using random values?

mlflow/pyspark/ml/__init__.py
WeichenXu123: Updated. Add both single param column and param map column.

WeichenXu123:
Ready for final review. One CI failure is "leaked run" error, which is not related to this PR.

tests/spark_autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -290,45 +332,150 @@ def test_pipeline(dataset_text):
...
+    metric_name, param_search_estimator, spark_session, dataset_regression
+):
+    mlflow.pyspark.ml.autolog()
+    lr = LinearRegression(solver="l-bfgs", regParam=0.01)
+    lrParamMaps = [
+        {lr.maxIter: 1, lr.standardization: False},
+        {lr.maxIter: 200, lr.standardization: True},
+        {lr.maxIter: 2, lr.standardization: False},
+    ]
+    best_params = {"LinearRegression.maxIter": 200, "LinearRegression.standardization": True}
```
harupy, line 368: Can we extract `best_params` from `model` instead of hard-coding it?

mlflow/pyspark/ml/__init__.py
```diff
@@ -134,40 +247,125 @@ def _get_instance_param_map_recursively(instance, level):
...
+
+def _log_parameter_search_results_as_artifact(param_maps, metrics, metric_name, run_id):
+    import pandas as pd
+    import json
+
+    result_dict = defaultdict(list)
+    result_dict["params"] = []
+    result_dict[metric_name] = metrics
+    for param_map in param_maps:
+        result_dict["params"].append(json.dumps(param_map))
```
harupy, line 360: Curious what happends if we remove `json.dumps`?

harupy (approved):
Left a small comment, otherwise LGTM

mlflow/pyspark/ml/__init__.py
```diff
@@ -257,29 +516,100 @@ def _log_pretraining_metadata(estimator, params):
...
+
+            metric_key, metrics = _get_param_search_metrics(estimator, spark_model)
+            _log_parameter_search_results_as_artifact(
+                estimator_param_maps, metrics, metric_key, mlflow.active_run().info.run_id
+            )
+
+            if estimator.getEvaluator().isLargerBetter():
+                best_index = np.argmax(metrics)
+            else:
+                best_index = np.argmin(metrics)
```
WeichenXu123, line 589: Note: these lines logic copied from CrossValidator/TrainValidationSplit code.

tests/spark_autologging/ml/test_pyspark_ml_autologging.py
WeichenXu123: We can. But here, I want to it always search out the best param : `{"LinearRegression.maxIter": 200, "LinearRegression.standardization": True}`, 
This is for testing whether the "best_parameter" selection logic works correctly.
Note: I copied some lines code from ml tuning, the test here is to ensure the copied code works correctly, see https://github.com/mlflow/mlflow/pull/4270/files#r623506659
That's also the reason I test 2 metrics "rmse" and "r2" here, one metric is "smaller better" and another "larger better", so they will cover the 2 branch in the copied code logic.

harupy: Ah I see. It makse sense now. Thanks for the explanation!

mlflow/pyspark/ml/__init__.py
```diff
@@ -107,24 +120,124 @@ def _get_estimator_info_tags(estimator):
...
+    Returns an AutologgingEstimatorMetadata object.
+    The AutologgingEstimatorMetadata object includes:
+     - hierarchy: the hierarchy of the estimator, it will expand
+         pipeline/meta estimator/param tuning estimator
+     - uid_to_indexed_name_map: a map of `uid` -> `name`, each nested instance uid will be
+         mapped to a fixed name. The naming rule is using `{class_name}` if the
+         instance type occurs once, or `{class_name}_{index}` if the instance type occurs
+         multiple times. The index order is in line with depth-first traversing.
+     - param_search_estimators: a list includes all param search estimators in the
+         hierarchy tree.
```
harupy, line 220: Can a pipeline contains multiple param search estimators?

WeichenXu123: It can. It is legal, but i think no user will do that.

mlflow/pyspark/ml/__init__.py
```diff
@@ -216,6 +448,38 @@ def autolog(
...
 
+    **How does autologging work for meta estimators?**
+          When a meta estimator (e.g. `Pipeline`_, `CrossValidator`_, `TrainValidationSplit`_,
+          `OneVsRest`_)
+          calls ``fit()``, it internally calls ``fit()`` on its child estimators. Autologging
+          does NOT perform logging on these constituent ``fit()`` calls.
+
+          A "estimator_info.json" artifact will be logged, the artifact include an item call
+          "hierarchy" which describe the hierarchy of the meta estimators. The hierarchy will
+          expand all nested stage such as nested pipeline stages.
```
dbczumar, line 459:
```suggestion
          A "estimator_info.json" artifact is logged, which includes a `hierarchy` entry
          describing the hierarchy of the meta estimator. The hierarchy includes expanded
          entries for all nested stages, such as nested pipeline stages.
```

dbczumar (approved):
LGTM with one small docs suggestion. Thanks @WeichenXu123 !

PR #4276, opened 2021-04-21
Fix: When top level call autologging failed, nested calls autologging are still enabled

WeichenXu123:

Fix: If _log_pretraining_metadata raise exception, the original patched fit will be called and then all nested fit will be autologged.

## How is this patch tested?

Unit test.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/autologging_utils/safety.py
+++ b/mlflow/utils/autologging_utils/safety.py
@@ -268,6 +268,9 @@ def is_testing():
     return os.environ.get(_AUTOLOGGING_TEST_MODE_ENV_VAR, "false") == "true"
 
 
+_AUTOLOGGING_INTEGRATIONS_TEMP_DISABLED_SET = set()
+
+
 def safe_patch(
     autologging_integration, destination, function_name, patch_function, manage_run=False
 ):
@@ -367,8 +370,9 @@ def safe_patch(
                 mlflow.active_run() and not _AutologgingSessionManager.active_session()
             )
 
-            if autologging_is_disabled(autologging_integration) or (
-                user_created_fluent_run_is_active and exclusive
+            if autologging_integration in _AUTOLOGGING_INTEGRATIONS_TEMP_DISABLED_SET or \
+                    autologging_is_disabled(autologging_integration) or (
+                    user_created_fluent_run_is_active and exclusive
             ):
                 # If the autologging integration associated with this patch is disabled,
                 # or if the current autologging integration is in exclusive mode and a user-created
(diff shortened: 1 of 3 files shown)


dbczumar (approved):
LGTM! Thanks @WeichenXu123 !

mlflow/utils/autologging_utils/safety.py
```diff
@@ -268,6 +268,9 @@ def is_testing():
     return os.environ.get(_AUTOLOGGING_TEST_MODE_ENV_VAR, "false") == "true"
 
 
+_AUTOLOGGING_INTEGRATIONS_TEMP_DISABLED_SET = set()
```
dbczumar, line 271: @WeichenXu123 As an alternative to defining a new set of temporarily disabled integrations, could we add a new `state` property to `AutologgingSession` (https://github.com/mlflow/mlflow/blob/94581a3fe4debd2deb9a53b53470c89d69291c29/mlflow/utils/autologging_utils/safety.py#L539) that we can update to `failed` when a failure occurs? We can then then check the state of the active session here: https://github.com/mlflow/mlflow/pull/4276/files#diff-7cea9a64941de9f0572df5ad42d7eef22d2915c2408bb0e59406fe2df20a8167R370.

The reason I prefer this approach is that it scopes the behavior to a single autologging session, rather than creating new global state. This way, we don't require a `try/finally` to ensure that an entry is always removed from this global list.

Let me know what your thoughts are here, @WeichenXu123

dbczumar: Regardless of the chosen approach, can we also introduce a test case ensuring that this nesting + failure scenario is handled properly?

WeichenXu123: emm, add AutologgingSession state property doesn't work.
because each new nested autologging call will create a new AutologgingSession, we cannot fetch the parent AutologgingSession.
otherwise we change AutologgingSession to something similar to the _TrainingSession

WeichenXu123: add test, but keep current fixing approach due to reason above.

dbczumar: @WeichenXu123 The current implementation doesn't create a new session every time (see https://github.com/mlflow/mlflow/blob/06c8015385d6f1e8c9e1f455e117b99bdbf55773/mlflow/utils/autologging_utils/safety.py#L549-L553)

WeichenXu123: my fault. your approach works. Updated code.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -536,7 +545,12 @@ def call_original(*og_args, **og_kwargs):
 # Represents an active autologging session using two fields:
 # - integration: the name of the autologging integration corresponding to the session
 # - id: a unique session identifier (e.g., a UUID)
-AutologgingSession = namedtuple("AutologgingSession", ["integration", "id"])
+# - state: init/succeeded/failed
+class AutologgingSession:
+    def __init__(self, integration, id_):
+        self.integration = integration
+        self.id = id_
+        self.state = "init"
```
dbczumar, line 553:
```suggestion
        self.state = "running"
```

dbczumar (approved):
Thanks @WeichenXu123 ! Left one very minor comment about the `AutologgingSession` initial state.

PR #4301, opened 2021-04-29
Run cross version tests when we update maximum versions

harupy:

When we update maximum versions for making a new release, run tests for udpated packages.

## How is this patch tested?

Existing check

Components 

Interface 

Language 

Integrations

### DIFF
--- a/dev/set_matrix.py
+++ b/dev/set_matrix.py
@@ -167,7 +167,7 @@ def filter_versions(versions, min_ver, max_ver, excludes=None):
 
     # Prevent specifying non-existent versions
     assert min_ver in versions
-    # assert max_ver in versions
+    assert max_ver in versions
     assert all(v in versions for v in excludes)
 
     versions = {Version(v): t for v, t in versions.items() if v not in excludes}
--- a/dev/set_matrix.py
+++ b/dev/set_matrix.py
@@ -419,7 +419,7 @@ def expand_config(config):
                         run=run,
                         package=pip_release,
                         version=ver,
-                        newer_than_max=Version(ver) > Version(max_ver),
+                        supported=Version(ver) <= Version(max_ver),
                     )
                 )
 
--- a/dev/set_matrix.py
+++ b/dev/set_matrix.py
@@ -440,7 +440,7 @@ def expand_config(config):
                         run=run,
                         package=pip_release,
                         version=DEV_VERSION,
-                        newer_than_max=True,
+                        supported=False,
                     )
                 )
     return matrix
--- a/dev/set_matrix.py
+++ b/dev/set_matrix.py
@@ -394,14 +394,14 @@ def expand_config(config):
         for key, cfg in cfgs.items():
             print("Processing", flavor_key, key)
             # Released versions
-            versions = filter_versions(
-                all_versions, cfg["minimum"], cfg["maximum"], cfg.get("unsupported"),
-            )
+            min_ver = cfg["minimum"]
+            max_ver = cfg["maximum"]
+            versions = filter_versions(all_versions, min_ver, max_ver, cfg.get("unsupported"),)
             versions = select_latest_micro_versions(versions)
 
             # Explicitly include the minimum supported version
-            if cfg["minimum"] not in versions:
-                versions.append(cfg["minimum"])
+            if min_ver not in versions:
+                versions.append(min_ver)
 
             pip_release = package_info["pip_release"]
             for ver in versions:
@@ -419,6 +419,7 @@ def expand_config(config):
                         run=run,
                         package=pip_release,
                         version=ver,
+                        newer_than_max=Version(ver) > Version(max_ver),
                     )
                 )
 
@@ -439,6 +440,7 @@ def expand_config(config):
                         run=run,
                         package=pip_release,
                         version=DEV_VERSION,
+                        newer_than_max=True,
                     )
                 )
     return matrix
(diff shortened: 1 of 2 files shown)


dev/set_matrix.py
```diff
@@ -419,6 +419,7 @@ def expand_config(config):
                         run=run,
                         package=pip_release,
                         version=ver,
+                        newer_than_max=Version(ver) > Version(max_ver),
```
harupy, line 422: We update `ml-package-versions.yml` when we make a release like below:

```diff
# ml-package-versions.yml

sklearn:

  autologging:
    minimum: "0.20.3"
-    maximum: "0.24.1"
+    maximum: "0.24.2"
```

For this change, we want to run tests for sklearn autologging using sklearn 0.24.2 to ensure sklearn autologging really works with 0.24.2. However, `set_matrix.py` skips running tests now. Here' why:


```python
# Assumption: the latest version of sklearn is 024.2 now
# NOTE: our cross-version-tests strategy fetches the latest version available ignoring the maximum version specified in `ml-package-versions.yml`.

# ---------- NOW ----------
# Matrix generated from master
[
    {"package": "sklearn", "version": "0.24.2"},
]

# Matrix generated from PR branch
[
    {"package": "sklearn", "version": "0.24.2"},
    # -> considered as not-updated
]

# ---------- PROPOSED ----------
# Matrix generated from master (where sklearn maximum version is 0.24.1
# so `newer_than_maximum` evaluates to True)
[
    {"package": "sklearn", "version": "0.24.2", "newer_than_maximum": True},
]

# Matrix generated from PR branch (where sklearn maximum version is 0.24.2
# so `newer_than_maximum` evaluates to False)
[
    {"package": "sklearn", "version": "0.24.2", "newer_than_maximum": False},
    # -> considered as "udpated"
]

# `newer_than_maximum` field is different -> run test for sklearn 0.2

```

mlflow/ml-package-versions.yml
```diff
@@ -13,7 +13,7 @@ sklearn:
 
   autologging:
     minimum: "0.20.3"
-    maximum: "0.24.1"
+    maximum: "0.24.2"
```
harupy, line 16: A dummy change to make sure the proposed change works fine.

harupy: It worked:

image

dev/set_matrix.py
harupy: This works but it's not straight forward.

WeichenXu123: The code logic is a little strange.

Why not do:
```
# Matrix generated from master 
matrix_master = [ {"package": "sklearn", "max_supported_version": "0.24.1"}, ]

Matrix generated from PR branch:
matrix_pr = [ {"package": "sklearn", "max_supported_version": "0.24.2"}, ]
```

Then we get master and PR's max_supported_version, and then we can generate a series of CI runs. We don't need the field `newer_than_maximum` which make the logic hard to understand.

harupy: If we did that, sklearn `0.24.2` would be always tested until  we update the max version to `0.24.2`.

harupy: How about renaming `newer_than_max` to `supported`?

dev/set_matrix.py
```diff
@@ -419,6 +419,7 @@ def expand_config(config):
                         run=run,
                         package=pip_release,
                         version=ver,
+                        supported=Version(ver) <= Version(max_ver),
```
harupy, line 422: @WeichenXu123 I renamed `newer_than_maximum` to `supported` (which I thin is clearer).

mlflow/ml-package-versions.yml
```diff
@@ -13,7 +13,7 @@ sklearn:
 
   autologging:
     minimum: "0.20.3"
-    maximum: "0.24.2"
+    maximum: "0.24.3"
```
harupy, line 16: This change should create a run for sklearn 0.24.3.

dev/set_matrix.py
harupy, line 422: When we update a package version, this field flips to `True` and the cross-version-tests script recognizes it as a new item. We could use this field to allow failure versions that we don't support by checking this field.

dev/set_matrix.py
```diff
@@ -439,6 +440,7 @@ def expand_config(config):
                         run=run,
                         package=pip_release,
                         version=DEV_VERSION,
+                        supported=False,
```
harupy, line 443: The dev version should always be unsupported.

dev/set_matrix.py
```diff
@@ -473,7 +475,10 @@ def main():
     diff_flavor = set(filter(lambda x: x["flavor"] in changed_flavors, matrix))
 
     # If this file contains changes, re-run all the tests, otherwise re-run the affected tests.
-    include = matrix if (__file__ in changed_files) else diff_config.union(diff_flavor)
+    # I'LL REVERT THIS CHANGE BEFORE MERGE
+    # include = matrix if (__file__ in changed_files) else diff_config.union(diff_flavor)
+    include = diff_config.union(diff_flavor)
+    # I'LL REVERT THIS CHANGE BEFORE MERGE
```
harupy, line 481: A temporary change to avoid running too many cross-version tests.

WeichenXu123 (approved):

mlflow/ml-package-versions.yml
```diff
@@ -236,9 +236,9 @@ fastai-1.x:
       pytest tests/fastai/test_fastai_model_export.py --large
 
   autologging:
-    pin_maximum: True
+    # pin_maximum: True
     minimum: "1.0.60"
-    maximum: "1.0.61"
+    maximum: "2.0.0"
```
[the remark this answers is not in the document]
WeichenXu123: Why CI triggered 2.1/2.2/2.3 version tests ? shouldn't it only trigger 2.0 version test ?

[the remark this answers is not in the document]
harupy: Yep, this is a temporary change to make sure the proposed logic works. The maximum version actually specifies the maximum **major** version. For example, `maximum: "2.0.0"` basically means `test all 2.x minor release but`.

Issue #4307, opened 2021-05-03
[FR] Adding a functionality to move runs to different experiments


edwardcqian:
## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary
Adding a functionality in both the UI and API to move a tracking run to a different experiment.

## Motivation
- What is the use case for this feature?
Often it's nice to be able to reorganize experiments. For example scope may change mid experiment, or by mistake a run is logged to the wrong place.
- Why is this use case valuable to support for MLflow users in general?
Currently there is no functionality to organize experiment. It's only create, log and delete.
- Why is this use case valuable to support for your project(s) or organization?
We currently move run folders manually, this is quite hacky and mistakes can happen, an formal solution.
- Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient)
There is currently no clean move functionality within MLflow (UI or API)

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [x] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interfaces
- [x] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Languages 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
[...]

PR #4309, opened 2021-05-04
Adding functionality to move logged runs

edwardcqian:

Implemented python API, Java client and tracking UI support for moving logged runs to a different experiment

A `/mlflow/runs/move` POST request is added to the REST API with parameters:
- `run_id`
- `src_experiment_id`: numeric experiment id for the source of the run
- `dest_experiment_id`: numeric experiment id for where the move the run

Links to #4307 

## How is this patch tested?
- Unit test
- Manual test by creating and moving an experiment via both local and db tracking backend

UI and REST API functionality added for moving logged runs to a different experiment

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/server/js/src/common/forms/validations.js
+++ b/mlflow/server/js/src/common/forms/validations.js
@@ -33,7 +33,7 @@ export const getExperimentIdValidator = (getCurrentExperimentId, getExistingExpe
       // no need to execute below validations when no value is entered
       // eslint-disable-next-line callback-return
       callback(undefined);
-    } else if (isNaN(value)) {
+    } else if (!(Number(value) === parseInt(value, 10))) {
       callback(`Experiment ID must be an integer`);
     } else if (getCurrentExperimentId() == value) {
       callback(`"${value}" is the current experiment ID`);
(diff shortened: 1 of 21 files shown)


mlflow/server/js/src/common/forms/validations.js
```diff
@@ -27,6 +27,27 @@ export const getExperimentNameValidator = (getExistingExperimentNames) => {
...
 };
 
+export const getExperimentIdValidator = (getCurrentExperimentId, getExistingExperimentIds) => {
+  return (rule, value, callback) => {
+    if (value.length === 0) {
+      // no need to execute below validations when no value is entered
+      // eslint-disable-next-line callback-return
+      callback(undefined);
+    } else if (isNaN(value)) {
+      callback(`Experiment ID must be an integer`);
```
akshaya-a, line 37: shouldn't this validate against string? perhaps just rely on GetExpById()'s response to validate?

edwardcqian: that sounds good. would also validate against floats.

edwardcqian: Upon some more thought. Changed to an integer check. By default the form gives string values, which does work for the API call.

Correct me if I'm wrong, `getExperiment` method would also get deleted experiment. It probably better to not include those for the move target.

edwardcqian: Ah I see, that makes sense. I think as long as I remove the integer condition it should validate string. It just checks that the Id is not the current experiment id and that it is in the list of existing experiment ids.

AveshCSingh:
I think we should hold off on this implementation for now, since this change needs some design work. See https://github.com/mlflow/mlflow/issues/4307#issuecomment-865452223.

Issue #4223, opened 2021-04-02
[BUG] Using a signature in a Tensorflow model throws an exception


saschaschramm:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for additional information about bug reports. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

**Please fill in this bug report template to ensure a timely and thorough response.**

### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ X] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: Custom
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: macOS Big Sur 11.2.3
- **MLflow installed from (source or binary)**: binary
- **MLflow version (run ``mlflow --version``)**: 1.15.0
- **Python version**: 3.8.8
- **npm version, if running the dev UI**:
- **Exact command to reproduce**:

### Describe the problem
**Actual behavior:**

When a TensorFlow model with a signature is serialized and used for a prediction the MLflow predict method throws an exception.

Model with signature being saved (this works):

```python
@tf.function(input_signature=[tf.TensorSpec(shape=[1, 1], dtype=tf.float32, name='input')])
        def serve_fn(example):
            outputs = model(example)
            return outputs

tf.saved_model.save(model, TF_MODEL_DIR,
                            signatures={'serving_default': serve_fn.get_concrete_function()})

mlflow.tensorflow.log_model(
                tf_saved_model_dir=TF_MODEL_DIR,
                tf_meta_graph_tags=['serve'],
                tf_signature_def_key='serving_default',
                artifact_path='mlflow_model'
            )

```

This throws an exception:
```python
mlflow_model = mlflow.pyfunc.load_model(model_uri)
prediction = mlflow_model.predict(feed_dict)
```

Interestingly an exception is not thrown when the TF2Wrapper is instantiated directly:
``` python
loaded_model = tf.saved_model.load(
                export_dir=f'mlruns/0/{run.info.run_id}/artifacts/mlflow_model/tfmodel',
[...]

PR #4319, opened 2021-05-06
Fix loading of TensorFlow SavedModel format (#4223)

saschaschramm:
Fix loading of TensorFlow SavedModel format (#4223)

This pull request fixes a bug that prevents loading a TensorFlow model that was saved in the TensorFlow SavedModel format. 

The bug is described in detail in #4223.

The problem is solved by passing the TensorFlow model to the _TF2Wrapper.
https://github.com/saschaschramm/mlflow/blob/d05ba3e41325b159945169dc6ed1a71755172f24/mlflow/tensorflow.py#L557-L563

## How is this patch tested?

* Executed existing unit tests
* Created a new unit test: tests/tensorflow/test_tensorflow2_saved_model.py

Fixes a bug that prevents loading a TensorFlow model that was saved in the SavedModel format. 

Components 
- [X] `area/scoring`: Local serving, model deployment tools, spark UDFs

Interface 

Language 

Integrations


mlflow/tensorflow.py
```diff
@@ -554,10 +554,12 @@ class _TF2Wrapper(object):
     ``predict(data: pandas.DataFrame) -> pandas.DataFrame``. For TensorFlow versions >= 2.0.0.
     """
 
-    def __init__(self, infer):
+    def __init__(self, model, infer):
         """
+        :param model: A Tensorflow SavedModel.
         :param infer: Tensorflow function returned by a saved model that is used for inference.
         """
+        self.model = model
```
WeichenXu123, line 562: Could you explain why add this attribute address the issue ? the `model` attribute is not used in `predict`

saschaschramm: Why the existing `_TF2Wrapper` doesn't work:
The TensorFlow model is loaded inside of the `_load_pyfunc` method. After loading the model the reference count of the `loaded_model` object is set to 1. When the `_load_pyfunc` method returns the reference count is decreased to 0 and the model gets deallocated. If you then try to call the infer signature, TensorFlow won't be able to find the `loaded_model` anymore.

Why the modified `_TF2Wrapper` works:
If you instead pass the model to the `_TF2Wrapper` also this object will hold a reference to the
`loaded_model` object. That means in this case the reference count is 2. When the `_load_pyfunc` method returns, the reference count is reduced to 1 but never reaches zero because the `_TF2Wrapper` still exists. This avoids that the model gets deallocated.

Hope this answers helps. If you have more questions don't hesitate to ask me.

WeichenXu123: Make sense.
This looks more like a TF2 bug , the returned signature should keep reference to the model, (but actually the reference count is decreased to 0 ), but the workaround here is fine.

saschaschramm: I think the reason is that ConcreteFunctions (the signature is such a function) only retain a weak reference to the model variables. The behavior is described here: https://www.tensorflow.org/guide/function

WeichenXu123: ok i see. Thanks!
Could you help add test for model serving on TF2 ?

saschaschramm: Sure I can help. 

A unit test ist already part of this pull request:
https://github.com/mlflow/mlflow/pull/4319/files/d05ba3e41325b159945169dc6ed1a71755172f24#diff-314e8120f884dcac5855f9ce3d1094d7871d4cc07d885d41b425d1f66c8d0e72


Do you need more?

WeichenXu123: I update your added test. and move it into tests/tensorflow/test_tensorflow2_model_export.py and add `@pytest.mark.large`

WeichenXu123: Follow-up question: could you explain why the existing test passed ? 
https://github.com/mlflow/mlflow/blob/b454522ecc7727e6d90b59969019428ee49e6779/tests/tensorflow/test_tensorflow2_model_export.py#L596

The existing test also run under Tensorflow2 and also do similar thing, the only difference is it use 'predict' signature, but you use 'serving_default' signature.

WeichenXu123 (approved):

mlflow/tensorflow.py
saschaschramm: The test passed because the `saved_tf_iris_model` was defined using the Estimator API (`tf.estimator.DNNClassifier`). The API already existed in TensorFlow 1 and  doesn’t use `tf.functions`, that’s why it works. I don’t know the TensorFlow Roadmap but I assume that Google will deprecate the Estimator API sometime in the future. The Estimator documentation (https://www.tensorflow.org/guide/estimator) says the following: 

> TensorFlow implements several pre-made Estimators. Custom estimators are still suported, but mainly as a backwards compatibility measure. Custom estimators should not be used for new code

Moreover, the TFX documentation (https://www.tensorflow.org/tfx/guide/keras) mentions:

> The Estimator API has been retained in TensorFlow 2.x, but is not the focus of new features and development.

Regarding `pyfunc_wrapper.predict`: At runtime `pyfunc_wrapper.predict` calls `_TF2Wrapper.predict` and this method also uses the signature (`self.infer`).

WeichenXu123: @saschaschramm Thanks for the explanation ! I add related comments into code.

mlflow/tensorflow.py
```diff
@@ -552,10 +552,16 @@ class _TF2Wrapper(object):
     ``predict(data: pandas.DataFrame) -> pandas.DataFrame``. For TensorFlow versions >= 2.0.0.
     """
 
-    def __init__(self, infer):
+    def __init__(self, model, infer):
         """
+        :param model: A Tensorflow SavedModel.
         :param infer: Tensorflow function returned by a saved model that is used for inference.
         """
+        # Note: we need retain the model reference in TF2Wrapper object, because the infer function
```
harupy, line 560:
```suggestion
        # Note: we need to retain the model reference in TF2Wrapper object, because the infer function
```

PR #4359, opened 2021-05-13
Adding explain option in mlflow deployments cli

shrinath-suresh:

https://pytorch.org/serve/inference_api.html#explanations-api 

To support torchserve explain api (Captum) , adding a new deployment cli option `explain`.

Note: the option is only available in cli mode. Explain method has not been added to `BaseDeploymentClient` as it would affect other flavours. 

## How is this patch tested?

Added necessary unit tests

https://github.com/chauhang/mlflow-torchserve-1/pull/7 - Separate activity going on to add few examples with `mlflow explain cli` commands. 

Components 

Interface 

Language 
- [x] `language/python`: Python APIs and clients

Integrations


tests/resources/mlflow-test-plugin/mlflow_test_plugin/fake_deployment_plugin.py
```diff
@@ -27,6 +27,10 @@ def get_deployment(self, name):
     def predict(self, deployment_name, df):
         return "1"
 
+    def explain(self, deployment_name, df):
+        # pylint: disable=unused-argument
```
shrinath-suresh, line 31: Since explain method is not availabe in basedeployment client, pylint throws unused argument warning. Ignoring the warning.

mlflow/deployments/cli.py
```diff
@@ -258,3 +258,24 @@ def predict(target, name, input_path, output_path):
...
+@parse_input
+@parse_output
+def explain(target, name, input_path, output_path):
+    """
+    predict the features importance for the deployed model for the given input(s)
+    """
+    import pandas as pd
+
+    df = pd.read_json(input_path)
+    client = interface.get_deploy_client(target)
```
shrinath-suresh, line 275: Just a thought:  Should we raise an exception here if target is not `torchserve` ?

smurching: @shrinath-suresh we probably don't want to condition logic in MLflow core on specific plugins that exist outside MLflow (you could imagine then we'd have a big branching if-statement for all deploy plugins/in order to add a new plugin folks would have to know to merge a PR modifying this file with details of their own plugin, vs being able to develop & release a plugin independently of MLflow.

Could we instead check whether the `explain` method is defined on the `client` object and raise an exception saying that the current plugin doesn't support the `explain` API if not? e.g. something like

```suggestion
    client = interface.get_deploy_client(target)
     if not hasattr(client, "explain"):
         raise MlflowException("Computing model explanations is not yet supported for deployment target %s" % target)
     result = client.explain(name, df)
     ... 
```

mlflow/deployments/cli.py
```diff
@@ -258,3 +258,24 @@ def predict(target, name, input_path, output_path):
...
+
+
+@commands.command("explain")
+@deployment_name
+@target_details
+@parse_input
+@parse_output
+def explain(target, name, input_path, output_path):
+    """
+    predict the features importance for the deployed model for the given input(s)
```
smurching, line 270: Could we enhance this docstring to make it clearer what the API does? Here's a stab at a docstring, let me know if I'm understanding the purpose of this API correctly and thanks in advance!:

```suggestion
    """
    Generate explanations of model predictions on the specified input for the deployed model for the given input(s). Explanation output formats vary by deployment target, and can include details like feature importance for understanding/debugging predictions. Run `mlflow deployments help` or consult the documentation for your plugin for details on explanation format. For information about the input data formats accepted by this function, see the following documentation: https://www.mlflow.org/docs/latest/models.html#built-in-deployment-tools
    """    
```

@shrinath-suresh it'd be good to check - given this explanation, could a new user successfully understand how to generate/interpret model explanations from a model deployed to torchserve?

tests/resources/mlflow-test-plugin/mlflow_test_plugin/fake_deployment_plugin.py
smurching: I think if we want to add this API to the MLflow core CLI, we'll need to add it to the base deployment client - it's a bit strange from a user perspective for this functionality to be CLI-only, as it's then not possible to write e.g. a Python script to perform deployment and all other deploy plugin functionality is available via both the API and CLI

KasirajanA: If we declare the explain method as abstract method in base deployment class then client instance creation for other flavour's will fail if the abstract method is not implemented in those plugin classes. Please refer to the attached snapshots. Can we add explain as a non-abstract method in base deployment class so that it won't break the other flavour's instance creation?

explain method as abstract method.
Screenshot from 2021-05-31 17-43-29

explain as non-abstract method
Screenshot from 2021-05-31 17-45-57

mlflow/deployments/cli.py
KasirajanA: If we add explain method as non-abstract method in base deployment class then "hasattr(client, "explain")" will always return True. So, Can we add the exception in the base class's explain method itself?

```
    def explain(self, deployment_name, df):
        raise MlflowException("Computing model explanations is not yet supported for this deployment target")
```

smurching: @KasirajanA yep that makes sense!

smurching: And I think adding it as a non-abstract method (good catch there as well) should allow us to make this change without breaking existing plugins, as you described :)

smurching:
BTW, thanks for this PR - per offline discussion, it seems reasonable to me to add an `explain` API as an optional method for plugins to implement (raising a clear exception indicating that explanation support isn't yet supported). In particular, it seems possible to implement explanation support for our existing built-in sagemaker and azureml deployment targets:
* [Sagemaker explanations doc link](https://sagemaker-examples.readthedocs.io/en/latest/sagemaker_processing/fairness_and_explainability/fairness_and_explainability.html#Detecting-Bias) 
* [AzureML explanations doc link](https://docs.microsoft.com/en-us/azure/machine-learning/how-to-machine-learning-interpretability-aml#interpretability-at-inference-time)

In particular:
* With Sagemaker, the workflow is that Sagemaker will spin up additional instances to compute explanations on a deployed model. As such the API seems similar to the one proposed here, with a few additional required parameters - you need to additionally specify an instance type and instance count for the compute launched for explanations (specified as part of a [ModelConfig](https://sagemaker.readthedocs.io/en/stable/api/training/processing.html#sagemaker.clarify.ModelConfig)), the desired input/output S3 path of the data/explanations (should be doable to convert this into local file paths in the MLflow case by uploading/downloading from a default, optionally user-specified bucket), and label/feature columns (can hopefully be inferred from the model schema in the MLflow case, or otherwise specified by the user)
* With AzureML, it seems you deploy an explainer alongside your main ML model (i.e. you pass both an explainer & your ML model to the `Webservice.deploy_from_model` API as in step f (at the time of writing) of [this example](https://docs.microsoft.com/en-us/azure/machine-learning/how-to-machine-learning-interpretability-aml#interpretability-at-inference-time)). We should be able to implement computing explanations by updating the Webservice to include an explainer ([see API](https://docs.microsoft.com/en-us/python/api/azureml-core/azureml.core.webservice.aci.aciwebservice?view=azure-ml-py#azureml_core_webservice_aci_AciWebservice_update)), optionally adding support for specifying an explainer at deploy time & later computing explanations against it.

KasirajanA:
@smurching @harupy, Can you please approve the workflows to run?

akshaya-a:
@smurching  ack, pointing the team at this PR

smurching:
@KasirajanA sorry for the delay, just triggered a CI run :)

tests/pytorch/test_pytorch_model_export.py
```diff
@@ -807,6 +808,11 @@ def test_load_model_succeeds_when_data_is_model_file_instead_of_directory(
 
 
 @pytest.mark.large
+@pytest.mark.skipif(
+    Version(torch.__version__) >= Version("1.9.0"),
+    reason="From pytorch 1.9.0, torch.load() creates a subclass from the passed Unpickler. "
+           "Refer https://github.com/pytorch/pytorch/pull/53139",
+)
```
KasirajanA, line 815: In pytorch 1.9.0, a subclass is created from the passed Unpickler in the torch.load function as pickle.Unpickler is marked readonly in pickle.
For more details refer to [this PR](https://github.com/pytorch/pytorch/pull/53139)

mlflow/deployments/base.py
```diff
@@ -190,3 +191,20 @@ def predict(self, deployment_name, df):
...
         pass
+
+    @experimental
+    def explain(self, deployment_name, df):  # pylint: disable=unused-argument
+        """
+        Generate explanations of model predictions on the specified input pandas Dataframe
+        ``df`` for the deployed model. Explanation output formats vary by deployment target,
+        and can include details like feature importance for understanding/debugging predictions.
+        Note that we have created this function as non-abstract method to not break
+        the existing downstream.
```
smurching, line 202:
```suggestion
```

Thanks for the updated docstring! We should remove these lines IMO, I don't think we want this to show up in user-facing docs (e.g. if I'm using the deploy API I might be confused by developer-facing comments about this method being non-abstract)

mlflow/deployments/base.py
```diff
@@ -190,3 +191,20 @@ def predict(self, deployment_name, df):
...
+
+        :param deployment_name: Name of deployment to predict against
+        :param df: Pandas DataFrame to use for explaining feature importance in model prediction
+        :return: An exception if the implementation is not available in deployment target's class
```
smurching, line 206: Could we clarify that the expected return type is a JSON-able object, e.g. a pandas dataframe or numpy array? 
```suggestion
        :return: A JSON-able object (pandas dataframe, numpy array, dictionary), or exception if unimplemented for the current deployment target
```

mlflow/deployments/cli.py
```diff
@@ -258,3 +258,31 @@ def predict(target, name, input_path, output_path):
...
+
+@commands.command("explain")
+@deployment_name
+@target_details
+@parse_input
+@parse_output
+def explain(target, name, input_path, output_path):
+    """
+    Generate explanations of model predictions on the specified input for
+    the deployed model for the given input(s).Explanation output formats vary
```
smurching, line 271:
```suggestion
    the deployed model for the given input(s). Explanation output formats vary
```

smurching (commented):
Mostly looks great! Just a few minor doc comments, otherwise looks good. Were you able to manually QA the new deployments API using both `mlflow` and the skinny client? @harupy may be able to provide guidance on how to do so. Thanks again for the PR :)

shrinath-suresh:
> Mostly looks great! Just a few minor doc comments, otherwise looks good. Were you able to manually QA the new deployments API using both `mlflow` and the skinny client? @harupy may be able to provide guidance on how to do so. Thanks again for the PR :)

Thanks for the review sid. With this branch, we have tested the api behaviour on `mlflow-torchserve` plugin side. @harupy can you please point us to documentation on how to test it in `mlflow` and the skinny client.

harupy:
For skinny-client, I think you can install it via: `pip install mlflow-skinny` and verify the CLI commands work as expected.

KasirajanA:
@smurching , @harupy, we installed mlflow-skinny and ran the pytorch examples successfully. 
mlflow_skinny

shrinath-suresh:
> @smurching , @harupy, we installed mlflow-skinny and ran the pytorch examples successfully.
> mlflow_skinny

@KasirajanA Can you please test the cli commands as well (create, predict and explain) ?

smurching (approved):
LGTM! If manual QA pans out then let's merge :)

shrinath-suresh:
@KasirajanA  has tested the example with explain API and its working as expected.

titanic_explain

shrinath-suresh:
@harupy @smurching Shall we go ahead and merge this PR ?

PR #4382, opened 2021-05-21
Limit # of child runs created for hyperparam search models

mohamad-arabi:

The ability to limit the number of child runs created when running hyper param search with auto-logging. 
The default number of child runs is set to 5.  

## How is this patch tested?
Added pytest unit tests.

A user can now specify a limit to the number child runs created when training an sklearn hyper parameter search estimator.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -694,8 +694,11 @@ def autolog(
     )
 
     if max_tuning_runs is not None and max_tuning_runs < 0:
-        raise ValueError(
-            "`max_tuning_runs` must be non-negative, instead got {}.".format(max_tuning_runs,)
+        raise MlflowException(
+            message=(
+                "`max_tuning_runs` must be non-negative, instead got {}.".format(max_tuning_runs)
+            ),
+            error_code=INVALID_PARAMETER_VALUE,
         )
 
     if not _is_supported_version():
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -693,6 +693,11 @@ def autolog(
         MAX_ENTITY_KEY_LENGTH,
     )
 
+    if max_tuning_runs is not None and max_tuning_runs < 0:
+        raise ValueError(
+            "`max_tuning_runs` must be non-negative, instead got {}.".format(max_tuning_runs,)
+        )
+
     if not _is_supported_version():
         warnings.warn(
             "Autologging utilities may not work properly on scikit-learn < {} ".format(
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -656,15 +656,15 @@ def autolog(
                    autologging. If ``False``, show all events and warnings during scikit-learn
                    autologging.
     :param max_tuning_runs: The maximum number of child Mlflow runs created for hyperparameter
-                            search estimators. To create an mlflow child run for the
-                            best k results(based on `rank_test_score` values), set max_tuning_runs
-                            to k. The default value is to track the best 5 search parameter sets.
-                            If `max_tuning_runs=None`, then a child run is created for each search
-                            parameter set. Note: In the case of multi-metric evaluation with a
-                            custom scorer, the first scorer’s `rank_test_score_<scorer_name>` will
-                            be used to select the best k results. To change metric used for
-                            selecting best k results, change ordering of dict passed as `scoring`
-                            parameter for estimator.
+                            search estimators. To create child runs for the best `k` results from
+                            the search, set `max_tuning_runs` to `k`. The default value is to track
+                            the best 5 search parameter sets. If `max_tuning_runs=None`, then
+                            a child run is created for each search parameter set. Note: The best k
+                            results is based on ordering in `rank_test_score`. In the case of
+                            multi-metric evaluation with a custom scorer, the first scorer’s
+                            `rank_test_score_<scorer_name>` will be used to select the best k
+                            results. To change metric used for selecting best k results, change
+                            ordering of dict passed as `scoring` parameter for estimator.
     """
     import pandas as pd
     import sklearn
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -463,6 +463,7 @@ def autolog(
     exclusive=False,
     disable_for_unsupported_versions=False,
     silent=False,
+    max_hyper_param_runs=5,
 ):  # pylint: disable=unused-argument
     """
     Enables (or disables) and configures autologging for scikit-learn estimators.
@@ -654,6 +655,11 @@ def autolog(
     :param silent: If ``True``, suppress all event logs and warnings from MLflow during scikit-learn
                    autologging. If ``False``, show all events and warnings during scikit-learn
                    autologging.
+    :param max_hyper_param_runs: The max number of child mlflow runs to be created for
+                                 hyper parameter search based estimators. If one only cares about
+                                 creating an mlflow child run for the best k results from the
+                                 search, then set max_hyper_param_runs to k. The default value is
+                                 to track the best 5 hyper paramters.
     """
     import pandas as pd
     import sklearn
@@ -824,6 +830,7 @@ def autolog(
                     _create_child_runs_for_parameter_search(
                         cv_estimator=estimator,
                         parent_run=mlflow.active_run(),
+                        max_hyper_param_runs=max_hyper_param_runs,
                         child_tags=child_tags,
                     )
                 except Exception as e:
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -463,7 +463,7 @@ def autolog(
     exclusive=False,
     disable_for_unsupported_versions=False,
     silent=False,
-    max_hyper_param_runs=5,
+    max_tuning_runs=5,
 ):  # pylint: disable=unused-argument
     """
     Enables (or disables) and configures autologging for scikit-learn estimators.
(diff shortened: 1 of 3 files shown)


mlflow/sklearn/__init__.py
```diff
@@ -463,6 +463,7 @@ def autolog(
     exclusive=False,
     disable_for_unsupported_versions=False,
     silent=False,
+    max_hyper_param_runs=5,
```
dbczumar, line 466:
```suggestion
    max_tuning_runs=5,
```

mlflow/sklearn/__init__.py
```diff
@@ -654,6 +655,11 @@ def fetch_logged_data(run_id):
     :param silent: If ``True``, suppress all event logs and warnings from MLflow during scikit-learn
                    autologging. If ``False``, show all events and warnings during scikit-learn
                    autologging.
+    :param max_hyper_param_runs: The max number of child mlflow runs to be created for
+                                 hyper parameter search based estimators. If one only cares about
```
dbczumar, line 659:
```suggestion
    :param max_hyper_param_runs: The maximum number of child MLflow runs to be created for
                                 hyperparameter search estimators. If one only cares about
```

mlflow/sklearn/__init__.py
```diff
@@ -654,6 +655,11 @@ def fetch_logged_data(run_id):
...
+                                 creating an mlflow child run for the best k results from the
+                                 search, then set max_hyper_param_runs to k. The default value is
+                                 to track the best 5 hyper paramters.
```
dbczumar, line 662: Can we refactor this phrasing to more succinctly tell the user that the best child runs are selected and also tell them how these runs are selected (i.e. via `rank_test_score` or the first available custom metric)?

mlflow/sklearn/__init__.py
dbczumar: Can we also add documentation indicating that the user can specify `None` to omit a maximum limit and log a child run for each set of explored hyperparameters?

mlflow/sklearn/utils.py
```diff
@@ -615,9 +624,19 @@ def _create_child_runs_for_parameter_search(cv_estimator, parent_run, child_tags
...
     cv_results_df = pd.DataFrame.from_dict(cv_estimator.cv_results_)
-    for _, result_row in cv_results_df.iterrows():
+    rank_column_name = "rank_test_score"
+    if rank_column_name not in cv_results_df.columns.values:
+        rank_column_name = first_custom_rank_column(cv_results_df)
+        warnings.warn(
+            "Top {} child runs will be created based on ordering in {} column. ".format(
+                max_hyper_param_runs, rank_column_name,
+            )
+            + "One can choose not to limit the number of child runs created if so desired."
```
dbczumar, line 635: How can users do this? I don't see where this PR provides that capability. I'd propose introducing `None` as a valid parameter value to satisfy this use case; we can check for `None` here and avoid limiting results.

dbczumar, line 635: Can we address the user in second person (`you`) and tell them specifically how to remove the child runs limit (i.e. by passing `None` for the autologging call option)? Can we also tell them that the metric being selected is the first one found & that they can reorder their custom metrics to in order to choose a different one for child run creation?

mlflow/sklearn/utils.py
```diff
@@ -597,6 +600,12 @@ def _create_child_runs_for_parameter_search(cv_estimator, parent_run, child_tags
     """
     import pandas as pd
 
+    def first_custom_rank_column(df):
+        column_names = df.columns.values
+        for i, col_name in enumerate(column_names):
+            if "rank_test_" in col_name:
+                return column_names[i]
```
dbczumar, line 607: Nit: Do we need the index `i` here? Can we do `for col_name in column_names` instead?

mlflow/sklearn/utils.py
```diff
@@ -597,6 +600,12 @@ def _create_child_runs_for_parameter_search(cv_estimator, parent_run, child_tags
     """
     import pandas as pd
 
+    def first_custom_rank_column(df):
```
dbczumar, line 603: What happens if  a custom rank column is not identified? Perhaps we should pick the first `n` runs and warn the user?

mohamad-arabi: It should be identified, that is what we expect from [cv_results_](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn.model_selection.GridSearchCV). If a rank column is not identified then I think something bigger has gone wrong in the training and this will be gracefully handled because all of [_create_child_runs_for_parameter_search](https://github.com/mlflow/mlflow/blob/master/mlflow/sklearn/__init__.py#L819) is wrapped in a try catch

mlflow/sklearn/__init__.py
```diff
@@ -654,6 +655,17 @@ def fetch_logged_data(run_id):
     :param silent: If ``True``, suppress all event logs and warnings from MLflow during scikit-learn
                    autologging. If ``False``, show all events and warnings during scikit-learn
                    autologging.
+    :param max_tuning_runs: The maximum number of child Mlflow runs to be created for
+                            hyperparameter search estimators. If one only cares about
+                            creating an mlflow child run for the best k results(based on
+                            `rank_test_score` values) from the search, then set max_tuning_runs
+                            to k. The default value is to track the best 5 search parameter sets.
```
dbczumar, line 662: As discussed offline, can we say something like

```
To create child runs for the best `k` results from the search, set `max_tuning_runs` to `k`.
```

instead?

mlflow/sklearn/__init__.py
```diff
@@ -654,6 +655,17 @@ def fetch_logged_data(run_id):
     :param silent: If ``True``, suppress all event logs and warnings from MLflow during scikit-learn
                    autologging. If ``False``, show all events and warnings during scikit-learn
                    autologging.
+    :param max_tuning_runs: The maximum number of child Mlflow runs to be created for
+                            hyperparameter search estimators. If one only cares about
+                            creating an mlflow child run for the best k results(based on
+                            `rank_test_score` values) from the search, then set max_tuning_runs
```
dbczumar, line 661: Can we remove (`based on rank_test_score values`) and move this discussion about how the best values are selected to the bottom of the docstring along with the information about custom scorers? These two are closely related.

mlflow/sklearn/utils.py
dbczumar: I don't see `rank_test` explicitly defined in the contract of [cv_results_](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn.model_selection.GridSearchCV), though it's used in an example.

Regardless, try/catch is fine as long as we have test coverage here. Can we confirm that we test our handling of child run creation failures for hyperparameter search and introduce coverage if we don't?

mlflow/sklearn/utils.py
```diff
@@ -615,9 +624,24 @@ def _create_child_runs_for_parameter_search(cv_estimator, parent_run, child_tags
...
+    else:
+        rank_column_name = "rank_test_score"
+        if rank_column_name not in cv_results_df.columns.values:
+            rank_column_name = first_custom_rank_column(cv_results_df)
+            warnings.warn(
+                "Top {} child runs will be created based on ordering in {} column. ".format(
+                    max_tuning_runs, rank_column_name,
+                )
+                + "You can choose not to limit the number of child runs created by"
+                + "setting `max_tuning_runs=None`."
```
dbczumar, line 640: Nit:

```suggestion
                + " setting `max_tuning_runs=None`."
```

mlflow/sklearn/utils.py
mohamad-arabi: Screen Shot 2021-05-24 at 12 29 37 PM

The example is the only location where they provide docs for `cv_results_`

mohamad-arabi: Also we already do have test coverage for graceful failing of [_create_child_runs_for_parameter_search()](https://github.com/mlflow/mlflow/blob/master/tests/sklearn/test_sklearn_autolog.py#L848)

dbczumar: Not to be pedantic, but the above text is just an example indicating how a Pandas DataFrame with particular columns (e.g. `rank_test_score`) is translated into a `cv_results_` dictionary; it does not explicitly stipulate that `rank_test_*` metrics are always computed, even if that is the case. Regardless, glad we have coverage! Thank you for auditing!

mohamad-arabi: No problem, thanks for the diligent review of my PR!

mohamad-arabi: That makes sense, they never explicitly listed what are the contents of `cv_results_` technically.

mlflow/sklearn/utils.py
```diff
@@ -615,9 +624,30 @@ def _create_child_runs_for_parameter_search(cv_estimator, parent_run, child_tags
...
+                + " You can choose not to limit the number of child runs created by"
+                + " setting `max_tuning_runs=None`."
+            )
+        cv_results_best_n_df = cv_results_df.nsmallest(max_tuning_runs, rank_column_name)
+        # Log how many child runs will be created vs omitted.
+        rest = len(cv_results_df) - max_tuning_runs
+        if rest > 0:
+            _logger.info(
+                "Logging the %s best runs, %s others will be omitted.", max_tuning_runs, rest,
+            )
```
dbczumar, line 648: Nit: This message becomes a bit odd when `max_tuning_runs` is zero or one.

"Logging the best 0 runs, 15 others will be omitted."

"Logging the best 1 runs, 14 others will be omitted."

Can we handle these cases?

mlflow/sklearn/__init__.py
```diff
@@ -682,6 +693,11 @@ def fetch_logged_data(run_id):
         MAX_ENTITY_KEY_LENGTH,
     )
 
+    if max_tuning_runs is not None and max_tuning_runs < 0:
+        raise ValueError(
```
dbczumar, line 697: Can we raise `MlflowException` with error code `INVALID_PARAMETER_VALUE` instead?

dbczumar (approved):
@mohamad-arabi LGTM once two comments have been addressed. Awesome work!

PR #4402, opened 2021-05-27
Use batch credential fetch APIs for Databricks artifact logging

dbczumar:

Use batch credential fetch APIs for Databricks artifact logging

## How is this patch tested?

- Unit tests
- Manual testing on Databricks to ensure that uploads / downloads work properly in the context of logging / loading models and logging / loading directories containing hundreds of files

## Release Notes

Improved artifact logging performance on Databricks

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/java/client/src/main/java/com/databricks/api/proto/mlflow/DatabricksArtifacts.java
+++ b/mlflow/java/client/src/main/java/com/databricks/api/proto/mlflow/DatabricksArtifacts.java
@@ -7235,17 +7235,17 @@ public final class DatabricksArtifacts {
       "\342?1\n/com.databricks.mlflow.api.MlflowTra" +
       "ckingMessage*V\n\026ArtifactCredentialType\022\021" +
       "\n\rAZURE_SAS_URI\020\001\022\025\n\021AWS_PRESIGNED_URL\020\002" +
-      "\022\022\n\016GCP_SIGNED_URL\020\0032\342\002\n DatabricksMlflo" +
-      "wArtifactsService\022\233\001\n\025getCredentialsForR" +
+      "\022\022\n\016GCP_SIGNED_URL\020\0032\344\002\n DatabricksMlflo" +
+      "wArtifactsService\022\234\001\n\025getCredentialsForR" +
       "ead\022\035.mlflow.GetCredentialsForRead\032&.mlf" +
-      "low.GetCredentialsForRead.Response\";\362\206\0317" +
-      "\n3\n\003GET\022&/mlflow/artifacts/credentials-f" +
-      "or-read\032\004\010\002\020\000\020\003\022\237\001\n\026getCredentialsForWri" +
-      "te\022\036.mlflow.GetCredentialsForWrite\032\'.mlf" +
-      "low.GetCredentialsForWrite.Response\"<\362\206\031" +
-      "8\n4\n\003GET\022\'/mlflow/artifacts/credentials-" +
-      "for-write\032\004\010\002\020\000\020\003B,\n\037com.databricks.api." +
-      "proto.mlflow\220\001\001\240\001\001\342?\002\020\001"
+      "low.GetCredentialsForRead.Response\"<\362\206\0318" +
+      "\n4\n\004POST\022&/mlflow/artifacts/credentials-" +
+      "for-read\032\004\010\002\020\000\020\003\022\240\001\n\026getCredentialsForWr" +
+      "ite\022\036.mlflow.GetCredentialsForWrite\032\'.ml" +
+      "flow.GetCredentialsForWrite.Response\"=\362\206" +
+      "\0319\n5\n\004POST\022\'/mlflow/artifacts/credential" +
+      "s-for-write\032\004\010\002\020\000\020\003B,\n\037com.databricks.ap" +
+      "i.proto.mlflow\220\001\001\240\001\001\342?\002\020\001"
     };
     com.google.protobuf.Descriptors.FileDescriptor.InternalDescriptorAssigner assigner =
         new com.google.protobuf.Descriptors.FileDescriptor.    InternalDescriptorAssigner() {
--- a/mlflow/java/client/src/main/java/com/databricks/api/proto/mlflow/DatabricksArtifacts.java
+++ b/mlflow/java/client/src/main/java/com/databricks/api/proto/mlflow/DatabricksArtifacts.java
@@ -2751,32 +2751,68 @@ public final class DatabricksArtifacts {
 
     /**
      * <pre>
-     * The artifact path, relative to the Run's artifact root location, for which to
+     * The artifact paths, relative to the Run's artifact root location, for which to
      * fetch artifact read credentials
      * </pre>
      *
-     * <code>optional string path = 2 [(.mlflow.validate_required) = true];</code>
+     * <code>repeated string path = 2;</code>
      */
-    boolean hasPath();
+    java.util.List<java.lang.String>
+        getPathList();
     /**
      * <pre>
-     * The artifact path, relative to the Run's artifact root location, for which to
+     * The artifact paths, relative to the Run's artifact root location, for which to
      * fetch artifact read credentials
      * </pre>
      *
-     * <code>optional string path = 2 [(.mlflow.validate_required) = true];</code>
+     * <code>repeated string path = 2;</code>
      */
-    java.lang.String getPath();
+    int getPathCount();
     /**
      * <pre>
-     * The artifact path, relative to the Run's artifact root location, for which to
+     * The artifact paths, relative to the Run's artifact root location, for which to
      * fetch artifact read credentials
      * </pre>
      *
-     * <code>optional string path = 2 [(.mlflow.validate_required) = true];</code>
+     * <code>repeated string path = 2;</code>
+     */
+    java.lang.String getPath(int index);
+    /**
+     * <pre>
+     * The artifact paths, relative to the Run's artifact root location, for which to
+     * fetch artifact read credentials
+     * </pre>
+     *
+     * <code>repeated string path = 2;</code>
      */
     com.google.protobuf.ByteString
-        getPathBytes();
+        getPathBytes(int index);
+
+    /**
+     * <pre>
+     * Token specifying the page of credentials to fetch for large requests that require pagination
+     * </pre>
+     *
+     * <code>optional string page_token = 3;</code>
+     */
+    boolean hasPageToken();
+    /**
+     * <pre>
+     * Token specifying the page of credentials to fetch for large requests that require pagination
+     * </pre>
+     *
+     * <code>optional string page_token = 3;</code>
+     */
+    java.lang.String getPageToken();
+    /**
+     * <pre>
+     * Token specifying the page of credentials to fetch for large requests that require pagination
+     * </pre>
+     *
+     * <code>optional string page_token = 3;</code>
+     */
+    com.google.protobuf.ByteString
+        getPageTokenBytes();
   }
   /**
    * Protobuf type {@code mlflow.GetCredentialsForRead}
@@ -2792,7 +2828,8 @@ public final class DatabricksArtifacts {
     }
     private GetCredentialsForRead() {
       runId_ = "";
-      path_ = "";
+      path_ = com.google.protobuf.LazyStringArrayList.EMPTY;
+      pageToken_ = "";
     }
 
     @java.lang.Override
@@ -2827,8 +2864,17 @@ public final class DatabricksArtifacts {
             }
             case 18: {
               com.google.protobuf.ByteString bs = input.readBytes();
+              if (!((mutable_bitField0_ & 0x00000002) == 0x00000002)) {
+                path_ = new com.google.protobuf.LazyStringArrayList();
+                mutable_bitField0_ |= 0x00000002;
+              }
+              path_.add(bs);
+              break;
+            }
+            case 26: {
+              com.google.protobuf.ByteString bs = input.readBytes();
               bitField0_ |= 0x00000002;
-              path_ = bs;
+              pageToken_ = bs;
               break;
             }
             default: {
@@ -2846,6 +2892,9 @@ public final class DatabricksArtifacts {
         throw new com.google.protobuf.InvalidProtocolBufferException(
             e).setUnfinishedMessage(this);
       } finally {
+        if (((mutable_bitField0_ & 0x00000002) == 0x00000002)) {
+          path_ = path_.getUnmodifiableView();
+        }
         this.unknownFields = unknownFields.build();
         makeExtensionsImmutable();
       }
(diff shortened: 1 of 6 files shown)


mlflow/protos/databricks_artifacts.proto
```diff
@@ -23,7 +23,7 @@ service DatabricksMlflowArtifactsService {
   rpc getCredentialsForRead (GetCredentialsForRead) returns (GetCredentialsForRead.Response) {
```
dbczumar, line 23: Update client proto definitions to use batch-supported POST endpoints exposed by the backend.

mlflow/protos/databricks_artifacts.proto
```diff
@@ -23,7 +23,7 @@ service DatabricksMlflowArtifactsService {
...
     option (rpc) = {
       endpoints: [{
-        method: "GET",
+        method: "POST",
```
dbczumar, line 26: Update client proto definitions to use batch-supported POST endpoints exposed by the Databricks backend.

mlflow/java/client/src/main/java/com/databricks/api/proto/mlflow/DatabricksArtifacts.java
```diff
@@ -2751,32 +2751,68 @@ public ArtifactCredentialInfo parsePartialFrom(
```
dbczumar, line 2751: Automatically regenerated to reflect changes in `mlflow/protos/databricks_artifacts.proto` (see https://github.com/mlflow/mlflow/pull/4402/files#r641035500).

mlflow/protos/databricks_artifacts_pb2.py
```diff
@@ -24,7 +24,7 @@
   package='mlflow',
```
dbczumar, line 24: Automatically regenerated to reflect changes in `mlflow/protos/databricks_artifacts.proto` (see https://github.com/mlflow/mlflow/pull/4402/files#r641035500).

mlflow/utils/__init__.py
```diff
@@ -63,6 +63,11 @@ def reraise(tp, value, tb=None):
         tb = None
 
 
+def chunk_list(l, chunk_size):
```
dbczumar, line 66: Define a utility to chunk a list, which is used to create GetCredentialsForRead and GetCredentialsForWrite requests that respect the maximum request size of 2000 artifact paths.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -139,17 +141,56 @@ def _get_run_artifact_root(self, run_id):
...
+                 run-relative artifact `paths` within the MLflow Run specified by `run_id`.
+        """
+        return self._get_credential_infos(GetCredentialsForWrite, run_id, paths)
+
+    def _get_read_credential_infos(self, run_id, paths):
+        """
+        :return: A list of `ArtifactCredentialInfo` objects providing read access to the specified
+                 run-relative artifact `paths` within the MLflow Run specified by `run_id`.
+        """
+        return self._get_credential_infos(GetCredentialsForRead, run_id, paths)
```
dbczumar, line 193: These methods are introduced / modified in order to leverage new batch request capabilities that have been added to `get-credentials-for-read` and `get-credentials-for-write`. Whereas each API call previously requested credentials for a single `path` within the specified `run_id`, each API call now requests credentials for up to 2000 paths in a single request.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -180,9 +221,9 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
                             "Failed to authorize request, possibly due to credential expiration."
                             " Refreshing credentials and trying again..."
                         )
-                        credentials = self._get_write_credentials(
-                            self.run_id, artifact_path
-                        ).credentials
+                        credentials = self._get_write_credential_infos(
+                            run_id=self.run_id, paths=[artifact_path]
+                        )[0]
```
dbczumar, line 226: Upon 403 errors from the cloud storage provider (S3, WASBS, etc), we fetch new credentials for the file being uploaded; this case rarely occurs and applies to large files which must be uploaded in many chunks. Credentials are fetched for a single path.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -223,33 +264,32 @@ def _signed_url_upload_file(self, credentials, local_file):
         except Exception as err:
             raise MlflowException(err)
 
-    def _upload_to_cloud(self, cloud_credentials, local_file, artifact_path):
-        if cloud_credentials.credentials.type == ArtifactCredentialType.AZURE_SAS_URI:
-            self._azure_upload_file(cloud_credentials.credentials, local_file, artifact_path)
-        elif cloud_credentials.credentials.type in [
+    def _upload_to_cloud(
+        self, cloud_credential_info, src_file_path, dst_run_relative_artifact_path
```
dbczumar, line 268: Tweaked some parameter names to be more descriptive and consistent with proto field naming (e.g. `cloud_credentials -> cloud_credential_info` to indicate a single entry within the `credential_infos` proto response field)

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -223,33 +264,32 @@ def _signed_url_upload_file(self, credentials, local_file):
...
         ]:
-            self._signed_url_upload_file(cloud_credentials.credentials, local_file)
+            self._signed_url_upload_file(cloud_credential_info, src_file_path)
         else:
             raise MlflowException(
                 message="Cloud provider not supported.", error_code=INTERNAL_ERROR
             )
 
-    def _download_from_cloud(self, cloud_credential, local_file_path):
+    def _download_from_cloud(self, cloud_credential_info, dst_local_file_path):
```
dbczumar, line 288: Tweaked some parameter names to be more descriptive and consistent with proto field naming (e.g. `cloud_credentials -> cloud_credential_info` to indicate a single entry within the `credential_infos` proto response field)

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -352,61 +452,79 @@ def list_artifacts(self, path=None):
         return infos
 
     def _download_file(self, remote_file_path, local_path):
-        run_relative_remote_file_path = posixpath.join(
```
dbczumar, line 355: In the interest of deduplicating credential resolution logic, I've centralized the logic for both file and directory downloads in `download_artifacts()` and left an explanatory comment in the now-empty-bodied `_download_file` function.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -352,61 +452,79 @@ def list_artifacts(self, path=None):
...
-                # root location, to futures representing their corresponding download operations
-                "src_paths_to_futures_map",
+                "dst_local_path",
+                # A future representing the artifact download operation
+                "download_future",
             ],
         )
 
-        def download_artifact(src_artifact_path, dst_local_dir_path):
+        def async_download_file_artifacts_from_paths(src_artifact_paths, dst_local_dir_path):
```
dbczumar, line 478: file and directory downloads now rely on this method, which accepts a collection of paths to be downloaded and returns a collection of inflight download objects, each of which specifies the source artifact path, destination local filesystem path, and future.

Directory artifact downloads call this for each level in the artifact directory tree, and file artifact downloads call this with a single file input.

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -197,22 +197,21 @@ def test_log_artifact_azure(
...
+            write_credential_infos_mock.return_value = [mock_credential_info]
             azure_upload_mock.return_value = None
             databricks_artifact_repo.log_artifact(test_file.strpath, artifact_path)
-            write_credentials_mock.assert_called_with(MOCK_RUN_ID, expected_location)
+            write_credential_infos_mock.assert_called_with(
+                run_id=MOCK_RUN_ID, paths=[expected_location]
+            )
             azure_upload_mock.assert_called_with(
-                mock_credentials, test_file.strpath, expected_location
+                mock_credential_info, test_file.strpath, expected_location
```
dbczumar, line 214: This change, and many others of the same form, just accommodate the renaming of `_get_write_credentials` and `_get_read_credentials_` to `_get_write_credential_infos` and `_get_read_credential_infos`, respectively, as well as the adjustment of these methods' return types from `GetCredentialsForRead.Response` and `GetCredentialsForWrite.Response` to lists of `ArtifactCredentialInfo` objects.

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -591,6 +556,226 @@ def test_paginated_list_artifacts(self, databricks_artifact_repo):
...
+                        GetCredentialsForRead(run_id=MOCK_RUN_ID, path=["testpath"], page_token="2")
+                    ),
+                    mock.call(
+                        GetCredentialsForRead(run_id=MOCK_RUN_ID, path=["testpath"], page_token="3")
+                    ),
+                ]
+            )
+            assert call_endpoint_mock.call_count == 3
+
+    def test_get_read_credential_infos_respects_max_request_size(self, databricks_artifact_repo):
```
dbczumar, line 612: As discussed in https://github.com/mlflow/mlflow/pull/4402/files#r641201273, these are 2 of 4 new test cases

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -591,6 +556,226 @@ def test_paginated_list_artifacts(self, databricks_artifact_repo):
...
+                    mock.call(
+                        GetCredentialsForWrite(
+                            run_id=MOCK_RUN_ID, path=["testpath"], page_token="3"
+                        )
+                    ),
+                ]
+            )
+            assert call_endpoint_mock.call_count == 3
+
+    def test_get_write_credential_infos_respects_max_request_size(self, databricks_artifact_repo):
```
dbczumar, line 724: As discussed in https://github.com/mlflow/mlflow/pull/4402/files#r641201273, these are the other 2 of 4 new test cases

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -429,46 +414,26 @@ def test_log_artifact_with_relative_path(self, test_file, artifact_path, expecte
...
-            write_credentials_mock.assert_called_with(MOCK_RUN_ID, expected_location)
+            write_credential_infos_mock.assert_called_with(
+                run_id=MOCK_RUN_ID, paths=[expected_location]
+            )
             upload_mock.assert_called_with(
-                write_credentials_response_proto, test_file.strpath, expected_location
-            )
-
-    @pytest.mark.parametrize("artifact_path", [None, "output/", ""])
-    def test_log_artifacts(self, databricks_artifact_repo, test_dir, artifact_path):
```
dbczumar, line 453: This test case previously depended on the assumption that `log_artifacts()` was implemented by repeatedly calling `log_artifact()`, which is no longer the case due to our use of batched credential generation. This test case has been obviated by `test_artifact_logging`.

mlflow/protos/databricks_artifacts.proto
```diff
@@ -100,14 +100,25 @@ message GetCredentialsForRead {
   // The ID of the MLflow Run for which to fetch artifact read credentials
   optional string run_id = 1 [(validate_required) = true];
 
-  // The artifact path, relative to the Run's artifact root location, for which to
+  // The artifact paths, relative to the Run's artifact root location, for which to
   // fetch artifact read credentials
```
jinzhang21, line 104: Add comment: must not be empty.

mlflow/protos/databricks_artifacts.proto
```diff
@@ -100,14 +100,25 @@ message GetCredentialsForRead {
...
+  optional string page_token = 3;
 
   message Response {
 
-    // Credentials for reading from the specified artifact location
-    optional ArtifactCredentialInfo credentials = 1;
+    // DEPRECATED: Credentials for reading from a single specified artifact location.
+    // Older versions of the MLflow client use this field to retrieve credentials via point fetches.
+    // This field is populated with the first element of the new `credential_infos` field
+    // optional ArtifactCredentialInfo credentials = 1 [deprecated = true];
```
jinzhang21, line 115: Do you intend to comment this line out? If you want to make it forward compatible only, it may worth just delete those lines, and reserve the field no. 1 in the end.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -54,6 +55,7 @@
     service: extract_api_info_for_service(service, _REST_API_PATH_PREFIX)
     for service in [MlflowService, DatabricksMlflowArtifactsService]
 }
+_MAX_CREDENTIALS_REQUEST_SIZE = 2000  # Max number of artifact paths in a single credentials request
```
jinzhang21, line 58: Leave a blank line between L57 and L58?

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -139,17 +141,56 @@ def _get_run_artifact_root(self, run_id):
...
+        """
+        credential_infos = []
+        page_token = None
+
+        for paths_chunk in chunk_list(paths, _MAX_CREDENTIALS_REQUEST_SIZE):
+            while True:
+                if page_token:
+                    json_body = message_to_json(
+                        request_message_class(
+                            run_id=run_id, path=paths_chunk, page_token=page_token
```
jinzhang21, line 162: If the page_token defaults to "" instead of None, could the if-else clause in this block be removed?

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -139,17 +141,56 @@ def _get_run_artifact_root(self, run_id):
...
+                        request_message_class(run_id=run_id, path=paths_chunk)
+                    )
+
+                response = self._call_endpoint(
+                    DatabricksMlflowArtifactsService, request_message_class, json_body
+                )
+                credential_infos += response.credential_infos
+                page_token = response.next_page_token
+
+                if not page_token or len(response.credential_infos) == 0:
```
jinzhang21, line 176: Under what conditions page_token is not empty but len(response.credential_infos) == 0?

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -195,9 +236,9 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
                         "Failed to authorize request, possibly due to credential expiration."
                         " Refreshing credentials and trying again..."
                     )
-                    credentials = self._get_write_credentials(
-                        self.run_id, artifact_path
-                    ).credentials
+                    credentials = self._get_write_credential_infos(
```
jinzhang21, line 239: credentials -> credential or credential_info

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -591,6 +556,226 @@ def test_paginated_list_artifacts(self, databricks_artifact_repo):
...
+        """
+        Verifies that the `_get_read_credential_infos` method, which is used to resolve read access
+        credentials for a collection of artifacts, handles paginated responses properly, issuing
+        incremental requests until all pages have been consumed
+        """
+        # Create 3 chunks of paths, two of which have the maximum request size and one of which
+        # is smaller than the maximum chunk size. Aggregate and pass these to
+        # `_get_read_credential_infos`, validating that this method decomposes the aggregate
+        # list into these expected chunks and makes 3 separate requests
+        paths_chunk_1 = ["path1"] * 2000
```
jinzhang21, line 622: Change 2000 to _MAX_CREDENTIALS_REQUEST_SIZE, ditto below.

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -591,6 +556,226 @@ def test_paginated_list_artifacts(self, databricks_artifact_repo):
...
+            call_endpoint_mock.side_effect = [
+                GetCredentialsForRead.Response(credential_infos=credential_infos_mock_1,),
+                GetCredentialsForRead.Response(credential_infos=credential_infos_mock_2,),
+                GetCredentialsForRead.Response(credential_infos=credential_infos_mock_3,),
+            ]
+
+            databricks_artifact_repo._get_read_credential_infos(
+                MOCK_RUN_ID, paths_chunk_1 + paths_chunk_2 + paths_chunk_3,
+            )
+            assert call_endpoint_mock.call_count == message_mock.call_count == 3
```
jinzhang21, line 658: asserting the 2 conditions separately in case one fails the other doesn't.
assert a == 3
assert b == 3

jinzhang21 (approved):
Overall LGTM. Some minor comments.

mlflow/protos/databricks_artifacts.proto
dbczumar: Done!

mlflow/protos/databricks_artifacts.proto
dbczumar: ooh, reserved is great! I've replaced this with `reserved 1` and shortened the comment. Thanks!

mlflow/store/artifact/databricks_artifact_repo.py
dbczumar: Good idea - the positioning of this definition is a bit odd. I've moved this up above `_SERVICE_AND_METHOD_TO_INFO` and below `_DOWNLOAD_CHUNK_SIZE` so that it's next to other integer constants.

mlflow/store/artifact/databricks_artifact_repo.py
dbczumar: Unfortunately, it doesn't seem to be the case. Specifying an empty string results in:

```
mlflow.exceptions.RestException: INVALID_PARAMETER_VALUE: Malformed token supplied with artifact credentials request.
```

However, `None` seems to work! Not sure why I had this logic here previously. I've removed the if / else here and in the definition of `list_artifacts()`.

mlflow/store/artifact/databricks_artifact_repo.py
dbczumar: If the backend introduces an error in page token population, this could happen. This just provides some added fault tolerance.

mlflow/store/artifact/databricks_artifact_repo.py
dbczumar: Fixed the naming here and in the block above. Thanks!

tests/store/artifact/test_databricks_artifact_repo.py
dbczumar: Good point! Done!

tests/store/artifact/test_databricks_artifact_repo.py
dbczumar: I'm concerned that this might come back to bite us of `_MAX_CREDENTIALS_REQUEST_SIZE` ends up being erroneously changed to some incorrect value like `3`.  Let me know if you feel strongly about this refactoring though!

dbczumar (commented):
@jinzhang21 Thanks so much for the very helpful feedback! I've addressed your comments, all of which resulted in changes except for https://github.com/mlflow/mlflow/pull/4402/files#r641717612, where I'm a bit concerned that using the conf value in the test may lead to missed diagnosis of a potential issue if the conf eventually becomes nonsensical.

mlflow/store/artifact/databricks_artifact_repo.py
jinzhang21: Hmm. This could obscure backend issues that should be fixed immediately. I'd prefer to actually expose the issue instead, e.g. assert len(response.credential_infos) somewhere in the while loop.

tests/store/artifact/test_databricks_artifact_repo.py
jinzhang21: Hardcoding is fine. It just might break if the size is changed to something bigger later, not a big deal though.

dbczumar: Based on offline discussion, I've changed the test to use the conf value and added an assert on the conf value with a descriptive message indicating that it should only be changed in response to a backend limit change

mlflow/store/artifact/databricks_artifact_repo.py
dbczumar: I think we should weigh the tradeoff of expose the backend issue quickly vs. disrupting production workflows. Such an assert has the potential to break all artifact logging on Databricks.

PR #4405, opened 2021-05-28
Revert patched code when disabling autolog

mohamad-arabi:

If `autolog` is invoked with `disable=True` param, then short circuit `autolog()` and revert all patched methods for that autologging integration. 

## How is this patch tested?
Added unit tests.

Components 

Interface 

Language 

Integrations


mlflow/utils/autologging_utils/safety.py
```diff
@@ -539,7 +541,16 @@ def call_original(*og_args, **og_kwargs):
                 else:
                     return original(*args, **kwargs)
 
-    _wrap_patch(destination, function_name, safe_patch_function)
+    new_patch = _wrap_patch(destination, function_name, safe_patch_function)
+    _store_patch(autologging_integration, new_patch)
+
+
+def revert_patches(autologging_integration):
+    if autologging_integration in _AUTOLOGGING_PATCHES:
```
mohamad-arabi, line 549: Added check for existance, because if not throws a  `None is non-iterable` exception.

mlflow/utils/gorilla.py
```diff
@@ -331,6 +334,41 @@ def apply(patch):
...
+            % (patch.destination.__name__,)
+        )
+
+    original_name = _ORIGINAL_NAME % (patch.name,)
+    if not getattr(patch.destination, original_name):
+        return
+
+    setattr(patch.destination, patch.name, original)
+
+    if original_name in patch.destination.__dict__:
```
mohamad-arabi, line 366: Only delete added attributes by gorilla monkey-patching, if attribute is defined on the class as opposed to inherited from a parent class.

mlflow/utils/autologging_utils/safety.py
dbczumar: @mohamad-arabi Can we do the following to avoid the duplicate existence check?

```
for patch in _AUTOLOGGING_PATCHES.get(autologging_integration, []):
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -539,7 +541,16 @@ def call_original(*og_args, **og_kwargs):
                 else:
                     return original(*args, **kwargs)
 
-    _wrap_patch(destination, function_name, safe_patch_function)
+    new_patch = _wrap_patch(destination, function_name, safe_patch_function)
+    _store_patch(autologging_integration, new_patch)
+
+
+def revert_patches(autologging_integration):
```
dbczumar, line 548: Can we add a small docstring for this method?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -618,6 +629,14 @@ def _wrap_patch(destination, name, patch, settings=None):
 
     patch = gorilla.Patch(destination, name, wrapped, settings=settings)
     gorilla.apply(patch)
+    return patch
+
+
+def _store_patch(autologging_integration, patch):
```
dbczumar, line 635: Can we add a small docstring for this method?

mlflow/utils/gorilla.py
dbczumar: Awesome solution! In accordance with that StackOverflow comment that we discussed offline, can we also add a `hasattr` check here too?

mlflow/utils/gorilla.py
```diff
@@ -331,6 +334,41 @@ def apply(patch):
     setattr(patch.destination, _ACTIVE_PATCH, patch)
 
 
+def revert(patch):
```
dbczumar, line 337: Can we add a notice to this method's docstring indicating that the method is taken from https://github.com/christophercrouzet/gorilla/blob/4dfa7215b8abacbdbd91e784da3d61d671a3a510/gorilla.py#L318-L351 with small additions for MLflow patching use cases?

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -343,6 +344,10 @@ def autolog(*args, **kwargs):
             config_to_store.update(kwargs)
             AUTOLOGGING_INTEGRATIONS[name] = config_to_store
 
+            if get_autologging_config(name, "disable", True):
+                revert_patches(name)
+                return
```
dbczumar, line 349: I think this logic works well for all autologging integrations except for the `mlflow.autolog()` fluent API, which makes use of the `@autologging_integration` decorator.

If we call `mlflow.autolog(disable=True)`, we should still call all of the active framework-specific autologging integrations so that their patches are reverted.

I think it's sufficient to add a special case here for `name == 'mlflow'` with an explanatory inline comment. Can we also make sure to add test coverage for this in `test_fluent_autolog.py`?

mlflow/utils/gorilla.py
mohamad-arabi: For class attributes stored in [slots](https://stackoverflow.com/questions/472000/usage-of-slots), if no additional attrbiutes are added then there will be no __dict__. Any [monkey patching will add attributes ](https://github.com/mlflow/mlflow/blob/de01ccf24474d8f9479851838292caa0d62865ec/mlflow/utils/gorilla.py#L334)forcing the class to have a __dict__ field on the class. If no code was mokey patched and the slight possibility of class to only have attributes defined via slots, i.e. __dict__  is undefined then the [getattr](https://github.com/mlflow/mlflow/blob/de01ccf24474d8f9479851838292caa0d62865ec/mlflow/utils/gorilla.py#L348) in the beginning of the method will short circuit. 

TLDR: No need to add  hasattr check for slots edge case.

mlflow/utils/autologging_utils/safety.py
mohamad-arabi: Done

mlflow/utils/autologging_utils/safety.py
mohamad-arabi: Done

mlflow/utils/autologging_utils/__init__.py
mohamad-arabi: Done, implemented logic in `revert_patches`

mlflow/utils/gorilla.py
mohamad-arabi: Done below

mlflow/utils/gorilla.py
```diff
@@ -328,7 +333,55 @@ def apply(patch):
...
+    Parameters
+    ----------
+    patch : gorilla.Patch
+        Patch.
+    Note
+    ----
+    This is only possible if the attribute :attr:`Settings.store_hit` was set
+    to ``True`` when applying the patch and overriding an existing attribute.
+
+    Notice:
```
mohamad-arabi, line 350: docstring comment

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -374,13 +380,39 @@ def patch_impl(original, *args, **kwargs):
...
+
+    # Destination method should not be wrapped if autolog is invoked with disable=True.
+    with mock.patch(
+        "mlflow.utils.autologging_utils.AutologgingEventLogger.log_patch_function_start"
+    ) as patched_fn_logger:
+        patch_destination.fn()
+        assert patch_impl_call_count == 1
+        assert not hasattr(patch_destination, "_gorilla_active_patch")
+        assert not hasattr(patch_destination, "_gorilla_original_fn")
+        assert patched_fn_logger.call_count == 0
```
dbczumar, line 415: Do we need to make this change to the test logic? Naively, it seems like the previous logic should work.

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -210,35 +214,37 @@ def log_autolog_called(self, integration, call_args, call_kwargs):
...
     mlflow.utils.import_hooks.notify_module_loaded(sklearn)
-    assert get_autologging_config("sklearn", "disable")
+    assert "sklearn" not in _AUTOLOGGING_PATCHES
 
     mlflow.autolog(disable=False)
     mlflow.utils.import_hooks.notify_module_loaded(sklearn)
     assert not get_autologging_config("sklearn", "disable")
+    assert "sklearn" in _AUTOLOGGING_PATCHES
     mlflow.sklearn.autolog(disable=True)
-    assert get_autologging_config("sklearn", "disable")
```
dbczumar, line 241: Are these changes necessary? The goal of this test is to verify that `autolog(disable=True)` propagates to framework-specific autologging integrations. It should be sufficient to check the autologging config to verify that this works properly. I don't think we need to check patch state.

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -172,7 +176,7 @@ def test_universal_autolog_calls_pyspark_immediately():
 
 
 @pytest.mark.large
-@pytest.mark.parametrize("config", [{"disable": False}, {"disable": True}])
+@pytest.mark.parametrize("config", [{"disable": False}])
```
dbczumar, line 179: Why did we remove `{"disable": True}`?

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -155,14 +155,18 @@ def test_universal_autolog_calls_specific_autologs_correctly(library, mlflow_mod
...
 def test_universal_autolog_calls_pyspark_immediately():
+    from mlflow.utils.autologging_utils.safety import _AUTOLOGGING_PATCHES
+
     mlflow.autolog()
     assert not autologging_is_disabled(mlflow.spark.FLAVOR_NAME)
+    assert mlflow.spark.FLAVOR_NAME in _AUTOLOGGING_PATCHES
 
     mlflow.autolog(disable=True)
-    assert autologging_is_disabled(mlflow.spark.FLAVOR_NAME)
+    assert mlflow.spark.FLAVOR_NAME not in _AUTOLOGGING_PATCHES
```
dbczumar, line 165: Can we add back `assert autologging_is_disabled(mlflow.spark.FLAVOR_NAME)`?

tests/autologging/test_autologging_utils.py
```diff
@@ -466,7 +466,8 @@ def test_autologging_integration_forwards_positional_and_keyword_arguments_as_ex
     def autolog(foo=7, bar=10, disable=False, silent=False):
         return foo, bar, disable
 
-    assert autolog(1, bar=2, disable=True) == (1, 2, True)
+    assert autolog(1, bar=2, disable=True) is None
```
dbczumar, line 469: I think we can remove this assert, as it isn't really testing the behavior being described in the test case.

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -346,6 +347,13 @@ def autolog(*args, **kwargs):
             config_to_store.update(kwargs)
             AUTOLOGGING_INTEGRATIONS[name] = config_to_store
 
+            # If disabling autologging using fluent api, then every active integration's autolog
+            # needs to be called with disable=True. So do not short circuit and let `mlflow.autolog()` envoke
+            # all active integrations with disable=True.
+            if name != "mlflow" and get_autologging_config(name, "disable", True):
```
mohamad-arabi, line 353: @dbczumar this is currently being tested in test_fluent_autolog.py 

[example: ](https://github.com/mlflow/mlflow/blob/bec8c2c60786ea997f6d36d2d7c5867ef1d0f8ad/tests/tracking/fluent/test_fluent_autolog.py#L162)
```
mlflow.autolog(disable=True)
 assert autologging_is_disabled(mlflow.spark.FLAVOR_NAME)
```

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -346,6 +347,13 @@ def autolog(*args, **kwargs):
             config_to_store.update(kwargs)
             AUTOLOGGING_INTEGRATIONS[name] = config_to_store
 
+            # If disabling autologging using fluent api, then every active integration's autolog
+            # needs to be called with disable=True. So do not short circuit and let `mlflow.autolog()` envoke
```
dbczumar, line 351: Extreme nit:

```suggestion
            # needs to be called with disable=True. So do not short circuit and let `mlflow.autolog()` invoke
```

mlflow/utils/autologging_utils/__init__.py
dbczumar: Fantastic!

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -346,6 +347,13 @@ def autolog(*args, **kwargs):
             config_to_store.update(kwargs)
             AUTOLOGGING_INTEGRATIONS[name] = config_to_store
 
+            # If disabling autologging using fluent api, then every active integration's autolog
+            # needs to be called with disable=True. So do not short circuit and let
+            # `mlflow.autolog()` envoke all active integrations with disable=True.
```
dbczumar, line 352:
```suggestion
            # `mlflow.autolog()` invoke all active integrations with disable=True.
```

mohamad-arabi: good catch

dbczumar (approved):
LGTM! Thanks @mohamad-arabi - great work!

PR #4409, opened 2021-05-28
Logging additional pip dependencies in requirements.txt 

stevenchen-db:

Logged additional pip dependences from the conda environment to a requirements.txt file for all flavors except PyTorch.

## How is this patch tested?

Added unit tests for all flavors that test if the conda_env additional pip dependences is the same as the written requirements.txt when calling save_model and log_model

This change logs the additional pip dependencies from `conda.yaml` in a separate `requirements.txt` file. 

Components 

Interface 

Language 

Integrations


mlflow/shap.py
```diff
@@ -459,6 +459,11 @@ def save_explainer(
     with open(os.path.join(path, conda_env_subpath), "w") as f:
         yaml.safe_dump(conda_env, stream=f, default_flow_style=False)
 
+    pip_req_subpath = "requirements.txt"
+    pip_deps = _get_additional_pip_dep(conda_env)
+    with open(os.path.join(path, pip_req_subpath), "w") as f:
```
ankit-db, line 464: Do we need requirements.txt for SHAP? cc: @harupy ?

mlflow/catboost.py
```diff
@@ -134,6 +134,11 @@ def save_model(
     with open(os.path.join(path, conda_env_subpath), "w") as f:
         yaml.safe_dump(conda_env, stream=f, default_flow_style=False)
 
+    pip_req_subpath = "requirements.txt"
+    pip_deps = _get_additional_pip_dep(conda_env)
+    with open(os.path.join(path, pip_req_subpath), "w") as f:
```
ankit-db, line 139: Since the code is exactly the same in each flavor, would it make sense to factor out the whole code snippet into `mlflow.utils.environment`

ankit-db (commented):
Will review the tests portion after the weekend, but a few comments on what's there so far

mlflow/shap.py
harupy: I think we do :)

mlflow/pytorch/__init__.py
```diff
@@ -522,6 +522,9 @@ class LinearNNModel(torch.nn.Module):
     with open(os.path.join(path, conda_env_subpath), "w") as f:
         yaml.safe_dump(conda_env, stream=f, default_flow_style=False)
 
+    pip_req_subpath = "additional_requirements.txt"
```
WeichenXu123, line 525: why this name different with other flavors ?

tests/keras/test_keras_model_export.py
```diff
@@ -438,6 +438,48 @@ def test_model_save_accepts_conda_env_as_dict(model, model_path):
...
+    # pyfunc_conf = _get_flavor_configuration(model_path=model_path, flavor_name=pyfunc.FLAVOR_NAME)
+    saved_pip_req_path = os.path.join(model_path, "requirements.txt")
+    assert os.path.exists(saved_pip_req_path)
+
+    with open(keras_custom_env, "r") as f:
+        keras_custom_env_parsed = yaml.safe_load(f)
+    with open(saved_pip_req_path, "r") as f:
+        requirements = f.read().split("\n")
+
+    assert keras_custom_env_parsed["dependencies"][-1]["pip"] == requirements
```
WeichenXu123, line 456: These code are similar in every suites. Can we create a helper method, to reduce duplicated code ?

WeichenXu123, line 456: These code are similar in every suites. Can we create a helper method, to reduce duplicated code ?

mlflow/pytorch/__init__.py
stevenchen-db: It seems that for only pytorch, there's an option to pass in a requirements_file to save_model, which was causing some tests to fail that were checking for the passed requirements file vs the one generated from the conda env. I could alternatively use a different path for that test or change all of the pip_req_subpath to "additional_requirements.txt." Not sure which option is best or if there is some other solution.

mlflow/gluon.py
```diff
@@ -222,6 +222,10 @@ def save_model(
             conda_env = yaml.safe_load(f)
     with open(os.path.join(path, conda_env_subpath), "w") as f:
         yaml.safe_dump(conda_env, stream=f, default_flow_style=False)
+
+    pip_req_subpath = "requirements.txt"
+    _log_pip_requirements(conda_env, os.path.join(path, pip_req_subpath))
```
ankit-db, line 227: It seems like `pip_req_subpath` will also be the same across all the flavors, so we may just want to not make this a parameter in the function and include it in the helper method itself.

mlflow/pyfunc/__init__.py
```diff
@@ -280,6 +282,8 @@ def add_to_model(model, loader_module, data=None, code=None, env=None, **kwargs)
         parms[DATA] = data
```
ankit-db, line 282: Should this just be `params` btw?

mlflow/pytorch/__init__.py
ankit-db: The right thing to do here is probably to merge these dependencies together, since the name `additional_requirements` is kind of nonstandard. Is this feasible to do? If so, we should consider doing this in a separate PR. If not, we can leave it like this in this PR

tests/catboost/test_catboost_model_export.py
```diff
@@ -236,6 +237,17 @@ def test_model_save_persists_specified_conda_env_in_mlflow_model_directory(
     assert read_yaml(saved_conda_env_path) == read_yaml(custom_env)
 
 
+@pytest.mark.large
+def test_model_save_persists_requirements_in_mlflow_model_directory(
+    reg_model, model_path, custom_env
+):
+    mlflow.catboost.save_model(cb_model=reg_model.model, path=model_path, conda_env=custom_env)
+
+    # pyfunc_conf = _get_flavor_configuration(model_path=model_path, flavor_name=pyfunc.FLAVOR_NAME)
```
ankit-db, line 246: Could we remove the commented out code?

tests/fastai/test_fastai_model_export.py
```diff
@@ -236,6 +236,19 @@ def test_model_save_persists_specified_conda_env_in_mlflow_model_directory(
...
 
+@pytest.mark.large
+def test_model_save_persists_requirements_in_mlflow_model_directory(
+    fastai_model, model_path, fastai_custom_env
+):
+    mlflow.fastai.save_model(
+        fastai_learner=fastai_model.model, path=model_path, conda_env=fastai_custom_env
+    )
+
+    # pyfunc_conf = _get_flavor_configuration(model_path=model_path, flavor_name=pyfunc.FLAVOR_NAME)
```
ankit-db, line 247: Same note about comment

tests/gluon/test_gluon_model_export.py
```diff
@@ -166,6 +166,17 @@ def test_model_save_persists_specified_conda_env_in_mlflow_model_directory(
     assert saved_conda_env_parsed == gluon_custom_env_parsed
 
 
+@pytest.mark.large
+def test_model_save_persists_requirements_in_mlflow_model_directory(
+    gluon_model, model_path, gluon_custom_env
+):
+    mlflow.gluon.save_model(gluon_model=gluon_model, path=model_path, conda_env=gluon_custom_env)
+
+    # pyfunc_conf = _get_flavor_configuration(model_path=model_path, flavor_name=pyfunc.FLAVOR_NAME)
```
ankit-db, line 175: Same note all the way down

mlflow/pytorch/__init__.py
ankit-db: Resolution offline was to move the Pytorch stuff out into its own smaller PR - there we'll implement the overriding behavior

ankit-db (commented):
Everything looks generally good -a few comments about some refactoring

mlflow/pytorch/__init__.py
harupy: @stevenchen-db @ankit-db The purpose of `requirements_file` in `mlflow.pytorch.{save_model, log_model}` is to make it easier to serve a saved model with [`mlflow-torchserve`](https://github.com/mlflow/mlflow-torchserve) (which serves a a model with [`torchserve`](https://github.com/pytorch/serve)). It works like this:

1. Log a pytorch model with `requirements_file`
2. Serve the model with `mlflow-torchserve`
3. `mlflow-torchserve` reads the logged `requirements_file` and pass it to `torchserve`
4. `torchserve` creates an environment based on the `requirements_file`

ankit-db (approved):
Looks good! Great job on this first OSS PR!

PR #4416, opened 2021-06-02
Efficient batching / parallel client utility for autologging implementations

dbczumar:

This PR introduces an `MlflowAutologgingQueueingClient` utility that is meant to be incorporated into autologging implementations for the purpose of more efficiently batching & asynchronously executing requests.

The PR also incorporates this client into the `mlflow.sklearn` autologging integration.

For single-estimator workflows (e.g. `LogisticRegression`), this PR reduces the number of logging operations and overall latency substantially. For example, the following workflow performs 3 fewer logging operations and takes ~ 40% less time on Databricks with this change applied:

```
import sklearn

from sklearn import datasets
iris = datasets.load_iris()

logistic = LogisticRegression(solver="saga", max_iter=200, random_state=0)
train_out = logistic.fit(iris.data, iris.target)
```

More substantial improvements are seen for meta-estimator use cases (e.g. GridSearch), with a speedup of 50 - 60%.

## How is this patch tested?

- New unit tests for `MlflowAutologgingQueueingClient` (introduced in `tests/autologging/test_autologging_client.py`)
- Existing framework-specific autologging tests from `mlflow.sklearn`.

## Release Notes

Improved performance for the `mlflow.sklearn` autologging integration.

Components 

Interface 

Language 

Integrations


mlflow/sklearn/__init__.py
```diff
@@ -37,6 +37,7 @@
     INPUT_EXAMPLE_SAMPLE_ROWS,
     resolve_input_example_and_signature,
     _get_new_training_session_class,
+    AutologgingBatchingClient,
```
dbczumar, line 40: Incorporate `AutologgingBatchingClient` into the `mlflow.sklearn` autologging flavor.

mlflow/sklearn/__init__.py
```diff
@@ -740,17 +745,17 @@ def _log_pretraining_metadata(estimator, *args, **kwargs):  # pylint: disable=un
         # process; accordingly, we avoid logging initial, untuned parameters
         # for these seed estimators.
         should_log_params_deeply = not _is_parameter_search_estimator(estimator)
-        # Chunk model parameters to avoid hitting the log_batch API limit
-        for chunk in _chunk_dict(
-            estimator.get_params(deep=should_log_params_deeply),
-            chunk_size=MAX_PARAMS_TAGS_PER_BATCH,
-        ):
-            truncated = _truncate_dict(chunk, MAX_ENTITY_KEY_LENGTH, MAX_PARAM_VAL_LENGTH)
-            try_mlflow_log(mlflow.log_params, truncated)
```
dbczumar, line 749: While not yet implemented, `AutologgingBatchingClient` will handle key / value truncation, chunking, and batching across params, tags, and metrics on a per-run basis.

mlflow/sklearn/__init__.py
```diff
@@ -740,17 +745,17 @@ def _log_pretraining_metadata(estimator, *args, **kwargs):  # pylint: disable=un
...
-
-        try_mlflow_log(mlflow.set_tags, _get_estimator_info_tags(estimator))
```
dbczumar, line 751: As part of this work, we should revisit the use of `try_mlflow_log`. As we begin to batch together logging operations, we must either find a way to permit certain operations to fail or revisit the UX / guarantees provided by autologging. As a strawman, I would argue that we should be less permissive with failed Tracking API calls and mark runs as `FAILED` when content cannot be logged successfully, as the absence of expected content may break user workflows or provide an incomplete record of training. I.e. `try_mlflow_log` seems to be used too liberally as-is.

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,133 @@
...
+    def create_run(
+        self,
+        experiment_id: str,
+        start_time: Optional[int] = None,
+        tags: Optional[Dict[str, Any]] = None,
+    ) -> Run:
+        """
+        Enqueues a CreateRun operation with the specified attributes.
+        """
+        return self._client.create_run(experiment_id, start_time, tags)
```
dbczumar, line 69: Needs to be converted to a queueable / flushable operation!

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def set_terminated(
+        self, run_id: str, status: Optional[str] = None, end_time: Optional[int] = None
+    ) -> None:
+        """
+        Enqueues an UpdateRun operation with the specified `status` and `end_time` attributes
+        for the specified `run_id`.
+        """
+        return self._client.set_terminated(run_id, status, end_time)
```
dbczumar, line 78: Needs to be converted to a queueable / flushable operation!

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def result(self):
+        for run_id, future in self._runs_to_futures_map.items():
+            future.result()
+
+
+class AutologgingBatchingClient:
+    """
+    Efficienty implements a subset of MLflow Tracking's  `MlflowClient` and fluent APIs to provide
+    automatic request batching, parallel execution, and parameter / tag truncation for autologging
```
dbczumar, line 46: Note that this sketch implementation doesn't provide comprehensive batching, parallel execution, or parameter truncation.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -701,56 +701,6 @@ def test_autolog_emits_warning_message_when_model_prediction_fails():
...
-            mock_log_metric.assert_called_once()
-            mock_set_tags.assert_called_once()
-            mock_log_model.assert_called_once()
-
-        query = "tags.{} = '{}'".format(MLFLOW_PARENT_RUN_ID, run.info.run_id)
-        assert len(mlflow.search_runs([run.info.experiment_id])) == 1
-        assert len(mlflow.search_runs([run.info.experiment_id], query)) == 0
-
-
-def test_meta_estimator_fit_performs_logging_only_once():
```
dbczumar, line 727: These test cases are now obviated by `test_flush_clears_pending_operations` from `tests/autologging/test_autologging_client.py`

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -895,43 +845,6 @@ def test_parameter_search_handles_large_volume_of_metric_outputs():
...
-        cv_model.fit(*get_iris())
-        mock_func.assert_called_once()
-
-
-@pytest.mark.usefixtures(test_mode_off.__name__)
-@pytest.mark.parametrize(
-    "func_to_fail",
-    ["mlflow.log_params", "mlflow.log_metric", "mlflow.set_tags", "mlflow.sklearn.log_model"],
-)
-def test_autolog_does_not_throw_when_mlflow_logging_fails(func_to_fail):
```
dbczumar, line 922: These tests have been unnecessary for some time, since the autologging patch safety framework is meant to handle all failures of this nature, with ample test coverage provided in https://github.com/mlflow/mlflow/blob/master/tests/autologging/test_autologging_safety_unit.py.

mlflow/sklearn/__init__.py
```diff
@@ -742,23 +750,24 @@ def _log_pretraining_metadata(estimator, *args, **kwargs):  # pylint: disable=un
         # process; accordingly, we avoid logging initial, untuned parameters
         # for these seed estimators.
         should_log_params_deeply = not _is_parameter_search_estimator(estimator)
-        # Chunk model parameters to avoid hitting the log_batch API limit
-        for chunk in _chunk_dict(
-            estimator.get_params(deep=should_log_params_deeply),
-            chunk_size=MAX_PARAMS_TAGS_PER_BATCH,
-        ):
-            truncated = _truncate_dict(chunk, MAX_ENTITY_KEY_LENGTH, MAX_PARAM_VAL_LENGTH)
-            try_mlflow_log(mlflow.log_params, truncated)
```
dbczumar, line 751: As part of integrating the new `MlflowAutologgingQueueingClient` into MLflow's sklearn autologging integration, I've opted to remove `try_mlflow_log` statements from MLflow Tracking API calls. `try_mlflow_log` was the predecessor to the autologging safety framework offered by `safe_patch`; `try_mlflow_log` was used to wrap any statements that were prone to failure (e.g. MLflow logging) to prevent exceptions from terminating model training. Now that `safe_patch` provides this safety, `try_mlflow_log` is used to wrap operations considered non-essential to the completeness / correctness of an MLflow run. However, parameter logging is clearly essential; an autologging run should not be marked as "finished successfully" (`FINISHED`) unless its parameters have been successfully logged. This also holds for metrics, artifacts, and tags. Longer term, we should significantly reduce / entirely remove instances of `try_mlflow_log` across all of our autologging integrations; this will aid users in identifying failures.

I've removed `try_mlflow_log` from sklearn as part of this PR because it has no equivalent in `MlflowAutologgingQueueingClient`, nor does it need an equivalent there.

mlflow/sklearn/__init__.py
```diff
@@ -33,10 +33,10 @@
 from mlflow.utils.autologging_utils import (
```
dbczumar, line 33: This portion of the change integrates the new `MlflowAutologgingQueueingClient` (defined in `mlflow/utils/autologging_utils/client.py`) into the `mlflow.sklearn` autologging integration.

mlflow/sklearn/utils.py
```diff
@@ -7,20 +7,9 @@
 import time
```
dbczumar, line 7: This portion of the change integrates the new `MlflowAutologgingQueueingClient` (defined in `mlflow/utils/autologging_utils/client.py`) into the `mlflow.sklearn` autologging integration.

mlflow/sklearn/utils.py
```diff
@@ -691,41 +677,17 @@ def first_custom_rank_column(df):
...
-            client.log_batch(
-                run_id=child_run.info.run_id,
-                params=[
-                    Param(str(key), str(value)) for key, value in truncated_params_batch.items()
-                ],
-                metrics=[
-                    Metric(key=str(key), value=value, timestamp=child_run_end_time, step=0)
-                    for key, value in truncated_metrics_batch.items()
-                ],
-            )
```
dbczumar, line 726: We can remove lots of complex batching logic from autologging integration implementations because `MlflowAutologgingQueueingClient` handles this in a unified location.

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,325 @@
...
+        except Exception as e:
+            return e
+
+    def _flush_pending_operations(self, pending_operations):
+        """
+        Synchronously and sequentially flushes the specified list of pending run operations.
+
+        NB: Operations are not parallelized on a per-run basis because MLflow's File Store, which
+        is frequently used for local ML development, does not support threadsafe metadata logging
+        within a given run.
```
dbczumar, line 215: This is fairly obnoxious. We could consider special-casing non-file-store tracking URIs. We've had discussions in the past about deprecating the file store, but we haven't yet taken that step.

For now, I'd opt for synchronous / sequential execution of logging within a run. This also reduces the likelihood of saturating / overloading the Databricks MLflow Tracking server (and other third-party Tracking implementations).

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,325 @@
...
+                message=(
+                    "The following failures occurred while performing one or more logging"
+                    " operations: {failures}".format(failures=failed_operations)
+                )
+            )
+
+
+class MlflowAutologgingQueueingClient:
+    """
+    Efficienty implements a subset of MLflow Tracking's  `MlflowClient` and fluent APIs to provide
```
harupy, line 64:
```suggestion
    Efficiently implements a subset of MLflow Tracking's  `MlflowClient` and fluent APIs to provide
```

nit

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,325 @@
...
+        self._pending_ops_by_run_id = {}
+
+        logging_operations = RunOperations(logging_futures)
+        if synchronous:
+            logging_operations.await_completion()
+        return logging_operations
+
+    def _get_pending_operations(self, run_id):
+        """
+        :return: A `_PendingRunOperations` containing all pending operations for the specified
```
harupy, line 192:
```suggestion
        :return: A `_PendingRunOperations` containing all pending operations for the specified `run_id`
```

Is `run_id` missing?

mlflow/utils/autologging_utils/client.py
jinzhang21: What is the reason to not simply improve the MlflowClient class to make it more efficient?

It seems the new autologging client has identical APIs but reimplemented some of them. Would it be better to inherit from MlflowClient class instead of creating something new?

mlflow/utils/autologging_utils/client.py
jinzhang21: This is probably fine for now. For file store we could potentially introduce chunked files to support concurrent write.

jinzhang21 (commented):
Overall looks good. I have not looked into the detailed implementation and tests yet. On a high level, my main question is whether the new client should be related to the standard MlflowClient. See my comment below.

harupy (approved):

mlflow/utils/autologging_utils/client.py
dbczumar: Good question! I agree that improvements to MlflowClient for more efficient batching / parallel execution are a great end goal; however, since `MlflowClient` is a user-facing API rather than a developer-facing API (like `MlflowAutologgingQueueingClient`), we'll need to put more thought into the API design / UX and user impact. For a developer API that's meant to be used for the purpose of building new autologging integrations, I think we can afford to take a best-effort approach and iterate if necessary, especially given that autologging integrations are native to MLflow and are therefore always reviewed / subject to CI.

As far as inheriting from `MlflowClient`, my view is that this would either: 1. require substantial effort to ensure that the entire `MlflowClient` interface supports batching / parallel execution or 2. create confusion / difficulty when using these APIs because callers (developers) must reason about which APIs behave sequentially / synchronously and which exhibit queueing behavior / parallel execution.

In the interest of short-term user impact and a reasonably well-documented / usable developer API, I've opted for a standalone `MlflowAutologgingQueueingClient` whose APIs operate with the same consistent queueing / parallel execution behavior. Let me know what your thoughts are here.

mlflow/utils/autologging_utils/client.py
dbczumar: It definitely was. I've fixed this. Thanks Haru!

dbczumar (commented):
@harupy @jinzhang21 Thank you for the reviews! I've addressed your comments - let me know if you can take another pass @jinzhang21 !

mlflow/utils/autologging_utils/client.py
jinzhang21: Thanks for the reasoning, Corey. I'm fine with current approach as long as we have the intention to merge the clients at some point in the future, when MlflowAutologgingQueueingClient becomes robust, and we have confidence in it. IMO this is a typical example of tech debt that need to be cleaned up at some point.

dbczumar: Absolutely! Great observation that we're taking on some debt for the benefit of more quickly providing performance improvements for autologging; fortunately, we're also prototyping improvements that are transferable to the `MlflowClient` in the process (as long as we undertake that work sooner than later).

I've added a module-level comment in `mlfflow/utils/autologging_utils/client.py` with a TODO to this effect.

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,370 @@
...
+        run_id: Union[str, PendingRunId],
+        metrics: Dict[str, float],
+        step: Optional[int] = None,
+    ) -> None:
+        """
+        Enqueues a collection of Metrics to be logged to the run specified by `run_id` at the
+        step specified by `step`.
+        """
+        metrics = _truncate_dict(metrics, max_key_length=MAX_ENTITY_KEY_LENGTH)
+        timestamp = int(time.time() * 1000)
```
jinzhang21, line 192: timestamp_ms to include the unit for better readability.

mlflow/utils/autologging_utils/client.py
```diff
@@ -0,0 +1,370 @@
...
+        param_batches_to_log = chunk_list(
+            pending_operations.params_queue, chunk_size=MAX_PARAMS_TAGS_PER_BATCH,
+        )
+        tag_batches_to_log = chunk_list(
+            pending_operations.tags_queue, chunk_size=MAX_PARAMS_TAGS_PER_BATCH,
+        )
+        for params_batch, tags_batch in zip_longest(
+            param_batches_to_log, tag_batches_to_log, fillvalue=[]
+        ):
+            metrics_batch_size = min(
```
jinzhang21, line 296: Could this number become negative? It seems not the intention to use it that way below.

jinzhang21 (approved):
Overall LGTM. Please address my comments below before submitting. Nice work!

mlflow/utils/autologging_utils/client.py
dbczumar: Changed to `timestamp_ms` - thanks for the suggestion!

mlflow/utils/autologging_utils/client.py
dbczumar: Good callout! I've added logic to ensure that the value is never negative.

PR #4431, opened 2021-06-07
[ML-15298] `mlflow.shap.log_explanation` creates extraneous run if autologging enabled

apurva-koti:

If autologging is enabled for a library, calling `mlflow.shap.log_explanation` on a function from that library will create an autologged run and print warning messages. This disables autologging in this use case as the user is not calling the function.

Test code run:
```python3
import pandas as pd
from sklearn.datasets import load_boston

import mlflow
mlflow.sklearn.autolog()

# prepare training data
dataset = load_boston()
X = pd.DataFrame(dataset.data[:50, :8], columns=dataset.feature_names[:8])
y = dataset.target[:50]

# train a model
model = LinearRegression()
model.fit(X, y)

# log an explanation
mlflow.shap.log_explanation(model.predict, X)
```

Results BEFORE this implementation:
image

Results AFTER this implementation (remaining run is from the `model.fit`):
image

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations


mlflow/shap.py
```diff
@@ -239,6 +240,12 @@ def predict_function(X) -> pred:
     import matplotlib.pyplot as plt
     import shap
 
+    original_config_dict = mlflow.utils.autologging_utils.AUTOLOGGING_INTEGRATIONS
+
+    new_config = original_config_dict.copy()
+    for integration in original_config_dict:
+        new_config[integration]["disable"] = True
+
```
apurva-koti, line 248: @mohamad-arabi I'm actually thinking it might be better to put this logic behind an `autologging.disable()` function so that we can use it like:
```
with autologging.disabled():
	...
```
because I suspect we'll need to do this in other places. Also it'd be nicer to keep the `AUTOLOGGING_INTEGRATIONS` abstracted away where it is.

mlflow/shap.py
```diff
@@ -253,6 +260,8 @@ def predict_function(X) -> pred:
     _log_matplotlib_figure(fig, _SUMMARY_BAR_PLOT_FILE_NAME, artifact_path)
     plt.close(fig)
 
+    mlflow.utils.autologging_utils.AUTOLOGGING_INTEGRATIONS = original_config_dict
```
mohamad-arabi, line 263: Try to iterate over all integrations `mlflow.integration.autolog(disable=True)`.

Then restore enabled ones.

mlflow/shap.py
```diff
@@ -239,6 +240,12 @@ def predict_function(X) -> pred:
     import matplotlib.pyplot as plt
     import shap
 
+    original_config_dict = mlflow.utils.autologging_utils.AUTOLOGGING_INTEGRATIONS
```
mohamad-arabi, line 243: Try to add  a util method in autologging_utils that would return only the list of actively enabled autologging integrations.

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -431,6 +434,19 @@ def autologging_is_disabled(integration_name):
     return False
 
 
+@contextlib.contextmanager
+def global_disable_autologging():
+    """
+    Disables autologging globally for all integrations.
+    This is useful in cases where the unpatched function is needed over the patched version.
+    Use in a with-block to disable autologging temporarily.
+    """
```
apurva-koti, line 443: RFC on name and docstring

tests/shap/test_log.py
```diff
@@ -106,6 +108,24 @@ def test_sklearn_log_explainer_pyfunc():
...
+    mlflow.sklearn.autolog(disable=False, exclusive=False)
+    dataset = sklearn.datasets.load_boston()
+    X = pd.DataFrame(dataset.data[:50, :8], columns=dataset.feature_names[:8])
+    y = dataset.target[:50]
+    model = sklearn.linear_model.LinearRegression()
+    model.fit(X, y)
+
+    with mlflow.start_run() as run:
+        mlflow.shap.log_explanation(model.predict, X)
+
```
mohamad-arabi, line 121: If not too difficult, to make sure that global disablement is contained:
train the model before(already doing that) and after call to shap. And confirm both training sessions have runs, and the shap call produced no runs.

apurva-koti: Solid idea, let's do it in utils test because now it's a utils thing, not a shap thing

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -43,6 +43,9 @@
 )
 _AUTOLOGGING_TEST_MODE_ENV_VAR = "MLFLOW_AUTOLOGGING_TESTING"
 
+# Flag indicating whether autologging is globally disabled for all integrations.
+AUTOLOGGING_GLOBALLY_DISABLED = False
```
dbczumar, line 47: Nit: Can we make this "internal" since it doesn't need to be referenced outside of the file in which it's defined?

```suggestion
_AUTOLOGGING_GLOBALLY_DISABLED = False
```

mlflow/utils/autologging_utils/__init__.py
dbczumar, line 443:
```suggestion
def disable_autologging():
    """
    Context manager that temporarily disables autologging globally for all integrations upon
    entry and restores the previous autologging configuration upon exit.
    """
```

mlflow/shap.py
```diff
@@ -240,8 +241,9 @@ def predict_function(X) -> pred:
     import shap
 
     artifact_path = _DEFAULT_ARTIFACT_PATH if artifact_path is None else artifact_path
-    background_data = shap.kmeans(features, min(_MAXIMUM_BACKGROUND_DATA_SIZE, len(features)))
-    explainer = shap.KernelExplainer(predict_function, background_data)
+    with mlflow.utils.autologging_utils.global_disable_autologging():
+        background_data = shap.kmeans(features, min(_MAXIMUM_BACKGROUND_DATA_SIZE, len(features)))
+        explainer = shap.KernelExplainer(predict_function, background_data)
     shap_values = explainer.shap_values(features)
```
dbczumar, line 247: Can we put lines 247 - 256 inside the context manager as well? Perhaps `explainer.shap_values()` or `shap.summary_plot()` may make use of sklearn in the future.

tests/shap/test_log.py
```diff
@@ -106,6 +108,24 @@ def test_sklearn_log_explainer_pyfunc():
...
+    run_data = MlflowClient().get_run(run.info.run_id).data
+    metrics, params, tags = run_data.metrics, run_data.params, run_data.tags
+    assert not metrics
```
dbczumar, line 124: can we also assert that the `mlflow.autologging` is absent?

mlflow/utils/autologging_utils/__init__.py
apurva-koti: I thought that as well, but it is primarily used in `safety.py`. Is it still ok to reference an internal variable like that since it (should be) the only usage?

dbczumar: I think that's okay :)

mlflow/shap.py
apurva-koti: Yup sounds good

tests/shap/test_log.py
apurva-koti: I think you mean the `mlflow.autologging` tag? sounds good

dbczumar (approved):
LGTM! Thanks @apurva-koti !

jinzhang21:
Nice change!

PR #4436, opened 2021-06-08
Update MLFlow max versions

sarthfrey:

As title suggests, using `python3 ./dev/update_ml_package_versions.py`

## How is this patch tested?

N/A

## Release Notes

N/A

No


harupy:
Thanks @sarthfrey for the PR, but we usually update `ml-package-versions` this when we make a new releaes usign our internal Jenkins job :)

PR #4439, opened 2021-06-09
[PaddlePaddle] support paddle model built by high level api

jinminhao:
Signed-off-by: Minhao Jin <minhaoj@andrew.cmu.edu>

Support paddle model trained with paddle high level api

## How is this patch tested?

(Details)
1. Add three unit tests and they all pass
2. Test writing a real example
3. run lint.sh to check code style

Components 

Interface 

Language 

Integrations


jinminhao:
Hi @harupy , 
I added support for paddle model trained with high level api. Please take a look at it. It should be supported since in PaddlePaddle, model.fit/model.evaluate (which will be in use when using autologging) is syntax that is only supported by paddle high level api.

tests/paddle/test_paddle_model_export.py
```diff
@@ -295,3 +295,129 @@ def test_model_log_without_specified_conda_env_uses_default_env_with_expected_de
...
+    test_dataset = pd_model_built_in_high_level_api.inference_dataframe
+    mlflow.paddle.save_model(pd_model=model, path=model_path)
+
+    reloaded_pd_model = mlflow.paddle.load_model(model_uri=model_path)
+    reloaded_pyfunc = pyfunc.load_pyfunc(model_uri=model_path)
+
+    low_level_test_dataset = []
+    for line in test_dataset:
+        data, _ = line
+        low_level_test_dataset.append(data)
```
harupy, line 341: Can this be simplified to:

```
low_level_test_dataset = [x[0] for x in test_dataset]
```

harupy:
@jinminhao Thanks for the update! I'm not familiar with paddlepaddle, when do we use the high level api?

jinminhao:
Hi @harupy, 
High level api in PaddlePaddle simplifies the code for model training and inference. The user can define a model by simply ```model = paddle.Model(...)```, then the training process is ```model.fit(...)```, and prediction is ```model.predict(...)```.  No matter users use paddle high or low level api, they can all use mlflow.paddle.load_model, save_model and log_model after this new feature is added.

 Moreover, this also helps in supporting mlflow autolog for paddlepaddle. Since ```autolog``` will detect class attributes like ```fit()``` or ```train()``` and is trigger when these attributes are called, only model that is implemented in paddle high level api can use ```autolog```.

tests/paddle/test_paddle_model_export.py
jinminhao: Yes! It's much better! Thank you for your suggestion

mlflow/paddle/__init__.py
```diff
@@ -196,7 +195,10 @@ def forward(self, inputs):
     model_data_subpath = "model"
     output_path = os.path.join(path, model_data_subpath)
 
-    paddle.jit.save(pd_model, output_path)
+    if pd_model.__class__.__name__ == "Model":
```
harupy, line 198:
```suggestion
    if ininstance(pd_model, paddle.Model):
```

Can we use `isinstance` here?

harupy:
Thanks for the explanation! so with the high level api, you can train your mode like skearn.

mlflow/paddle/__init__.py
jinminhao: Changed! Thx so much!

mlflow/paddle/__init__.py
```diff
@@ -196,7 +195,10 @@ def forward(self, inputs):
     model_data_subpath = "model"
     output_path = os.path.join(path, model_data_subpath)
 
-    paddle.jit.save(pd_model, output_path)
+    if isinstance(pd_model, paddle.Model):
+        pd_model.save(output_path, training=False)
```
harupy, line 199: Does `training=False` here mean saving a model in the evaluation mode?

jinminhao: Yes, it means that the model saved will be used for inference only and no longer support a second training process.

harupy: I see, is it possible to load a model saved with `training=False` and re-train it?

jinminhao: No. If we want to re-train it, we should set ```training=True```

harupy: Got it, would it make sense to make `training` configurable via an argument to allow re-training?

```diff
def save_model(
    ...
+   training=False
)
```

jinminhao: It makes sense. I will add a new example and unit tests  with the flag ```training=True```.

jinminhao: HI @harupy 
Now it supports model retrain. I also add a new example that is trained using high level paddle api. Please take a look at it. Thank you so much

jinminhao:
@harupy Now it supports model retrain. I also add a new example that is trained using high level paddle api. Please take a look at it. Thank you so much!

tests/paddle/test_paddle_model_export.py
```diff
@@ -295,3 +295,168 @@ def test_model_log_without_specified_conda_env_uses_default_env_with_expected_de
...
+    mlflow.paddle.save_model(pd_model=model, path=model_path, training=True)
+
+    class UCIHousing(paddle.nn.Layer):
+        def __init__(self):
+            super(UCIHousing, self).__init__()
+            self.fc_ = paddle.nn.Linear(13, 1, None)
+
+        def forward(self, inputs):  # pylint: disable=arguments-differ
+            pred = self.fc_(inputs)
+            return pred
```
harupy, line 437: Can we define this class in the __main__ scope?

jinminhao: Sure! The code is much cleaner now.

tests/paddle/test_paddle_model_export.py
```diff
@@ -54,7 +54,7 @@ def __init__(self):
             self.fc_ = Linear(in_features=13, out_features=1)
 
         @paddle.jit.to_static
-        def forward(self, inputs):  # pylint: disable=W0221
+        def forward(self, inputs):  # pylint: disable=arguments-differ
```
harupy, line 57: 👍

mlflow/paddle/__init__.py
```diff
@@ -254,7 +270,24 @@ def load_model(model_uri):
...
+                "The model doee not support re-train. \
+                Please set flag training=True when the paddle model is saved"
+            )
+        else:
+            model.load(
+                pd_model_artifacts_path,
+                skip_mismatch=skip_mismatch,
+                reset_optimizer=reset_optimizer,
+            )
+            return model
```
harupy, line 290:
```suggestion
        support_retrain = _check_if_model_supports_retrain(local_model_path)
        if not support_retrain:
            raise TypeError(
                "The model doee not support re-train. \
                Please set flag training=True when the paddle model is saved"
            )
        else:
            model.load(
                pd_model_artifacts_path,
                skip_mismatch=skip_mismatch,
                reset_optimizer=reset_optimizer,
            )
            return model
```

`model.load` works for a model logged with `training=False`, right?

mlflow/paddle/__init__.py
```diff
@@ -222,7 +228,7 @@ def forward(self, inputs):
     mlflow_model.save(os.path.join(path, MLMODEL_FILE_NAME))
 
 
-def load_model(model_uri):
+def load_model(model_uri, model=None, skip_mismatch=False, reset_optimizer=False):
```
harupy, line 231:
```suggestion
def load_model(model_uri, model=None, **kwargs):
```

Can we use `kwargs` here and pass it to `paddle.jit.load` or `model.load` in the function?

```python
if model is None:
    paddle.jit.load(..., **kwargs)
else:
    model.load(..., **kwrags)
```

mlflow/paddle/__init__.py
```diff
@@ -234,6 +240,16 @@ def load_model(model_uri):
             - ``models:/<model_name>/<model_version>``
             - ``models:/<model_name>/<stage>``
 
+    :param model: If it is not None, the type should be paddle.Model which is a model
+                  built with PaddlePaddle high level api and supports retraining
```
harupy, line 244:
```suggestion
    :param model: Required when loading a model saved using the high-level API. Must be an instance of `paddle.Model`.
```

mlflow/paddle/__init__.py
jinminhao: No, if we are using high level api and set ```training=False```. i.e ```model.save(..., training=False)```, then the model will be saved without any optimizer and we can just use low level api ```paddle.jit.load``` to load that model.

mlflow/paddle/__init__.py
jinminhao: Cool! It's much better and cleaner!

examples/paddle/train_high_level_api.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+model = paddle.Model(UCIHousing())
+optim = paddle.optimizer.Adam(learning_rate=0.01, parameters=model.parameters())
+model.prepare(optim, paddle.nn.MSELoss())
+
+model.fit(train_dataset, epochs=6, batch_size=8, verbose=1)
+
+sk_path_dir = "./test-out"
+mlflow.paddle.save_model(model, sk_path_dir)
```
harupy, line 26:
```suggestion
with mlflow.start_run():
    mlflow.paddle.log_model(model, "model")
```

Can we use `log_model`?

examples/paddle/train_high_level_api.py
```diff
@@ -0,0 +1,32 @@
...
+print("Model saved in run %s" % mlflow.active_run().info.run_uuid)
```
harupy, line 27:
```suggestion
with mlflow.start_run() as run:
    mlflow.paddle.log_model(model, "model")
    print("Model saved in run %s" % run.info.run_uuid)
```

Can we use `log_model`?

mlflow/paddle/__init__.py
harupy: Got it, thanks for the explanation :)

mlflow/paddle/__init__.py
```diff
@@ -234,6 +240,20 @@ def load_model(model_uri):
...
+    :param skip_mismatch: This only works when `model` is not None. Must be an instance
+                          of `paddle.Model`. Note that the instance of `paddle.Model`
+                          must be set `training=True` when saved. Default is False. If
+                          set true, `load_model` will ignore the parameters whose shape
+                          or name mismatch with those in model files that are saved
+                          previously.
+    :param reset_optimizer: This only works when `model` is not None. Must be an instance
+                            of `paddle.Model`. Note that the instance of `paddle.Model` must
+                            be set `training=True` when saved. Default is False. If set true,
+                            the `load_model` will ignore the optimizer provided.
```
harupy, line 255: Can we replace these arguments with `kwargs`?

mlflow/paddle/__init__.py
```diff
@@ -254,7 +274,22 @@ def load_model(model_uri):
     local_model_path = _download_artifact_from_uri(artifact_uri=model_uri)
     flavor_conf = _get_flavor_configuration(model_path=local_model_path, flavor_name=FLAVOR_NAME)
     pd_model_artifacts_path = os.path.join(local_model_path, flavor_conf["pickled_model"])
-    return paddle.jit.load(pd_model_artifacts_path)
+    if not model:
+        return paddle.jit.load(pd_model_artifacts_path)
```
harupy, line 278:
```suggestion
        return paddle.jit.load(pd_model_artifacts_path, **kwargs)
```

mlflow/paddle/__init__.py
jinminhao: For the docstring, do you think I need to list all the keywords for the ```kwargs``` that can be passed to ```paddle.jit.save``` and ```model.save```

mlflow/paddle/__init__.py
```diff
@@ -254,7 +274,22 @@ def load_model(model_uri):
...
+        if not support_retrain:
+            raise TypeError(
+                "The model doee not support re-train. \
+                Please set flag training=True when the paddle model is saved"
+            )
+        else:
+            model.load(
+                pd_model_artifacts_path, **kwargs,
+            )
+            return model
```
harupy, line 292:
```suggestion
        if not support_retrain:
            raise TypeError(
                "The model doee not support re-train. \
                Please set flag training=True when the paddle model is saved"
            )

        return model.load(
            pd_model_artifacts_path, **kwargs,
        )
```

nit

mlflow/paddle/__init__.py
harupy: No, it's not. You can just indicate that `kwargs` will be passed to `paddle.jit.save` or `model.load`.

jinminhao: I see. Thank you so much!

mlflow/paddle/__init__.py
```diff
@@ -47,6 +47,7 @@ def get_default_conda_env():
 def save_model(
     pd_model,
     path,
+    training=False,
```
harupy, line 50: Can we add this argument to `log_model` as well?

jinminhao: ohh yes. Sorry for carelessness

jinminhao: I will also add some new unit tests on ```log_model``` based on high level api.

jinminhao: Done

jinminhao: finished~

mlflow/paddle/__init__.py
```diff
@@ -234,6 +240,13 @@ def load_model(model_uri):
             - ``models:/<model_name>/<model_version>``
             - ``models:/<model_name>/<stage>``
 
+    :param model: Required when loading a `paddle.Model` model. Must be an instance of
+                  `paddle.Model`.
+                  NOTE: The instance of `paddle.Model` must be set `training=True` when
+                  saved.
```
harupy, line 246:
```suggestion
    :param model: Required when loading a `paddle.Model` model saved with `training=True`.
                  Must be an instance of the `paddle.Model` class.
```

mlflow/paddle/__init__.py
```diff
@@ -234,6 +240,13 @@ def load_model(model_uri):
...
+    :param kwargs: The keyword arguments passed to the PaddlePaddle model save method
+                   (i.e., `paddle.jit.save()```, ```model.save()```)
```
harupy, line 248:
```suggestion
    :param kwargs: The keyword arguments to pass to `paddle.jit.load` or `model.load`.
```

mlflow/paddle/__init__.py
```diff
@@ -254,12 +267,28 @@ def load_model(model_uri):
     local_model_path = _download_artifact_from_uri(artifact_uri=model_uri)
     flavor_conf = _get_flavor_configuration(model_path=local_model_path, flavor_name=FLAVOR_NAME)
     pd_model_artifacts_path = os.path.join(local_model_path, flavor_conf["pickled_model"])
-    return paddle.jit.load(pd_model_artifacts_path)
+    if not model:
+        return paddle.jit.load(pd_model_artifacts_path, **kwargs)
+    elif not isinstance(model, paddle.Model):
+        raise TypeError("Argument 'model' should be a paddle.Model, if it is not None")
```
harupy, line 273:
```suggestion
        raise TypeError("Invalid object type `{}` for `model`, must be `paddle.Model`".format(type(model)))
```

jinminhao: Got it!

mlflow/paddle/__init__.py
```diff
@@ -254,12 +265,30 @@ def load_model(model_uri):
...
+        raise TypeError(
+            "Invalid object type `{}` for `model`, must be `paddle.Model`".format(type(model))
+        )
+    else:
+        support_retrain = _check_if_model_supports_retrain(local_model_path)
+        if not support_retrain:
+            raise TypeError(
+                "The model doee not support re-train. \
+                Please set flag training=True when the paddle model is saved"
+            )
```
harupy, line 280: I'm trying to come up with a nice error message here.

jinminhao: Sorry for my pool English...

harupy:
```suggestion
            raise TypeError(
                "This model can't be loaded via `model.load` because a '.pdparams' file doesn't exist. "
                "Please leave `model` unspecified to load the model via `paddle.jit.save` "
                "or set `training` to True when saving a model."
            )
```

mlflow/paddle/__init__.py
```diff
@@ -234,6 +240,11 @@ def load_model(model_uri):
             - ``models:/<model_name>/<model_version>``
             - ``models:/<model_name>/<stage>``
 
+    :param model: Required when loading a `paddle.Model` model.
```
harupy, line 243:
```suggestion
    :param model: Required when loading a `paddle.Model` model saved with `traininig=True`.
```

I think you're missing "saved with `traininig=True`".

mlflow/paddle/__init__.py
jinminhao: How about "Saved model doesn't support re-train. Please set flag `training=True` when the `paddle.Model` model is saved i.e., `mlflow.paddle.train(..., training=True)`. Or set `model=None` for inference only."

harupy: > Sorry for my poor English...

Don't worry, I'm not good at Enlighs at all either and always struggling to write comments and docs.

harupy: The reason I avoided using `re-train` is we're not sure what the user does after loading the model.

jinminhao: I see. Yeah, it can make the users confusing.

harupy: btw if we try to load a model that doesn't have a `.pdparams` file, what error message do we get?

mlflow/paddle/__init__.py
```diff
@@ -393,3 +423,9 @@ def predict(self, data):
 
         predicted = self.pd_model(inp_data)
         return pd.DataFrame(predicted.numpy())
+
+
+def _check_if_model_supports_retrain(path):
+    print(path)
+    file_list = os.listdir(path)
+    return any(".pdparams" in file for file in file_list)
```
harupy, line 431:
```suggestion
def _check_if_model_supports_retrain(path):
    file_list = os.listdir(path)
    return any(".pdparams" in file for file in file_list)
```

Now I think we should rename this function and remove `print`.

jinminhao: How about ```_check_if_model_saved_in_pdparams_format```

harupy: What about `_contains_pdparams`? It matches what this function does.

jinminhao: Year, cool!

tests/paddle/test_paddle_model_export.py
```diff
@@ -295,3 +295,209 @@ def test_model_log_without_specified_conda_env_uses_default_env_with_expected_de
...
+    training_dataset, test_dataset = get_dataset_built_in_high_level_api()
+
+    model_retrain = paddle.Model(UCIHousing())
+    model_retrain = mlflow.paddle.load_model(model_uri=model_path, model=model_retrain)
+    optim = paddle.optimizer.Adam(learning_rate=0.015, parameters=model.parameters())
+    model_retrain.prepare(optim, paddle.nn.MSELoss())
+
+    model_retrain.fit(training_dataset, epochs=6, batch_size=8, verbose=1)
+
+    mlflow.paddle.save_model(pd_model=model_retrain, path=model_retrain_path, training=False)
```
harupy, line 440: Can we add an error check here?

```python
with pytest.raises(TypeError, match="This model can't be loaded"):
    mlflow.paddle.load_model(model_uri=model_path, model=model_retrain)
```

jinminhao: Sure!

harupy: Can we do the same thing for this error in `load_model`?:

```python
elif not isinstance(model, paddle.Model):
    raise TypeError(..)
```

mlflow/paddle/__init__.py
```diff
@@ -234,6 +240,11 @@ def load_model(model_uri):
             - ``models:/<model_name>/<model_version>``
             - ``models:/<model_name>/<stage>``
 
+    :param model: Required when loading a `paddle.Model` model.
+                  Must be an instance of `paddle.Model` saved with `training=True`.
```
harupy, line 244:
```suggestion
    :param model: Required when loading a `paddle.Model` model.
                  Must be an instance of `paddle.Model` saved with `training=True`.
```

The current error message sounds like we need to call `model.save` before `model.load`, but that's not true right?

jinminhao: Yes, then what about 
```
Required when loading a `paddle.Model` model.
                  Must be an instance of `paddle.Model` saved including files .pdopt and .pdparams. This can be achieved by set `training=True` when saved.
```

harupy: What about fixing the first sentence?

> Required when loading a `paddle.Model` model saved with `training=True`.

jinminhao: About checking if the model type supports re-train.
If the model is saved with `training=True`, the model will be saved in the two files, .pdopt and .pdparams. 
If set `training=False`, the model will be saved in three files, .pdiparams, .pdiparams.info and .pdmodel
When we are checking if the model supports re-train, can we just check if .pdparams exists? If .pdparams has issues or .pdopt doesn't exist, the paddle api will return error message.

jinminhao: > What about fixing the first sentence?
> 
> > Required when loading a `paddle.Model` model saved with `training=True`.

Year, cool

tests/paddle/test_paddle_model_export.py
jinminhao: Maybe it's better to use a low level paddle model here to check the type error message

jinminhao: Do you mean `str(0)`?

tests/paddle/test_paddle_model_export.py
```diff
@@ -295,3 +295,222 @@ def test_model_log_without_specified_conda_env_uses_default_env_with_expected_de
...
+
+    with pytest.raises(TypeError, match="This model can't be loaded"):
+        mlflow.paddle.load_model(model_uri=model_retrain_path, model=model_retrain)
+
+    error_model = str(0)
```
harupy, line 445: `str(0)` is fine but I'd simply use `0`.

jinminhao: Year! 0 is better. I found a potential bug because of this!

mlflow/paddle/__init__.py
```diff
@@ -62,6 +63,9 @@ def save_model(
 
     :param pd_model: paddle model to be saved.
     :param path: Local path where the model is to be saved.
+    :param training: This only works when model is trained using PaddlePaddle high level api.
+                     When flag `training` is set True, the model supports both re-training and
+                     inference. If set False, the model saved only supports inference.
```
harupy, line 68:
```suggestion
    :param training: Only valid when saving a model trained using the PaddlePaddle high level API.
                     If set to True, the saved model supports both re-training and
                     inference. If set to False, it only supports inference.
```

harupy: Can we add this doc to `log_model`?

jinminhao: Sure！ Added

harupy (approved):
@jinminhao Thanks for the hard work! LGTM!

jinminhao:
@harupy Thank you so much!

PR #4444, opened 2021-06-10
Make sure `import mlflow` doesn't import ML packages

harupy:

Make sure import mlflow doesn't import ML packages. Currently, `pyspark` is imported when running `import mlflow`.

## How is this patch tested?

Exisiting checks

Components 

Interface 

Language 

Integrations


mlflow/mleap.py
```diff
@@ -125,7 +125,7 @@ def save_model(
     spark_model,
     sample_input,
     path,
-    mlflow_model=Model(),
```
harupy, line 128: The `Model` class constructor internally uses a function that imports pyspark.

mlflow/mleap.py
```diff
@@ -125,7 +125,7 @@ def save_model(
...
+    mlflow_model=None,
```
harupy, line 128: The `Model` class constructor internally uses a function that imports pyspark.

dev/run-python-flavor-tests.sh
```diff
@@ -29,5 +29,6 @@ pytest tests/utils/test_model_utils.py --large
 pytest tests/tracking/fluent/test_fluent_autolog.py --large
 pytest tests/autologging --large
 find tests/spark_autologging/ml -name 'test*.py' | xargs -L 1 pytest --large
+python dev/test_mlflow_lazily_imports_pyspark.py
```
harupy, line 32: pytest executes import statements in test scripts when collecting testing functions. Some of our test scripts contain `import pyspark`. As a workaround, run `test_mlflow_lazily_imports_pyspark.py` in a separate process.

dev/test_mlflow_lazily_imports_pyspark.py
```diff
@@ -0,0 +1,5 @@
+import sys
+import mlflow  # pylint: disable=unused-import
+
+cached_packages = set(k.split(".")[0] for k in sys.modules.keys())
+assert "pyspark" not in cached_packages
```
harupy, line 5: We should do the samething for other frameworks, but the `python-small` check implicitly tests this by importing mlflow in an environment where no ML packages are imported.

harupy:
```
python -c '
import sys
import mlflow
cached_packages = set(k.split(".")[0] for k in sys.modules.keys())
assert "pyspark" not in cached_packages
'
```

mlflow/mleap.py
harupy:
```python

class Model(object):
    def __init__(
        self,
        artifact_path=None,
        run_id=None,
        utc_time_created=None,
        flavors=None,
        **kwargs
    ):
        ...
        databricks_runtime = get_databricks_runtime()  # <- this imports pyspark to read a spark conf
```

mlflow/models/signature.py
```diff
@@ -94,7 +96,7 @@ def __repr__(self) -> str:
 
 
 def infer_signature(
-    model_input: Any, model_output: MlflowInferableDataset = None
+    model_input: Any, model_output: "MlflowInferableDataset" = None
```
harupy, line 99: Used a [forward-reference](https://www.python.org/dev/peps/pep-0484/#forward-references) here to avoid `MlflowInferableDataset is not defined`

mlflow/models/signature.py
```diff
@@ -4,22 +4,24 @@
...
 import pandas as pd
 import numpy as np
 
 from mlflow.types.schema import Schema
 from mlflow.types.utils import _infer_schema
 
-try:
-    import pyspark.sql.dataframe
+# At runtime, we don't need  `pyspark.sql.dataframe`
+if TYPE_CHECKING:
```
WeichenXu123, line 16: Could you explain when "TYPE_CHECKING" is True ? When running type check ? When doing type checking will it import pyspark ?

WeichenXu123: I roughly search for this, seems when running type checking no code will be executed. 
> The type checker never executes your code: instead, it analyzes it. Type checkers are implemented in pretty much the same way compilers are implemented, minus the "generate bytecode/assembly/machine code" step.

But if it don't execute code, how does it enter the `if` branch ?

harupy: When `mypy` analyzes code, it does recognize modules under `if TYPE_CHECKING:`.

```python
from typing import TYPE_CHECKING


if TYPE_CHECKING:
    import foo
```

```
% mypy a.py
a.py:5: error: Cannot find implementation or library stub for module named 'foo'
a.py:5: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

dev/run-python-flavor-tests.sh
WeichenXu123: Could you add this comment into the script file ?

WeichenXu123 (approved):

dev/run-python-flavor-tests.sh
harupy: got it

PR #4467, opened 2021-06-17
Autologging: Set parent run tags using CreateRun API request, rather than separate batch logging request

dbczumar:

Set parent run tags using CreateRun API request, rather than separate batch logging request

## How is this patch tested?

- Existing unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/autologging_utils/safety.py
+++ b/mlflow/utils/autologging_utils/safety.py
@@ -191,9 +191,7 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
     """
 
     def create_managed_run():
-        managed_run = mlflow.start_run()
-        if tags:
-            try_mlflow_log(mlflow.set_tags, tags)
+        managed_run = mlflow.start_run(tags=tags)
         _logger.info(
             "Created MLflow autologging run with ID '%s', which will track hyperparameters,"
             " performance metrics, model artifacts, and lineage information for the"


mlflow/utils/autologging_utils/safety.py
```diff
@@ -191,9 +191,7 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
     """
 
     def create_managed_run():
-        managed_run = mlflow.start_run()
-        if tags:
-            try_mlflow_log(mlflow.set_tags, tags)
+        managed_run = mlflow.start_run(tags=tags)
```
harupy, line 194: I forgot why we needed to do this.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -191,9 +191,7 @@ def with_managed_run(autologging_integration, patch_function, tags=None):
     """
 
     def create_managed_run():
-        managed_run = mlflow.start_run()
-        if tags:
-            try_mlflow_log(mlflow.set_tags, tags)
```
harupy, line 196: I forgot why we need to do this.

harupy: One difference I rememeber is if we run create a run this way, the `managed_run` object doesn't contain `tags`.

harupy (approved):
LGTM!

mlflow/utils/autologging_utils/safety.py
dbczumar: I think we previously needed to do this because TensorFlow autologging checked for the `mlflow.autologging` tag on the active run to determine whether or not an autologging session was in progress, which is no longer necessary (removed here: https://github.com/mlflow/mlflow/commit/4e68f960d4520ade6b64a28c297816f622adc83e#diff-0ea13712d8aad26f69240bff7787b862ab88122c9c9e1936ac4af414efd58227L1063).

PR #4501, opened 2021-06-23
Added commandRunId request header

jimmyxu-db:

Added command run id as a request header created by DatabricksRequestHeaderProvider

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/request_header/databricks_request_header_provider.py
+++ b/mlflow/tracking/request_header/databricks_request_header_provider.py
@@ -25,7 +25,8 @@ class DatabricksRequestHeaderProvider(RequestHeaderProvider):
             request_headers["job_type"] = databricks_utils.get_job_type()
         if databricks_utils.is_in_cluster():
             request_headers["cluster_id"] = databricks_utils.get_cluster_id()
-        if databricks_utils.get_command_run_id():
-            request_headers["command_run_id"] = databricks_utils.get_command_run_id()
+        command_run_id = databricks_utils.get_command_run_id()
+        if command_run_id is not None:
+            request_headers["command_run_id"] = command_run_id
 
         return request_headers
--- a/mlflow/tracking/request_header/databricks_request_header_provider.py
+++ b/mlflow/tracking/request_header/databricks_request_header_provider.py
@@ -25,6 +25,7 @@ class DatabricksRequestHeaderProvider(RequestHeaderProvider):
             request_headers["job_type"] = databricks_utils.get_job_type()
         if databricks_utils.is_in_cluster():
             request_headers["cluster_id"] = databricks_utils.get_cluster_id()
-        request_headers["command_run_id"] = databricks_utils.get_command_run_id()
+        if databricks_utils.get_command_run_id():
+            request_headers["command_run_id"] = databricks_utils.get_command_run_id()
 
         return request_headers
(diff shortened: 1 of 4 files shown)


mlflow/utils/databricks_utils.py
```diff
@@ -173,6 +173,13 @@ def get_job_type():
         return _get_context_tag("jobTaskType")
 
 
+def get_command_run_id():
+    try:
```
smurching, line 177: I actually think we don't want to fall back to calling `_get_context_tag("commandRunId")`, as this will attempt to access a non-existent context tag in older runtimes. Instead, we should try-catch computing the command run ID (since it's only available in newer runtimes) and only set the `command_run_id` header if it's available/defined

Here's the backstory: the reason we have this pattern for other fields is that they're always present in older runtimes via tags on the command context (originally we'd just call e.g. `_get_context_tag("apiUrl")` to get the `apiUrl` field from the tags map in `get_webapp_url`), but we found that in particular security-enhanced runtimes, this led to errors as Python code didn't have permission to access the tags map on `CommandContext`. So in new runtimes we exposed allowlisted methods like ` apiUrl` on the `CommandContext` & in https://github.com/mlflow/mlflow/pull/3398 we added support for calling the new methods, falling back to the legacy approach of fetching values from the tags map for older runtimes.

mlflow/utils/databricks_utils.py
```diff
@@ -173,6 +173,14 @@ def get_job_type():
...
+        return _get_command_context().commandRunId().get()
+    except Exception:
+        # older runtimes may not have the commandRunId available
+        return "unknown_command_run_id"
```
smurching, line 181: Thanks for the change! It might be cleaner to just not pass a command run ID in this case. We could return `None`, and then in the caller check whether the command run ID is defined before including it in request headers. 

So maybe we could rename this method to `get_command_run_id_if_exists` and update the logic here to:
```suggestion
       # Older runtimes may not have the commandRunId available
        return None
```

And then in ``mlflow/tracking/request_header/databricks_request_header_provider.py``:

```
        command_run_id = databricks_utils.get_command_run_id_if_exists()
        if command_run_id is not None:
            request_headers["command_run_id"] = command_run_id
```

smurching (commented):
One last comment otherwise LGTM. Were you able to test this against new & old runtime versions? You should be able to build an MLflow wheel using `python setup.py bdist_wheel` as described [here](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst#building-a-distributable-artifact) (no need to run `npm run build` / build the JS bundle since we're not trying to run the UI) and then upload/install your wheel on a cluster as described [here](https://docs.databricks.com/libraries/workspace-libraries.html#upload-a-jar-python-egg-or-python-wheel). Then you can import/call `databricks_utils.get_command_run_id_if_exists()` and make sure the return value is as expected (we should as a follow-up think how to add tests that do the same)

mlflow/utils/databricks_utils.py
jimmyxu-db: i see that `get_cluster_id` can also return None, so I'm not gonna rename it to keep it simple

mlflow/tracking/request_header/databricks_request_header_provider.py
```diff
@@ -25,5 +25,7 @@ def request_headers(self):
             request_headers["job_type"] = databricks_utils.get_job_type()
         if databricks_utils.is_in_cluster():
             request_headers["cluster_id"] = databricks_utils.get_cluster_id()
+        if databricks_utils.get_command_run_id():
```
smurching, line 28: Tiny nit, could we avoid recomputing this? Maybe we can do something like:

```
command_id = databricks_utils.get_command_run_id()
if command_id is not None:
    request_headers["command_run_id"] = command_id
```

smurching (approved):
LGTM

PR #4518, opened 2021-07-01
Implement auto pip dependency inference for `mlflow.*.log_model`

harupy:

Implement auto pip dependency inference for `mlflow.*.log_model`.

This PR update the following flavors:

- keras
- tensorflow
- pytorch
- sklearn
- lightgbm
- catboost
- xgboost

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations


mlflow/utils/environment.py
```diff
@@ -82,3 +86,47 @@ def _get_additional_pip_dep(conda_env):
...
+        modules_names = json.loads(stdout.decode("utf-8"))
+        return _get_pip_requirements(modules_names)
+
+
+def _get_pip_requirements(module_names):
+    module_to_package = importlib_metadata.packages_distributions()
+    res = []
+    for module_name in module_names:
+        if module_name in module_to_package:
+            pkg = module_to_package[module_name][0]
```
harupy, line 128: Offline discussion on how to handle namespace packages:

```python
>>> importlib_metadata.packages_distributions()
{
    ...,
    # Namespace packages that belong to the namespace `google`
    "google": [
        "google-cloud-core",
        "protobuf",
        "google-api-core",
        "google-auth",
        "googleapis-common-protos",
        "google-resumable-media",
        "google-cloud-storage",
    ]
}
```

tests/utils/test_environment.py
```diff
@@ -50,3 +51,27 @@ def test_mlflow_conda_env_includes_pip_dependencies_and_pip_is_specified(pip_spe
...
     assert env["dependencies"].count("pip") == (2 if pip_specification == "pip" else 1)
+
+
+def _normalize_version(version):
+    """
+    Normalizes a package version (e.g. '1.5.0-dev0' -> '1.5.0.dev0')
+    """
+    from packaging.version import Version
+
+    return str(Version(version))
```
harupy, line 62: The `METADATA` file of `xgboost 1.5.0-dev.0` (note `-` before `dev`) looks like this:


```diff
  Metadata-Version: 2.1
  Name: xgboost
+ Version: 1.5.0.dev0
  Summary: XGBoost Python Package
  Home-page: https://github.com/dmlc/xgboost
  Maintainer: Hyunsu Cho
  Maintainer-email: chohyu01@cs.washington.edu
  ...
```

tests/utils/test_environment.py
```diff
@@ -50,3 +51,27 @@ def test_mlflow_conda_env_includes_pip_dependencies_and_pip_is_specified(pip_spe
...
+    import sklearn
+
+    model = xgb.XGBClassifier()
+
+    with mlflow.start_run():
+        mlflow.sklearn.log_model(model, artifact_path="model")
+        model_uri = mlflow.get_artifact_uri("model")
+
+    pip_reqs = _infer_pip_requirements(model_uri)
+    assert f"xgboost=={_normalize_version(xgb.__version__)}" in pip_reqs
```
harupy, line 76: The `METADATA` file of `xgboost 1.5.0-dev.0` (note `-` before `dev`) looks like this:


`importlib_metadata` reads a packge version from this file.

```python
>>> xgboost.__version__
1.5.0-dev.0
>>> importlib_metadata.version("xgboost")
1.5.0.dev0
```

mlflow/utils/environment.py
```diff
@@ -82,3 +89,61 @@ def _get_additional_pip_dep(conda_env):
...
+            code = code_template.format(mode_uri=model_uri)
+            f.write(code)
+
+        # Run the script in a separate process and parse the output
+        stdout = subprocess.check_output(["python", script_path])
+        modules_names = json.loads(stdout.decode("utf-8"))
+        return _get_pip_requirements(modules_names)
+
+
+def _safe_infer_pip_requirements(model_uri, fallback):
```
dbczumar, line 143: Nit: I think we should add the exception handling & fallback logic to `_infer_pip_requirements`, rather than defining a new method. Can we also make `_infer_pip_requirements` public? I.e. -> `infer_pip_requirements`? This will encourage users to leverage for additional use cases.

mlflow/utils/environment.py
```diff
@@ -82,3 +89,61 @@ def _get_additional_pip_dep(conda_env):
...
+
+def _get_pip_requirements(module_names):
+    module_to_packages = importlib_metadata.packages_distributions()
+    requirements = []
+    for module_name in module_names:
+        if module_name not in module_to_packages:
+            continue
+
+        for pkg in module_to_packages[module_name]:
+            req = "{}=={}".format(pkg, importlib_metadata.version(pkg))
```
dbczumar, line 102: We need to handle pyspark dev versions that aren't available on PyPI at some point as well by stripping away the `dev0` suffix, as is done in https://github.com/mlflow/mlflow/blob/bc86f1dfdb6444746470b7b12b6582b1390eb1aa/mlflow/spark.py#L83. 

Perhaps we should introduce a flag that controls whether or not known package dependency versions that we know are private should be swapped for publicly-available versions.

mlflow/utils/environment.py
```diff
@@ -82,3 +89,61 @@ def _get_additional_pip_dep(conda_env):
...
+import mlflow
+
+def get_loaded_modules():
+   loaded_modules = set()
+   for mod_name in sys.modules.keys():
+       if not mod_name.startswith("_"):
+           loaded_modules.add(mod_name.split(".")[0])
+   return loaded_modules
+
+model = mlflow.pyfunc.load_model("{mode_uri}")
```
dbczumar, line 125:
```suggestion
model = mlflow.pyfunc.load_model("{model_uri}")
```

dbczumar: This subprocess code may not have the correct tracking URI set, right? If that's the case, MLflow may not be able to download the model files from `model_uri`. Instead, I think we should call [mlflow.utils.artifact_utils._download_artifact_from_uri](https://github.com/mlflow/mlflow/blob/bc86f1dfdb6444746470b7b12b6582b1390eb1aa/mlflow/tracking/artifact_utils.py#L56) prior to creating the new model loading process.

mlflow/utils/environment.py
```diff
@@ -82,3 +89,61 @@ def _get_additional_pip_dep(conda_env):
 def _log_pip_requirements(conda_env, path, requirements_file="requirements.txt"):
     pip_deps = _get_additional_pip_dep(conda_env)
     _mlflow_additional_pip_env(pip_deps, path=os.path.join(path, requirements_file))
+
+
+def _get_pip_requirements(module_names):
+    module_to_packages = importlib_metadata.packages_distributions()
```
dbczumar, line 95: Does `packages_distributions` also give us transitive dependency versions? If not, can we extract that information using other APIs from `importlib_metadata`?

mlflow/utils/environment.py
harupy: Makes sense. I'll fix.

mlflow/utils/environment.py
harupy: Thanks for pointing this out. Agree on introducing such flag.

dbczumar:
> @dbczumar I'm wondering if we should introduce a boolean flag to control the enablement of this new feature (e.g. `infer_pip_requirements`) and set the default value to **`False`** to avoid affecting users who don't need this feature and are already satisfied with the existing `mlflow.*.log_model` behavior. We could set the default value to `True` once we're ready to do that.

@harupy If we trust our dependency inference procedure, I think it's acceptable to enable this by default. If users specify the now-deprecated `conda_env` argument, we should use the supplied conda environment instead of inferring dependencies. I think this is enough to achieve backwards compatibility. To evaluate our confidence in the dependency inference procedure, we should try it out with some simple and more complex models with the OSS MLflow server and other serving tools (e.g. SageMaker, Databricks Serving).

harupy:
@dbczumar Thanks for the feedback, +1 for doing more tests and make sure if our procedure really works and gain more confidence.

tests/xgboost/test_xgboost_model_export.py
```diff
@@ -324,3 +324,34 @@ def test_sagemaker_docker_model_scoring_with_default_conda_env(xgb_model, model_
...
+
+    inferred_pip_reqs = infer_pip_requirements(model_uri)
+    assert f"xgboost=={_canonicalize_version(xgb.__version__)}" in inferred_pip_reqs
+    assert f"scikit-learn=={_canonicalize_version(sklearn.__version__)}" in inferred_pip_reqs
+
+    # DEBUG
+    with capsys.disabled():
+        print("\n" + "+" * 30)
+        print("\n".join(inferred_pip_reqs))
+        print("+" * 30)
```
harupy, line 357: These lines print out the following in the `flavors` check **where lots of ML packages are installed**:

https://github.com/mlflow/mlflow/pull/4518/checks?check_run_id=2994199829#step:6:899

```diff
++++++++++++++++++++++++++++++
Bottleneck==1.3.2
PyYAML==5.4.1
Pygments==2.9.0
certifi==2021.5.30
chardet==4.0.0
click==7.1.2
cloudpickle==1.6.0
colorama==0.4.4
contextvars==2.4
databricks-cli==0.14.3
decorator==4.4.2
google-api-core==1.30.0
google-auth==1.32.1
google-cloud-core==1.7.1
google-cloud-storage==1.40.0
google-resumable-media==1.3.1
googleapis-common-protos==1.53.0
idna==2.10
immutables==0.15
importlib-metadata==4.6.1
ipython-genutils==0.2.0
ipython==7.16.1
jedi==0.18.0
joblib==1.0.1
matplotlib==3.3.4
mlflow-test-plugin==0.0.1
mlflow==1.18.1.dev0
numpy==1.19.5
packaging==21.0
pandas==1.1.5
parso==0.8.2
pexpect==4.8.0
pickleshare==0.7.5
prompt-toolkit==3.0.19
protobuf==3.17.3
py4j==0.10.9
pyspark==3.1.2
python-dateutil==2.8.1
pytz==2021.1
requests==2.25.1
ruamel.yaml.clib==0.2.6
ruamel.yaml==0.15.89
scikit-learn==0.24.2
scipy==1.5.4
setuptools==52.0.0.post20210125
six==1.15.0
traitlets==4.3.3
typing-extensions==3.7.4.3
urllib3==1.26.6
wcwidth==0.2.5
xgboost==1.4.2
zipp==3.5.0
++++++++++++++++++++++++++++++
```

- `importlib_medata`

harupy: One interesting behavior of `sys.modules` that I found is:

```diff
>>> import sys
>>> sys.modules.keys()
[
    "sys",
    "builtins",
    "_frozen_importlib",
    "_imp",
    "_thread",
    "_warnings",
    "_weakref",
    "zipimport",
    "_frozen_importlib_external",
    "_io",
    "marshal",
    "posix",
    "encodings",
    "codecs",
    "_codecs",
    "encodings.aliases",
    "encodings.utf_8",
    "_signal",
    "__main__",
    "encodings.latin_1",
    "io",
    "abc",
    "_abc",
    "site",
    "os",
    "stat",
    "_stat",
    "_collections_abc",
    "posixpath",
    "genericpath",
    "os.path",
    "_sitebuiltins",
    "_bootlocale",
    "_locale",
    "types",
    "importlib",
    "importlib._bootstrap",
    "importlib._bootstrap_external",
    "warnings",
    "importlib.util",
    "importlib.abc",
    "importlib.machinery",
    "contextlib",
    "collections",
    "operator",
    "_operator",
    "keyword",
    "heapq",
    "_heapq",
    "itertools",
    "reprlib",
    "_collections",
    "functools",
    "_functools",
+    "google",
+    "google.cloud",
+    "google.logging",
+    "mpl_toolkits",
    "ruamel",
+    "sphinxcontrib",
    "readline",
    "atexit",
    "rlcompleter",
]
```

harupy: Looks like some namespace packages are loaded during the REPL startup.

dbczumar (commented):
@harupy I tried inferring the pip requirements of an XGBoost model trained via the following script:

```
import mlflow

import xgboost as xgb
from sklearn import datasets

iris = datasets.load_iris()
X = iris.data
y = iris.target

params = {
    "objective": "multi:softprob",
    "num_class": 3,
    "eval_metric": "mlogloss",
    "seed": 42,
}
model = xgb.train(params, dtrain, evals=[(dtrain, "train")])

mlflow.xgboost.log_model(
    model,
    artifact_path="foobiz",
)
```

The inferred requirements were as follows:

```
['Bottleneck==1.3.2', 'PySocks==1.7.1', 'Pygments==2.7.2', 'brotlipy==0.7.0', 'certifi==2020.6.20', 'chardet==3.0.4', 'click==7.1.2', 'cloudpickle==1.6.0', 'colorama==0.4.3', 'databricks-cli==0.14.1', 'decorator==4.4.2', 'google-auth==1.26.1', 'idna==2.10', 'importlib-metadata==4.5.0', 'ipython==7.19.0', 'jedi==0.17.1', 'joblib==0.17.0', 'matplotlib==3.3.2', 'mkl-service==2.3.0', 'mlflow-skinny==1.14.0', 'mlflow==1.18.0', 'mypy-extensions==0.4.3', 'numpy==1.19.2', 'packaging==20.9', 'pandas==1.1.3', 'parso==0.7.0', 'pexpect==4.8.0', 'pickleshare==0.7.5', 'prompt-toolkit==3.0.8', 'protobuf==3.14.0', 'psutil==5.7.2', 'py4j==0.10.9', 'pyspark==3.1.1', 'python-dateutil==2.8.1', 'pytz==2020.1', 'requests==2.24.0', 'scikit-learn==0.24.1', 'scipy==1.5.2', 'setuptools==50.3.1.post20201107', 'six==1.15.0', 'sphinxcontrib-applehelp==1.0.2', 'sphinxcontrib-devhelp==1.0.2', 'sphinxcontrib-htmlhelp==1.0.3', 'sphinxcontrib-jsmath==1.0.1', 'sphinxcontrib-qthelp==1.0.3', 'sphinxcontrib-serializinghtml==1.1.4', 'sphinxcontrib-websupport==1.2.4', 'traitlets==5.0.5', 'typing-extensions==3.7.4.3', 'urllib3==1.25.11', 'wcwidth==0.2.5', 'xgboost==1.3.3', 'zipp==3.4.0', 'zope.event==4.5.0', 'zope.interface==5.1.2']
```

Unfortunately, serving the model with `mlflow models serve` produces the following error:

```
Pip subprocess error:
ERROR: Could not find a version that satisfies the requirement mkl-service==2.3.0 (from versions: none)
ERROR: No matching distribution found for mkl-service==2.3.0
```

Removing `mkl-service` produces an error with `setuptools`. There are a number of libraries that can't be installed due to missing versions. I suppose that, just because a particular library version has been installed on a system, we may not be able to install that particular version again from PyPI.

I played around with an approach that checks for uses of the `import` statement when reloading packages:

```
import functools
import importlib_metadata
import sys

model_path = sys.argv[1]
IMPORTED_MODULES = set()


def _get_pip_requirements(module_names):
    module_to_packages = importlib_metadata.packages_distributions()
    requirements = []
    for module_name in module_names:
        if module_name not in module_to_packages:
            continue

        for pkg in module_to_packages[module_name]:
            req = "{}=={}".format(pkg, importlib_metadata.version(pkg))
            if req in requirements:
                continue

            requirements.append(req)

    return sorted(requirements)


def _create_import_patch(python_builtin_import):
    """
    Wraps the built-in __import__ function to capture imported modules as usage logs.

    :param python_builtin_import: The built-in import function (`builtins.__import__`).
    :param allowlist: A set of string tuples that represents a module allowlist
                      (e.g. `{("pyspark",), ("pyspark", "ml"), ...}`).
    :param logger: An instance of a logging class that implements
                   the `com.databricks.backend.daemon.driver.PythonLogging` interface.
    :returns: The wrapped import function
    """
    @functools.wraps(python_builtin_import)
    def import_patch(name, globals=None, locals=None, fromlist=(), level=0):
        is_absolute_import = level == 0
        if not name.startswith("_") and is_absolute_import:
            IMPORTED_MODULES.add(name.split(".")[0])

        original_result = python_builtin_import(name, globals, locals, fromlist, level)
        return original_result

    return import_patch

import mlflow
original_load_pyfunc = mlflow.xgboost._load_pyfunc

# Patch the loader module's `_load_pyfunc` method to capture modules imported during the
# loading procedure by temporarily applying a patch to `builtins.__import__`
def patch_load_pyfunc(path):
    import builtins
    original_import = builtins.__import__
    builtins.__import__ = _create_import_patch(original_import)
    result = original_load_pyfunc(path)
    builtins.__import__ = original_import
    return result

mlflow.xgboost._load_pyfunc = patch_load_pyfunc

mlflow.pyfunc.load_model(model_path)

print(_get_pip_requirements(list(IMPORTED_MODULES)))
```

This XGBoost example yields the following requirements:

```
['joblib==0.17.0', 'mypy-extensions==0.4.3', 'numpy==1.19.2', 'pandas==1.1.3', 'psutil==5.7.2', 'scikit-learn==0.24.1', 'scipy==1.5.2', 'setuptools==50.3.1.post20201107', 'typing-extensions==3.7.4.3', 'xgboost==1.3.3']
```

`setuptools` is still here - perhaps we can denylist since it seems to be included by default in most Python distributions (see https://stackoverflow.com/a/64813092/11952869) - but the number of dependencies is significantly reduced. After manually removing `setuptools` and adding `mlflow`, serving works:

```
2021/07/08 23:41:34 INFO mlflow.pyfunc.backend: === Running command 'source /Users/corey.zumar/opt/anaconda3/bin/../etc/profile.d/conda.sh && conda activate mlflow-b6ecf73bc09f0108678380d09092e88bc31bbf3e 1>&2 && gunicorn --timeout=60 -b 127.0.0.1:5000 -w 1 ${GUNICORN_CMD_ARGS} -- mlflow.pyfunc.scoring_server.wsgi:app'
[2021-07-08 23:41:34 -0700] [85930] [INFO] Starting gunicorn 20.1.0
[2021-07-08 23:41:34 -0700] [85930] [INFO] Listening at: http://127.0.0.1:5000 (85930)
[2021-07-08 23:41:34 -0700] [85930] [INFO] Using worker: sync
[2021-07-08 23:41:34 -0700] [85932] [INFO] Booting worker with pid: 85932
```

How do you feel about this approach?

harupy:
@dbczumar Thanks! The proposed approach looks good to me, but one concern is this doesn't capture modules imported by mlflow.

> Unfortunately, serving the model with mlflow models serve produces the following error:

```
Pip subprocess error:
ERROR: Could not find a version that satisfies the requirement mkl-service==2.3.0 (from versions: none)
ERROR: No matching distribution found for mkl-service==2.3.0
```

Interesting.  This might be because `mkl-service 2.3.0` on **PyPI** doesn't provide a wheel for mac?
https://pypi.org/project/mkl-service/2.3.0/#files

harupy:
Code to remove transitive dependencies using https://github.com/naiquevin/pipdeptree:

```python
from itertools import chain
import pipdeptree

# We want to remove `numpy` and `joblib`
packages = ["scikit-learn", "xgboost", "numpy", "joblib"]
pkgs = pipdeptree.get_installed_distributions()
pkgs = [p for p in pkgs if p._key in packages]

tree = pipdeptree.PackageDAG.from_pkgs(pkgs)
branch_keys = set(r.key for r in chain.from_iterable(tree.values()))
nodes = [p for p in tree.keys() if p.key not in branch_keys]
top_level_packages = sorted(n.project_name for n in nodes)
print(top_level_packages)
# ['scikit-learn', 'xgboost']
```

harupy:
https://github.com/mlflow/mlflow/pull/4518/checks?check_run_id=3058978744#step:6:902

```
++++++++++++++++++++++++++++++
cloudpickle==1.6.0
pandas==1.1.5
pyspark==3.1.2
scikit-learn==0.24.2
setuptools==52.0.0.post20210125
typing-extensions==3.7.4.3
xgboost==1.4.2
++++++++++++++++++++++++++++++
```

Offline discussion: `typing-extensions` comes from `joblib`:

https://github.com/joblib/joblib/blob/754433f617793bc950be40cfaa265a32aed11d7d/joblib/externals/cloudpickle/cloudpickle.py#L62

but `typing-extensions` is not a hard dependency of `joblib` and setup.py doesn't contain it so `imporlib_metadata` or `pkg_resources` doesn't detect it.

harupy:
@dbczumar I think `pyspark` is imported in this line:

https://github.com/mlflow/mlflow/blob/28d8ca19e172c33304adbf521bd79cd35119e954/mlflow/models/model.py#L51

When we log/load an MLflow model, we use the `Model` class whose constructor attempts to import `pyspark` to obtain the DBR / MLR version.

harupy:
In python, it's pretty common to do `import if installed`.

tests/models/test_infer_pip_requirements.py
```diff
@@ -0,0 +1,169 @@
...
+    from transformers import BertModel, BertConfig
+
+    model = BertModel(BertConfig())
+    with mlflow.start_run():
+        mlflow.pytorch.log_model(model, artifact_path="model")
+        model_uri = mlflow.get_artifact_uri("model")
+
+    pip_reqs = mlflow.infer_pip_requirements(model_uri, mlflow.pytorch.FLAVOR_NAME)
+    assert _get_pinned_requirement("torch") in pip_reqs
+    assert _get_pinned_requirement("transformers") in pip_reqs
```
harupy, line 153: @dbczumar 

The inferred pip requirements of a pytorch-transformers model look like this:

https://github.com/mlflow/mlflow/pull/4518/checks?check_run_id=3153161565#step:6:882

```
tests/models/test_infer_pip_requirements.py::test_infer_pip_requirements_transformers_pytorch 
==============================
Keras-Preprocessing==1.1.2
Pillow==8.3.1
PyYAML==5.4.1
adal==1.2.7
bcrypt==3.2.0
boto3==1.18.6
google-cloud-storage==1.41.1
ipython==7.16.1
kubernetes==17.17.0
pandas==1.1.5
pyOpenSSL==20.0.1
scipy==1.5.4
tensorflow==2.4.2
torch==1.9.0
transformers==4.9.0
==============================
```

mlflow/utils/environment.py
```diff
@@ -188,6 +191,30 @@ def _is_iterable(x):
         )
 
 
+def infer_pip_requirements(model_uri, flavor, fallback=None):
```
harupy, line 194: Maybe this functions should be moved to `requirements_utils.py`

mlflow/xgboost.py
```diff
@@ -173,13 +173,17 @@ def save_model(
...
+                model_data_path, FLAVOR_NAME, fallback=default_requirements,
+            )
+        conda_env, pip_requirements, pip_constraints = _process_pip_requirements(
+            default_requirements, pip_requirements, extra_pip_requirements,
         )
-        if conda_env is None
-        else _process_conda_env(conda_env)
-    )
+    else:
+        conda_env, pip_requirements, pip_constraints = _process_conda_env(conda_env)
```
harupy, line 186: Fixed the code to obtain `conda_env`, `pip_requirements`, and `pip_constraints` to avoid making the inference when `pip_requirements` is specified. I'll apply this change to other flavors.

tests/models/test_infer_pip_requirements.py
```diff
@@ -0,0 +1,262 @@
...
+    with mlflow.start_run(), assert_did_not_fallback():
+        mlflow.sklearn.log_model(xgb_model, artifact_path="model")
+        model_uri = mlflow.get_artifact_uri("model")
+
+    reqs = _read_requirements(model_uri)
+    assert _get_pinned_requirement("xgboost") in reqs
+    assert _get_pinned_requirement("scikit-learn") in reqs
+
+    data = X.head(3)
+    resp = pyfunc_serve_and_score_model(model_uri, data, CONTENT_TYPE_JSON_SPLIT_ORIENTED)
```
dbczumar, line 117: Each of our model flavors have tests for model scoring, right? Can we incorporate this custom pipeline serving logic into those test modules instead?

tests/models/test_infer_pip_requirements.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def decorator(test_func):
+        should_skip = not all(map(_is_importable, module_names))
+        reason = f"This test requires {module_names}"
+        return pytest.mark.skipif(should_skip, reason=reason)(test_func)
+
+    return decorator
+
+
+# REMOVE THIS FIXTURE BEFORE MERGING THE PR
```
dbczumar, line 56: Reminder to remove this :)

setup.py
```diff
@@ -47,6 +47,9 @@ def package_files(directory):
     "pytz",
     "requests>=2.17.3",
     "packaging",
+    # 3.7.0 or newer is required to use `importlib_metadata.packages_distributions`:
```
dbczumar, line 50: Can we add an explanation about why we need to use `packages_distributions`?

mlflow/utils/requirements_utils.py
```diff
@@ -100,3 +112,109 @@ def _parse_requirements(requirements_file, is_constraint):
...
+    version = importlib_metadata.version(package)
+    # Strips a dev version suffix (e.g. '.dev0') for pyspark in Databricks.
+    if package == "pyspark" and is_in_databricks_runtime():
+        version = _strip_dev_version_suffix(version)
+    return version
+
+
+def _infer_requirements(model_uri, flavor):
+    """
+    Infers the pip requirements of the specified model.
```
dbczumar, line 194: Can we add a high-level explanation about how this method works so that callers know it's somewhat heavy-weight? I.e. this will create a subprocess and load the model in it to determine which libraries are imported

mlflow/utils/requirements_utils.py
```diff
@@ -100,3 +112,109 @@ def _parse_requirements(requirements_file, is_constraint):
...
+    reqs = package.requires()
+    if len(reqs) == 0:
+        return
+
+    for req in reqs:
+        yield req.name
+        yield from _get_requires_recursive(req.name)
+
+
+def _prune_packages(packages):
```
dbczumar, line 152: Can we write a unit test for this?

mlflow/utils/environment.py
dbczumar: I like this location, since it's a top-level environment utility function. `requirements_utils` are utilities that this method depends on.

mlflow/utils/environment.py
```diff
@@ -188,6 +191,30 @@ def _is_iterable(x):
...
+    """
+    Attempts to infer the pip requirements of the specified MLflow model.
+
+    :param model_uri: The URI of the MLflow model.
+    :param flavor: The flavor name of the MLflow model.
+    :param fallback: If provided, an unexpected error during the inference procedure is swallowed
+                     and the value of ``fallback`` will be returned.
```
dbczumar, line 201:
```suggestion
                     and the value of ``fallback`` is returned. Otherwise, the error is raised.
```

mlflow/utils/_capture_modules.py
```diff
@@ -0,0 +1,108 @@
...
+    output_file = args.output_file
+
+    flavor_module = getattr(mlflow, flavor)
+    wrapped_load_pyfunc = _wrap_load_pyfunc(flavor_module._load_pyfunc)
+    # Load the model and capture modules imported during the loading procedure.
+    if _is_mlflow_model(model_path):
+        flavor_module._load_pyfunc = wrapped_load_pyfunc
+        mlflow.pyfunc.load_model(model_path)
+    else:
+        wrapped_load_pyfunc(model_path)
```
dbczumar, line 100: Can we add some inline comments to explain the distinction between the two cases and why the logic for handling each is different?

dbczumar (commented):
@harupy Can we file an internal ticket for supporting PySpark MLlib dependency inference on Databricks?

mlflow/utils/requirements_utils.py
```diff
@@ -100,3 +112,109 @@ def _parse_requirements(requirements_file, is_constraint):
...
+                "--output-file",
+                output_file,
+            ]
+        )
+        with open(output_file) as f:
+            modules = f.read().splitlines()
+
+    packages = _flatten(map(_module_to_packages, modules))
+    packages = map(_canonicalize_package_name, packages)
+    packages = set(packages) - set(["setuptools", "pkg_resources"])
```
dbczumar, line 218: Can we add a comment explaining why we remove these?

dbczumar (approved):
@harupy Implementation looks great! I tried it out with XGBoost + sklearn and Keras + sklearn models on Databricks, works wonderfully!

We should add API docs and support for remaining flavors. A follow-up PR seems fine for that. Great work!

mlflow/utils/requirements_utils.py
harupy: Sure!
Done!

mlflow/utils/environment.py
harupy: Doen!

mlflow/utils/environment.py
harupy: Agree!

mlflow/utils/_capture_modules.py
harupy: Sure!

setup.py
harupy: Added the explanation!

mlflow/utils/requirements_utils.py
harupy: Got it!

harupy: Clarified how this function infers pip requirements!

mlflow/utils/requirements_utils.py
harupy: Sure!

tests/models/test_infer_pip_requirements.py
harupy: Thanks!

tests/models/test_infer_pip_requirements.py
harupy: Got it!

harupy: This is surprising but most of the flavors are missing serving tests.

harupy:
> @harupy Can we file an internal ticket for supporting PySpark MLlib dependency inference on Databricks?

@dbczumar Filed a ticket for this.

setup.py
```diff
@@ -47,6 +47,11 @@ def package_files(directory):
     "pytz",
     "requests>=2.17.3",
     "packaging",
+    # `mlflow.infer_pip_requirements` relies on `importlib_metadata.packages_distributions`
+    # to resolve a module name to its package name (e.g. 'sklearn' -> 'scikit-learn').
```
dbczumar, line 51:
```suggestion
    # Automated dependency detection in MLflow Models relies on `importlib_metadata.packages_distributions`
    # to resolve a module name to its package name (e.g. 'sklearn' -> 'scikit-learn').
```

harupy: 👍

mlflow/catboost.py
```diff
@@ -146,13 +146,29 @@ def save_model(
...
+        if pip_requirements is None:
+            # HACK: Temporarily create an MLmodel file because `mlflow.catboost._load_pyfunc`
+            # requires it to get model type and save format.
+            save_path = os.path.join(path, MLMODEL_FILE_NAME)
+            Model().add_flavor(FLAVOR_NAME, **flavor_conf).save(save_path)
+            inferred_reqs = mlflow.infer_pip_requirements(
+                model_data_path, FLAVOR_NAME, fallback=default_reqs,
+            )
+            os.remove(save_path)  # Clean up the MLmodel file
+            default_reqs = list(set(inferred_reqs).union(default_reqs))
```
harupy, line 165: This is hacky, but `load_pyfunc_` in several flavors read parameters required for the model deserialization from an `MLmodel` file.

mlflow/utils/environment.py
```diff
@@ -189,6 +192,31 @@ def _is_iterable(x):
         )
 
 
+_INFER_PIP_REQUIREMENTS_FALLBACK_MESSAGE = (
+    "Encountered an unexpected error while inferring pip requirements (model URI: %s, flavor: %s)"
+)
+
+
+def infer_pip_requirements(model_uri, flavor, fallback=None):
```
harupy, line 200: `infer_pip_requirements` significantly increase the `flavors` test execution time (by 10 ~ 15 min).

dbczumar: I don't think we should do anything non-obvious to change the behavior under the hood. Can we remove `sklearn`, `gluon`, etc from our model flavor tests as part of this work to keep CI time low? This will add some time to cross version tests, but I think that's okay.

harupy: Agree!

dev/run-python-flavor-tests.sh
```diff
@@ -8,16 +8,11 @@ export MLFLOW_HOME=$(pwd)
...
-pytest tests/pytorch --large
 pytest tests/h2o --large
 pytest tests/onnx --large
 pytest tests/pyfunc --large
-pytest tests/sklearn --large
 pytest tests/azureml --large
 pytest tests/models --large
-pytest tests/xgboost --large
-pytest tests/lightgbm --large
-pytest tests/catboost --large
```
harupy, line 20: @dbczumar Removed the flavor tests that

mlflow/catboost.py
harupy: I think we can clean up `dev/extract-ml-requirements.txt` in a follow-up PR.

mlflow/ml-package-versions.yml
```diff
@@ -101,8 +101,8 @@ keras:
     maximum: "2.4.3"
     requirements:
       "< 2.3.1": ["tensorflow==1.15.4", "h5py<3.0", "scikit-learn", "pyspark", "pyarrow"]
-      "== 2.3.1": ["tensorflow==2.2.1", "scikit-learn", "pyspark", "pyarrow"]
-      "> 2.3.1": ["tensorflow<2.5.0", "scikit-learn", "pyspark", "pyarrow"]
+      "== 2.3.1": ["tensorflow==2.2.1", "scikit-learn", "pyspark", "pyarrow", "transformers"]
+      "> 2.3.1": ["tensorflow<2.5.0", "scikit-learn", "pyspark", "pyarrow", "transformers"]
```
harupy, line 105: `transformers` doesn't seem to work with keras < 2.3.

dev/test-anaconda-compatibility-in-docker.sh
```diff
@@ -10,11 +10,13 @@ set -euo pipefail
 
 . ~/.bashrc
 
-pip freeze > /tmp/before.txt
+IGNORE_PATTERN="^\(importlib-metadata\|zipp\)"
```
harupy, line 13: Added this line avoid a failure caused by adding `importlib_metadata >= 3.7.0` as a dependency.

dev/test-anaconda-compatibility-in-docker.sh
```diff
@@ -10,11 +10,13 @@ set -euo pipefail
...
+
+pip freeze | grep -v "$IGNORE_PATTERN" > /tmp/before.txt
 pip install --upgrade-strategy only-if-needed -e /mnt/mlflow
-pip freeze > /tmp/after.txt
-diff /tmp/before.txt /tmp/after.txt > /tmp/diff.txt || true
-if [[ ! -z $(grep "<" /tmp/diff.txt) ]]; then
+pip freeze | grep -v "$IGNORE_PATTERN" > /tmp/after.txt
+diff --color=always /tmp/before.txt /tmp/after.txt > /tmp/diff.txt || true
+if [[ ! -z $(cat /tmp/diff.txt | grep "<") ]]; then
```
harupy, line 19: Added `--color=always` to make it easier to identify what's changed.

harupy:
@dbczumar Could you take a final look?

btw thanks for reviewing those large pull requests in the past few weeks!

mlflow/catboost.py
```diff
@@ -146,13 +146,29 @@ def save_model(
...
+        if pip_requirements is None:
+            # HACK: Temporarily create an MLmodel file because `mlflow.catboost._load_pyfunc`
+            # requires it to get model type and save format.
+            save_path = os.path.join(path, MLMODEL_FILE_NAME)
+            Model().add_flavor(FLAVOR_NAME, **flavor_conf).save(save_path)
+            inferred_reqs = mlflow.infer_pip_requirements(
+                model_data_path, FLAVOR_NAME, fallback=default_reqs,
+            )
+            os.remove(save_path)  # Clean up the MLmodel file
+            default_reqs = sorted(set(inferred_reqs).union(default_reqs))
```
harupy, line 165: This line unions the inferred requirements and the default requirements to ensure the default requirements are always included.

dev/test-anaconda-compatibility-in-docker.sh
dbczumar: What's the failure? Is this going to cause disruptions for real users / use cases?

mlflow/__init__.py
```diff
@@ -188,5 +189,6 @@
     "get_registry_uri",
     "set_registry_uri",
     "list_run_infos",
+    "infer_pip_requirements",
```
dbczumar, line 192: I don't think we want to make this part of our fluent API. Can we import this API in `mlflow.models` for documentation / visibility purposes instead?

mlflow/catboost.py
```diff
@@ -146,13 +146,29 @@ def save_model(
...
+        default_reqs = get_default_pip_requirements()
+        if pip_requirements is None:
+            # HACK: Temporarily create an MLmodel file because `mlflow.catboost._load_pyfunc`
+            # requires it to get model type and save format.
+            save_path = os.path.join(path, MLMODEL_FILE_NAME)
+            Model().add_flavor(FLAVOR_NAME, **flavor_conf).save(save_path)
+            inferred_reqs = mlflow.infer_pip_requirements(
+                model_data_path, FLAVOR_NAME, fallback=default_reqs,
+            )
+            os.remove(save_path)  # Clean up the MLmodel file
```
dbczumar, line 164: Instead of writing the MLmodel file and then removing it, can we save `mlflow_model` to a file before this `if` block (right below the definition of `flavor_conf`) and then reload it after? Can we move the inline comment as well?

Can we also clarify in the comment that `mlflow.catboost._load_pyfunc` is called for dependency inference purposes? Otherwise, it's unclear why we care about `_load_pyfunc`.

Finally, can we avoid describing this as a hack? It's part of the model persistence procedure :)

mlflow/catboost.py
```diff
@@ -168,14 +184,8 @@ def save_model(
...
-
-    flavor_conf = {
-        _MODEL_TYPE_KEY: cb_model.__class__.__name__,
-        _SAVE_FORMAT_KEY: kwargs.get("format", "cbm"),
-        **model_bin_kwargs,
-    }
     mlflow_model.add_flavor(
-        FLAVOR_NAME, catboost_version=cb.__version__, **flavor_conf,
+        FLAVOR_NAME, catboost_version=cb.__version__, **flavor_conf, **model_bin_kwargs,
     )
```
dbczumar, line 189: In accordance with https://github.com/mlflow/mlflow/pull/4518/files#r683751938, can we move this up to the part of the code right before we perform dependency resolution?

tests/keras/test_keras_model_export.py
```diff
@@ -651,3 +641,31 @@ def test_load_without_save_format(tf_keras_model, model_path):
 
     model_loaded = mlflow.keras.load_model(model_path)
     assert tf_keras_model.to_json() == model_loaded.to_json()
+
+
+@pytest.mark.large
+@pytest.mark.skipif(not _is_importable("transformers"), reason="This test requires transformers")
```
dbczumar, line 647:
```suggestion
@pytest.mark.skipif(not _is_importable("transformers"), reason="This test requires transformers, which is incompatible with Keras <= 2.3.0")
```

mlflow/tensorflow.py
```diff
@@ -293,15 +293,27 @@ def save_model(
...
+        meta_graph_tags=tf_meta_graph_tags,
+        signature_def_key=tf_signature_def_key,
     )
+    if conda_env is None:
+        default_reqs = get_default_pip_requirements()
+        if pip_requirements is None:
+            save_path = os.path.join(path, MLMODEL_FILE_NAME)
+            Model().add_flavor(FLAVOR_NAME, **flavor_conf).save(save_path)
+            inferred_reqs = mlflow.infer_pip_requirements(path, FLAVOR_NAME, fallback=default_reqs,)
+            os.remove(save_path)
```
dbczumar, line 310: Can we make the same adjustment recommended here? https://github.com/mlflow/mlflow/pull/4518/files#r683751938. Can we also add an inline comment?

dev/run-python-tf-tests.sh
```diff
@@ -6,17 +6,9 @@ err=0
...
-pytest tests/tensorflow_autolog/test_tensorflow2_autolog.py --large
-pytest tests/keras --large
-pytest tests/keras_autolog --large
-# Downgrade TensorFlow and Keras in order to test compatibility with older versions
-pip install 'tensorflow==1.15.4'
-pip install 'keras==2.2.5'
-pytest tests/keras --large
-pytest tests/tensorflow/test_tensorflow_model_export.py --large
-pytest tests/tensorflow_autolog/test_tensorflow_autolog.py --large
-pytest tests/keras_autolog --large
```
dbczumar, line 19: Nice! :D

dbczumar (changes requested):
@harupy I left a couple small comments - should be good to go once those are addressed!

mlflow/__init__.py
harupy: Sure!

mlflow/catboost.py
harupy: Got it. I misunderstood we need to save `conda_env.yml` and `requirements.txt` before `mlflow_model.save`, but we can save them after `mlflow_model.save`.

dbczumar (approved):
LGTM! Thanks @harupy !

PR #4519, opened 2021-07-01
[ML-16200] Add `pip_requirements` and `additional_pip_requirements` to `log_model`

harupy:

Add `pip_requirements` and `additional_pip_requirements` to `log_model`.

## How is this patch tested?

Unit tests

Add `pip_requirements` and `additional_pip_requirements` to `mlflow.*.log_model` and `mflow.*.save_model` to allows users to directly specify pip requirements.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/models/model.py
+++ b/mlflow/models/model.py
@@ -143,6 +143,8 @@ class Model(object):
         flavor,
         registered_model_name=None,
         await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
+        pip_requirements=None,
+        additional_pip_requirements=None,
         **kwargs
     ):
         """
@@ -184,7 +186,13 @@ class Model(object):
             local_path = tmp.path("model")
             run_id = mlflow.tracking.fluent._get_or_start_run().info.run_id
             mlflow_model = cls(artifact_path=artifact_path, run_id=run_id)
-            flavor.save_model(path=local_path, mlflow_model=mlflow_model, **kwargs)
+            flavor.save_model(
+                path=local_path,
+                mlflow_model=mlflow_model,
+                pip_requirements=pip_requirements,
+                additional_pip_requirements=additional_pip_requirements,
+                **kwargs
+            )
             mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
             try:
                 mlflow.tracking.fluent._record_logged_model(mlflow_model)
--- a/mlflow/models/model.py
+++ b/mlflow/models/model.py
@@ -144,7 +144,7 @@ class Model(object):
         registered_model_name=None,
         await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
         pip_requirements=None,
-        additional_pip_requirements=None,
+        extra_pip_requirements=None,
         **kwargs
     ):
         """
@@ -190,7 +190,7 @@ class Model(object):
                 path=local_path,
                 mlflow_model=mlflow_model,
                 pip_requirements=pip_requirements,
-                additional_pip_requirements=additional_pip_requirements,
+                extra_pip_requirements=extra_pip_requirements,
                 **kwargs
             )
             mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
--- a/mlflow/models/model.py
+++ b/mlflow/models/model.py
@@ -143,8 +143,6 @@ class Model(object):
         flavor,
         registered_model_name=None,
         await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
-        pip_requirements=None,
-        extra_pip_requirements=None,
         **kwargs
     ):
         """
@@ -186,13 +184,7 @@ class Model(object):
             local_path = tmp.path("model")
             run_id = mlflow.tracking.fluent._get_or_start_run().info.run_id
             mlflow_model = cls(artifact_path=artifact_path, run_id=run_id)
-            flavor.save_model(
-                path=local_path,
-                mlflow_model=mlflow_model,
-                pip_requirements=pip_requirements,
-                extra_pip_requirements=extra_pip_requirements,
-                **kwargs
-            )
+            flavor.save_model(path=local_path, mlflow_model=mlflow_model, **kwargs)
             mlflow.tracking.fluent.log_artifacts(local_path, artifact_path)
             try:
                 mlflow.tracking.fluent._record_logged_model(mlflow_model)
--- a/mlflow/utils/environment.py
+++ b/mlflow/utils/environment.py
@@ -1,7 +1,9 @@
 import yaml
+import tempfile
 import os
 
 from mlflow.utils import PYTHON_VERSION
+from mlflow.utils.requirements_utils import _parse_requirements
 
 _conda_header = """\
 name: mlflow-env
@@ -68,9 +70,9 @@ def _mlflow_additional_pip_env(
         return requirements
 
 
-def _get_additional_pip_dep(conda_env):
+def _get_pip_deps(conda_env):
     """
-    :return: The additional pip dependencies from the conda env
+    :return: The pip dependencies from the conda env
     """
     if conda_env is not None:
         for dep in conda_env["dependencies"]:
@@ -80,5 +82,75 @@ def _get_additional_pip_dep(conda_env):
 
 
 def _log_pip_requirements(conda_env, path, requirements_file="requirements.txt"):
-    pip_deps = _get_additional_pip_dep(conda_env)
+    pip_deps = _get_pip_deps(conda_env)
     _mlflow_additional_pip_env(pip_deps, path=os.path.join(path, requirements_file))
+
+
+def _parse_pip_requirements(pip_requirements):
+    """
+    Parses an iterable of pip requirement strings or a pip requirements file.
+
+    :param pip_requirements: Either an iterable of pip requirement strings
+        (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
+        file on the local filesystem (e.g. ``"requirements.txt"``). If ``None``, an empty list will
+        be returned.
+    :return: A list of pip requirement strings.
+    """
+    if pip_requirements is None:
+        return []
+
+    def _is_string(x):
+        return isinstance(x, str)
+
+    def _is_iterable(x):
+        try:
+            iter(x)
+            return True
+        except Exception:
+            return False
+
+    if _is_string(pip_requirements):
+        return list(_parse_requirements(pip_requirements))
+    elif _is_iterable(pip_requirements) and all(map(_is_string, pip_requirements)):
+        try:
+            # Create a temporary requirements file in the current working directory
+            tmp_req_file = tempfile.NamedTemporaryFile(
+                mode="w",
+                prefix="mlflow.",
+                suffix=".tmp.requirements.txt",
+                dir=os.getcwd(),
+                # Setting `delete` to True causes a permission-denied error on Windows
+                # while trying to read the generated temporary file.
+                delete=False,
+            )
+            tmp_req_file.write("\n".join(pip_requirements))
+            tmp_req_file.close()
+            return _parse_pip_requirements(tmp_req_file.name)
+        finally:
+            # Clean up the temporary requirements file
+            os.remove(tmp_req_file.name)
+    else:
+        raise TypeError(
+            "`pip_requirements` must be either a string path to a pip requirements file on the "
+            "local filesystem or an iterable of pip requirement strings, but got `{}`".format(
+                type(pip_requirements)
+            )
+        )
+
+
+def _validate_env_arguments(conda_env, pip_requirements, additional_pip_requirements):
+    """
+    Validates that only one or none of `conda_env`, `pip_requirements`, and
+    `additional_pip_requirements` is specified.
+    """
+    args = [
+        conda_env,
+        pip_requirements,
+        additional_pip_requirements,
+    ]
+    specified = [arg for arg in args if arg is not None]
+    if len(specified) > 1:
+        raise ValueError(
+            "Only one of `conda_env`, `pip_requirements`, and "
+            "`additional_pip_requirements` can be specified"
+        )
--- a/mlflow/utils/environment.py
+++ b/mlflow/utils/environment.py
@@ -138,19 +138,19 @@ def _parse_pip_requirements(pip_requirements):
         )
 
 
-def _validate_env_arguments(conda_env, pip_requirements, additional_pip_requirements):
+def _validate_env_arguments(conda_env, pip_requirements, extra_pip_requirements):
     """
     Validates that only one or none of `conda_env`, `pip_requirements`, and
-    `additional_pip_requirements` is specified.
+    `extra_pip_requirements` is specified.
     """
     args = [
         conda_env,
         pip_requirements,
-        additional_pip_requirements,
+        extra_pip_requirements,
     ]
     specified = [arg for arg in args if arg is not None]
     if len(specified) > 1:
         raise ValueError(
             "Only one of `conda_env`, `pip_requirements`, and "
-            "`additional_pip_requirements` can be specified"
+            "`extra_pip_requirements` can be specified"
         )
--- a/mlflow/utils/requirements_utils.py
+++ b/mlflow/utils/requirements_utils.py
@@ -1,0 +1,86 @@
+"""
+This module provides a set of utilities for interpreting and creating requirements files
+(e.g. pip's `requirements.txt`), which is useful for managing ML software environments.
+"""
+import os
+from itertools import filterfalse
+
+
+def _is_comment(line):
+    return line.startswith("#")
+
+
+def _is_empty(line):
+    return line == ""
+
+
+def _strip_inline_comment(line):
+    return line[: line.find(" #")].rstrip() if " #" in line else line
+
+
+def _is_requirements_file(line):
+    return line.startswith("-r ") or line.startswith("--requirement ")
+
+
+def _join_continued_lines(lines):
+    """
+    Joins lines ending with '\\'.
+
+    >>> _join_continued_lines["a\\", "b\\", "c"]
+    >>> 'abc'
+    """
+    continued_lines = []
+
+    for line in lines:
+        if line.endswith("\\"):
+            continued_lines.append(line.rstrip("\\"))
+        else:
+            continued_lines.append(line)
+            yield "".join(continued_lines)
+            continued_lines.clear()
+
+    # The last line ends with '\'
+    if continued_lines:
+        yield "".join(continued_lines)
+
+
+# TODO: Add support for constraint files:
+#       https://github.com/mlflow/mlflow/pull/4519#discussion_r668412179
+def _parse_requirements(requirements_file):
+    """
+    A simplified version of `pip._internal.req.parse_requirements` which performs the following
+    operations on the given requirements file and yields the parsed requirements.
+
+    - Remove comments and blank lines
+    - Join continued lines
+    - Resolve requirements file references (e.g. '-r requirements.txt')
+
+    References:
+    - `pip._internal.req.parse_requirements`:
+      https://github.com/pypa/pip/blob/7a77484a492c8f1e1f5ef24eaf71a43df9ea47eb/src/pip/_internal/req/req_file.py#L118
+    - Requirements File Format:
+      https://pip.pypa.io/en/stable/cli/pip_install/#requirements-file-format
+
+    :param requirements_file: A string path to a requirements file on the local filesystem.
+    :return: A list of parsed requirements (e.g. ``["scikit-learn==0.24.2", ...]``).
+    """
+    with open(requirements_file) as f:
+        lines = f.read().splitlines()
+
+    lines = map(str.strip, lines)
+    lines = map(_strip_inline_comment, lines)
+    lines = _join_continued_lines(lines)
+    lines = filterfalse(_is_comment, lines)
+    lines = filterfalse(_is_empty, lines)
+
+    for line in lines:
+        if _is_requirements_file(line):
+            req_file = line.split(maxsplit=1)[1]
+            abs_path = (
+                req_file
+                if os.path.isabs(req_file)
+                else os.path.join(os.path.dirname(requirements_file), req_file)
+            )
+            yield from _parse_requirements(abs_path)
+        else:
+            yield line
--- a/mlflow/xgboost.py
+++ b/mlflow/xgboost.py
@@ -91,7 +91,7 @@ def save_model(
     signature: ModelSignature = None,
     input_example: ModelInputExample = None,
     pip_requirements=None,
-    additional_pip_requirements=None,
+    extra_pip_requirements=None,
 ):
     """
     Save an XGBoost model to a path on the local file system.
@@ -148,7 +148,7 @@ def save_model(
         written to a ``requirements.txt`` file that is stored as part of the model. These
         requirements are also written to the ``pip`` section of the model's conda environment
         (``conda.yaml``) file.
-    :param additional_pip_requirements: Either an iterable of pip requirement strings
+    :param extra_pip_requirements: Either an iterable of pip requirement strings
         (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
         file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this specifies
         additional pip requirements that are appended to a default set of pip requirements generated
@@ -160,10 +160,10 @@ def save_model(
 
             - ``conda_env``
             - ``pip_requirements``
-            - ``additional_pip_requirements``
+            - ``extra_pip_requirements``
 
         :ref:`This example<pip-requirements-example>` demonstrates how to specify pip requirements
-        using ``pip_requirements`` and ``additional_pip_requirements``.
+        using ``pip_requirements`` and ``extra_pip_requirements``.
     """
     import xgboost as xgb
 
@@ -183,15 +183,13 @@ def save_model(
     # Save an XGBoost model
     xgb_model.save_model(model_data_path)
 
-    _validate_env_arguments(conda_env, pip_requirements, additional_pip_requirements)
+    _validate_env_arguments(conda_env, pip_requirements, extra_pip_requirements)
 
     conda_env_subpath = "conda.yaml"
     if conda_env is None:
         pip_requirements = _parse_pip_requirements(pip_requirements)
-        additional_pip_requirements = _parse_pip_requirements(additional_pip_requirements)
-        pip_reqs = pip_requirements or (
-            _get_default_pip_requirements() + additional_pip_requirements
-        )
+        extra_pip_requirements = _parse_pip_requirements(extra_pip_requirements)
+        pip_reqs = pip_requirements or (_get_default_pip_requirements() + extra_pip_requirements)
         conda_env = _mlflow_conda_env(additional_pip_deps=pip_reqs)
     elif not isinstance(conda_env, dict):
         with open(conda_env, "r") as f:
@@ -217,7 +215,7 @@ def log_model(
     input_example: ModelInputExample = None,
     await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
     pip_requirements=None,
-    additional_pip_requirements=None,
+    extra_pip_requirements=None,
     **kwargs
 ):
     """
@@ -279,7 +277,7 @@ def log_model(
         written to a ``requirements.txt`` file that is stored as part of the model. These
         requirements are also written to the ``pip`` section of the model's conda environment
         (``conda.yaml``) file.
-    :param additional_pip_requirements: Either an iterable of pip requirement strings
+    :param extra_pip_requirements: Either an iterable of pip requirement strings
         (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
         file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this specifies
         additional pip requirements that are appended to a default set of pip requirements generated
@@ -291,10 +289,10 @@ def log_model(
 
             - ``conda_env``
             - ``pip_requirements``
-            - ``additional_pip_requirements``
+            - ``extra_pip_requirements``
 
         :ref:`This example<pip-requirements-example>` demonstrates how to specify pip requirements
-        using ``pip_requirements`` and ``additional_pip_requirements``.
+        using ``pip_requirements`` and ``extra_pip_requirements``.
     :param kwargs: kwargs to pass to `xgboost.Booster.save_model`_ method.
     """
     Model.log(
@@ -307,7 +305,7 @@ def log_model(
         input_example=input_example,
         await_registration_for=await_registration_for,
         pip_requirements=pip_requirements,
-        additional_pip_requirements=additional_pip_requirements,
+        extra_pip_requirements=extra_pip_requirements,
         **kwargs
     )
 
--- a/mlflow/xgboost.py
+++ b/mlflow/xgboost.py
@@ -33,7 +33,12 @@ from mlflow.models.model import MLMODEL_FILE_NAME
 from mlflow.models.signature import ModelSignature
 from mlflow.models.utils import _save_example
 from mlflow.tracking.artifact_utils import _download_artifact_from_uri
-from mlflow.utils.environment import _mlflow_conda_env, _log_pip_requirements
+from mlflow.utils.environment import (
+    _mlflow_conda_env,
+    _log_pip_requirements,
+    _parse_pip_requirements,
+    _validate_env_arguments,
+)
 from mlflow.utils.model_utils import _get_flavor_configuration
 from mlflow.exceptions import MlflowException
 from mlflow.utils.annotations import experimental
@@ -64,17 +69,18 @@ FLAVOR_NAME = "xgboost"
 _logger = logging.getLogger(__name__)
 
 
+def _get_default_pip_requirements():
+    import xgboost as xgb
+
+    return ["xgboost=={}".format(xgb.__version__)]
+
+
 def get_default_conda_env():
     """
     :return: The default Conda environment for MLflow Models produced by calls to
              :func:`save_model()` and :func:`log_model()`.
     """
-    import xgboost as xgb
-
-    return _mlflow_conda_env(
-        # XGBoost is not yet available via the default conda channels, so we install it via pip
-        additional_pip_deps=["xgboost=={}".format(xgb.__version__)]
-    )
+    return _mlflow_conda_env(additional_pip_deps=_get_default_pip_requirements())
 
 
 def save_model(
@@ -84,6 +90,8 @@ def save_model(
     mlflow_model=None,
     signature: ModelSignature = None,
     input_example: ModelInputExample = None,
+    pip_requirements=None,
+    additional_pip_requirements=None,
 ):
     """
     Save an XGBoost model to a path on the local file system.
@@ -95,7 +103,9 @@ def save_model(
                       Conda environment yaml file. If provided, this describes the environment
                       this model should be run in. At minimum, it should specify the dependencies
                       contained in :func:`get_default_conda_env()`. If ``None``, the default
-                      :func:`get_default_conda_env()` environment is added to the model.
+                      :func:`get_default_conda_env()` environment is added to the model. pip
+                      requirements from ``conda_env`` are written to a pip ``requirements.txt``
+                      file and the full conda environment is written to ``conda.yaml``.
                       The following is an *example* dictionary representation of a Conda
                       environment::
 
@@ -130,7 +140,30 @@ def save_model(
                           model. The given example will be converted to a Pandas DataFrame and then
                           serialized to json using the Pandas split-oriented format. Bytes are
                           base64-encoded.
+    :param pip_requirements: Either an iterable of pip requirement strings
+        (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
+        file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this describes the
+        environment this model should be run in. If ``None``, a default list of requirements is
+        inferred from the current software environment. Requirements are automatically parsed and
+        written to a ``requirements.txt`` file that is stored as part of the model. These
+        requirements are also written to the ``pip`` section of the model's conda environment
+        (``conda.yaml``) file.
+    :param additional_pip_requirements: Either an iterable of pip requirement strings
+        (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
+        file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this specifies
+        additional pip requirements that are appended to a default set of pip requirements generated
+        automatically based on the user's current software environment. Requirements are also
+        written to the ``pip`` section of the model's conda environment (``conda.yaml``) file.
 
+        .. warning::
+            The following arguments can't be specified at the same time:
+
+            - ``conda_env``
+            - ``pip_requirements``
+            - ``additional_pip_requirements``
+
+        :ref:`This example<pip-requirements-example>` demonstrates how to specify pip requirements
+        using ``pip_requirements`` and ``additional_pip_requirements``.
     """
     import xgboost as xgb
 
@@ -150,9 +183,16 @@ def save_model(
     # Save an XGBoost model
     xgb_model.save_model(model_data_path)
 
+    _validate_env_arguments(conda_env, pip_requirements, additional_pip_requirements)
+
     conda_env_subpath = "conda.yaml"
     if conda_env is None:
-        conda_env = get_default_conda_env()
+        pip_requirements = _parse_pip_requirements(pip_requirements)
+        additional_pip_requirements = _parse_pip_requirements(additional_pip_requirements)
+        pip_reqs = pip_requirements or (
+            _get_default_pip_requirements() + additional_pip_requirements
+        )
+        conda_env = _mlflow_conda_env(additional_pip_deps=pip_reqs)
     elif not isinstance(conda_env, dict):
         with open(conda_env, "r") as f:
             conda_env = yaml.safe_load(f)
@@ -176,6 +216,8 @@ def log_model(
     signature: ModelSignature = None,
     input_example: ModelInputExample = None,
     await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
+    pip_requirements=None,
+    additional_pip_requirements=None,
     **kwargs
 ):
     """
@@ -188,7 +230,9 @@ def log_model(
                       Conda environment yaml file. If provided, this describes the environment
                       this model should be run in. At minimum, it should specify the dependencies
                       contained in :func:`get_default_conda_env()`. If ``None``, the default
-                      :func:`get_default_conda_env()` environment is added to the model.
+                      :func:`get_default_conda_env()` environment is added to the model. pip
+                      requirements from ``conda_env`` are written to a pip ``requirements.txt``
+                      file and the full conda environment is written to ``conda.yaml``.
                       The following is an *example* dictionary representation of a Conda
                       environment::
 
@@ -227,6 +271,30 @@ def log_model(
     :param await_registration_for: Number of seconds to wait for the model version to finish
                             being created and is in ``READY`` status. By default, the function
                             waits for five minutes. Specify 0 or None to skip waiting.
+    :param pip_requirements: Either an iterable of pip requirement strings
+        (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
+        file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this describes the
+        environment this model should be run in. If ``None``, a default list of requirements is
+        inferred from the current software environment. Requirements are automatically parsed and
+        written to a ``requirements.txt`` file that is stored as part of the model. These
+        requirements are also written to the ``pip`` section of the model's conda environment
+        (``conda.yaml``) file.
+    :param additional_pip_requirements: Either an iterable of pip requirement strings
+        (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
+        file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this specifies
+        additional pip requirements that are appended to a default set of pip requirements generated
+        automatically based on the user's current software environment. Requirements are also
+        written to the ``pip`` section of the model's conda environment (``conda.yaml``) file.
+
+        .. warning::
+            The following arguments can't be specified at the same time:
+
+            - ``conda_env``
+            - ``pip_requirements``
+            - ``additional_pip_requirements``
+
+        :ref:`This example<pip-requirements-example>` demonstrates how to specify pip requirements
+        using ``pip_requirements`` and ``additional_pip_requirements``.
     :param kwargs: kwargs to pass to `xgboost.Booster.save_model`_ method.
     """
     Model.log(
@@ -238,6 +306,8 @@ def log_model(
         signature=signature,
         input_example=input_example,
         await_registration_for=await_registration_for,
+        pip_requirements=pip_requirements,
+        additional_pip_requirements=additional_pip_requirements,
         **kwargs
     )
 
--- a/tests/helper_functions.py
+++ b/tests/helper_functions.py
@@ -22,7 +22,6 @@ from mlflow.tracking.artifact_utils import _download_artifact_from_uri
 from mlflow.utils.file_utils import read_yaml, write_yaml
 from mlflow.utils.environment import _get_pip_deps
 
-
 LOCALHOST = "127.0.0.1"
 
 
--- a/tests/helper_functions.py
+++ b/tests/helper_functions.py
@@ -18,7 +18,10 @@ import pytest
 import mlflow
 import mlflow.pyfunc.scoring_server as pyfunc_scoring_server
 import mlflow.pyfunc
+from mlflow.tracking.artifact_utils import _download_artifact_from_uri
 from mlflow.utils.file_utils import read_yaml, write_yaml
+from mlflow.utils.environment import _get_pip_deps
+
 
 LOCALHOST = "127.0.0.1"
 
@@ -317,12 +320,26 @@ def create_mock_response(status_code, text):
     return response
 
 
+def _read_yaml(path):
+    with open(path, "r") as f:
+        return yaml.safe_load(f)
+
+
+def _read_lines(path):
+    with open(path, "r") as f:
+        return f.read().splitlines()
+
+
 def _compare_conda_env_requirements(env_path, req_path):
     assert os.path.exists(req_path)
+    custom_env_parsed = _read_yaml(env_path)
+    requirements = _read_lines(req_path)
+    assert _get_pip_deps(custom_env_parsed) == requirements
 
-    with open(env_path, "r") as f:
-        custom_env_parsed = yaml.safe_load(f)
-    with open(req_path, "r") as f:
-        requirements = f.read().split("\n")
 
-    assert custom_env_parsed["dependencies"][-1]["pip"] == requirements
+def _assert_pip_requirements(model_uri, expected_pip_reqs):
+    local_path = _download_artifact_from_uri(model_uri)
+    txt_reqs = _read_lines(os.path.join(local_path, "requirements.txt"))
+    conda_reqs = _get_pip_deps(_read_yaml(os.path.join(local_path, "conda.yaml")))
+    assert txt_reqs == expected_pip_reqs
+    assert conda_reqs == expected_pip_reqs
--- a/tests/utils/test_environment.py
+++ b/tests/utils/test_environment.py
@@ -1,7 +1,11 @@
 import os
 import pytest
 
-from mlflow.utils.environment import _mlflow_conda_env
+from mlflow.utils.environment import (
+    _mlflow_conda_env,
+    _parse_pip_requirements,
+    _validate_env_arguments,
+)
 
 
 @pytest.fixture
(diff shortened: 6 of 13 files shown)


mlflow/utils/environment.py
```diff
@@ -80,5 +85,40 @@ def _get_additional_pip_dep(conda_env):
...
+    if isinstance(pip_requirements, str):
+        return [
+            req.requirement for req in parse_requirements(pip_requirements, session=PipSession())
+        ]
+    elif isinstance(pip_requirements, list):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmp_path = os.path.join(tmpdir, f"{uuid.uuid4().hex}.txt")
+            with open(tmp_path, "w") as f:
+                f.write("\n".join(pip_requirements))
+            return [req.requirement for req in parse_requirements(tmp_path, session=PipSession())]
```
harupy, line 102: I think we can directly resolve a list of packages rather than putting them in a tempoary text file, but it requires us to import more internal functions.

tests/utils/test_environment.py
```diff
@@ -50,3 +54,41 @@ def test_mlflow_conda_env_includes_pip_dependencies_and_pip_is_specified(pip_spe
         assert conda_dep in env["dependencies"]
     assert pip_specification in env["dependencies"]
     assert env["dependencies"].count("pip") == (2 if pip_specification == "pip" else 1)
+
+
+def test_resolve_pip_requirements(tmpdir):
+    assert _resolve_pip_requirements(["a", "b"]) == ["a", "b"]
+    assert _resolve_pip_requirements(["a==0.0", "b>1.1"]) == ["a==0.0", "b>1.1"]
```
harupy, line 61: The current implementation doesn't detect conflicting requirements:

```python
log_model(pip_requirements=["pandas==1.2.3", "pandas==4.5.6"])
```

Not sure if this is really necessary, we could detect them and raise an error / warning.

harupy: Tested how pip works for conflicting requirements:

```
% cat a.txt
scikit-learn==0.24.1
scikit-learn==0.24.2

% pip install -r a.txt
Requirement already satisfied: scikit-learn==0.24.1 in /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages (from -r a.txt (line 1)) (0.24.1)
ERROR: Cannot install scikit-learn==0.24.1 and scikit-learn==0.24.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested scikit-learn==0.24.1
    The user requested scikit-learn==0.24.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

Looks like it throws

mlflow/utils/environment.py
```diff
@@ -68,9 +73,9 @@ def _mlflow_additional_pip_env(
         return requirements
 
 
-def _get_additional_pip_dep(conda_env):
+def _get_pip_dep(conda_env):
```
harupy, line 76: Renamed this function since what this function does is just extract pip dependencies in the given conda env.

mlflow/utils/environment.py
```diff
@@ -80,5 +85,40 @@ def _get_additional_pip_dep(conda_env):
...
+    if isinstance(pip_requirements, str):
+        return [
+            req.requirement for req in parse_requirements(pip_requirements, session=PipSession())
+        ]
+    elif isinstance(pip_requirements, list):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmp_path = os.path.join(tmpdir, f"{uuid.uuid4().hex}.txt")
+            with open(tmp_path, "w") as f:
+                f.write("\n".join(pip_requirements))
+            return _parse_pip_requirements(tmp_path)
```
harupy, line 102: The current implementation fails in the following case

```python
def test_parse_pip_requirements_chdir(request, tmpdir):
    try:
        os.chdir(tmpdir)
        f = tmpdir.join("requirements.txt")
        f.write("b")
        assert "requirements.txt" in os.listdir()
        assert _parse_pip_requirements(f.strpath) == ["b"]
        assert _parse_pip_requirements(["a", "-r requirements.txt"]) == ["a", "b"]
    finally:
        os.chdir(request.config.invocation_dir)
```

```
url = '/var/folders/33/kkm875tn7cxb89b8mmcykrx00000gn/T/tmp68onj37q/requirements.txt'
session = <pip._internal.network.session.PipSession object at 0x7fd0b06d1650>

    def get_file_content(url, session):
        # type: (str, PipSession) -> Tuple[str, str]
        """Gets the content of a file; it may be a filename, file: URL, or
        http: URL.  Returns (location, content).  Content is unicode.
        Respects # -*- coding: declarations on the retrieved files.
    
        :param url:         File path or url.
        :param session:     PipSession instance.
        """
        scheme = get_url_scheme(url)
    
        if scheme in ['http', 'https']:
            # FIXME: catch some errors
            resp = session.get(url)
            raise_for_status(resp)
            return resp.url, resp.text
    
        elif scheme == 'file':
            url = url_to_path(url)
    
        try:
            with open(url, 'rb') as f:
                content = auto_decode(f.read())
        except OSError as exc:
            raise InstallationError(
>               f'Could not open requirements file: {exc}'
            )
E           pip._internal.exceptions.InstallationError: Could not open requirements file: [Errno 2] No such file or directory: '/var/folders/33/kkm875tn7cxb89b8mmcykrx00000gn/T/tmp68onj37q/requirements.txt'
```

mlflow/utils/environment.py
```diff
@@ -80,5 +84,61 @@ def _get_additional_pip_dep(conda_env):
...
+                suffix=".tmp.requirements.txt",
+                dir=os.getcwd(),
+                delete=False,
+            )
+            requirements.write("\n".join(pip_requirements))
+            requirements.close()
+            return _parse_pip_requirements(requirements.name)
+        finally:
+            # Clena up the temporary requirements file
+            os.remove(requirements.name)
```
harupy, line 121: `conda` does the same thing when installing pip dependencies

mlflow/xgboost.py
```diff
@@ -64,17 +68,18 @@
 _logger = logging.getLogger(__name__)
 
 
+def _get_default_pip_deps():
+    import xgboost as xgb
+
+    return ["xgboost=={}".format(xgb.__version__)]
```
harupy, line 74: Defined this function to make the code & logic a bit cleaner.

mlflow/xgboost.py
```diff
@@ -176,6 +186,8 @@ def log_model(
     signature: ModelSignature = None,
     input_example: ModelInputExample = None,
     await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
+    pip_requirements=None,
+    additional_pip_requirements=None,
```
harupy, line 190: Can we make some arguments for `log_model` keyword-only to pul related parameters closer?

mlflow/xgboost.py
harupy: We can reuse this function in #https://github.com/mlflow/mlflow/pull/4518 to obtain the pip dependencies we fallback to when the dependency inference fails.

mlflow/utils/environment.py
```diff
@@ -80,5 +84,61 @@ def _get_additional_pip_dep(conda_env):
...
+    elif _is_iterable(pip_requirements) and all(map(_is_string, pip_requirements)):
+        try:
+            # Create a temporary requirements file in the current working directory
+            requirements = tempfile.NamedTemporaryFile(
+                mode="w",
+                prefix="mlflow.",
+                suffix=".tmp.requirements.txt",
+                dir=os.getcwd(),
+                delete=False,
+            )
```
dbczumar, line 115: Nit: We can use [mlflow.utils.file_utils.TempDir](https://github.com/mlflow/mlflow/blob/bc86f1dfdb6444746470b7b12b6582b1390eb1aa/mlflow/utils/file_utils.py#L183) to avoid the try/catch

mlflow/utils/environment.py
```diff
@@ -80,5 +84,61 @@ def _get_additional_pip_dep(conda_env):
...
+        finally:
+            # Clena up the temporary requirements file
+            os.remove(requirements.name)
+
+    else:
+        raise TypeError(
+            "Invalid type for `pip_requirements`: {}, must be `str` or iterable of `str`".format(
+                type(pip_requirements)
+            )
+        )
```
dbczumar, line 128: To clarify, this error can be more specific: `pip_requirements` is either a string path to a pip requirements file on the local filesystem or a list of pip requirement strings.

mlflow/xgboost.py
dbczumar: This is a good callout. We can do this in MLflow 2.0. Unfortunately, it would be a breaking change for a pre-2.0 release.

mlflow/xgboost.py
dbczumar: Cool!

mlflow/utils/environment.py
harupy: Sound good, I'll update the error message

mlflow/xgboost.py
harupy: I'll define the same function in othere flavors.

mlflow/utils/environment.py
harupy: Thanks, I'll use `mlflow.utils.file_utils.TempDir` with `chdr`.

harupy: Found an issue related to the comment:

https://bugs.python.org/issue14243

harupy: Hmm `mlflow.utils.file_utils.TempDir` doesn't seem to work because we need to create a temporary requirements file in the current working directory

mlflow/ml-package-versions.yml
```diff
@@ -337,3 +337,15 @@ spark:
...
       find tests/spark_autologging/ml -name 'test*.py' | xargs -L 1 pytest --large
+
+pip:
+  package_info:
+    pip_release: "pip"
+    install_dev: |
+      pip install git+https://github.com/pypa/pip.git
+
+  models:
+    minimum: "20.1"
```
harupy, line 348: In pip >= 20.1, `pip._internal.req.parse_requirements` returns instances of `ParsedRequirement` which provides an attribute named `requirement` which stores a raw requirement string.

In pip < 20.1, `pip._internal.req.parse_requirements` returns instances of `InstallRequirement` which doesn't provide such attribute. It's a bit cumbersome to restore the raw requirement string.

harupy:
@dbczumar This PR is ready for another review. Could you take a look?

mlflow/utils/environment.py
harupy: One interesting behavior of `sys.modules` that I found is:

```diff
>>> import sys
>>> sys.modules.keys()
[
    "sys",
    "builtins",
    "_frozen_importlib",
    "_imp",
    "_thread",
    "_warnings",
    "_weakref",
    "zipimport",
    "_frozen_importlib_external",
    "_io",
    "marshal",
    "posix",
    "encodings",
    "codecs",
    "_codecs",
    "encodings.aliases",
    "encodings.utf_8",
    "_signal",
    "__main__",
    "encodings.latin_1",
    "io",
    "abc",
    "_abc",
    "site",
    "os",
    "stat",
    "_stat",
    "_collections_abc",
    "posixpath",
    "genericpath",
    "os.path",
    "_sitebuiltins",
    "_bootlocale",
    "_locale",
    "types",
    "importlib",
    "importlib._bootstrap",
    "importlib._bootstrap_external",
    "warnings",
    "importlib.util",
    "importlib.abc",
    "importlib.machinery",
    "contextlib",
    "collections",
    "operator",
    "_operator",
    "keyword",
    "heapq",
    "_heapq",
    "itertools",
    "reprlib",
    "_collections",
    "functools",
    "_functools",
+    "google",
+    "google.cloud",
+    "google.logging",
+    "mpl_toolkits",
    "ruamel",
+    "sphinxcontrib",
    "readline",
    "atexit",
    "rlcompleter",
]
```

harupy: Running `python -v` gives:

```
% python -v
import _frozen_importlib # frozen
import _imp # builtin
import '_thread' # <class '_frozen_importlib.BuiltinImporter'>
import '_warnings' # <class '_frozen_importlib.BuiltinImporter'>
import '_weakref' # <class '_frozen_importlib.BuiltinImporter'>
# installing zipimport hook
import 'zipimport' # <class '_frozen_importlib.BuiltinImporter'>
# installed zipimport hook
import '_frozen_importlib_external' # <class '_frozen_importlib.FrozenImporter'>
import '_io' # <class '_frozen_importlib.BuiltinImporter'>
import 'marshal' # <class '_frozen_importlib.BuiltinImporter'>
import 'posix' # <class '_frozen_importlib.BuiltinImporter'>
import _thread # previously loaded ('_thread')
import '_thread' # <class '_frozen_importlib.BuiltinImporter'>
import _weakref # previously loaded ('_weakref')
import '_weakref' # <class '_frozen_importlib.BuiltinImporter'>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/__init__.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__init__.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/__init__.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/codecs.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/codecs.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/codecs.cpython-37.pyc'
import '_codecs' # <class '_frozen_importlib.BuiltinImporter'>
import 'codecs' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8150dd0>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/aliases.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/aliases.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/aliases.cpython-37.pyc'
import 'encodings.aliases' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8174e90>
import 'encodings' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8150250>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/utf_8.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/utf_8.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/utf_8.cpython-37.pyc'
import 'encodings.utf_8' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8150d90>
import '_signal' # <class '_frozen_importlib.BuiltinImporter'>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/latin_1.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/latin_1.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/encodings/__pycache__/latin_1.cpython-37.pyc'
import 'encodings.latin_1' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8168490>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/io.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/io.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/io.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/abc.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/abc.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/abc.cpython-37.pyc'
import '_abc' # <class '_frozen_importlib.BuiltinImporter'>
import 'abc' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8168e50>
import 'io' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8168510>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/site.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/site.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/os.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/os.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/os.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/stat.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/stat.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/stat.cpython-37.pyc'
import '_stat' # <class '_frozen_importlib.BuiltinImporter'>
import 'stat' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c81ffa10>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/_collections_abc.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/_collections_abc.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/_collections_abc.cpython-37.pyc'
import '_collections_abc' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8203510>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/posixpath.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/posixpath.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/posixpath.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/genericpath.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/genericpath.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/genericpath.cpython-37.pyc'
import 'genericpath' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8233390>
import 'posixpath' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8212510>
import 'os' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c81f0a10>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/_sitebuiltins.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/_sitebuiltins.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/_sitebuiltins.cpython-37.pyc'
import '_sitebuiltins' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c81f0fd0>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/_bootlocale.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/_bootlocale.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/_bootlocale.cpython-37.pyc'
import '_locale' # <class '_frozen_importlib.BuiltinImporter'>
import '_bootlocale' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c81f6510>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/types.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/types.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/types.cpython-37.pyc'
import 'types' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8245490>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/__init__.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__init__.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/__init__.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/warnings.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/warnings.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/warnings.cpython-37.pyc'
import 'warnings' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c824ac90>
import 'importlib' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c8245c90>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/util.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/util.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/util.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/abc.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/abc.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/abc.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/machinery.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/machinery.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/importlib/__pycache__/machinery.cpython-37.pyc'
import 'importlib.machinery' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c825c150>
import 'importlib.abc' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c825b390>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/contextlib.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/contextlib.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/contextlib.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/collections/__pycache__/__init__.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/collections/__init__.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/collections/__pycache__/__init__.cpython-37.pyc'
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/operator.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/operator.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/operator.cpython-37.pyc'
import '_operator' # <class '_frozen_importlib.BuiltinImporter'>
import 'operator' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1b8067ed0>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/keyword.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/keyword.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/keyword.cpython-37.pyc'
import 'keyword' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1b806bc10>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/heapq.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/heapq.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/heapq.cpython-37.pyc'
# extension module '_heapq' loaded from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/lib-dynload/_heapq.cpython-37m-darwin.so'
# extension module '_heapq' executed from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/lib-dynload/_heapq.cpython-37m-darwin.so'
import '_heapq' # <_frozen_importlib_external.ExtensionFileLoader object at 0x7fb1b807f550>
import 'heapq' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1b8076690>
import 'itertools' # <class '_frozen_importlib.BuiltinImporter'>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/reprlib.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/reprlib.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/reprlib.cpython-37.pyc'
import 'reprlib' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1b807f890>
import '_collections' # <class '_frozen_importlib.BuiltinImporter'>
import 'collections' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1b80479d0>
# /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/functools.cpython-37.pyc matches /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/functools.py
# code object from '/Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/__pycache__/functools.cpython-37.pyc'
import '_functools' # <class '_frozen_importlib.BuiltinImporter'>
import 'functools' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1b804c110>
import 'contextlib' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c825bb50>
import 'importlib.util' # <_frozen_importlib_external.SourceFileLoader object at 0x7fb1c824fb50>
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google/cloud
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google/cloud
# destroy google.cloud
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google/logging
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/mpl_toolkits
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/google
# destroy google
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/ruamel
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/sphinxcontrib
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/sphinxcontrib
# destroy sphinxcontrib
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/sphinxcontrib
# destroy sphinxcontrib
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/sphinxcontrib
# destroy sphinxcontrib
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/sphinxcontrib
# destroy sphinxcontrib
# possible namespace for /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages/sphinxcontrib
# destroy sphinxcontrib
```

harupy: One interesting behavior of `sys.modules` that I found is:

```diff
>>> import sys
>>> sys.modules.keys()
[
    "sys",
    "builtins",
    "_frozen_importlib",
    "_imp",
    "_thread",
    "_warnings",
    "_weakref",
    "zipimport",
    "_frozen_importlib_external",
    "_io",
    "marshal",
    "posix",
    "encodings",
    "codecs",
    "_codecs",
    "encodings.aliases",
    "encodings.utf_8",
    "_signal",
    "__main__",
    "encodings.latin_1",
    "io",
    "abc",
    "_abc",
    "site",
    "os",
    "stat",
    "_stat",
    "_collections_abc",
    "posixpath",
    "genericpath",
    "os.path",
    "_sitebuiltins",
    "_bootlocale",
    "_locale",
    "types",
    "importlib",
    "importlib._bootstrap",
    "importlib._bootstrap_external",
    "warnings",
    "importlib.util",
    "importlib.abc",
    "importlib.machinery",
    "contextlib",
    "collections",
    "operator",
    "_operator",
    "keyword",
    "heapq",
    "_heapq",
    "itertools",
    "reprlib",
    "_collections",
    "functools",
    "_functools",
+    "google",
+    "google.cloud",
+    "google.logging",
+    "mpl_toolkits",
    "ruamel",
+    "sphinxcontrib",
    "readline",
    "atexit",
    "rlcompleter",
]
```

mlflow/ml-package-versions.yml
harupy: Todo: Check the version of pip installed on MLRs

mlflow/utils/environment.py
```diff
@@ -80,5 +86,77 @@ def _get_additional_pip_dep(conda_env):
...
+        ]
+    elif _is_iterable(pip_requirements) and all(map(_is_string, pip_requirements)):
+        try:
+            # Create a temporary requirements file in the current working directory
+            tmp_req_file = tempfile.NamedTemporaryFile(
+                mode="w",
+                prefix="mlflow.",
+                suffix=".tmp.requirements.txt",
+                dir=os.getcwd(),
+                # Setting `delete` to False causes a permission-denied error on Windows
```
harupy, line 116:
```suggestion
                # Setting `delete` to True causes a permission-denied error on Windows
```

mlflow/ml-package-versions.yml
harupy: Pip versions in MLRs:

| MLR | pip version |
| :-- | :---------- |
| 7.6 | 20.0.2      |
| 8.0 | 20.2.4      |
| 8.1 | 20.2.4      |
| 8.2 | 20.2.4      |

Source: https://docs.databricks.com/runtime/mlruntime.html#libraries-included-in-databricks-runtime-ml

dbczumar: Out of curiosity, what did older versions of conda do to support this?

mlflow/ml-package-versions.yml
```diff
@@ -337,3 +337,15 @@ spark:
     run: |
       # TODO: Add datasource autologging tests
       find tests/spark_autologging/ml -name 'test*.py' | xargs -L 1 pytest --large
+
+pip:
```
dbczumar, line 341: @harupy If we don't plan to explicitly require `pip >= 20.1` in `setup.py`, then we don't need to worry about MLR / Microsoft. However, how do we plan to handle the case where the user's installed pip version is less than the minimum? We could raise an exception if the user passes a file path to `pip_requirements` if `pip < 20.1`.

If MLR / MSFT agrees that it's okay, my preference would be to add `pip` to `setup.py`.

mlflow/models/model.py
```diff
@@ -143,6 +143,8 @@ def log(
         flavor,
         registered_model_name=None,
         await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
+        pip_requirements=None,
+        additional_pip_requirements=None,
```
dbczumar, line 147: Nice! Not sure why we didn't do this with `conda_env` in the first place.

mlflow/xgboost.py
```diff
@@ -130,7 +138,22 @@ def save_model(
                           model. The given example will be converted to a Pandas DataFrame and then
                           serialized to json using the Pandas split-oriented format. Bytes are
                           base64-encoded.
+    :param pip_requirements: Overwrites pip requirements of the conda environment this model should
```
dbczumar, line 141: `pip_requirements` should control which requirements are logged to `requirements.txt`. This is more important for users than the conda environment behavior. We should emphasize `requirements.txt` and say that the conda environment behavior is there for backwards compatibility.

mlflow/xgboost.py
```diff
@@ -130,7 +138,22 @@ def save_model(
...
+        be run in. Accepts either a string path to a pip requirements file on the local filesystem
+        or an iterable of pip requirement strings (e.g. :code:`["pandas", "-r requirements.txt"]`).
+    :param additional_pip_requirements: Works similarly to ``pip_requirements`` but specifies
+        additional pip requirements.
```
dbczumar, line 145: Can we clarify that this adds pip requirements to a default set of requirements generated automatically by the model flavor based on the user's current software environment?

mlflow/xgboost.py
```diff
@@ -227,6 +257,8 @@ def log_model(
     :param await_registration_for: Number of seconds to wait for the model version to finish
                             being created and is in ``READY`` status. By default, the function
                             waits for five minutes. Specify 0 or None to skip waiting.
+    :pip_requirements: <Add description>
+    :additional_pip_requirements: <Add description>
```
dbczumar, line 261: Can we add param docs here too?

tests/xgboost/test_xgboost_model_export.py
```diff
@@ -212,6 +213,58 @@ def test_model_save_persists_requirements_in_mlflow_model_directory(
     _compare_conda_env_requirements(xgb_custom_env, saved_pip_req_path)
 
 
+@pytest.mark.large
+def test_save_model_with_pip_requirements(xgb_model, tmpdir):
```
dbczumar, line 217: Can we add test coverage for log_model as well?

mlflow/xgboost.py
```diff
@@ -130,7 +138,22 @@ def save_model(
                           model. The given example will be converted to a Pandas DataFrame and then
                           serialized to json using the Pandas split-oriented format. Bytes are
                           base64-encoded.
+    :param pip_requirements: Overwrites pip requirements of the conda environment this model should
+        be run in. Accepts either a string path to a pip requirements file on the local filesystem
+        or an iterable of pip requirement strings (e.g. :code:`["pandas", "-r requirements.txt"]`).
```
dbczumar, line 143: If a user specifies `-r requirements.txt`, they'll need to bundle `requirements.txt` with the model manually (and presumably use an absolute path). Perhaps we should use an example that will work out of the box, since this one is tricky to get right. Can we replace this with a pinned library installation, such as `xgboost==1.4.2`?

tests/helper_functions.py
```diff
@@ -317,12 +318,18 @@ def create_mock_response(status_code, text):
...
+        return yaml.safe_load(f)
+
 
-    with open(env_path, "r") as f:
-        custom_env_parsed = yaml.safe_load(f)
-    with open(req_path, "r") as f:
-        requirements = f.read().split("\n")
+def _read_lines(path):
+    with open(path, "r") as f:
+        return f.read().split("\n")
```
dbczumar, line 328: Nit: how about `f.readlines()`?

mlflow/xgboost.py
dbczumar: I tried this out manually with the following code: 

```
import mlflow

import xgboost as xgb
from sklearn import datasets

iris = datasets.load_iris()
X = iris.data
y = iris.target

params = {
    "objective": "multi:softprob",
    "num_class": 3,
    "eval_metric": "mlogloss",
    "seed": 42,
}
model = xgb.train(params, dtrain, evals=[(dtrain, "train")])

mlflow.xgboost.log_model(
    model,
    additional_pip_requirements=[
        "foobaz==barbiz",
    ],
    artifact_path="foobiz",
)
```

The resulting requirements.txt doesn't contain my test requirement:

```
mlflow
xgboost==1.5.0-dev
```

However, the conda environment is correct:

```
channels:
- conda-forge
dependencies:
- python=3.8.5
- pip
- pip:
  - mlflow
  - xgboost==1.5.0-dev
  - foobaz==barbiz
name: mlflow-env
```

mlflow/xgboost.py
```diff
@@ -150,9 +173,14 @@ def save_model(
...
     xgb_model.save_model(model_data_path)
 
+    _validate_env_arguments(conda_env, pip_requirements, additional_pip_requirements)
+
     conda_env_subpath = "conda.yaml"
     if conda_env is None:
-        conda_env = get_default_conda_env()
+        final_pip_requirements = _determine_pip_requirements(
+            _get_default_pip_deps(), pip_requirements, additional_pip_requirements,
+        )
```
dbczumar, line 182: What happens if there's an error in this procedure? Should we throw a clear `MlflowException` indicating what happened?

mlflow/xgboost.py
harupy: They don't need to. The currently approach extract packages from a requirements file. For example,

```
# Mix of both shell and python

>>> cat my_requirements.txt
xgboost
scikit-learn

>>> save_model(pip_requiments=["-r my_requirements.txt"])
```

The code above logs a requirements file that looks like:

```
xgboost
scikit-learn
```

tests/helper_functions.py
harupy: I first tried `f.readlines()`, and then switched to `f.read().split("\n")` to drop line breaks, and then found `f.read().splitlines()` today:

```python
with open("a.txt") as f:
    print(f.read())
    f.seek(0)
    print(f.readlines())
    f.seek(0)
    print(f.read().split("\n"))
    f.seek(0)
    print(f.read().splitlines())
```

Result:

```
with open("a.txt") as f:
    print(f.read())
    f.seek(0)
    print(f.readlines())
    f.seek(0)
    print(f.read().split("\n"))
    f.seek(0)
    print(f.read().splitlines())
```

mlflow/ml-package-versions.yml
harupy: Conda just does the following when installing pip dependencies:

1. Store pip dependencies as a temporary requirements file.
2. And then, run `pip install -r` against the temporary file using `subprocess`.

https://github.com/conda/conda/blob/33a142c16530fcdada6c377486f1c1a385738a96/conda_env/installers/pip.py#L53

So conda doesn't have to think about how `pip._internal.req.parse_requirements` behaves.

tests/xgboost/test_xgboost_model_export.py
harupy: Absolutely!

mlflow/xgboost.py
harupy: Got it!

mlflow/xgboost.py
harupy: Yep!

mlflow/xgboost.py
harupy: Strange, I'll take a look.

mlflow/xgboost.py
harupy: Alternative approach is detect requirements files and bundle them with the model.

dbczumar: Got it - apologies, misread the logic a bit. Can we add some inline comments for clarity?

mlflow/xgboost.py
harupy: There are a couple cases where this procedure fails:

#### 1. Invalid value is passed to `pip_requirements` or `additional_pip_requirements`

In this case, an error indicating that they only accept a string or an interable of strings, is thrown.

#### 2. Given requirements file doesn't exist

In this case, `FileNotFoundError` is thrown in `_parse_pip_requirements`, which I think is clear enough.

So I don't think we need to throw clear `MlflowException` for now.

mlflow/xgboost.py
```diff
@@ -130,7 +138,29 @@ def save_model(
...
+        a pip requirements file on the local filesystem (e.g. ``"requirements.txt"``) or an iterable
+        of pip requirement strings (e.g. ``["scikit-learn", "-r requirements.txt"]``) is acceptable.
+    :param additional_pip_requirements: Specifies additional pip requirements on top of a default
+        set of pip requirements generated automatically by the model flavor on the user's current
+        software environment. The acceptable value type is the same as ``pip_requirements``.
+
+        .. note::
+            It is recommended to use ``pip_requirements`` and ``additional_pip_requirements``
+            when specifying pip requirements rather than ``conda_env`` which has been deprecated
+            and only exists for backwards compatibility.
```
harupy, line 153: @dbczumar I've addressed your comment on these parameters. What do you think? If this looks good, I'll address https://github.com/mlflow/mlflow/pull/4519/files#r665882073

mlflow/utils/req_parsing.py
```diff
@@ -0,0 +1,77 @@
+import os
```
dbczumar, line 1: Can we rename this file `requirements_utils.py`?

mlflow/utils/req_parsing.py
```diff
@@ -0,0 +1,77 @@
...
+
+    - Remove comments and blank lines
+    - Join continued lines
+    - Resolve requirements file references (e.g. '-r requirements.txt')
+
+    References:
+    - `pip._internal.req.parse_requirements`:
+      https://github.com/pypa/pip/blob/7a77484a492c8f1e1f5ef24eaf71a43df9ea47eb/src/pip/_internal/req/req_file.py#L118
+    - Requirements File Format:
+      https://pip.pypa.io/en/stable/cli/pip_install/#requirements-file-format
```
dbczumar, line 56: I noticed that constraint files are also supported by the requirements file format (`-c`). Do we need to add those?

mlflow/xgboost.py
```diff
@@ -130,7 +138,29 @@ def save_model(
                           model. The given example will be converted to a Pandas DataFrame and then
                           serialized to json using the Pandas split-oriented format. Bytes are
                           base64-encoded.
-
+    :param pip_requirements: Overwrites a default set of pip requirements generated automatically
```
dbczumar, line 141: Can we adjust the phrasing here so that it's similar to `conda_env`? How about:

```
Either an iterable of pip requirement strings (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements file on. the local filesystem (e.g. ``"requirements.txt"``). If provided, this describes the environment this model should be run in. If ``None``, a default list of requirements is inferred from the current software environment.
```

mlflow/xgboost.py
```diff
@@ -130,7 +138,29 @@ def save_model(
...
+        by the model flavor on the user's current software environment. The given requirements are
+        automatically parsed and logged as a single ``requirements.txt`` file. Either string path to
+        a pip requirements file on the local filesystem (e.g. ``"requirements.txt"``) or an iterable
+        of pip requirement strings (e.g. ``["scikit-learn", "-r requirements.txt"]``) is acceptable.
+    :param additional_pip_requirements: Specifies additional pip requirements on top of a default
+        set of pip requirements generated automatically by the model flavor on the user's current
+        software environment. The acceptable value type is the same as ``pip_requirements``.
+
+        .. note::
+            It is recommended to use ``pip_requirements`` and ``additional_pip_requirements``
```
dbczumar, line 151:
```suggestion
            It is recommended to use ``pip_requirements`` or ``additional_pip_requirements``
```

mlflow/xgboost.py
dbczumar: Sounds good!

mlflow/xgboost.py
```diff
@@ -130,7 +138,29 @@ def save_model(
...
+            when specifying pip requirements rather than ``conda_env`` which has been deprecated
+            and only exists for backwards compatibility.
```
dbczumar, line 153: On second thought, do we need to deprecate `conda_env`? As long as `conda_env` / `pip_requirements` / `additional_pip_requirements` write to *both* `requirements.txt` *and* `conda.yaml`, I think we can keep all three. Some users may want to use more sophisticated conda environments.

mlflow/xgboost.py
```diff
@@ -130,7 +138,29 @@ def save_model(
...
                           base64-encoded.
-
+    :param pip_requirements: Overwrites a default set of pip requirements generated automatically
+        by the model flavor on the user's current software environment. The given requirements are
+        automatically parsed and logged as a single ``requirements.txt`` file. Either string path to
+        a pip requirements file on the local filesystem (e.g. ``"requirements.txt"``) or an iterable
+        of pip requirement strings (e.g. ``["scikit-learn", "-r requirements.txt"]``) is acceptable.
+    :param additional_pip_requirements: Specifies additional pip requirements on top of a default
+        set of pip requirements generated automatically by the model flavor on the user's current
+        software environment. The acceptable value type is the same as ``pip_requirements``.
```
dbczumar, line 148:
```
Either an iterable of pip requirement strings (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string 
path to a pip requirements file on the local filesystem (e.g. ``"requirements.txt"``). Specifies additional
pip requirements that are appended to a default set of pip requirements generated automatically based
on the user's current software environment.
```

tests/xgboost/test_xgboost_model_export.py
```diff
@@ -212,6 +213,88 @@ def test_model_save_persists_requirements_in_mlflow_model_directory(
...
+    mlflow.xgboost.save_model(xgb_model.model, tmpdir1.strpath, pip_requirements=f.strpath)
+    reqs = _read_lines(tmpdir1.join("requirements.txt").strpath)
+    assert reqs == ["mlflow", "a"]
+
+    # List of requirements
+    tmpdir2 = tmpdir.join("2")
+    mlflow.xgboost.save_model(
+        xgb_model.model, tmpdir2.strpath, pip_requirements=[f"-r {f.strpath}", "b"]
+    )
+    reqs = _read_lines(tmpdir2.join("requirements.txt").strpath)
```
dbczumar, line 231: Nice! For each of these test cases, can we also add logic to confirm that the `pip` section of `conda.yaml` is populated with the expected dependencies?

dbczumar (commented):
@harupy Awesome! LGTM once a few remaining comments are addressed!

tests/utils/test_req_parsing.py
```diff
@@ -0,0 +1,144 @@
...
+    assert list(_join_continued_lines(["a\\"])) == ["a"]
+    assert list(_join_continued_lines(["\\", "a"])) == ["a"]
+
+
+def test_parse_requirements(request, tmpdir):
+    """
+    Ensures `_parse_requirements` returns the same result as `pip._internal.req.parse_requirements`
+    """
+    from pip._internal.req import parse_requirements as pip_parse_requirements
+    from pip._internal.network.session import PipSession
```
dbczumar, line 65: Do we need to add `pip` as a test-only requirement?

tests/xgboost/test_xgboost_model_export.py
harupy: Got it, I'll update the tests to ensure that.

mlflow/xgboost.py
dbczumar: Following up on this, let's make sure that each of these parameter description indicates that requirements are written in two formats: `conda.yaml` *and* `requirements.txt`.

mlflow/utils/req_parsing.py
harupy: Thanks for the catch, yeah I think we do.

```
# a.txt
scikit-learn
-c b.txt
```

```
# b.txt
xgboost < 1.5
```

```
>>> pip install -r a.txt
Requirement already satisfied: scikit-learn in /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages (from -r b.txt (line 1)) (0.24.1)
Requirement already satisfied: threadpoolctl>=2.0.0 in /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages (from scikit-learn->-r b.txt (line 1)) (2.1.0)
Requirement already satisfied: scipy>=0.19.1 in /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages (from scikit-learn->-r b.txt (line 1)) (1.6.2)
Requirement already satisfied: joblib>=0.11 in /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages (from scikit-learn->-r b.txt (line 1)) (1.0.1)
Requirement already satisfied: numpy>=1.13.3 in /Users/harutakakawamura/.pyenv/versions/miniconda3-4.7.12/envs/mlflow-dev-env/lib/python3.7/site-packages (from scikit-learn->-r b.txt (line 1)) (1.21.0)
```

We could merge `a.txt` and `b.txt` into one list like:

```
scikit-learn
xgboost < 1.5
```

but this is actually incorrect.

mlflow/xgboost.py
harupy: Good question, given we still rely on conda and don't provide a way to alternate `conda` (e.g. Docker), it actually makes sense to keep `conda_env`.

harupy: I'll close #4520.

mlflow/utils/req_parsing.py
harupy: One thing we could do is collect constraint requirements in a single constraint file and bundle it with model.

mlflow/utils/req_parsing.py
harupy: Sure!

mlflow/xgboost.py
harupy: Sure!

mlflow/utils/req_parsing.py
harupy: We can address this in a follow-up because this PR is getting larger.

harupy: Offline discussion: we'll address this as a follow-up.

I'll add a TODO comment for this.

mlflow/xgboost.py
harupy: Thanks!

mlflow/utils/requirements_utils.py
```diff
@@ -0,0 +1,78 @@
...
+            continued_lines.append(line)
+            yield "".join(continued_lines)
+            continued_lines.clear()
+
+    # The last line ends with '\'
+    if continued_lines:
+        yield "".join(continued_lines)
+
+# TODO: Add support for constraint files:
+#       https://github.com/mlflow/mlflow/pull/4519#discussion_r668412179
```
harupy, line 43: @dbczumar Add a TODO comment on support for constraint files.

mlflow/utils/environment.py
```diff
@@ -80,5 +82,78 @@ def _get_additional_pip_dep(conda_env):
 
 
 def _log_pip_requirements(conda_env, path, requirements_file="requirements.txt"):
-    pip_deps = _get_additional_pip_dep(conda_env)
+    pip_deps = _get_pip_deps(conda_env)
     _mlflow_additional_pip_env(pip_deps, path=os.path.join(path, requirements_file))
+
+
+def _parse_pip_requirements(pip_requirements):
```
dbczumar, line 89: Can we add some method documentation, including information about the accepted types for the `pip_requirements` parameter?

mlflow/utils/environment.py
```diff
@@ -80,5 +82,78 @@ def _get_additional_pip_dep(conda_env):
...
+    else:
+        raise TypeError(
+            "`pip_requirements` must be either a string path to a pip requirements file on the "
+            "local filesystem or an iterable of pip requirement strings, but got `{}`".format(
+                type(pip_requirements)
+            )
+        )
+
+
+def _determine_pip_requirements(
```
dbczumar, line 129: Can we add a docstring to this method?

mlflow/utils/requirements_utils.py
```diff
@@ -0,0 +1,78 @@
+import os
```
dbczumar, line 1: Can we add some module-level documentation to indicate that this module defines utilities for interpreting and creating requirements files (e.g. pip's `requirements.txt`), which is helpful for managing ML software environments?

mlflow/utils/requirements_utils.py
dbczumar: Awesome!

mlflow/utils/requirements_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    # The last line ends with '\'
+    if continued_lines:
+        yield "".join(continued_lines)
+
+# TODO: Add support for constraint files:
+#       https://github.com/mlflow/mlflow/pull/4519#discussion_r668412179
+def _parse_requirements(requirements_file):
+    """
+    A simplified version of `pip._internal.req.parse_requirements` which performs the following
+    operations on the given requirements file and yields the parsed requirements.
```
dbczumar, line 47: Can we indicate the parameter type for `requirements_file` - i.e. a string path to a requirements file on the local filesystem.

mlflow/xgboost.py
```diff
@@ -130,7 +138,29 @@ def save_model(
...
+        environment this model should be run in. If ``None``, a default list of requirements is
+        inferred from the current software environment. Requirements are automatically parsed and
+        written to a ``requirements.txt`` file that is stored as part of the model. These
+        requirements are also written to the ``pip`` section of the model's conda environment
+        (``conda.yaml``) file.
+    :param additional_pip_requirements: Either an iterable of pip requirement strings
+        (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
+        file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this specifies
+        additional pip requirements that are appended to a default set of pip requirements generated
+        automatically based on the user's current software environment.
```
dbczumar, line 153: Can we also include 

```
requirements are also written to the ``pip`` section of the model's conda environment (``conda.yaml``) file.
```

like we do in the docstring for `pip_requirements`?

mlflow/xgboost.py
```diff
@@ -227,6 +264,29 @@ def log_model(
...
```
dbczumar, line 279: Can we also include 

```
requirements are also written to the ``pip`` section of the model's conda environment (``conda.yaml``) file.
```

like we do in the docstring for `pip_requirements`?

mlflow/xgboost.py
```diff
@@ -130,7 +138,29 @@ def save_model(
...
                           base64-encoded.
-
+    :param pip_requirements: Either an iterable of pip requirement strings
+        (e.g. ``["scikit-learn", "-r requirements.txt"]``) or the string path to a pip requirements
+        file on the local filesystem (e.g. ``"requirements.txt"``). If provided, this describes the
+        environment this model should be run in. If ``None``, a default list of requirements is
+        inferred from the current software environment. Requirements are automatically parsed and
+        written to a ``requirements.txt`` file that is stored as part of the model. These
+        requirements are also written to the ``pip`` section of the model's conda environment
+        (``conda.yaml``) file.
```
dbczumar, line 148: This section is fantastic! Thank you! Can we include similar information for the `conda_env` parameter to indicate where dependency information is stored? I.e. indicate that pip requirements from `conda_env` are written to a pip `requirements.txt` file and that the full conda environment is written to `conda.yaml`. Can we do this for `log_model()` as well?

dbczumar (approved):
@harupy LGTM with a few comment / docs suggestions. Once those are addressed, feel free to commit changes for the other model flavors. We can also do that as part of a follow-up PR (which might help minimize the size of the diff). Awesome work!

mlflow/utils/requirements_utils.py
harupy: Sure!

mlflow/xgboost.py
harupy: Got it!

harupy: Done!

mlflow/xgboost.py
harupy: This looks like a dup of https://github.com/mlflow/mlflow/pull/4519#discussion_r669310782

mlflow/xgboost.py
harupy: Sure!

mlflow/utils/requirements_utils.py
harupy: Got it!

mlflow/utils/environment.py
harupy: I removed this function.

PR #4520, opened 2021-07-01
[WIP][ML-16201] Deprecate the use of `conda_env` argument in `log_model`

harupy:

[WIP][ML-16201] Deprecate the use of `conda_env` argument in `log_model`.

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/annotations.py
+++ b/mlflow/utils/annotations.py
@@ -66,21 +66,25 @@ def keyword_only(func):
     return wrapper
 
 
-def deprecate_conda_env(f):
+def deprecate_conda_env(func):
     """
     Wraps the given function to raise a deprecation warning when the `conda_env` argument is
     supplied.
     """
     conda_env_var_name = "conda_env"
-    spec = inspect.getfullargspec(f)
-    # Note `spec.args` contains positional, positional-only (introduced in Python 3.8),
-    # and keyword arguments.
-    conda_env_index = (spec.args + spec.kwonlyargs).index(conda_env_var_name)
+    parameters = inspect.signature(func).parameters
+    conda_env_index = list(parameters.keys()).index(conda_env_var_name)
+    # Assuming `conda_env` is defined as a keyword or keyword-only argument
+    default_value = parameters[conda_env_var_name].default
 
-    @wraps(f)
+    @wraps(func)
     def wrapper(*args, **kwargs):
-        is_conda_env_supplied = len(args) > conda_env_index or conda_env_var_name in kwargs
-        if is_conda_env_supplied:
+        conda_env_val = (
+            args[conda_env_index]
+            if len(args) > conda_env_index
+            else kwargs.get(conda_env_var_name, default_value)
+        )
+        if conda_env_val != default_value:
             warnings.warn(
                 (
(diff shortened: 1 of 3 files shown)


mlflow/utils/annotations.py
```diff
@@ -63,3 +64,25 @@ def wrapper(*args, **kwargs):
...
+    conda_env_index = spec.args.index(conda_env_var_name)
+
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        is_conda_env_supplied = len(args) > conda_env_index or conda_env_var_name in kwargs
+        if is_conda_env_supplied:
+            warnings.warn(
+                (
+                    "The `conda_env` argument has been deprecated, "
+                    "please use `pip_requirements` instead"
```
harupy, line 81:
```suggestion
                    "please use `pip_requirements` or `additional_pip_requirements` instead"
```

tests/utils/test_annotations.py
```diff
@@ -0,0 +1,26 @@
...
+
+def test_deprecate_conda_env():
+    @deprecate_conda_env
+    def f(a, conda_env=None):
+        """docstring"""
+
+    assert f.__name__ == "f"
+    assert f.__doc__ == "docstring"
+
+    with pytest.warns(FutureWarning, match="`conda_env` has been deprecated"):
```
harupy, line 15:
```suggestion
    with pytest.warns(FutureWarning, match="The `conda_env` argument has been deprecated"):
```

mlflow/xgboost.py
```diff
@@ -168,6 +169,7 @@ def save_model(
     mlflow_model.save(os.path.join(path, MLMODEL_FILE_NAME))
 
 
+@deprecate_conda_env
```
harupy, line 172: I'll decorate `log_model` and `save_model` in other flavors.

mlflow/utils/annotations.py
```diff
@@ -63,3 +64,25 @@ def wrapper(*args, **kwargs):
     notice = ".. Note:: This method requires all argument be specified by keyword.\n"
     wrapper.__doc__ = notice + wrapper.__doc__
     return wrapper
+
+
+def deprecate_conda_env(f):
```
harupy, line 69:
```suggestion
def deprecate_conda_env(f):
    """
    A decorator that makes the decorated function raise a deprecation warning when `conda_env` is supplied.
    """
```

dbczumar (commented):
@harupy We should figure out how to handle the fact that the MLflow model server still installs dependencies from `conda.yaml`, rather than `requirements.txt`. If we deprecate `conda_env`, then we need to provide users with a parameter that can be used to control the environment contents referenced at model serving time. 

Because the current MLflow behavior is to log the same set of dependencies to the pip requirements file and the `pip` section of the `conda.yaml` file, it should be sufficient to just tell users to use the new `pip_requirements` and `additional_pip_requirements` parameters. We should make it very clear in the documentation and deprecation warning that these new pip* parameters will also produce a `conda.yaml` file with dependencies in the `pip` section.

harupy:
We decided not to deprecate `conda_env`. See this comment: https://github.com/mlflow/mlflow/pull/4519#discussion_r668415592

PR #4551, opened 2021-07-13
Spark datasource autologging: Improve fidelity of REPL ID / context information

dbczumar:

This PR accomplishes the following goals related to Spark Datasource autologging in Databricks REPL-aware environments:

1. Ensure that Python subscribers resolve a REPL ID reliably, even when `spark.databricks.replId` is not immediately available due to asynchronous population

2. Ensure that a correct (REPL-aware) Java datasource publisher is configured reliably, even when `spark.databricks.replId` is not immediately available due to asynchronous population

## How is this patch tested?

- Included new unit tests
- Manually tested Spark autologging on Databricks in an environment where `dbutils.entry_point.getReplId()` is available. Manual test used the following code:

```
import mlflow.spark
import os
import shutil
from pyspark.sql import SparkSession
df = spark.createDataFrame([
        (4, "spark i j k"),
        (5, "l m n"),
        (6, "spark hadoop spark"),
        (7, "apache hadoop")], ["id", "text"])

import tempfile
csv_dir = os.path.join(tempfile.mkdtemp(), "NOTEBOOK1")
df.write.csv(csv_dir, header=True)

mlflow.spark.autolog()

loaded_df = spark.read.csv(csv_dir, header=True, inferSchema=True)
with mlflow.start_run() as active_run:
  # Call toPandas() to trigger a read of the Spark datasource. Datasource info
  # (path and format) is logged to the current active run, or the
  # next-created MLflow run if no run is currently active
  pandas_df = loaded_df.toPandas()
  import time
  time.sleep(5)
```

and produced the following results:

Screen Shot 2021-07-13 at 12 48 31 AM

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
+++ b/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
@@ -43,10 +43,6 @@ private[autologging] trait MlflowAutologEventPublisherImpl {
       "autologging."))
   }
 
-  // Exposed for testing
-  private[autologging] val databricksConfigProviderClassName =
-    "com.databricks.config.DatabricksClientSettingsProvider"
-
   /**
    * @returns True if Spark is running in a REPL-aware context. False otherwise.
    */
@@ -60,15 +56,14 @@ private[autologging] trait MlflowAutologEventPublisherImpl {
     }
 
     // If the `spark.databricks.replId` is absent, we may still be in a Databricks environment,
-    // which is REPL-aware. To check, we look for the presence of a Databricks-specific
-    // configuration class on the classpath
-    val configProviderFetchResult = Try {
-      Class.forName(databricksConfigProviderClassName);
+    // which is REPL-aware. To check, we look for the presence of a Databricks-specific cluster ID
+    // tag in the Spark configuration
+    val clusterId = spark.conf.getOption("spark.databricks.clusterUsageTags.clusterId")
+    if (clusterId.isDefined) {
+      return true
     }
-    configProviderFetchResult match {
-      case Success(_) => true
-      case Failure(_) => false
-    }
+
+    false
   }
 
   // Exposed for testing
--- a/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
+++ b/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
@@ -44,10 +44,13 @@ private[autologging] trait MlflowAutologEventPublisherImpl {
   }
 
   // Exposed for testing
+  private[autologging] val databricksConfigProviderClassName =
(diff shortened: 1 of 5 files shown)


mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
```diff
@@ -43,14 +44,39 @@ private[autologging] trait MlflowAutologEventPublisherImpl {
...
+    if (replId.isDefined) {
+      return true
+    }
+
+    // If the `spark.databricks.replId` is absent, we may still be in a Databricks environment,
+    // which is REPL-aware. To check, we look for the presence of a Databricks-specific
+    // configuration class on the classpath
+    val configProviderFetchResult = Try {
+      Class.forName(databricksConfigProviderClassName);
+    }
```
dbczumar, line 67: Ideally, we'd check for the presence of `getReplId()` in the Databricks entrypoint Java object. However, it was not immediately clear that this is resolvable from a Databricks Scala REPL. Input would be great here!

dbczumar: cc @WeichenXu123 @smurching

dbczumar: @WeichenXu123 We don't need / should not pass the repl ID to `ReplAwareSparkDataSourceListener`, since the listener needs to pull REPL ID information from Spark event properties. The listener fetches events from multiple REPLs.

smurching: I'll wait for Bago's confirmation but I think it makes sense that the entrypoint is difficult to access from Scala code here - the entrypoint is a Python-REPL-specific interface. Another way to check that we're in Databricks could be to check for the `spark.databricks.clusterUsageTags.clusterId` property (`spark.conf.get("spark.databricks.clusterUsageTags.clusterId")`), which I believe is set prior to cluster startup (it's how we fetch cluster ID to set as a tag for the "Reproduce Run" feature), but we should check/confirm. It may be more stable than relying on a particularly-named class - either way I guess we can add a test for the API we end up relying on to detect that we're in Databricks from this Scala code

smurching (approved):
Generally LGTM! Had a minor comment around how we check that we're in a Databricks cluster, but I think this is fine to merge

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala
dbczumar: Thanks for the suggestion @smurching ! I've gone ahead and replaced the logic that searches for `com.databricks.config.DatabricksClientSettingsProvider` with the proposed logic for fetching the `spark.databricks.clusterUsageTags.clusterId` conf. I've also updated tests accordingly.

smurching (approved):
LGTM!

PR #4575, opened 2021-07-19
Add `_get_default_pip_requirements` to all flavors

harupy:

Add the `_get_default_pip_requirements` fucntion to all flavors (except `xgboost`).

## How is this patch tested?

Existing checks.

Components 

Interface 

Language 

Integrations


mlflow/catboost.py
```diff
@@ -39,17 +39,18 @@
...
     :return: The default Conda environment for MLflow Models produced by calls to
              :func:`save_model()` and :func:`log_model()`.
     """
-    import catboost as cb
-
-    return _mlflow_conda_env(
-        # CatBoost is not yet available via the default conda channels, so we install it via pip
-        additional_pip_deps=["catboost=={}".format(cb.__version__)]
-    )
+    return _mlflow_conda_env(additional_pip_deps=_get_default_pip_requirements())
```
harupy, line 53: What this PR does is extract the code to prepare pip requirements from `get_default_conda_env` and redefine it as `_get_default_pip_requirements`. That's it.

mlflow/fastai.py
```diff
@@ -43,6 +43,18 @@
 FLAVOR_NAME = "fastai"
 
 
+def _get_default_pip_requirements(include_cloudpickle=False):
```
harupy, line 46:
```suggestion
def _get_default_pip_requirements(include_cloudpickle=False):
     """
     Returns a default set of pip requirements of this flavor.
     """
```

harupy: docstring may clarify what this function returns.

dbczumar (commented):
@harupy Do we need to do this if we're going to start inferring requirements?

harupy:
@dbczumar I think we do. We can fallback to what `_get_default_pip_requirements` returns when the inference procedure fails like this:

https://github.com/mlflow/mlflow/pull/4518/files#diff-f6bd607049d4969297151a812f447bdc13453a856df017983f7ea81470d25bd0R191-R193

```python
        default_pip_requirements = infer_pip_requirements(
            path, FLAVOR_NAME, fallback=_get_default_pip_requirements()
        )
```

mlflow/catboost.py
```diff
@@ -39,17 +39,18 @@
 _MODEL_BINARY_FILE_NAME = "model.cb"
 
 
+def _get_default_pip_requirements():
```
dbczumar, line 42: shall we make this a public function that users can reference? (and make a similar change for other flavors)

mlflow/fastai.py
dbczumar: How about 


```
:return: A list of default pip requirements for MLflow Models produced by this flavor.
             Calls to :func:`save_model()` and :func:`log_model()` produce a pip environment
             that, at minimum, contains these requirements.
```

dbczumar: When we implement pip dependency inference, can we also update `get_default_conda_env()` to indicate that the default conda env is a subset of the full set of dependencies generated by `save_model()` and `log_model()`? - we can do this as part of a follow-up PR.

dbczumar (approved):
LGTM with a couple of small comments!

mlflow/catboost.py
harupy: sure!

mlflow/fastai.py
harupy: > How about ...

Sounds good!

harupy: got it!

PR #4631, opened 2021-08-01
Fix some issues in gorillar patch

WeichenXu123:

- Modify `get_original_attribute` logic, search from children classes to parent classes,
  and for each class check "_gorilla_original_{attr_name}" attribute first.
  first. This will ensure get the correct original attribute in any cases, e.g.,
  the case some classes in the hierarchy haven't been patched, but some others are
  patched, this case the previous code is risky to get wrong original attribute.
- Make `get_original_attribute` support bypassing descriptor protocol.
- remove `get_attribute` method, use `get_original_attribute` with
  `bypass_descriptor_protocol=True` instead of calling it.
- After reverting patch, there will be no side-effect, restore object to be exactly the
  original status.
- Remove `create_patches` and `patches` methods.

## How is this patch tested?

Unit tests.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/autologging_utils/__init__.py
+++ b/mlflow/utils/autologging_utils/__init__.py
@@ -376,11 +376,12 @@ def autologging_integration(name):
             except Exception:
                 pass
 
+            revert_patches(name)
+
             # If disabling autologging using fluent api, then every active integration's autolog
             # needs to be called with disable=True. So do not short circuit and let
             # `mlflow.autolog()` invoke all active integrations with disable=True.
             if name != "mlflow" and get_autologging_config(name, "disable", True):
-                revert_patches(name)
                 return
 
             is_silent_mode = get_autologging_config(name, "silent", False)
--- a/mlflow/utils/gorilla.py
+++ b/mlflow/utils/gorilla.py
@@ -357,17 +357,13 @@ def revert(patch):
             % (patch.destination.__name__,)
         )
 
-    original = get_original_attribute(
-        patch.destination, patch.name, bypass_descriptor_protocal=True)
+    # get original attribute bypassing descriptor protocal
+    original = object.__getattribute__(patch.destination, original_name)
 
     if patch.is_inplace_patch:
         setattr(patch.destination, patch.name, original)
 
-    # We are only deleting the attribute if it is defined on the
-    # destination class as opposed to being inherited from parent class.
-    if original_name in patch.destination.__dict__:
-        delattr(patch.destination, original_name)
-
+    delattr(patch.destination, original_name)
     delattr(patch.destination, curr_active_patch)
 
 
--- a/mlflow/utils/gorilla.py
+++ b/mlflow/utils/gorilla.py
@@ -41,7 +41,6 @@ import logging
 import pkgutil
 import sys
 import types
-import warnings
 
 __version__ = "0.3.0"
 _logger = logging.getLogger(__name__)
--- a/mlflow/utils/gorilla.py
+++ b/mlflow/utils/gorilla.py
@@ -349,7 +349,8 @@ def revert(patch):
     # then there is no patching exist.
     curr_active_patch = _ACTIVE_PATCH % (patch.name,)
     if curr_active_patch not in patch.destination.__dict__:
-        raise RuntimeError(f"The patch does not exist or has been reverted.")
+        # already reverted.
+        return
 
     original_name = _ORIGINAL_NAME % (patch.name,)
     if original_name not in patch.destination.__dict__:
--- a/mlflow/utils/gorilla.py
+++ b/mlflow/utils/gorilla.py
@@ -297,14 +297,16 @@ def apply(patch):
     if curr_active_patch in patch.destination.__dict__:
         # safe guarding checking.
         raise RuntimeError(
-            f'Patch {patch.name} on {destination.__name__} already existed.'
-            f'Revert old patch first if you want to patch it again.')
+            f"Patch {patch.name} on {destination.__name__} already existed."
+            f"Revert old patch first if you want to patch it again."
+        )
 
     # When a hit occurs due to an attribute at the destination already existing
     # with the patch's name, the existing attribute is referred to as 'target'.
     try:
         target = get_original_attribute(
-            patch.destination, patch.name, bypass_descriptor_protocal=True)
+            patch.destination, patch.name, bypass_descriptor_protocal=True
+        )
     except AttributeError:
         pass
     else:
@@ -347,7 +349,7 @@ def revert(patch):
     # then there is no patching exist.
     curr_active_patch = _ACTIVE_PATCH % (patch.name,)
     if curr_active_patch not in patch.destination.__dict__:
-        raise RuntimeError(f'The patch does not exist or has been reverted.')
+        raise RuntimeError(f"The patch does not exist or has been reverted.")
 
     original_name = _ORIGINAL_NAME % (patch.name,)
     if original_name not in patch.destination.__dict__:
@@ -361,7 +363,11 @@ def revert(patch):
     original = object.__getattribute__(patch.destination, original_name)
 
     if patch.is_inplace_patch:
+        # restore original method
         setattr(patch.destination, patch.name, original)
+    else:
+        # delete patched method
+        delattr(patch.destination, patch.name)
 
     delattr(patch.destination, original_name)
     delattr(patch.destination, curr_active_patch)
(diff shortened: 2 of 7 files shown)


mlflow/utils/autologging_utils/__init__.py
```diff
@@ -376,11 +376,12 @@ def autolog(*args, **kwargs):
             except Exception:
                 pass
 
+            revert_patches(name)
```
WeichenXu123, line 379: Before setup new patches, we should revert all existing patches first. different autolog config may setup different patches, if we don't revert existing patches first, it  will make things messy.

mlflow/utils/gorilla.py
```diff
@@ -292,10 +293,20 @@ def apply(patch):
...
+            f"Revert old patch first if you want to patch it again."
+        )
+
     # When a hit occurs due to an attribute at the destination already existing
     # with the patch's name, the existing attribute is referred to as 'target'.
     try:
-        target = get_attribute(patch.destination, patch.name)
+        target = get_original_attribute(
+            patch.destination, patch.name, bypass_descriptor_protocal=True
+        )
```
WeichenXu123, line 309: This line changing address one bug: make patch on top of patch. Suppose the case:
```
class A:
    def predict(self):
        pass

class B(A):
    pass
```

Suppose we patch `A.predict` first, and then patch `B.predict`, when we do patching `B.predict`, the previous code `get_attribute(patch.destination, patch.name)` will get the patched `A.predict` method, so it will generate a new patch method on top of patched `A.predict`

I modify the `gorilla.get_original_attribute` impl to make it support class hierarchy correctly, and also make it support bypassing/not bypassing descriptor protocal.

mlflow/utils/gorilla.py
```diff
@@ -292,10 +293,20 @@ def apply(patch):
     """
     settings = Settings() if patch.settings is None else patch.settings
 
+    curr_active_patch = _ACTIVE_PATCH % (patch.name,)
+    if curr_active_patch in patch.destination.__dict__:
+        # safe guarding checking.
+        raise RuntimeError(
+            f"Patch {patch.name} on {destination.__name__} already existed."
+            f"Revert old patch first if you want to patch it again."
+        )
```
WeichenXu123, line 302: Fast fail if it already patched. Fast fail will help debug and prevent unpredictable bug.

mlflow/utils/gorilla.py
```diff
@@ -312,25 +323,7 @@ def apply(patch):
...
-            # (e.g., `sklearn.feature_extraction.text.CountVectorizer.fit_transform()`)
-            prev_patch = getattr(patch.destination, curr_active_patch, None)
-
-            if not hasattr(patch.destination, original_name) or (
-                prev_patch
-                and prev_patch.destination != patch.destination
-                and issubclass(patch.destination, prev_patch.destination)
-            ):
-                setattr(patch.destination, original_name, target)
+            setattr(patch.destination, original_name, target)
```
WeichenXu123, line 326: Simplify code logic here.

WeichenXu123, line 326: Simplify code logic here.

mlflow/utils/gorilla.py
```diff
@@ -353,35 +346,31 @@ def revert(patch):
...
-        delattr(patch.destination, curr_active_patch)
+    # get original attribute bypassing descriptor protocal
+    original = object.__getattribute__(patch.destination, original_name)
+
+    if patch.is_inplace_patch:
+        # restore original method
+        setattr(patch.destination, patch.name, original)
+    else:
+        # delete patched method
+        delattr(patch.destination, patch.name)
```
WeichenXu123, line 370: Update revert logic to avoid side effect after reverting patch.

Suppose the case:
```
class A:
    def predict(self):
        pass

class B(A):
    pass
```

When we patch `B.predict`, `B.__dict__` was appended a new attribute `predict`, but after reverting the patch, the new attribute `predict` still exists in `B.__dict__`, this is a risky side effect.

mlflow/utils/gorilla.py
```diff
@@ -788,7 +779,30 @@ def get_original_attribute(obj, name):
...
+
+    if inspect.isclass(obj):
+        for obj_ in inspect.getmro(obj):
+            if original_name in obj_.__dict__:
+                return _get_attr(obj_, original_name)
+            elif name in obj_.__dict__:
+                return _get_attr(obj_, name)
+            else:
+                # go on checking parent classes
+                continue
```
WeichenXu123, line 799: Search from children classes to parent classes, and check "original_name" attribute first. This will ensure get the correct original attribute in any cases (e.g., the case some classes in the hierarchy haven't been patched, but some others are patched, this case the previous code is risky to get wrong original attribute)

mlflow/utils/gorilla.py
WeichenXu123: I change this to a warning. raise error here breaks many tests.

mlflow/utils/gorilla.py
harupy: Can we create a test to verify this?

dbczumar (commented):
@WeichenXu123 This is fantastic - thanks for these fixes!

Can we create some unit tests for this behavior? A new file like `tests/utils/test_gorilla.py` makes sense here.

Can we also add a short note at the top of the `mlflow.utils.gorilla` module to indicate that this module has fixes / adaptations for MLflow use cases that make it different from the original gorilla library?

tests/tracking/fluent/test_fluent_autolog.py
```diff
@@ -291,6 +291,17 @@ def test_autolog_obeys_silent_mode(
     stream = StringIO()
     sys.stderr = stream
 
+    if (
```
dbczumar, line 294: Reminder to remove this before merging

mlflow/utils/autologging_utils/__init__.py
dbczumar: Makes sense! Can we add a test case for something that would fail if we didn't have this fix implemented?

mlflow/utils/gorilla.py
```diff
@@ -312,27 +323,10 @@ def apply(patch):
...
-                prev_patch
-                and prev_patch.destination != patch.destination
-                and issubclass(patch.destination, prev_patch.destination)
-            ):
-                setattr(patch.destination, original_name, target)
-
-    setattr(patch.destination, patch.name, patch.obj)
+            setattr(patch.destination, original_name, target)
+
+    patch_obj = object.__getattribute__(patch, "obj")  # bypass descriptor protocol
```
dbczumar, line 328: is this necessary? We're not subclassing `Patch` anywhere, right?

mlflow/utils/gorilla.py
```diff
@@ -353,35 +347,36 @@ def revert(patch):
...
 
-    original_name = _ORIGINAL_NAME % (patch.name,)
-    if not getattr(patch.destination, original_name, None):
-        return
+    # during reverting patch, we need restore the raw attribute to the patch point
+    # so get original attribute bypassing descriptor protocal
+    original = object.__getattribute__(patch.destination, original_name)
 
-    setattr(patch.destination, patch.name, original)
+    if patch.is_inplace_patch:
```
dbczumar, line 371: If the patch is an inplace patch, then won't the attribute with name `original_name` always be absent from the object referred to by the patch?

I.e. if `is_inplace_patch == True`, then `original_name not in patch.destination.__dict__`, right?

If that's true, then we should modify the condition on line 360 to only raise a runtime error if `original_name not in patch.destination.__dict__ and not is_inplace_patch`. Can we also add a test case for this?

tests/utils/test_gorilla.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env python
```
dbczumar, line 1:
```suggestion
```

mlflow/utils/gorilla.py
WeichenXu123: No. If patched, then always `original_name in patch.destination.__dict__`.  "is_inplace_patch" only means replace the existing method with the patched method, but it will still store the original method to the "original_name" attribute.

e.g.
```
class A:
    def predict(self):
        pass
```

We patch `A.predict`, then it is an in-place patch, and after patch, `A.predict` point to the patched_predict method, and `A._gorilla_original_predict` point to the original `A.predict`

WeichenXu123: Maybe "is_inplace_patch" name is bad, use the name "is_existing_method_replaced_by_patch" looks better ?

dbczumar: Are you sure? `original_name` should only be set when there's a hit and `store_hit = True` is passed to `apply()`. https://github.com/mlflow/mlflow/blob/66b4116b6afb66561b81554a1303ef9f11b1adc1/mlflow/utils/gorilla.py#L324-L326

dbczumar (approved):
LGTM! Thanks @WeichenXu123  - great work!

PR #4627, opened 2021-07-30
Introdcue `_is_available_on_pypi` and use it to determine whether we need to use `--no-conda` for `pyfunc_serve_and_score_model`

harupy:
Signed-off-by: harupy <hkawamura0130@gmail.com>

## What changes are proposed in this pull request?

Introdcue `_is_available_on_pypi` and use it to determine whether we need to use --no-conda for `pyfunc_serve_and_score_model`.

## Motivation

`_is_available_on_pypi` allows us to do the following:

- If we know conda will fail to create an env due to packages / versions that don't exist on PyPI, use `--no-conda` to ensure the env can be created.
- Otherwise, do no use `--no-conda` to encounter a did-not-find-matching-version error.

This PR should make it easier to test the auto dependency inference proposed in https://github.com/mlflow/mlflow/pull/4518.

## How is this patch tested?

Existing tests

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
[...]

PR #4632, opened 2021-08-02
Add `_get_pinned_requirement` to accurately obtain package versions

harupy:

See https://github.com/mlflow/mlflow/pull/4627#discussion_r680596126

## How is this patch tested?

Unit test

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pytorch/__init__.py
+++ b/mlflow/pytorch/__init__.py
@@ -40,7 +40,7 @@ from mlflow.utils.environment import (
     _REQUIREMENTS_FILE_NAME,
     _CONSTRAINTS_FILE_NAME,
 )
-from mlflow.utils.requirements_utils import _strip_local_version_identifier
+from mlflow.utils.requirements_utils import _get_pinned_requirement
 from mlflow.utils.docstring_utils import format_docstring, LOG_MODEL_PARAM_DOCS
 from mlflow.utils.file_utils import _copy_file_or_tree, TempDir, write_to
 from mlflow.utils.model_utils import _get_flavor_configuration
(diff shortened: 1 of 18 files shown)


dbczumar (approved):
LGTM! Nice catch, @harupy !

harupy:
@dbczumar Thanks for the review! I'll merge this PR and file a follow-up PR to use `_get_pinned_requirement` in other flavors!

mlflow/pytorch/__init__.py
```diff
@@ -67,8 +68,10 @@ def get_default_pip_requirements():
     import torchvision
 
     return [
-        "torch=={}".format(torch.__version__),
-        "torchvision=={}".format(torchvision.__version__),
+        # `torch.__version__` and `torchvision.__version__`contain a local version identifier
+        # (e.g. "cpu" in "1.9.0+cpu") which causes the installation from PyPI to fail.
+        "torch=={}".format(_strip_local_version_identifier(torch.__version__)),
+        "torchvision=={}".format(_strip_local_version_identifier(torchvision.__version__)),
```
harupy, line 74: I've switched to using `_strip_local_version_identifier` because `import torch; torch.__version__` is more robust than `importlib_metadata.version` which may fails when a user

harupy, line 74: @dbczumar 
I think `import torch; torch.__version__` is more stable than `importlib_metadata.version`. For example, we could download torch and add its path to `sys.path` (we rarely do this). In this case, `importlib_metadata.version` doesn't work, but `import torch; torch.__version__` does.

dbczumar: Other libraries may add local version identifiers in the future, so can we pass all of our version strings through a common function (like you had before) that performs `_strip_local_version_identifier` and can later be extended to perform additional operations?

harupy: Got it, I'll define `_get_pinned_requirement` (like I did) and run `_strip_local_version_identifier` in it.

harupy: Done in https://github.com/mlflow/mlflow/pull/4632/commits/5565b54bb6f04d7cb7ebb71d3b0e55c152befed0

dbczumar (approved):
LGTM again! Thanks @harupy !

harupy:
Thanks! I'll update other flavors to use `_get_pinned_requirement`.

Issue #4489, opened 2021-06-21
[BUG] Tracking: DB connection are not released when using log_param repeatedly


sebchr:
### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [x] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: yes
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 10 Version 10.0.19041 Build 19041
- **MLflow installed from (source or binary)**: binary
- **MLflow version (run ``mlflow --version``)**: mlflow, version 1.18.0
- **Python version**: Python 3.8.10
- **npm version, if running the dev UI**:
- **Exact command to reproduce**:

### Describe the problem
After logging ~80 parameters to the postgres tracking database, the number of available connections is exhausted, and I get the following error:
```
2021/06/21 14:21:00 WARNING mlflow.store.db.utils: SQLAlchemy engine could not be created. The following exception is caught.
(psycopg2.OperationalError) FATAL:  too many connections for role "zibvebrgldjyaa"

(Background on this error at: http://sqlalche.me/e/14/e3q8)
Operation will be retried in 0.1 seconds
```
Only 20 connections are available because I'm using a [Heroku Free-tier postgres database](https://elements.heroku.com/addons/heroku-postgresql). I'm the only one accessing the database and using `select * from pg_stat_activity where datname = 'dbname'` I can confirm that all connections are created by mlflow.

  I expected mlflow would use a constant and low number of connections. I did not see any improvement when setting `MLFLOW_SQLALCHEMYSTORE_POOL_SIZE` and `MLFLOW_SQLALCHEMYSTORE_MAX_OVERFLOW` environment variables.

The  version of the postgres database is: `PostgreSQL 13.3 (Ubuntu 13.3-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit`

### Code to reproduce issue
Minimal example:
```python
import logging
import mlflow

logging.basicConfig(level=logging.INFO)

if __name__ == "__main__":
[...]

PR #4663, opened 2021-08-05
Fixed SQLAlchemy backend not releasing DB connections by using a cache of SQLAlchemy Engines instead of creating new ones each time.

arpitjasa-db:
Updated SQLAlchemyStore to use and maintain a dictionary of SQLAlchemyEngine mapped by their database URI. This prevents multiple engines being created for the same database, exhausting its connections, but preserving functionality otherwise.

## How is this patch tested?
Unit testing as well as end-to-end testing using a local PostgreSQL database installation.

Updated SQLAlchemyStore to use and maintain a dictionary of SQLAlchemyEngine mapped by their database URI. This prevents multiple engines being created for the same database, exhausting its connections, but preserving functionality otherwise: https://github.com/mlflow/mlflow/issues/4489.

Components 

Interface 
- [X] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry

Language 

Integrations

<a name="rn/bug-fix"></a>

### DIFF
--- a/mlflow/store/db/utils.py
+++ b/mlflow/store/db/utils.py
@@ -15,7 +15,6 @@ from mlflow.store.db.db_types import SQLITE
 
 _logger = logging.getLogger(__name__)
 
-
 MLFLOW_SQLALCHEMYSTORE_POOL_SIZE = "MLFLOW_SQLALCHEMYSTORE_POOL_SIZE"
 MLFLOW_SQLALCHEMYSTORE_MAX_OVERFLOW = "MLFLOW_SQLALCHEMYSTORE_MAX_OVERFLOW"
 MAX_RETRY_COUNT = 15
@@ -184,7 +183,9 @@ def create_sqlalchemy_engine(db_uri):
         pool_kwargs["pool_size"] = int(pool_size)
     if pool_max_overflow:
         pool_kwargs["max_overflow"] = int(pool_max_overflow)
-    if pool_kwargs:
+    if pool_kwargs:  # If pool parameters are specified, create the SQLAlchemy engine with a QueuePool.
         _logger.info("Create SQLAlchemy engine with pool options %s", pool_kwargs)
         return sqlalchemy.create_engine(db_uri, pool_pre_ping=True, **pool_kwargs)
+    # Otherwise, create the SQLAlchemy engine with a NullPool to prevent connection pooling
+    # so that closing sessions will also close the connections.
     return sqlalchemy.create_engine(db_uri, poolclass=sqlalchemy.pool.NullPool)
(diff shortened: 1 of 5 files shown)


mlflow/store/db/utils.py
```diff
@@ -186,4 +186,5 @@ def create_sqlalchemy_engine(db_uri):
         pool_kwargs["max_overflow"] = int(pool_max_overflow)
     if pool_kwargs:
         _logger.info("Create SQLAlchemy engine with pool options %s", pool_kwargs)
-    return sqlalchemy.create_engine(db_uri, pool_pre_ping=True, **pool_kwargs)
+        return sqlalchemy.create_engine(db_uri, pool_pre_ping=True, **pool_kwargs)
+    return sqlalchemy.create_engine(db_uri, poolclass=sqlalchemy.pool.NullPool)
```
dbczumar, line 190: Can we add an inline comment explaining why we do this?

dbczumar (approved):
LGTM! Thanks @arpitjasa-db !

apurva-koti (approved):
Looks great thanks @arpitjasa-db !

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -101,7 +104,19 @@ def __init__(self, db_uri, default_artifact_root):
...
+                # Repeat check to prevent race conditions where one thread checks for an existing
+                # engine while another is creating the respective one, resulting in multiple
+                # engines being created. It isn't combined with the above check to prevent
+                # inefficiency from multiple threads waiting for the lock to check for engine
+                # existence if it has already been created.
+                if db_uri not in SqlAlchemyStore._db_uri_sql_alchemy_engine_map:
+                    SqlAlchemyStore._db_uri_sql_alchemy_engine_map[
+                        db_uri
+                    ] = mlflow.store.db.utils.create_sqlalchemy_engine_with_retry(db_uri)
+        self.engine = SqlAlchemyStore._db_uri_sql_alchemy_engine_map[db_uri]
```
apurva-koti, line 119: Do we want to move this assignment to the lock as well? Could we have a situation in which two threads encounter this block of code, but by the time one thread acquires the lock, creates the engine and assigns it, the other one is still acquiring the lock (because to that thread, `db_uri` is not yet in map)?

arpitjasa-db: @dbczumar and I talked about this where moving assignment inside of the lock could significantly slow down the threads since they each have to wait for the lock now. That's the purpose of the first check, which decreases the number of race conditions that could result in duplicate engines rather than eliminate. What do you think?

apurva-koti (approved):
Thanks for the fix :)

mlflow/store/tracking/sqlalchemy_store.py
apurva-koti: Discussed offline. One extra engine probably isn't a problem :)

dbczumar: @apurva-koti It could be a problem if the pool size is large. I'd rather keep the locking here, since it doesn't add much complexity.

dbczumar (approved):
LGTM! Thanks @arpitjasa-db !

dbczumar:
R Docs CI check seems to be transiently broken. Merging since all other tests pass! Thanks @arpitjasa-db !

PR #4682, opened 2021-08-11
Always use `importlib_metadata.version` when obtaining package version to prevent version mismatch between `importlib_metadata.version` and `<package>.__version__`

harupy:

Always use `importlib_metadata.version` when obtaining a package version to prevent version mismatch between `importlib_metadata.version` (used in auto dependency inference) and `<package>.__version__` (used in `get_default_pip_requirements`).

## How is this patch tested?

Existing tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/requirements_utils.py
+++ b/mlflow/utils/requirements_utils.py
@@ -12,12 +12,15 @@ import pkg_resources
 import importlib_metadata
 from itertools import filterfalse, chain
 from collections import namedtuple
+import logging
 
 from mlflow.exceptions import MlflowException
 from mlflow.tracking.artifact_utils import _download_artifact_from_uri
 from mlflow.utils.autologging_utils.versioning import _strip_dev_version_suffix
 from mlflow.utils.databricks_utils import is_in_databricks_runtime
 from packaging.version import Version, InvalidVersion
+
+_logger = logging.getLogger(__name__)
 
 
 def _is_comment(line):
@@ -255,12 +258,24 @@ def _infer_requirements(model_uri, flavor):
         "mlflow",
     ]
     packages = _prune_packages(packages) - set(excluded_packages)
-    return ["{}=={}".format(p, _get_installed_version(p)) for p in sorted(packages)]
+    return sorted(map(_get_pinned_requirement, packages))
 
 
-def _strip_local_version_identifier(version):
+def _get_local_version_label(version):
     """
-    Strips a local version identifer in `version`.
+    Extracts a local version label in `version`.
+
+    :param version: A version string.
+    """
+    try:
+        return Version(version).local
+    except InvalidVersion:
+        return None
+
+
+def _strip_local_version_label(version):
+    """
+    Strips a local version label in `version`.
 
     Local version identifiers:
     https://www.python.org/dev/peps/pep-0440/#local-version-identifiers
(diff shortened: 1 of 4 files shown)


mlflow/utils/requirements_utils.py
```diff
@@ -180,17 +180,30 @@ def _run_command(cmd):
...
-
     # In Databricks, strip a dev version suffix for pyspark (e.g. '3.1.2.dev0' -> '3.1.2')
     # and make it installable from PyPI.
     if package == "pyspark" and is_in_databricks_runtime():
-        version = _strip_dev_version_suffix(version)
-    return version
+        return _strip_dev_version_suffix(__import__("pyspark").__version__)
+
+    try:
+        return importlib_metadata.version(package)
```
harupy, line 194: If a user installs `torch` with:

```
pip install torch==1.8.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```

This line returns `1.8.0`.

mlflow/utils/requirements_utils.py
```diff
@@ -180,16 +180,31 @@ def _run_command(cmd):
...
-def _get_package_version_from_metadata(package):
+def _get_installed_version(package, module=None):
     """
-    Obtains the version of the specified package from its metadata.
+    Obtains the package version using `importlib_metadata.version`. If it fails, use
+    `__import__(module or package).__version__`.
     """
-    version = importlib_metadata.version(package)
+    try:
+        version = importlib_metadata.version(package)
```
harupy, line 189: If a user installs `torch` with:

```
pip install torch==1.8.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```

`importlib_metadata.version("torch")` returns `1.8.0`.

harupy: Or we could just log `torch==1.8.0+cpu`. If the user wants to specify `--find-links`,  he/she can use `pip_requirements`.

mlflow/utils/requirements_utils.py
```diff
@@ -180,16 +180,31 @@ def _run_command(cmd):
...
+        # `__import__(package).__version__`. See the example for pytorch below.
+        #
+        # Example
+        # -------
+        # $ pip install torch==1.9.0
+        # $ python -c "import torch; print(torch.__version__)"
+        # 1.9.0+cu102
+        # $ python -c "import importlib_metadata; print(importlib_metadata.version('torch'))"
+        # 1.9.0
+        version = __import__(module or package).__version__
```
harupy, line 201: The best effort to capture a package version.

dbczumar (commented):
@harupy Do you have an example of the mismatch? Would help to understand its significance.

harupy:
@dbczumar Sure!

```bash
% python -c "import mxnet; print(mxnet.__version__)"
1.8.0
                                                                                                                                                                        
% python -c "import importlib_metadata; print(importlib_metadata.version('mxnet'))"
1.8.0.post0
```

I think the same thing happens for `torch`.

mlflow/utils/requirements_utils.py
dbczumar: @harupy Ideally, we'd like to maximize the likelihood of serving succeeding. Can we ignore local version identifiers and print a warning indicating that the local version identifier was dropped & instructing customers to use `pip_requirements` or `conda_env` to specify the distribution containing the local version if necessary?

harupy: Ok, I'll update the code to drop local version identifiers and raise a warning for that.

mlflow/utils/requirements_utils.py
```diff
@@ -240,12 +258,24 @@ def _infer_requirements(model_uri, flavor):
...
+    :param version: A version string.
+    """
+    try:
+        return Version(version).local
+    except InvalidVersion:
+        return None
 
 
-def _strip_local_version_identifier(version):
+def _strip_local_version_label(version):
```
harupy, line 276: Renamed `_strip_local_version_identifier` to `_strip_local_version_label`. The following code describes why:

```
1.2.3+abc
     ^^^^ local version label (segment starting with '+')
^^^^^^^^^ local version identifier (version identifier containing a local version label)
```

dbczumar (approved):
LGTM! Thanks @harupy !

PR #4686, opened 2021-08-11
Pyspark ml autolog post-training metric

WeichenXu123:

Pyspark ml autolog post-training metric

## How is this patch tested?

Unit tests.

Components 

Interface 

Language 

Integrations


mlflow/pyspark/ml/__init__.py
```diff
@@ -427,6 +438,189 @@ def _log_estimator_params(param_map):
...
+
+    def should_log_post_training_metrics(self):
+        """
+        Check whether we should run patching code for autologging post training metrics.
+        This checking should surround the whole patched code due to the safe guard checking,
+        See following note.
+
+        Note: It includes checking `_SklearnTrainingSession.is_active()`, This is a safe guarding
+        for meta-estimator (e.g. GridSearchCV) case:
+          running GridSearchCV.fit, the nested `estimator.fit` will be called in parallel,
```
dbczumar, line 457: I think we need to update this documentation to refer to `_SparkTrainingSession` and a pyspark model type.

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -568,3 +568,12 @@ def get_instance_method_first_arg_value(method, call_pos_args, call_kwargs):
             inspect.Parameter.VAR_POSITIONAL,
         ]
         return call_kwargs.get(first_arg_name)
+
+
+def get_instance_method_arg_value(arg_index, arg_name, default_value, call_pos_args, call_kwargs):
```
dbczumar, line 573: What makes this specific to instance methods? Couldn't this apply to any function?

dbczumar: Can we add docs?

mlflow/pyspark/ml/__init__.py
```diff
@@ -427,6 +438,189 @@ def _log_estimator_params(param_map):
...
+          but, the _autolog_training_status is a global status without thread-safe lock protecting.
+          This safe guarding will prevent code run into this case.
```
dbczumar, line 459: `_autolog_training_status` refers to an instance of `_AutologgingMetricsManager` right? We should update the docs here & for sklearn if necessary

mlflow/pyspark/ml/__init__.py
```diff
@@ -643,19 +834,73 @@ def fit_mlflow(original, self, *args, **kwargs):
...
-            estimator = self.copy(params)
+            estimator = self.copy(params) if params is not None else self
             _log_pretraining_metadata(estimator, params)
             spark_model = original(self, *args, **kwargs)
             _log_posttraining_metadata(estimator, spark_model, params)
             return spark_model
 
     def patched_fit(original, self, *args, **kwargs):
+
+        status = _get_autologging_metrics_manager()
```
dbczumar, line 845:
```suggestion
        metrics_manager = _get_autologging_metrics_manager()
```

dbczumar: Can we make this change elsewhere in this file and for `sklearn` if necessary?

mlflow/pyspark/ml/__init__.py
```diff
@@ -643,19 +834,73 @@ def fit_mlflow(original, self, *args, **kwargs):
...
+        if status.should_log_post_training_metrics():
+            with status.disable_log_post_training_metrics():
+                metric = original(self, *args, **kwargs)
+
+            if status.is_metric_value_loggable(metric):
+
+                metric_name = self.getMetricName()
+
+                params = get_instance_method_arg_value(1, "params", None, args, kwargs)
+                evaluator = self.copy(params) if params is not None else self
```
dbczumar, line 880: Can you remind me why we perform this copying after producing the eval metric? Perhaps an inline comment explaining why this is necessary would help.

dbczumar (commented):
@WeichenXu123 Looks great! Just left some small questions / docs comments. Looking forward to taking a look at test cases. Awesome work!

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -568,3 +568,12 @@ def get_instance_method_first_arg_value(method, call_pos_args, call_kwargs):
...
+    if arg_name in call_kwargs:
+        return call_kwargs[arg_name]
+    elif arg_index < len(call_pos_args):
+        return call_pos_args[arg_index]
+    else:
+        return default_value
```
harupy, line 579: Is there any case where we reach this branch?

mlflow/pyspark/ml/__init__.py
```diff
@@ -427,6 +438,189 @@ def _log_estimator_params(param_map):
...
+
+_autologging_metrics_manager = _AutologgingMetricsManager()
+
+
+def _get_autologging_metrics_manager():
+    """
+    Get the global `_AutologgingMetricsManager` instance which holds information used in
+    post-training metric autologging. See doc of class `_AutologgingMetricsManager` for details.
+    """
+    return _autologging_metrics_manager
```
harupy, line 621: btw what is the purpose of this function? Can we just refer to `_autologging_metrics_manager` or am I missing something?

mlflow/pyspark/ml/__init__.py
harupy: https://spark.apache.org/docs/latest/api/python/_modules/pyspark/ml/param.html#Params.copy

```python
    def copy(self, extra=None):
        """
        Creates a copy of this instance with the same uid and some
        extra params. The default implementation creates a
        shallow copy using :py:func:`copy.copy`, and then copies the
        embedded and extra parameters over and returns the copy.
        Subclasses should override this method if the default approach
        is not sufficient.

        Parameters
        ----------
        extra : dict, optional
            Extra parameters to copy to the new instance

        Returns
        -------
        :py:class:`Params`
            Copy of this instance
        """
        if extra is None:
            extra = dict()
        that = copy.copy(self)
        that._paramMap = {}
        that._defaultParamMap = {}
        return self._copyValues(that, extra)
```

mlflow/pyspark/ml/__init__.py
```diff
@@ -643,19 +834,73 @@ def fit_mlflow(original, self, *args, **kwargs):
             _logger.warning(_get_warning_msg_for_fit_call_with_a_list_of_params(self))
             return original(self, *args, **kwargs)
         else:
-            estimator = self.copy(params)
+            estimator = self.copy(params) if params is not None else self
```
harupy, line 837: https://spark.apache.org/docs/latest/api/python/_modules/pyspark/ml/param.html#Params.copy

```python
    def copy(self, extra=None):
        """
        Creates a copy of this instance with the same uid and some
        extra params. The default implementation creates a
        shallow copy using :py:func:`copy.copy`, and then copies the
        embedded and extra parameters over and returns the copy.
        Subclasses should override this method if the default approach
        is not sufficient.

        Parameters
        ----------
        extra : dict, optional
            Extra parameters to copy to the new instance

        Returns
        -------
        :py:class:`Params`
            Copy of this instance
        """
        if extra is None:
            extra = dict()
        that = copy.copy(self)
        that._paramMap = {}
        that._defaultParamMap = {}
        return self._copyValues(that, extra)
```

`copy` seems to work for `None` or old versions of pyspark have different behaviors?

harupy: https://github.com/apache/spark/blame/44b7931936d9eff4d8f3054abdef3363af26afb6/python/pyspark/ml/param/__init__.py#L373

mlflow/pyspark/ml/__init__.py
harupy: `variable = _get_autologging_metrics_manager` looks like we create a **new** metric manager object.

mlflow/pyspark/ml/__init__.py
WeichenXu123: yes. but if params is None we don't need do the `copy.copy(self)` step. This is a small optimization.

mlflow/pyspark/ml/__init__.py
WeichenXu123: > Can we just refer to _autologging_metrics_manager 

Yes we can.

mlflow/utils/autologging_utils/__init__.py
WeichenXu123: this means using default value. e.g.

def f1(a=3):
  pass

f1() # then call_pos_args, call_kwargs are both empty, go into this branch and return default value 3

harupy: You're right. I think I misunderstood something.

mlflow/pyspark/ml/__init__.py
```diff
@@ -427,6 +438,189 @@ def _log_estimator_params(param_map):
         try_mlflow_log(mlflow.log_params, truncated)
 
 
+class _AutologgingMetricsManager:
```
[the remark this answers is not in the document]
WeichenXu123: I think we can do this in a follow-up refactor PR.

mlflow/pyspark/ml/__init__.py
WeichenXu123: yes. because we need generate evalautor params here. after self.copy(params) we can get the correct param maps from `evalautor.params`

mlflow/pyspark/ml/__init__.py
WeichenXu123: But I prefer keep this method `_get_autologging_metrics_manager`, it represent the pattern of "we can only get the global instance of AutologgingMetricsManager, we cannot reconstruct it or replace it.

harupy: I think a proper way to do this is to using the `Final` type + `mypy` ([ref](https://mypy.readthedocs.io/en/stable/final_attrs.html#final-names)), but we currently don't use `mypy`. Can we just capitalize `_autologging_metrics_manager` to `_AUTOLOGGING_METRICS_MANAGER` to indicate that it's a constant and should never be changed?

mlflow/ml-package-versions.yml
```diff
@@ -313,6 +313,7 @@ spark:
   autologging:
     minimum: "3.0.0"
     maximum: "3.1.2"
+    requirements: ["scikit-learn"]
```
WeichenXu123, line 316: Because in pyspark ml autologging tests, I use sklearn iris test dataset, so I add the dependency for sklearn here.

mlflow/pyspark/ml/__init__.py
```diff
@@ -643,19 +908,81 @@ def fit_mlflow(original, self, *args, **kwargs):
...
 
     def patched_fit(original, self, *args, **kwargs):
+        metrics_manager = _get_autologging_metrics_manager()
+        should_log_post_training_metrics = metrics_manager.should_log_post_training_metrics()
         with _SparkTrainingSession(clazz=self.__class__, allow_children=False) as t:
             if t.should_log():
-                return fit_mlflow(original, self, *args, **kwargs)
+                with metrics_manager.disable_log_post_training_metrics():
+                    fit_result = fit_mlflow(original, self, *args, **kwargs)
+                # In some cases the `fit_result` may be an iterator of spark models.
```
harupy, line 926: When does `fit` return an iterator?

mlflow/pyspark/ml/__init__.py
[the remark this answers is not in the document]
dbczumar: Awesome! Can you file an internal ticket?

mlflow/pyspark/ml/__init__.py
```diff
@@ -650,13 +695,42 @@ def autolog(
...
         - An MLflow Model with the :py:mod:`mlflow.spark` flavor containing a fitted estimator
           (logged by :py:func:`mlflow.spark.log_model()`). Note that large models may not be
           autologged for performance and storage space considerations, and autologging for
           Pipelines and hyperparameter tuning meta-estimators (e.g. CrossValidator) is not yet
           supported.
           See ``log_models`` param below for details.
+        - For post training metrics API calls, a "metric_info.json" artifact is logged. This is a
+          JSON object whose keys are MLflow post training metric names
+          (see "Post training metrics" section for the key format) and whose values are the
+          corresponding evaluator information, includes evaluator class name and evaluator params.
```
dbczumar, line 733: Just one doc nit:

```suggestion
          corresponding evaluator information, including evaluator class name and evaluator params.
```

Awesome job with the docs here!

mlflow/pyspark/ml/__init__.py
WeichenXu123: This case:
https://github.com/apache/spark/blob/7d823367348c0235ffb212b09afcf053d070068d/python/pyspark/ml/base.py#L156

dbczumar (approved):
LGTM with a tiny nit. Great work, @WeichenXu123 ! I tried this out manually using an example adapted from https://docs.databricks.com/applications/machine-learning/train-model/mllib/index.html#gbt-regression-using-mllib-pipelines (replaced `Pipeline` with a single `GBTRegressor`). Confirmed that the RMSE evaluator adds an expected metric to the run associated with `GBTRegressor.fit()`.

harupy (approved):

tests/spark_autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -582,3 +598,163 @@ def test_gen_estimator_metadata(spark_session):  # pylint: disable=unused-argume
...
+
+        model.transform(eval_dataset)
+
+        mock_register_model.assert_called_once()
+        mock_is_metric_value_loggable.assert_not_called()
+        mock_register_prediction_input_dataset.assert_not_called()
+        mock_log_post_training_metric.assert_not_called()
+
+
+def test_is_metrics_value_loggable():
```
WeichenXu123, line 751: This test is similar with the one in sklearn autolog test.

After the follow-up AutologgingMetricManager refactor, we should remove this duplicated test.

PR #4711, opened 2021-08-19
Added retry logic for recoverable errors in the HTTP request method

jinzhang21:
Introduced a general retry logic to the HTTP request method, which is used in most inter-service comms by MLflow.
Before the retry was manually implemented for 429 rate limited response. Temporary issues such as connection errors cause a raise instead of being retried. This change aims to extend the retry logic to most temporary recoverable errors. 

## How is this patch tested?
Modified unit tests to ensure the change work as intended.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/rest_utils.py
+++ b/mlflow/utils/rest_utils.py
@@ -1,6 +1,4 @@
 import base64
-import time
-import logging
 import json
 import requests
 from contextlib import contextmanager
@@ -14,27 +12,91 @@ from mlflow.utils.proto_json_utils import parse_dict
 from mlflow.utils.string_utils import strip_suffix
 from mlflow.exceptions import MlflowException, RestException
 
-_REST_API_PATH_PREFIX = "/api/2.0"
 RESOURCE_DOES_NOT_EXIST = "RESOURCE_DOES_NOT_EXIST"
-
-_logger = logging.getLogger(__name__)
-
+_REST_API_PATH_PREFIX = "/api/2.0"
 _DEFAULT_HEADERS = {"User-Agent": "mlflow-python-client/%s" % __version__}
+# Response codes that generally indicate transient network failures and merit client retries,
+# based on guidance from cloud service providers
+# (https://docs.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#general-rest-and-retry-guidelines)
+_TRANSIENT_FAILURE_RESPONSE_CODES = frozenset(
+    [
+        408,  # Request Timeout
+        429,  # Too Many Requests
+        500,  # Internal Server Error
+        502,  # Bad Gateway
+        503,  # Service Unavailable
+        504,  # Gateway Timeout
+    ]
+)
+
+
+def _get_http_response_with_retries(
+    method, url, max_retries, backoff_factor, retry_codes, **kwargs
+):
+    """
+    Performs an HTTP request using Python's `requests` module with an automatic retry policy.
+
+    :param method: a string indicating the method to use, e.g. "GET", "POST", "PUT".
+    :param url: the target URL address for the HTTP request.
+    :param max_retries: Maximum total number of retries.
+    :param backoff_factor: a time factor for exponential backoff. e.g. value 5 means the HTTP
+      request will be retried with interval 5, 10, 20... seconds. A value of 0 turns off the
+      exponential backoff.
+    :param retry_codes: a list of HTTP response error codes that qualifies for retry.
+    :param kwargs: Additional keyword arguments to pass to `requests.Session.request()`
+
+    :return: requests.Response object.
+    """
+    assert 0 <= max_retries < 10
+    assert 0 <= backoff_factor < 120
+    retry = Retry(
+        total=max_retries,
+        connect=max_retries,
+        read=max_retries,
+        redirect=max_retries,
+        status=max_retries,
+        status_forcelist=retry_codes,
+        method_whitelist=None,  # Allow retry on any HTTP methods.
+        backoff_factor=backoff_factor,
+    )
+    adapter = HTTPAdapter(max_retries=retry)
+    with requests.Session() as http:
+        http.mount("https://", adapter)
+        http.mount("http://", adapter)
+        response = http.request(method, url, **kwargs)
+        return response
 
 
 def http_request(
-    host_creds, endpoint, retries=3, retry_interval=3, max_rate_limit_interval=60, **kwargs
+    host_creds,
+    endpoint,
+    method,
+    max_retries=5,
+    backoff_factor=2,
+    retry_codes=_TRANSIENT_FAILURE_RESPONSE_CODES,
+    timeout=10,
+    **kwargs
 ):
     """
-    Makes an HTTP request with the specified method to the specified hostname/endpoint. Ratelimit
-    error code (429) will be retried with an exponential back off (1, 2, 4, ... seconds) for at most
-    `max_rate_limit_interval` seconds.  Internal errors (500s) will be retried up to `retries` times
-    , waiting `retry_interval` seconds between successive retries. Parses the API response
-    (assumed to be JSON) into a Python object and returns it.
+    Makes an HTTP request with the specified method to the specified hostname/endpoint. Transient
+    errors such as Rate-limited (429), service unavailable (503) and internal error (500) are
+    retried with an exponential back off with backoff_factor * (1, 2, 4, ... seconds).
+    The function parses the API response (assumed to be JSON) into a Python object and returns it.
 
     :param host_creds: A :py:class:`mlflow.rest_utils.MlflowHostCreds` object containing
         hostname and optional authentication.
-    :return: Parsed API response
+    :param endpoint: a string for service endpoint, e.g. "/path/to/object".
+    :param method: a string indicating the method to use, e.g. "GET", "POST", "PUT".
+    :param max_retries: maximum number of retries before throwing an exception.
+    :param backoff_factor: a time factor for exponential backoff. e.g. value 5 means the HTTP
+      request will be retried with interval 5, 10, 20... seconds. A value of 0 turns off the
+      exponential backoff.
+    :param retry_codes: a list of HTTP response error codes that qualifies for retry.
+    :param timeout: wait for timeout seconds for response from remote server for connect and
+      read request.
+    :param kwargs: Additional keyword arguments to pass to `requests.Session.request()`
+
+    :return: requests.Response object.
     """
     hostname = host_creds.host
     auth_str = None
@@ -58,45 +120,22 @@ def http_request(
     if host_creds.client_cert_path is not None:
         kwargs["cert"] = host_creds.client_cert_path
 
-    def request_with_ratelimit_retries(max_rate_limit_interval, **kwargs):
-        response = requests.request(**kwargs)
-        time_left = max_rate_limit_interval
-        sleep = 1
-        while response.status_code == 429 and time_left > 0:
-            _logger.warning(
-                "API request to {path} returned status code 429 (Rate limit exceeded). "
-                "Retrying in %d seconds. "
-                "Will continue to retry 429s for up to %d seconds.",
-                sleep,
-                time_left,
-            )
-            time.sleep(sleep)
-            time_left -= sleep
-            response = requests.request(**kwargs)
-            sleep = min(time_left, sleep * 2)  # sleep for 1, 2, 4, ... seconds;
-        return response
-
     cleaned_hostname = strip_suffix(hostname, "/")
     url = "%s%s" % (cleaned_hostname, endpoint)
-    for i in range(retries):
-        response = request_with_ratelimit_retries(
-            max_rate_limit_interval, url=url, headers=headers, verify=verify, **kwargs
+    try:
+        return _get_http_response_with_retries(
+            method,
+            url,
+            max_retries,
+            backoff_factor,
+            retry_codes,
+            headers=headers,
+            verify=verify,
+            timeout=timeout,
+            **kwargs
         )
-        if response.status_code >= 200 and response.status_code < 500:
-            return response
-        else:
-            _logger.error(
-                "API request to %s failed with code %s != 200, retrying up to %s more times. "
-                "API response body: %s",
-                url,
-                response.status_code,
-                retries - i - 1,
-                response.text,
-            )
-            time.sleep(retry_interval)
-    raise MlflowException(
-        "API request to %s failed to return code 200 after %s tries" % (url, retries)
-    )
+    except Exception as e:
+        raise MlflowException("API request to %s failed with exception %s" % (url, e))
 
 
 def _can_parse_as_json(string):
@@ -107,11 +146,11 @@ def _can_parse_as_json(string):
         return False
 
 
-def http_request_safe(host_creds, endpoint, **kwargs):
+def http_request_safe(host_creds, endpoint, method, **kwargs):
     """
     Wrapper around ``http_request`` that also verifies that the request succeeds with code 200.
     """
-    response = http_request(host_creds=host_creds, endpoint=endpoint, **kwargs)
+    response = http_request(host_creds=host_creds, endpoint=endpoint, method=method, **kwargs)
     return verify_rest_response(response, endpoint)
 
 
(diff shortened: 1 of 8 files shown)


mlflow/utils/rest_utils.py
```diff
@@ -14,26 +14,76 @@
 from mlflow.utils.string_utils import strip_suffix
 from mlflow.exceptions import MlflowException, RestException
 
-_REST_API_PATH_PREFIX = "/api/2.0"
 RESOURCE_DOES_NOT_EXIST = "RESOURCE_DOES_NOT_EXIST"
+_REST_API_PATH_PREFIX = "/api/2.0"
+_DEFAULT_HEADERS = {"User-Agent": "mlflow-python-client/%s" % __version__}
+# Response codes that generally indicate transient network failures and merit client retries,
+# based on guidance from cloud service providers
+# (https://docs.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#general-rest-and-retry-guidelines)
```
jinzhang21, line 22: Do we want to use this set of codes as the general retry-able cases? Currently this is only used in cloud_storage_http_request(). For http_request, we use a subset of it, namely, 429 / 500 / 503.

In addition, 500 seems not really retry-able but we can discuss what is retry-able another day.

mlflow/utils/rest_utils.py
```diff
@@ -14,26 +14,76 @@
...
+        connect=max_retries,
+        read=max_retries,
+        redirect=max_retries,
+        status=max_retries,
+        status_forcelist=retry_codes,
+        backoff_factor=backoff_factor,
+    )
+    adapter = HTTPAdapter(max_retries=retry)
+    with requests.Session() as http:
+        http.mount("https://", adapter)
```
jinzhang21, line 56: Are mounting https and http sufficient? Any other schema needs to be supported?

mlflow/utils/rest_utils.py
```diff
@@ -14,26 +14,76 @@
...
 
-    :param host_creds: A :py:class:`mlflow.rest_utils.MlflowHostCreds` object containing
+    :host_creds: A :py:class:`mlflow.rest_utils.MlflowHostCreds` object containing
         hostname and optional authentication.
+    :endpoint: a string for service endpoint, e.g. "/path/to/object".
+    :max_retries: maximum number of retries before throwing an exception.
+    :backoff_factor: a time factor for exponential backoff. e.g. value 5 means the HTTP request
+      will be retried with interval 5, 10, 20... seconds. A value of 0 turns off the exp-backoff.
+    :retry_codes: a list of HTTP response error codes that qualifies for retry.
+    :timeout: wait for timeout seconds for response from remote server for connect and read request.
```
jinzhang21, line 86: Added timeout = 10s to allow quicker failure on connection / read timeout. Otherwise it's indefinite by default.

mlflow/utils/rest_utils.py
```diff
@@ -14,26 +14,76 @@
...
+        response = http.request(method, url, **kwargs)
+        return response
 
 
 def http_request(
-    host_creds, endpoint, retries=3, retry_interval=3, max_rate_limit_interval=60, **kwargs
+    host_creds,
+    endpoint,
+    max_retries=3,
+    backoff_factor=5,
```
jinzhang21, line 68: With 3 max retries, and backoff_factor of 5. We expect to retry after failure at 5, 10, 20s. This should be sufficient for most internet conditions.

mlflow/utils/rest_utils.py
```diff
@@ -14,26 +14,76 @@
...
+      will be retried with interval 5, 10, 20... seconds. A value of 0 turns off the exp-backoff.
+    :retry_codes: a list of HTTP response error codes that qualifies for retry.
+    :kwargs: Keyword arguments to pass to `requests.Session.request()`
+    """
+    assert 0 <= max_retries < 10
+    assert 0 <= backoff_factor < 120
+    retry = Retry(
+        total=max_retries,
+        connect=max_retries,
+        read=max_retries,
```
jinzhang21, line 48: Here we don't distinguish / discriminate on error kinds. We retry for any kind of issue, and raise an exception when total = max_retries.

tests/utils/test_rest_utils.py
```diff
@@ -65,97 +65,106 @@ def test_malformed_json_error_response(response_mock):
...
 
-@mock.patch("requests.request")
+@mock.patch("requests.Session.request")
 def test_http_request_hostonly(request):
     host_only = MlflowHostCreds("http://my-host")
     response = mock.MagicMock()
     response.status_code = 200
     request.return_value = response
-    http_request(host_only, "/my/endpoint")
+    http_request(host_only, "/my/endpoint", method="GET")
```
jinzhang21, line 74: http_request() function needs method as an arg. This is the case before the change as well. I'm considering making it mandatory. Let me know what you think.

tests/utils/test_rest_utils.py
```diff
@@ -185,64 +196,34 @@ def test_ignore_tls_verification_not_server_cert_path():
         )
 
 
-@mock.patch("requests.request")
-def test_429_retries(request):
```
jinzhang21, line 189: Since we rely on the HTTPAdaptor and Retry function. This test is no longer necessary.

As a side note, it's also non-trivial to test the Retry functionality on the high level without an actual internet connection. https://github.com/getsentry/responses/issues/135

mlflow/utils/rest_utils.py
dbczumar: I think it makes sense to expand the retry-able cases for non-cloud-storage requests. Thanks Jin!

mlflow/utils/rest_utils.py
dbczumar: Makes sense! Can we update `cloud_storage_http_request` to use the same logic?

mlflow/utils/rest_utils.py
dbczumar: Using `max_retries=3` and `backoff_factor=5` reduces our retry duration for rate limiting from 60 seconds to ~ 35 seconds. This seems acceptable to me, but it's worth calling out.

mlflow/utils/rest_utils.py
```diff
@@ -58,45 +106,21 @@ def http_request(
...
-        time_left = max_rate_limit_interval
-        sleep = 1
-        while response.status_code == 429 and time_left > 0:
-            _logger.warning(
-                "API request to {path} returned status code 429 (Rate limit exceeded). "
-                "Retrying in %d seconds. "
-                "Will continue to retry 429s for up to %d seconds.",
-                sleep,
-                time_left,
-            )
```
dbczumar, line 72: It was kind of nice to have this warning behavior here previously so that users know why their workflows are taking so long or eventually failing. Is there a way to log failures using `HTTPAdapter`?

mlflow/utils/rest_utils.py
dbczumar: 5 seconds is a long time to wait for the first failure. Perhaps we can use a smaller backoff factor with more retries?

tests/utils/test_rest_utils.py
dbczumar: Sounds great! Let's do it. `method` is required by `requests.request`, so we should require it explicitly in the signature.

tests/utils/test_rest_utils.py
dbczumar: @jinzhang21 Can we conduct a manual test for this change, perhaps on Databricks?

mlflow/utils/rest_utils.py
dbczumar: I'm not aware of any other URI schemes that we need to support

mlflow/utils/rest_utils.py
dbczumar: Awesome! Can we add this to `cloud_storage_http_request` too?

mlflow/utils/rest_utils.py
jinzhang21: Done.

mlflow/utils/rest_utils.py
jinzhang21: Changed factor to 2 with 5 retries, total retry time is around 2^6-2 = 62 seconds.

mlflow/utils/rest_utils.py
jinzhang21: Done.

tests/utils/test_rest_utils.py
harupy: +1 for making `method` mandatory.

mlflow/utils/rest_utils.py
jinzhang21: Done.

mlflow/utils/rest_utils.py
jinzhang21: Not that I'm aware of. If the request fails ultimately owing to rate limit, the user can still see it, but I agree that the visibility on rate limit is lower in this case.

tests/utils/test_rest_utils.py
jinzhang21: Done.

tests/utils/test_rest_utils.py
jinzhang21: Love to. What's the standard way of testing client change on Databricks? Build a wheel and manually install it?

jinzhang21 (commented):
Thanks for the comment, Corey. PTAL.

mlflow/utils/rest_utils.py
```diff
@@ -14,26 +12,76 @@
...
+]
 
-_logger = logging.getLogger(__name__)
 
-_DEFAULT_HEADERS = {"User-Agent": "mlflow-python-client/%s" % __version__}
+def _get_http_response_with_retries(max_retries, backoff_factor, retry_codes, **kwargs):
+    """
+    Performs an HTTP request using Python's `requests` module with an automatic retry policy.
+
+    :max_retries: Maximum total number of retries.
```
harupy, line 41:
```suggestion
    :param max_retries: Maximum total number of retries.
```

Nit: can we add `param` to parameter descriptions?

mlflow/utils/rest_utils.py
```diff
@@ -14,26 +12,82 @@
...
+
+def _get_http_response_with_retries(
+    method, url, max_retries, backoff_factor, retry_codes, **kwargs
+):
+    """
+    Performs an HTTP request using Python's `requests` module with an automatic retry policy.
+
+    :method: a string indicating the method to use, e.g. "GET", "POST", "PUT".
+    :url: the target URL address for the HTTP request.
+    :param max_retries: Maximum total number of retries.
```
harupy, line 41: Sorry I wasn't clear. Can you add `param` to other parameters?

jinzhang21: It's unclear what style I should follow from adjacent functions. Can you point me a canonical function example for style guide?

harupy: Sure!

https://github.com/mlflow/mlflow/blob/8ef60e2dcac711a149fecac9e2b9ac2846610ad4/mlflow/tracking/fluent.py#L420-L430

tests/utils/test_rest_utils.py
dbczumar: You got it! You can also install from this branch using `%pip install git+https://github.com/mlflow/mlflow@ML-16770`, which I find to be faster.

jinzhang21: Ran a E2E xgboost training process on a test shard. I was able to verify the timeout behavior but couldn't verify the retry logic unfortunately. What's the best way to simulate a temporary error on Databricks backend? When I try to throw an exception on the backend, it registers as an error (permanent) on the RPC console. Not sure what error code / response was actually returned to the client.

mlflow/utils/rest_utils.py
harupy: Thanks!

tests/utils/test_rest_utils.py
jinzhang21: Update. This is because Retry does not work by default on POST method: https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry.DEFAULT_ALLOWED_METHODS

Updated the param to None to allow all HTTP methods to be retried. Verified the retry works as expected on test shard.

mlflow/utils/rest_utils.py
```diff
@@ -14,27 +12,91 @@
...
+    :param max_retries: maximum number of retries before throwing an exception.
+    :param backoff_factor: a time factor for exponential backoff. e.g. value 5 means the HTTP
+      request will be retried with interval 5, 10, 20... seconds. A value of 0 turns off the
+      exponential backoff.
+    :param retry_codes: a list of HTTP response error codes that qualifies for retry.
+    :param timeout: wait for timeout seconds for response from remote server for connect and
+      read request.
+    :param kwargs: Additional keyword arguments to pass to `requests.Session.request()`
+
+    :return requests.Response object.
```
harupy, line 99:
```suggestion
    :return: requests.Response object.
```

nit

mlflow/utils/rest_utils.py
```diff
@@ -14,27 +12,91 @@
...
+    :param method: a string indicating the method to use, e.g. "GET", "POST", "PUT".
+    :param url: the target URL address for the HTTP request.
+    :param max_retries: Maximum total number of retries.
+    :param backoff_factor: a time factor for exponential backoff. e.g. value 5 means the HTTP
+      request will be retried with interval 5, 10, 20... seconds. A value of 0 turns off the
+      exponential backoff.
+    :param retry_codes: a list of HTTP response error codes that qualifies for retry.
+    :param kwargs: Additional keyword arguments to pass to `requests.Session.request()`
+
+    :return requests.Response object.
```
harupy, line 48:
```suggestion
    :return: requests.Response object.
```

nit

mlflow/utils/rest_utils.py
```diff
@@ -14,27 +12,91 @@
...
+    assert 0 <= max_retries < 10
+    assert 0 <= backoff_factor < 120
+    retry = Retry(
+        total=max_retries,
+        connect=max_retries,
+        read=max_retries,
+        redirect=max_retries,
+        status=max_retries,
+        status_forcelist=retry_codes,
+        method_whitelist=None,  # Allow retry on any HTTP methods.
```
harupy, line 59:
```suggestion
        allowed_methods=None,  # Allow retry on any HTTP methods.
```

Can we use `allowed_methods` instead of `method_whitelist` that is deprecated now?

https://github.com/urllib3/urllib3/blob/main/CHANGES.rst#1260-2020-11-10

harupy: On the master branch of the `url` repository, `method_whitelist` has already been removed:

https://github.com/urllib3/urllib3/blob/25ed51f319a3c6920bb5bd959924ca80925ee120/src/urllib3/util/retry.py#L188-L206

mlflow/utils/rest_utils.py
harupy: If `:` is missing, it's rendered like below by sphinx:

image

mlflow/utils/rest_utils.py
jinzhang21: I tried allowed_methods on my Python 3.6 setup, it didn't work. But we can try to see if passes integration tests.

mlflow/utils/rest_utils.py
```diff
@@ -14,27 +12,91 @@
...
+    assert 0 <= max_retries < 10
+    assert 0 <= backoff_factor < 120
+    retry = Retry(
+        total=max_retries,
+        connect=max_retries,
+        read=max_retries,
+        redirect=max_retries,
+        status=max_retries,
+        status_forcelist=retry_codes,
+        allowed_methods=None,  # Allow retry on any HTTP methods.
```
jinzhang21, line 59: allowed_methods doesn't work on wheel install either. I have to change it back to method_whitelist:

MlflowException: API request to https://test-shard-jin-zhang.dev.databricks.com/api/2.0/mlflow/runs/create failed with exception __init__() got an unexpected keyword argument 'allowed_methods'

harupy: I see, I think we're using an old version of `urrlib3` in databricks. Can we use `allowed_methods` if the version of `urllib3` is >= 1.26.0?

```python
from packaging.version import Version

methods_arg = "allowed_methods" if Version(urllib3.__version__) >= Version("1.26.0") else "method_whitelist"
Retry(..., **{methods_arg: None})
```

dbczumar (approved):
LGTM! Thanks @jinzhang21 !

mlflow/utils/rest_utils.py
dbczumar: @harupy I installed urllib3 1.26.6, and I don't see `allowed_methods` defined on the `Retry()` API. I think it's a documentation issue.

dbczumar:
```
In [4]: import urllib3

In [5]: urllib3.__version__
Out[5]: '1.26.6'

In [6]: from urllib3.util import Retry

In [7]: help(Retry)

class Retry(builtins.object)
 |  Retry(total=10, connect=None, read=None, redirect=None, status=None, other=None, allowed_methods=<object object at 0x7fd1f09eec30>, status_forcelist=None, backoff_factor=0, raise_on_redirect=True, raise_on_status=True, history=None, respect_retry_after_header=True, remove_headers_on_redirect=<object object at 0x7fd1f09eec30>, method_whitelist=<object object at 0x7fd1f09eec30>)
```

harupy: @dbczumar Are you really sure that `Retry` doesn't have `allowed_methods`?

```python
>>> import urllib3
>>> urllib3.__version__
'1.26.6'
>>> from urllib3.util import Retry
>>> Retry()
Retry(total=10, connect=None, read=None, redirect=None, status=None)
>>> Retry(allowed_methods=None)
Retry(total=10, connect=None, read=None, redirect=None, status=None)

>>> Retry(method_whitelist=None)
__main__:1: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
Retry(total=10, connect=None, read=None, redirect=None, status=None)
```

harupy: https://github.com/urllib3/urllib3/blob/1.26.6/src/urllib3/util/retry.py#L233

dbczumar: Ah, missed it. Looks like it's there. I'll add this conditional logic.

harupy (approved):
LGTM once https://github.com/mlflow/mlflow/pull/4711#discussion_r695044896 is addressed!

PR #4188, opened 2021-03-19
Update compatibility with fastai v2

lluissalord:
## What changes are proposed in this pull request?

Currently MLFlow is supporting fastai_v1 which is already deprecated. Hence, here I propose several changes which updates MLFlow to be aligned with fastai_v2

## How is this patch tested?

Unit Tests updated here

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

Compatibility with last version of fastai framework.

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [x] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/server-infra`: MLflow server, JavaScript dev server
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

<a name="release-note-category"></a>
### How should the PR be classified in the release notes? Choose one:

- [ ] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section
- [ ] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section
- [x] `rn/feature` - A new user-facing feature worth mentioning in the release notes
[...]

PR #4715, opened 2021-08-20
Upgraded fastai from v1 to v2 (>=2.4.1)

jinzhang21:
Deprecated fastai v1 and upgraded it to v2 with minimum version 2.4.1. Closes https://github.com/mlflow/mlflow/pull/4188

## How is this patch tested?
Unit tests and integration tests.

## Release Notes
Deprecated fastai v1 and upgraded it to v2 (minimum version 2.4.1).

Components 

Interface 

Language 

Integrations


mlflow/fastai/callback.py
```diff
@@ -0,0 +1,157 @@
...
+                self.opt_func.keywords["opt"].__name__,
+            )
+        else:
+            try_mlflow_log(
+                mlflow.log_param, self.freeze_prefix + "opt_func", self.opt_func.__name__
+            )
+
+        params_not_to_log = []
+        for cb in self.cbs:
+            if isinstance(cb, ParamScheduler):
```
harupy, line 89: For param schedulers, can we also log the schedular name and its start& end values?

```
sched = {'lr': SchedLin(start=1e-3, end=1e-2)}
learn.fit(1, cbs=ParamScheduler(sched))
```

mlflow/ml-package-versions.yml
```diff
@@ -213,24 +213,21 @@ gluon:
     run: |
       pytest tests/gluon_autolog/test_gluon_autolog.py --large
 
-fastai-1.x:
+fastai-2.x:
```
harupy, line 216:
```suggestion
fastai:
```

I think we can remove `-2.x`

mlflow/fastai/callback.py
harupy: Ah the start and end values are logged as `<param>_init` and `<param>_final`

mlflow/fastai/callback.py
```diff
@@ -0,0 +1,157 @@
...
+                return False
+
+        # Do not record in case of predicting
+        if hasattr(self, "lr_finder") or hasattr(self, "gather_preds"):
+            return
+        # Remove non-float metrics and record the rest.
+        metrics = self.recorder.log
+        metrics = dict(zip(self.recorder.metric_names, metrics))
+        for key in [key for key, val in metrics.items() if not _is_float(val)]:
+            del metrics[key]
```
harupy, line 51:
```suggestion
        metrics = {k: v for k, v in zip(self.recorder.metric_names, metrics) if _is_float(v)}
```

Can we simplify this the code here?

mlflow/fastai/callback.py
```diff
@@ -0,0 +1,157 @@
...
+                Current model have only one param group which makes it impossible to freeze.
+                Because of this it will record some fitting params twice (overriding exception)
+                """
+            )
+
+        frozen = self.opt.frozen_idx != 0
+        if frozen and self.is_fine_tune:
+            self.freeze_prefix = "freeze_"
+        else:
+            self.freeze_prefix = ""
```
harupy, line 73: Can we just log `frozen_idx` (that indicates the model was trained with some layers frozen) rather than prefixing all logged parameters with `freeze_`?

examples/fastai/train.py
```diff
@@ -1,10 +1,10 @@
 #
 # This short example is based on the fastai GitHub Repository of vision examples
-# https://github.com/fastai/fastai/blob/master/examples/vision.ipynb
+# https://github.com/fastai/fastai/blob/master/nbs/examples/mnist_blocks.py
 # Modified here to show mlflow.fastai.autolog() capabilities
 #
 import argparse
-import fastai.vision as vis
+from fastai.vision.all import *
```
harupy, line 7: Can we only import used modules?

mlflow/fastai/callback.py
jinzhang21: It looks like both "freeze_{param}" and "{param}" are to be logged when fine_tune is used. I'm not familiar with fastai. Unsure if it's useful that way. You can see this behavior tested in tests/fastai/test_fastai_autolog.py.

mlflow/ml-package-versions.yml
jinzhang21: Looks like fastai-2.x is required.

examples/fastai/train.py
jinzhang21: Done.

mlflow/ml-package-versions.yml
harupy: I think we also need to fix this line:

https://github.com/mlflow/mlflow/pull/4715/files#diff-f7f812cf7b56d2001536c3463d4f4464ec4fa521f176b010dd1e28ed14b93bf4R13

mlflow/fastai/callback.py
harupy: Here's the source code of `fine_tune`:

```python
@patch
@delegates(Learner.fit_one_cycle)
def fine_tune(self:Learner, epochs, base_lr=2e-3, freeze_epochs=1, lr_mult=100,
              pct_start=0.3, div=5.0, **kwargs):
    "Fine tune with `Learner.freeze` for `freeze_epochs`, then with `Learner.unfreeze` for `epochs`, using discriminative LR."
    self.freeze()
    self.fit_one_cycle(freeze_epochs, slice(base_lr), pct_start=0.99, **kwargs)
    base_lr /= 2
    self.unfreeze()
    self.fit_one_cycle(epochs, slice(base_lr/lr_mult, base_lr), pct_start=pct_start, div=div, **kwargs)
```

https://github.com/fastai/fastai/blob/4b0785254fdece1a44859956b6e54eedb167a97e/fastai/callback/schedule.py#L155-L164

harupy: Logged params look like this:

mlflow/fastai/callback.py
```diff
@@ -0,0 +1,156 @@
...
+import matplotlib.pyplot as plt
+
+import mlflow.tracking
+from mlflow.utils.autologging_utils import try_mlflow_log
+from mlflow.fastai import log_model
+
+from fastai.callback.core import Callback
+
+
+# Move outside, because it cannot be pickled. Besides, ExceptionSafeClass was giving some issues
```
harupy, line 15: What happens if we use `ExceptionSafeClass`?

harupy: I locally ran the tests with the code below and they all passed.

```suggestion
class __MLflowFastaiCallback(Callback, metaclass=ExceptionSafeClass):
    ...
```

tests/fastai/test_fastai_autolog.py
```diff
@@ -36,25 +47,39 @@ def iris_data():
...
+
+
+@pytest.fixture(scope="session")
+def mnist_data():
+    mnist = untar_data(URLs.MNIST_TINY)
+    return ImageDataLoaders.from_folder(mnist, num_workers=0)
+
+
+def fastai_visual_model(data, **kwargs):
+    return cnn_learner(data, models.resnet18, normalize=False, **kwargs)
```
harupy, line 71: I think training a ResNet model with the MNIST dataset takes a long time. Can we use a smaller dataset and a simpler model?

mlflow/fastai/callback.py
harupy: A plot for `freeze_lr` looks like this:

tests/fastai/test_fastai_autolog.py
harupy: In this run, `pytest tests/fastai/test_fastai_autolog.py --large` takes about 10 minutes:
https://github.com/mlflow/mlflow/pull/4715/checks?check_run_id=3583140216

tests/autologging/test_autologging_utils.py
```diff
@@ -782,8 +782,8 @@ def test_violates_pep_440():
 @pytest.mark.parametrize(
     "flavor,module_version,expected_result",
     [
-        ("fastai", "1.0.60", True),
-        ("fastai", "1.0.50", False),
+        ("fastai", "2.4.1", True),
+        ("fastai", "1.0.60", False),
```
WeichenXu123, line 786: we can add a test case of:
`("fastai", "2.3.1", False),`

tests/fastai/test_fastai_autolog.py
```diff
@@ -94,50 +141,99 @@ def fastai_random_data_run(iris_data, fit_variant, manual_run):
 
 
 @pytest.mark.large
-@pytest.mark.parametrize("fit_variant", ["fit", "fit_one_cycle"])
-def test_fastai_autolog_logs_expected_data(fastai_random_data_run, fit_variant):
+@pytest.mark.parametrize("fit_variant", ["fit", "fit_one_cycle", "fine_tune"])
+def test_fastai_autolog_logs_expected_data(fastai_random_visual_data_run, fit_variant):
```
WeichenXu123, line 145: Seems missing `mlflow.fastai.autolog()` line for this test. i think this will cause singlely run this test fail, like `pytest tests/fastai/test_fastai_autolog.py -k test_fastai_autolog_logs_expected_data`

tests/fastai/test_fastai_model_export.py
```diff
@@ -40,20 +39,22 @@ def fastai_model():
...
 
 
 @pytest.fixture
 def model_path(tmpdir):
-    return os.path.join(str(tmpdir), "model")
+    path = os.path.join(str(tmpdir), "model")
+    # Seems like Windows absolute path are not currently supported by MLFlow
+    # then they are transformed to relative path
+    if path.lower().startswith("c:"):
+        path = path[2:]
```
WeichenXu123, line 56: do we need support run test on windows ? the 2 line logic looks weird, could it start with `d:` ?

tests/fastai/test_fastai_autolog.py
harupy: `mlflow.fastai.autolog` is called in the `fastai_random_visual_data_run` fixture.

mlflow/fastai/callback.py
```diff
@@ -0,0 +1,156 @@
...
+                        mlflow.log_param, self.freeze_prefix + param + "_init", values[0]
+                    )
+                    try_mlflow_log(
+                        mlflow.log_param, self.freeze_prefix + param + "_final", values[-1]
+                    )
+
+                    # Plot and save image of scheduling
+                    fig = plt.figure()
+                    plt.plot(values)
+                    plt.ylabel(param)
```
harupy, line 113: A plot for `freeze_lr` looks like this:



- There is no x-axis label.
- There is no legend.

tests/fastai/test_fastai_model_export.py
harupy: No, we don't need to.

tests/autologging/test_autologging_utils.py
jinzhang21: Done.

tests/fastai/test_fastai_autolog.py
jinzhang21: I didn't check every model there, but resnet18 has "only" 11M params. It appears to be the smallest model among those offered: https://fastai1.fast.ai/vision.models.html

jinzhang21:
There's some problem with sklearn versioning, causing errors mentioned here: https://stackoverflow.com/questions/62322882/load-iris-got-an-unexpected-keyword-argument-as-frame. Not sure how it's related to this PR or a general problem.

tests/fastai/test_fastai_autolog.py
harupy: @jinzhang21  
I found a way to create a fine-tunable tabular model. Can I push some commits?

harupy: @jinzhang21 Puhsed a commit to update the code. Now it takes about 3 min. to run tests for fastai autologging: https://github.com/mlflow/mlflow/pull/4715/checks?check_run_id=3605777911

jinzhang21: Super nice! Thanks for the change, Haru!

harupy (approved):
LGTM!

Issue #4745, opened 2021-08-26
[FR] Use mamba for creating environments


giacomov:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [x ] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

A following up PR will implement the following proposed modifications:

* allow for the use of a new environment variable, `MLFLOW_CONDA_CREATE_ENV_CMD`, that allows the user to specify an executable to be used for creating environments. If the user does not set it, `conda` will be used. If the user sets it, for example as `MLFLOW_CONDA_CREATE_ENV_CMD=mamba`, then `mamba` will be used (of course, the user will have to take care of installing `mamba` first).
* only one minor change is required, in the file `utils/conda.py`. Tests for this change will also be provided.

## Motivation
- What is the use case for this feature?

 [`mamba`](https://github.com/mamba-org/mamba) is a package manager compatible with Conda, but many times faster than conda when creating environments and installing packages. When using MLflow, creating the environment can be very time consuming, sometimes dominating the total runtime of an MLflow project. This feature allows to use `mamba` instead of `conda` to create the environments. 

It only requires the user to set the env variable `MLFLOW_CONDA_CREATE_ENV_CMD=mamba` (after having installed mamba, of course).

It also allows, for advanced users, to create a wrapper script or executable that runs only when creating environments. Just point `MLFLOW_CONDA_CREATE_ENV_CMD` to the wrapper.

- Why is this use case valuable to support for MLflow users in general?

It makes the execution of MLflow project and pipelines potentially much faster, by speeding up the environment creation step without requiring any change whatsoever to the MLflow project.

- Why is this use case valuable to support for your project(s) or organization?

[...]

PR #4746, opened 2021-08-26
[FR] Use mamba for creating environments

giacomov:

As described in #4745

* Add an optional environment variable (`MLFLOW_CONDA_CREATE_ENV_CMD`). The user can set it to a command that can be used in place of `conda` to *create* environments. Thanks to this facility, a user can for example use `mamba` [(link)](https://github.com/mamba-org/mamba) in place of `conda` to create environments. Mamba is much faster than conda in creating environments.

## How is this patch tested?

I made a minor refactoring of `utils/conda.py` to create a function that computes the path to `conda` and to the executable that should be used to create environments. If the user does not set `MLFLOW_CONDA_CREATE_ENV_CMD`, then the latter will point to `conda`, otherwise it will point to the required executable, like `mamba`.

I wrote a specific test for that function that checks that the executables for managing and creating enviroments are what is expected in different cases.

Add an optional environment variable (`MLFLOW_CONDA_CREATE_ENV_CMD`). The user can set it to a command that can be used in place of `conda` to *create* environments. Thanks to this facility, a user can for example use `MLFLOW_CONDA_CREATE_ENV_CMD=mamba` to use `mamba` [(link)](https://github.com/mamba-org/mamba) in place of `conda` to create environments. Mamba is much faster than conda in creating environments.

Components 
- [ x] `area/projects`: MLproject format, project running backends

Interface 

Language 

Integrations


CHANGELOG.rst
```diff
@@ -1,5 +1,12 @@
 Changelog
 =========
+1.20.1 (2021-08-26)
```
jinzhang21, line 3: Could you please pull from the master branch to remove these unintended changes? Thanks.

mlflow/utils/conda.py
```diff
@@ -60,6 +61,30 @@ def _get_conda_env_name(conda_env_path, env_id=None):
     return "mlflow-%s" % hashlib.sha1(conda_env_contents.encode("utf-8")).hexdigest()
 
 
+def _get_conda_executables():
```
jinzhang21, line 64: What's the reason for not using mamba for both operation and env creation? Allowing 2 conda executables could confuse most users without the context. I'd prefer to not doing so if possible.

CHANGELOG.rst
giacomov: ok, done.

mlflow/utils/conda.py
giacomov: yeah, I know, but unfortunately `mamba` provides only a subset of conda commands, specifically install, create, list, search, run, info and clean. For example, the `init` command is not available.

Hence, we must use this awkward thing of having two executables. On the other hand, `mamba` is installed using `conda`, so in most cases you have both.

jinzhang21: I see.

I'm a little confused reading the function comment below. It appears Mamba's usage is mostly limited to only create environments (as defined by MLFLOW_CONDA_CREATE_ENV_CMD). Is it correct? Do you mind listing the value of MLFLOW_CONDA_CREATE_ENV_CMD and MLFLOW_CONDA_HOME in the proposed use case when mamba and conda are both involved?

If my understanding is correct, how about a slightly different approach (mostly to reduce confusion):
1. define MLFLOW_MAMBA_HOME instead of MLFLOW_CONDA_CREATE_ENV_CMD
2. add a function _get_mamba_bin_executable(), similar to get_conda_bin_executable() to return mamba if defined, otherwise return conda instead (by calling get_conda_bin_executable() internally probably).
3. In get_or_create_conda_env(), call _get_mamba_bin_executable() and you are done.

giacomov: Yes, the usage of `mamba` is only limited to creating environment. 

If you want to use ``conda`` for both creating and activating environments, you don't need to do anything. This works as it has always worked. If you don't set anything, then ``conda`` is used from your ``PATH``. If you set ``MLFLOW_CONDA_HOME``, then ``${MLFLOW_CONDA_HOME}/bin/conda`` is used.

If you want to use mamba instead of conda to create environments, you just set:

```bash
export MLFLOW_CONDA_CREATE_ENV_CMD=mamba
```
Everything else works the same. So if you do not set `MLFLOW_CONDA_HOME`, then `mamba` will be assumed to be in one of the directories within the `PATH` env variable. If you do set it, then `mamba` will be executed from `${MLFLOW_CONDA_HOME}/bin/mamba`.

One reason to do it this way is to be a little future proof. Let's imagine that a third implementation of the `conda` package manager comes out in the future (i.e., an alternative to both conda and mamba), let's imagine it is called `boa`. Then you just set `MLFLOW_CONDA_CREATE_ENV_CMD=boa` and you are done. In other words, my implementation does not necessarily assume that the alternative implementation of `conda` is specifically `mamba`.

What do you think? I am open to go your route, although that would restrict us explicitly to mamba only.

tests/projects/test_projects.py
```diff
@@ -305,6 +305,38 @@ def test_conda_path(mock_env, expected_conda, expected_activate):
         assert mlflow.utils.conda.get_conda_bin_executable("activate") == expected_activate
 
 
+@pytest.mark.parametrize(
```
jinzhang21, line 308: Please add a test on failure scenario as well.

mlflow/utils/conda.py
jinzhang21: OK. How about this? Instead of making _get_conda_executables() returns 2 executables, add a separate function called _get_conda_executable_for_create_env() to return a dedicated executable for creating environment. With that it is clear what that executable is for, and how it can be obtained.

The rest of the code and remain mostly unchanged.

mlflow/utils/conda.py
```diff
@@ -84,18 +111,42 @@ def get_or_create_conda_env(conda_env_path, env_id=None):
...
             "to the path of the Conda executable".format(conda_path, MLFLOW_CONDA_HOME)
         )
+
+    try:
+        process.exec_cmd([conda_env_create_path, "--help"], throw_on_error=False)
+    except EnvironmentError:
+        raise ExecutionException(
+            "You have set the env variable {0}, but {1} does not exist or "
+            "it is not working properly. Note that {1} and the conda executable need to be "
+            "in the same conda environment. You can change the search path by"
```
jinzhang21, line 121: This exception message is a little confusing to me. Perhaps try to trigger the exception and see if it makes sense (see my comment below on adding test on failure scenarios).

mlflow/utils/conda.py
```diff
@@ -10,6 +10,7 @@
 # Environment variable indicating a path to a conda installation. MLflow will default to running
 # "conda" if unset
 MLFLOW_CONDA_HOME = "MLFLOW_CONDA_HOME"
+MLFLOW_CONDA_CREATE_ENV_CMD = "MLFLOW_CONDA_CREATE_ENV_CMD"
```
jinzhang21, line 13: MLFLOW_CONDA_CREATE_ENV_CMD --> MLFLOW_CONDA_HOME_FOR_CREATE_ENV for naming consistency.
Also add some comments explaining when and how it's used with an example illustrated with mamba.

jinzhang21 (commented):
LGTM overall. Some minor comments. Thanks a lot for contributing to MLflow!

tests/projects/test_projects.py
giacomov: ok, done.

mlflow/utils/conda.py
giacomov: ok, I added the test

mlflow/utils/conda.py
giacomov: I added the comment, but left the name as it was. I'm open to change that, but I wouldn't call it `*_HOME_*` because that refers to a path, while the `MLFLOW_CONDA_HOME_FOR_CREATE_ENV` is just a command.

For example, to use mamba for creating environments, you need to do:
> conda install mamba -n base -c conda-forge
> export MLFLOW_CONDA_HOME_FOR_CREATE_ENV=mamba
> mlflow run ...

jinzhang21 (approved):
Thanks a lot for contributing to MLflow, Giacomo!

tgy:
@giacomov i'd like to use this `MLFLOW_CONDA_CREATE_ENV_CMD` env var to use `micromamba` for setting up environments. i'm just a bit confused as to where the `MLFLOW_CONDA_CREATE_ENV_CMD` env var should be set. i don't want to always have to do `MLFLOW_CONDA_CREATE_ENV_CMD=micromamba mlflow [...]` but instead specify within the `MLproject` config file to always use `micromamba`. what's the right way to achieve that? thanks a lot for this PR btw! sorry for bumping up. maybe i'll add documentation about this because it's inexistant in the official mlflow doc.

giacomov:
I don't think you can do that with the MLProject file. Afaik it does not allow you to specify and environment _before_ processing your conda env file.

You can however obviously export your variable first then execute your commands:

```
> export MLFLOW_CONDA_CREATE_ENV_CMD=mamba
> mlflow command1
> mlflow command2
```

although I am not sure this covers your use case.

PR #4759, opened 2021-08-31
Enable auto dependency inference in spark flavor

harupy:

Enable auto dependency inference in spark flavor.

## How is this patch tested?

Existing tests

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/_capture_modules.py
+++ b/mlflow/utils/_capture_modules.py
@@ -13,6 +13,7 @@ import mlflow
 from mlflow.utils.file_utils import write_to
 from mlflow.pyfunc import MAIN
 from mlflow.models.model import MLMODEL_FILE_NAME, Model
+from mlflow.utils.databricks_utils import is_in_databricks_runtime
 
 
 def _get_top_level_module(full_module_name):
@@ -83,6 +84,11 @@ def main():
     sys.path = json.loads(args.sys_path)
 
     cap_cm = _CaptureImportedModules()
+
+    if flavor == "spark" and is_in_databricks_runtime():
+        from dbruntime.spark_connection import initialize_spark_connection
+
+        initialize_spark_connection()
 
     # If `model_path` refers to an MLflow model directory, load the model using
     # `mlflow.pyfunc.load_model`
--- a/mlflow/utils/_capture_modules.py
+++ b/mlflow/utils/_capture_modules.py
@@ -83,12 +83,16 @@ def main():
     # Mirror `sys.path` of the parent process
     sys.path = json.loads(args.sys_path)
 
+    if flavor == mlflow.spark.FLAVOR_NAME and is_in_databricks_runtime():
+        try:
+            # pylint: disable=import-error
+            from dbruntime.spark_connection import initialize_spark_connection
+
+            initialize_spark_connection()
+        except Exception:
+            pass
+
     cap_cm = _CaptureImportedModules()
-
-    if flavor == "spark" and is_in_databricks_runtime():
-        from dbruntime.spark_connection import initialize_spark_connection
-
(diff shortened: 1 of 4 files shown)


mlflow/utils/_capture_modules.py
```diff
@@ -84,6 +85,11 @@ def main():
 
     cap_cm = _CaptureImportedModules()
 
+    if flavor == "spark" and is_in_databricks_runtime():
+        from dbruntime.spark_connection import initialize_spark_connection
+
+        initialize_spark_connection()
```
WeichenXu123, line 91: if the case not in databricks runtime ?

harupy: @WeichenXu123 

If not, a new spark session is created in the following code:
https://github.com/mlflow/mlflow/blob/c4b8e849a3c718575e106f3499a12bea9b0b821e/mlflow/spark.py#L687-L700

mlflow/utils/_capture_modules.py
```diff
@@ -84,6 +85,11 @@ def main():
 
     cap_cm = _CaptureImportedModules()
 
+    if flavor == "spark" and is_in_databricks_runtime():
+        from dbruntime.spark_connection import initialize_spark_connection
```
harupy, line 89: This approach breaks when `initialize_spark_connection` is renamed or moved to a different module.

harupy: Should we make the spark initialization process modifiable via patching or an environment variable?

WeichenXu123: Add a MLR test to prevent it break.

mlflow/utils/_capture_modules.py
WeichenXu123: Shall we add try/catch here and add fallback handling ?
e.g. the case  User upgrade/downgrade the builtin mlflow of databrick runtime and found the `dbruntime.spark_connection.initialize_spark_connection` API does not exist.

dbczumar: Agree with @WeichenXu123 , I think a try/catch is a good idea here.

dbczumar (approved):
@harupy LGTM once https://github.com/mlflow/mlflow/pull/4759/files#r699873541 is addressed. Great work!

mlflow/utils/_capture_modules.py
```diff
@@ -82,6 +83,15 @@ def main():
...
     sys.path = json.loads(args.sys_path)
 
+    if flavor == mlflow.spark.FLAVOR_NAME and is_in_databricks_runtime():
+        try:
+            # pylint: disable=import-error
+            from dbruntime.spark_connection import initialize_spark_connection
+
+            initialize_spark_connection()
+        except Exception:
+            pass
```
harupy, line 93: @WeichenXu123 @dbczumar Added `try-catch`. We could optimistically run the subsequent process after the failure, but `mlflow.spark.load_pyfucn` will also propbably fail If `initialize_spark_connection` fails.

dbczumar: @harupy Makes sense that it will likely fail. Can we raise a clearer error message for debugging purposes?

harupy:
```suggestion
        try:
            # pylint: disable=import-error
            from dbruntime.spark_connection import initialize_spark_connection

            initialize_spark_connection()
        except Exception as e:
            raise Exception("Failed to initialize spark connection") from e
```

Can we just say `Failed to initialize spark connection`?

dbczumar: Sounds good to me!

PR #4773, opened 2021-09-03
Add prophet flavor

BenWilson2:

* Adds univariate Prophet flavor

Follow-ups
* Autologging support
* Multivariate forecasting support
* Example

## How is this patch tested?

Unit tests

Add support for univariate series Prophet forecasting

Components 
- [x] `area/examples`: Example code

Interface 

Language 

Integrations


mlflow/prophet.py
```diff
@@ -0,0 +1,296 @@
...
+    model_data_path = os.path.join(path, _MODEL_BINARY_FILE_NAME)
+    _save_model(pr_model, model_data_path)
+
+    model_bin_kwargs = {_MODEL_BINARY_KEY: _MODEL_BINARY_FILE_NAME}
+    pyfunc.add_to_model(
+        mlflow_model, loader_module="mlflow.prophet", env=_CONDA_ENV_FILE_NAME, **model_bin_kwargs
+    )
+    flavor_conf = {
+        _MODEL_TYPE_KEY: pr_model.__class__.__name__,
+        _SAVE_FORMAT_KEY: kwargs.get("format", "pr"),
```
dbczumar, line 137: Regardless of the `format` specified in `kwargs`, we seem to be unconditionally serializing the model via `model_to_json`. Given that `model_to_json` is the recommended approach for serialization, I don't think we need to support alternative save formats.

Can we drop the `kwargs` argument from the `save_model()` / `log_model()` APIs and remove this `_SAVE_FORMAT_KEY` config for the flavor?

Let me know if I'm missing something here :)

mlflow/prophet.py
```diff
@@ -0,0 +1,296 @@
...
+        **model_bin_kwargs,
+    }
+    mlflow_model.add_flavor(
+        FLAVOR_NAME,
+        numpy_version=numpy.__version__,
```
dbczumar, line 142: Is there something special about the numpy version that may impact prophet behavior? If so, should we also include it in the list of generated pip requirements? If not, can we drop it from the flavor conf?

mlflow/prophet.py
```diff
@@ -0,0 +1,296 @@
...
+        pystan_version=pystan.__version__,
+        prophet_version=prophet.__version__,
+        **flavor_conf,
+    )
+    mlflow_model.save(os.path.join(path, MLMODEL_FILE_NAME))
+
+    if conda_env is None:
+        if pip_requirements is None:
+            # cannot use inferred requirements due to prophet's build process
```
dbczumar, line 151: Nice inline comment! Can we elaborate here a bit more and indicate that prophet requires installation of several specific dependencies in a specific order, and dependency inference cannot guarantee that these dependencies are ordered properly?

cc @harupy who may also have input here. Another option we have is to add inferred dependencies after the list of hardcoded dependencies, though this may not be necessary since prophet doesn't support things like custom transformers.

mlflow/prophet.py
```diff
@@ -0,0 +1,296 @@
...
+    with open(path, "w") as f:
+        yaml.safe_dump(model_ser, stream=f, default_flow_style=False)
+
+
+def _load_model(path):
+
+    from prophet.serialize import model_from_json
+
+    with open(path, "r") as f:
+        model = yaml.safe_load(f)
```
dbczumar, line 253: Interesting that we're using the `yaml` module here to write / read the serialized model contents produced by `model_to_json`. Is there a reason for leveraging `yaml` rather than `json.dumps / loads`? If not, can we swap in `json`?

tests/prophet/test_prophet_model_export.py
```diff
@@ -0,0 +1,455 @@
...
+            metrics[metric] = cv_metrics[metric].mean()
+
+        mlflow.prophet.save_model(prophet_model.model, model_path)
+        mlflow.log_params(params)
+        mlflow.log_metrics(metrics)
+
+    client = MlflowClient()
+    run_id = client.get_run(run).info.run_id
+    logged_metrics = client.get_metric_history(run_id, "rmse")
+    assert logged_metrics[0].value == metrics["rmse"]
```
dbczumar, line 455: We probably don't need this test case since it isn't testing functionality introduced in the `mlflow.prophet` model flavor. This may be more applicable for autologging in the future.

tests/prophet/test_prophet_model_export.py
```diff
@@ -0,0 +1,455 @@
...
+def future_horizon_df(model, horizon):
+    return model.make_future_dataframe(periods=horizon)
+
+
+@pytest.fixture
+def generate_forecast(model, horizon):
+    return model.predict(model.make_future_dataframe(periods=horizon))[TARGET_FIELD_NAME]
+
+
+@pytest.mark.large
```
dbczumar, line 140: Can we add `pytestmark = pytest.mark.large` at the top of the file below the library imports, which will automatically mark all test cases as `pytest.mark.large` so that we don't have to do this for each test case & risk missing one?

Here's an example: https://github.com/mlflow/mlflow/blob/5f5fada1e1b423fefa5f99f6702377888b05f2a4/tests/pytorch/test_pytorch_autolog.py#L16

tests/prophet/test_prophet_model_export.py
```diff
@@ -0,0 +1,455 @@
...
+        mlflow.sklearn.log_model(
+            prophet_model.model, "model", pip_requirements=[f"-r {req_file.strpath}", "b"]
+        )
+        _assert_pip_requirements(
+            mlflow.get_artifact_uri("model"), ["mlflow", "a", "b"], strict=True
+        )
+
+    # Constraints file
+    with mlflow.start_run():
+        mlflow.sklearn.log_model(
```
dbczumar, line 327: Looks like we're accidentally logging the models with `mlflow.sklearnn.log_model()` here, rather than `mlflow.prophet.log_model()`. Seems like pickling works here though :D

dbczumar (commented):
@BenWilson2 Awesome stuff! I tried this out using the dataset from the Prophet quickstart (http://facebook.github.io/prophet/docs/quick_start.html) and confirmed that:

1. Prophet models can be logged & loaded using MLflow
2. Prophet model inference succeeds with both the native Prophet and pyfunc model representations
3. Prophet models can be served and scored successfully
4. Prophet models can be deployed for model serving on Databricks

I left a few initial comments on the tests / implementation. @harupy @WeichenXu123 can you take a look as well? I'd imagine we can get this merged soon!

One other thing we'll want to do is add the flavor to the docs here: https://mlflow.org/docs/latest/models.html#built-in-model-flavors.

As a follow-up task that @harupy can provide guidance on, we'll also probably want to add prophet to our cross version tests for backwards / forwards compatibility testing.

mlflow/prophet.py
```diff
@@ -0,0 +1,296 @@
...
+        "LunarCalendar",
+        "holidays",
+        "tqdm",
+        "cmdstanpy",
+        "matplotlib",
+        "setuptools-git",
+        "python-dateutil",
+        "pystan",
+        "prophet",
+    ]
```
harupy, line 54: Can we just include `prophet`?

mlflow/prophet.py
```diff
@@ -0,0 +1,296 @@
...
+                          - ``relative/path/to/local/model``
+                          - ``s3://my_bucket/path/to/model``
+                          - ``runs:/<mlflow_run_id>/run-relative/path/to/model``
+
+                          For more information about supported URI schemes, see
+                          `Referencing Artifacts <https://www.mlflow.org/docs/latest/tracking.html#
+                          artifact-locations>`_.
+
+        :return: A Prophet model instance
+        """
```
harupy, line 281: nit: indent

dev/extra-ml-requirements.txt
```diff
@@ -42,3 +42,14 @@ pyspark
...
+Cython
+convertdate
+LunarCalendar
+holidays
+tqdm
+cmdstanpy
+setuptools-git
+python-dateutil
+pystan~=2.19.1.1
+prophet==1.0.1
```
harupy, line 55:
```suggestion
prophet
```

Can we unpin `prophet`?

dev/extra-ml-requirements.txt
```diff
@@ -42,3 +42,14 @@ pyspark
...
+# Required by mlflow.prophet
+Cython
+convertdate
+LunarCalendar
+holidays
+tqdm
+cmdstanpy
+setuptools-git
+python-dateutil
+pystan~=2.19.1.1
```
harupy, line 54:
```suggestion
```

Can we remove these packages since they are installed with `prophet`?

examples/prophet/train.py
```diff
@@ -1,96 +1,49 @@
...
-        )
-
-    # Useful for multiple runs (only doing one run in this sample notebook)
-    with mlflow.start_run():
-        m = Prophet()
-        m.fit(df)
+    cv_metrics = performance_metrics(metrics_raw)
+    metrics = {}
+    for metric in metric_values:
+        metrics[metric] = cv_metrics[metric].mean()
```
harupy, line 38:
```suggestion
    metrics = {k: cv_metrics[k].mean() for k in metric_keys}
```

nit

examples/prophet/train.py
```diff
@@ -1,96 +1,49 @@
...
-        mlflow.log_metric("rmse", df_p.loc[0, "rmse"])
+forecast = loaded_model.predict(loaded_model.make_future_dataframe(60))
 
-        mlflow.pyfunc.log_model("model", conda_env=conda_env, python_model=FbProphetWrapper(m))
-        print(
-            "Logged model with URI: runs:/{run_id}/model".format(
-                run_id=mlflow.active_run().info.run_id
-            )
-        )
+print(f"forecast: ${forecast.head(30)}")
```
harupy, line 49:
```suggestion
print(f"forecast:\n${forecast.head(30)}")
```

nit

tests/prophet/test_prophet_model_export.py
BenWilson2: wow that was a big oversight. Of note: per their docs (and my testing) pickling works just fine so long as the version of pystan is consistent with the underlying built solvers - but if pystan version is different it blows up. However, if the ser/deser that's used with their approach is used, then the environment won't matter and reproducibility can work fine. Seems like the root of all of these woes is in pystan from what I can tell.

tests/prophet/test_prophet_model_export.py
BenWilson2: Changed

tests/prophet/test_prophet_model_export.py
BenWilson2: Removed

mlflow/prophet.py
BenWilson2: Definitely can! Changed.

mlflow/prophet.py
BenWilson2: Added some more notes to this comment. This was due to the exhaustive permutation testing that I did by inserting dependency after dependency to the install chain until I got it working. The only passing install conf was the one where I explicitly defined all base requirements to the process to get the build to work (NOTE: setup.py _still fails_ when the grouped pip install occurs for serving - it only succeeds after running a build retry when the remainder of the env is already installed. The install process for Prophet has direct import statements for all of those packages.

mlflow/prophet.py
BenWilson2: Removed this dead code

mlflow/prophet.py
BenWilson2: Lingering debugging red herring ;) removed.

mlflow/prophet.py
BenWilson2: Unfortunately, if we remove all of the dependencies for prophet it simply won't build. I went through exhaustive testing by removing each one, but the hard-dependencies in pystan's install on Cython mean that it won't build (gcc-g++ errors) if Cython isn't present before package installation and build. Similarly, if all of the other packages aren't already in the path prior to prophet install it won't install at all.

mlflow/prophet.py
BenWilson2: fixed

dev/extra-ml-requirements.txt
BenWilson2: changed

dev/extra-ml-requirements.txt
BenWilson2: As mentioned above, the build process for Prophet will fail if these aren't on the path prior to pystan and prophet being installed. They're using a rather non-standard way for installing dependencies in their lib and not having them present will cause the pip install of prophet and pystan to fail.

examples/prophet/train.py
BenWilson2: changed

mlflow/prophet.py
dbczumar: +1! Perhaps an inline comment would help here to clarify for future readers.

dev/extra-ml-requirements.txt
dbczumar: let's add an inline comment here to clarify why we're installing all of these dependencies for future readers

CHANGELOG.rst
```diff
@@ -1,5 +1,6 @@
 Changelog
 =========
+
```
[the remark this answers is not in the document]
BenWilson2: Reverted

examples/prophet/conda.yaml
```diff
@@ -1,9 +1,19 @@
...
+name: prophet-example
 channels:
   - conda-forge
 dependencies:
   - python=3.6
   - pip
   - pip:
-    - mlflow>=1.0
-    - fbprophet<=0.7
+    - mlflow>=1.20.0
```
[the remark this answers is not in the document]
BenWilson2: updated

examples/prophet/train.py
```diff
@@ -1,96 +1,49 @@
...
-        # Print out metrics
-        print("Prophet model (rolling_window=%f):" % (rolling_window))
-        print("  CV: \n%s" % df_cv.head())
-        print("  Perf: \n%s" % df_p.head())
+loaded_model = mlflow.prophet.load_model(model_uri)
 
-        # Log parameter, metrics, and model to MLflow
-        mlflow.log_param("rolling_window", rolling_window)
-        mlflow.log_metric("rmse", df_p.loc[0, "rmse"])
+forecast = loaded_model.predict(loaded_model.make_future_dataframe(60))
```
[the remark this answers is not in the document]
BenWilson2: added!

mlflow/prophet.py
```diff
@@ -0,0 +1,296 @@
...
+                              Pandas DataFrame and then serialized to json using the
+                              Pandas split-oriented format. Bytes are base64-encoded.
+
+        :param await_registration_for: Number of seconds to wait for the model version
+                            to finish being created and is in ``READY`` status.
+                            By default, the function waits for five minutes.
+                            Specify 0 or None to skip waiting.
+        :param pip_requirements: {{ pip_requirements }}
+        :param extra_pip_requirements: {{ extra_pip_requirements }}
+        :param kwargs: kwargs to pass to `Prophet.save_model`_ method.
```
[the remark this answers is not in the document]
BenWilson2: cleaned up (and fixed indent)

mlflow/prophet.py
BenWilson2: added comments in both locations with a note for a future pseudo TODO (if the build process gets refactored for Prophet, then remove all of the declared dependencies)

dev/extra-ml-requirements.txt
BenWilson2: added comment

tests/prophet/test_prophet_model_export.py
```diff
@@ -0,0 +1,414 @@
...
+
+@pytest.mark.small
+def test_model_save_without_specified_conda_env_uses_default_env_with_expected_dependencies(
+    prophet_model, model_path
+):
+    mlflow.prophet.save_model(prophet_model.model, model_path)
+    _assert_pip_requirements(model_path, mlflow.prophet.get_default_pip_requirements())
+
+
+@pytest.mark.small
```
dbczumar, line 372: @BenWilson2 Can we mark these as large by removing `@pytest.mark.small`? I realize that there may be similar tests that are miscategorized as `small` in other flavors.

dbczumar (approved):
@BenWilson2 LGTM once your commits are signed to get the DCO check passing, the docs error shown here is fixed - https://app.circleci.com/pipelines/github/mlflow/mlflow/7647/workflows/efd9b390-90cd-466d-a848-c90ef4eee4be/jobs/11613, and https://github.com/mlflow/mlflow/pull/4773/files#r704924056 is addressed. Awesome stuff! Great work!

dev/extra-ml-requirements.txt
harupy: Got it. If so, we need to install these before installing `prophet`, right?

```
pip install Cython convertdate ...
pip install -r dev/extra-ml-requirements.txt
```

dbczumar: Was hoping that we could pass ordered packages to the same `pip install` command. Looks like this fails with ```DEPRECATION: prophet was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. A possible replacement is to fix the wheel build issue reported above. You can find discussion regarding this at https://github.com/pypa/pip/issues/8368.
Successfully installed Cython-0.29.24 LunarCalendar-0.0.9 absl-py-0.13.0 aiohttp-3.7.4.post0 applicationinsights-0.11.10 astor-0.8.1 astunparse-1.6.3 async-timeout-3.0.1 attrs-21.2.0 azure-identity-1.6.1 azure-mgmt-resource-8.0.1 azureml-automl-core-1.2.0 azureml-core-1.2.0.post3 azureml-dataprep-1.3.6 azureml-dataprep-native-14.2.1 azureml-pipeline-1.2.0 azureml-pipeline-core-1.2.0 azureml-pipeline-steps-1.2.0 azureml-sdk-1.2.0 azureml-telemetry-1.2.0 azureml-train-1.2.0 azureml-train-automl-client-1.2.0 azureml-train-core-1.2.0 azureml-train-restclients-hyperdrive-1.2.0 beautifulsoup4-4.10.0 blis-0.7.4 bottleneck-1.3.2 cached-property-1.5.2 catalogue-2.0.6 catboost-0.26.1 chardet-4.0.0 clang-5.0 click-7.1.2 cmdstanpy-0.9.68 colorama-0.4.4 contextvars-2.4 convertdate-2.3.2 cymem-2.0.5 decorator-5.0.9 distro-1.6.0 dotnetcore2-2.1.21 ephem-4.0.0.2 fastai-1.0.60 fastprogress-1.0.0 flake8-3.7.9 fsspec-2021.8.1 fusepy-3.0.1 future-0.18.2 gast-0.4.0 google-auth-1.35.0 google-auth-oauthlib-0.4.6 google-pasta-0.2.0 graphviz-0.8.4 grpcio-1.40.0 h2o-3.32.1.7 h5py-3.1.0 hijri-converter-2.2.1 holidays-0.11.2 idna-ssl-1.1.0 immutables-0.16 keras-2.6.0 keras-preprocessing-1.1.2 korean-lunar-calendar-0.2.1 lightgbm-3.2.1 llvmlite-0.36.0 markdown-3.3.4 mccabe-0.6.1 msal-1.14.0 msal-extensions-0.3.0 multidict-5.1.0 murmurhash-1.0.5 mxnet-1.8.0.post0 numba-0.53.1 numexpr-2.7.3 nvidia-ml-py3-7.352.0 onnxruntime-1.8.1 opt-einsum-3.3.0 paddlepaddle-2.1.2 pathy-0.6.0 patsy-0.5.1 portalocker-1.7.1 preshed-3.0.5 prophet-1.0.1 py4j-0.10.9 pyDeprecate-0.3.1 pycodestyle-2.5.0 pydantic-1.8.2 pyflakes-2.1.1 pymeeus-0.5.11 pyspark-3.1.2 pystan-2.19.1.1 pytorch-lightning-1.4.5 ruamel.yaml-0.15.89 setuptools-git-1.2 shap-0.39.0 six-1.15.0 slicer-0.0.7 smart-open-5.2.1 soupsieve-2.2.1 spacy-3.1.2 spacy-legacy-3.0.8 srsly-2.4.1 statsmodels-0.12.2 tensorboard-2.6.0 tensorboard-data-server-0.6.1 tensorboard-plugin-wit-1.8.0 tensorflow-2.6.0 tensorflow-estimator-2.6.0 termcolor-1.1.0 thinc-8.0.10 torch-1.9.0 torchmetrics-0.5.1 torchvision-0.10.0 tqdm-4.62.2 typer-0.3.2 typing-extensions-3.7.4.3 ujson-4.1.0 wasabi-0.8.2 wrapt-1.12.1 xgboost-1.4.2 yarl-1.6.3```

However, prophet installation does ultimately succeed. Thoughts @BenWilson2 ?

harupy:
@BenWilson2 cc @dbczumar We can ignore the `build_doc_r` failure since it's unrelated to this PR.

dev/extra-ml-requirements.txt
harupy: https://github.com/mlflow/mlflow/pull/4773/checks?check_run_id=3550111462#step:5:1038

```
    Running setup.py install for prophet: started
    Running setup.py install for prophet: still running...
    Running setup.py install for prophet: finished with status 'done'
  DEPRECATION: prophet was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. A possible replacement is to fix the wheel build issue reported above. You can find discussion regarding this at https://github.com/pypa/pip/issues/8368.
```

The error log above indicates that pip attempted to install `prophet` by running `setup.py install`. I guess prophet's dependencies are already installed when running `setup.py install`.

harupy: The installation order is determined by `get_installation_order`:

https://github.com/pypa/pip/blob/5ab7c124db917625e55693773bcc224793d7ceed/src/pip/_internal/resolution/resolvelib/resolver.py#L185

```python
    def get_installation_order(
        self, req_set: RequirementSet
    ) -> List[InstallRequirement]:
        """Get order for installation of requirements in RequirementSet.

        The returned list contains a requirement before another that depends on
        it. This helps ensure that the environment is kept consistent as they
        get installed one-by-one.
        ...
        """
```

harupy: `prophet` can be installed by simply running `pip install prophet`:

```dockerfile
# Dockerfile

FROM python:3.7

RUN pip install prophet
RUN python -c "import prophet"

# `docker build .` succeeds
```

dbczumar: Interesting. IIRC we had trouble with just including `prophet` in the list of dependencies. Conda environment creation failed. @harupy @BenWilson2 can you try this out with just `prophet` and confirm that it fails or succeeds?

harupy:
```dockerfile
FROM continuumio/miniconda3

RUN echo '\
name: prophet\n\
channels:\n\
  - conda-forge\n\
dependencies:\n\
  - python=3.6\n\
  - pip\n\
  - pip:\n\
      - prophet\n\
' > conda.yaml
RUN cat conda.yaml
RUN conda env create --file conda.yaml
```

failed with:

```
  ModuleNotFoundError: No module named 'numpy'
  ----------------------------------------
  ERROR: Failed building wheel for prophet
    ERROR: Command errored out with exit status 1:
     command: /opt/conda/envs/prophet/bin/python -u -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-spruk13i/prophet_286c94955d8f495093472c3bf4fb6f23/setup.py'"'"'; __file__='"'"'/tmp/pip-install-spruk13i/prophet_286c94955d8f495093472c3bf4fb6f23/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-x79jrygt/install-record.txt --single-version-externally-managed --compile --install-headers /opt/conda/envs/prophet/include/python3.6m/prophet
         cwd: /tmp/pip-install-spruk13i/prophet_286c94955d8f495093472c3bf4fb6f23/
    Complete output (10 lines):
    running install
    running build
    running build_py
    creating build
    creating build/lib
    creating build/lib/prophet
    creating build/lib/prophet/stan_model
    Importing plotly failed. Interactive plots will not work.
    INFO:pystan:COMPILING THE C++ CODE FOR MODEL anon_model_dfdaf2b8ece8a02eb11f050ec701c0ec NOW.
    error: command 'gcc' failed with exit status 1
    ----------------------------------------
ERROR: Command errored out with exit status 1: /opt/conda/envs/prophet/bin/python -u -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-spruk13i/prophet_286c94955d8f495093472c3bf4fb6f23/setup.py'"'"'; __file__='"'"'/tmp/pip-install-spruk13i/prophet_286c94955d8f495093472c3bf4fb6f23/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-x79jrygt/install-record.txt --single-version-externally-managed --compile --install-headers /opt/conda/envs/prophet/include/python3.6m/prophet Check the logs for full command output.


CondaEnvException: Pip failed
```

dbczumar: @harupy Any idea how the current approach seems to work? Does gcc get installed?

harupy: This doesn't work either:

```python
FROM continuumio/miniconda3

RUN echo '\
name: prophet\n\
channels:\n\
  - conda-forge\n\
dependencies:\n\
  - python=3.6\n\
  - pip\n\
  - pip:\n\
      - Cython\n\
      - convertdate\n\
      - LunarCalendar\n\
      - holidays\n\
      - tqdm\n\
      - cmdstanpy\n\
      - setuptools-git\n\
      - python-dateutil\n\
      - pystan~=2.19.1.1\n\
      - prophet\n\
' > conda.yaml

RUN conda env create --file conda.yaml
```

harupy: This works:

```dockerfile
FROM continuumio/miniconda3

RUN echo '\
name: prophet\n\
channels:\n\
  - conda-forge\n\
dependencies:\n\
  - python=3.6\n\
  - pip\n\
  - pip:\n\
      - prophet\n\
' > conda.yaml

RUN apt-get update --yes
# Install gcc
RUN apt-get install build-essential --yes
RUN conda env create --file conda.yaml
```

harupy: I guess what we really need to install is gcc, not prophet's dependencies.

dbczumar: @harupy it passes CI, including serving tests, if these extra dependencies are included, but we don’t seem to be installing gcc in the test environment. That’s the part I still don’t understand - why does it fail in your docker container & succeed in our CI system when gcc isn’t installed in either place?

harupy: In GitHub Actions, lots of packages are pre-installed:

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#installed-apt-packages

image

harupy: Confirmed gcc is installed on github actions:

https://github.com/mlflow/mlflow/pull/4786/checks?check_run_id=3554454258#step:3:958

```
$ which gcc
/usr/bin/gcc
```

harupy: Testing conda env creation with only including prophet in this PR:
https://github.com/mlflow/mlflow/pull/4786/

harupy:
@BenWilson2 Could you add update `dev/run-python-flavor-tests.sh` to run prophet tests like this commit: https://github.com/mlflow/mlflow/pull/4787/commits/0efd0ee1b08c040d55dc9eb928aecd0b143e028f?

tests/prophet/test_prophet_model_export.py
BenWilson2: certainly!

dev/extra-ml-requirements.txt
harupy: The following two sets of requirements are equivalent:

```
Cython
convertdate
LunarCalendar
holidays
tqdm
cmdstanpy
setuptools-git
python-dateutil
pystan
prophet
```

```
prophet
```

In either case, pip fails to build a wheel for prophet but manages to install prophet using the legacy `setup.py install` method.

BenWilson2: Did some testing with removing all of the dependencies (I actually think that the condition that I was dealing with when I saw those issues was due to copying the sklearn log_model declaration in test_pyfunc_serve_and_score. 
Thanks for the 2nd set of eyes on this and validating that the setup.py install is going to be defaulted to regardless of whether we have the build env all set before issuing the prophet pip install. 
Removing those sub dependencies.

harupy (approved):
LGTM!

PR #4795, opened 2021-09-13
Add experiment tags to experiment creation APIs

apurva-koti:

Follow up on #4788, add `experiment_tag` parameters to the `create_experiment` APIs for file, REST and SQL stores. Additionally update the python server handlers and Java and R clients accordingly.

## How is this patch tested?

New unit tests

Experiment tags can now be passed into new experiments upon creation.

Components 

Interface 

Language 

Integrations


mlflow/tracking/client.py
```diff
@@ -461,12 +461,19 @@ def get_experiment_by_name(self, name: str) -> Optional[Experiment]:
...
+        name: str,
+        artifact_location: Optional[str] = None,
+        experiment_tags: Optional[Dict[str, Any]] = None,
+    ) -> str:
         """Create an experiment.
 
         :param name: The experiment name. Must be unique.
         :param artifact_location: The location to store run artifacts.
                                   If not provided, the server picks an appropriate default.
+        :param experiment_tags: A dictionary of key-value pairs that are converted into
```
dbczumar, line 475: Can we clarify that these tags are set when the experiment is created?

dbczumar (commented):
@apurva-koti looks like the diff includes a new binary file: `tests/resources/db/db_version_7ac759974ad8_with_metrics.sql`. Can this be removed?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -235,6 +235,14 @@ def create_experiment(self, name, artifact_location=None):
...
                     artifact_location=artifact_location,
                 )
+                experiment_tags_dict = {}
+                if experiment_tags:
+                    for tag in experiment_tags:
+                        experiment_tags_dict[tag.key] = tag.value
+                experiment.tags = [
+                    SqlExperimentTag(key=key, value=value)
+                    for key, value in experiment_tags_dict.items()
+                ]
```
dbczumar, line 245: Do we need to convert the list to a dictionary and convert the dictionary back to a list?

mlflow/tracking/_tracking_service/client.py
```diff
@@ -162,17 +162,25 @@ def get_experiment_by_name(self, name):
...
         _validate_experiment_name(name)
         _validate_experiment_artifact_location(artifact_location)
-        return self.store.create_experiment(name=name, artifact_location=artifact_location,)
+        experiment_tags = experiment_tags if experiment_tags else {}
+
+        return self.store.create_experiment(
+            name=name,
+            artifact_location=artifact_location,
+            experiment_tags=[ExperimentTag(key, value) for (key, value) in experiment_tags.items()],
+        )
```
dbczumar, line 183: Do we have a test case for this?

mlflow/store/tracking/sqlalchemy_store.py
apurva-koti: Oh my god, I just realized that 😂 I copied this from the `create_run` code so I will optimize that one as well

mlflow/tracking/_tracking_service/client.py
apurva-koti: Added a unit test to `test_client.py` that ensures the mock store's `create_experiment` function was called with the `ExperimentTag` proto object

mlflow/R/mlflow/R/tracking-experiments.R
```diff
@@ -5,17 +5,19 @@
 #' @param name The name of the experiment to create.
 #' @param artifact_location Location where all artifacts for this experiment are stored. If
 #'   not provided, the remote server will select an appropriate default.
+#' @param tags Experiment tags to set on the experiment upon experiment creation.
 #' @template roxlate-client
 #' @export
-mlflow_create_experiment <- function(name, artifact_location = NULL, client = NULL) {
+mlflow_create_experiment <- function(name, artifact_location = NULL, tags = NULL, client = NULL) {
```
dbczumar, line 11: This is a backwards-incompatible change due to argument ordering, right? At least it would be in Python. `tags` need to go at the end.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
```diff
@@ -286,6 +286,26 @@ public String createExperiment(String experimentName) {
...
   }
 
+  /**
+   * Create a new experiment. This method allows providing all possible
+   * fields of CreateExperiment, and can be invoked as follows:
+   *
+   *   <pre>
+   *   import org.mlflow.api.proto.Service.CreateExperiment;
+   *   CreateExperiment.Builder request = CreateExperiment.newBuilder();
+   *   request.setName(name);
```
dbczumar, line 296: shall we add `setTags` to the example here too?

mlflow/tracking/_tracking_service/client.py
```diff
@@ -162,17 +162,25 @@ def get_experiment_by_name(self, name):
...
         """
         _validate_experiment_name(name)
         _validate_experiment_artifact_location(artifact_location)
-        return self.store.create_experiment(name=name, artifact_location=artifact_location,)
+        experiment_tags = experiment_tags if experiment_tags else {}
+
+        return self.store.create_experiment(
+            name=name,
+            artifact_location=artifact_location,
+            experiment_tags=[ExperimentTag(key, value) for (key, value) in experiment_tags.items()],
```
dbczumar, line 182: Nit: Also probably don't need the conversion to dictionary and back to a list here. We can structure this as 

```
experiment_tags = experiment_tags or []
...
```

mlflow/R/mlflow/R/tracking-experiments.R
apurva-koti: I was trying to follow the order of arguments specified in the proto; not sure how R works, but wouldn't moving `tags` to the end break fluent usage? (ie. not specifiying `client`)

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
dbczumar: Can we add tests for this API?

mlflow/R/mlflow/R/tracking-experiments.R
dbczumar: moving tags to the end shouldn't break fluent usage. If client is unspecified, it'll still be NULL. Putting `tags` before client breaks usage patterns that specify client positionally. I.e. if I passed in the arguments ("foo", "location", client) in MLflow <= 1.20.2, client would start being interpreted as a collection of tags in MLflow > 1.20.2 once this change was released.

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
apurva-koti: Done - weird proto-based syntax derived from https://stackoverflow.com/a/66740363/13837474

mlflow/R/mlflow/R/tracking-experiments.R
apurva-koti: Right - but then wouldn't a future fluent usage of `create_experiment(name, location, tags)` be broken since tags would be interpreted as a client, if no keyword arguments are specified?

dbczumar: Future users of this updated API must specify either both tags and client positionally or pass tags as a named argument. This is better than enacting a backwards-incompatible change that breaks existing workflows.

apurva-koti: Fair enough! let me make that change

mlflow/java/client/src/main/java/org/mlflow/tracking/MlflowClient.java
apurva-koti: Done

dbczumar (approved):
LGTM, thanks @apurva-koti ! Can we test this on Databricks as well?

apurva-koti:
@dbczumar Let's do it :D

PR #4823, opened 2021-09-17
Fix spark session initialization in requirement inference for spark models

harupy:

Fix spark session initialization in requirement inference for spark models.

## How is this patch tested?

Installed MLflow from this PR branch on Databricks and verify a new spark session created in a subprocess does not affect the existing one.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/_capture_modules.py
+++ b/mlflow/utils/_capture_modules.py
@@ -83,16 +83,21 @@ def main():
     # Mirror `sys.path` of the parent process
     sys.path = json.loads(args.sys_path)
 
+    spark = None
     if flavor == mlflow.spark.FLAVOR_NAME and is_in_databricks_runtime():
-        try:
-            # pylint: disable=import-error
-            from dbruntime.spark_connection import initialize_spark_connection
+        from pyspark.sql import SparkSession
 
-            initialize_spark_connection()
-        except Exception as e:
-            raise Exception(
-                "Attempted to initialize a spark session to load the spark model, but failed"
-            ) from e
+        # Clear 'PYSPARK_GATEWAY_PORT' and 'PYSPARK_GATEWAY_SECRET' to enforce launching a new JVM
+        # gateway
+        os.environ.pop("PYSPARK_GATEWAY_PORT", None)
+        os.environ.pop("PYSPARK_GATEWAY_SECRET", None)
+        os.environ["SPARK_DIST_CLASSPATH"] = "/databricks/jars/*"
+        spark = (
+            SparkSession.builder.master("local")
+            .appName(__name__)
+            .config("spark.ui.enabled", "false")
+            .getOrCreate()
+        )
 
     cap_cm = _CaptureImportedModules()
 
@@ -119,6 +124,9 @@ def main():
 
     # Store the imported modules in `output_file`
     write_to(args.output_file, "\n".join(cap_cm.imported_modules))
+
+    if spark:
+        spark.stop()
 
 
 if __name__ == "__main__":
--- a/mlflow/utils/_capture_modules.py
+++ b/mlflow/utils/_capture_modules.py
@@ -125,6 +125,5 @@ def main():
             pass
 
 
-
 if __name__ == "__main__":
     main()
--- a/mlflow/utils/_capture_modules.py
+++ b/mlflow/utils/_capture_modules.py
@@ -83,7 +83,6 @@ def main():
     # Mirror `sys.path` of the parent process
     sys.path = json.loads(args.sys_path)
 
-    spark = None
     if flavor == mlflow.spark.FLAVOR_NAME and is_in_databricks_runtime():
         # Clear 'PYSPARK_GATEWAY_PORT' and 'PYSPARK_GATEWAY_SECRET' to enforce launching a new JVM
         # gateway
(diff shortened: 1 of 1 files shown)


mlflow/utils/_capture_modules.py
```diff
@@ -83,16 +83,21 @@ def main():
     # Mirror `sys.path` of the parent process
     sys.path = json.loads(args.sys_path)
 
+    spark = None
     if flavor == mlflow.spark.FLAVOR_NAME and is_in_databricks_runtime():
```
WeichenXu123, line 87: shall we remove the `is_in_databricks_runtime()` condition here ?

mlflow/utils/_capture_modules.py
```diff
@@ -120,6 +125,9 @@ def _load_pyfunc_patch(*args, **kwargs):
     # Store the imported modules in `output_file`
     write_to(args.output_file, "\n".join(cap_cm.imported_modules))
 
+    if spark:
+        spark.stop()
```
WeichenXu123, line 129: use 
```
try:
  spark.stop()
catch Exception:
  pass
```
would be better, to swallow minor exceptions

mlflow/utils/_capture_modules.py
harupy: why?

harupy: The code in this if caluse only makes sense in Databricks runtime, doesn't it?

WeichenXu123: Make sense.
I think we can remove lines of:
```
        spark = (
            SparkSession.builder.master("local")
            .appName(__name__)
            .config("spark.ui.enabled", "false")
            .getOrCreate()
        )
```
later in `mlflow.spark._load_pyfunc` it will execute similar create session code.

mlflow/utils/_capture_modules.py
```diff
@@ -83,16 +83,21 @@ def main():
...
+        # gateway
+        os.environ.pop("PYSPARK_GATEWAY_PORT", None)
+        os.environ.pop("PYSPARK_GATEWAY_SECRET", None)
+        os.environ["SPARK_DIST_CLASSPATH"] = "/databricks/jars/*"
+        spark = (
+            SparkSession.builder.master("local")
+            .appName(__name__)
+            .config("spark.ui.enabled", "false")
+            .getOrCreate()
+        )
```
WeichenXu123, line 100: https://github.com/mlflow/mlflow/pull/4823/files#r713672191

mlflow/utils/_capture_modules.py
harupy: Makes sense. Removed the lines to create a spark session!

mlflow/utils/_capture_modules.py
harupy: Done!

mlflow/utils/_capture_modules.py
```diff
@@ -120,6 +117,13 @@ def _load_pyfunc_patch(*args, **kwargs):
     # Store the imported modules in `output_file`
     write_to(args.output_file, "\n".join(cap_cm.imported_modules))
 
+    if spark:
```
WeichenXu123, line 120: We should add a line before this:
```
spark = pyspark.sql.SparkSession._instantiatedSession
```
and remove L86 line `spark = None`

mlflow/utils/_capture_modules.py
```diff
@@ -83,16 +83,13 @@ def main():
     # Mirror `sys.path` of the parent process
     sys.path = json.loads(args.sys_path)
 
+    spark = None
```
WeichenXu123, line 86: remove this line: https://github.com/mlflow/mlflow/pull/4823/files#r714627969

mlflow/utils/_capture_modules.py
```diff
@@ -83,16 +83,13 @@ def main():
...
-            initialize_spark_connection()
-        except Exception as e:
-            raise Exception(
-                "Attempted to initialize a spark session to load the spark model, but failed"
-            ) from e
+        # Clear 'PYSPARK_GATEWAY_PORT' and 'PYSPARK_GATEWAY_SECRET' to enforce launching a new JVM
+        # gateway
+        os.environ.pop("PYSPARK_GATEWAY_PORT", None)
+        os.environ.pop("PYSPARK_GATEWAY_SECRET", None)
+        os.environ["SPARK_DIST_CLASSPATH"] = "/databricks/jars/*"
```
WeichenXu123, line 92: @Ngone51 @HyukjinKwon Could you help review this part of the PR ?

Context:
The code here will be run in databricks environment, and it is in a **subprocess** of pyspark driver process.
In the subprocess, later it will create a **local spark session**, the 3 lines here is doing some preparation work:

(1)
```
        os.environ.pop("PYSPARK_GATEWAY_PORT", None)
        os.environ.pop("PYSPARK_GATEWAY_SECRET", None)
```
this is for cleaning gateway env variables inherits from parent process, otherwise the py4j will connect to the existing driver py4j gateway.

(2)
```
os.environ["SPARK_DIST_CLASSPATH"] = "/databricks/jars/*"
```
This line is because, databricks runtime put all runtime jars in `/databricks/jars/*`
and `SPARK_DIST_CLASSPATH` is for setting additional jars path https://github.com/databricks/runtime/blob/d8d2022d1e0a57644a71ff930ffbdc5528d24fca/bin/spark-class#L54

HyukjinKwon: wow complicated .. Yes, setting `PYSPARK_GATEWAY_PORT` and `PYSPARK_GATEWAY_SECRET` to `None` would work here, this part loos fine. But I wonder if we can just simply stick to using threads instead of processes. That would make our lives easier ..

WeichenXu123: @HyukjinKwon
Clarification: this PR is not related to the reuse py4j gateway server issue.
We create subprocess for capturing python package dependencies.
and here we create spark session of "local mode", which has nothing related to the existing py4j gateway server.

WeichenXu123 (approved):
LGTM

mlflow/utils/_capture_modules.py
HyukjinKwon: LGTM

PR #4856, opened 2021-09-29
Fix autologging event logger behavior when patch function failed before original function called

WeichenXu123:

When patch function raise exception before original function called, the expected behavior should be:
(1) if following original function call succeeds, then this is unexpected mlflow error, we should call `log_patch_function_error` and `log_original_function_success`
(2) if following original function call raise error, then this is **NOT** unexpected mlflow error we should **NOT** call `log_patch_function_error`, but we should call `log_original_function_error`

In contrast, the current behavior is:
(1) if following original function call succeeds, it forget to call `log_original_function_success`
(2) if following original function call raise error, then it still call `log_patch_function_error`, and it forget call `log_original_function_error`

## How is this patch tested?

Unit test.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/autologging_utils/safety.py
+++ b/mlflow/utils/autologging_utils/safety.py
@@ -559,7 +559,36 @@ def safe_patch(
                     return original_result
                 else:
                     try:
-                        call_original(*args, **kwargs)
+                        try_log_autologging_event(
+                            AutologgingEventLogger.get_logger().log_original_function_start,
+                            session,
+                            destination,
+                            function_name,
+                            args,
+                            kwargs,
+                        )
+                        original_result = original(*args, **kwargs)
+                        try_log_autologging_event(
+                            AutologgingEventLogger.get_logger().log_original_function_success,
+                            session,
+                            destination,
+                            function_name,
+                            args,
+                            kwargs,
+                        )
+                        return original_result
+                    except Exception as e:
+                        try_log_autologging_event(
+                            AutologgingEventLogger.get_logger().log_original_function_error,
+                            session,
+                            destination,
+                            function_name,
+                            args,
+                            kwargs,
+                            e,
+                        )
+                        failed_during_original = True
+                        raise
                     finally:
                         if not failed_during_original:
                             try_log_autologging_event(
--- a/mlflow/utils/autologging_utils/safety.py
+++ b/mlflow/utils/autologging_utils/safety.py
@@ -554,58 +554,58 @@ def safe_patch(
                         _validate_autologging_run(
                             autologging_integration, patch_function_run_for_testing.info.run_id
                         )
-
-                if original_has_been_called:
-                    return original_result
-                else:
-                    try:
-                        try_log_autologging_event(
-                            AutologgingEventLogger.get_logger().log_original_function_start,
-                            session,
-                            destination,
-                            function_name,
-                            args,
-                            kwargs,
-                        )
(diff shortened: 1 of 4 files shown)


tests/autologging/test_autologging_safety_unit.py
```diff
@@ -1627,3 +1627,47 @@ def autolog(disable=False, exclusive=False, silent=False):  # pylint: disable=un
     original_predict = Test1.predict
     autolog()
     assert get_func_attrs(Test1.predict) == get_func_attrs(original_predict)
+
+
+def test_log_behaivor_when_patch_fn_raise_error_before_original_called():
```
dbczumar, line 1632:
```suggestion
def test_log_behavior_when_patch_fn_raise_error_before_original_called():
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -442,6 +442,7 @@ def try_log_autologging_event(log_fn, *args):
                     )
 
             with _AutologgingSessionManager.start_session(autologging_integration) as session:
+                patch_function_exception = None
```
dbczumar, line 445: Nit: Can we move this up to lines 424-434 so that it's located with the other local variables used for tracking autologging state? Can we add an inline comment explaining its role like we do for the other variables?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -573,7 +558,54 @@ def call_original(*og_args, **og_kwargs):
...
+                        if not failed_during_original:
+                            try_log_autologging_event(
+                                AutologgingEventLogger.get_logger().log_patch_function_error,
+                                session,
+                                destination,
+                                function_name,
+                                args,
+                                kwargs,
+                                patch_function_exception,
+                            )
```
dbczumar, line 602: Shouldn't we only do this if `patch_function_exception` is not `None`?  Can we add test coverage to make sure we aren't emitting this event unless there's a patch function exception?

tests/autologging/fixtures.py
```diff
@@ -29,6 +29,9 @@ def recursive_fn(self, level, max_depth):
             else:
                 return self.recursive_fn(level + 1, max_depth)
 
+        def bad_fn(self, error_to_raise):
```
harupy, line 32:
```suggestion
        def throw_fn(self, error_to_raise):
```

Can we rename this function to clarify what it does?

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -662,30 +663,36 @@ def patch_impl(original, *args, **kwargs):
...
         "patch_error",
     ]
-    assert [call.method for call in mock_event_logger.calls] == expected_order_throw_after
-    patch_start, original_start, original_success, patch_error = mock_event_logger.calls
-    assert patch_start.exception is original_start.exception is None
-    assert original_success.exception is None
-    assert patch_error.exception == exc_to_raise
+
+    expected_order_bad_fn = ["patch_start", "original_start", "original_error"]
+    throw_location = "before"
```
harupy, line 676: Are we using this variable somewhere?

harupy: Ah it's used in `patch_impl`.

dbczumar (approved):
LGTM! Thanks @WeichenXu123 !

harupy (approved):
LGTM

PR #4877, opened 2021-10-07
Throw Exception When Invalid URI with Databricks Scheme Provided

yunpark93:
Signed-off-by: Yun Park <yun@databricks.com>

When a user uses a single slash in the URI when referencing the registry URI in Databricks, e.g. databricks:/profile:prefix, it incorrectly references the local registry. The user most likely intended to not use the local registry but the registry provided with a correctly formatted URI.

Screen Shot 2021-10-07 at 1 57 44 PM

## How is this patch tested?

(Details)

Tracking and registry URIs provided to the client that reference a Databricks endpoint (useing the `databricks` URI scheme) must be correctly formatted using 2 slashes after the colon instead of 1. Previously, if URIs such as `databricks://`, `databricks:/`, `databricks:/default` or `databricks:/scope:prefix` were provided, they defaulted to the local profile (in Databricks, this would be the default tracking server or registry in the workspace). This will no longer be the case, and you will instead get an error about an incorrectly formatted URI.

Components 
- [] `area/server-infra`: MLflow Tracking server backend

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/uri.py
+++ b/mlflow/utils/uri.py
@@ -69,7 +69,7 @@ def get_db_info_from_uri(uri):
     parsed_uri = urllib.parse.urlparse(uri)
     if parsed_uri.scheme == "databricks":
         # netloc should not be an empty string unless URI is formatted incorrectly.
-        if parsed_uri.netloc == '':
+        if parsed_uri.netloc == "":
             raise MlflowException(
                 "URI is formatted incorrectly: no netloc in URI '%s'." % parsed_uri.netloc
             )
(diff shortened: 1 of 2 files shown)


ankit-db (approved):
One small unit test I'd recommend adding

sueann (commented):
Could we add in the release notes which cases used to work (sort of as intended) but will now be given an error instead? I think `databricks:/` `databricks://` and `databricks:/default` (or `databricks:/databricks` ??) might be those? (should confirm the last two though)

sueann (approved):

PR #4946, opened 2021-10-26
[Proxied artifact operations] Implement REST API endpoints and artifact repository

harupy:

Implement REST API endpoints + artifact repository.

## How is this patch tested?

Unit tests

Components 

Interface 

Language 

Integrations


dbczumar (commented):
@harupy This looks like an excellent start! Awesome stuff!

examples/mlflow_artifacts/README.md
```diff
@@ -0,0 +1,34 @@
+This directory contains a set of files for demonstrating the mlflow artifacts service.
```
harupy, line 1: This is an end-to-end example to make sure everything works correctly.

mlflow/protos/mlflow_artifacts.proto
```diff
@@ -0,0 +1,86 @@
...
+      }],
+      visibility: PUBLIC,
+      rpc_doc_title: "Download Artifact",
+    };
+  }
+
+  rpc uploadArtifact (UploadArtifact) returns (UploadArtifact.Response) {
+    option (rpc) = {
+      endpoints: [{
+        method: "POST",
```
dbczumar, line 31: To comply with REST API conventions, I think this should be a `PUT` request instead of a `POST` request. See https://stackoverflow.com/questions/630453/what-is-the-difference-between-post-and-put-in-http#:~:text=POST%20means%20%22create%20new%22%20as,the%20server%20to%20create%20it.

mlflow/store/artifact/http_artifact_repo.py
```diff
@@ -0,0 +1,66 @@
...
+        sep = "/mlflow-artifacts/artifacts"
+        head, tail = self.artifact_uri.split(sep, maxsplit=1)
+        url = head + sep
+        tail = tail.lstrip("/")
+        params = {"path": posixpath.join(tail, path) if path else tail}
+        resp = requests.get(url, params=params)
+        resp.raise_for_status()
+        json = resp.json()
+        files = json.get("files")
+        if files:
```
dbczumar, line 47: Nit: 
```suggestion
        files = json.get("files", [])
        if files:
```

mlflow/protos/mlflow_artifacts.proto
harupy: Got it, thanks for the knowledge!

mlflow/cli.py
```diff
@@ -317,6 +317,9 @@ def _validate_static_prefix(ctx, param, value):  # pylint: disable=unused-argume
     "Default: Within file store, if a file:/ URI is provided. If a sql backend is"
     " used, then this option is required.",
 )
+@click.option(
+    "--artifacts-destination", metavar="URI", default="./artifacts", help="<Add description>",
```
dbczumar, line 321: How about `mlartifacts` in order to be consistent with `mlruns`?

```suggestion
    "--artifacts-destination", metavar="URI", default="./mlartifacts", help="<Add description>",
```

mlflow/server/handlers.py
```diff
@@ -799,6 +818,52 @@ def _delete_model_version_tag():
...
+                f.write(chunk)
+
+        artifact_repo = _get_artifact_repo_mlflow_artifacts()
+        artifact_repo.log_artifact(tmp_path, artifact_path=head or None)
+
+    return _wrap_response(UploadArtifact.Response())
+
+
+@catch_mlflow_exception
+def _list_artifacts_mlflow_artifacts():
```
dbczumar, line 851: Can we add short / simple docstrings for these methods, particularly `_list_artifacts_mlflow_artifacts()`, to explain what these handlers do?

mlflow/server/handlers.py
```diff
@@ -120,6 +128,17 @@ def _get_sqlalchemy_store(cls, store_uri):
 _model_registry_store_registry = ModelRegistryStoreRegistryWrapper()
 
 
+def _get_artifact_repo_mlflow_artifacts():
```
dbczumar, line 131: Can we add a docstring to this method explaining the use case?

mlflow/server/handlers.py
```diff
@@ -799,6 +818,52 @@ def _delete_model_version_tag():
     return _wrap_response(DeleteModelVersionTag.Response())
 
 
+@catch_mlflow_exception
+def _download_artifact(artifact_path):
+    basename = os.path.basename(artifact_path)
```
dbczumar, line 823: artifact paths for the `mlflow-artifacts` REST API are always given in POSIX path format, so I think we need to use `posixpath.basename` for parsing. When constructing local filesystem paths (e.g. 2 lines below), `os.path` is still appropriate.

mlflow/server/handlers.py
```diff
@@ -799,6 +818,52 @@ def _delete_model_version_tag():
...
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_path = os.path.join(tmp_dir, basename)
+        artifact_repo = _get_artifact_repo_mlflow_artifacts()
+        artifact_repo._download_file(artifact_path, tmp_path)
+        return send_from_directory(tmp_dir, basename)
+
+
+@catch_mlflow_exception
+def _upload_artifact(artifact_path):
+    head, tail = os.path.split(artifact_path)
```
dbczumar, line 833: Ditto here - I think we need `posixpath.split`

mlflow/server/handlers.py
```diff
@@ -799,6 +818,52 @@ def _delete_model_version_tag():
...
+
+@catch_mlflow_exception
+def _list_artifacts_mlflow_artifacts():
+    request_message = _get_request_message(ListArtifactsMlflowArtifacts())
+    response_message = ListArtifacts.Response()
+    path = request_message.path if request_message.HasField("path") else None
+    artifact_repo = _get_artifact_repo_mlflow_artifacts()
+    files = []
+    for file_info in artifact_repo.list_artifacts(path):
+        basename = os.path.basename(file_info.path)
```
dbczumar, line 858: I think we need `posixpath` here (same feedback as https://github.com/mlflow/mlflow/pull/4946/files#r740746815)

mlflow/store/artifact/http_artifact_repo.py
```diff
@@ -0,0 +1,66 @@
...
+                self.log_artifact(os.path.join(root, f), artifact_dir)
+
+    def list_artifacts(self, path=None):
+        sep = "/mlflow-artifacts/artifacts"
+        head, tail = self.artifact_uri.split(sep, maxsplit=1)
+        url = head + sep
+        tail = tail.lstrip("/")
+        params = {"path": posixpath.join(tail, path) if path else tail}
+        resp = requests.get(url, params=params)
+        resp.raise_for_status()
```
dbczumar, line 44: When we add tests for this artifact repo, it will be great to test this.

mlflow/store/artifact/http_artifact_repo.py
```diff
@@ -0,0 +1,66 @@
...
+            for f in files:
+                file_info = FileInfo(
+                    posixpath.join(path, f["path"]) if path else f["path"],
+                    f["is_dir"],
+                    f.get("file_size"),
+                )
+                file_infos.append(file_info)
+            return sorted(file_infos, key=lambda f: f.path)
+        else:
+            return []
```
dbczumar, line 58: We don't need the `if / else` now that we have `files = json.get("files", [])`, right?

mlflow/protos/mlflow_artifacts.proto
```diff
@@ -0,0 +1,86 @@
...
+}
+
+message UploadArtifact {
+  option (scalapb.message).extends = "com.databricks.rpc.RPC[$this.Response]";
+
+  message Response {}
+}
+
+message ListArtifacts {
+  option (scalapb.message).extends = "com.databricks.rpc.RPC[$this.Response]";
```
dbczumar, line 66: If we don't intend to implement these on Databricks, then I think we can remove the `(scalapb.message).extends` option from each of these protos.

mlflow/protos/mlflow_artifacts.proto
```diff
@@ -0,0 +1,86 @@
+syntax = "proto2";
+
+package mlflow.artifacts;
+
+import "scalapb/scalapb.proto";
+import "databricks.proto";
```
dbczumar, line 6: do we need to import `databricks.proto`?

mlflow/protos/mlflow_artifacts.proto
```diff
@@ -0,0 +1,92 @@
+syntax = "proto2";
+
+package mlflow.artifacts;
```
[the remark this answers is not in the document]
dbczumar: Does `.` have a special meaning in the proto DSL? I.e. does `mlflow.artifacts` indicate that this is a "sub package / submodule" of `mlflow`?

tests/tracking/test_mlflow_artifacts.py
```diff
@@ -0,0 +1,210 @@
...
+def upload_file(url, local_path):
+    with open(local_path, "rb") as f:
+        requests.put(url, data=f).raise_for_status()
+
+
+def download_file(url, local_path):
+    with requests.get(url, stream=True) as r:
+        r.raise_for_status()
+        with open(local_path, "wb") as f:
+            for chunk in r.iter_content(chunk_size=8192):
```
dbczumar, line 67: Where are we getting the chunk size of 3KB from? Should we potentially choose a larger chunk size?

mlflow/server/handlers.py
```diff
@@ -799,6 +818,52 @@ def _delete_model_version_tag():
...
+        return send_from_directory(tmp_dir, basename)
+
+
+@catch_mlflow_exception
+def _upload_artifact(artifact_path):
+    head, tail = os.path.split(artifact_path)
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_path = os.path.join(tmp_dir, tail)
+        with open(tmp_path, "wb") as f:
+            chunk_size = 8192
```
dbczumar, line 837: Where are we getting the chunk size of 3KB from? Should we potentially choose a larger chunk size?

examples/mlflow_artifacts/README.md
dbczumar: This is *extremely* cool!

mlflow/cli.py
dbczumar: Reminder to add a description :)

dbczumar (commented):
I tried this out locally by running `mlflow server` as follows:

```
$ mlflow server --artifacts-destination /tmp/bizfoobaz --default-artifact-root http://localhost:5000/api/2.0/mlflow-artifacts/artifacts
```

I then ran the following script:

```
import mlflow
mlflow.set_tracking_uri("http://localhost:5000")

class MyMod(mlflow.pyfunc.PythonModel):
    def predict(self, ctx, inp):
        return 7

with mlflow.start_run():
    mlflow.pyfunc.log_model(
        python_model=MyMod(),
        artifact_path="bizfoo",
    )
```

Logging succeeded :D! 

Finally, I launched the MLflow UI and attempted to view the logged artifacts. Unfortunately, this resulted in an error:

Screen Shot 2021-11-01 at 11 55 56 PM

Stacktrace from the server process:

```
2021/11/01 23:52:39 ERROR mlflow.server: Exception on /ajax-api/2.0/preview/mlflow/artifacts/list [GET]
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/corey.zumar/mlflow/mlflow/server/handlers.py", line 233, in wrapper
    return func(*args, **kwargs)
  File "/Users/corey.zumar/mlflow/mlflow/server/handlers.py", line 508, in _list_artifacts
    response_message.files.extend([a.to_proto() for a in artifact_entities])
  File "/Users/corey.zumar/mlflow/mlflow/server/handlers.py", line 508, in <listcomp>
    response_message.files.extend([a.to_proto() for a in artifact_entities])
  File "/Users/corey.zumar/mlflow/mlflow/entities/file_info.py", line 40, in to_proto
    proto.file_size = self.file_size
TypeError: '289' has type str, but expected one of: int, long
2021/11/01 23:55:52 ERROR mlflow.server: Exception on /ajax-api/2.0/preview/mlflow/artifacts/list [GET]
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/corey.zumar/mlflow/mlflow/server/handlers.py", line 233, in wrapper
    return func(*args, **kwargs)
  File "/Users/corey.zumar/mlflow/mlflow/server/handlers.py", line 508, in _list_artifacts
    response_message.files.extend([a.to_proto() for a in artifact_entities])
  File "/Users/corey.zumar/mlflow/mlflow/server/handlers.py", line 508, in <listcomp>
    response_message.files.extend([a.to_proto() for a in artifact_entities])
  File "/Users/corey.zumar/mlflow/mlflow/entities/file_info.py", line 40, in to_proto
    proto.file_size = self.file_size
TypeError: '289' has type str, but expected one of: int, long
```

We seem to be using an incorrect file_size representation in `http_artifact_repo.py` (https://github.com/mlflow/mlflow/pull/4946/files#diff-cd82ea3e604e2631623ace6d75186a42418a7936f804d653e46f78d618c3054cR53). Changing this from string to int locally seems to solve the problem, and the artifacts render correctly:

Screen Shot 2021-11-02 at 12 00 20 AM

Can we fix the representation and add a unit test?

dbczumar (commented):
@harupy This is so cool! Great progress!

mlflow/server/handlers.py
harupy: True, let me fix this.

mlflow/store/artifact/http_artifact_repo.py
harupy: Right, we can just return `files`.

mlflow/protos/mlflow_artifacts.proto
harupy: Makes senes. Let's remove them.

mlflow/protos/mlflow_artifacts.proto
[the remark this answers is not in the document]
harupy: https://developers.google.com/protocol-buffers/docs/style says:

> Package name should be in lowercase, and should correspond to the directory hierarchy. e.g., if a file is in my/package/, then the package name should be my.package.

Maybe we can use `mlflow-artifacts`

mlflow/protos/mlflow_artifacts.proto
harupy: It appears `databricks.proto` is required to use `option (rpc)`: 

```
$ ./generate-protos.sh

++ protoc --version
+ PROTOC_VERSION='libprotoc 3.6.0'
+ [[ libprotoc 3.6.0 != \l\i\b\p\r\o\t\o\c\ \3\.\6\.\0 ]]
+ PROTOS=mlflow/protos
+ protoc -I=mlflow/protos --python_out=mlflow/protos --java_out=mlflow/java/client/src/main/java mlflow/protos/databricks.proto mlflow/protos/service.proto mlflow/protos/model_registry.proto mlflow/protos/databricks_artifacts.proto mlflow/protos/mlflow_artifacts.proto mlflow/protos/scalapb/scalapb.proto
mlflow_artifacts.proto:16:12: Option "(rpc)" unknown.
mlflow_artifacts.proto:28:12: Option "(rpc)" unknown.
mlflow_artifacts.proto:40:12: Option "(rpc)" unknown.
```

mlflow/store/artifact/http_artifact_repo.py
harupy: Sure!

mlflow/cli.py
harupy: Thanks, I've updated the default value, I'll add the description soon!

mlflow/server/handlers.py
harupy: It's just a randomly-selected value. Using a larger chunk size does make sense.

harupy:
@dbczumar Thanks for catching the UI issue!

mlflow/protos/mlflow_artifacts.proto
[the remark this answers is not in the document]
dbczumar: Got it - I see you've crossed out "Maybe we can use mlflow-artifacts" - is this not supported? How about `mlflow_artifacts`?

mlflow/protos/mlflow_artifacts.proto
dbczumar: Got it! Thanks!

mlflow/server/handlers.py
dbczumar: Let's use a larger size! I think most network connections can handle it :D

mlflow/protos/mlflow_artifacts.proto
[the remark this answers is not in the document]
harupy: > Is this not supported?

No, `-` is not supported. I got this error with `mlflow-artifacts`:

```
$ ./generate-protos.sh
mlflow_artifacts.proto:3:15: Expected ";".
```

> How about `mlflow_artifacts`?

`mlflow_artifacts` didn't work either:

```
$ ./generate-protos.sh
mlflow_artifacts.proto:17:12: Option "(rpc)" unknown.
mlflow_artifacts.proto:29:12: Option "(rpc)" unknown.
mlflow_artifacts.proto:41:12: Option "(rpc)" unknown.
```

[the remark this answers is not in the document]
harupy: Can we change `package` to `mlflow` and prefix `FileInfo` and `ListArtifacts` with `MlflowArtifacts` 

```diff
- message ListArtifacts {
+ message MlflowArtifactsListArtifacts {
  // Filter artifacts matching this path (a relative path from the root artifact directory).
  optional string path = 1;

  message Response {
    // File location and metadata for artifacts.
    repeated FileInfo files = 1;
  }
}

- message FileInfo {
+ message MlflowArtifactsFileInfo {
  // Path relative to the root artifact directory run.
  optional string path = 1;

  // Whether the path is a directory.
  optional bool is_dir = 2;

  // Size in bytes. Unset for directories.
  optional int64 file_size = 3;
}
```

mlflow/server/handlers.py
harupy: Added simple docstrings!

mlflow/cli.py
```diff
@@ -317,6 +317,9 @@ def _validate_static_prefix(ctx, param, value):  # pylint: disable=unused-argume
     "Default: Within file store, if a file:/ URI is provided. If a sql backend is"
     " used, then this option is required.",
 )
+@click.option(
+    "--artifacts-destination", metavar="URI", default="./mlartifacts", help="<Add description>",
+)
```
harupy, line 322: Reminder to myself to add a description.

mlflow/server/handlers.py
```diff
@@ -799,6 +823,64 @@ def _delete_model_version_tag():
...
+    """
+    A request handler for `GET /mlflow-artifacts/artifacts/<artifact_path>` to download an artifact
+    from `artifact_path` (a relative path from the root artifact directory).
+    """
+    basename = posixpath.basename(artifact_path)
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_path = os.path.join(tmp_dir, basename)
+        artifact_repo = _get_artifact_repo_mlflow_artifacts()
+        artifact_repo._download_file(artifact_path, tmp_path)
+        return send_from_directory(tmp_dir, basename)
```
harupy, line 837: @dbczumar On windows, this code doesn't work:

https://github.com/mlflow/mlflow/runs/4102743134?check_suite_focus=true#step:5:1887

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpoug8eouu\\a.txt'
```

This SO question reports the exact same issue:

https://stackoverflow.com/questions/56581717/in-flask-how-can-i-send-a-temporary-file-and-delete-it-after-upload-is-finished?noredirect=1&lq=1

> It looks like `send_file` is asynchronous, so Python tries to delete the temporary directory while another thread is still uploading the file. On Linux, this doesn't happen because only the filenames are deleted, while the file node will still be alive until the other thread finishes uploading the file.

harupy: This SO answer provides a few workarounds: https://stackoverflow.com/a/24613980

mlflow/store/artifact/http_artifact_repo.py
```diff
@@ -0,0 +1,70 @@
...
+from mlflow.store.artifact.artifact_repo import ArtifactRepository, verify_artifact_path
+from mlflow.utils.file_utils import relative_path_to_artifact_path
+
+
+class HttpArtifactRepository(ArtifactRepository):
+    """Stores artifacts in a remote artifact storage using HTTP requests"""
+
+    def __init__(self, artifact_uri):
+        super().__init__(artifact_uri)
+        self._session = requests.Session()
```
harupy, line 15: Replaced `request.xxx` to with `self._session.xxx` to avoid re-establishing a new connection everytime we make a request.

mlflow/store/artifact/http_artifact_repo.py
```diff
@@ -0,0 +1,70 @@
...
+            file_infos.append(file_info)
+
+        return sorted(file_infos, key=lambda f: f.path)
+
+    def _download_file(self, remote_file_path, local_path):
+        url = posixpath.join(self.artifact_uri, remote_file_path)
+        with self._session.get(url, stream=True) as resp:
+            resp.raise_for_status()
+            with open(local_path, "wb") as f:
+                chunk_size = 1024 * 1024  # 1 MB
```
harupy, line 68: Increased `chunk_size` to 1 MB.

mlflow/utils/proto_json_utils.py
```diff
@@ -11,6 +11,8 @@
 
 def message_to_json(message):
     """Converts a message to JSON, using snake_case for field names."""
+    # Note protobuf encodes int64, fixed64, and uint64 to string:
+    # https://developers.google.com/protocol-buffers/docs/proto3#json
```
harupy, line 15: Added this note for those encountering the same issue in the future.

mlflow/utils/proto_json_utils.py
```diff
@@ -11,6 +11,8 @@
...
     return MessageToJson(message, preserving_proto_field_name=True)
```
harupy, line 16: Added this comment for those encountering the same issue in the future.

examples/mlflow_artifacts/README.md
```diff
@@ -0,0 +1,45 @@
...
+```sh
+# Clean up existing wheels
+rm dist/*
+
+# Build a wheel for the development version of mlflow
+pip wheel --no-deps --wheel-dir dist ../..
+
+# Build a wheel for the latest version of mlflow on PyPI
+pip wheel --no-deps --wheel-dir dist mlflow
+```
```
dbczumar, line 16: Why are these steps required?

mlflow/cli.py
```diff
@@ -317,6 +317,15 @@ def _validate_static_prefix(ctx, param, value):  # pylint: disable=unused-argume
     "Default: Within file store, if a file:/ URI is provided. If a sql backend is"
     " used, then this option is required.",
 )
+@click.option(
+    "--artifacts-destination",
+    metavar="URI",
+    default="./mlartifacts",
+    help=(
+        "The base artifact location from which to resolve artifact upload/download/list requests "
```
dbczumar, line 325: Love it! Can we also clarify that this only applies when the tracking server is configured to stream artifacts and the experiment artifact root is `http` or `mlflow-artifacts`?

mlflow/server/handlers.py
dbczumar: "Stream file, then delete" from https://stackoverflow.com/a/24613980/11952869 seems good!

mlflow/server/handlers.py
```diff
@@ -799,6 +823,64 @@ def _delete_model_version_tag():
...
+    A request handler for `PUT /mlflow-artifacts/artifacts/<artifact_path>` to upload an artifact
+    to `artifact_path` (a relative path from the root artifact directory).
+    """
+    head, tail = posixpath.split(artifact_path)
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_path = os.path.join(tmp_dir, tail)
+        with open(tmp_path, "wb") as f:
+            chunk_size = 1024 * 1024  # 1 MB
+            while True:
+                chunk = request.stream.read(chunk_size)
```
dbczumar, line 852: Is there any timeout on the stream read operation? Else, we may get stuck here forever.

mlflow/store/artifact/http_artifact_repo.py
dbczumar: neat!

mlflow/store/artifact/http_artifact_repo.py
```diff
@@ -0,0 +1,70 @@
...
+                f["is_dir"],
+                int(f["file_size"]) if ("file_size" in f) else None,
+            )
+            file_infos.append(file_info)
+
+        return sorted(file_infos, key=lambda f: f.path)
+
+    def _download_file(self, remote_file_path, local_path):
+        url = posixpath.join(self.artifact_uri, remote_file_path)
+        with self._session.get(url, stream=True) as resp:
```
dbczumar, line 65: Is there a timeout we should specify here? Else, we may get stuck here

mlflow/utils/proto_json_utils.py
dbczumar: Thank you!

mlflow/protos/mlflow_artifacts.proto
```diff
@@ -0,0 +1,80 @@
+syntax = "proto2";
```
dbczumar, line 1: Can we add a docstring to this proto file explaining which use cases it serves?

mlflow/cli.py
dbczumar: In general, we should figure out how to tell users about the `http` and `mlflow-artifacts` URIs for `--default-artifact-root` as well.

dbczumar (commented):
@harupy Looks great! Happy to approve once remaining comments are addressed and tests pass.

examples/mlflow_artifacts/README.md
```diff
@@ -0,0 +1,45 @@
+This directory contains a set of files for demonstrating the MLflow Artifacts service.
```
dbczumar, line 1: Can we explain what the artifacts service does and provide a small example command that users could run manually to serve artifacts via `mlflow server` in the README?

mlflow/protos/mlflow_artifacts.proto
harupy: Sure!

mlflow/server/handlers.py
harupy: Let's try that!

examples/mlflow_artifacts/README.md
harupy: This is mainly for development purposes. We can simplify the workflow by preparing two Docker files (`Dockerfile` and `Dockerfile.dev`). 

`docker-compose.yml` would look like:

```
# docker-compose.yml
  tracking-server:
    build:
      context: .
      # Defaults to `Dockerfile` if `DOCKERFILE` environment variable doesn't exist.
      dockerfile: "${DOCKERFILE:-Dockerfile}"
    depends_on:
      - postgres
      - artifacts-server
```

`README.md` would look like:

````md
This directory contains a set of files for demonstrating the MLflow Artifacts service.

## Run the example

```sh
# Build services
docker-compose build

# Launch tracking and artifacts servers
docker-compose up -d

# Run `run.py` that uploads, downloads, and list artifacts
docker-compose run -v ${PWD}/run.py:/app/run.py client python run.py
```

## Explore the logging results:

```sh
# Make sure both tracking and artifacts servers are running
docker-compose ps
```

- MLflow UI is available at http://localhost:5000 to explore tracking results.
- MinIO Console is available at http://localhost:9001 to explore logged artifacts. The login username and password are `user` and `password`.

## Reset tracking and artifacts servers

```
docker-compose down --volumes --remove-orphans
```

👇👇👇

## Development

```
pip wheel --no-deps --wheel-dir dist ../..
DOCKERFILE=Dockerfile.dev docker-compose build
docker-compose run -v ${PWD}/run.py:/app/run.py client python run.py
```

````

What do you think about this approach?

File: mlflow/store/artifact/http_artifact_repo.py (hunk 26 above)
Comment (harupy) [reply mlflow/store/artifact/http_artifact_repo.py:65]: Yes, what about 10 minutes?

Comment (harupy) [reply mlflow/store/artifact/http_artifact_repo.py:65]: Found a related discussion:

https://github.com/psf/requests/issues/1803#issuecomment-325934820

image

Comment (harupy) [reply mlflow/store/artifact/http_artifact_repo.py:65]: Can we set `timeout=10`?

File: mlflow/store/artifact/http_artifact_repo.py [hunk 29]
```diff
@@ -0,0 +1,70 @@
...
+        self._session.close()
+
+    def log_artifact(self, local_file, artifact_path=None):
+        verify_artifact_path(artifact_path)
+
+        file_name = os.path.basename(local_file)
+        paths = (artifact_path, file_name) if artifact_path else (file_name,)
+        url = posixpath.join(self.artifact_uri, *paths)
+        with open(local_file, "rb") as f:
+            resp = self._session.put(url, data=f)
```
harupy, line 27: For `put`, I think we need to set a larger timeout (say 10 minutes).

mlflow/server/handlers.py
harupy: The stream-then-delete approach worked on windows!

mlflow/server/handlers.py
harupy: - `request.stream` is an instance of `gunicorn.http.body.Body`.
- `gunicorn.http.body.Body.read` doesn't seem to have an option related to `timeout`: 
 
Source: https://github.com/benoitc/gunicorn/blob/ff58e0c6da83d5520916bc4cc109a529258d76e1/gunicorn/http/body.py#L202

```python
    def read(self, size=None):
        size = self.getsize(size)
        ...
```

examples/mlflow_artifacts/README.md
harupy: Dockerfile:

```
FROM python:3.6

WORKDIR /app

RUN pip install psycopg2 boto3
RUN pip install mlflow
```

Dockerfile.dev:

```
FROM python:3.6

WORKDIR /app

RUN pip install psycopg2 boto3
RUN pip install mlflow
COPY dist ./dist
RUN pip install dist/mlflow-*.whl
```

mlflow/cli.py
harupy: Can we address this in a follow-up that adds support for `mlflow-artifacts` scheme?

examples/mlflow_artifacts/README.md
```diff
@@ -0,0 +1,54 @@
+This directory contains a set of files for demonstrating the MLflow Artifacts Service.
+
```
harupy, line 2: I'll insert a new section here that briefly explains what the MLflow Artifacts Service does.

```suggestion
## What does the MLflow Artifacts Service do?
...
```

examples/mlflow_artifacts/README.md
harupy: Updated the README!

mlflow/utils/proto_json_utils.py
harupy: Once https://github.com/mlflow/mlflow/pull/5010 is merged, we no longer need this comment.

mlflow/cli.py
dbczumar: > In general, we should figure out how to tell users about the http and mlflow-artifacts URIs for --default-artifact-root as well.

That's a good idea. We should also make it easy for users to learn about these through the `mlflow server` CLI in a follow-up PR.

mlflow/server/handlers.py
dbczumar: Got it. This probably isn't too big a deal because we set gunicorn worker timeouts. Worst case, we may hang here until the worker times out, which is fine :).

dbczumar (approved):
LGTM again, thanks @harupy ! Really awesome!

harupy:
@dbczumar Thanks for reviewing this PR!

PR #4885, opened 2021-10-12
[WIP] [Draft] Autologging functionality for scikit-learn integration with XGBoost and LightGBM

jwyyy:
Signed-off-by: Junwen Yao, jwyiao@gmail.com.

## What changes are proposed in this pull request?

This PR will enable autologging for XGBoost and LightGBM sklearn estimators. Resolves https://github.com/mlflow/mlflow/issues/4296.

(Part 1) autologging for XGBoost sklearn estimators. 
(Part 2) working on autologging for LightGBM sklearn estimators.

## How is this patch tested?

A short example is provided. Tests will be added later.

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

This PR will enable autologging for XGBoost (LightGBM) sklearn estimators using `mlflow.xgboost.autolog()` (`mlflow.lightgbm.autolog()`).

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [x] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [x] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

<a name="release-note-category"></a>
[...]

PR #4954, opened 2021-10-28
Autologging functionality for scikit-learn integration with XGBoost (Part 1)

jwyyy:
Signed-off-by: Junwen Yao <jwyiao@gmail.com>

This is the first PR to add autologging for XGBoost sklearn models. It revises model saving / loading and adds `model_class` to specify XGBoost model class, such as Booster, or XGBRegressor. 

A separate PR will be made to add autologging for XGBoost sklearn models using `mlflow.sklearn` routine.

(Draft + discussion: https://github.com/mlflow/mlflow/pull/4885)

## How is this patch tested?

New test functions are added. See PR files.

This PR will enable saving / loading XGBoost models, including sklearn models, with model class specification. 
Functions `save_model()` / `load_model()` in `mlflow.xgboost` can be used as before.

Components 

Interface 

Language 

Integrations


dbczumar (commented):
@jwyyy Looks awesome! I left a few tiny comments - I think this is very close!

mlflow/xgboost.py
```diff
@@ -257,8 +257,27 @@ def log_model(
...
+        # specify the ``data`` field within its flavor configuration.
+        # We use ``model_class`` to specify its sklearn model class.
+        # For these models, the ``path`` parameter of ``load_pyfunc()``
+        # refers to the top-level MLflow Model directory.
+        # In this case, we first get the xgboost sklearn model from
+        # its flavor configuration and then create an instance based on its class.
+        flavor_conf = _get_flavor_configuration(model_path=path, flavor_name=FLAVOR_NAME)
+        model_class = flavor_conf.get("model_class", "Booster")
+        model_path = os.path.join(path, "model.xgb")
+        model = getattr(xgb, model_class)()
```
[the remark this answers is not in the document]
jwyyy: Hi @dbczumar, we can definitely save the fully-qualified class name. Just a short question: We can save `xgboost.XGBRegressor` or `xgboost.sklearn.XGBRegressor`. Which one is preferred? Since both are valid: 
```
import xgboost
regressor = xgboost.XGBRegressor(args)
# also valid
regressor = xgboost.sklearn.XGBRegressor(args)
```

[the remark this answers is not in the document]
jwyyy: On a second though, maybe `xgboost.sklearn.XGBRegressor` is better. It specifies the class is a sklearn model.

mlflow/xgboost.py
```diff
@@ -152,14 +152,16 @@ def save_model(
 
     # Save an XGBoost model
     xgb_model.save_model(model_data_path)
-
+    xgb_model_class = xgb_model.__class__.__name__
     pyfunc.add_to_model(
         mlflow_model,
         loader_module="mlflow.xgboost",
-        data=model_data_subpath,
```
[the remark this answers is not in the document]
dbczumar: @harupy @jwyyy I think this is a good idea for MLflow 2.0 (whenever we decide to release a 2.0 version). This would be a breaking change for all MLflow Model flavors / loader modules defined outside of the main MLflow code base, as these loader modules would need to be updated to expect a directory rather than a file.

jwyyy:
Hi @dbczumar @harupy, thank you for your feedback! I updated this PR and doc according to your suggestions. Please let me know if they look good. Thanks!

mlflow/xgboost.py
```diff
@@ -286,12 +309,11 @@ def load_model(model_uri):
...
 
-    :return: An XGBoost model (an instance of `xgboost.Booster`_)
+    :return: An XGBoost model. An instance of either `xgboost.Booster`_ or XGBoost scikit-learn
+             models, depending on the saved model class specification.
     """
     local_model_path = _download_artifact_from_uri(artifact_uri=model_uri)
-    flavor_conf = _get_flavor_configuration(model_path=local_model_path, flavor_name=FLAVOR_NAME)
-    xgb_model_file_path = os.path.join(local_model_path, flavor_conf.get("data", "model.xgb"))
-    return _load_model(path=xgb_model_file_path)
+    return _load_model(path=local_model_path)
```
jwyyy, line 316: @dbczumar @harupy Is it a good idea to apply `_XGBModelWrapper` here to unify the input format no matter what XGBoost model is loaded?

harupy: I don't think we should do that because `mlflow.xgboost.log_model` should return what the user logs (e.g. if the user logs `xgboost.XGBClassifier`, `xgboost.XGBClassifier` should be returned).

jwyyy: Thanks for your explanation! I think it is good to leave as it is.

mlflow/xgboost.py
```diff
@@ -255,10 +255,32 @@ def log_model(
...
+        model_class = "Booster"
+        xgb_model_path = os.path.join(path, flavor_conf["data"])
+    else:
+        # In contrast, XGBoost models saved in new MLflow (>=x.x.x) do not
+        # specify the ``data`` field within its flavor configuration.
+        # We use ``model_class`` to specify its XGBoost model class.
+        # In this case, we first get the XGBoost model from
+        # its flavor configuration and then create an instance based on its class.
+        model_class = flavor_conf.get("model_class", "xgboost.core.Booster").split(".")[-1]
+        xgb_model_path = os.path.join(path, "model.xgb")
```
harupy, line 280:
```suggestion
    model_dir = os.path.dirname(path)
    flavor_conf = _get_flavor_configuration(model_path=model_dir, flavor_name=FLAVOR_NAME)
    model_class = flavor_conf.get("model_class", "xgboost.core.Booster").split(".")[-1]
        xgb_model_path = os.path.join(path, "model.xgb")
```

Can we use `os.path.dirname` to get the directory that stores model artifacts?

harupy: Can we load a model class using `importlib`? 

```python
>>> model_class = "xgboost.core.Booster"
>>> module, cls = model_class.rsplit(".", maxsplit=1)
>>> getattr(importlib.import_module(module), cls)
<class 'xgboost.core.Booster'>
```

mlflow/xgboost.py
```diff
@@ -152,14 +152,14 @@ def save_model(
 
     # Save an XGBoost model
     xgb_model.save_model(model_data_path)
-
+    xgb_model_class = ".".join([xgb_model.__module__, xgb_model.__class__.__name__])
```
harupy, line 155: Can we use `_get_fully_qualified_class_name` here?

https://github.com/mlflow/mlflow/blob/4195541c7f6a79fba4ccad2295534f16eaa8db39/mlflow/utils/__init__.py#L126-L130

mlflow/xgboost.py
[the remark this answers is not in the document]
harupy: I agree :)

tests/xgboost/test_xgboost_model_export.py
```diff
@@ -49,6 +49,16 @@ def xgb_model():
     return ModelWithData(model=model, inference_dataframe=X, inference_dmatrix=dtrain)
 
 
+@pytest.fixture(scope="session")
+def xgb_sklearn_model():
+    boston = datasets.load_boston()
```
harupy, line 54: Can we use a different dataset because `load_boston` is deprecated?

https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html:

> DEPRECATED: load_boston is deprecated in 1.0 and will be removed in 1.2.

mlflow/xgboost.py
jwyyy: Thanks for your suggestion! It is what I was looking for :laughing:

mlflow/xgboost.py
jwyyy: Thanks for the suggestion! I will update them in the next commit shortly.

jwyyy: > Can we use `os.path.dirname` to get the directory that stores model artifacts, and pass the directory to `_get_flavor_configuration`?

In the case where `path` is a folder not a model file, it could lead to its parent directory, causing loading error. But I got your idea. Trying to figure out a solution. Thanks for your suggestion!

tests/xgboost/test_xgboost_model_export.py
jwyyy: Thanks for your suggestion! I didn't notice that boston will be removed. I use the wine dataset in the latest commit.

tests/xgboost/test_xgboost_model_export.py
```diff
@@ -452,3 +480,49 @@ def test_pyfunc_serve_and_score_sklearn(model):
...
+    assert "model_class" in pyfunc_conf
+    assert pyfunc.DATA not in pyfunc_conf
+
+    pyfunc.add_to_model(model_conf, loader_module="mlflow.xgboost", data="model.xgb")
+    model_conf.add_flavor(mlflow.xgboost.FLAVOR_NAME, data="model.xgb")
+    model_conf.save(model_conf_path)
+    model_conf = Model.load(model_conf_path)
+    xgboost_conf = model_conf.flavors.get(mlflow.xgboost.FLAVOR_NAME)
+    assert "data" in xgboost_conf
+    assert xgboost_conf["data"] == "model.xgb"
```
jwyyy, line 513: I have to add `data` to MLmodel and save it again to test if `_load_model` can successfully load previously logged XGBoost Booster models. I am not sure if [L465](https://github.com/mlflow/mlflow/commit/aa563fbdb8dc35cc99cb4e73e7e057a16b76acff#diff-5e72f29b7d4bc3e2962472dea49037d8585cd85b1cc3f0dc3b23bce0c27b0459R465) really tests backward compatibility, since `conf` is reloaded inside `pyfunc.load_model()`, and updating `pyfunc_conf` outside doesn't change `conf`.

jwyyy:
Hi @harupy @dbczumar, thank you for initializing the auto check! According to the error messages, two unsuccessful checks were due to network connection exception, not related to the changes in this PR. I am not sure if I can fix these two problems efficiently. Can you take a look and re-try the check? Thanks! Also please let me know if the latest commit look good.

mlflow/xgboost.py
```diff
@@ -255,19 +256,48 @@ def log_model(
...
+        model_class = "xgboost.core.Booster"
+        xgb_model_path = os.path.join(model_dir, flavor_conf["data"])
+    else:
+        # In contrast, XGBoost models saved in new MLflow (>=x.x.x) do not
+        # specify the ``data`` field within its flavor configuration.
+        # We use ``model_class`` to specify its XGBoost model class.
+        # In this case, we first get the XGBoost model from
+        # its flavor configuration and then create an instance based on its class.
+        model_class = flavor_conf.get("model_class", "xgboost.core.Booster")
+        xgb_model_path = os.path.join(model_dir, "model.xgb")
```
harupy, line 284: Is it possible to keep the `data` field and remove if-else here? 

```python
model_class = flavor_conf.get("model_class", "xgboost.core.Booster")
xgb_model_path = os.path.join(model_dir, flavor_conf["data"])
module, cls = model_class.rsplit(".", maxsplit=1)
model = getattr(importlib.import_module(module), cls)()
model.load_model(xgb_model_path)
```

jwyyy: > ```
> xgb_model_path = os.path.join(model_dir, flavor_conf["data"])
> ...
> model.load_model(xgb_model_path)
> ```
Thanks for your suggestion! I think we can do:
> xgb_model_path = os.path.join(model_dir, flavor_conf.get("data", "model.xgb"))

When `model_class` is saved, `data` is not a key of `flavor_conf` anymore. So `flavor_conf["data"]` returns None. In that case, we can manually append the saved model file to `xgb_model_path`.

jwyyy: I kept the comments in the latest commit. They are meant to explain the difference in old / new model saving / loading. But if they look redundant, we can remove them.

harupy: > When `model_class` is saved, data is not a key of flavor_conf anymore.

Can we save both `model_class` and `data`?

jwyyy: > > When `model_class` is saved, data is not a key of flavor_conf anymore.
> 
> Can we save both `model_class` and `data`?

I think this is also fine. But in `pyfunc.load_model()`, `DATA` is already added a model path if it exists in MLmodel conf ([L666](https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/__init__.py#L666)). It adds additional work to [retrieving model directory](https://github.com/mlflow/mlflow/pull/4954/files#diff-f6bd607049d4969297151a812f447bdc13453a856df017983f7ea81470d25bd0R268) every time `_load_pyfunc()` is called.

jwyyy: Or we can save `data` in xgboost flavor but not in pyfunc flavor to avoid adding `DATA` path in `pyfunc.load_model()`. Please let me know which is preferred. Thanks!

jwyyy: It came to me that we don't need to save `model_class` in pyfunc flavor as well, since it is not necessary when calling `pyfunc.load_model`, and loading is handled by local `_load_pyfunc`.

jwyyy:
Hi @harupy @dbczumar, I notice that this PR #4997 tries to add a new argument to `load_model()` in every flavor. This will have an impact on how we handle model loading in `mlflow.xgboost` (although the revision should be very easy). I am wondering which PR will be addressed first. Or maybe we can directly add the xgboost part in this PR. Please let you know your plan / ideas. Thanks! 

(The previous failed check (35/36) was due to Keras 2.7 compatibility, which should be fixed now.)

harupy:
> I notice that this PR #4997 tries to add a new argument to load_model() in every flavor.
Yes it does, but it doesn't change how to load the model, it just allows users to decide where to load the model.

mlflow/xgboost.py
```diff
@@ -255,19 +256,43 @@ def log_model(
...
+    model_dir = os.path.dirname(path) if os.path.isfile(path) else path
+    flavor_conf = _get_flavor_configuration(model_path=model_dir, flavor_name=FLAVOR_NAME)
+
+    # XGBoost Booster models saved in MLflow (<x.x.x) specify
+    # the ``data`` field within its pyfunc and XGBoost flavor configurations.
+    # In contrast, XGBoost models saved in new MLflow (>=x.x.x) do not
+    # specify the ``data`` field within its pyfunc flavor configuration.
+    # We also add ``model_class`` in XGBoost flavor configuration to specify
+    # its XGBoost model class. When loading models, we first get the XGBoost
+    # model from its flavor configuration and then create an instance based on its class.
```
harupy, line 277: Can we fix this comment because we'll keep the data field?

jwyyy: The `data`field in `pyfunc` flavor was removed [(L157)](https://github.com/mlflow/mlflow/pull/4954/files#diff-f6bd607049d4969297151a812f447bdc13453a856df017983f7ea81470d25bd0R157) to avoid extra work in `_load_pyfunc()` ([discussion](https://github.com/mlflow/mlflow/pull/4954#discussion_r742956137)). We can remove the comment if the change is clear. Or add back the `data` field to the `pyfunc` flavor. Please let me know which is preferred. Thanks!

harupy: Sorry I misread the code diff. `pyfunc` flavor no longer has the `data` field. I'd prefer to keep the `data` field since removing it might lead to a breaking change.

@dbczumar Any thoughts?

dbczumar: @harupy What's the backwards-incompatible change here? Is the issue that older versions of MLflow won't be able to load XGBoost Booster models saved after this change is applied? If so, I don't think we typically guarantee that new MLflow models are loadable by older MLflow versions.

If we want to maintain this guarantee by preserving the `data` field, how do we propose to find the model class within `_load_pyfunc`?

jwyyy: > If we want to maintain this guarantee by preserving the `data` field, how do we propose to find the model class within `_load_pyfunc`?

We can still get the directory path by retrieving the parent directory of `model.xgb` first as done in [L268](https://github.com/mlflow/mlflow/pull/4954/files#diff-f6bd607049d4969297151a812f447bdc13453a856df017983f7ea81470d25bd0R268).

dbczumar: @jwyyy Sounds good! I think it's okay to assume that `model.xgb` always lives within the same directory as `MLmodel`, since this flavor is in charge of determining the directory structure for the model.

jwyyy: @dbczumar Thank you for your suggestion! @harupy I think it is settled now. I will add back `data` in the `pyfunc` flavor after the current check finishes. (It failed Keras 2.7.0 autologging / models, and the cause seems to be the compatibility issue that happened in other PRs a few days ago. I guess it may also fail TF 2.7.0 tests as well.)

mlflow/xgboost.py
```diff
@@ -255,19 +261,40 @@ def log_model(
...
 def _load_pyfunc(path):
     """
     Load PyFunc implementation. Called by ``pyfunc.load_pyfunc``.
 
-    :param path: Local filesystem path to the MLflow Model with the ``xgboost`` flavor.
+    :param path: Local filesystem path to
+                    the MLflow Model with the ``xgboost`` flavor (MLflow < x.x.x) or
+                    the top-level MLflow Model directory (MLflow >= x.x.x).
     """
+
```
harupy, line 297: nit: can we revert this since it's unrelated?

harupy (approved):
Looks good to me once https://github.com/mlflow/mlflow/pull/4954#discussion_r745308893 is addressed!

mlflow/xgboost.py
harupy: Sounds good, thanks for the updates!

harupy: We can ignore keras 2.7.0 failures since we solved them in #5024.

mlflow/xgboost.py
jwyyy: Sure. I will update the commit once the current check finishes!

mlflow/xgboost.py
```diff
@@ -255,19 +261,40 @@ def log_model(
...
+    """
+    import importlib
+
+    model_dir = os.path.dirname(path) if os.path.isfile(path) else path
+    flavor_conf = _get_flavor_configuration(model_path=model_dir, flavor_name=FLAVOR_NAME)
+
+    # XGBoost models saved in new MLflow (>=x.x.x) have ``model_class``
+    # in the XGBoost flavor configuration to specify its XGBoost model class.
+    # When loading models, we first get the XGBoost model from
+    # its flavor configuration and then create an instance based on its class.
```
harupy, line 279:
```suggestion
    # XGBoost models saved in MLflow >=1.22.0 have `model_class`
    # in the XGBoost flavor configuration to specify its XGBoost model class.
    # When loading models, we first get the XGBoost model from
    # its flavor configuration and then create an instance based on its class.
```

jwyyy:
> Looks good to me once [#4954 (comment)](https://github.com/mlflow/mlflow/pull/4954#discussion_r745308893) is addressed!

@harupy I just updated the comments in this PR, including [#4954 (review)](https://github.com/mlflow/mlflow/pull/4954#discussion_r745311926).

jwyyy:
@harupy @dbczumar Thank you very much for your feedback and suggestions! It seems we are at the final stage of this PR. Once it is merged, I will make the second PR which completes the autologging for XGBoost sklearn models based on our previous discussion. It should be ready within the next few days. Meanwhile I will also submit the first PR for LightGBM models (similar to this PR). Thanks again!

dbczumar (approved):
LGTM! Thanks so much @jwyyy !

jwyyy:
Two failed tests seem unrelated to this PR. The error messages are the same as unsuccessful checks in recent commits to master #5032 #5031 #4999 #5029.

harupy:
> Two failed tests seem unrelated to this PR.

Yep, we're already investigating them. Let's merge this PR!

Issue #4816, opened 2021-09-16
[FR] Migrate MLflow SageMaker deployment functionality to deployments interface


dbczumar:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [ ] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [X] No. I cannot contribute this feature at this time.

## Proposal Summary

MLflow provides a deployments API / CLI for CRUDing real-time scoring deployments from MLflow models. This is a pluggable / extendable interface, and plugins currently exist for deployment to Azure ML, TorchServe, Redis, and more. MLflow's tool for deploying models to SageMaker isn't currently exposed through the deployments API / CLI; rather, it has a separate `mlflow sagemaker` CLI / `mlflow.sagemaker` API for legacy reasons. For consistency, we propose to move SageMaker deployment functionality into the `mlflow deployments` CLI / API.

## Motivation
- What is the use case for this feature? This feature unifies MLflow's real-time scoring deployment integrations behind a single API.
- Why is this use case valuable to support for MLflow users in general? It's difficult for users to reason about separate APIs for different deployment environments. A unified API will reduce friction.
- Why is this use case valuable to support for your project(s) or organization? Databricks customers deploy models to a variety of locations. It's difficult for them to reason about separate APIs for SageMaker deployment.
- Why is it currently difficult to achieve this use case? There is no SageMaker deployment functionality conforming to the MLflow plugin specification.

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
[...]

PR #4923, opened 2021-10-21
Allow Deploying SageMaker Endpoints to Another AWS Account #4921

andresionek91:
## What changes are proposed in this pull request?

Currently, MLFlow will only allows deploying SageMaker endpoints to the currently active AWS account. This feature request aims to allow deployment to a different account by assuming a cross-account role.

## How is this patch tested?

Created unit tests to verify if:
* Assume role operation is executed successfully
* Deployment with assume role works
* Deploy with non existent assume role fails

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

Allow assuming a role in another AWS Account to deploy SageMaker Endpoints.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [x] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

<a name="release-note-category"></a>
### How should the PR be classified in the release notes? Choose one:

[...]

PR #4971, opened 2021-10-30
Move the code that deploys a SageMaker Endpoint to a plugin that implements BaseDeploymentClient

jamestran201:
Signed-off-by: James Tran <tran.james2001@gmail.com>

This PR adds a deployment plugin for SageMaker called `SageMakerDeploymentClient`. Deployments to SageMaker can now be managed through the `mlflow deployments` CLI. The target URI for this command can be:
- ``sagemaker``: This will set the default region to `us-west-2` and the default assumed role ARN to `None`.
- ``sagemaker:/region_name``: This will set the default region to `region_name` and the default assumed role ARN to `None`.
- ``sagemaker:/region_name/assumed_role_arn``: This will set the default region to `region_name` and the default assumed role ARN to `assumed_role_arn`.

For example, to create a SageMaker deployment: `mlflow deployments create --target sagemaker:/us-east-1/arn:aws:iam::123456789012:role/assumed_role --name dummy --model-uri ...`

Related issue: https://github.com/mlflow/mlflow/issues/4816

## How is this patch tested?

This patch is tested by the tests in `tests/sagemaker/test_sagemaker_deployment_client.py`.

Add a deployment plugin for SageMaker. The only implemented function at the moment is `create_deployment()`.

Components 

Interface 

Language 

Integrations
- [x] `integrations/sagemaker`: SageMaker integrations


docs/source/conf.py
```diff
@@ -331,6 +331,7 @@
     ("py:class", "matplotlib.figure.Figure"),
     ("py:class", "plotly.graph_objects.Figure"),
     ("py:class", "PIL.Image.Image"),
+    ("py:class", "mlflow.deployments.base.BaseDeploymentClient"),
```
tmnhat2001, line 334: This is to avoid this error when building the documentation:
```
/home/teppi/mlflow/mlflow/sagemaker/__init__.py:docstring of mlflow.sagemaker.SageMakerPlugin:1:py:class reference target not found: mlflow.deployments.base.BaseDeploymentClient
```

mlflow/sagemaker/__init__.py
```diff
@@ -1656,6 +1657,182 @@ def _does_model_exist(model_name, sage_client):
...
+        config = self._default_config()
+
+        for key in custom_config:
+            if key in config:
+                config[key] = custom_config[key]
+
+        return config
+
+    @experimental
+    def create_deployment(self, name, model_uri, flavor=None, config=None):
```
tmnhat2001, line 1689: Marking this as experimental because not all methods in this class have been implemented yet.

tests/sagemaker/test_sagemaker_plugin.py
```diff
@@ -0,0 +1,615 @@
+import os
```
tmnhat2001, line 1: Most of the content in this file are copied from https://github.com/mlflow/mlflow/blob/master/tests/sagemaker/test_deployment.py.

The only changes are:
- Call `sagemaker_plugin.create_deployment()` instead of `mfs.deploy()`
- Create a fixture for the SageMakerPlugin

tests/sagemaker/test_sagemaker_plugin.py
```diff
@@ -0,0 +1,615 @@
...
+    deployment_end_time = time.time()
+
+    assert (deployment_end_time - deployment_start_time) < endpoint_creation_latency
+    endpoint_description = sagemaker_client.describe_endpoint(EndpointName=name)
+    assert endpoint_description["EndpointStatus"] == Endpoint.STATUS_CREATING
+
+
+# @pytest.mark.large
+# @mock_sagemaker_aws_services
+# def test_deploy_replace_in_asynchronous_mode_returns_before_endpoint_creation_completes(
```
tmnhat2001, line 296: Some tests are commented out in this file because they are testing the ability to update a deployment. This functionality will be handled by `update_deployment()` in a follow-up PR.

dbczumar (commented):
@tmnhat2001 Apologies for the delay. This looks like a great start for the implementation! A discussed on the issue, can we explore embedding the region name in the target URI?

tmnhat2001:
@dbczumar No worries! I was leaving it for a later PR, but I've added them here now. Please review again when you have the chance.

dbczumar:
@tmnhat2001 Thank you for your updates - I'll take a look first thing tomorrow morning.

mlflow/sagemaker/__init__.py
```diff
@@ -1656,6 +1659,204 @@ def _does_model_exist(model_name, sage_client):
         return True if response else False
 
 
+class SageMakerPlugin(BaseDeploymentClient):
+    """
+    Initialize a deployment client for SageMaker
+
+    :param target_uri: A URI that has the value of ``sagemaker`` or ``sagemaker:/region_name``.
+                       The provided region name will be used as the default region for subsequent function calls.
+                       When no region name is provided, the default region will be set to `us-west-2`.
```
dbczumar, line 1668: Nice! Apologies for continuing to make things harder, but we recently merged https://github.com/mlflow/mlflow/pull/4923 to enable users to specify an IAM role to use when interfacing with SageMaker deployments, which makes it easier to manage deployments across accounts. It would be great to incorporate IAM roles into the target URI to support this use case as well.

Perhaps `sagemaker:/<region_name>/<iam_role_arn>`? Users may have to URL encode the role ARN if it contains `?` or `#`, but I think this is probably uncommon.

If an IAM role ARN is specified, the user must also specify the region. Let me know how this sounds.

dbczumar (commented):
@tmnhat2001 This is looking great! Left a few comments. Once those are addressed, I think we can move forward with implementing the other deployment client methods. Thanks so much for your contribution!

mlflow/sagemaker/__init__.py
```diff
@@ -1656,6 +1659,204 @@ def _does_model_exist(model_name, sage_client):
...
+            mode=DEPLOYMENT_MODE_CREATE,
+            archive=config["archive"],
+            instance_type=config["instance_type"],
+            instance_count=config["instance_count"],
+            vpc_config=config["vpc_config"],
+            synchronous=config["synchronous"],
+            timeout_seconds=config["timeout_seconds"],
+        )
+
+    def update_deployment(self, name, model_uri=None, flavor=None, config=None):
```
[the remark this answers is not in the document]
tmnhat2001: Yes, that's right.

docs/source/conf.py
tmnhat2001: I believe it comes from this line: https://github.com/mlflow/mlflow/blob/bacc5d5aa903a9c9484ca75c218818159f8e641c/mlflow/sagemaker/__init__.py#L1748

The error goes away if BaseDeploymentClient is not declared as the parent class.

mlflow/sagemaker/__init__.py
tmnhat2001: That sounds great!

mlflow/sagemaker/__init__.py
```diff
@@ -1656,6 +1659,204 @@ def _does_model_exist(model_name, sage_client):
         return True if response else False
 
 
+class SageMakerPlugin(BaseDeploymentClient):
```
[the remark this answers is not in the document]
tmnhat2001: I had to register the module that contains the SageMaker client (see https://github.com/mlflow/mlflow/pull/4971/commits/059d64f3f77f17f802da3692a6d16a7e5a6c50c4) because the plugin manager requires some functions are defined at the module level when initializing a client object:
https://github.com/mlflow/mlflow/blob/9faf3b04c9f6b3766a64d7433dd3e85bcdf27fff/mlflow/deployments/plugin_manager.py#L93-L94

tmnhat2001:
@dbczumar Thanks for the review! I have addressed the comments. Please review it again when you have time.

docs/source/conf.py
dbczumar: Got it! Thanks for clarifying :)

mlflow/sagemaker/__init__.py
```diff
@@ -1064,6 +1066,33 @@ def _sigterm_handler(*_):
     proc.wait()
 
 
+def target_help(_):
```
dbczumar, line 1069: Can we remove the `_` arg from `target_help()` to resolve the following issue?:

```
$ mlflow deployments help -t sagemaker
/Users/corey.zumar/mlflow/mlflow/server/handlers.py:119: UserWarning: Failure attempting to register store for scheme "file-plugin": No module named 'mlflow_test_plugin.sqlalchemy_store'
  self.register_entrypoints()
<module 'mlflow.sagemaker' from '/Users/corey.zumar/mlflow/mlflow/sagemaker/__init__.py'>
  File "/Users/corey.zumar/opt/anaconda3/bin/mlflow", line 33, in <module>
    sys.exit(load_entry_point('mlflow', 'console_scripts', 'mlflow')())
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/corey.zumar/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/corey.zumar/mlflow/mlflow/deployments/cli.py", line 216, in target_help
    click.echo(interface._target_help(target))
  File "/Users/corey.zumar/mlflow/mlflow/deployments/interface.py", line 83, in _target_help
    return plugin_store[target].target_help()
TypeError: target_help() missing 1 required positional argument: '_'
```

mlflow/sagemaker/__init__.py
```diff
@@ -1064,6 +1066,33 @@ def _sigterm_handler(*_):
...
+    - sagemaker:/region_name
+    - sagemaker:/region_name/assume_role_arn
+
+    When the region_name or assume_role_arn are provided, they will be used as the default region
+    and assumed role ARN when executing the commands.
+
+    The `create` and `update` commands require a deployment name and a model_uri. The model flavor
+    and deployment configuration can be optionally provided.
+
+    The `delete` command accepts configurations to archive a model instead of deleting, execute
```
dbczumar, line 1090: Awesome! To support this, we'll need to extend the interface to accomplish this - e.g. by adding `config` as an optional kwarg to `delete_deployment`.

mlflow/sagemaker/__init__.py
```diff
@@ -1064,6 +1066,33 @@ def _sigterm_handler(*_):
...
+    The target URI must follow the following formats:
+    - sagemaker
+    - sagemaker:/region_name
+    - sagemaker:/region_name/assume_role_arn
+
+    When the region_name or assume_role_arn are provided, they will be used as the default region
+    and assumed role ARN when executing the commands.
+
+    The `create` and `update` commands require a deployment name and a model_uri. The model flavor
+    and deployment configuration can be optionally provided.
```
dbczumar, line 1088: Nice docstring! Can we also clarify that sync & async execution are supported, since we highlight that for `delete` below as well?

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1776,238 @@ def _does_model_exist(model_name, sage_client):
...
+            return
+
+        separator_index = values_str.find("/")
+        if separator_index == -1:
+            # values_str would look like us-east-1
+            self.region_name = values_str
+        else:
+            # values_str would look like us-east-1/arn:aws:1234:role/assumed_role
+            self.region_name = values_str[:separator_index]
+            self.assumed_role_arn = values_str[separator_index + 1 :]
```
dbczumar, line 1823: Can we handle the case where there may be multiple interior slashes in a path? e.g. `sagemaker:/us-west-2////arn:iam:my_role`?

tests/sagemaker/test_sagemaker_deployment_client.py
```diff
@@ -0,0 +1,722 @@
...
+            name="test-app",
+            model_uri=pretrained_model.model_uri,
+        )
+
+    assert "deployment operation failed" in exc.value.message
+    assert exc.value.error_code == ErrorCode.Name(INTERNAL_ERROR)
+
+
+# @pytest.mark.large
+# @mock_sagemaker_aws_services
```
dbczumar, line 472: Let me know if I can help debug the commented tests or if they're a work in progress or just used for debugging during your implementation.

dbczumar (commented):
@tmnhat2001 This looks fantastic! Sorry for the delay here again - the holiday season has been quite busy. Barring a few comments, I think we're all set to implement the remaining APIs.

tmnhat2001:
@dbczumar Sorry for the late reply. I need to wrap up a couple of things so I'll get back to this PR later this week.

tmnhat2001:
@dbczumar Sorry for the late response 🙇‍♂️ . I've addressed the comments, please review again when you have time.

Should I implement the remaining API's in this PR as well?

dbczumar:
> @dbczumar Sorry for the late response 🙇‍♂️ . I've addressed the comments, please review again when you have time.
> 
> Should I implement the remaining API's in this PR as well?

Hi @tmnhat2001 , no problem at all! If possible, it would be great to implement the remaining APIs in this PR :). Thank you for your contribution!

tmnhat2001:
@dbczumar Could you please review again? I have implemented the `update_deployment()` function.

jamestran201:
The commits between this comment and the previous one are for implementing the `delete_deployment()` function.

jamestran201:
The commits between this comment and the previous one are for implementing the `get_deployment()` function.

jamestran201:
@dbczumar The `list_deployments()` function has been implemented. Should we implement the `predict()` function as well?

dbczumar:
> @dbczumar The `list_deployments()` function has been implemented. Should we implement the `predict()` function as well?

@jamestran201 Fantastic! Thanks so much! I'll take a look! Perhaps we can implement `predict()` as a separate PR since we don't currently have an equivalent function and it might require some more iterations. Would you be able to work on this in a separate PR?

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1779,480 @@ def _does_model_exist(model_name, sage_client):
...
+            assume_role_arn=final_config["assume_role_arn"],
+            bucket=final_config["bucket"],
+            image_url=final_config["image_url"],
+            region_name=final_config["region_name"],
+            mode=DEPLOYMENT_MODE_CREATE,
+            archive=final_config["archive"],
+            instance_type=final_config["instance_type"],
+            instance_count=final_config["instance_count"],
+            vpc_config=final_config["vpc_config"],
+            synchronous=final_config["synchronous"],
```
dbczumar, line 1997: From the command line, `synchronous` is being parsed as a String and always evaluates to `True` because the string value `"False"` is truthy. Can we fix this here and in the other methods that use synchronous (e.g. perhaps by adding special case logic to `apply_custom_config`)?

We could define a variable above as:

```
synchronous = final_config["synchronous"]
if isinstance(synchronous, str):
    synchronous = (synchronous == True)
```

and pass it in here (though we may want to do this in `apply_custom_config` to avoid code repetition):

```suggestion
            synchronous=synchronous,
```

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1779,480 @@ def _does_model_exist(model_name, sage_client):
...
+        app_name, flavor = deploy(
+            app_name=name,
+            model_uri=model_uri,
+            flavor=flavor,
+            execution_role_arn=final_config["execution_role_arn"],
+            assume_role_arn=final_config["assume_role_arn"],
+            bucket=final_config["bucket"],
+            image_url=final_config["image_url"],
+            region_name=final_config["region_name"],
+            mode=DEPLOYMENT_MODE_CREATE,
```
dbczumar, line 1992: On second thought, can we check if `mode` is specified in the `config` and, if it is, pass it in? If `mode` is not specified, we can fall back to `DEPLOYMENT_MODE_CREATE`.

There are a couple reasons I think we should do this:

1. Right now, if I try to use this method while specifying a deployment name that already exists, I get the following error message:

```
mlflow.exceptions.MlflowException: You are attempting to deploy an application with name: mydep2 in 'create' mode. However, an application with the same name already exists. If you want to update this application, deploy in 'add' or 'replace' mode.
```

This exception comes from `mlflow.sagemaker.deploy()` and would be inaccurate if the `replace` and `add` `mode`s couldn't be specified.

2. For users who are automating deployments and don't want to add logic to check whether or not a deployment exists before creating a new one, supporting `mode=replace` would provide users with a convenient single-method "create or update" capability.

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1779,480 @@ def _does_model_exist(model_name, sage_client):
...
+        except Exception as exc:
+            raise MlflowException(
+                message=(f"There was an error while retrieving the deployment: {exc}\n")
+            )
+
+    def predict(self, deployment_name, df):
+        pass
+
+    def explain(self, deployment_name, df):  # pylint: disable=unused-argument
+        pass
```
dbczumar, line 2253: Can we raise `NotImplementedError` for now?

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1779,480 @@ def _does_model_exist(model_name, sage_client):
...
+            bucket=final_config["bucket"],
+            image_url=final_config["image_url"],
+            region_name=final_config["region_name"],
+            mode=DEPLOYMENT_MODE_CREATE,
+            archive=final_config["archive"],
+            instance_type=final_config["instance_type"],
+            instance_count=final_config["instance_count"],
+            vpc_config=final_config["vpc_config"],
+            synchronous=final_config["synchronous"],
+            timeout_seconds=final_config["timeout_seconds"],
```
dbczumar, line 1998: attempting to use this via the command line yields:

```
TypeError: '<' not supported between instances of 'float' and 'str'
```

Can we cast to `int()`?

mlflow/sagemaker/__init__.py
dbczumar: The same is true for `archive`. Can we update it as well?

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1779,480 @@ def _does_model_exist(model_name, sage_client):
...
+            flavor=flavor,
+            execution_role_arn=final_config["execution_role_arn"],
+            assume_role_arn=final_config["assume_role_arn"],
+            bucket=final_config["bucket"],
+            image_url=final_config["image_url"],
+            region_name=final_config["region_name"],
+            mode=DEPLOYMENT_MODE_CREATE,
+            archive=final_config["archive"],
+            instance_type=final_config["instance_type"],
+            instance_count=final_config["instance_count"],
```
dbczumar, line 1995: When instance count is passed via the command line, it should still be interpreted as an int, e.g.

```suggestion
            instance_count=int(final_config["instance_count"]),
```

Perhaps it would be better to make this change in `apply_custom_config`.

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1779,480 @@ def _does_model_exist(model_name, sage_client):
...
+            vpc_config=final_config["vpc_config"],
```
dbczumar, line 1996: If `vpc_config` is passed via the commandline, its type will be `str`. In that case, can we parse it to a dictionary via `json.parse()`?

dbczumar (commented):
@jamestran201 Thank you so much for your hard work on this. The plugin works extremely well; I've tested out all of the CRUD functionality, and it's very smooth in general. I've left a few small comments, mostly related to docs and data type casting when using the tool through the `mlflow deployments` CLI. After these addressed, I'd imagine we'll be ready or very close to ready to merge. Let me know if you have any questions!

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1769,574 @@ def _does_model_exist(model_name, sage_client):
...
+        for key, value in custom_config.items():
+            if key not in config:
+                continue
+
+            if key in int_fields and not isinstance(value, int):
+                value = int(value)
+            elif key in bool_fields and not isinstance(value, bool):
+                value = value == "True"
+            elif key in dict_fields and not isinstance(value, dict):
+                value = json.parse(value)
```
jamestran201, line 1872: There is some duplication in the conditionals. I attempted to DRY it up but the result seemed harder to follow. Please let me know if the duplication needs to be addressed.

dbczumar: Love it! Looks great!

mlflow/sagemaker/__init__.py
```diff
@@ -1116,12 +1133,17 @@ def _get_assumed_role_arn(**assume_role_credentials):
 def _assume_role_and_get_credentials(assume_role_arn):
     """
     Assume a new role in AWS and return the credentials for that role.
+    When ``assume_role_arn`` is ``None`` or an empty string,
+    this function does nothing and returns an empty dictionary.
 
     :param assume_role_arn: The ARN of the role that will be assumed
```
dbczumar, line 1139:
```suggestion
def _assume_role_and_get_credentials(assume_role_arn=None):
    """
    Assume a new role in AWS and return the credentials for that role.
    When ``assume_role_arn`` is ``None`` or an empty string,
    this function does nothing and returns an empty dictionary.

    :param assume_role_arn: Optional ARN of the role that will be assumed
```

mlflow/sagemaker/__init__.py
```diff
@@ -1064,6 +1053,34 @@ def _sigterm_handler(*_):
...
 
+def target_help():
+    """
+    Provide help information for the SageMaker deployment client.
+
+    :return:
+    :rtype: str
+    """
+    help_str = """For detailed documentation on the SageMaker deployment client, please visit
+    https://mlflow.org/docs/latest/python_api/mlflow.sagemaker.html#mlflow.sagemaker.SageMakerDeploymentClient
```
dbczumar, line 1064: Indentation nit:

```suggestion
    help_str = """\
    For detailed documentation on the SageMaker deployment client, please visit
    https://mlflow.org/docs/latest/python_api/mlflow.sagemaker.html#mlflow.sagemaker.SageMakerDeploymentClient
```

mlflow/sagemaker/__init__.py
```diff
@@ -1747,6 +1769,704 @@ def _does_model_exist(model_name, sage_client):
...
+        except Exception as exc:
+            raise MlflowException(
+                message=(f"There was an error while retrieving the deployment: {exc}\n")
+            )
+
+    def predict(self, deployment_name, df):
+        raise NotImplementedError("This function is not implemented yet.")
+
+    def explain(self, deployment_name, df):  # pylint: disable=unused-argument
+        raise NotImplementedError("This function is not implemented yet.")
```
dbczumar, line 2467: Can we remove these methods for now until they're implemented? This way, they won't show up in the docs.

dbczumar (approved):
@jamestran201 This is fantastic! I've tried it out and confirmed that roles are now being assumed for all operations and commandline data types are being parsed correctly. The docs render really well, and the examples are very helpful. Thank you for all your hard work here and for introducing thorough test coverage as well.

I've left 3 tiny documentation comments. I'm very excited to merge this once they are addressed. LGTM!

jamestran201:
@dbczumar Nice! Thank you for your guidance!

PR #4999, opened 2021-11-03
Add POST version of GetLatestVersions api

stevenchen-db:
For some backends, when calling the GetLatestVersions api with a stages parameter, only the first stage would be used due to how GET requests handle lists of query params (`name=model&stages=production&stages=staging`) . This PR adds a POST version of the GetLatestVersions api and a new `call_endpoints` util function that tries all endpoints for an API.

## How is this patch tested?

* Unit tests
* Manually tested on Databricks and the OSS server
    * note: the OSS server properly handles the array of stages query param, but some backends do not
 

Components 

Interface 

Language 

Integrations


mlflow/utils/rest_utils.py
```diff
@@ -220,6 +232,15 @@ def call_endpoint(host_creds, endpoint, method, json_body, response_proto):
     return response_proto
 
 
+def call_endpoints(host_creds, endpoints, json_body, response_proto):
+    for i, (endpoint, method) in enumerate(endpoints):
+        try:
+            return call_endpoint(host_creds, endpoint, method, json_body, response_proto)
+        except RestException as e:
+            if e.error_code != ErrorCode.Name(ENDPOINT_NOT_FOUND) or i == len(endpoints) - 1:
```
wentinghu, line 240: how do we guarantee that we try the POST endpoint before the GET endpoint for GetLatestVersions? why do we only check for ENDPOINT_NOT_FOUND here?

stevenchen-db: The order that the request is made is defined by the order in the service proto, so I added in the POST endpoint before the GET. Yeah the ENDPOINT_NOT_FOUND is a good callout. The databricks backend returns ENDPOINT_NOT_FOUND, so I wanted to make sure only this exception would be allowed to pass to prevent this function from masking other exceptions. I'm not sure how other backends handle ENDPOINT_NOT_FOUND though, so maybe there's a better condition to check for. cc @sueann

wentinghu: ooh i see, makes sense!

wentinghu (approved):
lgtm!

Issue #4037, opened 2021-02-01
[BUG]


denisergashbaev:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for additional information about bug reports. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

**Please fill in this bug report template to ensure a timely and thorough response.**

### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [x] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 18.04
- **MLflow installed from (source or binary)**: binary
- **MLflow version (run ``mlflow --version``)**: 1.13.1
- **Python version**: 3.8.5 (but same error on 3.6.5)
- **npm version, if running the dev UI**:
- **Exact command to reproduce**:
follows in the reproducible test case

### Describe the problem
Describe the problem clearly here. Include descriptions of the expected behavior and the actual behavior.
When querying MlFlow installed via Rest API the json returns string values for the attributes that according to documentation should be integers. Consider, for instance, the array of registered models `https://www.mlflow.org/docs/latest/rest-api.html#mlflowregisteredmodel` which is returned when calling `https://www.mlflow.org/docs/latest/rest-api.html#list-registeredmodels`.  The fields `creation_timestamp` and `last_updated_timestamp` are set to strings as shown below:

```
curl http://localhost:5000/api/2.0/mlflow/registered-models/list
{
  "registered_models": [
    {
      "name": "test-model-1",
      "creation_timestamp": "1612177566287",
      "last_updated_timestamp": "1612177566757",
      "latest_versions": [
        {
          "name": "test-model-1",
          "version": "3",
          "creation_timestamp": "1612177566757",
          "last_updated_timestamp": "1612177566757",
          "current_stage": "None",
          "description": "",
          "source": "./mlruns/1/2ced310fbdfe479cb4539a7073a57519/artifacts/test-model-1",
[...]

PR #5010, opened 2021-11-04
Fix a bug that converts int64 to string when converting Protobuf to JSON

lichenran1234:
Signed-off-by: Chenran Li <chenran.li@databricks.com>

According to issue #4037, the returned JSON of the endpoints has `creation_timestamp` and `last_updated_timestamp` as strings, not numbers. It's different from what was documented in the [official doc](https://www.mlflow.org/docs/latest/rest-api.html#mlflowregisteredmodel).

The reason is we are calling Google's `MessageToJson` API to convert protobuf to json, which implicitly converts int64/fixed64/unit64 fields to strings. And they claimed it's a feature not a bug (see the [discussion](https://github.com/protocolbuffers/protobuf/issues/2954#issuecomment-292671661)).

According to the bug reporter, this bug doesn't exist in Azure ML mlflow server (which is essentially our Databricks mlflow server). That's because we are using ScalaPB's `ToJson()` API for all the Databricks endpoints, and it doesn't convert int64 to string.

There is no way to let `MessageToJson` API not convert int64 to strings. Nor are there any other good Python proto-to-json libraries. So to fix this bug, we have to choose from:
* (what I'm doing in this PR) manually converting the int64/uint64/fixed64 fields back to numbers after calling `MessageToJson`
* (too risky so I chose not to do) writing our customized `MessageToJson` API

## How is this patch tested?

unit tests

Components 

Interface 

Language 

Integrations


tests/utils/test_proto_json_utils.py
```diff
@@ -28,6 +29,75 @@ def test_message_to_json():
         "lifecycle_stage": "active",
```
lichenran1234, line 29: The bug wasn't caught by this test case, because there is no int fields here.

tests/utils/test_proto_json_utils.py
```diff
@@ -28,6 +29,75 @@ def test_message_to_json():
...
+                "current_stage": "Staging",
+                "description": "v 2",
+                "user_id": "u2",
+                "source": "A/C",
+                "run_id": "123",
+                "status": "READY",
+            },
+        ],
+    }
+
```
jinzhang21, line 100: Add another line to convert the json back to proto and dict and assert it works?

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,79 @@
...
+            to_dict[key] = from_dict[key]
+    return to_dict
 
 
 def message_to_json(message):
     """Converts a message to JSON, using snake_case for field names."""
-    return MessageToJson(message, preserving_proto_field_name=True)
+
+    # Google's MessageToJson API converts int64/fixed64/unit64 proto fields to JSON strings.
+    json_dict_with_int64_converted_to_str = json.loads(
```
jinzhang21, line 65: json_dict_with_int64_converted_to_str -> json_dict_with_int64_as_str to be consistent with json_dict_with_int64_as_numbers below

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,79 @@
...
+        else:
+            to_dict[key] = from_dict[key]
+    return to_dict
 
 
 def message_to_json(message):
     """Converts a message to JSON, using snake_case for field names."""
-    return MessageToJson(message, preserving_proto_field_name=True)
+
+    # Google's MessageToJson API converts int64/fixed64/unit64 proto fields to JSON strings.
```
jinzhang21, line 64: Add context to why, e.g. citing the comment https://github.com/protocolbuffers/protobuf/issues/2954#issuecomment-292671661

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,79 @@
...
+            for v in value:
+                json_value.append(ftype(v))
+        else:
+            json_value = ftype(value)
+
+        json_dict[field.name] = json_value
+    return json_dict
+
+
+def _merge_json_dicts(from_dict, to_dict):
```
jinzhang21, line 43: Would a package work better here? Not sure about its quality though. https://pypi.org/project/jsonmerge/

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,79 @@
...
+        else:
+            # Skip all non-int64 fields.
+            continue
+
+        if field.label == FieldDescriptor.LABEL_REPEATED:
+            json_value = []
+            for v in value:
+                json_value.append(ftype(v))
+        else:
+            json_value = ftype(value)
```
jinzhang21, line 37: json_value = [ftype(v) for v in value] if field.label == FieldDescriptor.LABEL_REPEATED else ftype(value)

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,79 @@
...
+    """Converts a proto message to JSON, preserving only the int64/uint64/fixed64 fields."""
+
+    json_dict = {}
+    for field, value in proto_message.ListFields():
+        if field.type == FieldDescriptor.TYPE_MESSAGE:
+            ftype = partial(_mark_int64_fields)
+        elif field.type in [
+            FieldDescriptor.TYPE_INT64,
+            FieldDescriptor.TYPE_UINT64,
+            FieldDescriptor.TYPE_FIXED64,
```
jinzhang21, line 25: What about these types?

TYPE_FIXED32 = 7
TYPE_INT32 = 5
TYPE_SFIXED32 = 15
TYPE_SFIXED64 = 16
TYPE_SINT32 = 17
TYPE_SINT64 = 18
TYPE_UINT32 = 13
TYPE_UINT64 = 4

And CPP types:
CPPTYPE_INT32 = 1
CPPTYPE_INT64 = 2
CPPTYPE_UINT32 = 3
CPPTYPE_UINT64 = 4

https://googleapis.dev/python/protobuf/latest/google/protobuf/descriptor.html

jinzhang21 (commented):
Thanks for making the change! A couple minor comments.

tests/utils/test_proto_json_utils.py
lichenran1234: Done. Thanks!

mlflow/utils/proto_json_utils.py
lichenran1234: Done. Thanks!

mlflow/utils/proto_json_utils.py
lichenran1234: Good point! Done

mlflow/utils/proto_json_utils.py
lichenran1234: Done, thanks!

mlflow/utils/proto_json_utils.py
lichenran1234: For int32 types, according to the [doc](https://developers.google.com/protocol-buffers/docs/proto3#json), they won't be converted to JSON strings so we don't need to add them here.

For CPP types, they are from a difference enum `FieldDescriptor::CppType` rather than `FieldDescriptor::Type`. They are used for `field.cpp_type()`. But here we only care about `field.type()`.
Screen Shot 2021-11-04 at 19 11 17

I added two more int64 types (`TYPE_SFIXED64` and `TYPE_SINT64`) to cover all the int64 types.

mlflow/utils/proto_json_utils.py
lichenran1234: Unfortunately that package has some weird bugs: it dropped lots of fields when merging.

jinzhang21: Np. Just to clarify options. Custom implementation is fine.

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,81 @@
...
+        )
+    return json_dict
+
+
+def _merge_json_dicts(from_dict, to_dict):
+    """Merges the json elements of from_dict into to_dict. Only works for json dicts
+    converted from proto messages
+    """
+
+    for key in from_dict:
```
jinzhang21, line 47: for key in from_dict: -> for key, value in from_dict.items():

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,81 @@
...
+        value = to_dict[key]
+        if isinstance(value, dict):
+            _merge_json_dicts(from_dict[key], to_dict[key])
+        elif isinstance(value, list):
```
jinzhang21, line 51: Could value be a tuple?

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,81 @@
...
+            for i in range(len(value)):
+                if isinstance(value[i], dict):
+                    _merge_json_dicts(from_dict[key][i], to_dict[key][i])
+                else:
+                    to_dict[key][i] = from_dict[key][i]
```
jinzhang21, line 56: Enumerate seems to be simpler here:

for i, v in enumerate(value):
  if isinstance(v, dict):
     _merge_json_dicts(v, to_dict[key][i])
  else:
     to_dict[key][i] = v

jinzhang21 (commented):
LGTM! Thanks for the change, Chenran! There are some tests are failing. Please fix those.

dbczumar:
@harupy FYI

mlflow/utils/proto_json_utils.py
```diff
@@ -1,17 +1,81 @@
...
+
+        json_dict[field.name] = (
+            [ftype(v) for v in value]
+            if field.label == FieldDescriptor.LABEL_REPEATED
+            else ftype(value)
+        )
+    return json_dict
+
+
+def _merge_json_dicts(from_dict, to_dict):
```
harupy, line 42: Hi @lichenran1234, can we add a test for `_merge_json_dicts` to make sure it works properly?

mlflow/utils/proto_json_utils.py
lichenran1234: Good point! Done

mlflow/utils/proto_json_utils.py
lichenran1234: Good question! According to [this page](https://realpython.com/python-json/), Python dict constructed from a JSON string cannot have tuples:
Screen Shot 2021-11-05 at 10 49 31

mlflow/utils/proto_json_utils.py
lichenran1234: Done, thanks!

mlflow/utils/proto_json_utils.py
lichenran1234: Thanks for the comment, Harupy! Usually we don't write unit tests for private functions. Especially for this `_merge_json_dicts` function: the code inside it should be embedded inside `message_to_json` function, but I'm extracting it out for readability. So I guess we should follow the [Test Behavior, Not Implementation](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html) principle.

Can you think of more test cases for `message_to_json` if you are worried that `_merge_json_dicts` may not work properly?

jinzhang21: I agree with @lichenran1234 that don't test private function. But perhaps add more types of field to the unit test to verify that the public API works for all kinds of fields? e.g. we might want to verify that the following aspects are correctly translated:

- default values (I worry about this one)
- extensions (and this one)
- proto maps
- oneof
- enums

harupy: @lichenran1234 Makes sense to not test `_merge_json_dicts`, thanks for the knowledge!

lichenran1234: Thanks @harupy and @jinzhang21 ! I added a new test proto message so I can test this function extensively. I also added support for proto maps.

mlflow/protos/protos_for_test/test_message.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto2";
```
harupy, line 1: Is it possible to move this file under tests?

jinzhang21 (approved):
Thanks a lot for addressing this issue, @lichenran1234 ! Could you please move the test protos to mlflow/tests/protos as @harupy suggested? The rest LGTM! Feel free to merge after the change.

harupy (approved):
LGTM!

mlflow/protos/protos_for_test/test_message.proto
lichenran1234: Good point! Done!

PR #5012, opened 2021-11-05
Add optional experiment_id parameter to `mlflow.set_experiment`

dbczumar:

Add optional experiment_id parameter to `mlflow.set_experiment`, allowing users to set an experiment by ID.

## How is this patch tested?

Included unit tests

## Release Notes

Add an optional `experiment_id` parameter to `mlflow.set_experiment()`, enabling users to set an experiment by ID.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -65,6 +65,8 @@ def set_experiment(experiment_name: str = None, experiment_id: str = None) -> No
                             created.
     :param experiment_id: ID of the experiment to be activated. If an experiment with this ID
                           does not exist, an exception is thrown.
+    :return: An instance of :py:class:`mlflow.entities.Experiment` representing the new active
+             experiment.
 
     .. code-block:: python
         :caption: Example
@@ -97,24 +99,10 @@ def set_experiment(experiment_name: str = None, experiment_id: str = None) -> No
             error_code=INVALID_PARAMETER_VALUE,
         )
 
-    def verify_experiment_active(experiment):
-        if experiment.lifecycle_stage != LifecycleStage.ACTIVE:
-            raise MlflowException(
-                message=(
-                    "Cannot set a deleted experiment '%s' as the active experiment."
-                    " You can restore the experiment, or permanently delete the "
-                    " experiment to create a new one." % experiment.name
-                ),
-                error_code=INVALID_PARAMETER_VALUE,
-            )
-
     client = MlflowClient()
     if experiment_id is None:
         experiment = client.get_experiment_by_name(experiment_name)
-        if experiment:
-            verify_experiment_active(experiment)
-            experiment_id = experiment.experiment_id
-        else:
+        if not experiment:
             _logger.info(
                 "Experiment with name '%s' does not exist. Creating a new experiment.",
                 experiment_name,
@@ -123,6 +111,7 @@ def set_experiment(experiment_name: str = None, experiment_id: str = None) -> No
             # simultaneously, a race condition may be encountered here wherein experiment creation
             # fails
             experiment_id = client.create_experiment(experiment_name)
+            experiment = client.get_experiment(experiment_id)
     else:
         experiment = client.get_experiment(experiment_id)
         if experiment is None:
@@ -130,10 +119,20 @@ def set_experiment(experiment_name: str = None, experiment_id: str = None) -> No
                 message=f"Experiment with ID '{experiment_id}' does not exist.",
                 error_code=RESOURCE_DOES_NOT_EXIST,
             )
-        verify_experiment_active(experiment)
+
+    if experiment.lifecycle_stage != LifecycleStage.ACTIVE:
+        raise MlflowException(
+            message=(
+                "Cannot set a deleted experiment '%s' as the active experiment."
+                " You can restore the experiment, or permanently delete the "
+                " experiment to create a new one." % experiment.name
+            ),
+            error_code=INVALID_PARAMETER_VALUE,
+        )
 
     global _active_experiment_id
-    _active_experiment_id = experiment_id
+    _active_experiment_id = experiment.experiment_id
+    return experiment
 
 
 class ActiveRun(Run):  # pylint: disable=W0223
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -89,7 +89,9 @@ def set_experiment(experiment_name: str = None, experiment_id: str = None) -> No
         Tags: {}
         Lifecycle_stage: active
     """
-    if (experiment_name is not None and experiment_id is not None) or (experiment_name is None and experiment_id is None):
+    if (experiment_name is not None and experiment_id is not None) or (
+        experiment_name is None and experiment_id is None
+    ):
         raise MlflowException(
             message="Must specify exactly one of: `experiment_id` or `experiment_name`.",
             error_code=INVALID_PARAMETER_VALUE,
--- a/mlflow/tracking/fluent.py
+++ b/mlflow/tracking/fluent.py
@@ -98,7 +98,7 @@ def set_experiment(experiment_name: str = None, experiment_id: str = None) -> No
         )
 
     def verify_experiment_active(experiment):
-        if experiment.lifecycle_stage == LifecycleStage.DELETED:
+        if experiment.lifecycle_stage != LifecycleStage.ACTIVE:
             raise MlflowException(
                 message=(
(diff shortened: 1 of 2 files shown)


tests/tracking/test_tracking.py
```diff
@@ -100,16 +91,66 @@ def test_set_experiment():
...
+    nonexistent_id = "-1337"
+    with pytest.raises(MlflowException) as exc:
+        mlflow.set_experiment(experiment_id=nonexistent_id)
+    assert exc.value.error_code == ErrorCode.Name(RESOURCE_DOES_NOT_EXIST)
+    with start_run() as run:
+        assert run.info.experiment_id == exp_id
+
+
+def test_set_experiment_parameter_validation():
+    with pytest.raises(MlflowException) as exc:
```
BenWilson2, line 111: Perhaps use the optional match= in `pytest.raises` to do regex matching for the exception error message?
```python
with pytest.raises.(MLflowException, match="Must specify exactly one") as exc:
    mlflow.set_experiment()
assert exc.value.error_Code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
```
Haru asked that I change a few newer unit tests to use this pattern.

mlflow/tracking/fluent.py
```diff
@@ -80,20 +89,48 @@ def set_experiment(experiment_name: str) -> None:
...
         raise MlflowException(
-            "Cannot set a deleted experiment '%s' as the active experiment."
-            " You can restore the experiment, or permanently delete the "
-            " experiment to create a new one." % experiment.name
+            message="Must specify exactly one of: `experiment_id` or `experiment_name`.",
+            error_code=INVALID_PARAMETER_VALUE,
         )
+
+    def verify_experiment_active(experiment):
+        if experiment.lifecycle_stage == LifecycleStage.DELETED:
```
jinzhang21, line 101: if experiment.lifecycle_stage != LifecycleStage.ACTIVE:
to be more specific, since the name of the function is to verify it's active or not?

mlflow/tracking/fluent.py
```diff
@@ -80,20 +89,48 @@ def set_experiment(experiment_name: str) -> None:
...
+                    " experiment to create a new one." % experiment.name
+                ),
+                error_code=INVALID_PARAMETER_VALUE,
+            )
+
+    client = MlflowClient()
+    if experiment_id is None:
+        experiment = client.get_experiment_by_name(experiment_name)
+        if experiment:
+            verify_experiment_active(experiment)
```
jinzhang21, line 115: How about verify outside of the if-else block?

mlflow/tracking/fluent.py
```diff
@@ -80,20 +89,48 @@ def set_experiment(experiment_name: str) -> None:
...
+            experiment_id = experiment.experiment_id
+        else:
+            _logger.info(
+                "Experiment with name '%s' does not exist. Creating a new experiment.",
+                experiment_name,
+            )
+            experiment_id = client.create_experiment(experiment_name)
```
jinzhang21, line 122: There's a potential race condition here between client.get_experiment_by_name() and client.create_experiment(experiment_name). It could cause issues in distributed logging situation. I don't think you need to fix it here, but might be good to annotate the code to make future life easier.

mlflow/tracking/fluent.py
```diff
@@ -80,20 +89,48 @@ def set_experiment(experiment_name: str) -> None:
...
+    else:
+        experiment = client.get_experiment(experiment_id)
```
jinzhang21, line 124: Does this function throw if experiment_id doesn't exist?

mlflow/tracking/fluent.py
```diff
@@ -80,20 +89,48 @@ def set_experiment(experiment_name: str) -> None:
...
         Lifecycle_stage: active
     """
-    client = MlflowClient()
-    experiment = client.get_experiment_by_name(experiment_name)
-    exp_id = experiment.experiment_id if experiment else None
-    if exp_id is None:  # id can be 0
-        print("INFO: '{}' does not exist. Creating a new experiment".format(experiment_name))
-        exp_id = client.create_experiment(experiment_name)
-    elif experiment.lifecycle_stage == LifecycleStage.DELETED:
+    if (experiment_name is not None and experiment_id is not None) or (
```
jinzhang21, line 92: Is empty string legit for either name or id? I guess not. In that case, it might be better to verify
if not (experiment_name and experiment_id)

jinzhang21 (approved):
Thanks for making the change, Corey! LGTM!

mlflow/tracking/fluent.py
sunishsheth2009: I am curious on why do we need this condition? We just need `experiment_name is None and experiment_id is None` right? Let me know if I am missing something.

mlflow/tracking/fluent.py
sunishsheth2009: I think we might need to check if experiment is present to check outside of the if block or we need to replicate that if experiment check in the `verify_experiment_active` function as far as I can tell.

sunishsheth2009 (approved):

mlflow/tracking/fluent.py
BenWilson2: 404 in REST and throws in file_store and sqlalchemy_store

mlflow/tracking/fluent.py
BenWilson2: Wouldn't the guarded exclusivity check protect against conflicting submission behavior. For an invalid id but a valid name it would raise an Exception based on the id validation. Could be super confusing for users.

dbczumar: > Is empty string legit for either name or id? I guess not. In that case, it might be better to verify
if not (experiment_name and experiment_id)

I would imagine that it's invalid on most if not all backends, but I'm hesitant to enforce this on the off chance that someone's third-party backend has a legitimate use case for this. For example, MLflow's default experiment used to be a falsey integer value (`0`) before the change to string representations.

mlflow/tracking/fluent.py
dbczumar: Done! Good idea :)

mlflow/tracking/fluent.py
dbczumar: The challenge here is that we'd need to re-fetch the experiment in order to perform the activity check in the case where a new experiment was created, since `client.create_experiment` returns an experiment ID, rather than the whole experiment entity. In the interest of saving a network call, I've embedded `verify_experiment_active` within a couple interior branches.

mlflow/tracking/fluent.py
dbczumar: This is a good call-out. It's worth noting that the MLflow fluent API in general is not meant to be safe across threads or processes. I've added a note here nonetheless.

mlflow/tracking/fluent.py
dbczumar: Thanks @BenWilson2 ! Yes, while this throws for MLflow-native stores, it may not always throw for alternative stores. I don't think the `None` check is particularly problematic here as an extra layer of defense.

tests/tracking/test_tracking.py
dbczumar: Awesome suggestion! Done!

mlflow/tracking/fluent.py
dbczumar: Because the same logic applies for experiment names, I've removed the test case asserting that an empty string name is invalid, as that's backend-dependent.

dbczumar (commented):
@jinzhang21 @BenWilson2 @sunishsheth2009 Thank you for your thorough reviews! I've addressed your comments. I'm going to push one more change so that `set_experiment` also returns the experiment entity.

mlflow/tracking/fluent.py
jinzhang21: SGTM.

mlflow/tracking/fluent.py
```diff
@@ -123,17 +111,28 @@ def verify_experiment_active(experiment):
...
+                " You can restore the experiment, or permanently delete the "
+                " experiment to create a new one." % experiment.name
+            ),
+            error_code=INVALID_PARAMETER_VALUE,
+        )
 
     global _active_experiment_id
-    _active_experiment_id = experiment_id
+    _active_experiment_id = experiment.experiment_id
+    return experiment
```
BenWilson2, line 135: This actually simplifies some MLOps pipeline flows! A bonus feature is always a great thing

BenWilson2 (approved):
LGTM!

PR #5014, opened 2021-11-07
Clean up pylint command

harupy:

Clean up pylint command by specifying required options in `pylintrc`.

## How is this patch tested?

NA

Components 

Interface 

Language 

Integrations

### DIFF
--- a/dev/set_matrix.py
+++ b/dev/set_matrix.py
@@ -316,8 +316,7 @@ def process_requirements(requirements, version=None):
             )
             if match_all:
                 return packages
-        else:
-            return []
+        return []
 
     raise TypeError("Invalid object type for `requirements`: '{}'".format(type(requirements)))
 
--- a/dev/show_package_release_dates.py
+++ b/dev/show_package_release_dates.py
@@ -6,7 +6,7 @@ import traceback
 
 
 def get_distributions():
-    res = subprocess.run(["pip", "list"], stdout=subprocess.PIPE)
+    res = subprocess.run(["pip", "list"], stdout=subprocess.PIPE, check=True)
     pip_list_stdout = res.stdout.decode("utf-8")
     # `pip_list_stdout` looks like this:
     # ``````````````````````````````````````````````````````````
--- a/lint.sh
+++ b/lint.sh
@@ -5,30 +5,6 @@
 err=0
 trap 'err=1' ERR
 
-FWDIR="$(cd "`dirname $0`"; pwd)"
-cd "$FWDIR"
-
-# https://stackoverflow.com/a/17841619
-function join {
-  local d=$1
-  shift
-  echo -n "$1"
-  shift
-  printf "%s" "${@/#/$d}"
-}
-
-include_dirs=(
-  "mlflow"
-  "tests"
-)
-
-exclude_dirs=(
-  "mlflow/protos"
-  "mlflow/server/js"
-  "mlflow/store/db_migrations"
-  "mlflow/temporary_db_migrations_for_pre_1_users"
-)
-
 echo -e "\n========== black ==========\n"
 # Exclude proto files because they are auto-generated
 black --check .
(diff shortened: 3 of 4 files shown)


lint.sh
```diff
@@ -38,23 +14,8 @@ if [ $? -ne 0 ]; then
...
-# This behavior might cause us to unintentionally ignore some files.
-# To avoid this issue, select files to lint using `git ls-files` and `grep`.
-# Another advantage of this approach is we can apply pylint to all python scripts
-# without creating `__init__.py` in all directories.
-exclude="^\($(join "\|" "${exclude_dirs[@]}")\)/.\+\.py$"
-include="^\($(join "\|" "${include_dirs[@]}")\)/.\+\.py$"
-msg_template="{path} ({line},{column}): [{msg_id} {symbol}] {msg}"
-
-pylint --jobs=0 --msg-template="$msg_template" --rcfile="$FWDIR/pylintrc" $(git ls-files | grep $include | grep -v $exclude)
+pylint $(git ls-files | grep '\.py$')
```
harupy, line 18: This allows us to run `pylint $(git diff --name-only| grep '\.py$')` before making a commit to make sure there is no pylint error.

harupy: You can also run `pylint $(git diff --name-only | grep '\.py$')` using a pre-commit for Git.

lint.sh
```diff
@@ -38,23 +14,8 @@ if [ $? -ne 0 ]; then
...
-# see: http://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options
-# This behavior might cause us to unintentionally ignore some files.
-# To avoid this issue, select files to lint using `git ls-files` and `grep`.
-# Another advantage of this approach is we can apply pylint to all python scripts
-# without creating `__init__.py` in all directories.
-exclude="^\($(join "\|" "${exclude_dirs[@]}")\)/.\+\.py$"
-include="^\($(join "\|" "${include_dirs[@]}")\)/.\+\.py$"
-msg_template="{path} ({line},{column}): [{msg_id} {symbol}] {msg}"
-
-pylint --jobs=0 --msg-template="$msg_template" --rcfile="$FWDIR/pylintrc" $(git ls-files | grep $include | grep -v $exclude)
```
harupy, line 57: I made the following changes:

- Move `--jobs` is specified in `pylintrc`.
- Removed `--msg-template` since the default template is good enough.
- Removed `--rcfile` because `pylint` automatically detects `pylintrc`.

pylintrc
```diff
@@ -7,19 +7,28 @@ extension-pkg-whitelist=
 
 # Add files or directories to the denylist. They should be base names, not
 # paths.
-ignore=build,protos,sdk,db_migrations,temporary_db_migrations_for_pre_1_users
-
+ignore=
```
harupy, line 10: Switched to using `ignore-paths` which is more accurate because it matches against paths, not base names.

dev/set_matrix.py
```diff
@@ -316,8 +316,7 @@ def process_requirements(requirements, version=None):
             )
             if match_all:
                 return packages
-        else:
-            return []
+        return []
```
harupy, line 319: Required this change because `pylint` now checks scripts in the `dev` directory.

dev/show_package_release_dates.py
```diff
@@ -6,7 +6,7 @@
 
 
 def get_distributions():
-    res = subprocess.run(["pip", "list"], stdout=subprocess.PIPE)
+    res = subprocess.run(["pip", "list"], stdout=subprocess.PIPE, check=True)
```
harupy, line 9: Fixed this for the same reason as https://github.com/mlflow/mlflow/pull/5014/files#r744341750.

dev/set_matrix.py
WeichenXu123: So pylint print warning for `for ... else` usage ?

WeichenXu123 (approved):
LGTM

dev/set_matrix.py
harupy: No, pylint raises a warning when we use `for-else` without a `break` statement in the loop:

https://docs.quantifiedcode.com/python-anti-patterns/correctness/else_clause_on_loop_without_a_break_statement.html

```python
# Example

def contains_magic_number(list, magic_number):
    for i in list:
        if i == magic_number:
            print("This list contains the magic number")
    else:
        print("This list does NOT contain the magic number")
```

PR #5017, opened 2021-11-07
Automatically update metric plots for in-progress runs #2099

### LINKED ISSUE #2099: [FR] Automatically update metric plots for in-progress runs
## MLflow Roadmap Item

This is an MLflow Roadmap item that has been prioritized by the MLflow maintainers. We're seeking help with the implementation of roadmap items tagged with the `help wanted` label. Items tagged with the `needs design` label are awaiting a design sketch from an MLflow maintainer: @harupy.

For requirements clarifications and implementation questions, or to request a PR review, please tag @harupy in your communications related to this issue.

## Describe the proposal
Automatically update MLflow metric plots for in-progress runs periodically, enabling users to track the performance of in-progress runs without having to perform periodic page refreshes and reset UI state.

### Motivation
It would be great to be able to monitor training progress via metric graphs for a particular set of runs.

### Proposed Changes
Currently when i plot a chart say for loss of an in-progress run, the MLflow UI creates a snapshot of the current status of the training. However the chart will not automatically update with new data. The browser must be manually refreshed in order to display updated data. However, updating the chart causes certain plot state & settings to be lost.

cedkoffeto:
Signed-off-by: Cedric Koffeto cedkoffeto@gmail.com

setInterval function added in metricsPlotPanel component which retrieves execution status and metrics history from  API  every 8 seconds to update the plot and automatically stops when all executions are complete. closes #2099

## How is this patch tested?

- Unit tests
- Manually verified the proposed feature works correctly using this script:

```python
import time
import numpy as np
import mlflow
import multiprocessing as mp


def log(run_id, slope, repeat):
    sleep = 10
    with mlflow.start_run(run_id=run_id):
        for epoch in range(1, repeat + 1):
            print(epoch)
            mlflow.log_metric(key="metric1", value=slope * epoch * np.log(epoch), step=epoch)
            mlflow.log_metric(key="metric2", value=slope * (1 / epoch) * np.log(epoch), step=epoch)
            time.sleep(sleep)


run_uuids = [client.create_run("0").info.run_id for _ in range(2)]
runs_param = "[" + ",".join(map(lambda s: f"%22{s}%22", run_uuids)) + "]"

print(
    "URL:",
    r"http://localhost:3000/#/metric/metric1?runs=<<< runs_param >>>&experiment=0&plot_metric_keys=[%22metric1%22]&plot_layout={%22autosize%22:true,%22xaxis%22:{},%22yaxis%22:{}}&x_axis=step&y_axis_scale=linear&line_smoothness=1&show_point=true&deselected_curves=[]&last_linear_y_axis_range=[]".replace(
        "<<< runs_param >>>", runs_param
    ),
)

args_list = [(run_uuid, idx + 1, 5 + idx * 3) for idx, run_uuid in enumerate(run_uuids)]

with mp.Pool() as pool:
    pool.starmap(log, args_list)
```

It can be tested with a unit test to check if the plot is actually updated

Automatically update metric plots for in-progress runs

Components 

Interface 

Language 

Integrations

<a name="UI"></a>


harupy:
Hi @cedkoffeto, thanks for the PR! I'll review it soon :)

cedkoffeto:
> Hi @cedkoffeto, thanks for the PR! I'll review it soon :)

@harupy Glad to be able to help 😉

harupy:
@cedkoffeto btw could you take a screen record of how the plot automatically gets updated?

cedkoffeto:
> @cedkoffeto btw could you take a screen record of how the metric plot automatically gets updated?

Here it is @harupy 


https://user-images.githubusercontent.com/38736306/140674522-d039267c-f832-4ead-a7f3-19c28abb0ee4.mp4

harupy:
Thanks for the screen recording. It looks like the entire plot gets rerendered.

dbczumar:
@cedkoffeto @harupy Awesome stuff! Does the proposal from https://github.com/harupy/mlflow/tree/5017-harupy also preserve plot customizations and zoom?

cedkoffeto:
> this.setState((prevState) => ({historyRequestIds: [...]}))

Hi @harupy,
In fact, I also think that saving requests seems unnecessary in our case.

cedkoffeto:
> @cedkoffeto @harupy Awesome stuff! Does the proposal from https://github.com/harupy/mlflow/tree/5017-harupy also preserve plot customizations and zoom?

Thanks, @dbczumar
I think it does preserve plot customizations and zoom but let @harupy  confirm.

harupy:
Yep, it does. Here's a guick demo.

https://user-images.githubusercontent.com/17039389/140839788-166fe709-dc97-4477-8e9d-2002c0289f15.mov

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
```diff
@@ -63,6 +65,8 @@ export class MetricsPlotPanel extends React.Component {
   // a single-click event.
   SINGLE_CLICK_EVENT_DELAY_MS = this.MAX_DOUBLE_CLICK_INTERVAL_MS + 10;
 
+  DURATION_BETWEEN_HISTORY_UPDATES_MS = 8000;
```
harupy, line 68: @cedkoffeto Is there any specific reason that you chose 8000?

cedkoffeto: Hi @harupy  
Since the refresh of the plot was not as interesting to see, 5 or 3 seconds was a little too fast so I increased to 8 because 10 was also a little too long as the waiting time so I was trying to find a trade-off

but now with your version, the refresh seems to be way better so there is no need to have a high delay.

Hope it makes sense

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
```diff
@@ -459,6 +498,18 @@ export class MetricsPlotPanel extends React.Component {
...
+
+    let boolean = false;
+    runs &&
+      runs.forEach((run) => {
+        if (run.getStatus() === 'RUNNING') boolean = true;
+      });
+
+    return boolean;
+  }
+
```
harupy, line 512:
```suggestion
  checkOnRunUnfinished() {
    const { runs } = this.props;

    let boolean = false;
    runs &&
      runs.forEach((run) => {
        if (run.getStatus() === 'RUNNING') boolean = true;
      });

    return boolean;
  }

```

Can we remove this function since it's no longer used?

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
harupy: I think this does depend on how often you log metrics. For example, when you train a large deep learning model and logs metrics by epoch, 3 seconds might be too short.

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
```diff
@@ -71,11 +76,45 @@ export class MetricsPlotPanel extends React.Component {
...
+      this.intervalId = setInterval(() => {
+        const { runUuids } = this.props;
+        // Use `message` for demo purposes.
+        message.success({
+          content: 'Updated',
+          duration: 0.5,
+          style: {
+            marginTop: '20vh',
+          },
+        });
```
harupy, line 103: @cedkoffeto I added this block for demo purposes, but do you think it's useful if the UI shows a message every time it updates the plot?

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
cedkoffeto: @cedkoffeto  I do agree with you 
So we can keep 3 or increase to 5 to handle quick updates cases (3s is ok for me btw).

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
cedkoffeto: @harupy  Not that useful in my point of view

I think we can show the status (live) somewhere maybe in a green box something like that to specify that there are executions not yet completed and only keep the completion message.
```python
message.success('All runs completed');
```

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
cedkoffeto: Sure

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
harupy: > keep the completion message.

I agree!

> I think we can show the status (live) somewhere maybe in a green box something like that to specify that there are executions not yet completed and only keep the completion message.

This sounds good to me :) I'll take to the team about this.

cedkoffeto: @harupy Ok  good 👍🏽

cedkoffeto:
Hi @harupy, any update?

harupy:
Hi @cedkoffeto, sorry for the late reply.

https://user-images.githubusercontent.com/17039389/143550744-79d5244d-e7eb-4691-b8e2-0e67ef13c5ae.mov

cedkoffeto:
> Hi @cedkoffeto, sorry for the late reply. We internally discussed this feature. Here's our latest prototype:
> 
>  automatic-metric-plot-update.mov 
> code: https://github.com/harupy/mlflow/pull/28/files

Hi @harupy, that's great!

harupy:
@cedkoffeto I pushed some commits to update the PR.

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
```diff
@@ -22,10 +22,16 @@ import { getUUID } from '../../common/utils/ActionUtils';
 export const CHART_TYPE_LINE = 'line';
 export const CHART_TYPE_BAR = 'bar';
 
+// Polling interval
+export const METRICS_PLOT_POLLING_INTERVAL_MS = 5000;
```
dbczumar, line 26: Can we increase this to 10 seconds?

dbczumar: In general, what happens if the refresh fails? Does the page crash?

cedkoffeto:
> @cedkoffeto I pushed some commits to update the PR.
👍🏽

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
harupy: > Can we increase this to 10 seconds? 5 seems aggressive.

Sure!

harupy: > In general, what happens if the refresh fails? Does the page crash?

Let me test.

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
```diff
@@ -22,10 +22,16 @@ import { getUUID } from '../../common/utils/ActionUtils';
...
+// Stop polling when the polling duration exceeds this value
+export const METRICS_PLOT_POLLING_DURATION_MS = 3600 * 1000; // 1 hour
```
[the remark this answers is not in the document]
harupy: @dbczumar I remember we discussed that we should set an appropriate polling threshold when a run never ends, but I don't think we need to because runs usually end.

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
harupy: https://user-images.githubusercontent.com/17039389/143964115-8fdd77c0-6b67-4ecf-afcd-61cf4c158bdd.mov

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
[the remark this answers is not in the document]
harupy: Offline discussion: check the timestamp of the last metric, and if it's more than 1 week, then we won't refresh.

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
```diff
@@ -22,10 +22,16 @@ import { getUUID } from '../../common/utils/ActionUtils';
 export const CHART_TYPE_LINE = 'line';
 export const CHART_TYPE_BAR = 'bar';
 
+export const METRICS_PLOT_POLLING_INTERVAL_MS = 10 * 1000; // 10 seconds
+// A run is considered as 'hanging' if its status is 'RUNNING' but its latest metric was logged
+// prior to this threshold. The metrics plot doesn't automatically update hanging runs.
+export const METRICS_PLOT_HANGING_RUN_THRESHOLD_MS = 3600 * 24 * 7 * 1000; // 1 week
```
harupy, line 28: Does "hanging" make sense?

mlflow/server/js/src/i18n/default/en.json
```diff
@@ -611,6 +611,10 @@
     "defaultMessage": "Registered Models",
     "description": "Text for registered model link in the title for model comparison page"
   },
+  "UEDu0c": {
+    "defaultMessage": "MLflow UI automatically fetches metric histories for active runs and updates the metrics plot with a {interval} second interval.",
```
harupy, line 615: Included `interval` so a user doesn't need to guess or measure how long the interval is.

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.js
dbczumar: Yep!

mlflow/server/js/src/experiment-tracking/components/MetricsPlotPanel.test.js
```diff
@@ -22,100 +37,123 @@ describe('unit tests', () => {
...
       metricsWithRunInfoAndHistory: [
+        // Metrics for runUuid1
         {
           metricKey: 'metric_1',
           history: [
             /* Intentionally reversed timestamp and step here for testing */
-            { key: 'metric_1', value: 100, step: 2, timestamp: 1556662044000 },
-            { key: 'metric_1', value: 50, step: 1, timestamp: 1556662043000 },
+            { key: 'metric_1', value: 100, step: 2, timestamp: now },
+            { key: 'metric_1', value: 50, step: 1, timestamp: now - 1 },
```
harupy, line 67: Replaced hardcoded timestamps with `now` to prevent the metrics plot from considering these runs as hanging.

dbczumar (approved):
LGTM! Awesome work, @cedkoffeto, @harupy ! Thank you so much for this contribution, @cedkoffeto!

cedkoffeto:
> much

Thanks! It was a pleasure :)  
Thanks also to @harupy for your great help 🙏🏽

Issue #4987, opened 2021-11-02
[SETUP-BUG] Issue in pip dependency resolution of test-requirements.txt


mhemendra:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for information on what types of issues we address.

**Please fill in this installation issue template to ensure a timely and thorough response.**

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Windows 11
- **MLflow installed from (source or binary)**: NA
- **MLflow version (run ``mlflow --version``)**: NA
- **Python version**: 3.6.13
- **Exact command to reproduce**:
- 
conda create --name mlflow-dev-env python=3.6
conda activate mlflow-dev-env
pip install -r dev-requirements.txt
pip install -r test-requirements.txt

### Describe the problem
Provide the exact sequence of commands / steps that you executed before running into the problem.

The setup step ran fine till dev-requirements.txt, but when i ran for test-requirements.txt, pip setup runs for hours but never finishes, it was downloading all the versions of almost all libraries. I could see lots of messages like "pip is looking at multiple versions of <library_name>" for many libraries.
When i tried to use pip-compile test-requirements.txt as suggested in https://stackoverflow.com/questions/65122957/resolving-new-pip-backtracking-runtime-issue, I'm getting the below error

Could not find a version that matches gast==0.3.3,==0.4.0 (from tensorflow==2.6.1->-r dev/extra-ml-requirements.txt (line 18))
Tried: 0.1, 0.1.1, 0.1.2, 0.1.3, 0.1.4, 0.1.5, 0.2.0, 0.2.1, 0.2.1.post0, 0.2.1.post1, 0.2.2, 0.3.0, 0.3.1, 0.3.2, 0.3.3, 0.3.3, 0.4.0, 0.4.0, 0.5.0, 0.5.0, 0.5.1, 0.5.1, 0.5.2, 0.5.2
There are incompatible versions in the resolved dependencies:
  gast==0.4.0 (from tensorflow==2.6.1->-r dev/extra-ml-requirements.txt (line 18))
  gast==0.3.3 (from paddlepaddle==2.1.3->-r dev/extra-ml-requirements.txt (line 44))

Please let me know how to resolve this issue.

### Other info / logs
Include any logs or source code that would be helpful to diagnose the problem. If including tracebacks, please include the full traceback. Large logs and files should be attached.

[...]

PR #5048, opened 2021-11-11
Resolve getting-started package install issues #4987

maitre-matt:
Signed-off-by: Matthieu Maitre <mmaitre@microsoft.com>

The change addresses the near-infinite pip backtracking when following the getting-started instructions (#4987). `pip-compile` is introduced in the getting-started steps to resolve all the dependencies at once. Dependencies are pinned and adjusted to resolve conflicts.

Caveats:
- I was not able to install on Windows as PyStan was not building.
- I was able to install on Ubuntu on Windows (WSL). There I ended up having to elevate `sudo pip install -e .[extras]` to resolve access-denied errors on output directories. Not sure if this is specific to my setup or something that should be mentioned in the getting-started.

I removed the `pip install pytest` as it is already covered by `test-requirements.txt`.

## How is this patch tested?

Ran the getting-started steps on Ubuntu in WSL.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -147,8 +147,9 @@ If you plan on doing development and testing, you will also need to install the 
 
 .. code-block:: bash
 
-    pip install -r dev-requirements.txt
-    pip install -r test-requirements.txt
+    pip install pip-tools
+    pip-compile --output-file compiled-requirements.txt dev-requirements.txt test-requirements.txt
+    pip install -r compiled-requirements.txt
     pip install -e .[extras]  # installs mlflow from current checkout
     pip install -e tests/resources/mlflow-test-plugin # installs `mlflow-test-plugin` that is required for running certain MLflow tests
 
@@ -156,11 +157,7 @@ You may need to run ``conda install cmake`` for the test requirements to properl
 
 Ensure `Docker <https://www.docker.com/>`_ is installed.
 
-Finally, we use ``pytest`` to test all Python contributed code. Install ``pytest``:
-
-.. code-block:: bash
-
-    pip install pytest
+Finally, we use ``pytest`` to test all Python contributed code.
 
 JavaScript and UI
 ~~~~~~~~~~~~~~~~~
--- a/dev/extra-ml-requirements.txt
+++ b/dev/extra-ml-requirements.txt
@@ -8,14 +8,13 @@ keras
 # Required by mlflow.sklearn
 scikit-learn
 # Required by mlflow.gluon
-# TODO: Unpin once https://github.com/apache/incubator-mxnet/issues/20068 is fixed
-mxnet!=1.8.0
+mxnet==1.8.0.post0
 # Required by mlflow.fastai
 fastai>=2.4.1
 # Required by mlflow.spacy
 spacy
 # Required by mlflow.tensorflow
-tensorflow
+tensorflow==2.4.4
 # TODO: Stop excluding tensorflow-estimator 2.7.0 once tensorflow 2.7.0 is released:
 # https://github.com/tensorflow/tensorflow/issues/52883#issuecomment-955802164
 tensorflow-estimator!=2.7.0
--- a/dev/lint-requirements.txt
+++ b/dev/lint-requirements.txt
@@ -1,5 +1,6 @@
 pyarrow==0.17.0
-pylint==2.11.1
+pylint==2.1.0
+astroid==2.7.3
 rstcheck==3.2
 black==19.10b0
 # Pin isort version to prevent pylint from raising an error


dev/lint-requirements.txt
```diff
@@ -1,5 +1,6 @@
 pyarrow==0.17.0
-pylint==2.11.1
+pylint==2.1.0
+astroid==2.7.3
```
dbczumar, line 3: Do we need to update these? Downgrading pylint may require reformatting portions of the codebase. pinning astroid doesn't seem strictly essential for linting.

dev/extra-ml-requirements.txt
```diff
@@ -8,14 +8,13 @@ keras
...
-# TODO: Unpin once https://github.com/apache/incubator-mxnet/issues/20068 is fixed
-mxnet!=1.8.0
+mxnet==1.8.0.post0
 # Required by mlflow.fastai
 fastai>=2.4.1
 # Required by mlflow.spacy
 spacy
 # Required by mlflow.tensorflow
-tensorflow
+tensorflow==2.4.4
```
dbczumar, line 17: MLflow Models supports the latest versions of TensorFlow (2.7.0). I don't think we should pin to 2.4.4.

dev/extra-ml-requirements.txt
```diff
@@ -8,14 +8,13 @@ keras
 # Required by mlflow.sklearn
 scikit-learn
 # Required by mlflow.gluon
-# TODO: Unpin once https://github.com/apache/incubator-mxnet/issues/20068 is fixed
-mxnet!=1.8.0
+mxnet==1.8.0.post0
```
dbczumar, line 11: Can we unpin mxnet rather than pinning to `1.8.0.post0`?

CONTRIBUTING.rst
```diff
@@ -147,20 +147,17 @@ If you plan on doing development and testing, you will also need to install the
 
 .. code-block:: bash
 
-    pip install -r dev-requirements.txt
-    pip install -r test-requirements.txt
+    pip install pip-tools
+    pip-compile --output-file compiled-requirements.txt dev-requirements.txt test-requirements.txt
```
dbczumar, line 151: Is `pip-compile` strictly necessary for correctly installing dev and test requirements?

maitre-matt: We can likely do more user friendly. `pip-compile` is needed to avoid `pip` entering a near-infinite loop due to its package-conflict resolution logic. `pip-compile` does not need to be run by new users though. The latest guidance I am seeing from PyPI is to commit the compiled requirements file [[PyCon2021](https://www.youtube.com/watch?v=VWWgkF-0cDQ)][[pip-tools FAQ](https://github.com/jazzband/pip-tools#should-i-commit-requirementsin-and-requirementstxt-to-source-control)]. That ensures reproducible dev environments, reproducing builds, protects against package tampering, and removes the need for new users to use `pip-compile`. It does require existing committers to run `pip-compile` when updating the repo requirements, so it changes their workflow a bit. I could switch to committing the compiled requirements if this is fine with other committers.

Another improvement for new users could be to trim down the number of requirements to install upon first use. Version conflicts seem to primarily come from `extra-ml-requirements.txt`, which is likely not critical for a first commit doing a touch up in the code, unlike `lint-requirements.txt` which looks required to ensure success of a first PR.

dbczumar: @maitre-matt Using `pip-compile` sounds good - thank you for the explanation. I also agree that taking `extra-ml-requirements` off the critical path for developer environment setup is a good idea.

dbczumar: `extra-ml-requirements` is now a somewhat-outdated set of dependency recommendations for certain libraries as well. https://github.com/mlflow/mlflow/blob/master/mlflow/ml-package-versions.yml is the source of truth for our CI systems.

harupy:
@maitre-matt 

> The change addresses the near-infinite pip backtracking

Is it possible to reproduce this issue?

mmaitre314:
> @maitre-matt
> 
> > The change addresses the near-infinite pip backtracking
> 
> Is it possible to reproduce this (e.g. using `Dockerfile`)?

It's fairly easy to repro using an empty environment: see https://github.com/mlflow/mlflow/issues/4987 for details. Pip's backtracking downloads various package versions to resolve requirement conflicts, and packages like PyTorch/Tensorflow/etc. are really big. @mhemendra stopped at 40GB of downloads. I stopped much earlier and switched to pip-compile which performs conflict resolution in a less network-intensive manner.

dbczumar:
> > @maitre-matt
> > > The change addresses the near-infinite pip backtracking
> > 
> > 
> > Is it possible to reproduce this (e.g. using `Dockerfile`)?
> 
> It's fairly easy to repro using an empty environment: see #4987 for details. Pip's backtracking downloads various package versions to resolve requirement conflicts, and packages like PyTorch/Tensorflow/etc. are really big. @mhemendra stopped at 40GB of downloads. I stopped much earlier and switched to pip-compile which performs conflict resolution in a less network-intensive manner.

Sounds good! Let's use pip compile!

dbczumar:
@maitre-matt pip-compile takes quite a long time to run, as observed by @harupy. Can we adjust the developer guidelines to recommend first trying the existing installation approach and then position `pip-compile` as an alternative if users run into issues?

maitre-matt:
> 
> @maitre-matt pip-compile takes quite a long time to run, as observed by @harupy. Can we adjust the developer guidelines to recommend first trying the existing installation approach and then position `pip-compile` as an alternative if users run into issues?

I am thinking the output of pip-compile could be checked in as some `requirements.txt` file so that new developers do not have to deal with pip-compile (i.e. keep getting-started guidelines unchanged) , and pip install also becomes much faster (no version ranges to deal with). I need to do some more work to address your feedback about not downgrading package versions. Once I have a new set of compatible versions I'll push a new change proposal to discuss.

harupy:
> I am thinking the output of pip-compile could be checked in as some requirements.txt file so that new developers do not have to deal with pip-compile (i.e. keep getting-started guidelines unchanged).

Does the result of `pip-compile` depend on OS and python version?

dbczumar:
> > @maitre-matt pip-compile takes quite a long time to run, as observed by @harupy. Can we adjust the developer guidelines to recommend first trying the existing installation approach and then position `pip-compile` as an alternative if users run into issues?
> 
> I am thinking the output of pip-compile could be checked in as some `requirements.txt` file so that new developers do not have to deal with pip-compile (i.e. keep getting-started guidelines unchanged) , and pip install also becomes much faster (no version ranges to deal with). I need to do some more work to address your feedback about not downgrading package versions. Once I have a new set of compatible versions I'll push a new change proposal to discuss.

Got it. Whichever solution we adopt shouldn't involve strict pinning of packages in our `requirements` files, since the libraries we integrate with are constantly evolving & being updated.

maitre-matt:
> > I am thinking the output of pip-compile could be checked in as some requirements.txt file so that new developers do not have to deal with pip-compile (i.e. keep getting-started guidelines unchanged).
> 
> Does the result of `pip-compile` depend on OS and python version? I guess it does.

It may: that depends on whether some packages have dependencies that differ based on OS and Python version. For instance `mxnet==1.8.0.post0` was not published for Windows but `mxnet==1.7.0.post2` was, so package version resolution will differ based on OS.

An alternative solution could be to not require new developers to install `extra-ml-requirements.txt`. The problem with the current setup is that there are too many packages bringing too many version requirements that cannot be jointly resolved.

maitre-matt:
I just did a clean test from an empty conda environment on the latest master and it looks like the issue does not repro anymore. `pip install -r requirements/test-requirements.txt` went through and resolved all the package versions (on WSL2 Ubuntu with Python 3.6). Things are looking good without this PR, so discarding it.

PR #5069, opened 2021-11-15
Implement model evaluation API (part 1)

WeichenXu123:

Implement basic APIs:
class EvaluationDataset/EvaluationResult/EvaluationMetrics/EvaluationArtifact, evaluate API, class ModelEvaluator (base evaluator class) and plugin interface

## How is this patch tested?

Unit tests.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mlflow/evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    Represents an input dataset for model evaluation. This is intended for
+    use with the `mlflow.evaluate()`API.
+    """
+
+    def __init__(self, data, labels=None, name=None):
+        """
+        :param data: One of the following:
+         - A numpy array or list of evaluation features, excluding labels.
+         - A Pandas DataFrame, or the path to a serialized DataFrame,
```
WeichenXu123, line 68: Question:
For Pandas DataFrame case, does it mean we regards all columns as feature columns except the "labels" column ? @dbczumar

mlflow/evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+                   one does not exist.
+    :param evaluators: The name of the evaluator to use for model evaluations, or
+                       a list of evaluator names. If unspecified, all evaluators
+                       capable  of evaluating the specified model on the specified
+                       dataset are used.
+    :param evaluator_config: A dictionary of additional configurations to supply
+                             to the evaluator. If multiple evaluators are
+                             specified, each configuration should be supplied as
+                             a nested dictionary whose key is the evaluator name.
+    :return: An `EvaluationResult` instance containing evaluation results.
```
WeichenXu123, line 192: @dbczumar When we specify a list of evaluators, What shall we return ? In this PR, I return a dict of "evaluator_name" --> evaluation_result

WeichenXu123:
I will add more tests, but the PR is ready for first pass review.

mlflow/evaluation.py
BenWilson2: A lot of DS teams put identification tags and additional information in the "raw input" that goes into an ML pipeline, filtering out the X (effective vector for Pandas ML use cases) columns before submitting to a processing pipeline and retain those identifying columns on batch inference for identification purposes (i.e. a member_id, a sku, a product_id, etc..) Most of these are either primary or surrogate keys and would be problematic to include in any evaluation of the feature set.

dbczumar: I think we should leave it up to the user to filter out features that are unnecessary for model evaluation. @WeichenXu123 let's regard all columns except the `labels` column as features and document this so that users know they need to filter out everything that isn't a feature or a label.

mlflow/evaluation.py
dbczumar: @WeichenXu123 Can we merge the evaluation results into a single object? Otherwise, if there are multiple evaluators registered, users won't know whether the return type is `dict` or `EvaluationResult` when `evaluators=None`. Returning `EvaluationResult` in both cases is better than always returning a dictionary because, usually, only the default evaluator will be used, so we shouldn't have to make users unpack a `dict`.

mlflow/evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        raise NotImplementedError()
+
+    @property
+    def artifacts(self) -> Dict[str, EvaluationArtifact]:
+        """
+        A dictionary mapping standardized artifact names (e.g. "roc_data") to
+        artifact content and location information
+        """
+        raise NotImplementedError()
```
dbczumar, line 55: Can we provide a default implementation for each of these methods and a default constructor? This way, plugin implementers don't have to redefine `EvaluationResult` themselves unless they want to add new methods.

mlflow/evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+        The content of the artifact (representation varies)
+        """
+        raise NotImplementedError()
+
+    @property
+    def location(self) -> str:
+        """
+        The location of the artifact
+        """
+        raise NotImplementedError()
```
dbczumar, line 26: Can we provide a default implementation for each of these methods too?

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+                         arguments that may be added to the evaluation interface
+                         in the future.
+        :return: An `EvaluationResult` instance containing evaluation results.
+        """
+        client = mlflow.tracking.MlflowClient()
+        with GetOrCreateRunId(run_id) as run_id:
+            metrics_dict = self.compute_metrics(predict, dataset)
+            timestamp = int(time.time() * 1000)
+            dataset_id = dataset.name if dataset.name is not None else dataset.hash
+            # TODO: log tags of dataset metadata
```
WeichenXu123, line 240: In design doc, the tags format is a list of dataset metadata
But implementing it make performance bad. When we append a new dataset metadata into the tag, we need first read tag `mlflow.datasets`, then append the new metadata into the tag string, and then overwrite the tag.

shall we change the design to log tag key like `mlflow.datasets.{name/hash}` and with each dataset metadata value ?

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+            client.log_batch(
+                run_id,
+                metrics=[
+                    Metric(key=f'{key}_on_{dataset_id}', value=value, timestamp=timestamp, step=0)
+                    for key, value in metrics_dict
+                ],
+            )
+            artifact_dict = self.compute_and_log_artifact(predict, dataset, run_id, client)
```
WeichenXu123, line 248: When log artifact, the artifact path shall we also include the dataset name/hash ?

mlflow/evaluation.py
dbczumar: @WeichenXu123 We need to make sure that users can search for runs based on dataset names, hashes, and paths. We don't have partial search capabilities for tag keys, so it might be tricky to make this work. Let me know if you have thoughts about supporting search here.

In general, we don't expect that users will call `mlflow.evaluate()` frequently enough for the performance to matter. Concurrency is a concern, but the likelihood of a race condition seems low; we can handle it with locking later.

mlflow/evaluation.py
dbczumar: we should definitely include the dataset name. I don't think we need to include the hash.

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+from mlflow.utils.file_utils import TempDir
+from mlflow.entities import Metric
+from mlflow.tracking.artifact_utils import get_artifact_uri, _download_artifact_from_uri
+
+
+class EvaluationMetrics(dict):
+    pass
+
+
+class EvaluationArtifacts:
```
dbczumar, line 20:
```suggestion
class EvaluationArtifact:
```

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def __init__(self, location, content=None):
+        self._content = content
+        self._location = location
+
+    def load_content_from_file(self, local_artifact_file):
```
dbczumar, line 26: Awesome! For the initial implementation, I'm assuming we'll create subclasses of `EvaluationArtifact` for images, CSVs, etc?

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+        with open(path, 'r') as f:
+            obj = pickle.load(f)
+        return obj
+
+    def save(self, path):
+        """Write the evaluation results to the specified local filesystem path"""
+        # We will likely avoid serializing artifacts themselves, just locations.
+        # Deserialization will resolve locations to artifact contents.
+        with open(path, 'w') as f:
+            pickle.dump(self, f)
```
dbczumar, line 77: If artifact content has already been loaded, I don't think we want to pickle it. Ideally, it would be nice if we could save the evaluation result as a JSON-formatted file rather than relying on pickle.

dbczumar: To assist with the process, let's have each `EvaluationArtifact` implement its own `save()` method that we can call.

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+            EvaluationDataset._gen_md5_for_arraylike_obj(md5_gen, self.data)
+            md5_gen.update(self.labels.encode('UTF-8'))
+        return md5_gen.digest()
+
+    @property
+    def metadata(self):
+        return {
+            'name': self.name,
+            'hash': self.hash,
+            'path': self.path,
```
dbczumar, line 155: If path is unspecified (`None`), can we exclude it? Same with `name`

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+         - A numpy array or list of evaluation labels, if `data` is also a numpy array or list.
+         - The string name of a column from `data` that contains evaluation labels, if `data`
+           is a DataFrame.
+
+        :param name: (Optional) The name of the dataset (must not contain ").
+
+        :param path: (Optional) the path to a serialized DataFrame
+          (e.g. a delta table, parquet file)
+        """
+        self.name = name
```
dbczumar, line 121: if `name` is unspecified, can we set  the `name` to the `hash`?

mlflow/evaluation.py
WeichenXu123: Yes.

mlflow/evaluation.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        The location of the artifact
+        """
+        return self._location
+
+    def __getstate__(self, state):
+        state = state.__dict__.copy()
+        # skip pickling artifact content
+        del state['_content']
+        return state
```
WeichenXu123, line 53: Skip pickling `_content` attribute.

mlflow/evaluation.py
WeichenXu123: what do you think of the approach skip pickling _content in here? (so we can still pickle the whole EvaluationResult instance
https://github.com/mlflow/mlflow/pull/5069/files#r751814783

dbczumar: I think it would be nice to save all of the content as files. That way, users can easily inspect the files outside of Python, share them across Python versions etc.

mlflow/evaluation.py
```diff
@@ -0,0 +1,415 @@
...
+        return an tuple of:
+         - an instance of EvaluationMetrics
+         - a dict of artifact_name -> instance_of_EvaluationArtifact
+        and log artifacts into run specified by run_id
+        """
+        raise NotImplementedError()
+
+    def evaluate(
+            self,
+            model: PyFuncModel,
```
WeichenXu123, line 254: I propose to change `predict` argument to be `model` argument and force it to be `PyFuncModel` type, reason:
* force to containerize the model
* Currently `PyFuncModel` only support `predict`, if user want to support `predict_proba`, he can define custom PyFuncModel which contains `predict_proba` methods, and in concrete Evaluator implementation, we will check `hasattr(model, 'predict_proba')` to decide whether to output probability related metrics and artifacts.

And another proposal:
Shall we make `model_type` to be an attribute of `PyFuncModel` ?

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+        """
+        The location of the artifact
+        """
+        return self._location
+
+    def __getstate__(self, state):
+        state = state.__dict__.copy()
+        # skip pickling artifact content
+        del state["_content"]
+        return state
```
dbczumar, line 61: Now that we don't intend to pickle evaluation results, can we remove this?

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+
+    @property
+    def hash(self):
+        """
+        Compute a hash from the specified dataset by selecting the first 5 records, last 5 records,
+        dataset size and feeding them through a cheap, low-collision hash function
+        """
+        if self._hash is not None:
+            return self._hash
+        else:
```
dbczumar, line 185: Can we assign `self._hash` to the hash when we compute it?

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+            return md5_gen.hexdigest()
+
+    @property
+    def metadata(self):
+        metadata = {
+            "hash": self.hash,
+            "path": self.path,
+        }
+        if self.user_specified_name is not None:
+            metadata["name"] = self.user_specified_name
```
dbczumar, line 202: We should always include `name` in the metadata, since it always has a value (either `self.user_specified_name` or `self._hash`). We should only include the path if it's not `None`.

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+         - A numpy array or list of evaluation labels, if `data` is also a numpy array or list.
+         - The string name of a column from `data` that contains evaluation labels, if `data`
+           is a DataFrame.
+
+        :param name: (Optional) The name of the dataset (must not contain ").
+
+        :param path: (Optional) the path to a serialized DataFrame
+          (e.g. a delta table, parquet file)
+        """
+        self.user_specified_name = name
```
dbczumar, line 149:
```suggestion
        self._user_specified_name = name
```

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+        self.data = data
+        self.labels = labels
+        self.path = path
+        self._hash = None
+
+    def extract_features_and_labels(self):
```
dbczumar, line 155:
```suggestion
    def _extract_features_and_labels(self):
```

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+        if isinstance(self.data, np.ndarray):
+            return self.data, self.labels
+        elif isinstance(self.data, pd.DataFrame):
+            feature_cols = [x for x in self.data.columns if x != self.labels]
+            return self.data[feature_cols], self.data[self.labels]
+        else:
+            raise ValueError(f'Unsupported data type: {type(self.data)}')
```
dbczumar, line 162: can we add support for Spark DataFrames with <= 10,000 rows? We can check the size of the dataframe and, if it's < 10,000 rows, call `toPandas()`.

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+         - A Pandas DataFrame, or the path to a serialized DataFrame,
+           containing evaluation features and labels. All columns will be regarded as feature
+           columns except the "labels" column.
+
+        :param labels: One of the following:
+         - A numpy array or list of evaluation labels, if `data` is also a numpy array or list.
+         - The string name of a column from `data` that contains evaluation labels, if `data`
+           is a DataFrame.
+
+        :param name: (Optional) The name of the dataset (must not contain ").
```
dbczumar, line 144: Can we add validation to ensure that the `name` and `path` do not contain `"`?

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+
+        :param path: (Optional) the path to a serialized DataFrame
```
dbczumar, line 146: The path must also not contain `"`.

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+            raise ValueError(f'Unsupported data type: {type(self.data)}')
+
+    @staticmethod
+    def _gen_md5_for_arraylike_obj(md5_gen, data):
+        md5_gen.update(pickle.dumps(len(data)))
+        if len(data) < EvaluationDataset.NUM_SAMPLE_ROWS_FOR_HASH * 2:
+            md5_gen.update(pickle.dumps(data))
+        else:
+            md5_gen.update(pickle.dumps(data[: EvaluationDataset.NUM_SAMPLE_ROWS_FOR_HASH]))
+            md5_gen.update(pickle.dumps(data[-EvaluationDataset.NUM_SAMPLE_ROWS_FOR_HASH :]))
```
dbczumar, line 171: If we use pickle, this hash will change across Python versions, right? It could also change depending on other aspects of the current environment, e.g. the point in the callstack where `mlflow.evaluate()` is called. Could we instead use something like https://pandas.pydata.org/docs/reference/api/pandas.util.hash_pandas_object.html for pandas dataframes and https://stackoverflow.com/a/5386770/11952869 for numpy arrays? Feel free to do your own research here.

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+            if isinstance(self.data, np.ndarray):
+                EvaluationDataset._gen_md5_for_arraylike_obj(md5_gen, self.data)
+                EvaluationDataset._gen_md5_for_arraylike_obj(md5_gen, self.labels)
+            elif isinstance(self.data, pd.DataFrame):
+                EvaluationDataset._gen_md5_for_arraylike_obj(md5_gen, self.data)
+                md5_gen.update(self.labels.encode("UTF-8"))
+            return md5_gen.hexdigest()
+
+    @property
+    def metadata(self):
```
dbczumar, line 196:
```suggestion
    def _metadata(self):
```

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+    def __enter__(self):
+        if self.run_id is not None:
+            return self.run_id
+        else:
+            self.managed_run_context = mlflow.start_run()
+            return self.managed_run_context.__enter__().info.run_id
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.managed_run_context is not None:
+            return self.managed_run_context.__exit__(exc_type, exc_val, exc_tb)
```
dbczumar, line 232: Instead of `GetOrCreateRunId`, could we just use 

`with mlflow.start_run(run_id=run_id)`? From the docs, if `run_id` is `not None`, `start_run()` will activate and resume the specified run. See https://github.com/mlflow/mlflow/blob/9faf3b04c9f6b3766a64d7433dd3e85bcdf27fff/mlflow/tracking/fluent.py#L174-L177.

mlflow/evaluation.py
dbczumar: @WeichenXu123 Changing `predict()` to `model` with type `PyFuncModel` sounds good to me! 

For model_type, how would we find the right type to add as an attribute to `PyFuncModel`? Users would need some place to specify this when saving the model, right?

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+                    Metric(key=f"{key}_on_{dataset.name}", value=value, timestamp=timestamp, step=0)
+                    for key, value in metrics_dict.items()
+                ],
+                tags=[RunTag("mlflow.datasets", dataset_metadata_str)],
+            )
+
+            return EvaluationResult(metrics_dict, artifacts_dict)
+
+
+class ModelEvaluatorRegistry:
```
dbczumar, line 324: Can we add this to a separate file called `evaluator_registry.py` in a common directory? E.g. we could have `mlflow/tracking/evaluation` as a directory with `__init__.py` containing the evaluator interface and `evaluator_registry.py` containing the evaluator registry.

I think putting files in `tracking` makes sense since this will be an MLflow Tracking API.

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+def evaluate(
+        model: Union[str, PyFuncModel],
+        model_type, dataset,
+        run_id=None,
+        evaluators=None,
+        evaluator_config=None
+) -> Union[EvaluationResult, Dict[str, EvaluationResult]]:
+    """
```
dbczumar, line 375: In addition to the parameter docstrings, we'll want to explain the overall purpose and behavior of `mlflow.evaluate()`. We can do this once the rest of the PR is ready to merge, just leaving this here as a note.

tests/protos/test_message_pb2.py
```diff
@@ -2,544 +2,932 @@
 # source: test_message.proto
```
dbczumar, line 2: We don't need these changes, right?

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+                       capable  of evaluating the specified model on the specified
+                       dataset are used.
+    :param evaluator_config: A dictionary of additional configurations to supply
+                             to the evaluator. If multiple evaluators are
+                             specified, each configuration should be supplied as
+                             a nested dictionary whose key is the evaluator name.
+    :return: An `EvaluationResult` instance containing evaluation results.
+    """
+    if evaluators is None:
+        evaluators = "default_evaluator"
```
dbczumar, line 398: Can we use `default` for the default evaluator instead?

```suggestion
        evaluators = "default"
```

mlflow/evaluation.py
```diff
@@ -0,0 +1,424 @@
...
+                    labels (optional) for model evaluation.
+    :param run_id: The ID of the MLflow Run to which to log results. If
+                   unspecified, behavior depends on the specified `evaluator`.
+                   When `run_id` is unspecified, the default evaluator logs
+                   results to the current active run, creating a new active run if
+                   one does not exist.
+    :param evaluators: The name of the evaluator to use for model evaluations, or
+                       a list of evaluator names. If unspecified, all evaluators
+                       capable  of evaluating the specified model on the specified
+                       dataset are used.
```
dbczumar, line 390: Can we indicate that the default evaluator can be referred to by `default`?

dbczumar (commented):
@WeichenXu123 This is looking great! Awesome stuff!

tests/protos/test_message_pb2.py
harupy: @WeichenXu123 Looks like you applied black formatting on this file.

mlflow/evaluation.py
WeichenXu123: There's a risk that, computing spark dataframe size will trigger spark jobs, which may took long time.

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,375 @@
...
+        self._content = content
+        self._location = location
+
+    @classmethod
+    def load_content_from_file(cls, local_artifact_path):
+        raise NotImplementedError()
+
+    @classmethod
+    def save_content_to_file(cls, content, output_artifact_path):
+        raise NotImplementedError()
```
dbczumar, line 34:
```suggestion
    @classmethod
    def _load_content_from_file(cls, local_artifact_path):
        raise NotImplementedError()

    @classmethod
    def _save_content_to_file(cls, content, output_artifact_path):
        raise NotImplementedError()
```

dbczumar: Can we make these instance methods in case developers want to reference other instance attributes when deciding how to save content?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+class EvaluationMetrics(dict):
+    pass
+
+
+class EvaluationArtifact:
+    def __init__(self, location, content=None):
+        self._content = content
+        self._location = location
```
dbczumar, line 26: Can we write docstrings for these classes?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,375 @@
...
+                "location": artifact.location,
+                "class_name": _get_fully_qualified_class_name(artifact),
+            }
+            for artifact_name, artifact in self.artifacts.items()
+        }
+        with open(os.path.join(path, "artifacts_metadata.json"), "w") as fp:
+            json.dump(artifacts_metadata, fp)
+
+        for artifact_name, artifact in self.artifacts.items():
+            artifact.save_content_to_file(artifact.content, os.path.join(path, artifact_name))
```
dbczumar, line 98: Can we place these in an `artifacts` subdirectory?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,375 @@
...
+
+        :param name: (Optional) The name of the dataset (must not contain ").
+
+        :param path: (Optional) the path to a serialized DataFrame (must not contain ").
+          (e.g. a delta table, parquet file)
+        """
+        if name is not None and '"' in name:
+            raise ValueError(f'Dataset name cannot include " but get name {name}')
+        if path is not None and '"' in path:
+            raise ValueError(f'Dataset path cannot include " but get name {path}')
```
dbczumar, line 145:
```suggestion
        if name is not None and '"' in name:
            raise ValueError(f'Dataset name cannot include " but got name {name}')
        if path is not None and '"' in path:
            raise ValueError(f'Dataset path cannot include " but got name {path}')
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,375 @@
...
+            raise ValueError(f'Unsupported data type: {type(self.data)}')
+
+    @staticmethod
+    def _gen_md5_for_arraylike_obj(md5_gen, data):
+        md5_gen.update(pickle.dumps(len(data)))
+        if len(data) < EvaluationDataset.NUM_SAMPLE_ROWS_FOR_HASH * 2:
+            md5_gen.update(pickle.dumps(data))
+        else:
+            md5_gen.update(pickle.dumps(data[: EvaluationDataset.NUM_SAMPLE_ROWS_FOR_HASH]))
+            md5_gen.update(pickle.dumps(data[-EvaluationDataset.NUM_SAMPLE_ROWS_FOR_HASH :]))
```
dbczumar, line 185: Reminder to address https://github.com/mlflow/mlflow/pull/5069#discussion_r753861133

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+def evaluate(
+        model: Union[str, PyFuncModel],
+        model_type, dataset,
+        run_id=None,
+        evaluators=None,
+        evaluator_config=None
+) -> EvaluationResult:
+    """
```
dbczumar, line 325: Reminder to add a docstring explaining the overall behavior of this method and use cases.

mlflow/models/evaluation/base.py
WeichenXu123: change `_save_content_to_file` to be instance method. But I prefer `_load_content_from_file` to be class method, then we loading EvaluationResults we can load content via class, which is easier to use

mlflow/evaluation.py
WeichenXu123: The function https://pandas.pydata.org/docs/reference/api/pandas.util.hash_pandas_object.html will return a pandas Series object which contains a list of uint values, which is not convenient. I rethink on this, I suggest using this:
```
    def _array_like_obj_to_bytes(data):
        if isinstance(data, pd.DataFrame):
            return data.to_numpy().tobytes()
        elif isinstance(data, np.ndarray):
            return data.tobytes()
        elif isinstance(data, list):
            return np.array(data).tobytes()
```
This helper method convert pandas dataframe/numpy array/list into a deterministic bytes array, and it is independent with python version/platform.

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+        elif isinstance(data, np.ndarray):
+            return data.tobytes()
+        elif isinstance(data, list):
+            return np.array(data).tobytes()
+        else:
+            raise ValueError('Unsupported data type.')
+
+    @staticmethod
+    def _gen_md5_for_arraylike_obj(md5_gen, data):
+        md5_gen.update(pickle.dumps(len(data)))
```
dbczumar, line 197: Do we need to call `pickle.dumps()` on `len(data)`, or can we just pass in the len since it's an int?

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+        metadata = {
+            "name": self.name,
+            "hash": self.hash,
+        }
+        if self.path is not None:
+            metadata["path"] = self.path
+        return metadata
+
+
+class ModelEvaluator:
```
dbczumar, line 244: Can we place this object in a separate file, e.g. `evaluator.py`?

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+def evaluate(
+        model: Union[str, PyFuncModel],
+        model_type, dataset,
+        run_id=None,
+        evaluators=None,
+        evaluator_config=None
+) -> EvaluationResult:
+    """
```
dbczumar, line 346: Reminder to add a docstring explaining the overall behavior of this method and use cases.

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+
+    eval_results = []
+    for evaluator_name in evaluators:
+        config = evaluator_config[evaluator_name]
+        try:
+            evaluator = _model_evaluation_registry.get_evaluator(evaluator_name)
+        except MlflowException:
+            continue
+
+        if evaluator.can_evaluate(model_type, config):
```
dbczumar, line 390: Can we add a `logger.info()` call before each evaluator.evaluate() call? E.g. `"Evaluating the model with the 'default' evaluator"`

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+            return EvaluationResult(metrics_dict, artifacts_dict)
+
+        if mlflow.active_run() is not None:
+            return do_evaluate(mlflow.active_run().info.run_id)
+        else:
+            with mlflow.start_run(run_id=run_id) as run:
+                return do_evaluate(run.info.run_id)
+
+
+def evaluate(
```
dbczumar, line 339: Can we also introduce a `list_evaluators()` API to list all of the available evaluators?

dbczumar: After we implement the default evaluator, can we also import this `evaluate` API in https://github.com/mlflow/mlflow/blob/master/mlflow/__init__.py and  https://github.com/mlflow/mlflow/blob/master/mlflow/models/__init__.py so that users can call this via `mlflow.evaluate()` and `mlflow.models.evaluate()`?

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+                tags=[RunTag("mlflow.datasets", dataset_metadata_str)],
+            )
+
+            return EvaluationResult(metrics_dict, artifacts_dict)
+
+        if mlflow.active_run() is not None:
+            return do_evaluate(mlflow.active_run().info.run_id)
+        else:
+            with mlflow.start_run(run_id=run_id) as run:
+                return do_evaluate(run.info.run_id)
```
dbczumar, line 336: Can we put this logic inside `mlflow.models.evaluation.evaluate()` and pass the run ID to `ModelEvaluator.evaluate()`? This way, developers don't have to manage runs if they want to override `ModelEvaluator.evaluate()`.

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+        and log artifacts into run specified by run_id
+        """
+        raise NotImplementedError()
+
+    def evaluate(
+            self,
+            model: PyFuncModel,
+            model_type,
+            dataset,
+            run_id=None,
```
dbczumar, line 275: From https://github.com/mlflow/mlflow/pull/5069/files#r755590594, `run_id` should never be `None`.

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+                    metadata["hash"] == dataset.hash
+                    and metadata["name"] == dataset._user_specified_name
+                ):
+                    metadata_exists = True
+                    break
+
+            if not metadata_exists:
+                dataset_metadata_list.append(dataset._metadata)
+
+            dataset_metadata_str = json.dumps(dataset_metadata_list)
```
dbczumar, line 316: Can we compute and set the dataset metadata str in `mlflow.models.evaluation.evaluate()` so that we don't set it multiple times if multiple evaluators are used?

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+                                 the evaluator.
+        :param **kwargs: For forwards compatibility, a placeholder for additional
+                         arguments that may be added to the evaluation interface
+                         in the future.
+        :return: True if the evaluator can evaluate the specified model on the
+                 specified dataset. False otherwise.
+        """
+        raise NotImplementedError()
+
+    def compute_metrics_and_compute_and_log_artifacts(
```
dbczumar, line 259: Once https://github.com/mlflow/mlflow/pull/5069/files#r755590594 and https://github.com/mlflow/mlflow/pull/5069/files#r755593278 are addressed, I don't think we need this method. Once `mlflow.models.evaluation.evaluate()` initializes the run and sets the `mlflow.datasets` tag *before* calling `ModelEvaluator.evaluate()`, the `ModelEvaluator.evaluate()` method just has to compute and log metrics, params, etc. The developer can implement this in `ModelEvaluator.evaluate()` directly.

mlflow/models/evaluation/__init__.py
jinzhang21: Actually I'd recommend to move all base classes to their own class.py file to keep __ init __.py clean:)

mlflow/models/evaluation/__init__.py
jinzhang21: +1

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+                                 the evaluator.
+        :param **kwargs: For forwards compatibility, a placeholder for additional
+                         arguments that may be added to the evaluation interface
+                         in the future.
+        :return: An `EvaluationResult` instance containing evaluation results.
+        """
+        client = mlflow.tracking.MlflowClient()
+        self.mlflow_client = client
+
+        def do_evaluate(_run_id):
```
jinzhang21, line 296: How about just call it _evaluate()?

mlflow/models/evaluation/evaluator_registry.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self):
+        self._registry = {}
+
+    def register(self, scheme, evaluator):
+        """Register model evaluator provided by other packages"""
+        self._registry[scheme] = evaluator
+
+    def register_entrypoints(self):
+        # Register artifact repositories provided by other packages
```
jinzhang21, line 20: Update comment.

mlflow/models/evaluation/__init__.py
jinzhang21: +1.

jinzhang21 (commented):
Very nice implementation, Weichen! LGTM!

jinzhang21 (approved):

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+         - The string name of a column from `data` that contains evaluation labels, if `data`
+           is a DataFrame.
+
+        :param name: (Optional) The name of the dataset (must not contain ").
+
+        :param path: (Optional) the path to a serialized DataFrame (must not contain ").
+          (e.g. a delta table, parquet file)
+        """
+        if name is not None and '"' in name:
+            raise ValueError(f'Dataset name cannot include " but got name {name}')
```
harupy, line 149:
```suggestion
            raise ValueError(f'Dataset name cannot include a double quote (") but got name {name}')
```

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+        if path is not None and '"' in path:
+            raise ValueError(f'Dataset path cannot include " but got name {path}')
```
harupy, line 151:
```suggestion
            raise ValueError(f'Dataset path cannot include " but got path {path}')
```

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+        self.data = data
+        self.labels = labels
+        self.path = path
+        self._hash = None
+
+    def _extract_features_and_labels(self):
+        if isinstance(self.data, np.ndarray):
+            return self.data, self.labels
+        elif isinstance(self.data, pd.DataFrame):
+            feature_cols = [x for x in self.data.columns if x != self.labels]
```
harupy, line 179: Does this line work correctly when `self.labels` is an array?

mlflow/models/evaluation/__init__.py
harupy: +1

mlflow/models/evaluation/__init__.py
```diff
@@ -0,0 +1,399 @@
...
+    use with the `mlflow.evaluate()`API.
+    """
+
+    NUM_SAMPLE_ROWS_FOR_HASH = 5
+
+    def __init__(self, data, labels, name=None, path=None):
+        """
+        :param data: One of the following:
+         - A numpy array or list of evaluation features, excluding labels.
+         - A Pandas DataFrame, or the path to a serialized DataFrame,
```
dbczumar, line 134: Can we add support for Spark DataFrames? The default evaluator should call `limit(10000)` and print a warning indicating that only the first 10000 rows are examined.

mlflow/models/evaluation/__init__.py
WeichenXu123: in this case `isinstance(self.data, pd.DataFrame)==True` the labels must be a string, having verification in constructor.

mlflow/models/evaluation/__init__.py
WeichenXu123: changed to `md5_gen.update(np.int64(len(data)).tobytes())` to avoid calling pickle here

mlflow/models/evaluation/__init__.py
WeichenXu123: I refactor the logic, add a context manager with verification:
```
class StartRunOrReuseActiveRun:

    def __init__(self, run_id):
        self.user_specified_run_id = run_id
        self.managed_run = None

    def __enter__(self):
        if mlflow.active_run() is not None:
            active_run_id = mlflow.active_run().info.run_id
            if self.user_specified_run_id is not None and \
                    self.user_specified_run_id != active_run_id:
                raise ValueError("An active run exists, you cannot specify another run_id when "
                                 "evaluating.")
            return active_run_id
        else:
            if self.user_specified_run_id is None:
                raise ValueError("Active run does not exist, you need specify a run_id when "
                                 "evaluating.")
            self.managed_run = \
                mlflow.start_run(run_id=self.user_specified_run_id).__enter__()
            return self.user_specified_run_id

    def __exit__(self, exc_type, exc_val, exc_tb):
        if self.managed_run is not None:
            return self.managed_run.__exit__(exc_type, exc_val, exc_tb)
```

mlflow/evaluation.py
```diff
@@ -0,0 +1,410 @@
...
+    pass
+
+
+class EvaluationArtifact:
+    def __init__(self, location, content=None):
+        self._content = content
+        self._location = location
+
+    @classmethod
+    def load_content_from_file(cls, local_artifact_path):
```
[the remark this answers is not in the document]
WeichenXu123: we can support load by downloading from uri in future.

mlflow/evaluation.py
```diff
@@ -0,0 +1,410 @@
...
+    model, model_type, dataset, run_id=None, evaluators=None, evaluator_config=None
+) -> Union[EvaluationResult, Dict[str, EvaluationResult]]:
+    """
+    :param model: A model supported by the specified `evaluator`, or a URI
+                  referring to such a model. The default evaluator supports the
+                  following:
+
+                  - A pyfunc model instance (an instance of class `PyFuncModel`)
+
+    :param model_type: A string describing the model type. The default evaluator
```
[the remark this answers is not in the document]
WeichenXu123: We can do this in future, but not now.

mlflow/evaluation.py
```diff
@@ -0,0 +1,410 @@
...
+                   results to the current active run, creating a new active run if
+                   one does not exist.
+    :param evaluators: The name of the evaluator to use for model evaluations, or
+                       a list of evaluator names. If unspecified, all evaluators
+                       capable  of evaluating the specified model on the specified
+                       dataset are used.
+    :param evaluator_config: A dictionary of additional configurations to supply
+                             to the evaluator. If multiple evaluators are
+                             specified, each configuration should be supplied as
+                             a nested dictionary whose key is the evaluator name.
```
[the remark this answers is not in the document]
WeichenXu123: Could you rephrase your description here ? @jinzhang21

mlflow/utils/__init__.py
```diff
@@ -172,3 +173,11 @@ def _inspect_original_var_name(var, fallback_name):
...
     except Exception:
         return fallback_name
+
+
+def load_class(kls):
+    parts = kls.split(".")
+    module = ".".join(parts[:-1])
+    m = importlib.import_module(module)
+    name = parts[-1]
+    return getattr(m, name)
```
harupy, line 183: https://github.com/mlflow/mlflow/pull/5092#discussion_r755647142

WeichenXu123:
@dbczumar @harupy 
Shall we export `evaluate` API as `mflow.evaluate` ?
This will cause importing mlflow trigger importing mlflow.models, and it will add many dependencies into "mlflow-skinny"

WeichenXu123:
> We should also make sure that users can inspect the mlflow.evaluate() method without triggering / requiring additional imports; if we need to move imports inside the evaluate() function, then let's do that.

"if we need to move imports inside the evaluate() function, then let's do that." I think this cannot address it, we need:
 (1) make other imports inside `mlflow.models` to be lazy import.
Or 
(2) move evaluate/EvaluationResults/EvaluataionDataset implementation out of `mlflow.models`.

WeichenXu123:
@dbczumar When importing mlflow.models, it will import numpy, pandas, etc, which are not in `SKINNY_REQUIREMENTS` but in `CORE_REQUIREMENTS` .

WeichenXu123:
> Users should be able to import mlflow.models in mlflow-skinny without triggering any additional imports

This looks hard. It have to import numpy and pandas, moving `import numpy/pandas` into functions does not address it, e.g. see code here:
https://github.com/mlflow/mlflow/blob/1955682dc0f762f4ae40cdd81729646eae09bc9d/mlflow/types/schema.py#L37

When importing class, it have to import numpy

@dbczumar

dbczumar:
> 



> > Users should be able to import mlflow.models in mlflow-skinny without triggering any additional imports
> 
> This looks hard. It have to import numpy and pandas, moving `import numpy/pandas` into functions does not address it, e.g. see code here:
> 
> https://github.com/mlflow/mlflow/blob/1955682dc0f762f4ae40cdd81729646eae09bc9d/mlflow/types/schema.py#L37
> 
> When importing class, it have to import numpy
> 
> @dbczumar

@WeichenXu123 Can we try / catch imports of signature, schema, and infer_pip_requirements from `mlflow.models.__init__`?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+    Represent a dict of metrics.
+    """
+
+    pass
+
+
+class EvaluationArtifact:
+    """
+    Represent a artifact. Contains artifact uri and content.
```
dbczumar, line 31:
```suggestion
    A model evaluation artifact containing an artifact uri and content.
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+from mlflow.utils.class_utils import _get_class_from_string
+import logging
+
+
+_logger = logging.getLogger(__name__)
+
+
+class EvaluationMetrics(dict):
+    """
+    Represent a dict of metrics.
```
dbczumar, line 23:
```suggestion
    A dictionary of model evaluation metrics.
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+            artifacts_metadata = json.load(fp)
+
+        artifacts = {}
+
+        artifacts_dir = os.path.join(path, "artifacts")
+
+        for artifact_name, meta in artifacts_metadata:
+            location = meta["location"]
+            ArtifactCls = _get_class_from_string(meta["class_name"])
+            content = ArtifactCls.load_content_from_file(os.path.join(artifacts_dir, artifact_name))
```
dbczumar, line 101: Can we call `ArtifactCls.load()` and pass in `local_artifact_path`, rather than calling the internal method?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+        If `local_artifact_path` is None, download artifact from the artifact uri and load it.
+        otherwise load artifact content from specified path.
+        """
+        if local_artifact_path is None:
+            return self._load_content_from_file(local_artifact_path)
+        else:
+            with TempDir() as temp_dir:
+                local_artifact_file = temp_dir.path("local_artifact")
+                _download_artifact_from_uri(self._uri, local_artifact_file)
+                self._load_content_from_file(local_artifact_file)
```
dbczumar, line 52: Can we assign the output of `load_content_from_file` to `self._content`?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+class EvaluationArtifact:
+    """
+    Represent a artifact. Contains artifact uri and content.
+    """
+
+    def __init__(self, uri, content=None):
+        self._uri = uri
+        self._content = content
+
+    def _load_content_from_file(self, local_artifact_path):
```
dbczumar, line 38: Can we document that this method should return loaded artifact content? In https://github.com/mlflow/mlflow/pull/5069/files#r759662089, we'll assign the loaded content to `self._content`.

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+        """
+        return self._artifacts
+
+
+_cached_mlflow_client = None
+
+
+class EvaluationDataset:
+    """
+    Represents an input dataset for model evaluation. This is intended for
```
dbczumar, line 149:
```suggestion
    An input dataset for model evaluation. This is intended for
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+                )
+        elif isinstance(data, supported_dataframe_types):
+            if not isinstance(labels, str):
+                raise ValueError(
+                    "If data is a Pandas DataFrame or Spark DataFrame, labels must be the "
+                    "string name of a column from `data` that contains evaluation labels"
+                )
+        else:
+            raise ValueError(
+                "The data argument must be a numpy array, a list or a " "Pandas DataFrame."
```
dbczumar, line 200:
```suggestion
                "The data argument must be a numpy array, a list or a Pandas DataFrame."
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+    def _extract_features_and_labels(self):
+        """
+        Extract features data and labels data.
+        For spark dataframe, will only extract the first SPARK_DATAFRAME_LIMIT rows data
+        and emit warning.
+        """
+        if isinstance(self.data, np.ndarray):
+            return self.data, self.labels
+        else:
+            if not isinstance(self.data, pd.DataFrame):
```
dbczumar, line 218: Shall we check if `isinstance(self.data, SparkDataFrame)` instead?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+                data = self.data.limit(EvaluationDataset.SPARK_DATAFRAME_LIMIT).toPandas()
+                _logger.warning(
+                    f"Only the first {EvaluationDataset.SPARK_DATAFRAME_LIMIT} rows in the "
+                    f"spark dataframe are examined."
```
dbczumar, line 222:
```suggestion
                    f"Specified Spark DataFrame is too large for model evaluation. Only the first {EvaluationDataset.SPARK_DATAFRAME_LIMIT} rows will be used."
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+        """
+        Compute a hash from the specified dataset by selecting the first 5 records, last 5 records,
+        dataset size and feeding them through a cheap, low-collision hash function
+        """
+        if self._hash is None:
+            md5_gen = hashlib.md5()
+            if isinstance(self.data, np.ndarray):
+                EvaluationDataset._gen_md5_for_arraylike_obj(md5_gen, self.data)
+                EvaluationDataset._gen_md5_for_arraylike_obj(md5_gen, self.labels)
+            elif isinstance(self.data, pd.DataFrame):
```
dbczumar, line 286: Can we handle the Spark DataFrame hashing here as well?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+            ):
+                raise ValueError(
+                    "An active run exists, you cannot specify another run_id when " "evaluating."
+                )
+            return active_run_id
+        else:
+            if self.user_specified_run_id is None:
+                raise ValueError(
+                    "Active run does not exist, you need specify a run_id when " "evaluating."
+                )
```
dbczumar, line 425: If an active run doesn't exist, let's create one for the user.

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+
+    def __enter__(self):
+        if mlflow.active_run() is not None:
+            active_run_id = mlflow.active_run().info.run_id
+            if (
+                self.user_specified_run_id is not None
+                and self.user_specified_run_id != active_run_id
+            ):
+                raise ValueError(
+                    "An active run exists, you cannot specify another run_id when " "evaluating."
```
dbczumar, line 418:
```suggestion
                    "An active run exists, you cannot specify another run_id when evaluating."
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+    """
+    Return a name list for all available Evaluators.
+    """
+    # import _model_evaluation_registry inside function to avoid circuit importing
+    from mlflow.models.evaluation.evaluator_registry import _model_evaluation_registry
+
+    return list(_model_evaluation_registry._registry.keys())
+
+
+class StartRunOrReuseActiveRun:
```
dbczumar, line 399:
```suggestion
class _StartRunOrReuseActiveRun:
```

tests/models/test_evaluation.py
```diff
@@ -0,0 +1,188 @@
...
+    X, y = get_iris()
+    d1 = EvaluationDataset(data=X, labels=y, name="a1")
+    assert d1.name == "a1"
+    d2 = EvaluationDataset(data=X, labels=y)
+    d2.name == d2.hash
+
+
+def test_dataset_hash(iris_dataset, iris_pandas_df_dataset):
+    assert iris_dataset.hash == "49a04c127e5441e0f27e63a325b5fa69"
+    assert iris_pandas_df_dataset.hash == "d6770fd5fffe651cb95e965854920df9"
```
dbczumar, line 164: Have we verified that this works across Python versions? Also, can we add a test for Spark DataFrames?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+        """
+        client = mlflow.tracking.MlflowClient()
+        timestamp = int(time.time() * 1000)
+        client.log_batch(
+            run_id,
+            metrics=[
+                Metric(key=f"{key}_on_{dataset_name}", value=value, timestamp=timestamp, step=0)
+                for key, value in metrics.items()
+            ],
+        )
```
dbczumar, line 359: I don't think it's necessary to provide this method for now. Implementers can leverage `MlflowClient.log_batch()`.

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+                "evaluator config dict."
+            )
+
+    if isinstance(model, str):
+        model = mlflow.pyfunc.load_model(model)
+    elif isinstance(model, PyFuncModel):
+        pass
+    else:
+        raise ValueError(
+            "The model argument must be a URI str referring to mlflow model or "
```
dbczumar, line 501:
```suggestion
            "The model argument must be a string URI referring to an MLflow model or "
```

mlflow/models/evaluation/base.py
WeichenXu123: Good catch!

tests/models/test_evaluation.py
```diff
@@ -0,0 +1,188 @@
...
+        assert saved_metrics == expected_saved_metrics
+        assert eval_result.metrics == expected_metrics
+
+
+def test_dataset_name():
+    X, y = get_iris()
+    d1 = EvaluationDataset(data=X, labels=y, name="a1")
+    assert d1.name == "a1"
+    d2 = EvaluationDataset(data=X, labels=y)
+    d2.name == d2.hash
```
harupy, line 159:
```suggestion
    assert d2.name == d2.hash
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+            dataset_metadata_list = []
+
+        metadata_exists = False
+        for metadata in dataset_metadata_list:
+            if metadata["hash"] == self.hash and metadata["name"] == self._user_specified_name:
+                metadata_exists = True
+                break
+
+        if not metadata_exists:
+            dataset_metadata_list.append(self._metadata)
```
harupy, line 323:
```suggestion
        for metadata in dataset_metadata_list:
            if metadata["hash"] == self.hash and metadata["name"] == self._user_specified_name:
                break
        else:
            dataset_metadata_list.append(self._metadata)
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+            if not isinstance(self.data, pd.DataFrame):
+                data = self.data.limit(EvaluationDataset.SPARK_DATAFRAME_LIMIT).toPandas()
+                _logger.warning(
+                    f"Only the first {EvaluationDataset.SPARK_DATAFRAME_LIMIT} rows in the "
+                    f"spark dataframe are examined."
+                )
+            else:
+                data = self.data
+            feature_cols = [x for x in data.columns if x != self.labels]
+            return data[feature_cols], data[self.labels]
```
harupy, line 227:
```suggestion
            return data.drop(self.labels, axis=1), data[self.labels]
```

does the same thing.

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+
+    @staticmethod
+    def _array_like_obj_to_bytes(data):
+        """
+        Helper method to convert pandas dataframe/numpy array/list into bytes for
+        MD5 calculation purpose.
+        """
+        if isinstance(data, pd.DataFrame):
+            return data.to_numpy().tobytes() + ",".join(list(data.columns)).encode("UTF-8")
```
harupy, line 236:
```suggestion
            return data.to_numpy().tobytes() + ",".join(data.columns).encode("UTF-8")
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+
+    with StartRunOrReuseActiveRun(run_id) as actual_run_id:
+        client = mlflow.tracking.MlflowClient()
+        dataset._log_dataset_tag(client, actual_run_id)
+
+        eval_results = []
+        for evaluator_name in evaluators:
+            config = evaluator_config.get(evaluator_name)
+            if config is None:
+                config = {}
```
harupy, line 513:
```suggestion
            config = evaluator_config.get(evaluator_name, {})
```

Can we set the default value?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+            run_id,
+            metrics=[
+                Metric(key=f"{key}_on_{dataset_name}", value=value, timestamp=timestamp, step=0)
+                for key, value in metrics.items()
+            ],
+        )
+
+    def evaluate(
+        self,
+        model: "mlflow.pyfunc.PyFuncModel",
```
harupy, line 363: Do we need type hints?

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+    def _log_dataset_tag(self, client, run_id):
+        """
+        Log dataset metadata as a tag "mlflow.datasets", if the tag already exists, it will
+        append current dataset metadata into existing tag content.
+        """
+        existing_dataset_metadata_str = client.get_run(run_id).data.tags.get("mlflow.datasets")
+        if existing_dataset_metadata_str is not None:
+            dataset_metadata_list = json.loads(existing_dataset_metadata_str)
+        else:
+            dataset_metadata_list = []
```
harupy, line 314:
```suggestion
        existing_dataset_metadata_str = client.get_run(run_id).data.tags.get("mlflow.datasets", "[]")
        dataset_metadata_list = json.loads(existing_dataset_metadata_str)
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+            from pyspark.sql import DataFrame as SparkDataFrame
+
+            supported_dataframe_types = (pd.DataFrame, SparkDataFrame)
+        except ImportError:
+            supported_dataframe_types = (pd.DataFrame,)
+
+        if name is not None and '"' in name:
+            raise ValueError(f'Dataset name cannot include a double quote (") but got name {name}')
+        if path is not None and '"' in path:
+            raise ValueError(f'Dataset path cannot include a double quote (") but got name {path}')
```
harupy, line 184:
```suggestion
            raise ValueError(f'Dataset path cannot include a double quote (") but got path {path}')
```

harupy, line 184:
```suggestion
        if name is not None and '"' in name:
            raise ValueError(f'Dataset name cannot include a double quote (") but got {name}')
        if path is not None and '"' in path:
            raise ValueError(f'Dataset path cannot include a double quote (") but got {path}')
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+                raise ValueError(
+                    "If data is a numpy array or list of evaluation features, "
+                    "labels must be a numpy array or list of evaluation labels"
+                )
+        elif isinstance(data, supported_dataframe_types):
+            if not isinstance(labels, str):
+                raise ValueError(
+                    "If data is a Pandas DataFrame or Spark DataFrame, labels must be the "
+                    "string name of a column from `data` that contains evaluation labels"
+                )
```
harupy, line 197: In multi-target regression, `labels` can become a list of strings:
https://scikit-learn.org/stable/modules/generated/sklearn.multioutput.MultiOutputRegressor.html

harupy: I don't think we need to address this in the initial version though.

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,530 @@
...
+
+    :param model_type: A string describing the model type. The default evaluator
+                       supports "regressor" and "classifier" as model types.
+    :param dataset: An instance of `EvaluationDataset` containing features
+                    labels (optional) for model evaluation.
+    :param run_id: The ID of the MLflow Run to which to log results. If
+                   unspecified, behavior depends on the specified `evaluator`.
+                   When `run_id` is unspecified, the default evaluator logs
+                   results to the current active run, creating a new active run if
+                   one does not exist.
```
harupy, line 457: What should happen if `run_id` is specified, but there is no active run?

dbczumar: Then the specified run should be resumed, assuming it already exists. If it does not exist, we should throw.

mlflow/models/evaluation/base.py
WeichenXu123: The design doc has decide the labels must be a string (only one label column allowed)
but using pandas_df.drop make code more simple, I adopt it.

mlflow/models/evaluation/base.py
WeichenXu123: we still need `if config is None: config = {}` because :

```
v = {'a': None}
v.get('a', {}) # will return None
```

mlflow/models/evaluation/base.py
harupy: Thanks for the explanation!

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,555 @@
...
+    """
+    Return a name list for all available Evaluators.
+    """
+    # import _model_evaluation_registry inside function to avoid circuit importing
+    from mlflow.models.evaluation.evaluator_registry import _model_evaluation_registry
+
+    return list(_model_evaluation_registry._registry.keys())
+
+
+class _StartRunOrReuseActiveRun:
```
harupy, line 427:
```python
from contextlib import contextmanager

@contextmanager
def _start_run_if_not_exists(run_id):
    active_run = mlflow.active_run()
    if not active_run:
        # Note `mlflow.start_run` throws if `run_id` is not found.
        with mlflow.start_run(run_id=run_id) as run:
            yield run.info.run_id
    else:
        if run_id and active_run.info.run_id != run_id:
            raise ValueError("...")

        yield active_run.info.run_id
```

Can we use `contextmanager` to simplify the implementation?

mlflow/models/evaluation/base.py
harupy: That makes sense. 

```suggestion
            config = evaluator_config.get(evaluator_name) or {}
```

mlflow/models/__init__.py
```diff
@@ -21,17 +21,26 @@
...
+    __all__ += [
+        "Model",
+        "ModelSignature",
+        "ModelInputExample",
+        "infer_signature",
+        "FlavorBackend",
+        "infer_pip_requirements",
+    ]
+except ImportError:
+    pass
```
harupy, line 46: Why do we need this change?

WeichenXu123: https://github.com/mlflow/mlflow/pull/5069#issuecomment-982880567

tests/models/test_evaluation.py
WeichenXu123: I revisit the hash calculation.
the hash calc over `numpy_array.tobytes` is risky because the tobytes return raw data in memory, which might change between different platform/environment.

I replace it with `pandas.util.hash_pandas_object`, `pandas.util.hash_array` and `struct.pack` (pack uint64 numbers into platform-independent bytes)

Now I can confirm the latest hash calculation is deterministic and platform-independent. if so I think we don't need test for "verified that this works across Python versions"

mlflow/models/__init__.py
dbczumar: @WeichenXu123 Can we include `infer_pip_requirements`, `Model`, and `FlavorBackend` unconditionally here - https://github.com/mlflow/mlflow/pull/5069/files#diff-433650164140d873b2734597cd9015aeb9f8bad5563230eaca543df1a117b8d4R27 and only try / catch `ModelInputSignature`, `infer_signature`, and `ModelInputExample`?

dbczumar: @WeichenXu123 Can we add a test case ensuring that `mlflow.evaluate` is importable from the skinny client? For reference, our skinny test suites are run here: https://github.com/mlflow/mlflow/blob/master/dev/run-python-skinny-tests.sh

mlflow/models/evaluation/base.py
dbczumar: @WeichenXu123 This method should not be expected to assign to `self._content`. We should assign to `self._content` here: https://github.com/mlflow/mlflow/pull/5069/files#diff-7b34221ea99c68bbb283c3a4058bec08a43fcbf72392b8e9692853aa3ac6a4daR56. Sorry if my previous comment was unclear.

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,559 @@
...
+    model_type: str,
+    dataset: "mlflow.models.evaluation.EvaluationDataset",
+    run_id=None,
+    evaluators=None,
+    evaluator_config=None,
+) -> "mlflow.models.evaluation.EvaluationResult":
+    """
+    Evaluate a pyfunc model on specified dataset, log evaluation results (metrics and
+    artifacts) into active run or specified mlflow run), and return evaluation results
+    containing metrics and artifacts.
```
dbczumar, line 475:
```suggestion
    Evaluate a PyFunc model on the specified dataset using one or more specified evaluators, and
    log resulting metrics & artifacts to MLflow Tracking.
```

mlflow/models/evaluation/base.py
```diff
@@ -0,0 +1,559 @@
...
+        )
+        dataset_metadata_list = json.loads(existing_dataset_metadata_str)
+
+        for metadata in dataset_metadata_list:
+            if metadata["hash"] == self.hash and metadata["name"] == self._user_specified_name:
+                break
+        else:
+            dataset_metadata_list.append(self._metadata)
+
+        dataset_metadata_str = json.dumps(dataset_metadata_list)
```
dbczumar, line 382: Can we remove spaces from the datasets JSON so that it's easier to write MLflow Search queries with ILIKE syntax? This way, we don't have to write queries that assume there's a single space between each attribute type (e.g. `name`) and the attribute value (e.g. `my_dataset`).

```suggestion
        dataset_metadata_str = json.dumps(dataset_metadata_list, separators=(',', ':'))
```

Should do it - reference https://stackoverflow.com/a/16311587/11952869.

Can we also add a test case to make sure there isn't any whitespace in the `mlflow.datasets` tag content?

dbczumar (approved):
@WeichenXu123 LGTM once remaining comments are addressed. Can we also address https://github.com/mlflow/mlflow/pull/5069/files#r760773789? This is excellent work! I tried it out manually with the test plugin, and it behaves as expected.

harupy (approved):
LGTM!

PR #5070, opened 2021-11-15
Enable mlflow-artifacts scheme as wrapper around http artifact scheme

BenWilson2:

Support mlflow-artifacts: scheme for proxied mlflow server artifact handling functionality. 
The implementation is a wrapper around the HTTP artifact repository scheme. 
This supports both a 'basic scheme' (user leaving the resolving server address out of the uri) and approaches that retain the host (and/or port) for replacing the scheme with the originating registered uri for the artifact server.

## How is this patch tested?

Unit tests

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Support for mlflow-artifacts scheme added (i.e., mlflow.create_experiment(name="my-experiment", artifact_location="mlflow-artifacts:/root/models/my-experiment-models/model-1"). Defining host, port, or both is also supported.

Components 

Interface 

Language 

Integrations


mlflow/store/artifact/mlflow_artifacts_repo.py
```diff
@@ -0,0 +1,25 @@
...
+class MlflowArtifactsRepository(HttpArtifactRepository):
+    """Scheme wrapper around HttpArtifactRepository for mlflow-artifacts server functionality"""
+
+    def __init__(self, artifact_uri):
+        parsed = _resolve_connection_params(artifact_uri)
+        tracking_uri = get_tracking_uri()
+        resolved_artifacts_uri = (
+            artifact_uri.replace("mlflow-artifacts:", f"{tracking_uri}")
+            .replace(f"mlflow-artifacts:{parsed.host}", f"{tracking_uri}")
+            .replace(f"mlflow-artifacts:{parsed.host}:{parsed.port}", f"{tracking_uri}")
```
dbczumar, line 22: We should only swap in the tracking URI if the artifact URI doesn't specify a hostname.

harupy: Just to confirm, if `artifact_uri` is `mlflow-artifacts:/my/artifact/path`, it should be convert to `<tracking uri>/api/2.0/mlflow-artifacts/artifacts/my/artifact/path`, correct?

BenWilson2: Yep! I added some additional parameter tests for the mock harness to validate the user-defined pathing following the scheme patch.

mlflow/store/artifact/mlflow_artifacts_repo.py
```diff
@@ -0,0 +1,12 @@
+from mlflow.store.artifact.http_artifact_repo import HttpArtifactRepository
+from mlflow.tracking._tracking_service.utils import get_tracking_uri
+
+
+class MlflowArtifactsRepository(HttpArtifactRepository):
+    """Scheme wrapper around HttpArtifactRepository for mlflow-artifacts server functionality"""
+
+    def __init__(self, artifact_uri):
+        tracking_uri = get_tracking_uri()
+        resolved_artifacts_uri = artifact_uri.replace("mlflow-artifacts:", f"{tracking_uri}")
```
dbczumar, line 10: We should only perform this replacement if the URI doesn't specify a host. We also need to handle the case where the URI has one or three slashes after the scheme.

Right now, if the URI specifies a host (e.g. `mlflow-artifacts://myhostname:5000`), the following behavior occurs:
```
In [11]: MlflowArtifactsRepository("mlflow-artifacts://foo:5050/my/artifact/path").artifact_uri
Out[11]: 'http://localhost:5000//foo:5050/my/artifact/path'
```

Similarly, if a hostname-less `mlflow-artifacts` URI is specified with one or three slashes after the scheme, the following behaviors occur:

```
In [14]: mlflow.set_tracking_uri("http://localhost:5000/")

In [15]: MlflowArtifactsRepository("mlflow-artifacts:/my/artifact/path").artifact_uri
Out[15]: 'http://localhost:5000//my/artifact/path'

In [16]: MlflowArtifactsRepository("mlflow-artifacts:///my/artifact/path").artifact_uri
Out[16]: 'http://localhost:5000////my/artifact/path'
```

While the extra slashes in the path shouldn't be harmful, it would be nice to remove them.

To clarify re:hostnames - when a user supplies an artifact URI of the form `mlflow-artifacts://<hostname>/<path>` or `mlflow-artifacts:/<hostname>:<port>/<path>`, we should resolve the URI to `http://<hostname>/api/2.0/mlflow-artifacts/artifacts/<path>` or ``http://<hostname>:<port>/api/2.0/mlflow-artifacts/artifacts/<path>` respectively.

Can we add test cases for each of the above scenarios as well?

BenWilson2: Some... (not so minor) changes to handle resolving all of the different ways that this uri can be configured. I have a validation check in here for raising an Exception if only a port is passed; I'd really like an opinion on whether this is needed or if it's a silly thing to check for.

Also, should I take those private functions and make them class methods? I don't see those functions being particularly useful outside of this sub class of HttpArtifactRepository. 
@dbczumar @harupy any thoughts / advice on these?

mlflow/store/artifact/mlflow_artifacts_repo.py
```diff
@@ -0,0 +1,81 @@
...
+            f"{uri_parse.host}, but no host was defined."
+        )
+
+
+class MlflowArtifactsRepository(HttpArtifactRepository):
+    """Scheme wrapper around HttpArtifactRepository for mlflow-artifacts server functionality"""
+
+    def __init__(self, artifact_uri):
+
+        self._session = requests.Session()
```
harupy, line 45: Do we need this line? `HttpArtifactRepository.__init__` does the same thing, right?

BenWilson2: I was wondering the same thing. When I don't create a requests.Session() object, I get this Exception (that is ignored in the test suite, but it throws nonetheless):

```Traceback (most recent call last):
  File "/Users/benwilson/Repos/mlflow/mlflow/store/artifact/http_artifact_repo.py", line 18, in __del__
    self._session.close()
AttributeError: 'MlflowArtifactsRepository' object has no attribute '_session'
```

harupy: Interesting. It looks like `__del__` even when `__init__` fails.

harupy: `__del__` of `HttpArtifactRepository` should look like:

```
del __del__(self):
    if hasattr(self. "_session"):
        self.close()
```

BenWilson2: I think it's because requests keeps a session alive by default https://2.python-requests.org/en/master/user/advanced/#keep-alive so that call to __del__ is the only thing saving our connection pool from exhaustion ;) I think the other way to 'auto-handle it' is by wrapping the request call in a with statement, but that really wouldn't work with how HttpArtifactRepository needs to function.
The interesting thing about it is that the subclass of MlflowArtifactsRepository can't inherit the `requests.Session()` instantiation from `super()`. It will throw saying that it's not accessible. Must be some sort of protection in requests that keep it instance-scoped to protect people from leaving sessions active inadvertently?

BenWilson2: great point! Making that validation check and I'll push in a minute :)

mlflow/store/artifact/mlflow_artifacts_repo.py
harupy: > I have a validation check in here for raising an Exception if only a port is passed.

This looks good to me.

> should I take those private functions and make them class methods?

I don't have a strong preference between private functions and class methods. The current code looks ok to me.

mlflow/store/artifact/mlflow_artifacts_repo.py
```diff
@@ -0,0 +1,79 @@
...
+        tracking_uri = get_tracking_uri()
+
+        track_parse = _parse_artifact_uri(tracking_uri)
+
+        uri_parse = _parse_artifact_uri(artifact_uri)
+
+        # Check to ensure that a port is present with no hostname
+        _validate_port_mapped_to_hostname(uri_parse)
+
+        api_path = "/api/2.0/mlflow-artifacts/artifacts"
```
harupy, line 57: I think we need to consider a case where a user runs `mlflow server` with `--static-prefix`.

tests/store/artifact/test_mlflow_artifact_repo.py
```diff
@@ -0,0 +1,281 @@
...
+
+def test_mlflow_artifact_uri_formats_resolved():
+
+    tracking_uri = urlparse(get_tracking_uri())
+
+    conditions = [
+        (
+            "mlflow-artifacts://myhostname:4242/my/artifact/path/hostport",
+            f"{tracking_uri.scheme}:"
+            f"//myhostname:4242/api/2.0/mlflow-artifacts/artifacts/my/artifact/path/hostport",
```
harupy, line 26: It took me a while to understand this is one string 😅

mlflow/store/artifact/mlflow_artifacts_repo.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def _validate_port_mapped_to_hostname(uri_parse):
+    # This check is to catch an mlflow-artifacts uri that has a port designated but no
+    # hostname specified. `urllib.parse.urlparse` will treat such a uri as a filesystem
+    # definition, mapping the provided port as a hostname value if this condition is not
+    # validated.
+    if uri_parse.host and _check_if_host_is_numeric(uri_parse.host) and not uri_parse.port:
+        raise MlflowException(
+            f"The mlflow-artifacts uri was supplied with a port number: "
```
harupy, line 34:
```suggestion
            "The mlflow-artifacts uri was supplied with a port number: "
```

nit

tests/store/artifact/test_mlflow_artifact_repo.py
```diff
@@ -0,0 +1,275 @@
...
+        artifact_repo = MlflowArtifactsRepository(submit)
+        assert artifact_repo.resolve_uri(submit) == resolved
+    with pytest.raises(
+        MlflowException,
+        match="The mlflow-artifacts uri was supplied with a port number: 5000, but no "
+        "host was defined.",
+    ):
+        uri = MlflowArtifactsRepository(  # pylint: disable=unused-variable
+            failing_condition
+        ).artifact_uri
```
harupy, line 54:
```suggestion
        MlflowArtifactsRepository(failing_condition)
```

nit

mlflow/store/artifact/mlflow_artifacts_repo.py
BenWilson2: changed

tests/store/artifact/test_mlflow_artifact_repo.py
BenWilson2: changed to just the class instantiation

mlflow/store/artifact/mlflow_artifacts_repo.py
```diff
@@ -0,0 +1,79 @@
...
+class MlflowArtifactsRepository(HttpArtifactRepository):
+    """Scheme wrapper around HttpArtifactRepository for mlflow-artifacts server functionality"""
+
+    def __init__(self, artifact_uri):
+
+        super().__init__(self.resolve_uri(artifact_uri))
+
+    @classmethod
+    def resolve_uri(cls, artifact_uri):
+        tracking_uri = get_tracking_uri()
```
harupy, line 48: Can we raise an exception if the scheme of `tracking_uri` is not `http` or `https` (e.g. `sqlite:///foo/bar`)?

BenWilson2: great idea :) added! I also simplified the parsing logic and temporarily set the tracking uri in the test function to validate against a valid server address (then reverted the global variable to not affect other tests with an inadvertent side-effect)

BenWilson2:
I installed the new version of black after merging master into this branch. Black on local (21.10b0) suggests no changes, but the lint test is failing in CI. @harupy have you seen this since the merge of the black update?

harupy:
@BenWilson2 

Does `black --version` print out `21.10b0`?

```
black --version
```

tests/store/artifact/test_mlflow_artifact_repo.py
```diff
@@ -0,0 +1,300 @@
...
+        ),
+        (
+            f"mlflow-artifacts:{base_path}/nohost",
+            f"{tracking_uri.scheme}://{tracking_uri.netloc}/{base_api_uri}{base_path}/nohost",
+        ),
+        (
+            f"mlflow-artifacts://{base_path}/redundant",
+            f"{tracking_uri.scheme}://{tracking_uri.netloc}/{base_api_uri}{base_path}/redundant",
+        ),
+        ("mlflow-artifacts:/", uri_temp,),
```
harupy, line 50:
```suggestion
        ("mlflow-artifacts:/", uri_temp,),
```

@BenWilson2 Removing the command here should fix the lint issue.

tests/store/artifact/test_mlflow_artifact_repo.py
```diff
@@ -0,0 +1,303 @@
...
+from mlflow.exceptions import MlflowException
+from mlflow.tracking._tracking_service.utils import get_tracking_uri, set_tracking_uri
+
+ORIGINAL_TRACKING_URI = get_tracking_uri()
+
+
+def _set_temp_tracking_uri():
+    temp_uri = "http://localhost:5000/api/2.0/mlflow-artifacts/artifacts"
+    set_tracking_uri(temp_uri)
+    return temp_uri
```
harupy, line 18: Using a autouse fixture might cleaner:

```python
@pytest.fixture(scope="module", autouse=True)
def set_tracking_uri():
    with mock.patch(
        "mlflow.store.artifact.mlflow_artifacts_repo.get_tracking_uri",
        return_value="http://localhost:5000/api/2.0/mlflow-artifacts/artifacts",
    ):
        yield
```

mlflow/store/artifact/mlflow_artifacts_repo.py
```diff
@@ -0,0 +1,96 @@
...
+
+        track_parse = _parse_artifact_uri(tracking_uri)
+
+        uri_parse = _parse_artifact_uri(artifact_uri)
+
+        print(f"\nuri_parse: {uri_parse}")
+        print(f"track_parse: {track_parse}")
```
harupy, line 63:
```suggestion
```

mlflow/store/artifact/mlflow_artifacts_repo.py
harupy: Thanks for the update! Tracking URI looks like `http://localhost:5000` and doesn't contain `/api/2.0/mlflow-artifacts/artifacts`.

tests/store/artifact/test_mlflow_artifact_repo.py
BenWilson2: so much cleaner. Thanks Haru!

mlflow/store/artifact/mlflow_artifacts_repo.py
BenWilson2: updated the tests and cleaned all of that up

mlflow/store/artifact/mlflow_artifacts_repo.py
BenWilson2: deleted my debug ;)

mlflow/store/artifact/mlflow_artifacts_repo.py
harupy: Sorry I wasn't clear. The resolved artifact URI needs to have `/api/2.0/mlflow-artifacts/artifacts` (e.g. `http://localhost:5000/api/2.0/mlflow-artifacts/artifacts/path/to/dir`).

harupy (approved):
LGTM!

PR #5167, opened 2021-12-15
Only generate model uuid when logging model

WeichenXu123:

Only generate model uuid when logging model.

Before:
When loading model from a old version mlflow model, a new model uuid will be generated. Loading multiple times will generate multiple different model uuid.

After:
When loading model from a old version mlflow model, the model uuid attribute will be None.

## How is this patch tested?

Unit tests.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/models/model.py
+++ b/mlflow/models/model.py
@@ -54,10 +54,7 @@ class Model(object):
         self.flavors = flavors if flavors is not None else {}
         self.signature = signature
         self.saved_input_example_info = saved_input_example_info
-        if callable(model_uuid):
-            self.model_uuid = model_uuid()
-        else:
-            self.model_uuid = model_uuid
+        self.model_uuid = model_uuid() if callable(model_uuid) else model_uuid
         self.__dict__.update(kwargs)
 
     def __eq__(self, other):
--- a/mlflow/models/model.py
+++ b/mlflow/models/model.py
@@ -6,7 +6,7 @@ import yaml
 import os
 import uuid
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union, Callable
 
 import mlflow
 from mlflow.exceptions import MlflowException
@@ -42,6 +42,7 @@ class Model(object):
         flavors=None,
         signature=None,  # ModelSignature
         saved_input_example_info: Dict[str, Any] = None,
+        model_uuid: Union[str, Callable, None] = lambda: uuid.uuid4().hex,
         **kwargs,
     ):
         # store model id instead of run_id and path to avoid confusion when model gets exported
@@ -53,7 +54,10 @@ class Model(object):
         self.flavors = flavors if flavors is not None else {}
         self.signature = signature
         self.saved_input_example_info = saved_input_example_info
-        self.model_uuid = uuid.uuid4().hex
+        if callable(model_uuid):
+            self.model_uuid = model_uuid()
+        else:
+            self.model_uuid = model_uuid
         self.__dict__.update(kwargs)
 
     def __eq__(self, other):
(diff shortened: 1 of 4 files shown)


tests/pyfunc/test_model_export_with_loader_module_and_data_path.py
```diff
@@ -102,13 +112,17 @@ def test_model_save_load(sklearn_knn_model, iris_data, tmpdir, model_path):
 
     reloaded_model_config = Model.load(os.path.join(model_path, "MLmodel"))
     assert model_config.__dict__ == reloaded_model_config.__dict__
+    assert model_config.model_uuid is not None and _is_valid_uuid(model_config)
```
harupy, line 115:
```suggestion
    assert model_config.model_uuid is not None
    assert _is_valid_uuid(model_config.model_uuid)
```

harupy:
Can you add a test that `model_uuid` is None when we load `Model` from `MLmodel` file without `model_uuid`?

mlflow/models/model.py
```diff
@@ -186,7 +186,8 @@ def log(
         with TempDir() as tmp:
             local_path = tmp.path("model")
             run_id = mlflow.tracking.fluent._get_or_start_run().info.run_id
-            mlflow_model = cls(artifact_path=artifact_path, run_id=run_id)
+            model_uuid = uuid.uuid4().hex
+            mlflow_model = cls(artifact_path=artifact_path, run_id=run_id, model_uuid=model_uuid)
```
WeichenXu123, line 190: We should not set uuid here. Otherwise there're many other place we need also set uuid. 
e.g. 
https://github.com/mlflow/mlflow/blob/f9b5745b68c815adb411453048317ee46ac52a31/mlflow/pyfunc/__init__.py#L1072
https://github.com/mlflow/mlflow/blob/f9b5745b68c815adb411453048317ee46ac52a31/mlflow/catboost.py#L122
etc.

I move the set uuid code into Model constructor again.

mlflow/models/model.py
```diff
@@ -42,7 +42,6 @@ def __init__(
         flavors=None,
         signature=None,  # ModelSignature
         saved_input_example_info: Dict[str, Any] = None,
-        model_uuid=None,
```
dbczumar, line 45: Instead of removing `model_uuid` from the constructor and modifying the attribute during `load()`, can we define a third type of value for `model_uuid`?

1. By default, `model_uuid` should be a function that generates a UUID, e.g. `lambda: uuid.uuid4().hex`
2. Model uuid could be `None`, indicating that the model has no ID
3. Model uuid could be a string, indicating that the model already has a UUID

In the constructor, we can check if the input is a function and, if it is, call it to generate an ID. Otherwise, set `self.model_uuid = model_uuid`.

mlflow/models/model.py
```diff
@@ -54,7 +53,7 @@ def __init__(
         self.flavors = flavors if flavors is not None else {}
         self.signature = signature
         self.saved_input_example_info = saved_input_example_info
-        self.model_uuid = uuid.uuid4().hex if model_uuid is None else model_uuid
+        self.model_uuid = uuid.uuid4().hex
```
harupy, line 56: This basically means every `Model` instance has a different `model_uuid`. Is this really desired behavior?

mlflow/models/model.py
WeichenXu123: sounds good

mlflow/models/model.py
WeichenXu123: Every new constructed model has a different ID, this is desired.
We only need keep the rule that the model load back from saved model should load old ID back or set it None if saved model don't have ID

mlflow/models/model.py
```diff
@@ -54,7 +54,10 @@ def __init__(
         self.flavors = flavors if flavors is not None else {}
         self.signature = signature
         self.saved_input_example_info = saved_input_example_info
-        self.model_uuid = uuid.uuid4().hex if model_uuid is None else model_uuid
+        if callable(model_uuid):
+            self.model_uuid = model_uuid()
+        else:
+            self.model_uuid = model_uuid
```
harupy, line 60:
```suggestion
        self.model_uuid = model_uuid() if callable(model_uuid) else model_uuid
```

dbczumar (approved):
LGTM! Thanks @WeichenXu123 !

PR #5206, opened 2021-12-28
Remove param 'evals_result' 'early_stopping_rounds' in lightgbm version > 3.3.1

liangz1:
Signed-off-by: Liang Zhang <liang.zhang@databricks.com>

The parameter `evals_result` is removed in the master of lightgbm: https://github.com/microsoft/LightGBM/pull/4882.
The parameter `early_stopping_rounds` is removed in the master of lightgbm: https://github.com/microsoft/LightGBM/pull/4908.

We should also remove this param in our test.

This PR also fixed the sagemaker test failure.

## How is this patch tested?

Existing tests.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -25,6 +25,7 @@ import shutil
 import logging
 import functools
 from copy import deepcopy
+from packaging.version import Version
 
 import mlflow
 from mlflow import pyfunc
@@ -447,10 +448,13 @@ def autolog(
             "fobj",
             "feval",
             "init_model",
-            "evals_result",
             "learning_rates",
             "callbacks",
         ]
+        if Version(lightgbm.__version__) <= Version("3.3.1"):
+            # The parameter `evals_result` in `lightgbm.train` is removed in this PR:
+            # https://github.com/microsoft/LightGBM/pull/4882
+            unlogged_params.append("evals_result")
 
         params_to_log_for_fn = get_mlflow_run_params_for_fn_args(
             original, args, kwargs, unlogged_params
--- a/mlflow/lightgbm.py
+++ b/mlflow/lightgbm.py
@@ -487,25 +487,9 @@ def autolog(
             # training model
             model = original(*args, **kwargs)
 
-            # If early_stopping_rounds is present, logging metrics at the best iteration
+            # If early stopping is activated, logging metrics at the best iteration
             # as extra metrics with the max step + 1.
-            if Version(lightgbm.__version__) <= Version("3.3.1"):
-                # The parameter `early_stopping_rounds` in `lightgbm.train` is removed in this PR:
-                # https://github.com/microsoft/LightGBM/pull/4908
-                early_stopping_index = all_arg_names.index("early_stopping_rounds")
-                early_stopping = (
-                    num_pos_args >= early_stopping_index + 1 or "early_stopping_rounds" in kwargs
-                )
-            else:
-                early_stopping = False
-                if "callbacks" in kwargs and kwargs["callbacks"] is not None:
-                    for cb in kwargs["callbacks"]:
-                        if (
-                            hasattr(cb, "__qualname__")
-                            and cb.__qualname__ == "early_stopping.<locals>._callback"
-                        ):
-                            early_stopping = True
-                            break
+            early_stopping = model.best_iteration > 0
             if early_stopping:
                 extra_step = len(eval_results)
                 autologging_client.log_metrics(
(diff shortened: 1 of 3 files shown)


mlflow/lightgbm.py
```diff
@@ -484,10 +489,23 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+                early_stopping = (
+                    num_pos_args >= early_stopping_index + 1 or "early_stopping_rounds" in kwargs
+                )
+            else:
+                early_stopping = False
+                if "callbacks" in kwargs and kwargs["callbacks"] is not None:
+                    for cb in kwargs["callbacks"]:
+                        if (
+                            hasattr(cb, "__qualname__")
+                            and cb.__qualname__ == "early_stopping.<locals>._callback"
```
WeichenXu123, line 505: This checking is a little tricky , is there better way ?  I think you want to check whether the callback is `lgb.early_stopping(5)` , can we directly compare the function object with it ?

WeichenXu123 (approved):
LGTM except one comment

mlflow/lightgbm.py
liangz1: Directly comparing function object seems not working:
```python
lightgbm.early_stopping(stopping_rounds=1) is lightgbm.early_stopping(stopping_rounds=1) # False
lightgbm.early_stopping(stopping_rounds=1) == lightgbm.early_stopping(stopping_rounds=1) # False
```

mlflow/lightgbm.py
```diff
@@ -484,10 +489,23 @@ def log_feature_importance_plot(features, importance, importance_type):
...
+                        ):
+                            early_stopping = True
+                            break
             if early_stopping:
```
liangz1, line 509: https://github.com/microsoft/LightGBM/blob/ce486e5b45a6f5e67743e14765ed139ff8d532e5/python-package/lightgbm/engine.py#L226-L263
`best_iteration` is initiated as 0 and is only assigned with a positive int when `EarlyStoppingException` is raised (this means early stopping is activated, including using `early_stopping_rounds` param, `lightgbm.early_stopping` callback, or even user-defined early_stopping callbacks that raises `EarlyStoppingException`).
iff early stopping is not activated, `best_iteration` will be 0.

mlflow/lightgbm.py
```diff
@@ -482,12 +487,9 @@ def log_feature_importance_plot(features, importance, importance_type):
...
             model = original(*args, **kwargs)
 
-            # If early_stopping_rounds is present, logging metrics at the best iteration
+            # If early stopping is activated, logging metrics at the best iteration
             # as extra metrics with the max step + 1.
-            early_stopping_index = all_arg_names.index("early_stopping_rounds")
-            early_stopping = (
-                num_pos_args >= early_stopping_index + 1 or "early_stopping_rounds" in kwargs
-            )
+            early_stopping = model.best_iteration > 0
```
liangz1, line 492: `model.best_iteration > 0` can be used to determine whether early stopping is activated.

`model.best_iteration` is initiated as 0 and is assigned with a positive int when `EarlyStoppingException` is raised (this means early stopping is activated, including using the `early_stopping_rounds` param, `lightgbm.early_stopping` callback function, or even a user-defined early_stopping callback that raises `EarlyStoppingException`).
`model.best_iteration` is returned as 0 when early stopping is not activated.

Reference: https://github.com/microsoft/LightGBM/blob/ce486e5b45a6f5e67743e14765ed139ff8d532e5/python-package/lightgbm/engine.py#L226-L263

liangz1:
I executed the sagemaker tests on my laptop and they all passed.

tests/sagemaker/mock/__init__.py
```diff
@@ -1076,10 +1075,7 @@ def response_object(self):
         return response
 
 
-# Create a SageMaker backend for each EC2 region
-sagemaker_backends = {}
-for region, ec2_backend in ec2_backends.items():
-    new_backend = SageMakerBackend()
-    sagemaker_backends[region] = new_backend
+# Create a SageMaker backend for EC2 region: "us-west-2"
+sagemaker_backends = {"us-west-2": SageMakerBackend()}
```
liangz1, line 1079: According to the description of https://github.com/spulec/moto/pull/4699:
> all backends are now generated on-request, rather then on start-up.

So we need to create a SageMaker backend for "us-west-2", which is used in our test. Since "us-west-2" is **hardcoded** in our tests as the default AWS region, I also hardcoded it here.

With moto==2.3.0, the old implementation only returns one region "us-east-1", which is the default region used by moto.

BenWilson2: We can probably update the small-requirements.txt now to instead of excluding version 2.0.7, just force use of >=2.3.0.

liangz1: This change is also compatible with moto<2.3.0, so we don't have to change it now.

PR #5230, opened 2022-01-06
Return Model from log_model

liangz1:

Provide a return value from `mlflow.*.log_model()` that contains model metadata such as `model_uri`, `run_id`,...
This metadata makes it easier for the user to do follow-up model loading.

## How is this patch tested?

Existing tests + tests/models/test_model.py test_model_info()

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

`mlflow.*.log_model()` returns a ModelInfo instance that contains the metadata of the logged model.

Components 

Interface 

Language 

Integrations


mlflow/models/model.py
```diff
@@ -109,6 +127,21 @@ def saved_input_example_info(self, value: Dict[str, Any]):
         # pylint: disable=attribute-defined-outside-init
         self._saved_input_example_info = value
 
+    def get_model_info(self) -> ModelInfo:
```
harupy, line 130: Do we need this method? Can we just add `model_uri` as an attribute of `Model`?

WeichenXu123: We can add `model_uri` as a property of `Model` like:

@property
def model_uri():
  return artifact_utils.get_artifact_uri(self.run_id, self.artifact_path)

mlflow/models/model.py
```diff
@@ -109,6 +127,21 @@ def saved_input_example_info(self, value: Dict[str, Any]):
...
+        """Create a ModelInfo instance that contains the model metadata."""
+        return ModelInfo(
+            run_id=self.run_id,
+            artifact_path=self.artifact_path,
+            model_uri="runs:/{}/{}".format(self.run_id, self.artifact_path),
```
WeichenXu123, line 135: Let's use `artifact_utils.get_artifact_uri` to generate uri.

liangz1: I think the `runs:/` scheme is easy to read (shorter) and also consistent with our examples in the documentation ([example](https://mlflow.org/docs/latest/python_api/mlflow.pytorch.html#mlflow.pytorch.load_model)), while the absolute artifact uri returned by `get_artifact_uri` is much longer.
Would you mind elaborating on the reasons for using `get_artifact_uri`?

mlflow/models/model.py
liangz1: @harupy After offline discussions, we decide to use `ModelInfo` instead of `Model` for the following reasons:
1. `Model` contains many other methods that do not make sense in the return value, e.g. `add_flavor(...)` since `Model` is mutable while the return value should be immutable.
2. `ModelInfo` k-v contents can be printed since it's implemented as a `namedtuple`, while `Model` k-v contents cannot be printed out-of-the-box.

```
m = mlflow.xxx_yyy.log_model(...)
m.model_uri
```
is the current usage pattern, where m is a `ModelInfo` instance.

mlflow/models/model.py
WeichenXu123: why not add a helper method in `artifact_utils` modules for this ?

liangz1: I think the new helper method will not add much value because 1) it's a pretty simple format to manually type it. 2) the string literal is mentioned in many places in our document, so it's easy for the user to discover and understand, while the proposed helper function will need an extra item in our documentation, which seems to be less discoverable.

dbczumar (approved):
LGTM once all model flavors are updated!

mlflow/catboost.py
```diff
@@ -226,8 +226,12 @@ def log_model(
     :param pip_requirements: {{ pip_requirements }}
     :param extra_pip_requirements: {{ extra_pip_requirements }}
     :param kwargs: kwargs to pass to `CatBoost.save_model`_ method.
+    :return: A `ModelInfo` namedtuple instance that contains the metadata of the logged model,
+             including: `run_id`, `artifact_path`, `model_uri`, `utc_time_created`, `flavors`,
+             `model_uuid`, `saved_input_example_info`, `signature`, `input_schema`, and
+             `output_schema`.
```
harupy, line 232: Can we document `ModelInfo` somewhere to simplify the param description here?

harupy: This post (written by the sphinx author) describes how to write a document for `namedtuple`:

https://qiita.com/tk0miya/items/d13781658ed38045a693

harupy (approved):
LGTM once https://github.com/mlflow/mlflow/pull/5230#discussion_r781713636 is addressed.

mlflow/catboost.py
liangz1: Currently, the namedtuple class is not captured by sphinx, so rewriting the `__doc__` field does not show up in the page either.
I tried `sphinx-toolbox` pip package to support namedtuple, but it has a bug with python 3.7.3 [`_ForwardRef` does not exist in python 3.7.3 while it's supposed to exist](https://github.com/sphinx-toolbox/sphinx-toolbox/blob/c40be459eabee0df4b4d15bd300d015a6fe58c2d/sphinx_toolbox/more_autodoc/typehints.py#L129).

dbczumar: In that case, can we make this a Python class?

liangz1: `class ModelInfo(NamedTuple):` seems to be working. Let me try that.

mlflow/models/model.py
```diff
@@ -22,12 +23,37 @@
...
+ModelInfo.__doc__ = "The metadata of a logged MLflow Model."
+ModelInfo.run_id.__doc__ = "The ``run_id`` associated with the logged model."
+ModelInfo.artifact_path.__doc__ = "Run relative path identifying the logged model."
+ModelInfo.model_uri.__doc__ = "The ``model_uri`` of the logged model."
+ModelInfo.utc_time_created.__doc__ = "The UTC time that the logged model is created."
+ModelInfo.flavors.__doc__ = "Flavor module to save the model with."
+ModelInfo.model_uuid.__doc__ = "The ``model_uuid`` of the logged model."
+ModelInfo.saved_input_example_info.__doc__ = ""
+ModelInfo.signature.__doc__ = (
+    "A :py:class:`ModelSignature <mlflow.models.ModelSignature>` that "
```
liangz1, line 52: @harupy 
```
:py:class:`ModelSignature <mlflow.models.ModelSignature>`
```
^^^This works while the following does not:
```
:py:class:`ModelSignature`
```

mlflow/models/model.py
```diff
@@ -109,6 +135,22 @@ def saved_input_example_info(self, value: Dict[str, Any]):
         # pylint: disable=attribute-defined-outside-init
         self._saved_input_example_info = value
 
+    def get_model_info(self):
+        """
+        Create a :py:class:`ModelInfo <mlflow.models.model.ModelInfo>` instance that contains the
```
liangz1, line 140: @harupy Also, 
```
:py:class:`ModelInfo <mlflow.models.model.ModelInfo>`
```
works while the following does not:
```
:py:class:`ModelInfo`
```

Thanks for helping!

mlflow/models/model.py
harupy, line 52:
```suggestion
class ModelInfo(NamedTuple):
    """
    The metadata of a logged MLflow Model.
    """

    #: The metadata of a logged MLflow Model.
    run_id: str
    #: The metadata of a logged MLflow Model.
    artifact_path: str
    #: The metadata of a logged MLflow Model.
    model_uri: str
    #: The metadata of a logged MLflow Model.
    utc_time_created: str
    #: The metadata of a logged MLflow Model.
    flavors: Dict[str, Any]
    #: The metadata of a logged MLflow Model.
    model_uuid: str
    #: The metadata of a logged MLflow Model.
    saved_input_example_info: Dict[str, Any]
    #: A :py:class:`ModelSignature <mlflow.models.ModelSignature>` that
    #: describes the model input and output :py:class:`Schema <mlflow.types.Schema>`.
    signature: ModelSignature
```

harupy: I think this is much cleaner than overwriting `__doc__`.

harupy: Note `#:` is a special syntax for sphinx.

harupy: And we should use `Optional` for properties that can be None (e.g. `signature`)

mlflow/models/model.py
```diff
@@ -22,12 +23,37 @@
...
 
+class ModelInfo(NamedTuple):
+    run_id: str
+    artifact_path: str
+    model_uri: str
+    utc_time_created: str
+    flavors: Dict[str, Any]
+    model_uuid: str
+    saved_input_example_info: Dict[str, Any]
+    signature: ModelSignature
```
harupy, line 40:
```suggestion
    signature: Optional[ModelSignature]
```

Can we use `Optional` for properties that can be None?

docs/source/python_api/mlflow.models.rst
```diff
@@ -6,3 +6,7 @@ mlflow.models
     :undoc-members:
     :show-inheritance:
 
+.. autoclass:: mlflow.models.model.ModelInfo
+    :members:
+    :undoc-members:
+    :exclude-members: ModelSignature
```
liangz1, line 12: This line removes the duplicate entry of `ModelSignature` (should not appear in `ModelInfo`) imported by https://github.com/mlflow/mlflow/blob/00bf853a4d832e168c3e7ff4f28e7b48bccd7ccf/mlflow/models/model.py#L36, which is to avoid a global import since pandas is not a dependency of mlflow-skinny.

mlflow/models/model.py
```diff
@@ -22,12 +22,62 @@
...
+    #: ``{"artifact_path": "input_example.json", "type": "dataframe", "pandas_orient": "split"}``.
+    saved_input_example_info: Optional[Dict[str, Any]]
+
+    #: A :py:class:`ModelSignature <mlflow.models.ModelSignature>` that
+    #: describes the model input and output :py:class:`Schema <mlflow.types.Schema>`.
+    signature: Optional[ModelSignature]
+
+    #: The UTC time that the logged model is created, e.g., ``'2022-01-12 05:17:31.634689'``.
+    utc_time_created: str
+
```
liangz1, line 79: I added example values for some attributes.

mlflow/models/model.py
liangz1: I changed it to a native python dictionary that is `ModelSignature.to_dict()` to avoid importing pandas in mlflow skinny. Another reason is that for the user's information, there is no need to return the `ModelSignature` object. A dictionary should be sufficient for the purpose of information reading.

mlflow/models/model.py
liangz1: Thanks! I updated the syntax.

mlflow/models/model.py
```diff
@@ -102,13 +150,33 @@ def signature(self, value):
 
     @property
     def saved_input_example_info(self) -> Optional[Dict[str, Any]]:
+        """
+        A dictionary that contains the metadata of the saved input example, e.g.,
+        ``{"artifact_path": "input_example.json", "type": "dataframe", "pandas_orient": "split"}``.
+        """
```
liangz1, line 156: Improve the `Model` class property docstring.

PR #5306, opened 2022-01-25
Condense run comparison table

WeichenXu123:

Condense run comparison table. 
1. Add hover text for cell element.
2. For parameters and metrics section, limit the max height and make it vertical scrollable.
3. Make parameters and metrics section collapsable.
4. Add "Show diff only" toggle button for parameters and metrics section.
5. Fix the left header column when horizontally scroll the parameters/metrics section.

Demostration:

https://user-images.githubusercontent.com/19235986/151132312-4b68784c-183a-4315-afd7-013951f78bba.mov

## How is this patch tested?

Run following testing code to generate experiments:
```
import mlflow
from sklearn.datasets import make_classification

mlflow.sklearn.autolog()

X, y = make_classification(n_samples=1000, n_classes=10, n_informative=5, random_state=1)

for i in range(10):
    with mlflow.start_run(run_name=f'linear_model_run_{i}') as run:
        model = LogisticRegression(
            solver='liblinear',
            max_iter=(i + 3),
            C=(i + 10),
        ).fit(X, y)
```

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Condense run comparison table.

Add hover text for cell element.
For parameters and metrics section, limit the max height and make it vertical scrollable.
Make parameters and metrics section collapsable.
Add "Show diff only" toggle button for parameters and metrics section.
Fix the left header column when horizontally scroll the parameters/metrics section.

Components 

Interface 

Language 

Integrations


mlflow/server/js/src/experiment-tracking/components/CompareRunView.css
```diff
@@ -1,29 +1,65 @@
...
+  background-color: yellow;
+  color: black;
+  text-align: center;
+  padding: 5px 0;
+  border-radius: 6px;
+  overflow: hidden;
+  pointer-events: none;
+  position:absolute;
+  z-index: 1;
+}
```
harupy, line 46: Do we really need these styels?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -230,10 +310,44 @@ export class CompareRunView extends Component {
...
   }
 
+  onMouseEnterTableCell(e) {
+    var hoverTextElem = document.getElementById("table-cell-hover-text");
+    hoverTextElem.style.visibility = 'visible';
+    hoverTextElem.style.display = 'block';
+    hoverTextElem.style.left = `${e.clientX + window.scrollX}px`;
+    hoverTextElem.style.top = `${e.clientY + window.scrollY}px`;
+    hoverTextElem.innerHTML = e.target.innerHTML
+  }
```
harupy, line 320: Looks like we tweak css a lot. Is there any ant design component that provides features we require?

WeichenXu123: Removed this, replace it with antd Tooltip component.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.css
WeichenXu123: Removed this, replace it with antd Tooltip component.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -67,51 +96,90 @@ export class CompareRunView extends Component {
...
+
+    function onTableBlockScrollHanlder(e) {
+      const blocks = document.querySelectorAll('.compare-table .table-block');
+      for (let index = 0; index < blocks.length; ++index) {
+        const block = blocks[index];
+        if (block !== e.target) {
+          block.scrollLeft = e.target.scrollLeft;
+        }
+      }
+    }
```
WeichenXu123, line 112: Why need this handler:

here we have several table blocks: “run_id/name/.. block”, “parameter block”, “metric block”,
when any of them scroll **horizontally**, then other block must scroll together.

previously, they are always scroll together because they are in the same table,

But, after we add the vertical scroll bar to “parameter block”, “metric block”, things changed, these blocks they scroll individually in horizon direction, when one scroll, others do not scroll.

So we need js code to control this, the js code triggerred by one DOM element “onscroll”, but it need control other DOM element scroll offset

WeichenXu123: This recording show how horizontally scroll looks like
https://user-images.githubusercontent.com/19235986/151130077-dbf4bde3-c8d2-486b-aab7-e30210ff1aed.mov

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -49,6 +50,34 @@ export class CompareRunView extends Component {
     Utils.updatePageTitle(pageTitle);
   }
 
+  adjustTableColumnWidth(numRuns) {
```
WeichenXu123, line 53: Why need this handler:

Previously the parameters / metrics table block, all columns are aligned exactly, and when page width changed, the columns will auto change size.

But, after add the vertical scroll bar to parameters/metrics block, things changed, the columns in the 2 blocks they are not aligned, and it cannot respond to the page size changing. Now we need additional js to do the work, i.e. this function.

WeichenXu123: https://user-images.githubusercontent.com/19235986/151131306-af5923ad-46f8-4f37-805c-e25c3827a925.mov

dbczumar (commented):
@WeichenXu123 This looks great! Before approving / merging, can we add horizontal and vertical scroll bars to the "Metrics" and "Parameters" sections? Right now, users have to click on the table cells and drag to move around. Scroll bars would help.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -230,6 +336,13 @@ export class CompareRunView extends Component {
     );
   }
 
+  switchNonDiffRowsDisplay(blockClass, showDiffOnly) {
+    const nonDiffRows = document.querySelectorAll(`.compare-table .${blockClass} .non-diff-row`);
+    for (let index = 0; index < nonDiffRows.length; ++index) {
+      nonDiffRows[index].style.display = showDiffOnly ? 'none' : '';
+    }
+  }
```
harupy, line 344: Instead of mutating the style, can we define a state and hide non-diff rows if it's true like:

```js
const rowsToRender = hideNonDiffRows ? rows.filter(...) : rows;
return render(rowsToRender)
```

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -246,41 +359,34 @@ export class CompareRunView extends Component {
     });
 
     return keys.map((k) => {
-      let row_class = undefined;
-      if (highlightChanges) {
-        const all_equal = data[k].every((x) => x === data[k][0]);
-        if (!all_equal) {
-          row_class = 'row-changed';
-        }
+      const all_equal = data[k].every((x) => x === data[k][0]);
```
harupy, line 362:
```suggestion
      const allEqual = data[k].every((x) => x === data[k][0]);
```

dbczumar (approved):
Looks fantastic! Thanks @WeichenXu123 !

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -68,6 +117,12 @@ export class CompareRunView extends Component {
       <Link to={Routes.getExperimentPageRoute(experimentId)}>{experiment.getName()}</Link>,
       title,
     ];
+
+    function onCollapsibleSectionChanged(blockClass, hideBlock) {
+      const blockElem = document.querySelectorAll(`.compare-table .${blockClass}`)[0];
+      blockElem.style.display = hideBlock ? 'none' : 'block';
+    }
```
harupy, line 124: Controlling visibility like this is not straightforward. Can we define a state, and toggle that state with a callback?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -271,41 +382,34 @@ export class CompareRunView extends Component {
     });
 
     return keys.map((k) => {
-      let row_class = undefined;
-      if (highlightChanges) {
-        const all_equal = data[k].every((x) => x === data[k][0]);
-        if (!all_equal) {
-          row_class = 'row-changed';
-        }
+      const all_equal = data[k].every((x) => x === data[k][0]);
```
harupy, line 385:
```suggestion
      const allEqual = data[k].every((x) => x === data[k][0]);
```

Can we use camel case for variable names?

harupy (changes requested):

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: I know setting a state is a more react-js code way. But seems it is not as efficient here. After state changed, the render will be called again to generate all virtual doms, and then we need call `adjustTableColumnWidth` method again to adjust all table cells width again. In this scenario directly setting the real DOM style is a more straightforward way.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -48,6 +48,55 @@ export class CompareRunView extends Component {
...
+    );
+
+    function onTableBlockScrollHanlder(e) {
+      const blocks = document.querySelectorAll('.compare-table .table-block');
+      for (let index = 0; index < blocks.length; ++index) {
+        const block = blocks[index];
+        if (block !== e.target) {
+          block.scrollLeft = e.target.scrollLeft;
+        }
+      }
```
harupy, line 65: `forEach` can simplify the code here.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -48,6 +48,55 @@ export class CompareRunView extends Component {
...
+    }
+    const blocks = document.querySelectorAll('.compare-table .table-block');
+    for (let index = 0; index < blocks.length; ++index) {
+      const block = blocks[index];
+      block.onscroll = onTableBlockScrollHanlder;
+    }
```
harupy, line 71: Same here (https://github.com/mlflow/mlflow/pull/5306/files#r794556512)

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -255,6 +359,13 @@ export class CompareRunView extends Component {
     );
   }
 
+  switchNonDiffRowsDisplay(blockClass, showDiffOnly) {
+    const nonDiffRows = document.querySelectorAll(`.compare-table .${blockClass} .non-diff-row`);
+    for (let index = 0; index < nonDiffRows.length; ++index) {
+      nonDiffRows[index].style.display = showDiffOnly ? 'none' : '';
+    }
```
harupy, line 366: Same here (https://github.com/mlflow/mlflow/pull/5306/files#r794556512)

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -48,6 +48,55 @@ export class CompareRunView extends Component {
...
+      true,
+    );
+
+    function onTableBlockScrollHanlder(e) {
+      const blocks = document.querySelectorAll('.compare-table .table-block');
+      for (let index = 0; index < blocks.length; ++index) {
+        const block = blocks[index];
+        if (block !== e.target) {
+          block.scrollLeft = e.target.scrollLeft;
+        }
```
harupy, line 64: Same here (https://github.com/mlflow/mlflow/pull/5306/files#r794556512)

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -135,48 +190,57 @@ export class CompareRunView extends Component {
...
                   {this.props.runInfos.map((r) => (
-                    <th scope='column' className='data-value' key={r.run_uuid}>
-                      <Link to={Routes.getRunPageRoute(r.getExperimentId(), r.getRunUuid())}>
-                        {r.getRunUuid()}
-                      </Link>
+                    <th scope='row' className='data-value' key={r.run_uuid}>
+                      <Tooltip
+                        title={r.getRunUuid()}
+                        color='blue'
+                        overlayStyle={{ 'max-width': '400px' }}
```
harupy, line 208:
```suggestion
                        overlayStyle={{ maxWidth: '400px' }}
```

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -135,48 +190,57 @@ export class CompareRunView extends Component {
...
-                          className='truncate-text single-line'
-                          style={styles.compareRunTableCellContents}
-                        >
-                          {runName}
+                      <td className='data-value' key={runInfos[i].run_uuid}>
+                        <div className='truncate-text single-line'>
+                          <Tooltip
+                            title={runName}
+                            color='blue'
+                            overlayStyle={{ 'max-width': '400px' }}
```
harupy, line 233:
```suggestion
                            overlayStyle={{ maxWidth: '400px' }}
```

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -188,43 +252,83 @@ export class CompareRunView extends Component {
                       ? Utils.formatTimestamp(run.getStartTime())
                       : '(unknown)';
                     return (
-                      <td className='meta-info' key={run.run_uuid}>
-                        {startTime}
+                      <td className='data-value' key={run.run_uuid}>
+                        <Tooltip
+                          title={startTime}
+                          color='blue'
+                          overlayStyle={{ 'max-width': '400px' }}
```
harupy, line 259:
```suggestion
                          overlayStyle={{ maxWidth: '400px' }}
```

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -48,6 +48,55 @@ export class CompareRunView extends Component {
...
     );
     Utils.updatePageTitle(pageTitle);
+    this.adjustTableColumnWidth(this.props.runInfos.length);
+    window.addEventListener(
+      'resize',
+      (e) => this.adjustTableColumnWidth(this.props.runInfos.length),
+      true,
+    );
+
+    function onTableBlockScrollHanlder(e) {
```
harupy, line 58:
```suggestion
    function onTableBlockScrollHandler(e) {
```

typo

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -48,6 +48,55 @@ export class CompareRunView extends Component {
...
+      const widthValue = `${width}px`;
+      for (let index = 0; index < cells.length; ++index) {
+        cells[index].style.width = widthValue;
+        cells[index].style.minWidth = widthValue;
+        cells[index].style.maxWidth = widthValue;
+      }
+    }
+    setWidth('head-value', colWidth);
+    setWidth('data-value', colWidth);
+    setWidth('compact-data-value', colWidth * numRuns);
```
harupy, line 98: Where do we use `compact-data-value`?

WeichenXu123: removed

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: Use forEach will cause lint error `Assignment to property of function parameter`. I don't want to break lint rule here

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -37,6 +37,40 @@ export class CompareRunView extends Component {
...
+      collapseParamBlock: false,
+      collapseMetricBlock: false,
+      onlyShowParamDiff: false,
+      onlyShowMetricDiff: false,
+    };
+    // onResizeHandler need access `this.state`, so bind `this`.
+    this.onResizeHandler = this.onResizeHandler.bind(this);
+    this.onTableBlockScrollHandler = this.onTableBlockScrollHandler.bind(this);
+
+    this.tableContainerRef = React.createRef();
```
WeichenXu123, line 53: Using this Ref, we don't need the `getElementById`

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -37,6 +37,40 @@ export class CompareRunView extends Component {
...
+  onResizeHandler(e) {
+    const container = this.tableContainerRef.current;
+    if (container !== null) {
+      const containerWidth = container.clientWidth;
+      this.setState({ tableContainerWidth: containerWidth });
+    }
+  }
+
+  onTableBlockScrollHandler(e) {
+    const blocks = this.tableContainerRef.current.querySelectorAll('.compare-table .table-block');
```
WeichenXu123, line 65: This is the only place I preserve the "querySelectorAll".
Because:
* We cannot set scrollbar offset via CSS. So we have to put it in the onScroll handler.
* We can use multiple react Refs to holds these table-blocks, but it make code more complex, using the selector might be the easiest way.
* I limit the `querySelectorAll` to select inside the `this.tableContainerRef` element so it will be safe.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -48,12 +82,48 @@ export class CompareRunView extends Component {
...
     Utils.updatePageTitle(pageTitle);
+
+    window.addEventListener('resize', this.onResizeHandler, true);
+    window.dispatchEvent(new Event('resize'));
+  }
+
+  componentWillUnmount() {
+    // Because onResizeHandler is registered in `componentDidMount`,
+    // We need unregister it here otherwise every time mount will add a new listener.
+    window.removeEventListener('resize', this.onResizeHandler, true);
```
WeichenXu123, line 93: A fix: we need removeEventListener if we install it in componentDidMount

mlflow/server/js/src/experiment-tracking/components/CompareRunView.css
```diff
@@ -1,12 +1,28 @@
-.compare-table th.row-header {
-  min-width: 120px;
+.sticky_header {
```
harupy, line 1:
```suggestion
.sticky-header {
```

Use the kebab style

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -188,65 +276,121 @@ export class CompareRunView extends Component {
...
+                        />
+                      }
+                      onChange={(activeKeys) =>
+                        this.setState({ collapseParamBlock: activeKeys.length === 0 })
+                      }
+                    >
+                      <Switch
+                        checkedChildren='Show diff only'
+                        unCheckedChildren='Show diff only'
+                        onChange={(isShowDiffOnly, e) =>
```
harupy, line 315:
```suggestion
                        onChange={(checked, e) =>
```

I prefer using `checked` here since it describes what this argument represents.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -37,6 +37,40 @@ export class CompareRunView extends Component {
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      tableContainerWidth: null,
+      collapseParamBlock: false,
+      collapseMetricBlock: false,
+      onlyShowParamDiff: false,
+      onlyShowMetricDiff: false,
+    };
+    // onResizeHandler need access `this.state`, so bind `this`.
```
harupy, line 49: we probably don't need this comment.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -37,6 +37,40 @@ export class CompareRunView extends Component {
...
+    const container = this.tableContainerRef.current;
+    if (container !== null) {
+      const containerWidth = container.clientWidth;
+      this.setState({ tableContainerWidth: containerWidth });
+    }
+  }
+
+  onTableBlockScrollHandler(e) {
+    const blocks = this.tableContainerRef.current.querySelectorAll('.compare-table .table-block');
+    for (let index = 0; index < blocks.length; ++index) {
```
harupy, line 66:
```suggestion
    for (let index = 0; index < blocks.length; ++index) {
```

Does `forEach((_, index) => {...})` work?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -188,65 +276,121 @@ export class CompareRunView extends Component {
...
+                    className='inter-title sticky_header'
                     colSpan={this.props.runInfos.length + 1}
+                    style={tableBlockWidthStyle}
                   >
-                    <h2>
-                      <FormattedMessage
-                        defaultMessage='Parameters'
-                        // eslint-disable-next-line max-len
-                        description='Row group title for parameters of runs on the experiment compare runs page'
+                    <CollapsibleSection
```
harupy, line 300: Why do we only wrap the Switch component with `CollapsibleSection`?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -37,6 +37,40 @@ export class CompareRunView extends Component {
     intl: PropTypes.shape({ formatMessage: PropTypes.func.isRequired }).isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      tableContainerWidth: null,
+      collapseParamBlock: false,
+      collapseMetricBlock: false,
```
harupy, line 45:
```suggestion
```

We probably don't need these two states once https://github.com/mlflow/mlflow/pull/5306/files#r795070842 is addressed.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -188,65 +276,121 @@ export class CompareRunView extends Component {
...
-                        <i className='fas fa-chart-line' style={{ paddingLeft: '6px' }} />
-                      </Link>
-                    );
-                  },
-                  Utils.formatMetric,
-                )}
+              </tbody>
+              <tbody
+                className='table-block metric-block'
+                style={Object.assign({}, tableBlockWidthStyle, { maxHeight: '300px' })}
```
harupy, line 365:
```suggestion
                style={{...tableBlockWidthStyle, maxHeight: '300px' )}
```

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -188,65 +276,121 @@ export class CompareRunView extends Component {
...
                       />
-                    </h2>
+                    </CollapsibleSection>
                   </th>
                 </tr>
-                {this.renderDataRows(this.props.paramLists, true)}
+              </tbody>
+              <tbody
+                className='table-block param-block'
+                style={Object.assign({}, tableBlockWidthStyle, { maxHeight: '500px' })}
```
harupy, line 325: https://github.com/mlflow/mlflow/pull/5306/files#r795071670

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -188,65 +276,121 @@ export class CompareRunView extends Component {
...
+                          defaultMessage='Parameters'
+                          // eslint-disable-next-line max-len
+                          description='Row group title for parameters of runs on the experiment compare runs page'
+                        />
+                      }
+                      onChange={(activeKeys) =>
+                        this.setState({ collapseParamBlock: activeKeys.length === 0 })
+                      }
+                    >
+                      <Switch
```
harupy, line 312: Why is `Switch` in `tr`? Does this component have to be part of the table?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -258,54 +402,66 @@ export class CompareRunView extends Component {
...
+      width: `${colWidth}px`,
+      minWidth: `${colWidth}px`,
+      maxWidth: `${colWidth}px`,
+    };
+
+    function isAllEqual(k) {
+      return data[k].every((x) => x === data[k][0]);
+    }
+
+    if (this.state[onlyShowDiffStateKey]) {
```
harupy, line 429: I think accessing a state like this makes the code harder to read. If I was you, I'd refactor this function to the following:


```javascript
  getData(list) {
    const keys = CompareRunUtil.getKeys(list);
    const data = {};
    keys.forEach((k) => (data[k] = []));
    list.forEach((records, i) => {
      keys.forEach((k) => data[k].push(undefined));
      records.forEach((r) => (data[r.key][i] = r.value));
    });
    const isAllEqual = (arr) => arr.every((x) => x === arr[0]);
    return Object.entries(data).reduce(
      (obj, [key, values]) => ({ ...obj, [key]: { values, hasDiff: !isAllEqual(values) } }),
      {},
    );
  }

  getWidthStyle = (width) => ({
    width: `${width}px`,
    minWidth: `${width}px`,
    maxWidth: `${width}px`,
  });

  // eslint-disable-next-line no-unused-vars
  renderDataRows(
    list,
    colWidth,
    showDiffOnly = false,
    highlightDiff = false,
    headerMap = (key, data) => key,
    formatter = (value) => value,
  ) {
    const data = this.getData(list);
    const colWidthStyle = this.getWidthStyle(colWidth);
    return Object.entries(data)
      .filter((_, { hasDiff }) => !showDiffOnly || hasDiff)
      .map(([key, { values, hasDiff }]) => {
        let rowClass = undefined;
        if (highlightDiff && hasDiff) {
          rowClass = 'diff-row';
        }

        return (
          <tr key={key} className={rowClass}>
            <th scope='row' className='head-value sticky_header' style={colWidthStyle}>
              {headerMap(key, values)}
            </th>
            {values.map((value, i) => {
              const cellText = value === undefined ? '' : formatter(value);
              return (
                <td
                  className='data-value'
                  key={this.props.runInfos[i].run_uuid}
                  style={colWidthStyle}
                >
                  <Tooltip title={cellText} color='blue' overlayStyle={{ maxWidth: '400px' }}>
                    <span className='truncate-text single-line'>{cellText}</span>
                  </Tooltip>
                </td>
              );
            })}
          </tr>
        );
      });
  }
```

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -135,48 +206,65 @@ export class CompareRunView extends Component {
...
                       description='Row title for the run id on the experiment compare runs page'
                     />
                   </th>
                   {this.props.runInfos.map((r) => (
-                    <th scope='column' className='data-value' key={r.run_uuid}>
-                      <Link to={Routes.getRunPageRoute(r.getExperimentId(), r.getRunUuid())}>
-                        {r.getRunUuid()}
-                      </Link>
+                    <th scope='row' className='data-value' key={r.run_uuid} style={colWidthStyle}>
+                      <Tooltip
```
harupy, line 225: Instead of `Tooltip`, can we use [the title attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)?

https://user-images.githubusercontent.com/17039389/151673078-4bd6e8ed-9198-41c4-b41e-d2553e11e430.mov

harupy: `Tooltip` always instantly shows up when you hover on the cell, but I feel like this behavior unnecessarily draws the user's attention, especially when the cell content is NOT truncated.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -48,12 +82,48 @@ export class CompareRunView extends Component {
...
     );
     Utils.updatePageTitle(pageTitle);
+
+    window.addEventListener('resize', this.onResizeHandler, true);
+    window.dispatchEvent(new Event('resize'));
+  }
+
+  componentWillUnmount() {
+    // Because onResizeHandler is registered in `componentDidMount`,
+    // We need unregister it here otherwise every time mount will add a new listener.
```
harupy, line 92:
```suggestion
    // Avoid registering `onResizeHandler` every time this component mounts
```

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: Then we have to split the whole table to be 3 tables (run-info table, param table, metric table), now they are in a table.

WeichenXu123: Shall we split ? feel like kind of big change but I can do that.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: Does  'title' support only display when truncated ?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
harupy: I think that's simpler.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: +1

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: @dbczumar 
Do you prefer use "title" or "Tooltip" ?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: The refactor code seems buggy on "only show diff" case. 
But I update to avoid access state inside the function.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: i remember corey prefer "blue" background for the hovor text. So I keep tooltip. But I increase the mouseEnterDelay to be 1.0 second (i.e. only mouse hover above 1 second, then the tooltip will show)

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -258,54 +377,71 @@ export class CompareRunView extends Component {
...
+      keys = keys.filter((k) => !isAllEqual(k));
+    }
     return keys.map((k) => {
-      let row_class = undefined;
-      if (highlightChanges) {
-        const all_equal = data[k].every((x) => x === data[k][0]);
-        if (!all_equal) {
-          row_class = 'row-changed';
-        }
+      const allEqual = isAllEqual(k);
```
harupy, line 408: can we avoid calling `isAllEqual` twice?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -188,45 +265,87 @@ export class CompareRunView extends Component {
...
+          }
+        >
+          <Switch
+            checkedChildren='Show diff only'
+            unCheckedChildren='Show diff only'
+            onChange={(checked, e) => this.setState({ onlyShowParamDiff: checked })}
+          />
+          <br />
+          <br />
+          <div className='responsive-table-container'>
```
harupy, line 303: Do we need to wrap all the tables with this div?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -37,6 +37,39 @@ export class CompareRunView extends Component {
...
+    const container = this.tableContainerRef.current;
+    if (container !== null) {
+      const containerWidth = container.clientWidth;
+      this.setState({ tableContainerWidth: containerWidth });
+    }
+  }
+
+  onCompareRunTableScrollHandler(e) {
+    const blocks = this.compareRunViewRef.current.querySelectorAll('.compare-run-table');
+    for (let index = 0; index < blocks.length; ++index) {
```
harupy, line 65: Does forEach((_, index) => {...}) work?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: delete this div. But I don't delete `responsive-table-container`, this css class is used in other pages.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: works.

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
WeichenXu123: yes. we can simply return `null` in the forEach function for hidden rows.

harupy:
Found a minor issue:


https://user-images.githubusercontent.com/17039389/151800585-7bc6d884-bd8c-4712-9834-f394ea5a2716.mov

The tooltip shows up outside of the cell.

harupy:
This is not a big issue, but if the background color of the tooltip is blue, it conflicts with the text Selection color. Is it possible to use a different text selection color?

https://user-images.githubusercontent.com/17039389/151802985-1b4c65c1-7010-4dc5-a4b0-e3723c10170e.mov

mlflow/server/js/src/experiment-tracking/components/CompareRunView.js
```diff
@@ -135,121 +198,172 @@ export class CompareRunView extends Component {
...
+                    </td>
+                  );
+                })}
+              </tr>
+            </tbody>
+          </table>
+        </CollapsibleSection>
+        <CollapsibleSection
+          title={
+            <h1>
```
harupy, line 286:
```suggestion
            <h1 style={{ marginTop: 0, marginBottom: 0 }}>
```

Can we remove vertical margins to align the arrow icon and title?

BEFORE

Screen Shot 2022-01-31 at 22 53 03

AFTER

Screen Shot 2022-01-31 at 22 52 40

mlflow/server/js/src/experiment-tracking/components/CompareRunView.css
```diff
@@ -1,12 +1,29 @@
...
+  left: 0;
+}
+
+.compare-run-table {
+  display: block;
+  overflow: auto;
+  width: 100%;
+}
+
+.no-scrollbar::-webkit-scrollbar {
```
harupy, line 13: Is this class used?

WeichenXu123: No use now. removed.

harupy (approved):
LGTM

mlflow/server/js/src/experiment-tracking/components/CompareRunView.css
```diff
@@ -19,16 +32,33 @@
...
   width: 100%;
   max-width: 100%;
   margin-bottom: 20px;
+
+  /* The border-collapse and border-spacing setting is to make header column "position sticky"
+   * works correctly with the border-bottom style.
+   * https://stackoverflow.com/a/61412922/12301610
+   */
+  border-collapse: separate;
+  border-spacing: 0;
```
harupy, line 61: What would happen if we removed these?

harupy: I commented out these styles but didn't see the difference.

WeichenXu123: It is a small style issue. without this some border lines looks thicker.

WeichenXu123: But added this, in chrome (only in chrome, might be bug), there's also a bug display header column: 
image

WeichenXu123: Or we can hide header column border to avoid this issue
image

harupy: image

harupy:
why did you remove `border-collapse: separate; border-spacing: 0;`?

mlflow/server/js/src/experiment-tracking/components/CompareRunView.css
WeichenXu123: emm， interesting

WeichenXu123: offline discussion, remove this because border line thicker is a tiny issue from customer side.

Issue #5308, opened 2022-01-25
[BUG] ONNX deployment on AzureML not working with onnxruntime > v1.9.0


ecm200:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for additional information about bug reports. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

**Please fill in this bug report template to ensure a timely and thorough response.**

### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [x] Yes. I can contribute a fix for this bug independently.
- [x] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Ubuntu 18.04
- **MLflow installed from (source or binary)**: Both source and binaries.
- **MLflow version (run ``mlflow --version``)**:  1.18.0 and 1.23.0 (latest
- **Python version**: 3.8.12
- **npm version, if running the dev UI**:
- **Exact command to reproduce**:

### Describe the problem

I have saved a PyTorch Detectron 2 Object Detection model as an ONNX format, and registered it with the MLFlow tracking serving as an artifact. I did this using the `mlflow.onnx.log_model()` functionality, so the model artifact was uploaded as a MLFlow pyfunc type model. This process happens correctly and registers a model as expected, with the following meta data in the `MLModel` file:

```yaml
artifact_path: model/tracing_detectron2_mask_rcnn_R_50_FPN_1X__COCO_trained__model_zoo
flavors:
  onnx:
    data: model.onnx
    onnx_version: 1.10.2
  python_function:
    data: model.onnx
    env: conda.yaml
    loader_module: mlflow.onnx
    python_version: 3.8.12
run_id: 205584e9973d4077822e5a333fa4aec8
signature:
  inputs: '[{"name": "data", "type": "tensor", "tensor-spec": {"dtype": "int8", "shape":
    [3, -1, -1]}}]'
  outputs: '[{"name": "bboxes", "type": "tensor", "tensor-spec": {"dtype": "float32",
[...]

PR #5317, opened 2022-01-26
onnxruntime InferenceSession fix to bug #5308

ecm200:
Some distributions of onnxruntime require the specification of the providers argument on calling
InferenceSession. E.g. onnxruntime-gpu.
The package import call does not differnetiate which architecture specific version has been installed,
as all are imported with onnxruntime.
onnxruntime documentation says that
from v1.9.0 some distributions require the providers list to be
provided on calling an InferenceSession.
Therefore the try catch nested structure below attempts to create an inference session
with just the model path as pre v1.9.0.
If that fails, it will use the providers list
 as part of the InferenceSession call.
At the moment this is just CUDA and CPU, and probably
should be expanded.

Fix for bug #5308 

In the `onnx.py` module, the class `_OnnxModelWrapper` init method has been adapted to cater for the changes in how an ONNX model is loaded into an InferenceSession. Previously to v1.9.0 of onnxruntime, only the path to the model file was needed to start an inference session, however post v1.9.0, some builds require the explicit passing of a providers list.

The original implementation of loading the onnx model was:

```python
self.rt = onnxruntime.InferenceSession(path)
```

The following code shows the updated version, which if the original call fails, it falls back to providing a list of executor providers:

```python

try:
  self.rt = onnxruntime.InferenceSession(path)
except:
  try:
      self.rt = onnxruntime.InferenceSession(path, providers=providers)
  except:
      raise
```

## How is this patch tested?

I have tested this on pre and post v1.9.0 versions of the onnxruntime and onnxruntime-gpu packages, and this works in all instances.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations
- [x] `integrations/sagemaker`: SageMaker integrations

### DIFF
--- a/mlflow/onnx.py
+++ b/mlflow/onnx.py
@@ -40,7 +40,7 @@ from mlflow.utils.model_utils import _get_flavor_configuration
 from mlflow.tracking._model_registry import DEFAULT_AWAIT_MAX_SLEEP_SECONDS
 
 FLAVOR_NAME = "onnx"
-ONNX_EXECUTION_PROVIDERS = ['CUDAExecutionProvider', 'CPUExecutionProvider']
+ONNX_EXECUTION_PROVIDERS = ["CUDAExecutionProvider", "CPUExecutionProvider"]
 
 
 def get_default_pip_requirements():
@@ -148,7 +148,12 @@ def save_model(
     pyfunc.add_to_model(
         mlflow_model, loader_module="mlflow.onnx", data=model_data_subpath, env=_CONDA_ENV_FILE_NAME
     )
-    mlflow_model.add_flavor(FLAVOR_NAME, onnx_version=onnx.__version__, data=model_data_subpath, providers=onnx_execution_providers)
+    mlflow_model.add_flavor(
+        FLAVOR_NAME,
+        onnx_version=onnx.__version__,
+        data=model_data_subpath,
+        providers=onnx_execution_providers,
+    )
     mlflow_model.save(os.path.join(path, MLMODEL_FILE_NAME))
 
     if conda_env is None:
@@ -201,23 +206,23 @@ class _OnnxModelWrapper:
         model_meta = Model.load(os.path.join(local_path, MLMODEL_FILE_NAME))
 
         # Check if the MLModel config has the providers meta data
-        if 'providers' in model_meta.flavors.get(FLAVOR_NAME).keys():
-            providers = model_meta.flavors.get(FLAVOR_NAME)['providers']
+        if "providers" in model_meta.flavors.get(FLAVOR_NAME).keys():
+            providers = model_meta.flavors.get(FLAVOR_NAME)["providers"]
         # If not, then default to the predefined list.
         else:
             providers = ONNX_EXECUTION_PROVIDERS
 
-        # NOTE: Some distributions of onnxruntime require the specification of the providers 
+        # NOTE: Some distributions of onnxruntime require the specification of the providers
         # argument on calling. E.g. onnxruntime-gpu. The package import call does not differnetiate
-        #  which architecture specific version has been installed, as all are imported with 
+        #  which architecture specific version has been installed, as all are imported with
         # onnxruntime. onnxruntime documentation says that from v1.9.0 some distributions require
         #  the providers list to be provided on calling an InferenceSession. Therefore the try
         #  catch structure below attempts to create an inference session with just the model path
-        #  as pre v1.9.0. If that fails, it will use the providers list call. 
-        # At the moment this is just CUDA and CPU, and probably should be expanded. 
+        #  as pre v1.9.0. If that fails, it will use the providers list call.
+        # At the moment this is just CUDA and CPU, and probably should be expanded.
         # A method of user customisation has been provided by adding a variable in the save_model()
-        # function, which allows the ability to pass the list of execution providers via a 
-        # optional argument e.g. 
+        # function, which allows the ability to pass the list of execution providers via a
+        # optional argument e.g.
         #
         # mlflow.onnx.save_model(..., providers=['CUDAExecutionProvider'...])
         #
@@ -235,7 +240,6 @@ class _OnnxModelWrapper:
             self.rt = onnxruntime.InferenceSession(path, providers=providers)
         except:
             raise
-
 
         assert len(self.rt.get_inputs()) >= 1
         self.inputs = [(inp.name, inp.type) for inp in self.rt.get_inputs()]
@@ -378,7 +382,7 @@ def log_model(
     await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
     pip_requirements=None,
     extra_pip_requirements=None,
-    onnx_execution_providers=None
+    onnx_execution_providers=None,
 ):
     """
     Log an ONNX model as an MLflow artifact for the current run.
@@ -433,5 +437,5 @@ def log_model(
         await_registration_for=await_registration_for,
         pip_requirements=pip_requirements,
         extra_pip_requirements=extra_pip_requirements,
-        onnx_execution_providers=onnx_execution_providers
+        onnx_execution_providers=onnx_execution_providers,
     )
--- a/mlflow/onnx.py
+++ b/mlflow/onnx.py
@@ -116,7 +116,7 @@ def save_model(
     :param extra_pip_requirements: {{ extra_pip_requirements }}
     :param onnx_execution_providers: List of strings defining onnxruntime execution providers.
                                      Defaults to example:
-                                     ['CUDAExecutionProvider', 'CPUExecutionProvider']
+                                     ``['CUDAExecutionProvider', 'CPUExecutionProvider']``
                                      This uses GPU preferentially over CPU.
                                      See onnxruntime API for further descriptions.
     """
--- a/mlflow/onnx.py
+++ b/mlflow/onnx.py
@@ -182,10 +182,27 @@ def _load_model(model_file):
 
 
 class _OnnxModelWrapper:
-    def __init__(self, path):
+    def __init__(self, path, providers=None):
         import onnxruntime
 
-        self.rt = onnxruntime.InferenceSession(path)
+        if providers is None:
+            providers = ['CUDAExecutionProvider', 'CPUExecutionProvider']
+
+        # NOTE: Some distributions of onnxruntime require the specification of the providers argument on calling. E.g. onnxruntime-gpu.
+        #       The package import call does not differnetiate which architecture specific version has been installed, as all are imported with onnxruntime.
+        #       onnxruntime documentation says that from v1.9.0 some distributions require the providers list to be provided on calling an InferenceSession.
+        #       Therefore the try catch nested structure below attempts to create an inference session with just the model path as pre v1.9.0.
+        #       If that fails, it will use the providers list call. 
+        #       At the moment this is just CUDA and CPU, and probably should be expanded. It's been added to the class init call as an optional argument for further user customisation.
+        try:
+            self.rt = onnxruntime.InferenceSession(path)
+        except:
+            try:
+                self.rt = onnxruntime.InferenceSession(path, providers=providers)
+            except:
+                raise
+
+
         assert len(self.rt.get_inputs()) >= 1
         self.inputs = [(inp.name, inp.type) for inp in self.rt.get_inputs()]
         self.output_names = [outp.name for outp in self.rt.get_outputs()]
--- a/mlflow/onnx.py
+++ b/mlflow/onnx.py
@@ -40,7 +40,7 @@ from mlflow.utils.model_utils import _get_flavor_configuration
 from mlflow.tracking._model_registry import DEFAULT_AWAIT_MAX_SLEEP_SECONDS
 
 FLAVOR_NAME = "onnx"
-ONNX_EXECUTION_PROVIDERS = ['CUDAExecutionProvider', 'CPUExecutionProvider']
+ONNX_EXECUTION_PROVIDERS = ["CUDAExecutionProvider", "CPUExecutionProvider"]
 
 
 def get_default_pip_requirements():
@@ -149,7 +149,12 @@ def save_model(
     pyfunc.add_to_model(
         mlflow_model, loader_module="mlflow.onnx", data=model_data_subpath, env=_CONDA_ENV_FILE_NAME
     )
-    mlflow_model.add_flavor(FLAVOR_NAME, onnx_version=onnx.__version__, data=model_data_subpath, providers=onnx_execution_providers)
+    mlflow_model.add_flavor(
+        FLAVOR_NAME,
+        onnx_version=onnx.__version__,
+        data=model_data_subpath,
+        providers=onnx_execution_providers,
+    )
     mlflow_model.save(os.path.join(path, MLMODEL_FILE_NAME))
 
     if conda_env is None:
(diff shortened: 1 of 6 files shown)


mlflow/onnx.py
```diff
@@ -182,10 +182,27 @@ def _load_model(model_file):
...
+
+        # NOTE: Some distributions of onnxruntime require the specification of the providers argument on calling. E.g. onnxruntime-gpu.
+        #       The package import call does not differnetiate which architecture specific version has been installed, as all are imported with onnxruntime.
+        #       onnxruntime documentation says that from v1.9.0 some distributions require the providers list to be provided on calling an InferenceSession.
+        #       Therefore the try catch nested structure below attempts to create an inference session with just the model path as pre v1.9.0.
+        #       If that fails, it will use the providers list call. 
+        #       At the moment this is just CUDA and CPU, and probably should be expanded. It's been added to the class init call as an optional argument for further user customisation.
+        try:
+            self.rt = onnxruntime.InferenceSession(path)
+        except:
```
BenWilson2, line 199: Could you catch the appropriate exception here to initiate the alternate processing logic?

mlflow/onnx.py
```diff
@@ -182,10 +182,27 @@ def _load_model(model_file):
...
 class _OnnxModelWrapper:
-    def __init__(self, path):
+    def __init__(self, path, providers=None):
         import onnxruntime
 
-        self.rt = onnxruntime.InferenceSession(path)
+        if providers is None:
+            providers = ['CUDAExecutionProvider', 'CPUExecutionProvider']
+
+        # NOTE: Some distributions of onnxruntime require the specification of the providers argument on calling. E.g. onnxruntime-gpu.
```
BenWilson2, line 191: Please adjust line length to 100 chars max.

mlflow/onnx.py
```diff
@@ -182,10 +182,27 @@ def _load_model(model_file):
...
+        #       The package import call does not differnetiate which architecture specific version has been installed, as all are imported with onnxruntime.
+        #       onnxruntime documentation says that from v1.9.0 some distributions require the providers list to be provided on calling an InferenceSession.
```
BenWilson2, line 193: Please add link in the docstring to their docs that discusses these providers and how they are defined / used

mlflow/onnx.py
```diff
@@ -182,10 +182,27 @@ def _load_model(model_file):
 
 
 class _OnnxModelWrapper:
-    def __init__(self, path):
+    def __init__(self, path, providers=None):
```
BenWilson2, line 185: for this to be accessible to an end-user, the `_load_pyfunc()` function would need to be updated to support loading with a particular intended end-use environment provider configuration which would in turn change the signature of `mlflow.pyfunc.load_model()` to make this a modifiable argument. 
This is probably not what we want to do here. 
We might want to think about setting this property for an onnx flavor in a different way (during logging of the model and have this property stored in the serialized artifact?) 
WDYT @harupy @dbczumar ?

ecm200: @BenWilson2 completely understand the reservations about changing the `mlflow.pyfunc.load_model()` functionality given it's application to all model flavours.

I think your suggestion about modifying the saving metadata information on saving of the model is quite a nice suggestion as this means that only the `onnx.py` module code would need to be changed, is that right?

So under the `model_flavours` section of the `MLmodel` yaml:

```yaml
flavors:
  onnx:
    data: model.onnx
    onnx_version: 1.10.2
    providers: ['CUDAExecutionProvider', CPUExecutionProvider']
```

From my assessment, that would be facilitated through an additional argument added to the `add_flavour` method call in line 140 of `onnx.py`:

From:

`mlflow_model.add_flavor(FLAVOR_NAME, onnx_version=onnx.__version__, data=model_data_subpath)`

to:

`mlflow_model.add_flavor(FLAVOR_NAME, onnx_version=onnx.__version__, data=model_data_subpath, providers=providers)`

Could the variable `providers` be an additional optional argument for the `save_model` function, giving the user the ability to specify their own preference of providers (with a default list if not specified)?

```python
def save_model(
    onnx_model,
    path,
    conda_env=None,
    mlflow_model=None,
    pip_requirements=None,
    extra_pip_requirements=None,
    providers=None
):
   if not providers:
      providers = ['CUDAExecutionProvider', 'CPUExecutionProvider']

```

mlflow/onnx.py
ecm200: So, if I understand you correctly, catch the exception explicitly by type and then call the alternative loading, instead of a nested try implementation that I have done? Did I understand you right?

In the case where the `providers` variable is needed, the exception thrown is a `ValueError`.

So the refactored version would look as follows:

```python
try:
   self.rt = onnxruntime.InferenceSession(path)
except ValueError:
   self.rt = onnxruntime.InferenceSession(path, providers=providers)
except:
   raise
```

mlflow/onnx.py
ecm200: Will gladly adjust in the morning and submit commits to update this.

mlflow/onnx.py
ecm200: Will gladly add in the morning and submit commits to update this.

mlflow/onnx.py
BenWilson2: exactly! Just so that we can make sure we're only trying the other logic in the case of the exact type of error that we're expecting (Sorry, I was hoping that the onnx package had a custom exception other than the blanket ValueError, but it's generally a good practice to catch only the Exceptions that we want to catch to do any further processing).

mlflow/onnx.py
dbczumar: @ecm200 @BenWilson2 This proposal sounds great to me!

mlflow/onnx.py
```diff
@@ -182,10 +182,27 @@ def _load_model(model_file):
...
 
-        self.rt = onnxruntime.InferenceSession(path)
+        if providers is None:
+            providers = ['CUDAExecutionProvider', 'CPUExecutionProvider']
+
+        # NOTE: Some distributions of onnxruntime require the specification of the providers argument on calling. E.g. onnxruntime-gpu.
+        #       The package import call does not differnetiate which architecture specific version has been installed, as all are imported with onnxruntime.
+        #       onnxruntime documentation says that from v1.9.0 some distributions require the providers list to be provided on calling an InferenceSession.
+        #       Therefore the try catch nested structure below attempts to create an inference session with just the model path as pre v1.9.0.
+        #       If that fails, it will use the providers list call.
```
dbczumar, line 195: @ecm200 Would it be possible to always instantiate the InferenceSession with `providers`, or should `providers` only be used as a fallback?

mlflow/onnx.py
ecm200: @dbczumar @BenWilson2 

Great. I’ll draft a change and commit and update the PR.

Only thing I need to figure is how to get the info into the ONNX private class that handles the inference session setup. 

My proposal would be that given this class gets the path to the onnx model, then the expectation is that this model will always be an MLModel type format? Therefore we can expect to have an MLModel yaml file located in the same directory? Is that a correct assumption?

If that’s a reasonable assumption then the `__init__` method of the aforementioned class could simply load the model meta data which would have the `providers` parameters set. For backwards comparability, where the MLModel file does not have a entry for `providers` then the fallback option could be to have a default list which covers a reasonable range of execution types.

mlflow/onnx.py
ecm200: @dbczumar That’s a very good point and only occurred to me last night.

Certainly v1.8.0 does have the variable but it is not required. I am not sure how far back the variable is present in the `onnxruntime` releases before that.

Potentially I think it’s easier to define it this way, as it’s a clearly defined breaking change in the `onnxruntime` documentation.

I suppose we could just pin to a minimum version of `onnxruntime` to `mlflow`?

So I’m erring to the current implementation but happy to do what you admins think is more appropriate or online with `mlflow` style and policy.

dbczumar: @ecm200 Great point! In the interest of supporting a broader set of onnxruntime versions, I'm also in favor of the current implementation. Thank you for the detailed explanation.

mlflow/onnx.py
ecm200: @BenWilson2 @dbczumar 

Refactor committed with the suggested changes catching the ValueError and then running the post v1.9.0 call with providers.

Added functionality in the `__init__` method of the `_OnnxModelWrapper` class. 
This now reads the `MLModel` file from the same directory as the `model.onnx` has been downloaded to. **The assumption is that the MLModel file will always be present, as the model will be logged using the onnx flavour**.
It attempts to find the field "providers" in the onnx model flavour fields of the MLModel yaml.
If does not find it (for backwards compatability), it uses the default list which is hardwired as an attribute `ONNX_EXECUTION_PROVIDES` of the onnx model flavour.

mlflow/onnx.py
ecm200: On new commit.
Text notes have been limited to 100 characters and rewritten to according to the altered implementation.

mlflow/onnx.py
ecm200: Links have been added to the `onnxruntime` descriptions of execution providers and to the specific API description of the `InferenceSession` class.

mlflow/onnx.py
ecm200: @dbczumar  New commit implements refactored approach as discussed.

mlflow/onnx.py
ecm200: New commit af21a11 has this implemented.
See below commetns.

ecm200:
@BenWilson2 @dbczumar 

After commit af21a11.

Updated the `log_model` and `save_model` functions of the `mlflow.onnx` model flavor.

The `onnx_execution_providers` argument was passed through the `onnx.pyfunc.log_model` function by using the kwargs structure.

The `save_model` function of `mlflow.onnx` was updated to handle the additional argument. If the `onnx_execution_providers` variable remains uninitiated (None), then the variable is initialised with the `ONNX_EXECUTION_PROVIDERS` module parameter with default GPU then CPU settings.

The additional `providers` variable is supplied to the `onnx.Model.add_flavor` and appears in the resulting saved `MLModel` yaml configuration file for the uploaded model. It saves as follows:

```yaml
artifact_path: model/tracing_detectron2_mask_rcnn_R_50_FPN_1X__COCO_trained__model_zoo
flavors:
  onnx:
    data: model.onnx
    onnx_version: 1.8.1
    providers:
    - CUDAExecutionProvider
    - CPUExecutionProvider
  python_function:
    data: model.onnx
    env: conda.yaml
    loader_module: mlflow.onnx
    python_version: 3.8.12
model_uuid: 3efce24bba61429aa9543a6f48bff00b
run_id: 06491f3c2c1d422d98c0d45ddc75ffb0
signature:
```

The MLModel file has been successfully read in as part of the `mlflow.onnx` model loading function, and verified as picking up the `providers` list from the MLModel yaml file.

BenWilson2:
Hi @ecm200 could you sign your commit (It's the DCO check that's failing in the Github Actions CI checks)?
You can set up your local git to automatically apply signed commits. See: [the docs on signing commits](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst#sign-your-work) for more information and how to do this. 
For the currently running CI check, you can manually affirm that you're the author by click on that failed task to sign the commit being tested.

mlflow/onnx.py
```diff
@@ -111,9 +114,17 @@ def save_model(
                           by converting it to a list. Bytes are base64-encoded.
     :param pip_requirements: {{ pip_requirements }}
     :param extra_pip_requirements: {{ extra_pip_requirements }}
+    :param onnx_execution_providers: List of strings defining onnxruntime execution providers.
+                                     Defaults to example:
+                                     ['CUDAExecutionProvider', 'CPUExecutionProvider']
```
BenWilson2, line 119: For docs rendering
```suggestion
                                     ``['CUDAExecutionProvider', 'CPUExecutionProvider']``
```

mlflow/onnx.py
```diff
@@ -111,9 +114,17 @@ def save_model(
...
+                                     This uses GPU preferentially over CPU.
+                                     See onnxruntime API for further descriptions.
```
BenWilson2, line 121:
```suggestion
                                     See onnxruntime API for further descriptions:
                                     https://onnxruntime.ai/docs/execution-providers/
```

mlflow/onnx.py
```diff
@@ -362,6 +414,11 @@ def log_model(
                             waits for five minutes. Specify 0 or None to skip waiting.
     :param pip_requirements: {{ pip_requirements }}
     :param extra_pip_requirements: {{ extra_pip_requirements }}
+    :param onnx_execution_providers: List of strings defining onnxruntime execution providers.
+                                     Defaults to example:
+                                     ['CUDAExecutionProvider', 'CPUExecutionProvider']
+                                     This uses GPU preferentially over CPU.
+                                     See onnxruntime API for further descriptions.
```
BenWilson2, line 421:
```suggestion
                                     See onnxruntime API for further descriptions:
                                     https://onnxruntime.ai/docs/execution-providers/
```

BenWilson2:
Once the linting is fixed and the commit is signed off LGTM!

ecm200:
@BenWilson2 @dbczumar 

I have tried retrospectively signing the unsigned commits, but it doesn't appear to have worked with all of them.

You mention an override button, but I believe I need write access to the repository in order for this to be shown.
I think this might be the easiest solution to fix it, but I can't access that button to do it unless I get write access.
Github docs say:

>Additionally, the DCO creates an override button accessible to only those with write access to the repository to create a successful check.

image

mlflow/onnx.py
```diff
@@ -182,10 +199,49 @@ def _load_model(model_file):
...
+        # see the API page below:
+        # https://onnxruntime.ai/docs/api/python/api_summary.html#id8
+        #
+
+        try:
+            self.rt = onnxruntime.InferenceSession(path)
+        except ValueError:
+            self.rt = onnxruntime.InferenceSession(path, providers=providers)
+        except:
+            raise
```
BenWilson2, line 243: We don't need this final except catch block. If anything other than a ValueError raises, it'll throw an Exception normally. 
All that's needed here is:

```
try:
    self.rt = onnxruntime.InferenceSession(path)
except ValueError:
    self.rt = onnxruntime.InferenceSession(path, providers=providers)
```
Since we're catching the ValueError and handling it with a retry, we should be good to go with just that logic.

mlflow/onnx.py
```diff
@@ -182,10 +199,49 @@ def _load_model(model_file):
 
 
 class _OnnxModelWrapper:
-    def __init__(self, path):
+    def __init__(self, path, providers=None):
         import onnxruntime
 
-        self.rt = onnxruntime.InferenceSession(path)
+        # Get the model meta data from the MLModel yaml file which may contain the providers specification
```
BenWilson2, line 205:
```suggestion
        # Get the model meta data from the MLModel yaml file which may contain the 
        # providers specification
```
To pacify the lint requirement of 100 lines

BenWilson2:
autoformat

BenWilson2 (approved):
LGTM! 
Thank you very much for the contribution and I hope it's the first of many!

ecm200:
> LGTM!
> Thank you very much for the contribution and I hope it's the first of many!

Happy to help and give back a little.

Do I need to do anything further to finish the PR?

I do have some thoughts on the ‘mlflow.azureml’ module and for potential to extend the functionality of the boilerplate entry script so that it supports more than just pandas dataframe tables, I’d like it to support the ‘TFServing’ inferencing format for Tensor type data so that it supports image data agnostically.

I’ve had some success editing the ‘mlflow.azureml.__init__.py’ file which contains the boilerplate entry script text as a string variable at the bottom of the file (line 657) . Replacing ‘parse_json_input’ with ‘ infer_and_parse_json_input’ would potentially go some of the way to solving that issue.

Should I scope some changes and put in another PR?

BenWilson2:
For the ideas that you mentioned, feel free to create issues with feature requests (even if you're the one who will be working on them, it's a good medium for getting feedback / share design thoughts / canvas for other ideas in implementation). After the issue has received some comments then you can work on implementations :) 
The only thing left for this one is to merge it.

ecm200:
> For the ideas that you mentioned, feel free to create issues with feature requests (even if you're the one who will be working on them, it's a good medium for getting feedback / share design thoughts / canvas for other ideas in implementation). After the issue has received some comments then you can work on implementations :) The only thing left for this one is to merge it.

@BenWilson2 this might be a dumb question, but I'll ask it just in case. 

I haven't actually done a PR into a repository I cannot edit before, so I am assuming the merge process is not on me, but either someone who does, or it is part of the CI process?

BenWilson2:
I was in a meeting ;) 
It's merged now! Thanks again!

PR #5347, opened 2022-02-04
Add warning in MLflow pytorch docs to include signature

mehtayogita:

Updates MLFlow pytorch documentation to add warning suggesting to add signature while logging model to avoid float precision errors. 

## How is this patch tested?

Building docs locally and verifying the change.

- [X] Yes. Make sure the changed pages / sections render correctly by following the steps below.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Updates the MLFlow pytorch documentation.

Components 
- [X] `area/docs`: MLflow documentation pages

Interface 

Language 

Integrations


mehtayogita:
autoformat

mehtayogita:
autoformat

mehtayogita:
autoformat

mlflow/pytorch/__init__.py
```diff
@@ -128,6 +128,13 @@ def log_model(
     """
     Log a PyTorch model as an MLflow artifact for the current run.
 
+        .. warning::
+
+            Log the model with signature to avoid inference errors. Pytorch float precision default
```
andreakress, line 133:
```suggestion
            Log the model with signature to avoid inference errors. PyTorch float precision default
```

mlflow/pytorch/__init__.py
```diff
@@ -128,6 +128,13 @@ def log_model(
...
+            is float32, while numpy float precision default is float64. Adding the signature will
```
andreakress, line 134:
```suggestion
            is float32, while NumPy float precision default is float64. Adding the signature will
```

mlflow/pytorch/__init__.py
```diff
@@ -128,6 +128,13 @@ def log_model(
     """
     Log a PyTorch model as an MLflow artifact for the current run.
 
+        .. warning::
```
andreakress, line 131: Just an FYI -- typically in tech writing, "warning" is used only for serious situations, such as physical injury or data loss. I think it's probably actually appropriate here, since there is a possibility of causing the model to produce inaccurate results.

mlflow/pytorch/__init__.py
```diff
@@ -128,6 +128,13 @@ def log_model(
...
+
+            Log the model with signature to avoid inference errors. Pytorch float precision default
+            is float32, while numpy float precision default is float64. Adding the signature will
+            ensure correct precision is used when the model is logged and the model can be queried
+            accurately.
+
```
andreakress, line 137: WDYT of:
Adding the signature will ensure correct precision is used when the model is logged and the model can be queried accurately.
--> Include the signature to ensure that the model is logged with the correct precision and produces accurate results.

andreakress, line 137: WDYT of:
Adding the signature will ensure correct precision is used when the model is logged and the model can be queried accurately.
--> Include the signature to ensure that the model is logged with the correct precision and can produce accurate results.

mlflow/pytorch/__init__.py
ankit-db: Agreed, @andreakress I would prefer that we use a warning class warning here. This kind of error took me multiple hours to debug and I have a full understanding of our entire stack. Our users could really get turned around here if we don't offer a warning.

mlflow/pytorch/__init__.py
ankit-db, line 133:
```suggestion
            Log the model with a signature to avoid inference errors. Pytorch float precision default
```

mlflow/pytorch/__init__.py
ankit-db, line 134: I would maybe re-frame this a bit and say:

For models without signatures, the MLflow Model Server relies on the default inferred data type from NumPy. However, PyTorch often expects different defaults, particularly when parsing floats.

mlflow/pytorch/__init__.py
ankit-db: I'd modify that a bit to:

Include the signature to ensure that the model is logged with the correct data type so that the MLflow model server can correctly produce valid input to the model.

ankit-db (commented):
Good start - a few notes

mlflow/pytorch/__init__.py
mehtayogita: Ack.

mlflow/pytorch/__init__.py
mehtayogita: Ack.

mlflow/pytorch/__init__.py
mehtayogita: Updated as suggested.

mlflow/pytorch/__init__.py
```diff
@@ -128,6 +128,15 @@ def log_model(
     """
     Log a PyTorch model as an MLflow artifact for the current run.
 
+        .. warning::
+
+            Log the model with a signature to avoid inference errors.
+            For models without signatures, the MLflow Model Server relies on the default inferred
+            data type from NumPy. However, PyTorch often expects different defaults, particularly
+            when parsing floats. Include the signature to ensure that the model is logged with the
+            correct data type so that the MLflow model server can correctly produce valid input
```
ankit-db, line 137: My bad here, since I made the earlier suggestion, but could we actually do "provide" instead of "produce"

ankit-db (approved):
One minor nit, but looks great otherwise!

mlflow/pytorch/__init__.py
mehtayogita: Ack, done.

mlflow/pytorch/__init__.py
```diff
@@ -128,6 +128,15 @@ def log_model(
     """
     Log a PyTorch model as an MLflow artifact for the current run.
 
+        .. warning::
+
+            Log the model with a signature to avoid inference errors.
+            For models without signatures, the MLflow Model Server relies on the default inferred
+            data type from NumPy. However, PyTorch often expects different defaults, particularly
+            when parsing floats. Include the signature to ensure that the model is logged with the
+            correct data type so that the MLflow model server can correctly provide valid input
```
andreakress, line 137:
```suggestion
            correct data type so that the MLflow model server correctly provides valid input
```

andreakress (approved):
Approved with one suggestion.

mlflow/pytorch/__init__.py
ankit-db: Just to clarify @andreakress - one thing I think would be good to emphasize is that by logging a signature, the user is making it *possible* for the model server to provide valid input. Correctly inferring the correct data types without the signature is an impossible problem. Maybe just me, but, in your suggested phrasing, it kind of feels like we're saying that there's a bug where it won't provide it correctly.

andreakress: WDYT of something like this:

If the model is logged without a signature, the MLflow Model Server relies on the default inferred data type from NumPy. However, PyTorch often expects different defaults, particularly when parsing floats. You must include the signature to ensure that the model is logged with the correct data type so that the MLflow model server can correctly provide valid input.

ankit-db: Sounds good!

ankit-db: Thanks for helping iterate on this one!

Issue #5208, opened 2021-12-29
[FR] Validate request JSON with JSON schema or similar


mrkaye97:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [ ] Yes. I can contribute this feature independently.
- [x] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

API calls should return `HTTP 400` when the parameters (e.g.) don't match expected data types instead of failing with a `500`. Creating a JSON schema -- using [jsonschema](https://github.com/Julian/jsonschema), for example -- for the MLFlow REST API to check requests against would fix these issues. This would result in far friendlier UX, easier debugging, more predictable responses, and a generally more RESTful API.

## Motivation

I keep getting `500` errors for things like supplying a parameter to an API call that's the wrong data type. See [this issue](https://github.com/mlflow/mlflow/issues/5199) for an example. This has also happened with calls to logging parameters (both individually and in batches) and all kinds of other functions.

Right now, this means that an end user of a running MLFlow service get an error message like this back when something goes wrong:

```
Response [https://<<host>>/api/2.0/mlflow/runs/log-batch]
  Date: 2021-12-29 20:15
  Status: 500
  Content-Type: text/html; charset=utf-8
  Size: 290 B
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
```

This error was caused by providing a `timestamp` value to `log-batch` that was a character string as opposed to a numeric timestamp.

[...]

PR #5350, opened 2022-02-06
WIP: JSON Schema Validation, Part 1

mrkaye97:
WIP. Opening this PR to refer back to in #5208 

This PR is a first pass at adding JSON validation against a predetermined schema, in an effort to make error handling more friendly for users. The end goal is to return `HTTP 400` for bad parameters instead of `500`, which happens currently (at least for most endpoints)

## How is this patch tested?

WIP -- Not tested yet. What is the best way to test changes like this?

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mrkaye97:
See [this comment](https://github.com/mlflow/mlflow/issues/5208#issuecomment-1030858836) for some context on the issue I'm grappling with here. I may just be missing something about `parse_dict` since I'm not familiar with it, but at least afaict it looks like we might need to do extra type conversion or JSON parsing for this to work.

mlflow/server/handlers.py
```diff
@@ -323,6 +345,31 @@ def _not_implemented():
...
+        {
+          "title": "experiments/create",
+          "properties": {
+            "name": {
+              "type": "string",
+              "description": "Experiment name."
+            },
+            "artifact_location": {
+              "type": "string",
+              "description": "Location where all artifacts for the experiment are stored. If not provided, the remote server will select an appropriate default."
```
mrkaye97, line 359: Copied from the REST API doc

mrkaye97:
@dbczumar Moving discussion over here.
1. What is the best way to test these changes? Is CI passing (waiting on that currently) enough to be confident in them? Are there tests living somewhere that I should be running on local before pushing? 
2. [This](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R225-R229) seems like a reasonable approach to me, for now at least. It'll let me keep developing in smaller batches. WDYT?
3. [Is this the proper way to raise a 400 for bad params](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R239-R243)?
4. I'm not sure how I feel about [including the schema in the methods](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R349-R371). I think it'd probably be better to define a big object somewhere with _all_ of the schemas, and then look up the one we need instead of defining it inside of each method. WDYT?

dbczumar:
> @dbczumar Moving discussion over here.
> 
> 1. What is the best way to test these changes? Is CI passing (waiting on that currently) enough to be confident in them? Are there tests living somewhere that I should be running on local before pushing?
> 2. [This](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R225-R229) seems like a reasonable approach to me, for now at least. It'll let me keep developing in smaller batches. WDYT?
> 3. [Is this the proper way to raise a 400 for bad params](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R239-R243)?
> 4. I'm not sure how I feel about [including the schema in the methods](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R349-R371). I think it'd probably be better to define a big object somewhere with _all_ of the schemas, and then look up the one we need instead of defining it inside of each method. WDYT?

@mrkaye97 Great questions!

1. In addition to CI on this PR, running https://github.com/mlflow/mlflow/blob/master/tests/tracking/test_rest_tracking.py via `cd tests/tracking && pytest test_rest_tracking.py` with `pytest==6.2.5` will provide a good sense of whether or not things are working.

2. Sounds great

3. We recommend raising `MlflowException` with error code `INVALID_PARAMETER_VALUE`. Here's an example: https://github.com/mlflow/mlflow/blob/9c70732d4f380ded5a31a2ca76686306956189d8/mlflow/pyfunc/__init__.py#L819-L823

4. Do we need to store the full schema? Perhaps it would be sufficient to have the validators access attributes of `request_message` and assert on them? This logic can be specific to each validator.

mrkaye97:
Great, thanks for the feedback. That all makes sense. On 4), are you thinking something [like this](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R283-R287) and [this](https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R222-R224) where I've just factored out some schema logic into a dictionary? This seems problematic to me because it's too automagical, but I'm not 100% sure if I was understanding your suggestion correctly.

Another thing that's probably cleaner would be to do something like this:

```python
def _name_schema(**kwargs):
  base = {"type": "string"}
  for k, v in **kwargs:
    base[k] = v
```

(excuse ugly python -- not much of a pythoner 🙂)

This way we would be able to pass additional things, like `"required": true` or `"maxItems": 10` or something, and then could validate similarly:

```python
Draft202012Validator(_name_schema()).is_valid(request_json["name"])
```

I'm sure this could get cleaned up or made a little more Pythonic, but it seems like a reasonable strategy to me

dbczumar:
> 

@mrkaye97 This looks fantastic (https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R283-R287 / https://github.com/mlflow/mlflow/pull/5350/files#diff-5c10a4e2ca47745f06fa9e7201087acfc102849756cb8d85e774a5ac468cb037R222-R224). I don't have a strong preference either way between the automagical approach and the second one you've outlined. Both seem to translate a dictionary of field names and types into a validation routine, right?

dbczumar:
Hi @mrkaye97 , any updates here? Let me know if there's anything we can help with.

mrkaye97:
Workin on it! Been busy at the day job for the past week or so. I'll plan on taking another look tomorrow or Thursday and report back, but I think I've got a pretty good idea of how to design an initial pass at this point. 

Separately, I'm curious about the test failures I'm seeing in CI. For instance, in the R tests it seems like there are a lot of flaky tests failing after timing out waiting for a server to spin up. I'm mostly just wondering is this is the usual state of affairs for these tests -- was seeing similar things for the Python tests too, where things were failing for completely unrelated reasons to any of these changes that were unpredictable.

mrkaye97:
Closing in favor of #5458

PR #5378, opened 2022-02-14
Introducing abstract endpoint methods

trangevi:

This PR is to introduce the concept of "endpoints" into MLflow. Endpoints are a layer above deployments, intended to allow for the addition of features such as traffic splitting for the sake of deployment upgrades and AB testing.

## How is this patch tested?

Abstract methods only, no tests.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Adds an abstract BaseEndpointClient for plugins to utilize for supporting Endpoint objects.

Components 

Interface 

Language 

Integrations


mlflow/deployments/base.py
```diff
@@ -87,7 +87,81 @@ def __init__(self, target_uri):
...
+        name.
+        The dict is guaranteed to contain an 'name' key containing the endpoint name.
+        The other fields of the returned dictionary and their types may vary across targets.
+
+        :param name: Name of endpoint to fetch
+        """
+        pass
+
+    @abc.abstractmethod
+    def create_deployment(self, name, model_uri, flavor=None, config=None, endpoint_uri=None):
```
jinzhang21, line 164: Could `endpoint_uri` be folded into `config`? It decouples the concepts of deployment from endpoint. For use cases w/o endpoint, a user doesn't have to worry about it at all.

Ditto below for other *_deployment() functions.

mlflow/deployments/base.py
```diff
@@ -87,7 +87,81 @@ def __init__(self, target_uri):
         self.target_uri = target_uri
 
     @abc.abstractmethod
-    def create_deployment(self, name, model_uri, flavor=None, config=None):
+    def create_endpoint(self, name, config=None):
+        """
```
jinzhang21, line 91: Instead of directly modifying `BaseDeploymentClient` class, which is meant to be used by all possible deployment plugins / implementations (some of which doesn't support the concept of endpoint), how about creating a new class deriving from `BaseDeploymentClient`? e.g. `BaseEndpointClient`, or `BaseDeploymentClientWithEndpoint`

mlflow/deployments/base.py
```diff
@@ -126,72 +202,105 @@ def update_deployment(self, name, model_uri=None, flavor=None, config=None):
...
 
         :param name: ID of deployment to fetch
+        :param endpoint_uri: URI of the Endpoint containing the deployment to get. May not be
+                             supported by all targets.
         """
         pass
 
     @abc.abstractmethod
-    def predict(self, deployment_name, df):
+    def predict(self, uri, df):
```
jinzhang21, line 259: If a subclass is created, there's no need to worry about backward-compatibility issues here:)

mlflow/deployments/base.py
```diff
@@ -126,72 +202,105 @@ def update_deployment(self, name, model_uri=None, flavor=None, config=None):
...
-        :param deployment_name: Name of deployment to predict against
+        :param uri: The URI of the endpoint or deployment to predict against. This can take one of
+                    three forms:
+
+                    - ``endpoints:/<endpoint name>`` for predicting against the default deployment
+                    in an endpoint
+                    - ``deployments:/<endpoint name>/<deployment name>`` for predicting against a
+                    specific deployment in an endpoint
+                    - ``deployments:/<deployment name>`` for predicting against a deployment for
+                    targets which don't support endpoints
```
jinzhang21, line 274: Since the support for endpoint (new) and deployment (existing) is loose, we have not touched whether and how to officially support the use of the scheme `"endpoints:/"` or `"deployments:/"` in the MLflow URI. For example, how could we tell `deployments:/<endpoint name>/<deployment name>` from `deployments:/<deployment name>`? To do so would require both deployment name and endpoint name not contain any "/".

To make the client support broadly applicable. It might be better to not use URI here at all but simply specify a generic name that can be interpreted properly by plugins. I apologize for contradicting myself on the design doc comments.

mlflow/deployments/base.py
dbczumar: @trangevi @jinzhang21 As we discussed offline, I think that deployments and endpoints are strongly coupled on certain platforms (e.g. AzureML) and pretty strongly linked on other platforms (SageMaker, TorchServe). Accordingly, providing users with a single function for creating a deployment and adding it to an endpoint seems like a good idea because it's broadly applicable. If we're having trouble figuring out how to refer to an endpoint (i.e. should it be a string name? should it be a URI? should the URI embed hostname information?), perhaps we can call this parameter `endpoint` for deployment-related methods (e.g. `create_deployment`). This provides us the flexibility to introduce references to endpoints by name, URI, and / or something else in the future. We could initially start with supporting names (this is consistent with the CRUD methods for endpoints introduced in this PR). Thoughts?

jinzhang21: Naming it as `endpoint` SGTM.

mlflow/deployments/base.py
```diff
@@ -70,6 +70,226 @@ def target_help():
...
+        .. Note::
+            Subclasses should raise :py:class:`mlflow.exceptions.MlflowException` in error cases (e.g.
+            on failure to deploy a model).
+        """
+
+    def __init__(self, target_uri):
+        self.target_uri = target_uri
+
+    @abc.abstractmethod
+    def create_endpoint(self, endpoint_name, config=None):
```
jinzhang21, line 89: `endpoint_name` seems not necessary here. Generic `name` should be sufficient. In any case please update the comments (at the moment it contains both `endpoint_name` and `name`).
Ditto below for `update_endpoint`, `delete_endpoint` and `get_endpoint`

mlflow/deployments/base.py
```diff
@@ -70,6 +70,226 @@ def target_help():
...
+        name.
+        The dict is guaranteed to contain an 'name' key containing the endpoint name.
+        The other fields of the returned dictionary and their types may vary across targets.
+
+        :param endpoint_name: Name of endpoint to fetch
+        """
+        pass
+
+    @abc.abstractmethod
+    def create_deployment(self, deployment_name, endpoint_name, model_uri, flavor=None, config=None):
```
jinzhang21, line 163: I don't have strong opinion here, but changing
`def create_deployment(self, deployment_name, endpoint_name, ...` to
`def create_deployment(self, name, endpoint, ...` might be cleaner and more future compatible:
Imagine we introduce a URI/URL to identify an endpoint, `endpoint` can still be used in this case.

mlflow/deployments/base.py
```diff
@@ -70,6 +70,226 @@ def target_help():
...
+        The other fields of the returned dictionary and their types may vary across
+        deployment targets.
+
+        :param deployment_name: Name of deployment to fetch
+        :param endpoint_name: Name of the Endpoint containing the deployment to fetch
+        """
+        pass
+
+    @abc.abstractmethod
+    def predict(self, endpoint_name, df, deployment_name=None):
```
jinzhang21, line 254: A similar argument as above on the naming convention:
`def predict(self, endpoint, df, deployment=None):`
may offer more flexibility.

mlflow/deployments/base.py
trangevi: Switched to endpoint, as discussed

mlflow/deployments/base.py
trangevi: I switched to `name` for this method, but left the others with the `endpoint` parameter, as we'd talked about previously. Please let me know if you think it should be handled differently.

mlflow/deployments/base.py
```diff
@@ -70,6 +70,226 @@ def target_help():
...
+
+        :param endpoint: Name of the endpoint to issue the prediction against
+        :param df: Pandas DataFrame to use for inference
+        :param deployment: If provided issue a prediction against the specified endpoint,
+                           bypassing the endpoint traffic settings
+        :return: A pandas DataFrame, pandas Series, or numpy array
+        """
+        pass
+
+    @experimental
```
jinzhang21, line 274: Is explain method strictly required?

trangevi: This was another one that existed in the BaseDeploymentClient, so I added it here as well. I think it's probably good to maintain that parity.

mlflow/deployments/base.py
```diff
@@ -70,6 +70,226 @@ def target_help():
...
+        ``BaseEndpointClient`` within the plugin module, and customize method docstrings with
+        target-specific information.
+
+        .. Note::
+            Subclasses should raise :py:class:`mlflow.exceptions.MlflowException` in error cases
+            (e.g. on failure to deploy a model).
+        """
+
+    def __init__(self, target_uri):
+        self.target_uri = target_uri
```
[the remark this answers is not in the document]
jinzhang21: It's not documented here an implementer would have no idea what it means and what to do with it. If it's a use case for AML I'd recommend to keep it for AML specialization.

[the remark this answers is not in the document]
trangevi: Sure, I can remove it

mlflow/deployments/base.py
trangevi: Not currently, possibly in the future. But given we decided that any endpoint related deployment operations should go through the endpoint client, I'm assuming we want to match the current deployment client for existing methods

jinzhang21: It's much easier to add the method back than taking it away. Shall we file another PR when the need arises? We might have a different idea by then on the API (since the requirements are much more concrete). WDYT?

jinzhang21 (commented):
LGTM! Thanks for the addition, @trangevi ! Just one more comment on the explain method:)

trangevi:
autoformat

jinzhang21 (approved):

mlflow/deployments/base.py
```diff
@@ -70,6 +70,208 @@ def target_help():
...
+    ``BaseEndpointClient`` within the plugin module, and customize method docstrings with
+    target-specific information.
+
+    .. Note::
+        Subclasses should raise :py:class:`mlflow.exceptions.MlflowException` in error cases
+        (e.g. on failure to deploy a model).
+    """
+
+    def __init__(self):
+        pass
```
dbczumar, line 87: @jinzhang21 I saw that we removed `target_uri` based on feedback from https://github.com/mlflow/mlflow/pull/5378#discussion_r842247310. `target_uri` is used to specify where deployment resources should be created - e.g. a particular AML workspace, AWS account, etc. Without such a capability, everything needs to be passed via `config`. I think it should be included. This gets propagated down in the deployment client case via https://mlflow.org/docs/latest/python_api/mlflow.deployments.html#mlflow.deployments.get_deploy_client.

@trangevi Independently of this, how do we envision users interacting with implementations of `BaseEndpointClient`? If I want to interact with a deployment client via the Python API, I run something like the following:

```
client = mlflow.deployments.get_deploy_client("<azureml_tracking_uri>")
client.create_deployment(...)
```

Will users do something like the following for an endpoint client?

```
endpoint_client = mlflow.deployments.get_endpoint_client("<azureml_tracking_uri>")
endpoint_client.create_endpoint(...)
```

To me, exposing these methods on the user-facing deployment client seems more natural:

```
client = mlflow.deployments.get_deploy_client("<azureml_tracking_uri>")
client.create_deployment(...)
client.create_endpoint(...)
```

Is this the intention? If so, should `BaseEndpointClient` be a distinct abstract class? I think that's fine as long as there's a plan for exposing `BaseEndpointClient` methods in the user-facing result of `mlflow.deployments.get_endpoint_client`, but we should make sure to align on the UX and implementation details for this before moving forward since it isn't obvious.

dbczumar (changes requested):
Can we resolve https://github.com/mlflow/mlflow/pull/5378/files#r845638368 before merging?

mlflow/deployments/base.py
trangevi: @dbczumar, the discussion that we'd had to add a distinct BaseEndpointClient was based on not ever plugin implementer having an endpoints concept. This was decided as the best way to support both scenarios while being clearest to users of the clients what they can interact with (ie not having deployment methods with optional endpoint parameters that may or may not make sense).

I can add a `get_endpoint_client` method to match the deployment one, if that's the missing piece for exposing it as a distinct entity. And add back target_uri to match.

jinzhang21: On adding back target_uri, can we add comments on why it's needed and how it will be used? It wasn't clear in the initial commit, leading to a suggested removal.

dbczumar: @trangevi I think we need to consider / align on the user experience before merging this code. As it stands, the code introduces a developer-facing `BaseEndpointClient` without creating any corresponding user-facing APIs for interacting with it. I'm also hesitant to bifurcate the user-facing API surface into deployment APIs exposed by `DeploymentClient` and separate endpoint APIs exposed by `EndpointClient`, since they two are conceptually intertwined. I'd favor the following UX:

```
client = mlflow.deployments.get_deploy_client("<azureml_tracking_uri>")
client.create_deployment(...)
client.create_endpoint(...)
```

If this means that we have to merge `BaseEndpointClient` with `BaseDeploymentClient`, that's okay; instead of marking these methods as explicitly abstract via `abc.abstractmethod`, we can have them raise `NotImplemented` and include informative docstrings explaining their intended use cases and telling developers when they should be implemented.

trangevi:
autoformat

mlflow/deployments/base.py
```diff
@@ -70,6 +70,213 @@ def target_help():
...
+        The other fields of the returned dictionary and their types may vary across
+        deployment targets.
+
+        :param deployment: Name of deployment to fetch
+        :param endpoint: Name of the Endpoint containing the deployment to fetch
+        """
+        pass
+
+    @abc.abstractmethod
+    def predict(self, endpoint, df, deployment=None):
```
dbczumar, line 259: Regarding the extensibility of `BaseDeploymentClient.predict()` (https://github.com/databricks/universe/pull/163268) to endpoints, I think the following signature should work:

```
def predict(self, deployment=None, endpoint=None):
     """
     Exactly one of `deployment` or `endpoint` must be specified.
     """
```

Consider preexisting code that calls `predict("my_deployment")`. This will continue to behave as it did before, since `deployment` remains the first positional arg. New code for predicting on an endpoint can be constructed as `predict(endpoint="my_endpoint").`

mlflow/deployments/base.py
trangevi: Discussed offline and made corresponding changes

jinzhang21 (approved):

mlflow/sagemaker/__init__.py
```diff
@@ -1881,7 +1881,7 @@ def _apply_custom_config(self, config, custom_config):
             config[key] = value
 
     @experimental
-    def create_deployment(self, name, model_uri, flavor=None, config=None):
+    def create_deployment(self, name, model_uri, flavor=None, config=None, endpoint=None):
```
dbczumar, line 1884: Nit: Can we document that `endpoint` is currently unused

mlflow/deployments/base.py
```diff
@@ -182,22 +196,103 @@ def predict(self, deployment_name, df):
...
         :param df: Pandas DataFrame to use for explaining feature importance in model prediction
+        :param endpoint: Endpoint to predict against. May not be supported by all targets
         :return: A JSON-able object (pandas dataframe, numpy array, dictionary), or
                  an exception if the implementation is not available in deployment target's class
         """
         raise MlflowException(
             "Computing model explanations is not yet supported for this deployment target"
         )
+
+    @abc.abstractmethod
```
dbczumar, line 223: Decorating these with `@abc.abstractmethod` will break existing plugin implementations that inherit from `BaseDeploymentClient`, e.g.:

```
 Can't instantiate abstract class MyDeploymentClient with abstract methods create_endpoint, delete_endpoint, get_endpoint, list_endpoints, update_endpoint
```


I've confirmed that preexisting abstract method implementations that don't define the `endpoint` kwarg (e.g. the ones in SageMaker on `master`) continue to work just fine. However, I think we should keep the changes to SageMaker to fully comply with the new deployment client structure and incentivize contributors to add support for it.

dbczumar (commented):
@trangevi Looks great! Left a couple small inline comments. Additionally, can we update the deployments cli to expose these endpoint capabilities? https://github.com/mlflow/mlflow/blob/master/mlflow/deployments/cli.py

dbczumar (approved):
LGTM! Thanks @trangevi !

dbczumar:
autoformat

PR #5386, opened 2022-02-16
ignore micro-version mismatch for databricks pyspark

WeichenXu123:

ignore micro-version mismatch for databricks pyspark

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

ignore micro-version mismatch for databricks pyspark

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/requirements_utils.py
+++ b/mlflow/utils/requirements_utils.py
@@ -442,14 +442,14 @@ def _check_requirement_satisfied(requirement_str):
     Otherwise, returns an instance of `_MismatchedPackageInfo`.
     """
     _init_packages_to_modules_map()
-    req = pkg_resources.Requirement.parse(requirement_str)
-    pkg_name = req.name
 
-    if pkg_name == "pyspark" and is_in_databricks_runtime():
+    if re.match(r"pyspark[^\w-]", requirement_str) and is_in_databricks_runtime():
         # for databricks runtime pyspark, ignore micro-version mismatch.
         # replace == specifier to be ~= specifier, but keep === specifier unchanged.
         requirement_str = re.sub("([^=])==([^=])", r"\1~=\2", requirement_str)
-        req = pkg_resources.Requirement.parse(requirement_str)
+
+    req = pkg_resources.Requirement.parse(requirement_str)
+    pkg_name = req.name
 
     try:
         installed_version = _get_installed_version(pkg_name, _PACKAGES_TO_MODULES.get(pkg_name))
--- a/mlflow/utils/requirements_utils.py
+++ b/mlflow/utils/requirements_utils.py
@@ -445,7 +445,7 @@ def _check_requirement_satisfied(requirement_str):
 
     if re.match(r"pyspark[^\w-]", requirement_str) and is_in_databricks_runtime():
         # for databricks runtime pyspark, ignore micro-version mismatch.
-        # replace == specifier to be ~= specifier, but keep === specifier unchanged.
+        # replace == operator with ~= operator, but keep === operator unchanged.
         requirement_str = re.sub("([^=])==([^=])", r"\1~=\2", requirement_str)
 
     req = pkg_resources.Requirement.parse(requirement_str)
--- a/mlflow/utils/requirements_utils.py
+++ b/mlflow/utils/requirements_utils.py
@@ -445,7 +445,7 @@ def _check_requirement_satisfied(requirement_str):
     req = pkg_resources.Requirement.parse(requirement_str)
     pkg_name = req.name
 
-    if is_in_databricks_runtime() and pkg_name == 'pyspark':
+    if pkg_name == 'pyspark' and is_in_databricks_runtime():
         # for databricks runtime pyspark, ignore micro-version mismatch.
         # replace == specifier to be ~= specifier, but keep === specifier unchanged.
         requirement_str = re.sub('([^=])==([^=])', r'\1~=\2', requirement_str)
--- a/mlflow/utils/requirements_utils.py
+++ b/mlflow/utils/requirements_utils.py
@@ -445,6 +445,12 @@ def _check_requirement_satisfied(requirement_str):
     req = pkg_resources.Requirement.parse(requirement_str)
     pkg_name = req.name
 
+    if is_in_databricks_runtime() and pkg_name == 'pyspark':
+        # for databricks runtime pyspark, ignore micro-version mismatch.
+        # replace == specifier to be ~= specifier, but keep === specifier unchanged.
+        requirement_str = re.sub('([^=])==([^=])', r'\1~=\2', requirement_str)
+        req = pkg_resources.Requirement.parse(requirement_str)
+
     try:
         installed_version = _get_installed_version(pkg_name, _PACKAGES_TO_MODULES.get(pkg_name))
     except ModuleNotFoundError:
--- a/mlflow/utils/requirements_utils.py
+++ b/mlflow/utils/requirements_utils.py
@@ -445,10 +445,10 @@ def _check_requirement_satisfied(requirement_str):
     req = pkg_resources.Requirement.parse(requirement_str)
     pkg_name = req.name
 
-    if pkg_name == 'pyspark' and is_in_databricks_runtime():
+    if pkg_name == "pyspark" and is_in_databricks_runtime():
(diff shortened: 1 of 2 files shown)


dbczumar (approved):
LGTM! Thanks @WeichenXu123 !

mlflow/utils/requirements_utils.py
```diff
@@ -445,6 +445,12 @@ def _check_requirement_satisfied(requirement_str):
     req = pkg_resources.Requirement.parse(requirement_str)
     pkg_name = req.name
 
+    if pkg_name == 'pyspark' and is_in_databricks_runtime():
```
harupy, line 448: Can we add a test?

mlflow/utils/requirements_utils.py
```diff
@@ -445,6 +445,12 @@ def _check_requirement_satisfied(requirement_str):
...
+        # for databricks runtime pyspark, ignore micro-version mismatch.
+        # replace == specifier to be ~= specifier, but keep === specifier unchanged.
+        requirement_str = re.sub('([^=])==([^=])', r'\1~=\2', requirement_str)
```
harupy, line 451: Can we run this operation after `_init_packages_to_modules_map`?

harupy:
```suggestion
        requirement_str = re.sub(r'==([^=])', r'~=\1', requirement_str)
```

Can we only replace the version specifier?

WeichenXu123: This seems does not work for `===`
`re.sub(r'==([^=])', r'~=\1', 'abc===def')` get `'abc=~=def'`

tests/pyfunc/test_warn_dependency_requirement_mismatches.py
```diff
@@ -9,6 +9,18 @@
...
+def _gen_mock_get_installed_version_fn(mock_versions):
+    original_get_installed_version_fn = mlflow.utils.requirements_utils._get_installed_version
+
+    def mock_get_installed_version_fn(package, module=None):
+        if package in mock_versions:
+            return mock_versions[package]
+        else:
+            return original_get_installed_version_fn(package, module)
+
+    return mock_get_installed_version_fn
```
harupy, line 21:
```python
@contextmanager
def mock_get_installed_version(mock_versions):
    original_get_installed_version = mlflow.utils.requirements_utils._get_installed_version

    def new_get_installed_version(package, module=None):
        if package in mock_versions:
            return mock_versions[package]
        else:
            return original_get_installed_version(package, module)

    with mock.patch(
         "mlflow.utils.requirements_utils._get_installed_version",
         mocked_get_installed_version) as mock_get_installed_version:
         yield mock_get_installed_version
```

Can we define a context manager like this?

mlflow/utils/requirements_utils.py
```diff
@@ -442,6 +442,12 @@ def _check_requirement_satisfied(requirement_str):
     Otherwise, returns an instance of `_MismatchedPackageInfo`.
     """
     _init_packages_to_modules_map()
+
+    if re.match(r"pyspark[^\w-]", requirement_str) and is_in_databricks_runtime():
+        # for databricks runtime pyspark, ignore micro-version mismatch.
+        # replace == specifier to be ~= specifier, but keep === specifier unchanged.
```
harupy, line 448:
```suggestion
        # replace == operator with ~= operator, but keep === operator unchanged.
```

tests/pyfunc/test_warn_dependency_requirement_mismatches.py
```diff
@@ -110,3 +111,30 @@ def mock_get_installed_version_fn(package, module=None):
...
             )
+
+
+@pytest.mark.large
+def test_warn_dependency_requirement_mismatches_ignore_databricks_runtime_micro_version(tmpdir):
+    req_file = tmpdir.join("requirements.txt")
+    req_file.write("pyspark==3.2.1")
+
+    with mock.patch("mlflow.pyfunc._logger.warning") as mock_warning:
+        with mock.patch("mlflow.utils.requirements_utils.is_in_databricks_runtime", lambda: True):
```
harupy, line 122:
```suggestion
        with mock.patch("mlflow.utils.requirements_utils.is_in_databricks_runtime", return_value=True):
```

you could also use `return_value`

tests/pyfunc/test_warn_dependency_requirement_mismatches.py
```diff
@@ -110,3 +111,30 @@ def mock_get_installed_version_fn(package, module=None):
...
+
+        with mock.patch(
+            "mlflow.utils.requirements_utils._get_installed_version",
+            _gen_mock_get_installed_version_fn({"pyspark": "3.2.2"}),
+        ):
+            _warn_dependency_requirement_mismatches(model_path=tmpdir)
+            mock_warning.assert_called_once_with(
+                AnyStringWith(" - pyspark (current: 3.2.2, required: pyspark==3.2.1)")
+            )
+            mock_warning.reset_mock()
```
harupy, line 140: `3.2.2` and `3.2.1` should be considered as matched, right?

mlflow/utils/requirements_utils.py
harupy: Got it, I misunderstood `===` was a typo, but it's a valid operator.

tests/pyfunc/test_warn_dependency_requirement_mismatches.py
WeichenXu123: this is test for non-databricks pyspark.

mlflow/utils/requirements_utils.py
```diff
@@ -442,6 +442,12 @@ def _check_requirement_satisfied(requirement_str):
     Otherwise, returns an instance of `_MismatchedPackageInfo`.
     """
     _init_packages_to_modules_map()
+
+    if re.match(r"pyspark[^\w-]", requirement_str) and is_in_databricks_runtime():
+        # for databricks runtime pyspark, ignore micro-version mismatch.
+        # replace == operator with ~= operator, but keep === operator unchanged.
+        requirement_str = re.sub("([^=])==([^=])", r"\1~=\2", requirement_str)
```
harupy, line 449: Can this handle a requirement string that looks like `pyspark>=1.2.3,==1.2.*`?

WeichenXu123: emm. this is an issue.

WeichenXu123: i will update regex to cover the case ==3.2 or ==3.2.*

dbczumar: I don't think it needs to, right? The purpose of this change is to make sure that exact requirement specifiers generated automatically by MLflow, e.g. `pyspark==3.2.1`, are not interpreted too strictly on Databricks. If the user wants to supply a custom / stricter specifier, then they should be able to without us interfering / interpreting the specifier differently.

harupy: I'm trying to come up with a more robust approach:

```python
import pkg_resources
from packaging.requirements import Requirement
from packaging.specifiers import SpecifierSet
import re


def is_satisfied(req, version):
    req = pkg_resources.Requirement.parse(req)
    if req.name == "pyspark" and len(req.specifier) == 1:
        m = re.match(r"==([^=]+)", str(req.specifier))
        if m:
            return SpecifierSet(f"~={m.group(1)}").contains(version)
    return req.specifier.contains(version)


print(is_satisfied("pyspark == 1.2.2", "1.2.3"))  # True
print(is_satisfied("pyspark>=1.2.3,==1.2.*", "1.2.3"))  # True
print(is_satisfied("pyspark === abc", "1.2.3"))  # False
print(is_satisfied("scikit-learn == 1.2.2", "1.2.3"))  # False

```

dbczumar: Actually, @WeichenXu123 , do we need this PR? The micro version isn't going to change within the same MLR / DBR version, right?

dbczumar (changes requested):
On second thought, not sure we need to do this. See https://github.com/mlflow/mlflow/pull/5386/files#r808837619

mlflow/utils/requirements_utils.py
WeichenXu123: Confirmed with @hyukjin  the pyspark version is not changed within the same MLR/DBR version

PR #5389, opened 2022-02-18
[Custom Metrics] Enabled support for logging of numerical metrics

MarkYHZhang:

Added support for users to pass in custom metric functions into mlflow.evaluate. This PR is a part of more PRs to come for custom metrics support. Specifically, this PR focuses on enabling logging of numerical metrics generated by user's custom metric functions into MLflow.

## How is this patch tested?

Wrote unit tests located in mlflow/tests/models/test_default_evaluator.py

Updates to the documentation will be done once the entire feature of custom metrics is complete.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Enables MLflow tracking of custom metrics produced by user-provided metric functions.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -483,7 +483,7 @@ class ModelEvaluator(metaclass=ABCMeta):
         :param run_id: The ID of the MLflow Run to which to log results.
         :param evaluator_config: A dictionary of additional configurations for
                                  the evaluator.
-        :param custom_metrics: A list of callable custom metric functions
+        :param custom_metrics: A list of callable custom metric functions.
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
         :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
@@ -837,8 +837,9 @@ def evaluate(
                                       metrics: Dict[AnyStr, Union[int, float, np.number]] = ???
                                       artifacts: Dict[AnyStr, Any] = ???
                                       ...
-                                      # If no artifacts, replace the below with `return metrics`
-                                      return metrics, artifacts
+                                      if artifacts is not None:
+                                          return metrics, artifacts
+                                      return metrics
                                   ``
 
                               Example Usage:
@@ -858,8 +859,8 @@ def evaluate(
 
                               with mlflow.start_run():
                                   mlflow.evaluate(
-                                      model, X, targets, custom_metrics=(
-                                      squared_diff_plus_one,scatter_plot)...)
+                                      model, X, targets, custom_metrics=[
+                                      squared_diff_plus_one,scatter_plot,...])
 
     :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
              evaluation results.
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -469,7 +469,7 @@ class ModelEvaluator(metaclass=ABCMeta):
         dataset,
         run_id,
         evaluator_config,
-        custom_metric_fns=None,
+        custom_metrics=None,
         **kwargs,
     ):
         """
@@ -483,7 +483,7 @@ class ModelEvaluator(metaclass=ABCMeta):
         :param run_id: The ID of the MLflow Run to which to log results.
         :param evaluator_config: A dictionary of additional configurations for
                                  the evaluator.
-        :param custom_metric_fns: A list of callable custom metric functions
+        :param custom_metrics: A list of callable custom metric functions
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
         :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
@@ -608,7 +608,7 @@ def _evaluate(
     run_id,
     evaluator_name_list,
     evaluator_name_to_conf_map,
-    custom_metric_fns,
+    custom_metrics,
 ):
     """
     The public API "evaluate" will verify argument first, and then pass normalized arguments
@@ -642,7 +642,7 @@ def _evaluate(
                 dataset=dataset,
                 run_id=run_id,
                 evaluator_config=config,
-                custom_metric_fns=custom_metric_fns,
+                custom_metrics=custom_metrics,
             )
             eval_results.append(result)
 
@@ -674,7 +674,7 @@ def evaluate(
     feature_names: list = None,
     evaluators=None,
     evaluator_config=None,
-    custom_metric_fns=None,
+    custom_metrics=None,
 ):
     """
     Evaluate a PyFunc model on the specified dataset using one or more specified ``evaluators``, and
@@ -798,7 +798,7 @@ def evaluate(
                              If multiple evaluators are specified, each configuration should be
                              supplied as a nested dictionary whose key is the evaluator name.
 
-    :param custom_metric_fns: (Optional) A list of custom metric functions. A custom metric
+    :param custom_metrics: (Optional) A list of custom metric functions. A custom metric
                               function is required to take in two parameters:
                               - Union[pandas.Dataframe, pyspark.sql.DataFrame]: The first being a
                                 Pandas or Spark DataFrame containing ``prediction`` and ``target``
@@ -897,5 +897,5 @@ def evaluate(
             run_id=run_id,
             evaluator_name_list=evaluator_name_list,
             evaluator_name_to_conf_map=evaluator_name_to_conf_map,
-            custom_metric_fns=custom_metric_fns,
+            custom_metrics=custom_metrics,
         )
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -815,52 +815,27 @@ def evaluate(
                               - Dict[AnyStr, Union[int, float, np.number]: a singular dictionary of
                                 custom metrics, where the keys are the names of the metrics, and the
                                 values are the scalar values of the metrics.
-                              - Tuple[Dict[AnyStr, Union[int,float,np.number]], Dict[AnyStr,Any]]:
-                                a tuple of a dict containing the custom metrics, and a dict of
-                                artifacts, where the keys are the names of the artifacts, and the
-                                values are objects representing the artifacts.
 
-                              Object types that artifacts be represented as:
-                              - A string uri representing the file path to the artifact. MLflow will
-                                infer the type of the artifact based on the file extension.
-                              - ``EvaluationArtifact`` type objects. i.e. CsvEvaluationArtifact,
-                                ImageEvaluationArtifact.
-                              - Pandas DataFrame. This will be resolved as a CSV artifact.
-                              - Numpy array. This will be saved as a .npy artifact.
-                              - Matplotlib Figure. This will be saved as an image artifact.
-                              - Any JSON serializable object. This will be saved as a JSON artifact.
-
-                              Custom Metric Function Boilerplate:
-                                  ``
+                              .. code-block:: python
+                                  :caption: Custom Metric Function Boilerplate:
                                   def custom_metrics_boilerplate(eval_df, builtin_metrics):
                                       ...
                                       metrics: Dict[AnyStr, Union[int, float, np.number]] = ???
-                                      artifacts: Dict[AnyStr, Any] = ???
                                       ...
-                                      if artifacts is not None:
-                                          return metrics, artifacts
                                       return metrics
-                                  ``
 
-                              Example Usage:
-                              def squared_diff_plus_one(eval_df, builtin_metrics):
-                                  return {
-                                      "sdiff_plus_one": (np.sum(np.abs(eval_df["prediction"]
-                                                        - eval_df["target"]+1)**2))
-                                  }
+                              .. code-block:: python
+                                  :caption: Example usage of custom metrics:
+                                  def squared_diff_plus_one(eval_df, builtin_metrics):
+                                      return {
+                                          "sdiff_plus_one": (np.sum(np.abs(eval_df["prediction"]
+                                                            - eval_df["target"]+1)**2))
+                                      }
 
-                              def scatter_plot(eval_df, builtin_metrics):
-                                  plt.scatter(eval_df['prediction'], eval_df['target'])
-                                  plt.xlabel('Targets')
-                                  plt.ylabel('Predictions')
-                                  plt.title("Targets vs. Predictions")
-                                  plt.savefig("some/path/example.png")
-                                  return {}, {“pred_target_scatter”: “some/path/example.png”}
-
-                              with mlflow.start_run():
-                                  mlflow.evaluate(
-                                      model, X, targets, custom_metrics=[
-                                      squared_diff_plus_one,scatter_plot,...])
+                                  with mlflow.start_run():
+                                      mlflow.evaluate(
+                                          model, X, targets, custom_metrics=[
+                                          squared_diff_plus_one,...])
 
     :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
              evaluation results.
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -462,15 +462,7 @@ class ModelEvaluator(metaclass=ABCMeta):
 
     @abstractmethod
     def evaluate(
-        self,
-        *,
-        model,
-        model_type,
-        dataset,
-        run_id,
-        evaluator_config,
-        custom_metrics=None,
-        **kwargs,
+        self, *, model, model_type, dataset, run_id, evaluator_config, custom_metrics=None, **kwargs
     ):
         """
         The abstract API to log metrics and artifacts, and return evaluation results.
@@ -799,43 +791,55 @@ def evaluate(
                              supplied as a nested dictionary whose key is the evaluator name.
 
     :param custom_metrics: (Optional) A list of custom metric functions. A custom metric
-                              function is required to take in two parameters:
-                              - Union[pandas.Dataframe, pyspark.sql.DataFrame]: The first being a
-                                Pandas or Spark DataFrame containing ``prediction`` and ``target``
-                                column. The ``prediction`` column contains the predictions made by
-                                the model. The ``target`` column contains the corresponding labels
-                                to the predictions made on that row.
-                              - Dict: The second is a dictionary containing the metrics calculated
-                                by the default evaluator. The keys are the names of the metrics
-                                and the values are the scalar values of the metrics. Refer to the
-                                DefaultEvaluator behavior section for what metrics will be returned
-                                based on the type of model (i.e. classifier or regressor).
+                           function is required to take in two parameters:
 
-                              A custom metric function can return in two different formats:
-                              - Dict[AnyStr, Union[int, float, np.number]: a singular dictionary of
-                                custom metrics, where the keys are the names of the metrics, and the
-                                values are the scalar values of the metrics.
+                           - Union[pandas.Dataframe, pyspark.sql.DataFrame]: The first being a
+                             Pandas or Spark DataFrame containing ``prediction`` and ``target``
+                             column. The ``prediction`` column contains the predictions made by
+                             the model. The ``target`` column contains the corresponding labels
+                             to the predictions made on that row.
+                           - Dict: The second is a dictionary containing the metrics calculated
+                             by the default evaluator. The keys are the names of the metrics
+                             and the values are the scalar values of the metrics. Refer to the
+                             DefaultEvaluator behavior section for what metrics will be returned
+                             based on the type of model (i.e. classifier or regressor).
 
-                              .. code-block:: python
-                                  :caption: Custom Metric Function Boilerplate:
-                                  def custom_metrics_boilerplate(eval_df, builtin_metrics):
-                                      ...
-                                      metrics: Dict[AnyStr, Union[int, float, np.number]] = ???
-                                      ...
-                                      return metrics
+                           A custom metric function can return in the following format:
 
-                              .. code-block:: python
-                                  :caption: Example usage of custom metrics:
-                                  def squared_diff_plus_one(eval_df, builtin_metrics):
-                                      return {
-                                          "sdiff_plus_one": (np.sum(np.abs(eval_df["prediction"]
-                                                            - eval_df["target"]+1)**2))
-                                      }
+                           - Dict[AnyStr, Union[int, float, np.number]: a singular dictionary of
+                             custom metrics, where the keys are the names of the metrics, and the
+                             values are the scalar values of the metrics.
 
-                                  with mlflow.start_run():
-                                      mlflow.evaluate(
-                                          model, X, targets, custom_metrics=[
-                                          squared_diff_plus_one,...])
+                           .. code-block:: python
+                               :caption: Custom Metric Function Boilerplate
+
+                               def custom_metrics_boilerplate(eval_df, builtin_metrics):
+                                   # ...
+                                   metrics: Dict[AnyStr, Union[int, float, np.number]] = some_dict
+                                   # ...
+                                   return metrics
+
+                           .. code-block:: python
+                               :caption: Example usage of custom metrics
+
+                               def squared_diff_plus_one(eval_df, builtin_metrics):
+                                 return {
+                                     "squared_diff_plus_one": (
+                                         np.sum(
+                                             np.abs(
+                                                 eval_df["prediction"] - eval_df["target"] + 1
+                                             ) ** 2
+                                         )
+                                     )
+                                 }
+
+                               with mlflow.start_run():
+                                   mlflow.evaluate(
+                                       model,
+                                       X,
+                                       targets,
+                                       custom_metrics=[squared_diff_plus_one, ...],
+                                   )
 
     :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
              evaluation results.
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -825,7 +825,7 @@ def evaluate(
                                 infer the type of the artifact based on the file extension.
                               - ``EvaluationArtifact`` type objects. i.e. CsvEvaluationArtifact,
                                 ImageEvaluationArtifact.
-                              - Pandas DataFrame. This will be resolved as a parquet artifact.
+                              - Pandas DataFrame. This will be resolved as a CSV artifact.
                               - Numpy array. This will be saved as a .npy artifact.
                               - Matplotlib Figure. This will be saved as an image artifact.
                               - Any JSON serializable object. This will be saved as a JSON artifact.
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -462,7 +462,15 @@ class ModelEvaluator(metaclass=ABCMeta):
 
     @abstractmethod
     def evaluate(
-        self, *, model, model_type, dataset, run_id, evaluator_config, custom_metrics=None, **kwargs
+        self,
+        *,
+        model,
+        model_type,
+        dataset,
+        run_id,
+        evaluator_config,
+        custom_metric_fns=None,
+        **kwargs,
     ):
         """
         The abstract API to log metrics and artifacts, and return evaluation results.
@@ -475,7 +483,7 @@ class ModelEvaluator(metaclass=ABCMeta):
         :param run_id: The ID of the MLflow Run to which to log results.
         :param evaluator_config: A dictionary of additional configurations for
                                  the evaluator.
-        :param custom_metrics: A list of custom metric functions
+        :param custom_metric_fns: A list of callable custom metric functions
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
         :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
@@ -600,7 +608,7 @@ def _evaluate(
     run_id,
     evaluator_name_list,
     evaluator_name_to_conf_map,
-    custom_metrics,
+    custom_metric_fns,
 ):
     """
     The public API "evaluate" will verify argument first, and then pass normalized arguments
@@ -634,7 +642,7 @@ def _evaluate(
                 dataset=dataset,
                 run_id=run_id,
                 evaluator_config=config,
-                custom_metrics=custom_metrics,
+                custom_metric_fns=custom_metric_fns,
             )
             eval_results.append(result)
 
@@ -666,7 +674,7 @@ def evaluate(
     feature_names: list = None,
     evaluators=None,
     evaluator_config=None,
-    custom_metrics=None,
+    custom_metric_fns=None,
 ):
     """
     Evaluate a PyFunc model on the specified dataset using one or more specified ``evaluators``, and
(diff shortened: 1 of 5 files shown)


MarkYHZhang:
Hey @apurvakoti @dbczumar @jinzhang21 @WeichenXu123, can you all take a look? It seems like I don't have permission to add reviewers. So I'm tagging you all here. Thanks!

mlflow/models/evaluation/base.py
```diff
@@ -473,6 +475,7 @@ def evaluate(self, *, model, model_type, dataset, run_id, evaluator_config, **kw
         :param run_id: The ID of the MLflow Run to which to log results.
         :param evaluator_config: A dictionary of additional configurations for
                                  the evaluator.
+        :param custom_metrics: A list of custom metric functions
```
jinzhang21, line 478: Show what's required for the custom metric function signatures, and give an example on how it could be used (similar to what's discussed in the design discussion).

jinzhang21 (commented):
Added some initial comment on the API. I'll review the rest later today.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,6 +580,17 @@ def plot_pr_curve():
...
+    def _evaluate_custom_metric_fns(self):
+        for custom_metric_fn in self.custom_metric_fns:
+            # recreating the dataframe for each custom metric function call,
+            # in case the user modifies the eval_df inside their custom function.
+            eval_df = pd.DataFrame({"prediction": self.y_pred, "target": self.y})
+            metric_results, _ = _evaluate_custom_metric_fn(custom_metric_fn, eval_df, self.metrics)
+            # skip logging metric functions that doesn't return anything
+            if metric_results is None:
+                continue
+            self.metrics.update(metric_results)
```
apurva-koti, line 592: Can we add a TODO to log artifacts?

apurva-koti (commented):
Some small comments, will do a full review again

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +259,45 @@ def _do_plot(**kwargs):
 }
 
 
+def _evaluate_custom_metric_fn(custom_metric_fn, eval_df, builtin_metrics):
+    result = custom_metric_fn(eval_df, builtin_metrics)
+    if result is None:
+        _logger.warning(
+            f"Custom metric function '{custom_metric_fn.__name__}' returned None. Logging ignored."
```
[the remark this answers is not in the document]
MarkYHZhang: It would just return "<lambda>" as its "name".

mlflow/models/evaluation/base.py
```diff
@@ -461,7 +461,17 @@ def can_evaluate(self, *, model_type, evaluator_config, **kwargs) -> bool:
...
-    def evaluate(self, *, model, model_type, dataset, run_id, evaluator_config, **kwargs):
+    def evaluate(
+        self,
+        *,
+        model,
+        model_type,
+        dataset,
+        run_id,
+        evaluator_config,
+        custom_metric_fns=None,
```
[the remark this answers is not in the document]
jinzhang21: I'm generally not a big fan of allowing a mix of several types in a single function arg. i.e. for future extensibility, I'd rather prefer to have both `custom_metric_fns` and `custom_metrics` defined. On the other hand, allowing an input arg to have multiple types has been a tradition in MLflow (e.g. `dataset` can be a dataframe or ndarray). Purely for the sake of being internally consistent, I'm OK with using `custom_metrics` here.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +259,45 @@ def _do_plot(**kwargs):
...
+    ):
+        return result
+
+    raise MlflowException(
+        f"Custom metric '{custom_metric.__name__}' did not return in an expected format."
+        f"The two acceptable return types are:"
+        f"1. Dict[AnyStr, Union[int, float, np.number]: a dictionary of metrics"
+        f"2. Tuple[Dict[AnyStr, Union[int, float, np.number]], Dict[AnyStr, Any]]: a"
+        f"   dictionary of metrics and a dictionary of artifacts."
+        f"For more, check out the docstring for mlflow.evaluate"
```
apurva-koti, line 297:
```suggestion
        f"Custom metric '{custom_metric.__name__}' did not return an expected format."
```

apurva-koti (commented):
Looks good! Let's use `[]` as the default argument for `custom_metrics` rather than `None`.

MarkYHZhang:
> Looks good! Let's use `[]` as the default argument for `custom_metrics` rather than `None`.

I think we should avoid using mutable default argument in python due to issues described [here](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/). Although, I could initialize it to a list inside the constructor, but considering other default arguments like https://github.com/mlflow/mlflow/blob/2c0d13db6c677f1023a9a2a1776aca9e32aebe3d/mlflow/models/evaluation/base.py#L655
are not initialized, I think it might be good idea just to leave it as is.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Is it bad practice to have long exception messages? Are users expected to refer to the docstring?

apurva-koti: Just seemed like a duplication of information that could get annoying to maintain since it's not in an obvious place.
I think we can instead link them to the actual docs once they're published.

apurva-koti:
Ah good catch @MarkYHZhang ! Let's keep it `None` then.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,6 +580,18 @@ def plot_pr_curve():
...
+    def _evaluate_custom_metrics(self):
+        for custom_metric in self.custom_metrics:
+            # recreating the dataframe for each custom metric function call,
+            # in case the user modifies the eval_df inside their custom function.
+            eval_df = pd.DataFrame({"prediction": self.y_pred, "target": self.y})
+            metric_results, _ = _evaluate_custom_metric(custom_metric, eval_df, self.metrics)
+            # skip logging metric functions that doesn't return anything
+            if metric_results is None:
+                continue
+            self.metrics.update(metric_results)
```
apurva-koti, line 592:
```suggestion
            if metric_results is not None:
                self.metrics.update(metric_results)
```

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Make sense!

tests/models/test_default_evaluator.py
```diff
@@ -499,3 +503,155 @@ def test_gen_multiclass_roc_curve():
...
+        with pytest.raises(
+            MlflowException,
+            match=(
+                re.escape(
+                    f"Custom metric '{test_fn.__name__}' did not return in an expected format."
+                    f"The two acceptable return types are:"
+                    f"1. Dict[AnyStr, Union[int, float, np.number]: a dictionary of metrics"
+                    f"2. Tuple[Dict[AnyStr, Union[int, float, np.number]], Dict[AnyStr, Any]]: a"
+                    f"   dictionary of metrics and a dictionary of artifacts."
+                    f"For more, check out the docstring for mlflow.evaluate"
```
apurva-koti, line 553: We should only assert on the exact substring that we know will probably persist, like `did not return in an expected format.` Maybe the fn name also.

apurva-koti (approved):
LGTM once all comments addressed!

mlflow/models/evaluation/base.py
```diff
@@ -778,6 +798,70 @@ def evaluate(
...
+                                      ...
+                                      metrics: Dict[AnyStr, Union[int, float, np.number]] = ???
+                                      artifacts: Dict[AnyStr, Any] = ???
+                                      ...
+                                      if artifacts is not None:
+                                          return metrics, artifacts
+                                      return metrics
+                                  ``
+
+                              Example Usage:
```
dbczumar, line 845: Can we use `.. code-block: python`, as in https://github.com/mlflow/mlflow/blob/2c0d13db6c677f1023a9a2a1776aca9e32aebe3d/mlflow/tracking/fluent.py#L208-L235?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +259,38 @@ def _do_plot(**kwargs):
 }
 
 
+def _evaluate_custom_metric(custom_metric, eval_df, builtin_metrics):
```
dbczumar, line 262: Can we add a short docstring for this internal method?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,6 +573,17 @@ def plot_pr_curve():
 
         self._log_pandas_df_artifact(per_class_metrics_collection_df, "per_class_metrics")
 
+    def _evaluate_custom_metrics(self):
+        for custom_metric in self.custom_metrics:
+            # recreating the dataframe for each custom metric function call,
+            # in case the user modifies the eval_df inside their custom function.
```
dbczumar, line 579: in theory, it's possible for `self.y_pred` / `self.y` to be an object that gets modified during `_evaluate_custom_metric`. Embedding the object in a new pandas dataframe won't help in that case. Perhaps we want to use `deepcopy()` on `self.y_pred` and `self.y`.

mlflow/models/evaluation/base.py
```diff
@@ -778,6 +798,70 @@ def evaluate(
...
+
+                              Object types that artifacts be represented as:
+                              - A string uri representing the file path to the artifact. MLflow will
+                                infer the type of the artifact based on the file extension.
+                              - ``EvaluationArtifact`` type objects. i.e. CsvEvaluationArtifact,
+                                ImageEvaluationArtifact.
+                              - Pandas DataFrame. This will be resolved as a CSV artifact.
+                              - Numpy array. This will be saved as a .npy artifact.
+                              - Matplotlib Figure. This will be saved as an image artifact.
+                              - Any JSON serializable object. This will be saved as a JSON artifact.
```
dbczumar, line 831: We have a branch cut for MLflow 1.24.0 coming up within the next 1-2 days. In case we want to include support for numerical metrics in 1.24.0 and follow up with artifacts in 1.25.0, can we remove this portion of the docstring since artifacts aren't yet supported?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +259,38 @@ def _do_plot(**kwargs):
...
+    def __validate_artifacts_dict(artifacts):
+        return isinstance(artifacts, dict) and all(
+            isinstance(artifacts_name, str) for artifacts_name in artifacts.keys()
+        )
+
+    if isinstance(result, dict) and __validate_metrics_dict(result):
+        return result, None
+
+    if (
+        isinstance(result, tuple)
```
dbczumar, line 283: Shall we check the length of the tuple to make sure that it's `2`?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +259,38 @@ def _do_plot(**kwargs):
...
+        and __validate_metrics_dict(result[0])
+        and __validate_artifacts_dict(result[1])
+    ):
+        return result
+
+    raise MlflowException(
+        f"Custom metric '{custom_metric.__name__}' did not return in an expected format."
```
dbczumar, line 290: Can we include `error_code=INVALID_PARAMETER_VALUE` here, indicating that the supplied `custom_metric` is invalid?

dbczumar, line 290: Can we link to the docs for expected input formats and / or provide a brief summary of the supported formats?

apurva-koti: ^This existed but I asked @MarkYHZhang to take it out to avoid duplicating information that's hard to maintain (because this particular error isn't in an obvious place).
We will link to the docs once they're published

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +259,38 @@ def _do_plot(**kwargs):
 }
 
 
+def _evaluate_custom_metric(custom_metric, eval_df, builtin_metrics):
+    result = custom_metric(eval_df, builtin_metrics)
+    if result is None:
+        _logger.warning(f"Custom metric function '{custom_metric.__name__}' returned None.")
```
dbczumar, line 265: Not all callables have a `__name__` attribute. Can we use `getattr(custom_metric, '__name__', <default>)` here and below instead? For <default> perhaps `"at index 0"`, `"at index 1"`, ... would be helpful, as it would tell the user the index of the custom metric function in the supplied `custom_metrics` list.

https://stackoverflow.com/a/33578357/11952869

dbczumar (commented):
@MarkYHZhang This looks great! Left a few small comments - should be ready to merge soon!

mlflow/models/evaluation/default_evaluator.py
dbczumar: @apurva-koti We should link to the docs in this PR, else initial users of the API won't see this helpful doc link in the error message. @MarkYHZhang can you include a link to the expected docs location that will exist once this PR is merged and docs are published? Happy to help.

apurva-koti: Honestly it's probably fine to restore your earlier message describing the expected types @MarkYHZhang and then just replace it with a docs link when they're published. That way there's never an information gap

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Good catch!

mlflow/models/evaluation/base.py
MarkYHZhang: I think I should also modify the example to exclude artifacts support.

dbczumar: 👍

mlflow/models/evaluation/default_evaluator.py
dbczumar: Can we include a link to the docs? We just need to link to `mlflow.evaluate()`. We already do this here: https://mlflow.org/docs/latest/models.html#model-evaluation.

mlflow/models/evaluation/base.py
MarkYHZhang: I have more written more detailed documentation along with example in the function docstring of the mlflow.evaluate function here https://github.com/mlflow/mlflow/blob/7b26a88d51d59ece67aca3a8db6a0140baf65822/mlflow/models/evaluation/base.py#L801-L839

dbczumar (commented):
@MarkYHZhang Do we plan to add an example to the `examples` section or extend one of the existing examples? Can we do that as part of this PR?

mlflow/models/evaluation/base.py
```diff
@@ -778,6 +798,45 @@ def evaluate(
...
+                              - Dict: The second is a dictionary containing the metrics calculated
+                                by the default evaluator. The keys are the names of the metrics
+                                and the values are the scalar values of the metrics. Refer to the
+                                DefaultEvaluator behavior section for what metrics will be returned
+                                based on the type of model (i.e. classifier or regressor).
+
+                              A custom metric function can return in two different formats:
+                              - Dict[AnyStr, Union[int, float, np.number]: a singular dictionary of
+                                custom metrics, where the keys are the names of the metrics, and the
+                                values are the scalar values of the metrics.
```
dbczumar, line 817: Can we add spaces between the list titles and the elements? The current rendering is probably not quite what we intend:

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
 }
 
 
+def _evaluate_custom_metric(index, custom_metric, eval_df, builtin_metrics):
+    """
+    This function calls the `custom_metric` function and performs validations on the returned
+    result to ensure that they are in the expected format. It will raise a MlflowException if
+    the result is not in the expected format.
+
+    :param index
```
dbczumar, line 270: Can we add a docstring for this parameter?

mlflow/models/evaluation/base.py
```diff
@@ -778,6 +798,45 @@ def evaluate(
...
+                                  def squared_diff_plus_one(eval_df, builtin_metrics):
+                                      return {
+                                          "sdiff_plus_one": (np.sum(np.abs(eval_df["prediction"]
+                                                            - eval_df["target"]+1)**2))
+                                      }
+
+                                  with mlflow.start_run():
+                                      mlflow.evaluate(
+                                          model, X, targets, custom_metrics=[
+                                          squared_diff_plus_one,...])
```
dbczumar, line 838: Nit: Can we clean up the formatting here? Perhaps running this through `black` would be helpful.

dbczumar (approved):
LGTM when small docs nits have been addressed and example has been included. Thanks @MarkYHZhang ! Awesome stuff!

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
...
+    :param custom_metric: A user provided custom metric function
+    :param eval_df: A Pandas dataframe object containing a prediction and a target column
+    :param builtin_metrics: A dictionary of metrics produced by the default evaluator
+    :return: A tuple of dictionaries. The first is a dictionary of metrics, the second is
+             a dictionary of artifacts (which can be null if the custom metric function did
+             not produce any).
```
harupy, line 276:
```suggestion
    :return: A tuple of dictionaries. The first is a dictionary of metrics, the second is
             a dictionary of artifacts (which can be None if the custom metric function did
             not produce any).
```

Can we replace `null` with `None`? or are we using `null` intentionally?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
...
+    result = custom_metric(eval_df, builtin_metrics)
+    if result is None:
+        _logger.warning(
+            f"Custom metric function "
+            f"'{getattr(custom_metric, '__name__', repr(custom_metric))}' at index "
+            f"{index} in the `custom_metrics` parameter returned None."
+        )
+        return None, None
+
+    def __validate_metrics_dict(metrics):
```
jinzhang21, line 287: __validate_metrics_dict -> _validate_metrics

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
...
+
+    if (
+        isinstance(result, tuple)
+        and len(result) == 2
+        and __validate_metrics_dict(result[0])
+        and __validate_artifacts_dict(result[1])
+    ):
+        return result
+
+    raise MlflowException(
```
jinzhang21, line 309: The way the exception is raised is not informative. If I'm developing a custom metrics function, I have no way to know where it went wrong other than comparing against the documentation in mlflow.evaluate() very carefully. I added a few comments on where an exception should be raised above.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
...
+
+    def __validate_metrics_dict(metrics):
+        return isinstance(metrics, dict) and all(
+            isinstance(metric_name, str) and isinstance(metric_val, (int, float, np.number))
+            for metric_name, metric_val in metrics.items()
+        )
+
+    def __validate_artifacts_dict(artifacts):
+        return isinstance(artifacts, dict) and all(
+            isinstance(artifacts_name, str) for artifacts_name in artifacts.keys()
```
jinzhang21, line 295: If any of these condition is not true, raise in place with proper error message.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
...
+            f"Custom metric function "
+            f"'{getattr(custom_metric, '__name__', repr(custom_metric))}' at index "
+            f"{index} in the `custom_metrics` parameter returned None."
+        )
+        return None, None
+
+    def __validate_metrics_dict(metrics):
+        return isinstance(metrics, dict) and all(
+            isinstance(metric_name, str) and isinstance(metric_val, (int, float, np.number))
+            for metric_name, metric_val in metrics.items()
```
jinzhang21, line 290: If any of these condition is not true, raise in place with proper error message.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
...
+        )
+
+    def __validate_artifacts_dict(artifacts):
```
jinzhang21, line 293: __validate_artifacts_dict -> _validate_artifacts

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,6 +596,19 @@ def plot_pr_curve():
...
         self._log_pandas_df_artifact(per_class_metrics_collection_df, "per_class_metrics")
 
+    def _evaluate_custom_metrics(self):
+        for i, custom_metric in enumerate(self.custom_metrics):
+            # recreating the dataframe for each custom metric function call,
+            # in case the user modifies the eval_df inside their custom function.
+            eval_df = pd.DataFrame(
+                {"prediction": copy.deepcopy(self.y_pred), "target": copy.deepcopy(self.y)}
+            )
+            metric_results, _ = _evaluate_custom_metric(i, custom_metric, eval_df, self.metrics)
```
jinzhang21, line 606: Why is user modifying self.metrics not concerning?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,6 +596,19 @@ def plot_pr_curve():
 
         self._log_pandas_df_artifact(per_class_metrics_collection_df, "per_class_metrics")
 
+    def _evaluate_custom_metrics(self):
+        for i, custom_metric in enumerate(self.custom_metrics):
+            # recreating the dataframe for each custom metric function call,
+            # in case the user modifies the eval_df inside their custom function.
+            eval_df = pd.DataFrame(
+                {"prediction": copy.deepcopy(self.y_pred), "target": copy.deepcopy(self.y)}
+            )
```
jinzhang21, line 605: This could be expensive. I recommend to only do it once outside of the loop.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -258,6 +261,59 @@ def _do_plot(**kwargs):
...
+    :param index
+    :param custom_metric: A user provided custom metric function
+    :param eval_df: A Pandas dataframe object containing a prediction and a target column
+    :param builtin_metrics: A dictionary of metrics produced by the default evaluator
+    :return: A tuple of dictionaries. The first is a dictionary of metrics, the second is
+             a dictionary of artifacts (which can be null if the custom metric function did
+             not produce any).
+    """
+    result = custom_metric(eval_df, builtin_metrics)
+    if result is None:
```
jinzhang21, line 279: We raise on incorrect return format from `custom_metric` function. Why is None considered special here since it's not a correct format?

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Right, thanks!

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Right, that's a good point.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Oh, because I was following this https://docs.google.com/document/d/1lC2PMENiT4G7o-QklRdeukGBON_hv9D47G1Q5GCt3GE/edit#bookmark=id.cod7z43dnwqy. I realized I didn't state this in the design document, should I just remove this behavior?

mlflow/models/evaluation/base.py
MarkYHZhang: Fixed! 
image

mlflow/models/evaluation/base.py
```diff
@@ -590,7 +593,14 @@ def _get_last_failed_evaluator():
...
 def _evaluate(
-    *, model, model_type, dataset, run_id, evaluator_name_list, evaluator_name_to_conf_map
+    *,
+    model,
+    model_type,
+    dataset,
+    run_id,
+    evaluator_name_list,
+    evaluator_name_to_conf_map,
+    custom_metrics,
```
jinzhang21, line 603: Rm ","

mlflow/models/evaluation/base.py
```diff
@@ -655,6 +666,7 @@ def evaluate(
     feature_names: list = None,
     evaluators=None,
     evaluator_config=None,
+    custom_metrics=None,
```
jinzhang21, line 669: rm ","

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,6 +614,21 @@ def plot_pr_curve():
...
+    def _evaluate_custom_metrics(self):
+        if self.custom_metrics is None:
+            return
+        copy_of_y_pred = copy.deepcopy(self.y_pred)
+        copy_of_y = copy.deepcopy(self.y)
+        copy_of_metrics = copy.deepcopy(self.metrics)
+        for i, custom_metric in enumerate(self.custom_metrics):
+            # recreating the dataframe for each custom metric function call,
+            # in case the user modifies the eval_df inside their custom function.
+            eval_df = pd.DataFrame({"prediction": copy_of_y_pred, "target": copy_of_y})
```
jinzhang21, line 626: Will this step create another copy from copy_of_y_pred and copy_of_y (and a copy of dict is temporarily created)? Why not just moving this line outside of the loop?

tests/models/test_default_evaluator.py
```diff
@@ -499,3 +502,160 @@ def test_gen_multiclass_roc_curve():
 
     expected_auc = [0.75, 0.75, 0.3333]
     assert np.allclose(results.auc, expected_auc, rtol=1e-3)
+
+
+def test_evaluate_custom_metric_incorrect_return_formats():
```
jinzhang21, line 507: The test looks thorough, but it could be improved by using the parametrize method from pytest: https://docs.pytest.org/en/6.2.x/example/parametrize.html#parametrizing-conditional-raising

tests/models/test_default_evaluator.py
```diff
@@ -499,3 +502,160 @@ def test_gen_multiclass_roc_curve():
...
+    eval_df = pd.DataFrame({"prediction": [1.2, 1.9, 3.2], "target": [1, 2, 3]})
+    metrics = _get_regressor_metrics(eval_df["target"], eval_df["prediction"])
+
+    def dummy_fn(*_):
```
jinzhang21, line 511: Also add a test on lambda function. e.g. legal case: `custom_metrics=(lambda x: {'square': x**2},)` and illegal case: `custom_metrics=(lambda x: x,)`

mlflow/models/evaluation/base.py
harupy: This "," is automatically inserted by black. I'm curious why we need to remove it.

harupy: Is it for preventing this issue?
https://stackoverflow.com/questions/33350454/syntaxerror-with-passing-kwargs-and-trailing-comma

tests/models/test_default_evaluator.py
```diff
@@ -499,3 +502,160 @@ def test_gen_multiclass_roc_curve():
...
+    )
+    assert res_metrics_2 == {
+        "example_count_times_1_point_5": metrics["example_count"] * 1.5,
+        "sum_on_label_minus_5": metrics["sum_on_label"] - 5,
+        "example_np_metric_1": np.float32(123.2),
+        "example_np_metric_2": np.ulonglong(10000000),
+    }
+    assert "pred_target_abs_diff" in res_artifacts_2 and res_artifacts_2[
+        "pred_target_abs_diff"
+    ].equals(np.abs(eval_df["prediction"] - eval_df["target"]))
```
harupy, line 610:
```suggestion
    assert "pred_target_abs_diff" in res_artifacts_2
    assert res_artifacts_2[
        "pred_target_abs_diff"
    ].equals(np.abs(eval_df["prediction"] - eval_df["target"]))
```

Can we split this assertion to simplify the assertion error message?

### Example

```python
def test_1():
    res_artifacts_2 = {"example_dictionary_artifact": {"a": 1, "b": 3}}
    assert "example_dictionary_artifact" in res_artifacts_2
    assert res_artifacts_2["example_dictionary_artifact"] == {"a": 1, "b": 2}


"""
# pytest error message

E       AssertionError: assert {'a': 1, 'b': 3} == {'a': 1, 'b': 2}
E         Omitting 1 identical items, use -vv to show
E         Differing items:
E         {'b': 3} != {'b': 2}
E         Full diff:
E         - {'a': 1, 'b': 2}
E         ?               ^
E         + {'a': 1, 'b': 3}...
E         
E         ...Full output truncated (2 lines hidden), use '-vv' to show
"""


def test_2():
    res_artifacts_2 = {"example_dictionary_artifact": {"a": 1, "b": 3}}
    assert "example_dictionary_artifact" in res_artifacts_2 and res_artifacts_2[
        "example_dictionary_artifact"
    ] == {"a": 1, "b": 2}


"""
# pytest error message

E       AssertionError: assert ('example_dictionary_artifact' in {'example_dictionary_artifact': {'a': 1, 'b': 3}} and {'a': 1, 'b': 3} == {'a': 1, 'b': 2}
E         Omitting 1 identical items, use -vv to show
E         Differing items:
E         {'b': 3} != {'b': 2}
E         Full diff:
E         - {'a': 1, 'b': 2}
E         ?               ^
E         + {'a': 1, 'b': 3}...
E         
E         ...Full output truncated (2 lines hidden), use '-vv' to show)
"""

```

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Actually I just remembered that Corey brought up a good point [here](https://github.com/mlflow/mlflow/pull/5389#discussion_r813242477) that each custom metric function could modify `eval_df`. So it might be a good idea to perform deep copy on each. Namely reverting back to:
```python
eval_df = {"prediction": copy.deepcopy(self.y_pred), "target": copy.deepcopy(self.y)}
```

tests/models/test_default_evaluator.py
MarkYHZhang: Oh that's interesting, I didn't know about the parametrize method. But I wonder if I should use it here because this test only contain cases that raise exceptions. And seems like from the pytest documentation referenced above parametrize method is used in tests that contain cases that raise and don't raise exceptions. Perhaps did you mean I should rewrite my tests here so that they contain both exception-raising and exception-not-raising cases?

jinzhang21: No need to rewrite if it's difficult to factor them out. Leave them as is is fine.

mlflow/models/evaluation/base.py
jinzhang21: I recall there was some issue with it long time ago but don't remember exactly why. If black adds it, and it runs fine, I'm cool with it.

jinzhang21 (approved):

harupy (approved):

PR #5405, opened 2022-02-23
[Custom Metrics] Artifact type detection and logging

MarkYHZhang:

<del>
## Note
Updates to examples and further documentations will be done in a separate PR due to the size of this PR.
</del>

## A notable implementation detail:
When a user's custom metric function produces an artifact file and pass its local path, for example `/some/dir/image.csv` with the artifact name `someplot` for a dataset `example_dataset`. The behavior described in this PR is to log the artifact with the following file name on an artifact repository:
```
someplot_on_data_example_dataset.csv
``` 
This is the convention followed by all of the internally generated artifacts (from DefaultEvaluator). An implementation detail to note is that we first make a copy of the artifact file supplied by the user and save it to a temporary directory. Then, we rename the copy of that artifact to the defined format:
https://github.com/mlflow/mlflow/blob/2c0d13db6c677f1023a9a2a1776aca9e32aebe3d/mlflow/models/evaluation/default_evaluator.py#L77
Finally, we upload the copy through `mlflow.log_artifact`. Although it is possible to directly upload to the repository from the user provided artifact file by modifying the implementation of `log_artifact` for every artifact repository. But I think that requires an unnecessary amount of changes. Let me know if there are better ways of handling this. Thanks!

Added function for custom metrics to support auto detection and loading of user produced artifacts.

## How is this patch tested?

Wrote many unit tests in mlflow/tests/models/test_default_evaluator.py

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Extends custom metrics support for `mlflow.evaluate` first introduced in https://github.com/mlflow/mlflow/pull/5389 by enabling the ability to detect and load artifacts produced by user's custom metric functions.

Components 

Interface 

Language 

Integrations

Tagging: @apurva-koti @dbczumar @jinzhang21 @WeichenXu123.


apurva-koti (commented):
Nice tests!

mlflow/models/evaluation/artifacts.py
```diff
@@ -21,3 +23,30 @@ def _save(self, output_artifact_path):
...
+class ParquetEvaluationArtifact(EvaluationArtifact):
+    def _save(self, output_artifact_path):
+        self._content.to_parquet(output_artifact_path, compression="brotli")
+
+    def _load_content_from_file(self, local_artifact_path):
+        self._content = pd.read_parquet(local_artifact_path)
+        return self._content
+
+
+class NpyEvaluationArtifact(EvaluationArtifact):
```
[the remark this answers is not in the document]
MarkYHZhang: Good idea.

mlflow/models/evaluation/artifacts.py
```diff
@@ -21,3 +23,30 @@ def _save(self, output_artifact_path):
...
+        np.save(output_artifact_path, self._content, allow_pickle=False)
+
+    def _load_content_from_file(self, local_artifact_path):
+        self._content = np.load(local_artifact_path, allow_pickle=False)
+        return self._content
+
+
+class JsonEvaluationArtifact(EvaluationArtifact):
+    def _save(self, output_artifact_path):
+        json.dump(self._content, open(output_artifact_path, "w"))
```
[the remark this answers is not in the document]
MarkYHZhang: I think the only difference between `w` and `w+` is that `w+` allows reading, source: https://mkyong.com/python/python-difference-between-r-w-and-a-in-open.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,95 @@ def __validate_artifacts(artifacts):
...
+
+    type_to_artifact_map = {
+        pd.DataFrame: CsvEvaluationArtifact,
+        np.ndarray: NpyEvaluationArtifact,
+        pyplot.Figure: ImageEvaluationArtifact,
+    }
+
+    artifact_file_name = (
+        _gen_log_key(artifact_name, dataset_name)
+        + f"{type_to_ext_map.get(type(raw_artifact), '.json')}"
```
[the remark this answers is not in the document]
MarkYHZhang: Did you mean a `_` before `artifact_file_name`? I used `.get` because in the event that the type of `raw_artifact` does not exist in `type_to_ext_map`, it will default to use the `.json` extension. I realized I don't really need the string formatting, it should be simplified to:
```python
artifact_file_name = _gen_log_key(artifact_name, dataset_name) + type_to_ext_map.get(type(raw_artifact), '.json')
```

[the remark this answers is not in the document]
MarkYHZhang: Did you mean a `_` before `artifact_file_name`? I used `.get` because in the event that the type of `raw_artifact` does not exist in `type_to_ext_map`, it will default to use the `.json` extension. I realized I don't really need the string formatting, it should be simplified to:
```python
artifact_file_name = _gen_log_key(artifact_name, dataset_name) + type_to_ext_map.get(type(raw_artifact), '.json')
```

mlflow/models/evaluation/artifacts.py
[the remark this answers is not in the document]
apurva-koti: 🤦 Of course we're trying to create this file lol. Never mind me

mlflow/models/evaluation/default_evaluator.py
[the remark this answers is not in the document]
apurva-koti: Nvm, looked through `gen_log_key`. We don't want to separate the generated name from the extension :)

tests/models/test_default_evaluator.py
```diff
@@ -15,11 +22,13 @@
     _get_classifier_per_class_metrics,
     _gen_classifier_curve,
     _evaluate_custom_metric,
+    _load_custom_metric_artifact,
 )
 import mlflow
 from sklearn.linear_model import LogisticRegression
 
 # pylint: disable=unused-import
+from mlflow.utils.file_utils import TempDir
```
harupy, line 31: Can we use pytest's `tmpdir` fixture instead?
https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture

MarkYHZhang: I think the usage of `tmpdir` [is no longer recommended](https://py.readthedocs.io/en/latest/path.html#:~:text=Note%3A%20The%20%E2%80%98py%E2%80%99%20library%20is%20in%20%E2%80%9Cmaintenance%20mode%E2%80%9D%20and%20so%20is%20not%20recommended%20for%20new%20projects.%20Please%20check%20out%20pathlib%20or%20pathlib2%20for%20path%20operations.) since it is based on a `py.path` object. Also, `TempDir` is a parameter that the new function `_load_custom_metric_artifact` expects from calls from an `DefaultEvaluator` object. I think I could use a [`tmp_path`](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture) fixture to provide a separate temporary path for the files the tests generate. And that will co-exist with the import of `TempDir` from MLflow.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,94 @@ def __validate_artifacts(artifacts):
...
+    )
+
+    artifact_file_local_path = temp_dir.path(artifact_file_name)
+
+    if isinstance(raw_artifact, pd.DataFrame):
+        raw_artifact.to_csv(artifact_file_local_path, index=False)
+    elif isinstance(raw_artifact, np.ndarray):
+        np.save(artifact_file_local_path, raw_artifact, allow_pickle=False)
+    elif isinstance(raw_artifact, pyplot.Figure):
+        raw_artifact.savefig(artifact_file_local_path)
```
jinzhang21, line 412: This is probably out of scope for this PR, but have we discussed the option whether to save the entire figure object for interactive usage later? @WeichenXu123 @dbczumar

apurva-koti (approved):

tests/models/test_default_evaluator.py
harupy: @MarkYHZhang Thanks for the explanation! I didn't know `tmpdir` is no longer recommended. Using `tmp_path` sounds good to me.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,95 @@ def __validate_artifacts(artifacts):
...
+
+    exception_header = _get_custom_metric_exception_header(custom_metric_index, custom_metric_name)
+
+    # If path is str, convert to pathlib.Path. Then, type inference through file extension
+    if isinstance(raw_artifact, str):
+        raw_artifact = pathlib.Path(raw_artifact)
+    if isinstance(raw_artifact, pathlib.Path):
+        ext_to_artifact_map = {
+            ".png": ImageEvaluationArtifact,
+            ".jpg": ImageEvaluationArtifact,
```
BenWilson2, line 364: Should we support `.svg` format here for an additional `ImageEvaluationArtifact`? It's pretty widely used.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,95 @@ def __validate_artifacts(artifacts):
...
+    )
+
+    artifact_file_local_path = temp_dir.path(artifact_file_name)
+
+    if isinstance(raw_artifact, pd.DataFrame):
+        raw_artifact.to_csv(artifact_file_local_path, index=False)
+    elif isinstance(raw_artifact, np.ndarray):
+        np.save(artifact_file_local_path, raw_artifact, allow_pickle=False)
+    elif isinstance(raw_artifact, pyplot.Figure):
+        raw_artifact.savefig(artifact_file_local_path)
```
BenWilson2, line 413: If we're not permitting `**kwargs` here, we should probably mention in the docs that the only way to modify (globally) any generated plots (dpi, scaling, whitespace bbox, padding, format) during `.savefig()` is through setting values to the env variable dictionary `rcParams`.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,95 @@ def __validate_artifacts(artifacts):
     )
 
 
+def _load_custom_metric_artifact(
```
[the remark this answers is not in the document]
MarkYHZhang: Ah yeah, when I first named it I thought of it as loading an `raw_artifact` into an `EvaluationArtifact` object. But you're right, given the operations done in the method, it's probably better to name it as `_log_custom_metric_artifact`.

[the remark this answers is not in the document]
MarkYHZhang: I will split this function into a static function `_infer_artifact_type_and_ext` and a member function `_log_custom_metric_artifact` of the `DefaultEvaluator`. To test `_log_custom_metric_artifact`, I think I will need to update `_evaluate_custom_metrics` to include saving of artifacts, and then call `evaluate` to perform the required tests.

tests/models/test_default_evaluator.py
```diff
@@ -684,3 +693,221 @@ def example_custom_metric(eval_df, given_metrics):
 
     assert "positive_count" in result.metrics
     assert np.isclose(result.metrics["positive_count"], np.sum(y_pred), rtol=1e-3)
+
+
+def test_load_custom_metric_artifact_returns_given_evaluation_artifact_as_is():
```
dbczumar, line 698: Rather than testing a private / internal function, can we instead structure these as tests of the `mlflow.evaluate()` public API with `evaluators="default"`?

MarkYHZhang: Yes, I am doing that refactor right now. Will update this PR soon :)

MarkYHZhang: Yes, it was intentional that nowhere else calls `_load_custom_metric_artifact` in this PR. Because my original intention was to have a follow up PR that calls this function and to connect it with `mlflow.evaluate`. But now it seems like it would make more sense to have that follow up PR combined with this.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Good point, I will note this down in the docs. If users want customizations to their figure, they could opt to call `savefig()` first with their desired arguments, then return the file path of their image artifact in their custom metric function.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: `pillow` currently does not support `.svg` files as documented [here](https://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html). We could extend support for this by adding additional dependencies like `cairosvg` or `svglib` to MLflow as suggested [here](https://stackoverflow.com/questions/15130670/pil-and-vectorbased-graphics). Should we add such dependencies?

mlflow/models/evaluation/base.py
```diff
@@ -793,52 +793,82 @@ def evaluate(
...
-                           - Dict[AnyStr, Union[int, float, np.number]: a singular dictionary of
+                           - ``Dict[AnyStr, Union[int, float, np.number]``: a singular dictionary of
                              custom metrics, where the keys are the names of the metrics, and the
                              values are the scalar values of the metrics.
+                           - ``Tuple[Dict[AnyStr, Union[int,float,np.number]], Dict[AnyStr,Any]]``:
+                             a tuple of a dict containing the custom metrics, and a dict of
+                             artifacts, where the keys are the names of the artifacts, and the
+                             values are objects representing the artifacts.
+
+                           Object types that artifacts be represented as:
```
BenWilson2, line 817:
```suggestion
                           Object types that artifacts can be represented as:
```

mlflow/models/evaluation/default_evaluator.py
BenWilson2: Perhaps not as part of this PR.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,48 @@ def __validate_artifacts(artifacts):
     )
 
 
+def _infer_artifact_type_and_ext(artifact_name, raw_artifact, exception_header):
+    # Given a path-like object: a string representation of path, or a pathlib.Path object
+    if isinstance(raw_artifact, str):
+        raw_artifact = pathlib.Path(raw_artifact)
+    if isinstance(raw_artifact, pathlib.Path):
+        ext_to_artifact_map = {
```
BenWilson2, line 355: should these enumeration dictionaries be stored externally to these functions (perhaps as constants in `mlflow.models.evaluation.artifacts` so that they could be easily referenced elsewhere?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,48 @@ def __validate_artifacts(artifacts):
...
+            ".npy": NumpyEvaluationArtifact,
+            ".csv": CsvEvaluationArtifact,
+            ".parquet": ParquetEvaluationArtifact,
+        }
+        if raw_artifact.is_file() and raw_artifact.suffix in ext_to_artifact_map.keys():
+            return True, ext_to_artifact_map[raw_artifact.suffix], raw_artifact.suffix
+        raise MlflowException(
+            f"{exception_header} produced an unsupported artifact '{artifact_name}' with path "
+            f"'{raw_artifact}'. The supported file extensions are:"
+            f".csv, .jpeg, .jpg, .png, .json, .npy, .parquet"
```
BenWilson2, line 369:
```suggestion
            f"{', '.join(ext_to_artifact_map.keys())}"
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +347,48 @@ def __validate_artifacts(artifacts):
...
+    type_to_ext_map = {
+        pd.DataFrame: ".csv",
+        np.ndarray: ".npy",
+        pyplot.Figure: ".png",
+    }
+    type_to_artifact_map = {
+        pd.DataFrame: CsvEvaluationArtifact,
+        np.ndarray: NumpyEvaluationArtifact,
+        pyplot.Figure: ImageEvaluationArtifact,
+    }
```
BenWilson2, line 384: Similar to above. Could we extract these constant enum-like dicts externally?

apurva-koti (commented):
I think the tests are a bit hard to parse. Where relevant (ie for repeated code with individual bits changed like type, expected return, etc) is it possible to define a generic test function that runs over several arguments?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -347,15 +347,8 @@ def __validate_artifacts(artifacts):
...
-def _load_custom_metric_artifact(
-    dataset_name, temp_dir, artifact_name, raw_artifact, custom_metric_index, custom_metric_name
-):
-    if isinstance(raw_artifact, EvaluationArtifact):
-        return raw_artifact
-
-    exception_header = _get_custom_metric_exception_header(custom_metric_index, custom_metric_name)
-
-    # If path is str, convert to pathlib.Path. Then, type inference through file extension
+def _infer_artifact_type_and_ext(artifact_name, raw_artifact, exception_header):
```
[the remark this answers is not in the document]
MarkYHZhang: Just added docstring earlier this morning :).

BenWilson2:
> > I think the tests are a bit hard to parse. Where relevant (ie for repeated code with individual bits changed like type, expected return, etc) is it possible to define a generic test function that runs over several arguments?
> 
> Just cleaned up all of the `_infer_artifact_type_and_ext` tests by rewriting them with [`pytest.parametrize`](https://docs.pytest.org/en/6.2.x/parametrize.html). While I think it is possible to shorten the `evaluate` tests for custom metrics, I feel like it would cause those tests to be not self-contained. Which I think could be a readability problem. Let me know if you have any suggestions though!

The only opportunity would be to create a fixture for generating the image but that doesn't solve the usage pattern in the last test case. I think the parameterization that you did in the one test looks great. Good job @MarkYHZhang !

BenWilson2 (approved):
Nice work! LGTM!

tests/models/test_default_evaluator.py
```diff
@@ -636,25 +646,110 @@ def example_custom_metric_with_artifacts(given_df, given_metrics):
...
+        artifact_representation = artifact
+    inferred_from_path, inferred_type, inferred_ext = _infer_artifact_type_and_ext(
+        f"{ext}_{artifact_type.__name__}_artifact", artifact_representation, ""
+    )
+    assert not (is_file ^ inferred_from_path)
+    assert inferred_type is artifact_type
+    assert inferred_ext == f".{ext}"
+
+
+def _start_run_for_custom_metric_tests(model_uri, dataset, custom_metrics):
```
apurva-koti, line 714: Nit: I'd rename this function to something like `_get_results_for_custom_metrics_tests` since the start run isn't the main point

tests/models/test_default_evaluator.py
```diff
@@ -669,7 +764,6 @@ def example_custom_metric(eval_df, given_metrics):
         expected_metrics["true_negatives"] + expected_metrics["true_positives"],
         rtol=1e-3,
     )
-
```
apurva-koti, line 672: Nit: avoid noop changes

tests/models/test_default_evaluator.py
```diff
@@ -681,6 +775,158 @@ def example_custom_metric(eval_df, given_metrics):
     assert np.isclose(
         metrics["positive_count_on_data_breast_cancer_dataset"], np.sum(y_pred), rtol=1e-3
     )
-
```
apurva-koti, line 684: Nit: avoid noop changes

apurva-koti (approved):
LGTM!

tests/models/test_default_evaluator.py
MarkYHZhang: I removed the empty line because before I forgot to group these lines together. These lines tests for the same thing, and the assertions below, also follows this grouping pattern. I'll make sure to look out to avoid these next time!

tests/models/test_default_evaluator.py
MarkYHZhang: ditto above

tests/models/test_default_evaluator.py
apurva-koti: Oh got it, you wrote these 😂

mlflow/models/evaluation/base.py
```diff
@@ -793,52 +793,82 @@ def evaluate(
...
+                                           )
+                                       )
+                                   }
+
+                               def scatter_plot(eval_df, builtin_metrics):
+                                   plt.scatter(eval_df['prediction'], eval_df['target'])
+                                   plt.xlabel('Targets')
+                                   plt.ylabel('Predictions')
+                                   plt.title("Targets vs. Predictions")
+                                   plt.savefig("some/path/example.png")
```
jinzhang21, line 863: We need to make sure user can actually write to a path in custom function. Update it with a tempdir + "/example.png" or something functional as is.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +349,37 @@ def __validate_artifacts(artifacts):
     )
 
 
+def _infer_artifact_type_and_ext(artifact_name, raw_artifact, exception_header):
```
jinzhang21, line 352: This function is all about artifact type inference. I suggest to move it to mlflow/models/evaluation/artifacts.py instead.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +349,37 @@ def __validate_artifacts(artifacts):
...
+            f"'{raw_artifact}'. The supported file extensions are: "
+            f"{', '.join(EXT_TO_ARTIFACT_MAP.keys())}"
+        )
+
+    # Given as other python object
+    return (
+        False,
+        TYPE_TO_ARTIFACT_MAP.get(type(raw_artifact), JsonEvaluationArtifact),
+        TYPE_TO_EXT_MAP.get(type(raw_artifact), ".json"),
+    )
```
dbczumar, line 380: Can we create a `namedtuple` for this? `namedtuples` are more self-documenting and are easier to maintain.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -614,7 +662,53 @@ def plot_pr_curve():
...
+        artifact_file_local_path = self.temp_dir.path(artifact_file_name)
+
+        if inferred_from_path:
+            shutil.copyfile(raw_artifact, artifact_file_local_path)
+        elif inferred_type is CsvEvaluationArtifact:
+            raw_artifact.to_csv(artifact_file_local_path, index=False)
+        elif inferred_type is NumpyEvaluationArtifact:
+            np.save(artifact_file_local_path, raw_artifact, allow_pickle=False)
+        elif inferred_type is ImageEvaluationArtifact:
+            raw_artifact.savefig(artifact_file_local_path)
```
dbczumar, line 688: Are we missing cases for `ParquetEvaluationArtifact` and `JsonArtifact` here? If not, can we add an inline comment explaining why (also, if these classes aren't necessary, can we delete them?)?

Are we missing corresponding test coverage?

tests/models/test_default_evaluator.py
```diff
@@ -636,25 +646,110 @@ def example_custom_metric_with_artifacts(given_df, given_metrics):
     assert res_artifacts_2["example_dictionary_artifact"] == {"a": 1, "b": 2}
 
 
-def test_custom_metric(binary_logistic_regressor_model_uri, breast_cancer_dataset):
-    def example_custom_metric(eval_df, given_metrics):
-        return {
-            "true_count": given_metrics["true_negatives"] + given_metrics["true_positives"],
-            "positive_count": np.sum(eval_df["prediction"]),
-        }
+def test_infer_artifact_type_and_ext_raise_exception_for_non_file_artifact_path(tmp_path):
```
dbczumar, line 649: Do we need to be testing these internal utils, or can we rely on integration tests of public API behavior?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -614,7 +662,53 @@ def plot_pr_curve():
...
+                    f"- Matplotlib Figure. This will be saved as an image artifact."
+                    f"- Any JSON serializable object. This will be saved as a JSON file."
+                )
+
+        mlflow.log_artifact(artifact_file_local_path)
+        artifact = inferred_type(uri=mlflow.get_artifact_uri(artifact_file_name))
+        artifact._load(artifact_file_local_path)
+        return artifact
+
+    def _log_custom_metrics(self):
```
jinzhang21, line 711: custom metrics are not logged in this function (it's done in`self. _log_metrics`). What happened inside is "evaluate custom functions" and "log custom artifacts". I suggest to separate this into 2 functions to make it really clear what's been done in each. Or at least name it accurately if separating the logic is complicated.

mlflow/models/evaluation/default_evaluator.py
dbczumar, line 352: Not a big fan of passing strings around as method parameters purely for exception content. Can we instead pass the custom metric and index (ideally these could be represented as a `namedtuple` as well) as a parameter to this method and use them to construct the exception content?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -614,7 +662,53 @@ def plot_pr_curve():
...
+                    f"type '{type(raw_artifact)}'. Supported object types for artifacts are:"
+                    f"- A string uri representing the file path to the artifact. MLflow"
+                    f"  will infer the type of the artifact based on the file extension."
+                    f"- EvaluationArtifact type objects. i.e. ImageEvaluationArtifact."
+                    f"- Pandas DataFrame. This will be resolved as a CSV artifact."
+                    f"- Numpy array. This will be saved as a .npy artifact."
+                    f"- Matplotlib Figure. This will be saved as an image artifact."
+                    f"- Any JSON serializable object. This will be saved as a JSON file."
+                )
+
```
dbczumar, line 705: Why is JSON our fallback case? Can't we save the string representation of the object as a text file? This seems like a better catch-all.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -708,7 +809,7 @@ def plot_confusion_matrix():
                 "confusion_matrix",
             )
 
-        self._evaluate_custom_metrics()
+        self._log_custom_metrics()
         self._log_metrics()
         self._log_model_explainability()
         return EvaluationResult(self.metrics, self.artifacts)
```
dbczumar, line 815: This logic is present in both `_evaluate_regressor` and `_evaluate_classifier`. Can we factor it out?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -332,6 +349,37 @@ def __validate_artifacts(artifacts):
...
+    # Given a path-like object: a string representation of path, or a pathlib.Path object
+    if isinstance(raw_artifact, str):
+        raw_artifact = pathlib.Path(raw_artifact)
+    if isinstance(raw_artifact, pathlib.Path):
+        if raw_artifact.is_file() and raw_artifact.suffix in EXT_TO_ARTIFACT_MAP.keys():
+            return True, EXT_TO_ARTIFACT_MAP[raw_artifact.suffix], raw_artifact.suffix
+        raise MlflowException(
+            f"{exception_header} produced an unsupported artifact '{artifact_name}' with path "
+            f"'{raw_artifact}'. The supported file extensions are: "
+            f"{', '.join(EXT_TO_ARTIFACT_MAP.keys())}"
```
dbczumar, line 372: This message only reflects the case where the artifact is of an unrecognized type (`raw_artifact.suffix not in EXT_TO_ARTIFACT_MAP.keys()`), not the case where the file doesn't exist or is a directory (`not raw_artifact.is_file()`). Shall we update it?

dbczumar (commented):
@MarkYHZhang Awesome stuff! Left a few comments. Can we also extend https://github.com/mlflow/mlflow/blob/master/examples/evaluation/evaluate_with_custom_metrics.py to include artifacts as part of this PR?

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: `ParquetEvaluationArtifact` isn't here because it can only be supplied through file paths. Specifically, if user supplies a `DataFrame` object, we default to saving it as a `CsvEvaluationArtifact`. `JsonEvaluationArtifact` isn't here because    the last `else` clause handles that case. And since `_infer_artifact_type_and_ext` defaults to `JsonEvaluationArtifact` if it can't match other types, this case is handled. I'll explain this in the code. We're not missing test coverage, if you refer to the test cases for both `_infer_artifact_type_and_ext` and `evaluate` with custom metrics, both have cases that tests for parquet and json artifacts.

tests/models/test_default_evaluator.py
MarkYHZhang: I feel like there's no harm in having more verbose tests, I think this makes it easier to quickly spot behavior changes in case of future changes to custom metrics. Also, it seems like there's existing tests that [test for internal functions](https://github.com/mlflow/mlflow/blob/91ed65c0a756758f3bb7cfa4d41071d5d1393962/tests/models/test_default_evaluator.py#L281-L505)

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: JSON is used as fallback because I think it is more structured and standardized than using string representation. For example, if a user have an arbitrary class that they want to serialize. They can either opt to provide an implementation to `__repr__` and define their own format (which most of the time I think it's a dictionary that is JSON serializable anyways), or they can instead take a more standardized and more established approach of making their class [JSON serializable](https://pythontic.com/serialization/json/introduction). 

I think what we could do is that we first try to JSON serialize it, if it fails, we save the string representation. But imo this is just passing down the problem. Because if their class doesn't have an implementation to `__repr__` they'll see that their object is "serialized" to something like `<__main__.some.package.TheirObject object at 0x7fb7b08ab290>`.

Using string is a better catch-all in the sense that it won't raise an exception when the user doesn't provide a way to serialize their object. I think if we want to be verbose and let them know immediately that their class isn't serializable, then we should stick with JSON. But if we want to be more accommodating, having string as an option could be a good idea.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: I can't think of a fitting function name. Especially after renaming `_log_custom_metrics` to `_evaluate_custom_metrics_and_log_produced_artifacts` as [suggested by Jin](https://github.com/mlflow/mlflow/pull/5405#discussion_r819277286). Does `_prepare_and_generate_evaluation_result` sound good?

mlflow/models/evaluation/default_evaluator.py
dbczumar: I'd rather we be flexible than rigid with what users can log. If the object is of a string type, can we write it out as txt. If the object is anything else, can we just pickle it? In either case, let's print a warning.

tests/models/test_default_evaluator.py
dbczumar: I don't see what these tests add by themselves. I'd rather we focus on the behavior of the user-facing `mlflow.evaluate()` API.

mlflow/models/evaluation/default_evaluator.py
dbczumar: I see. In that case, `ParquetEvaluationArtifact` and `JsonEvaluationArtifact` aren't used anywhere. Can we remove them?

dbczumar: Discussed offline, these are returned to users.

mlflow/models/evaluation/default_evaluator.py
dbczumar: `_log_and_return_evaluation_result`

docs/source/models.rst
```diff
@@ -922,7 +922,89 @@ and behavior:
 .. |eval_importance_img| image:: _static/images/model_evaluation_feature_importance.png
    :width: 69%
 
-More information about model evaluation behaviors and outputs is available in the
+
+Evaluating on Custom Metrics
```
dbczumar, line 926:
```suggestion
Evaluating with Custom Metrics
```

docs/source/models.rst
```diff
@@ -922,7 +922,89 @@ and behavior:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the default set of metrics is insufficient, you can specify a list of functions to the ``custom_metrics``
+argument of :py:func:`mlflow.evaluate()` to evaluate your models in ways that you see fit. Custom metric
+functions should accept two arguments: a DataFrame containing ``prediction`` and ``target`` columns, 
+and a dictionary of the default set of metrics. For a full list of default metrics, refer to the documentation
```
dbczumar, line 932:
```suggestion
and a dictionary containing the default set of metrics. For a full list of default metrics, refer to the documentation 
```

docs/source/models.rst
```diff
@@ -922,7 +922,89 @@ and behavior:
...
+functions should accept two arguments: a DataFrame containing ``prediction`` and ``target`` columns, 
+and a dictionary of the default set of metrics. For a full list of default metrics, refer to the documentation 
+of :py:func:`mlflow.evaluate()`.
+
+The following `short example from the MLflow GitHub Repository
+<https://github.com/mlflow/mlflow/blob/master/examples/evaluation/evaluate_with_custom_metrics.py>`_ 
+uses :py:func:`mlflow.evaluate()` with a custom metric function to evaluate the performance of a regressor on the
+`California Housing Dataset <https://www.dcc.fc.up.pt/~ltorgo/Regression/cal_housing.html>`_.
+Note that custom metric functions can return both metrics and artifacts. It can either return a single
+dictionary of metrics, or two dictionaries representing metrics and artifacts.
```
dbczumar, line 940: The first sentence uses a plural (`custom metric functions`) and the second sentence uses a singular pronoun (`it`). Can we make these agree?

docs/source/models.rst
```diff
@@ -922,7 +922,89 @@ and behavior:
...
+            data=eval_data,
+            targets="target",
+            model_type="regressor",
+            dataset_name="cali_housing",
+            evaluators=["default"],
+            custom_metrics=[example_custom_metric_fn],
+        )
+
+
+For a more comprehensive usage of custom metrics, refer to `this example from the MLflow GitHub Repository
```
dbczumar, line 1004:
```suggestion
For a more comprehensive custom metrics usage example, refer to `this example from the MLflow GitHub Repository
```

docs/source/models.rst
```diff
@@ -922,7 +922,89 @@ and behavior:
...
+<https://github.com/mlflow/mlflow/blob/master/examples/evaluation/evaluate_with_custom_metrics_comprehensive.py>`_.
+
+Any additional information about model evaluation behaviors and outputs is available in the
```
dbczumar, line 1007:
```suggestion
Additional information about model evaluation behaviors and outputs is available in the
```

examples/evaluation/README.md
```diff
@@ -9,6 +9,10 @@ specified dataset using builtin default evaluator, and log resulting metrics & a
   generated by `sklearn.datasets.make_classification`.
 - Example `evaluate_on_regressor.py` evaluate as scikit-learn `LinearRegression` model on dataset loaded by
   `sklearn.datasets.fetch_california_housing`
+- Example `evaluate_with_custom_metrics.py` evaluates a custom metric function on a scikit-learn `LinearRegression`
```
dbczumar, line 12: "Evaluate a custom metric function" isn't quite correct terminology. Models are evaluated on datasets, producing a collection of metrics, which may be computed using custom routines. Can we rephrase?

docs/source/models.rst
```diff
@@ -922,7 +922,89 @@ and behavior:
 .. |eval_importance_img| image:: _static/images/model_evaluation_feature_importance.png
    :width: 69%
 
-More information about model evaluation behaviors and outputs is available in the
+
+Evaluating on Custom Metrics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the default set of metrics is insufficient, you can specify a list of functions to the ``custom_metrics``
+argument of :py:func:`mlflow.evaluate()` to evaluate your models in ways that you see fit. Custom metric
```
dbczumar, line 930:
```suggestion
If the default set of metrics is insufficient, you can specify a list of ``custom_metrics`` functions to :py:func:`mlflow.evaluate()` to produce custom performance metrics for the models that you're evaluating. Custom metric
```

docs/source/models.rst
```diff
@@ -922,7 +922,89 @@ and behavior:
...
+    eval_data = X_test.copy()
+    eval_data["target"] = y_test
+
+
+    def example_custom_metric_fn(eval_df, builtin_metrics):
+        """
+        This example custom metric function creates a metric based on the ``prediction`` and
+        ``target`` columns in ``eval_df`` and a metric derived from existing metrics in
+        ``builtin_metrics``. In addition, it also generates and saves a scatter plot to a file
+        as an image artifact.
```
dbczumar, line 974:
```suggestion
        ``builtin_metrics``. It also generates and saves a scatter plot to a file
        as an image artifact.
```

dbczumar: Can we also explain what this scatter plot represents?

examples/evaluation/README.md
```diff
@@ -24,4 +28,6 @@ Run in this directory with Python.
 python evaluate_on_binary_classifier.py
 python evaluate_on_multiclass_classifier.py
 python evaluate_on_regressor.py
+python evaluate_with_custom_metrics.py
+python evaluate_with_custom_metrics_comprehensive.py
```
dbczumar, line 32: While we're at it, can we add these to our CI? https://github.com/mlflow/mlflow/blob/5784e7e833385e59cd194fd63e4ae5e456abd779/tests/examples/test_examples.py#L138-L191

mlflow/models/evaluation/artifacts.py
```diff
@@ -21,3 +30,154 @@ def _save(self, output_artifact_path):
...
+        return self._content
+
+
+class PickleEvaluationArtifact(EvaluationArtifact):
+    def _save(self, output_artifact_path):
+        with open(output_artifact_path, "wb") as f:
+            # Using protocol 4 because it is the highest supported protocol as of the
+            # introduction of PickleEvaluationArtifact. For more details regarding pickle
+            # protocols, refer to: https://docs.python.org/3/library/pickle.html#data-stream-format
+            pickle.dump(self._content, f, protocol=4)
```
dbczumar, line 81: Instead of hardcoding the protocol, can we use the default protocol via `pickle.DEFAULT_PROTOCOL`? The default protocol is `4` for recent Python versions.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -614,21 +625,100 @@ def plot_pr_curve():
...
+        elif inferred_type is ImageEvaluationArtifact:
+            raw_artifact.savefig(artifact_file_local_path)
+        else:
+            # storing as pickle
+            try:
+                # Using protocol 4 because it is the highest supported protocol as of the
+                # introduction of PickleEvaluationArtifact. For more regarding pickle protocols,
+                # refer to: https://docs.python.org/3/library/pickle.html#data-stream-format
+                with open(artifact_file_local_path, "wb") as f:
+                    pickle.dump(raw_artifact, f, protocol=4)
```
dbczumar, line 667: Can we reuse the `_save` routines that we already define for each of these artifact classes, rather than re-implementing them here?

examples/evaluation/evaluate_with_custom_metrics_comprehensive.py
```diff
@@ -0,0 +1,166 @@
...
+    example_dict.update(builtin_metrics)
+    example_dict_2 = '{"a": 3, "b": [1, 2, 3]}'
+    example_image = plt.figure()
+    plt.scatter(eval_df["prediction"], eval_df["target"])
+    plt.xlabel("Targets")
+    plt.ylabel("Predictions")
+    plt.title("Targets vs. Predictions")
+    example_custom_class = ExampleClass(10)
+
+    # In rare cases, you might already have constructed a EvaluationArtifact-like object
```
dbczumar, line 105: Why is this rare?

dbczumar: The common use case I can see for this is letting the user return a custom evaluation artifact type that isn't already supported. Do we have a recommended workflow for that?

mlflow/models/evaluation/base.py
```diff
@@ -793,52 +793,90 @@ def evaluate(
...
+                           - ``EvaluationArtifact`` type objects. i.e. ``CsvEvaluationArtifact``,
+                             ``ImageEvaluationArtifact``.
+                           - Pandas DataFrame. This will be resolved as a CSV artifact.
+                           - Numpy array. This will be saved as a .npy artifact.
+                           - Matplotlib Figure. This will be saved as an image artifact. Note that
+                             ``matplotlib.pyplot.savefig`` is called behind the scene with default
+                             configurations. To customize, either save the figure with the desired
+                             configurations and return its file path or define customizations
+                             through environment variables in ``matplotlib.rcParams``.
+                           - Other objects will be attempted to be pickled with protocol 4.
```
dbczumar, line 832: Can we not hardcode the protocol version? Rather, we should refer to the "default protocol"

mlflow/models/evaluation/base.py
```diff
@@ -793,52 +793,90 @@ def evaluate(
...
+                             values are objects representing the artifacts.
+
+                           Object types that artifacts can be represented as:
+
+                           - A string uri representing the file path to the artifact. MLflow will
+                             infer the type of the artifact based on the file extension.
+                           - A string representation of a JSON object. This will be saved as a
+                             .json artifact.
+                           - ``EvaluationArtifact`` type objects. i.e. ``CsvEvaluationArtifact``,
+                             ``ImageEvaluationArtifact``.
```
dbczumar, line 824: Can we explain that, in these cases, the user is expected to log the associated artifact content to MLflow manually from within the `custom_metric` function?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -614,21 +625,100 @@ def plot_pr_curve():
...
-            # TODO: artifact detection and logging.
+            if artifact_results is not None:
+                for artifact_name, raw_artifact in artifact_results.items():
+                    self.artifacts[artifact_name] = self._log_custom_metric_artifact(
+                        artifact_name,
+                        raw_artifact,
+                        custom_metric_tuple,
+                    )
+
+    def _log_and_return_evaluation_result(self):
```
dbczumar, line 717: Can we add a docstring for this?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -614,21 +625,100 @@ def plot_pr_curve():
 
         self._log_pandas_df_artifact(per_class_metrics_collection_df, "per_class_metrics")
 
-    def _evaluate_custom_metrics(self):
+    def _log_custom_metric_artifact(self, artifact_name, raw_artifact, custom_metric_tuple):
```
dbczumar, line 628: Can we add a docstring for this?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -261,28 +271,29 @@ def _do_plot(**kwargs):
 }
 
 
-def _evaluate_custom_metric(index, custom_metric, eval_df, builtin_metrics):
+_CustomMetricFnTuple = namedtuple("_CustomMetricFnTuple", ["fn", "name", "index"])
```
dbczumar, line 274:
```suggestion
_CustomMetric = namedtuple("_CustomMetric", ["function", "name", "index"])
```

dbczumar: Can you also add a docstring for this class?

docs/source/models.rst
MarkYHZhang: Will change to plural. Thanks!

mlflow/models/evaluation/artifacts.py
MarkYHZhang: I didn't do that because I thought it might be a good idea to maintain a fixed version to ensure future compatibility. i.e. in the event that DEFAULT_PROTOCOL changes to 5 in a future update, and that would cause MLflow unable to read pickle artifacts generated from prior versions.  Should I be concerned about this?

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: It would be tricky to do that. Because from my understanding, in order to construct an `EvaluationArtifact`-like object, you first need to save it to a local file, then call `mlflow.log_artifact`. In this case, we're saving user provided objects to local files. Is there ways around this?

mlflow/models/evaluation/base.py
MarkYHZhang: ditto: https://github.com/mlflow/mlflow/pull/5405#discussion_r821244734

mlflow/models/evaluation/artifacts.py
BenWilson2: It was 11 years of releases between minor version default assignment between Protocol 3 and 4. I think it's safe to use the default (also considering that a Py minor version is typically maintained for ~ 4 years, any model that's that old and hasn't retrained is likely a) not running in production any more or b) not providing any value to a company).

mlflow/models/evaluation/default_evaluator.py
dbczumar: We should introduce a way for developers to construct before saving.

mlflow/models/evaluation/base.py
dbczumar: We should use the default protocol, as @BenWilson2 has discussed. Most tools use the default protocol.

mlflow/models/evaluation/artifacts.py
MarkYHZhang: Sounds good!

mlflow/models/evaluation/base.py
MarkYHZhang: Got it.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: I think this is a good idea since that's a better user experience when creating and logging custom `EvaluationArtifact` classes. Although I think this is pretty significant and could be out of scope of this PR. What do you all think about removing the ability for users to log ``EvaluationArtifact`` objects in this PR, and once I added the ability to construct `EvaluationArtifact` before saving I'll add back this functionality? @dbczumar @jinzhang21 @BenWilson2 @apurva-koti

jinzhang21: Agreed. I like the idea to provide general support on `_save`.

MarkYHZhang:
Removed support for custom metric functions to log `EvaluationArtifact` objects for now. This support will be added once we have completed revisioning of `EvaluationArtifacts` such that they can be constructed [without the need to save](https://github.com/mlflow/mlflow/pull/5405#discussion_r821265963).

examples/evaluation/evaluate_with_custom_metrics_comprehensive.py
MarkYHZhang: Will include an example for this in the follow up PR discussed [here](https://github.com/mlflow/mlflow/pull/5405#issuecomment-1062114812)

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Will remove the ability to log `EvaluationArtifact` for now. Once revisions to have general support on `_save` is done this ability will be added back.

docs/source/models.rst
```diff
@@ -922,7 +922,90 @@ and behavior:
...
+        """
+        metrics = {
+            "squared_diff_plus_one": np.sum(np.abs(eval_df["prediction"] - eval_df["target"] + 1) ** 2),
+            "sum_on_label_divided_by_two": builtin_metrics["sum_on_label"] / 2,
+        }
+        plt.scatter(eval_df["prediction"], eval_df["target"])
+        plt.xlabel("Targets")
+        plt.ylabel("Predictions")
+        plt.title("Targets vs. Predictions")
+        plot_path = os.path.join(tmp_dir, "example_scatter_plot.png")
```
jinzhang21, line 985: Can we create tmp_dir inside the custom metric function in this example? In general the function should not access anything other than eval_df and builtin_metrics, to make pickling the function much easier. This would be useful for distributed processing in the future.

examples/evaluation/evaluate_with_custom_metrics.py
```diff
@@ -3,28 +3,49 @@
...
+    plt.ylabel("Predictions")
+    plt.title("Targets vs. Predictions")
+    plot_path = os.path.join(tmp_dir, "example_scatter_plot.png")
+    plt.savefig(plot_path)
+    artifacts = {"example_scatter_plot_artifact": plot_path}
+    return metrics, artifacts
 
 
-with mlflow.start_run() as run:
+with mlflow.start_run() as run, tempfile.TemporaryDirectory() as tmp_dir:
```
jinzhang21, line 48: ditto for moving tmp_dir inside the function.

examples/evaluation/evaluate_with_custom_metrics_comprehensive.py
```diff
@@ -0,0 +1,155 @@
...
+
+    plt.savefig(os.path.join(tmp_dir, "example2.png"))
+    artifacts = {
+        "example_dict_2_from_obj_saved_as_csv": example_dict,
+        "example_image_2_from_png_file": os.path.join(tmp_dir, "example2.png"),
+    }
+    return metrics, artifacts
+
+
+with mlflow.start_run() as run, tempfile.TemporaryDirectory() as tmp_dir:
```
jinzhang21, line 136: ditto

mlflow/models/evaluation/artifacts.py
```diff
@@ -21,3 +30,151 @@ def _save(self, output_artifact_path):
...
+    ".jpg": ImageEvaluationArtifact,
+    ".jpeg": ImageEvaluationArtifact,
+    ".json": JsonEvaluationArtifact,
+    ".npy": NumpyEvaluationArtifact,
+    ".csv": CsvEvaluationArtifact,
+    ".parquet": ParquetEvaluationArtifact,
+    ".txt": TextEvaluationArtifact,
+}
+
+TYPE_TO_EXT_MAP = {
```
jinzhang21, line 97: Add _ to make it private. I don't think need to expose it. -> `_TYPE_TO_EXT_MAP`

mlflow/models/evaluation/artifacts.py
```diff
@@ -21,3 +30,151 @@ def _save(self, output_artifact_path):
...
+    pd.DataFrame: ".csv",
+    np.ndarray: ".npy",
+    plt.Figure: ".png",
+}
+
+TYPE_TO_ARTIFACT_MAP = {
```
jinzhang21, line 103: ditto

mlflow/models/evaluation/artifacts.py
```diff
@@ -21,3 +30,151 @@ def _save(self, output_artifact_path):
...
+    np.ndarray: NumpyEvaluationArtifact,
+    plt.Figure: ImageEvaluationArtifact,
+}
+
+_InferredArtifactProperties = namedtuple(
+    "_InferredArtifactProperties", ["from_path", "type", "ext"]
+)
+
+
+def _infer_artifact_type_and_ext(artifact_name, raw_artifact, custom_metric_tuple):
```
jinzhang21, line 114: Nice. It's very clean now.

mlflow/models/evaluation/base.py
```diff
@@ -793,52 +793,89 @@ def evaluate(
...
+                                       )
+                                   }
+
+                               def scatter_plot(eval_df, builtin_metrics):
+                                   import tempfile
+                                   plt.scatter(eval_df['prediction'], eval_df['target'])
+                                   plt.xlabel('Targets')
+                                   plt.ylabel('Predictions')
+                                   plt.title("Targets vs. Predictions")
+                                   plt.savefig(os.path.join(tmp_dir, "example.png"))
```
jinzhang21, line 865: ditto for tmp_dir

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -261,28 +270,42 @@ def _do_plot(**kwargs):
 }
 
 
-def _evaluate_custom_metric(index, custom_metric, eval_df, builtin_metrics):
+_CustomMetric = namedtuple("_CustomMetric", ["function", "name", "index"])
```
jinzhang21, line 273: Have you considered making a callable class with name and index attributes? It may look cleaner

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -614,21 +637,116 @@ def plot_pr_curve():
...
+        exception_and_warning_header = (
+            f"Custom metric function '{custom_metric_tuple.name}' at index "
+            f"{custom_metric_tuple.index} in the `custom_metrics` parameter"
+        )
+
+        inferred_from_path, inferred_type, inferred_ext = _infer_artifact_type_and_ext(
+            artifact_name, raw_artifact, custom_metric_tuple
+        )
+        artifact_file_name = _gen_log_key(artifact_name, self.dataset_name) + inferred_ext
+        artifact_file_local_path = self.temp_dir.path(artifact_file_name)
```
jinzhang21, line 663: What happens if there's a collision on `artifact_file_local_path`? i.e. 2 or more custom metric functions are creating the same artifact?

jinzhang21 (commented):
Mostly LGTM! I'll do a final pass after comments are addressed.

docs/source/models.rst
MarkYHZhang: Uhmm, yeah, I did that originally. However, the problem with having the tmp_dir context inside the function is that once the function finishes execution, it deletes the directory. This is problematic because later on in our code, we will need to reference to those produced artifacts. Perhaps we could introduce another argument to custom metric functions called `tmp_dir`.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Oh that's true. I went with `namedtuple` because Corey suggested it to me, and also I saw that there's another similar usage of `namedtuple` in this file for [`_Curve`](https://github.com/mlflow/mlflow/blob/561eeffc8dd9a62585821065d5ee8c0fbc186e24/mlflow/models/evaluation/default_evaluator.py#L166).

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: Right, I think it's a good idea to throw an exception to let the user know if there's a collision.

docs/source/models.rst
jinzhang21: I see. In this case, I think we need to change the API to provide this directory as part of the custom metric function argument, since we do expect users to write data to disk, in addition to the tuple of dicts returned.

mlflow/models/evaluation/default_evaluator.py
MarkYHZhang: I opted instead to use `typing.NamedTuple`. It is a much cleaner approach imo.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -261,28 +274,48 @@ def _do_plot(**kwargs):
...
-        f"'{getattr(custom_metric, '__name__', repr(custom_metric))}' at index {index} in the "
-        f"`custom_metrics` parameter"
+        f"Custom metric function '{custom_metric_tuple.name}' at index {custom_metric_tuple.index}"
+        " in the `custom_metrics` parameter"
     )
 
-    result = custom_metric(eval_df, builtin_metrics)
+    if len(inspect.signature(custom_metric_tuple.function).parameters) == 3:
+        result = custom_metric_tuple.function(
+            eval_df, builtin_metrics, custom_metric_tuple.artifacts_dir
```
jinzhang21, line 314: I thought we agreed to not add artifacts_dir to the user-supplied function args. User can treat it as a global variable that always exists during the function runtime. Just note it in the comments sections and examples should be sufficient.

jinzhang21 (approved):
LGTM! Let's wait for some feedback from ml-api@ before merging this change.

PR #5428, opened 2022-02-28
Switched rest_utils to use cached Requests.Session objects

jinzhang21:
Signed-off-by: jinzhang21 <jin.zhang@databricks.com>

Currently for every single HTTP request, we create a new Session object, which has to re-establish a new TCP connection to remote server before sending and receiving data. This is inefficient and sometimes can cause client be rate-limited by the server.

## How is this patch tested?

Existing unit tests.
Manual test on Databricks simulated environment:
- Log 1000 metrics
- Log 1000 params
- Log 100 artifacts
- Log 100 models 

Some anecdotal performance improvements are observed:
- From 1.23.1, total run time about **410 seconds**
- From current PR, total run time about **340 seconds**.
Above measurements are repeated twice.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/rest_utils.py
+++ b/mlflow/utils/rest_utils.py
@@ -3,6 +3,7 @@ import json
 import requests
 import urllib3
 from contextlib import contextmanager
+from functools import lru_cache
 from packaging.version import Version
 from requests.adapters import HTTPAdapter
 from urllib3.util import Retry
@@ -31,30 +32,42 @@ _TRANSIENT_FAILURE_RESPONSE_CODES = frozenset(
         504,  # Gateway Timeout
     ]
 )
-# A dictionary caching Requests.Sessions during runtime. This is useful for TCP connection re-use.
-_REQUEST_SESSIONS = {}
 
 
-def _get_request_session(retry_kwargs):
+@lru_cache(maxsize=64)
+def _get_request_session(max_retries, backoff_factor, retry_codes):
     """
     Returns a cached Requests.Session object for making HTTP request.
 
-    :param retry_kwargs: a dict containing urllib3.Retry instantiation arguments. A change will
-      result in creating and caching a new Requests.Session object.
+    :param max_retries: Maximum total number of retries.
+    :param backoff_factor: a time factor for exponential backoff. e.g. value 5 means the HTTP
+      request will be retried with interval 5, 10, 20... seconds. A value of 0 turns off the
+      exponential backoff.
+    :param retry_codes: a list of HTTP response error codes that qualifies for retry.
     :return: requests.Session object.
     """
-    # If a session already exists for the retry configuration, simply returns it.
-    retry_key = hash(frozenset(retry_kwargs.items()))
-    if retry_key in _REQUEST_SESSIONS:
-        return _REQUEST_SESSIONS.get(retry_key)
+    assert 0 <= max_retries < 10
+    assert 0 <= backoff_factor < 120
 
-    # Otherwise create a new session with current retry configuration.
+    retry_kwargs = {
+        "total": max_retries,
+        "connect": max_retries,
+        "read": max_retries,
+        "redirect": max_retries,
+        "status": max_retries,
+        "status_forcelist": retry_codes,
+        "backoff_factor": backoff_factor,
+    }
+    if Version(urllib3.__version__) >= Version("1.26.0"):
+        retry_kwargs["allowed_methods"] = None
+    else:
+        retry_kwargs["method_whitelist"] = None
+
     retry = Retry(**retry_kwargs)
     adapter = HTTPAdapter(max_retries=retry)
     session = requests.Session()
     session.mount("https://", adapter)
     session.mount("http://", adapter)
-    _REQUEST_SESSIONS[retry_key] = session
     return session
 
 
@@ -75,24 +88,7 @@ def _get_http_response_with_retries(
 
     :return: requests.Response object.
     """
-    assert 0 <= max_retries < 10
-    assert 0 <= backoff_factor < 120
-
-    retry_kwargs = {
-        "total": max_retries,
-        "connect": max_retries,
-        "read": max_retries,
-        "redirect": max_retries,
-        "status": max_retries,
-        "status_forcelist": retry_codes,
-        "backoff_factor": backoff_factor,
-    }
-    if Version(urllib3.__version__) >= Version("1.26.0"):
-        retry_kwargs["allowed_methods"] = None
-    else:
-        retry_kwargs["method_whitelist"] = None
-
-    session = _get_request_session(retry_kwargs)
+    session = _get_request_session(max_retries, backoff_factor, retry_codes)
     return session.request(method, url, **kwargs)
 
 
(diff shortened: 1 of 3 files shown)


mlflow/utils/rest_utils.py
```diff
@@ -31,6 +31,31 @@
         504,  # Gateway Timeout
     ]
 )
+# A dictionary caching Requests.Sessions during runtime. This is useful for TCP connection re-use.
+_REQUEST_SESSIONS = {}
+
+
+def _get_request_session(retry_kwargs):
```
harupy, line 38: Can we use [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache)?

mlflow/utils/rest_utils.py
```diff
@@ -67,13 +92,8 @@ def _get_http_response_with_retries(
...
 
-    retry = Retry(**retry_kwargs)
-    adapter = HTTPAdapter(max_retries=retry)
-    with requests.Session() as http:
-        http.mount("https://", adapter)
-        http.mount("http://", adapter)
-        response = http.request(method, url, **kwargs)
-        return response
+    session = _get_request_session(retry_kwargs)
+    return session.request(method, url, **kwargs)
```
harupy, line 96:
```suggestion
    with _get_request_session(retry_kwargs) as session:
        return session.request(method, url, **kwargs)
```

Can we use `with`?

harupy: The reason we don't use `with` is to avoid closing the session?

tests/utils/test_rest_utils.py
```diff
@@ -4,6 +4,7 @@
 import numpy
 import pytest
 
+import mlflow.utils.rest_utils as rest_utils
```
harupy, line 7:
```suggestion
from mlflow.utils import rest_utils
```

nit: Can we use from-import?

mlflow/utils/rest_utils.py
harupy:
```python
>>> hash({"a": 0})
TypeError: unhashable type: 'dict'
```

`dict` is

harupy: So we need a trick like this if we use `lru_cache`:

```python
from functools import lru_cache


class HashableDict(dict):
    def __hash__(self):
        return hash(frozenset(self.items()))


@lru_cache(maxsize=100)
def func(kwargs):
    print(kwargs)


func(HashableDict({"a": 0}))
func(HashableDict({"a": 0}))
func(HashableDict({"a": 0, "b": 1}))
```

BenWilson2: Won't we have to set a cache expiry time in order to fetch updated page results server-side and prevent stale cached data from being displayed if a previous request's data updates? (I imagine this might be applicable to certain pages and not others, though)

`@lru_cache(maxsize=<x>, seconds=30)` (or some other reasonable time period based on updating periodicities of run data while navigating). 

Would we want to cache Put requests or just Get requests?

Potential future food for thought for this: [requests-cache](https://requests-cache.readthedocs.io/en/latest/user_guide.html) which can give endpoint isolation for caching certain endpoints with different settings and a bunch of other potentially useful features to lessen the tracking server burden.

jinzhang21: Not sure if I fully understand your concern, but the added function LRU only caches the `requests.Session` object. All outgoing requests (initiated by `session.request`) are not cached and should never be.

The objective here is to persist the TCP channel across HTTP requests to the same destination IP/Port combo, nothing else.

mlflow/utils/rest_utils.py
```diff
@@ -33,22 +34,17 @@
 )
 
 
-def _get_http_response_with_retries(
-    method, url, max_retries, backoff_factor, retry_codes, **kwargs
-):
+@lru_cache(maxsize=64)
+def _get_request_session(max_retries, backoff_factor, retry_codes):
```
dbczumar, line 38: We use the same max retries, backoff factor, and retry codes for nearly all requests, implying that only a single session will be used. Shall we scope this based on URL as well?

mlflow/utils/rest_utils.py
BenWilson2: I was just curious if a NAT router / Firewall change between the server and the client introducing a network change would lead a long-running session into defaulting to maximum backoff retry event waiting before invalidating the cache value and establishing a new TCP connection. Maybe I'm thinking about this wrong though.

mlflow/utils/rest_utils.py
jinzhang21: It's very much true. And I think it is working as intended -- most of HTTP requests are sent from a single client (current runtime) to a single destination server, which should and could be considered as in the same session.

Session uses urlib3 to manage the connection pool (on TCP level) under the hood. If url differs, a new connection will be created there. I don't see a specific reason to create new sessions for that purpose.

mlflow/utils/rest_utils.py
jinzhang21: See the top section of this page: https://docs.python-requests.org/en/latest/user/advanced/#session-objects
I think we can safely leave the network management and TCP connection re-use to urllib3, which is managed by the session object.

mlflow/utils/rest_utils.py
dbczumar: We make cloud storage requests to S3, Azure, GCP, etc. Is there anything problematic with reusing the same session object across all of these platforms? I'm just wondering why this wouldn't be the default behavior for requests if single session use across an entire application was recommended it most cases.

jinzhang21: We can certainly cache sessions based on URLs. My point is that it's less efficient, in the sense that Session is just a thin wrapper used to persist a few settings across requests. More specifically, it "Provides cookie persistence, connection-pooling, and configuration.", according to the class documentation: https://github.com/psf/requests/blob/main/requests/sessions.py#L327

From https://github.com/psf/requests/blob/main/requests/sessions.py#L343, among all stuff a Session persists currently
```
        'headers', 'cookies', 'auth', 'proxies', 'hooks', 'params', 'verify',
        'cert', 'adapters', 'stream', 'trust_env', 'max_redirects',
```
the only setting that's being persisted in our use case is the `adapters` (which could change depending on the retry logic), which does the TCP connection pooling and re-use.

In the future, if we decide to persist auth and cookies, it probably makes sense to change the cache logic to take these into account, which should happen automatically since auth and cookie settings have to be passed into the session generator function `_get_request_session` as part of the argument, which will change the session cache key signature.

mlflow/utils/rest_utils.py
BenWilson2: 100%. I forgot about how requests will drop the connection and re-validate with a new connection from the pool if a connection cannot be established.

mlflow/utils/rest_utils.py
dbczumar: Understood! Thanks for the clarification. Can we manually confirm that logging both models and metadata (metrics, params, etc) works properly on Databricks before merging? I see we've already tested metrics.

dbczumar (approved):
LGTM pending results of manual testing with model logging + metadata logging. Thanks @jinzhang21 !

mlflow/utils/rest_utils.py
jinzhang21: Tested on Databricks with logging metrics, params, artifacts and models repeatedly. No problem was observed. We see some performance improvements, especially on repeated small requests such as metric logging.

harupy (approved):
LGTM!

PR #5434, opened 2022-02-28
Prototype restoring model dependencies for `pyfunc.load_model` and `pyfunc.spark_udf`

WeichenXu123:

Prototype restoring model dependencies for `pyfunc.load_model` and `pyfunc.spark_udf`

## How is this patch tested?

Manually.

(1) Log a model first:
```
import mlflow
from sklearn.datasets import make_classification
import sklearn.metrics

X, y = make_classification(n_samples=100, n_classes=3, n_informative=3, random_state=1)

"""
mlflow.sklearn.autolog()

    model = LogisticRegression(
        solver='liblinear',
        max_iter=1,
        C=0.5,
    ).fit(X, y)

print(run.info.run_id)
```

(2): load model with `restore_python_env` set to be `True`

```
import mlflow
from sklearn.datasets import make_classification
import sklearn.metrics

X, y = make_classification(n_samples=100, n_classes=3, n_informative=3, random_state=1)

m = mlflow.pyfunc.load_model(
    'runs:/{run_id}/model',  # sklearn model, sklearn version 1.0.2
    restore_python_env=True,
)
print(type(m))

y_pred = m.predict(X)

print(f"precision={sklearn.metrics.precision_score(y, y_pred, average='micro')}")
```

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -274,6 +274,7 @@ from mlflow.utils.requirements_utils import (
     _check_requirement_satisfied,
     _parse_requirements,
 )
+from mlflow.utils._spark_utils import _SparkBroadcastFileCache
 
 FLAVOR_NAME = "python_function"
 MAIN = "loader_module"
@@ -695,10 +696,15 @@ _PICKLE_PROTOCOL_FOR_RESTORE_PY_ENV = 3
 
 
 class _RestoredPythonEnvSubprocessMgr:
-
-    def __init__(self, py_version, requirements_file_path):
+    """
+    We need create _RestoredPythonEnvSubprocessMgr instance
+    from spark application driver side.
+    """
+    def __init__(self, py_version, requirements_file_path, loader_module, model_path):
         self._py_version = py_version
         self._requirements_file_path = requirements_file_path
+        self._loader_module = loader_module
+        self._model_path = model_path
 
         self._subproc = None
         self._python_installation_dir = None
@@ -749,10 +755,10 @@ class _RestoredPythonEnvSubprocessMgr:
         pyenv_home_path = os.path.join(
             env_cache_dir, 'pythons'
         )
-        python_installation_path = os.path.join(
+        self._python_installation_dir = os.path.join(
             pyenv_home_path, '.pyenv', 'versions', self._py_version
         )
-        if not os.path.exists(python_installation_path):
+        if not os.path.exists(self._python_installation_dir):
             # install specified python version
             sub_env = os.environ.copy()
             sub_env['HOME'] = pyenv_home_path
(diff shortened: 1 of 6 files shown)


mlflow/pyfunc/__init__.py
```diff
@@ -656,15 +672,235 @@ def _warn_dependency_requirement_mismatches(model_path):
...
+                 os.path.dirname(os.path.dirname(mlflow.__file__))]
+            )
+        else:
+            _logger.info(f'Found python virtual env cache at {virtual_env_path}')
+
+        self._virtual_env_python_bin = os.path.join(
+            virtual_env_path, 'bin/python'
+        )
+        # setup tcp connection with subproc
+        listen_sock = socket.socket()
```
WeichenXu123, line 794: The reason I use socket to transferring data between processes is:
* using stdin/stdout is risky, logger or some other routine might also write data to stdout.
* using pipe seems not possible because subprocess is not forked from parent process.

then socket become best choice. Also note spark python UDF implementation it also using socket to transfer data for sub-process executing python UDF routine.

mlflow/pyfunc/__init__.py
```diff
@@ -656,15 +674,285 @@ def _warn_dependency_requirement_mismatches(model_path):
...
+        return e
+
+
+_PYENV_CMD = 'pyenv'
+_VIRTUALENV_CMD = 'virtualenv'
+
+_DATABRICKS_NFS_CACHE_ROOT_DIR = "/local_disk0/.ephemeral_nfs/mlflow/cache"
+
+
+_PICKLE_PROTOCOL_FOR_RESTORE_PY_ENV = 3
```
BenWilson2, line 695: Do we want to stick with Protocol 4 (supported since Py3.4) which is the default in Py3.8+ to support, as the docs (https://docs.python.org/3/library/pickle.html#data-stream-format) mention "adds support for very large objects, pickling more types of objects, and some data format optimizations"?

mlflow/pyfunc/__init__.py
```diff
@@ -656,15 +674,285 @@ def _warn_dependency_requirement_mismatches(model_path):
...
+        self._python_installation_dir = os.path.join(
+            pyenv_home_path, '.pyenv', 'versions', self._py_version
+        )
+        if not os.path.exists(self._python_installation_dir):
+            # install specified python version
+            sub_env = os.environ.copy()
+            sub_env['HOME'] = pyenv_home_path
+            subprocess.run(
+                [_PYENV_CMD, 'install', '-v', self._py_version],
+                check=True, stdout=sys.stdout, stderr=sys.stderr,
```
BenWilson2, line 767: will this subprocess sys.stdout and sys.stderr be visible on the driver for Spark? 
Should we have this install information piped to a log file so that users don't get confused if seeing conflicting versions of libraries that are not actually installed in their session repl?

mlflow/pyfunc/__init__.py
BenWilson2: It's just a one-time hit to a relatively slow process since it's only going to transfer this environment a single time per execution session though, right? It seems like a smart design to me! :)

mlflow/pyfunc/__init__.py
```diff
@@ -656,15 +674,285 @@ def _warn_dependency_requirement_mismatches(model_path):
...
+                 '--model_path', self._model_path,
+                 ],
+                stdout=sys.stdout, stderr=sys.stderr,
+            )
+
+            comm_sock, _ = listen_sock.accept()
+            self._comm_sock = comm_sock
+            self._comm_stream = comm_sock.makefile('rwb')
+        except socket.timeout:
+            raise RuntimeError('Launch sub process for virtual python environment failed.')
```
BenWilson2, line 829: Are we going to have an ability to capture this exception and run with current system environment if something goes wrong with the environment re-build (i.e. some OSS packages delisting a particular version of a package due to bugs)?

mlflow/pyfunc/__init__.py
WeichenXu123: I think we can automatically generate best Protocol number, according to the virtual env python version.

mlflow/pyfunc/__init__.py
WeichenXu123: > i.e. some OSS packages delisting

Is this possible ? I think version published to pypi cannot be withdrawn.

And if subprocess failed due to any reason, I would like to let the error raised and let user determine what to do next.

mlflow/pyfunc/__init__.py
WeichenXu123: Yes I will improve how to handle the subprocess output. 
redirect them to parent output is easier for user debugging, we can prepend string like "subprocess output: " to the output logs.

PR #5448, opened 2022-03-03
Add `code_paths` parameter to all flavors

stevenchen-db:
Signed-off-by: Steven Chen <s.chen@databricks.com>

This PR adds a `code_paths` parameters to all flavors. `code_paths` allow users to log custom modules for model classes that are defined in external modules, which currently is only possible with pytorch and pyfunc models. pyfunc's load_model already handles adding code_paths to the system conf when `code` is specified in the pyfunc conf.

## How is this patch tested?
* unit tests
* the existing pytorch and pyfunc tests should verify that the new helper methods work e2e in serving
* manually tested logging and loading models for different flavors

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mlflow/paddle/__init__.py
```diff
@@ -296,6 +303,7 @@ def load_model(model_uri, model=None, dst_path=None, **kwargs):
     import paddle
 
     local_model_path = _download_artifact_from_uri(artifact_uri=model_uri, output_path=dst_path)
+    pyfunc.utils._add_code_from_conf_to_system_path(local_model_path)
```
ankit-db, line 306: Weird...is the `load_model` for this flavor in a different spot in the file than other flavors?

tests/catboost/test_catboost_model_export.py
```diff
@@ -419,3 +420,11 @@ def test_pyfunc_serve_and_score_sklearn(reg_model):
     )
     scores = pd.read_json(resp.content.decode("utf-8"), orient="records").values.squeeze()
     np.testing.assert_array_almost_equal(scores, model.predict(inference_dataframe.head(3)))
+
+
+def test_save_model_with_code_paths(cb_model, model_path):
```
ankit-db, line 425: Love the unit tests! Can we add `log_model` to the call as well? I know that it always calls `save_model` under the hood, but that seems like something the unit tests shouldn't know

ankit-db (commented):
One small comment about unit tests, but otherwise, LGTM

mlflow/paddle/__init__.py
stevenchen-db: what do you mean by a different spot? Isn't this still in the `load_model` function?

ankit-db: I just meant literally in the order of the file. Only noticed cause you were modifying all the files

stevenchen-db: ohhh yeah that is weird how it's located above `log_model`

mlflow/spark.py
```diff
@@ -678,7 +688,10 @@ def load_model(model_uri, dfs_tmpdir=None):
         model_uri = ModelsArtifactRepository.get_underlying_uri(model_uri)
         _logger.info("'%s' resolved as '%s'", runs_uri, model_uri)
     flavor_conf = _get_flavor_configuration_from_uri(model_uri, FLAVOR_NAME)
+
     model_uri = append_to_uri_path(model_uri, flavor_conf["model_data"])
+    local_model_path = _download_artifact_from_uri(model_uri)
+    pyfunc.utils._add_code_from_conf_to_system_path(local_model_path)
```
stevenchen-db, line 694: not too sure how model_uris work with spark, but the unit seemed to pass

mlflow/catboost.py
```diff
@@ -86,6 +87,9 @@ def save_model(
                      or `CatBoostRegressor`_) to be saved.
     :param path: Local path where the model is to be saved.
     :param conda_env: {{ conda_env }}
+    :param code_paths: A list of local filesystem paths to Python file dependencies (or directories
+                       containing file dependencies). These files are *prepended* to the system
+                       path when the model is loaded.
```
dbczumar, line 92: @stevenchen-db Can we clarify that these files are only prepended to the system path when the model is loaded as pyfunc via `mlflow.pyfunc.load_model()`?

stevenchen-db: Are you referring to all flavors or catboost specifically? Is there any reason why the flavor specific load_model shouldn't prepend the code_paths to the system path?

dbczumar: Oh, I didn't realize we were doing this for the flavor-specific case. `code` is an attribute of the `pyfunc` flavor specification of a model, so it's kind of strange that we're using / loading it when a model is restored with other flavors, e.g. `sklearn`. However, this feature seems pretty useful for users. How hard would it be to add the `code` property to each individual flavor spec?

stevenchen-db: @dbczumar Ah I see, I was going off of the way `load_model` is handled for pytorch and noticed that the pyfunc conf gets read to set the code path. For serving, we will need `code` to be in pyfunc since we use `mlflow.pyfunc.load_model`, which already has the logic to add code paths to the system path.

ankit-db: @stevenchen-db - maybe it makes sense to add code to both the pyfunc and flavor for each model and when the flavor is using flavor-specific logging, it can refer to the one in the flavor portion of MLModel?

stevenchen-db: @ankit-db is there a scenario where the `code` attribute would be different between pyfunc and the flavor? I thought the workflow would be 1) user calls `mlflow.<flavor>.log_model` and then 2) serving will use `mlflow.pyfunc.load_model`, so whatever code was logged through the flavor should be loaded back from the pyfunc conf right?

stevenchen-db: Oh @dbczumar were you asking if we should log `code` in both pyfunc and the flavor? I guess it might be cleaner to read `code` from the flavor conf, but I think the flavor's `code` conf would also always be a duplicate of the pyfunc `code` conf

dbczumar: @stevenchen-db Yep! Ideally, it would be nice to log to both, even if the content is duplicated. This helps preserve the flavor spec abstraction.

mlflow/pyfunc/utils.py
```diff
@@ -1,24 +0,0 @@
-import os
```
stevenchen-db, line 1: still figuring out how to get `black` to not complain when deleting this file

hooks/pre-commit
```diff
@@ -25,7 +25,7 @@ validate_version() {
   fi
 }
 
-diff=$(git diff --cached --name-only | grep '\.py$' || true)
+diff=$(git diff --cached --diff-filter ACMRTUXB --name-only | grep '\.py$' || true)
```
stevenchen-db, line 28: This is necessary to get past the pre commit hook if a file gets deleted. `black` will try to run on a deleted file since it shows up in the `git diff`, which results in an error. This filter selects all changes except file deletion (thanks @harupy!)

mlflow/catboost.py
stevenchen-db: Makes sense, just logged both and modified `load_model`!

ankit-db (approved):
Thanks for the change @stevenchen-db !

mlflow/utils/file_utils.py
```diff
@@ -21,11 +21,13 @@
 except ImportError:
     from yaml import SafeLoader as YamlSafeLoader, SafeDumper as YamlSafeDumper
 
+
```
dbczumar, line 24: I think the changes in this file belong in `model_utils`. Most of these utils either don't operate on files (e.g. `validate_code_paths`) or rely on other `file_utils` methods to interact with files without adding any new file-specific logic (e.g. `_copy_code_paths`).

mlflow/utils/file_utils.py
```diff
@@ -21,11 +21,13 @@
...
 from mlflow.entities import FileInfo
 from mlflow.exceptions import MissingConfigException
 from mlflow.utils.rest_utils import cloud_storage_http_request, augmented_raise_for_status
 
 ENCODING = "utf-8"
+CODE = "code"
```
dbczumar, line 30: Following from https://github.com/mlflow/mlflow/pull/5448/files#r822044736, let's move this to `model_utils` and use a clearer variable name, e.g. `FLAVOR_CONFIG_CODE`

mlflow/utils/file_utils.py
dbczumar: Can we also add docstrings to the new utility methods?

mlflow/utils/file_utils.py
```diff
@@ -357,6 +359,22 @@ def _copy_file_or_tree(src, dst, dst_dir=None):
...
+            _copy_file_or_tree(src=code_path, dst=path, dst_dir=code_dir_subpath)
+    else:
+        code_dir_subpath = None
+    return code_dir_subpath
+
+
+def _validate_code_paths(code_paths):
+    if code_paths is not None:
+        if not isinstance(code_paths, list):
+            raise TypeError("Argument code_paths should be a list, not {}".format(type(code_paths)))
```
dbczumar, line 375: Can we combine this validation logic with `copy_code_paths` to create a `validate_and_copy_code_paths` method? This way, flavor implementations don't need to import & call 2 separate methods. This is more robust, since it's somewhat easy for a developer to miss the validation method and still achieve working tests.

tests/catboost/test_catboost_model_export.py
```diff
@@ -419,3 +420,15 @@ def test_pyfunc_serve_and_score_sklearn(reg_model):
     )
     scores = pd.read_json(resp.content.decode("utf-8"), orient="records").values.squeeze()
     np.testing.assert_array_almost_equal(scores, model.predict(inference_dataframe.head(3)))
+
+
+def test_log_model_with_code_paths(cb_model):
+    artifact_path = "model"
+    with mlflow.start_run(), mock.patch(
+        "mlflow.catboost._add_code_from_conf_to_system_path"
```
dbczumar, line 428: Is there a way to test that `_add_code_from_conf_to_system_path` works? Can we introduce this test coverage?

dbczumar (changes requested):
@stevenchen-db Generally looks great! Left a few small comments about code structure, docstrings, and test coverage.

tests/catboost/test_catboost_model_export.py
stevenchen-db: I believe the pytorch tests with the module_scoped_subclassed_model will test that adding code to system paths work since the logged code needs to be added to the system path to serve the model

dbczumar: Seems pretty indirect. Can we test this directly?

stevenchen-db: Yep, just added a test in test_model_utils.py!

mlflow/utils/file_utils.py
```diff
@@ -483,3 +484,21 @@ def _handle_readonly_on_windows(func, path, exc_info):
         raise exc_value
     os.chmod(path, stat.S_IWRITE)
     func(path)
+
+
+def _get_code_dirs(src_code_path, dst_code_path=None):
```
dbczumar, line 489: Looks like this was left over after the migration to `model_utils`. Can we revert changes to `file_utils`?

dbczumar (approved):
LGTM once changes to `file_utils` are reverted. Thanks @stevenchen-db !

harupy (approved):
LGTM for https://github.com/mlflow/mlflow/pull/5448/files#r821204199

PR #5458, opened 2022-03-05
Adds JSON Schema Validation

### LINKED ISSUE #5208: [FR] Validate request JSON with JSON schema or similar

mrkaye97:

Closes #5208

This PR is a first pass at adding JSON validation against a predetermined schema, in an effort to make error handling more friendly for users. The end goal is to return HTTP 400 for bad parameters instead of 500, which happens currently (at least for most endpoints).

I'm sure there are many error handling cases I've missed, but I think this should cover the most common issues (misspecified parameter types). And of course, we can always expand on this in the future.

## How is this patch tested?

Tested with `pytest` as per @dbczumar's [suggestion](https://github.com/mlflow/mlflow/pull/5350#issuecomment-1039594468). I added a few tests to check if errors were coming back as expected, but we can always add more.

I also spun up a local MLFlow instance and ran some test API calls from an R session on local.

**Before this change:**

Lots of `500`s with nothing helpful.
Screen Shot 2022-03-06 at 2 37 37 PM

Screen Shot 2022-03-06 at 2 35 02 PM

Screen Shot 2022-03-06 at 2 34 44 PM

In order to debug, you would need to dig into your instance's logs, where you'd see something like this (which is helpful, but is hard to parse and doesn't indicate exactly what went wrong. And also, logs aren't always easily available to MLFlow users):

```
2022-03-06T17:44:56.390215+00:00 app[web.1]: Traceback (most recent call last):
2022-03-06T17:44:56.390216+00:00 app[web.1]:   File "/opt/conda/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app

...

2022-03-06T17:44:56.390219+00:00 app[web.1]:   File "/opt/conda/lib/python3.9/site-packages/mlflow/server/handlers.py", line 284, in _create_experiment

...

2022-03-06T17:44:56.390221+00:00 app[web.1]:   File "/opt/conda/lib/python3.9/site-packages/google/protobuf/json_format.py", line 445, in ParseDict
2022-03-06T17:44:56.390221+00:00 app[web.1]:     parser.ConvertMessage(js_dict, message)
2022-03-06T17:44:56.390221+00:00 app[web.1]:   File "/opt/conda/lib/python3.9/site-packages/google/protobuf/json_format.py", line 476, in ConvertMessage
2022-03-06T17:44:56.390222+00:00 app[web.1]:     self._ConvertFieldValuePair(value, message)
2022-03-06T17:44:56.390222+00:00 app[web.1]:   File "/opt/conda/lib/python3.9/site-packages/google/protobuf/json_format.py", line 594, in _ConvertFieldValuePair
2022-03-06T17:44:56.390222+00:00 app[web.1]:     raise ParseError('Failed to parse {0} field: {1}.'.format(name, e))
2022-03-06T17:44:56.390223+00:00 app[web.1]: google.protobuf.json_format.ParseError: Failed to parse name field: expected string or bytes-like object.
```

**After this change:**

Helpful error messages!

Screen Shot 2022-03-13 at 5 43 30 PM
Screen Shot 2022-03-13 at 5 43 42 PM
Screen Shot 2022-03-13 at 5 43 51 PM

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

This change adds validation of API request bodies against predetermined JSON schemas, which will result in friendlier error messages for MLFlow users.

See #5208 

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/R/mlflow/R/tracking-rest.R
+++ b/mlflow/R/mlflow/R/tracking-rest.R
@@ -60,37 +60,57 @@ mlflow_rest <- function( ..., client, query = NULL, data = NULL, verb = "GET", v
   get_response <- switch(
     verb,
     GET = function() {
-      GET( api_url, query = query, mlflow_rest_timeout(), config = rest_config$config,
-           req_headers)
+     GET(
+       api_url,
+       query = query,
+       mlflow_rest_timeout(),
+       config = rest_config$config,
+       req_headers
+     )
     },
-    POST = function(){
-      POST( api_url,
-            body = if (is.null(data)) NULL else rapply(data, as.character, how = "replace"),
-            encode = "json",
-            mlflow_rest_timeout(),
-            config = rest_config$config,
-            req_headers
-      )
+    POST = function() {
+     POST(
+       api_url,
+       body = toJSON(
+         data,
+         pretty = TRUE,
+         auto_unbox = TRUE,
+         na = "null"
+       ),
+       mlflow_rest_timeout(),
+       config = rest_config$config,
+       req_headers
+     )
     },
-    PATCH = function(){
-      httr::PATCH( api_url,
-            body = if (is.null(data)) NULL else rapply(data, as.character, how = "replace"),
-            encode = "json",
-            mlflow_rest_timeout(),
-            config = rest_config$config,
-            req_headers
-      )
+    PATCH = function() {
+     PATCH(
+       api_url,
+       body = toJSON(
+         data,
+         pretty = TRUE,
+         auto_unbox = TRUE,
+         na = "null"
+       ),
+       mlflow_rest_timeout(),
+       config = rest_config$config,
+       req_headers
+     )
     },
     DELETE = function() {
-      httr::DELETE(api_url,
-              body = if (is.null(data)) NULL else rapply(data, as.character, how = "replace"),
-              encode = "json",
-              mlflow_rest_timeout(),
-              config = rest_config$config,
-              req_headers
-      )
+     DELETE(
+       api_url,
+       body = toJSON(
+         data,
+         pretty = TRUE,
+         auto_unbox = TRUE,
+         na = "null"
+       ),
+       mlflow_rest_timeout(),
+       config = rest_config$config,
+       req_headers
+     )
     },
-    stop("Verb '", verb, "' is unsupported.", call. = FALSE)
+    abort("Verb '", verb, "' is unsupported.")
   )
   sleep_for <- 1
   time_left <- max_rate_limit_interval
--- a/mlflow/R/mlflow/R/tracking-rest.R
+++ b/mlflow/R/mlflow/R/tracking-rest.R
@@ -60,57 +60,37 @@ mlflow_rest <- function( ..., client, query = NULL, data = NULL, verb = "GET", v
   get_response <- switch(
     verb,
     GET = function() {
-     GET(
-       api_url,
-       query = query,
-       mlflow_rest_timeout(),
-       config = rest_config$config,
-       req_headers
-     )
+      GET( api_url, query = query, mlflow_rest_timeout(), config = rest_config$config,
+           req_headers)
     },
-    POST = function() {
-     POST(
-       api_url,
-       body = toJSON(
-         data,
-         pretty = TRUE,
-         auto_unbox = TRUE,
-         na = "null"
-       ),
-       mlflow_rest_timeout(),
-       config = rest_config$config,
-       req_headers
-     )
+    POST = function(){
+      POST( api_url,
+            body = if (is.null(data)) NULL else rapply(data, as.character, how = "replace"),
+            encode = "json",
+            mlflow_rest_timeout(),
+            config = rest_config$config,
+            req_headers
+      )
     },
-    PATCH = function() {
-     PATCH(
-       api_url,
-       body = toJSON(
-         data,
-         pretty = TRUE,
-         auto_unbox = TRUE,
-         na = "null"
-       ),
-       mlflow_rest_timeout(),
-       config = rest_config$config,
-       req_headers
-     )
+    PATCH = function(){
+      httr::PATCH( api_url,
+            body = if (is.null(data)) NULL else rapply(data, as.character, how = "replace"),
+            encode = "json",
+            mlflow_rest_timeout(),
+            config = rest_config$config,
+            req_headers
+      )
     },
     DELETE = function() {
-     DELETE(
-       api_url,
-       body = toJSON(
-         data,
-         pretty = TRUE,
-         auto_unbox = TRUE,
-         na = "null"
-       ),
-       mlflow_rest_timeout(),
-       config = rest_config$config,
-       req_headers
-     )
+      httr::DELETE(api_url,
+              body = if (is.null(data)) NULL else rapply(data, as.character, how = "replace"),
+              encode = "json",
+              mlflow_rest_timeout(),
+              config = rest_config$config,
+              req_headers
+      )
     },
-    abort("Verb '", verb, "' is unsupported.")
+    stop("Verb '", verb, "' is unsupported.", call. = FALSE)
   )
   sleep_for <- 1
   time_left <- max_rate_limit_interval
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -1086,40 +1086,17 @@ def _get_artifact_repo(run):
 @_disable_if_artifacts_only
 def _log_batch():
 
-    def _assert_metrics_correctly_constructed(metrics):
+    def _assert_metrics_fields_present(metrics):
         for m in metrics:
-            print("This is a metric: ", m)
-            ## Key
-            k = m["key"]
-            _assert_required(k)
-            _assert_string(k)
+            _assert_required(m["key"])
+            _assert_required(m["value"])
+            _assert_required(m["timestamp"])
+            _assert_required(m["step"])
 
-            ## Value
-            v = m["value"]
-            _assert_required(v)
-            _assert_floatlike(v)
-
-            ## Timestamp
-            t = m["timestamp"]
-            _assert_required(t)
-            _assert_intlike(t)
-
-            ## Step
-            s = m["step"]
-            _assert_required(s)
-            _assert_intlike(s)
-
-    def _assert_params_tags_correctly_constructed(params):
+    def _assert_params_tags_fields_present(params):
         for p in params:
-            ## Key
-            k = p["key"]
-            _assert_required(k)
-            _assert_string(k)
-
-            ## Value
-            v = p["value"]
-            _assert_required(v)
-            _assert_string(v)
+            _assert_required(p["key"])
+            _assert_required(p["value"])
 
     _validate_batch_log_api_req(_get_request_json())
     request_message = _get_request_message(
@@ -1131,15 +1108,15 @@ def _log_batch():
             ],
             "metrics": [
                 _assert_array,
-                _assert_metrics_correctly_constructed
+                _assert_metrics_fields_present
             ],
             "params": [
                 _assert_array,
-                _assert_params_tags_correctly_constructed
+                _assert_params_tags_fields_present
             ],
             "tags": [
                 _assert_array,
-                _assert_params_tags_correctly_constructed
+                _assert_params_tags_fields_present
             ]
         }
     )
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -288,7 +288,7 @@ def _assert_item_type_string(x):
         _assert_string(item)
 
 
-def _validate_param_against_schema(schema, param, value, parsing_succeeded = False):
+def _validate_param_against_schema(schema, param, value, proto_parsing_succeeded = False):
     for f in schema:
         # If the validator is a type checker and dict parsing already
         #  succeeeded, we can safely skip re-checking types
@@ -300,7 +300,7 @@ def _validate_param_against_schema(schema, param, value, parsing_succeeded = Fal
             _assert_float,
             _assert_array,
             _assert_item_type_string
-        ] and parsing_succeeded:
+        ] and proto_parsing_succeeded:
             continue
 
         try:
@@ -357,15 +357,15 @@ def _get_request_message(request_message, flask_request=request, schema = {}):
     if request_json is None:
         request_json = {}
 
-    parsing_succeeded = True
+    proto_parsing_succeeded = True
     try:
         parse_dict(request_json, request_message)
     except ParseError as e:
-        parsing_succeeded = False
+        proto_parsing_succeeded = False
 
     for k, v in request_json.items():
         if (k in schema.keys()):
-            _validate_param_against_schema(schema[k], k, v, parsing_succeeded)
+            _validate_param_against_schema(schema[k], k, v, proto_parsing_succeeded)
 
     return request_message
 
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -308,16 +308,16 @@ def _build_schema(name, additional_schema = {}):
 
     return schema
 
-def _validate_param_against_schema(schema, param):
+def _validate_param_against_schema(schema, param, value):
     ## If the schema is empty, do nothing
     if not schema:
         return True
 
     try:
-        Draft202012Validator(schema).validate(param)
+        Draft202012Validator(schema).validate(value)
     except ValidationError as e:
         raise MlflowException(
-            message=e.message,
+            message="Invalid value supplied for '{}''".format(param) + e.message,
             error_code=INVALID_PARAMETER_VALUE
         )
 
@@ -369,7 +369,7 @@ def _get_request_message(request_message, flask_request=request, **additional_sc
 
     for k, v in request_json.items():
         schema = _build_schema(k, additional_schema.get(k, {}))
-        _validate_param_against_schema(schema, v)
+        _validate_param_against_schema(schema, k, v)
 
     parse_dict(request_json, request_message)
 
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -292,7 +292,7 @@ def _validate_param_against_schema(schema, param, value):
             f(value)
         except AssertionError as e:
             raise MlflowException(
-                message="Invalid value {} for '{}' supplied.".format(value, param) + "\nSee the API docs for more information on parameters.",
+                message="Invalid value {} for parameter '{}' supplied.".format(value, param) + "\nSee the API docs for more information on specifying parameters.",
                 error_code=INVALID_PARAMETER_VALUE
             )
 
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -629,6 +629,9 @@ def _update_run():
             ],
             "end_time": [
                 _assert_int
+            ],
+            "status": [
+                _assert_string
             ]
         }
     )
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -294,8 +294,7 @@ def _assert_less_than_or_equal(x, max):
     assert x <= max
 
 def _assert_item_type_string(x):
-    for item in x:
-        _assert_string(item)
+    assert all(map(lambda item: isinstance(item, str), x))
 
 _TYPE_VALIDATORS = {
     _assert_intlike,
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -317,7 +317,7 @@ def _validate_param_against_schema(schema, param, value):
         Draft202012Validator(schema).validate(value)
     except ValidationError as e:
         raise MlflowException(
-            message="Invalid value supplied for '{}''".format(param) + e.message,
+            message="Invalid value for '{}' supplied.".format(param) + "\n" + e.message,
             error_code=INVALID_PARAMETER_VALUE
         )
 
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -265,13 +265,23 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
 def _assert_string(x):
     assert isinstance(x, str)
 
-def _assert_int(x):
+def _assert_intlike(x):
+    try:
+        x = int(x)
+    except ValueError:
+        pass
+
     assert isinstance(x, int)
 
 def _assert_bool(x):
     assert isinstance(x, bool)
 
-def _assert_float(x):
+def _assert_floatlike(x):
+    try:
+        x = float(x)
+    except ValueError:
+        pass
+
     assert isinstance(x, float)
 
 def _assert_array(x):
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -1155,7 +1155,8 @@ def _log_model():
                 _assert_required
             ],
             "model_json": [
-                _assert_string
+                _assert_string,
+                _assert_required
             ]
         }
     )
--- a/mlflow/server/handlers.py
+++ b/mlflow/server/handlers.py
@@ -13,9 +13,6 @@ from functools import wraps
 from flask import Response, request, current_app, send_file
 from google.protobuf import descriptor
 
-from jsonschema import validate, Draft202012Validator
-from jsonschema.exceptions import ValidationError
-
 from mlflow.entities import Metric, Param, RunTag, ViewType, ExperimentTag, FileInfo
 from mlflow.entities.model_registry import RegisteredModelTag, ModelVersionTag
 from mlflow.exceptions import MlflowException
@@ -265,69 +262,41 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
     except UnsupportedModelRegistryStoreURIException:
         pass
 
-_STRING_TYPE_DEFAULT_SCHEMA = {"type": "string"}
-_INTEGER_TYPE_DEFAULT_SCHEMA = {"type": "integer"}
-_REQUIRED_PARAM_DEFAULT_SCHEMA = {"required": True}
+def _assert_string(x):
+    assert isinstance(x, str)
 
-_DEFAULT_SCHEMAS = {
-    "name": _STRING_TYPE_DEFAULT_SCHEMA,
-    "artifact_location": _STRING_TYPE_DEFAULT_SCHEMA,
-    "tags": {"type": "array", "minItems": 0},
-    "max_results": _INTEGER_TYPE_DEFAULT_SCHEMA,
-    "page_token": _STRING_TYPE_DEFAULT_SCHEMA,
-    "next_page_token": _STRING_TYPE_DEFAULT_SCHEMA,
-    "experiment_id": _STRING_TYPE_DEFAULT_SCHEMA,
-    "experiment_name": _STRING_TYPE_DEFAULT_SCHEMA,
-    "new_name": _STRING_TYPE_DEFAULT_SCHEMA,
-    "timestamp": _INTEGER_TYPE_DEFAULT_SCHEMA,
-    "start_time": _INTEGER_TYPE_DEFAULT_SCHEMA,
-    "end_time": _INTEGER_TYPE_DEFAULT_SCHEMA,
-    "run_id": _STRING_TYPE_DEFAULT_SCHEMA,
-    "key": _STRING_TYPE_DEFAULT_SCHEMA,
-    "step": _INTEGER_TYPE_DEFAULT_SCHEMA,
-    "model_json": _STRING_TYPE_DEFAULT_SCHEMA,
-    "metric_key": _STRING_TYPE_DEFAULT_SCHEMA,
-    "filter": _STRING_TYPE_DEFAULT_SCHEMA,
-    "path": _STRING_TYPE_DEFAULT_SCHEMA,
-    "description": _STRING_TYPE_DEFAULT_SCHEMA,
-    "stages": {"type": "array", "items": _STRING_TYPE_DEFAULT_SCHEMA},
-    "source": _STRING_TYPE_DEFAULT_SCHEMA,
-    "run_link": _STRING_TYPE_DEFAULT_SCHEMA,
-    "version": _STRING_TYPE_DEFAULT_SCHEMA,
-    "order_by": {"type": "array", "items": _STRING_TYPE_DEFAULT_SCHEMA},
-    "archive_existing_versions": {"type": "boolean"}
-}
+def _assert_int(x):
+    assert isinstance(x, int)
 
+def _assert_bool(x):
+    assert isinstance(x, bool)
 
-def _build_schema(name, additional_schema = {}):
+def _assert_float(x):
+    assert isinstance(x, float)
 
-    schema = copy.deepcopy(_DEFAULT_SCHEMAS.get(name, {}))
+def _assert_array(x):
+    assert isinstance(x, list)
 
-    for k, v in additional_schema.items():
-        schema[k] = v
-
-    return schema
+def _assert_required(x):
+    assert x is not None
 
 def _validate_param_against_schema(schema, param, value):
-    ## If the schema is empty, do nothing
-    if not schema:
-        return True
+    for f in schema:
+        try:
+            f(value)
+        except AssertionError as e:
+            raise MlflowException(
+                message="Invalid value {} for '{}' supplied.".format(value, param) + "\nSee the API docs for more information on parameters.",
+                error_code=INVALID_PARAMETER_VALUE
+            )
 
-    try:
-        Draft202012Validator(schema).validate(value)
-    except ValidationError as e:
-        raise MlflowException(
-            message="Invalid value for '{}' supplied.".format(param) + "\n" + e.message,
-            error_code=INVALID_PARAMETER_VALUE
-        )
-
-    return True
+    return None
 
 
 def _get_request_json(flask_request=request):
     return flask_request.get_json(force=True, silent=True)
 
-def _get_request_message(request_message, flask_request=request, **additional_schema):
+def _get_request_message(request_message, flask_request=request, schema = {}):
     from querystring_parser import parser
 
     if flask_request.method == "GET" and len(flask_request.query_string) > 0:
@@ -368,8 +337,8 @@ def _get_request_message(request_message, flask_request=request, **additional_sc
         request_json = {}
 
     for k, v in request_json.items():
-        schema = _build_schema(k, additional_schema.get(k, {}))
-        _validate_param_against_schema(schema, k, v)
+        if (k in schema.keys()):
+            _validate_param_against_schema(schema[k], k, v)
 
     parse_dict(request_json, request_message)
 
@@ -487,7 +456,16 @@ def _not_implemented():
 @_disable_if_artifacts_only
 def _create_experiment():
 
-    request_message = _get_request_message(CreateExperiment(), name = _REQUIRED_PARAM_DEFAULT_SCHEMA)
+    request_message = _get_request_message(
+        CreateExperiment(),
+        schema = {
+            "name": [
+                _assert_required,
+                _assert_string
+            ]
+        }
+    )
+
     tags = [ExperimentTag(tag.key, tag.value) for tag in request_message.tags]
     experiment_id = _get_tracking_store().create_experiment(
         request_message.name, request_message.artifact_location, tags
@@ -501,7 +479,15 @@ def _create_experiment():
 @catch_mlflow_exception
 @_disable_if_artifacts_only
 def _get_experiment():
-    request_message = _get_request_message(GetExperiment(), experiment_id = _REQUIRED_PARAM_DEFAULT_SCHEMA)
+    request_message = _get_request_message(
+        GetExperiment(),
+        schema = {
+            "experiment_id": [
+                _assert_required,
+                _assert_string
+            ]
+        }
+    )
     response_message = GetExperiment.Response()
     experiment = _get_tracking_store().get_experiment(request_message.experiment_id).to_proto()
     response_message.experiment.MergeFrom(experiment)
@@ -513,7 +499,15 @@ def _get_experiment():
 @catch_mlflow_exception
 @_disable_if_artifacts_only
 def _get_experiment_by_name():
-    request_message = _get_request_message(GetExperimentByName(), experiment_name = _REQUIRED_PARAM_DEFAULT_SCHEMA)
+    request_message = _get_request_message(
+        GetExperimentByName(),
+        schema = {
+            "experiment_name": [
+                _assert_required,
+                _assert_string
+            ]
+        }
+    )
     response_message = GetExperimentByName.Response()
     store_exp = _get_tracking_store().get_experiment_by_name(request_message.experiment_name)
     if store_exp is None:
@@ -531,7 +525,15 @@ def _get_experiment_by_name():
 @catch_mlflow_exception
 @_disable_if_artifacts_only
 def _delete_experiment():
-    request_message = _get_request_message(DeleteExperiment(), experiment_id = _REQUIRED_PARAM_DEFAULT_SCHEMA)
+    request_message = _get_request_message(
+        DeleteExperiment(),
+        schema = {
+            "experiment_id": [
+                _assert_required,
+                _assert_string
+            ]
+        }
+    )
     _get_tracking_store().delete_experiment(request_message.experiment_id)
     response_message = DeleteExperiment.Response()
     response = Response(mimetype="application/json")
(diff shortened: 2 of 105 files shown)


mlflow/server/handlers.py
```diff
@@ -261,12 +265,67 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
     except UnsupportedModelRegistryStoreURIException:
         pass
 
+_STRING_TYPE_DEFAULT_SCHEMA = {"type": "string"}
+_INTEGER_TYPE_DEFAULT_SCHEMA = {"type": "integer"}
+_REQUIRED_PARAM_DEFAULT_SCHEMA = {"required": True}
```
mrkaye97, line 270: Adds some defaults so we can reuse these in a bunch of places

mlflow/server/handlers.py
```diff
@@ -261,12 +265,67 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+    "run_link": _STRING_TYPE_DEFAULT_SCHEMA,
+    "version": _STRING_TYPE_DEFAULT_SCHEMA,
+    "order_by": {"type": "array", "items": _STRING_TYPE_DEFAULT_SCHEMA},
+    "archive_existing_versions": {"type": "boolean"}
+}
+
+
+def _build_schema(name, additional_schema = {}):
+
+    schema = copy.deepcopy(_DEFAULT_SCHEMAS.get(name, {}))
```
mrkaye97, line 304: Is there a better way to do this than with a `deepcopy`?

mlflow/server/handlers.py
```diff
@@ -261,12 +265,67 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+
+    for k, v in additional_schema.items():
```
mrkaye97, line 306: The idea here is to first look up the default schema associated with whatever the param we're trying to validate is, and then if there are any additional pieces to the schema to add, add them before running the validator

mlflow/server/handlers.py
```diff
@@ -261,12 +265,67 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+            error_code=INVALID_PARAMETER_VALUE
+        )
+
 
 def _get_request_json(flask_request=request):
     return flask_request.get_json(force=True, silent=True)
 
-
-def _get_request_message(request_message, flask_request=request):
+def _get_request_message(request_message, flask_request=request, **additional_schema):
```
mrkaye97, line 328: using kwargs here seemed like a good option -- then we can just specify things like `name = {"required": True}` to add to the schema

mlflow/server/handlers.py
mrkaye97: Also: I'm not actually sure if it should be `True` or `"true"` here, but I think `True` works

mlflow/server/handlers.py
```diff
@@ -261,12 +265,67 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+    if not schema:
+        return True
+
+    try:
+        Draft202012Validator(schema).validate(param)
+    except ValidationError as e:
+        raise MlflowException(
+            message=e.message,
+            error_code=INVALID_PARAMETER_VALUE
+        )
```
mrkaye97, line 322: Design nits: This will return the exception that the validator throws as the message, which I think is okay and probably easier than returning some custom message.

Separately, I don't love throwing an error here as a side effect, but sometimes returning `True`. I can't think of a better pattern though. The only other thing that comes to mind would be to either return the error message or a placeholder like "ok", and then check if the return of this call is `"ok"` and pass, otherwise raise the exception. Open to ideas on this one

mrkaye97:
@dbczumar Moved this over to here. I was having some issues with huge diffs when trying to rebase on top of `master` which was causing GH to hang ¯\_(ツ)_/¯

I left some comments about design decisions I made and a couple of open questions. There are also two failing tests that I was seeing on local, caused by trying to log `NaN` as a number.

Let me know what the next steps for this should be, or if you've got any feedback! Also, who should I be requesting for review?

mrkaye97:
Oh and another question: Do I need to add `jsonschema` to the `requirements.txt` files? Or which, if any?

mlflow/server/handlers.py
```diff
@@ -261,12 +265,69 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+def _build_schema(name, additional_schema = {}):
+
+    schema = copy.deepcopy(_DEFAULT_SCHEMAS.get(name, {}))
+
+    for k, v in additional_schema.items():
+        schema[k] = v
+
+    return schema
+
+def _validate_param_against_schema(schema, param, value):
```
mrkaye97, line 311: I opted to thread both the parameter (name) and the value through, so that we could supply a more helpful error to the user by telling them _which parameter_ was the one causing the failure

dbczumar:
> Oh and another question: Do I need to add `jsonschema` to the `requirements.txt` files? Or which, if any?

Hi @mrkaye97 ! Thanks for the awesome PR. Is it possible to implement this validation without introducing additional library dependencies? We try to avoid new library dependencies in core MLflow modules (such as `mlflow server`) unless it's absolutely necessary because MLflow is used in a wide variety of system / software environments.

mrkaye97:
> > Oh and another question: Do I need to add `jsonschema` to the `requirements.txt` files? Or which, if any?
> 
> Hi @mrkaye97 ! Thanks for the awesome PR. Is it possible to implement this validation without introducing additional library dependencies? We try to avoid new library dependencies in core MLflow modules (such as `mlflow server`) unless it's absolutely necessary because MLflow is used in a wide variety of system / software environments.

@dbczumar Yeah, I definitely can. But just to push back a little, how strongly do you feel about this? On one hand, I agree with you that I'd prefer to not introduce a dependency. But I guess it depends on what "absolutely necessary" means. In this case, it'd be pretty easy to roll our own validation. But that also means we'd increase the maintenance costs of this code, in addition to increasing the complexity of adding additional validation for future endpoints, schemas, etc.

Again, I'm happy to go either way, but just wanted to throw this out there before leaning into rolling our own

dbczumar:
> > > Oh and another question: Do I need to add `jsonschema` to the `requirements.txt` files? Or which, if any?
> > 
> > 
> > Hi @mrkaye97 ! Thanks for the awesome PR. Is it possible to implement this validation without introducing additional library dependencies? We try to avoid new library dependencies in core MLflow modules (such as `mlflow server`) unless it's absolutely necessary because MLflow is used in a wide variety of system / software environments.
> 
> @dbczumar Yeah, I definitely can. But just to push back a little, how strongly do you feel about this? On one hand, I agree with you that I'd prefer to not introduce a dependency. But I guess it depends on what "absolutely necessary" means. In this case, it'd be pretty easy to roll our own validation. But that also means we'd increase the maintenance costs of this code, in addition to increasing the complexity of adding additional validation for future endpoints, schemas, etc.
> 
> Again, I'm happy to go either way, but just wanted to throw this out there before leaning into rolling our own

@mrkaye97 Thanks for digging further here. Apologies for the vague "absolutely necessary" designation. The general guidance is that we shouldn't add a new dependency if: 1. there's a relatively straightforward way to correctly implement the same logic provided by a third-party library, or 2. we can easily extract the relevant modules from the third-party library into the project directly without significant ramifications (e.g. breach of licensing / TOS) or maintenance challenges (e.g. perhaps the library is updated frequently with critical fixes / improvements, and it would be cumbersome to continually update the inlined modules).

In this case, (1) seems to apply, given the limited set of validation functionality that we require for this use case. Let me know if you have any questions, and thanks for your flexibility :D

mrkaye97:
> > > > Oh and another question: Do I need to add `jsonschema` to the `requirements.txt` files? Or which, if any?
> > > 
> > > 
> > > Hi @mrkaye97 ! Thanks for the awesome PR. Is it possible to implement this validation without introducing additional library dependencies? We try to avoid new library dependencies in core MLflow modules (such as `mlflow server`) unless it's absolutely necessary because MLflow is used in a wide variety of system / software environments.
> > 
> > 
> > @dbczumar Yeah, I definitely can. But just to push back a little, how strongly do you feel about this? On one hand, I agree with you that I'd prefer to not introduce a dependency. But I guess it depends on what "absolutely necessary" means. In this case, it'd be pretty easy to roll our own validation. But that also means we'd increase the maintenance costs of this code, in addition to increasing the complexity of adding additional validation for future endpoints, schemas, etc.
> > Again, I'm happy to go either way, but just wanted to throw this out there before leaning into rolling our own
> 
> @mrkaye97 Thanks for digging further here. Apologies for the vague "absolutely necessary" designation. The general guidance is that we shouldn't add a new dependency if: 1. there's a relatively straightforward way to correctly implement comparable logic, or 2. we can easily extract the relevant modules from the third-party library into the project directly without significant ramifications (e.g. breach of licensing / TOS) or maintenance challenges (e.g. perhaps the library is updated frequently with critical fixes / improvements, and it would be cumbersome to continually update the inlined modules).
> 
> In this case, (1) seems to apply, given the limited set of validation functionality that we require for this use case. Let me know if you have any questions, and thanks for your flexibility :D

Sure, that makes sense! Let me report back in a few days on this. Thanks for the feedback!

mrkaye97:
alright @dbczumar, what do you think of this approach? IMO, it's a lot less elegant, but at least it rolls our own instead of relying on a library. Still have the same two failing tests on local (`NaN` supplied, number required) but everything else seems to be working.

Also a TODO: I'm not sure how to best extend this for `log_batch`, since it takes a list of (e.g.) metrics, which seems like a pain. It expects a list of metrics, where each metric is a dictionary, and we want to validate each k-v pair of that dictionary. Haven't figured out the best way to do that yet.

LMK what you think whenever you get a chance :+1:

mlflow/server/handlers.py
```diff
@@ -261,12 +261,47 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+    assert isinstance(x, list)
+
+def _assert_required(x):
+    assert x is not None
+
+def assert_max_1k(x):
+    assert x <= 1000
+
+def _assert_max_50k(x):
+    assert x <= 50000
```
dbczumar, line 286: Nit: Can we consolidate these as:

```
def _assert_less_than_or_equal(x, max_value):
    assert x <= max_value
```

Then, we can pass `lambda x: _assert_less_than_or_equal(x, 50000)`, etc. as part of the `schema` parameter of `_get_request_message`.

mlflow/server/handlers.py
```diff
@@ -522,7 +623,18 @@ def _create_run():
...
 def _update_run():
-    request_message = _get_request_message(UpdateRun())
+    request_message = _get_request_message(
+        UpdateRun(),
+        schema = {
+            "run_id": [
+                _assert_required,
+                _assert_string
+            ],
+            "end_time": [
```
dbczumar, line 633: Can we validate `status` too?

dbczumar (commented):
> alright @dbczumar, what do you think of this approach? IMO, it's a lot less elegant, but at least it rolls our own instead of relying on a library. Still have the same two failing tests on local (`NaN` supplied, number required) but everything else seems to be working.
> 
> Also a TODO: I'm not sure how to best extend this for `log_batch`, since it takes a list of (e.g.) metrics, which seems like a pain. It expects a list of metrics, where each metric is a dictionary, and we want to validate each k-v pair of that dictionary. Haven't figured out the best way to do that yet.
> 
> LMK what you think whenever you get a chance 👍

@mrkaye97 This is fantastic! IMO, your implementation is quite elegant. Let me know if you'd like help / input regarding NaNs (as far as I can tell, NaNs are interpreted as floats when parsing JSON).

Regarding LogBatch, I think we can add comparable per-field validation logic to the body of the LogBatch handler. It may not look super nice, but I think it's the simplest approach.

Can't wait to merge this once NaNs and LogBatch have been addressed. Thanks again!

mlflow/server/handlers.py
mrkaye97: hehe, nice idea. will add this. I was trying to do this to no avail 🤦 :

```python
lambda x: assert x < 50000
```

mrkaye97:
> > alright @dbczumar, what do you think of this approach? IMO, it's a lot less elegant, but at least it rolls our own instead of relying on a library. Still have the same two failing tests on local (`NaN` supplied, number required) but everything else seems to be working.
> > Also a TODO: I'm not sure how to best extend this for `log_batch`, since it takes a list of (e.g.) metrics, which seems like a pain. It expects a list of metrics, where each metric is a dictionary, and we want to validate each k-v pair of that dictionary. Haven't figured out the best way to do that yet.
> > LMK what you think whenever you get a chance 👍
> 
> @mrkaye97 This is fantastic! IMO, your implementation is quite elegant. Let me know if you'd like help / input regarding NaNs (as far as I can tell, NaNs are interpreted as floats when parsing JSON).
> 
> Regarding LogBatch, I think we can add comparable per-field validation logic to the body of the LogBatch handler. It may not look super nice, but I think it's the simplest approach.
> 
> Can't wait to merge this once NaNs and LogBatch have been addressed. Thanks again!

Works for me! I might need a few days -- going to be a busy week for me. Hopefully should have this over the finish line this weekend though. Thanks for the feedback.

On the `NaN` thing: I guess the question is whether we want to allow them or not. IMO, if MLFlow expects a number for a field, then `NaN` _should_ throw an error -- it's explicitly _not_ a number, after all. If we decide that not allowing `NaN` is the right way to go, I can probably just update those two tests. If we want to allow `NaN` as a float, I can allow them

mrkaye97:
@dbczumar Update on this: 
* I've added validation for `log_batch` in the best way I could think of (🤮 )
* I've used some lambdas to clean up repeated code a little, per your suggestion
* I _think_ I've got `status` being validated right (can I just check if it's a string? I think it's a special class)

On the `NaN`s: I'm not actually sure why that's failing tests -- I verified that Python parses `"NaN"` and `"nan"` as floats, and in a repl, running `isinstance(float("nan"), float)` works fine. I also tried parsing JSON in the repl directly, but that also seemed fine. Here's what I tried:

```python
import json

x = json.loads('{"Key": NaN}')

print(x)
print(type(x["Key"]))

## Checking the same assertion that the validator is performing
print(isinstance(x["Key"], float))

## To make sure that Python does, in fact, consider NaN to be a float when we tell it
print(float("nan"))
print(isinstance(float("nan"), float))

## To make sure it's actually the float validator failing
print(float("nan") is not None)
```

So yeah, I'm not actually sure what's going wrong here. Any ideas?

mrkaye97:
Another thing: The R tests are failing because of a bug in the R package itself, not in this change.

The issue is here: https://github.com/mlflow/mlflow/blob/82d402173880991f0139b5f0268e4d28c7672645/mlflow/R/mlflow/R/tracking-rest.R#L68

Basically, your R package converts all of the elements of the body to character vectors, which is now failing because it tries to make an API call with (e.g.) that start time as `"100"` instead of `100`, which now fails. Honestly, I'm not sure why the type conversion was happening in the first place though. My suggestion would be to refactor how `mlflow_rest` works as follows (how we've done it in our `lightMLFlow` package): https://github.com/collegevine/lightMLFlow/blob/main/R/rest.R#L113-L118 I will put in that change.

See this `reprex`:
``` r
data <- list(
    foo = "bar",
    baz = 314159
)

rapply(data, as.character, how = "replace")
#> $foo
#> [1] "bar"
#> 
#> $baz
#> [1] "314159"
```

<sup>Created on 2022-03-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mlflow/R/mlflow/R/tracking-rest.R
```diff
@@ -60,37 +60,57 @@ mlflow_rest <- function( ..., client, query = NULL, data = NULL, verb = "GET", v
...
     GET = function() {
-      GET( api_url, query = query, mlflow_rest_timeout(), config = rest_config$config,
-           req_headers)
+     GET(
+       api_url,
+       query = query,
+       mlflow_rest_timeout(),
+       config = rest_config$config,
+       req_headers
+     )
```
mrkaye97, line 69: Proposed drive-by fixing these. Honestly, I'm not sure why these were implemented how they were in the first place

dbczumar:
> Another thing: The R tests are failing because of a bug in the R package itself, not in this change.
> 
> The issue is here:
> 
> https://github.com/mlflow/mlflow/blob/82d402173880991f0139b5f0268e4d28c7672645/mlflow/R/mlflow/R/tracking-rest.R#L68
> 
> Basically, your R package converts all of the elements of the body to character vectors, which is now failing because it tries to make an API call with (e.g.) that start time as `"100"` instead of `100`, which now fails. Honestly, I'm not sure why the type conversion was happening in the first place though. My suggestion would be to refactor how `mlflow_rest` works as follows (how we've done it in our `lightMLFlow` package): https://github.com/collegevine/lightMLFlow/blob/main/R/rest.R#L113-L118 I will put in that change.
> 
> See this `reprex`:
> 
> ```r
> data <- list(
>     foo = "bar",
>     baz = 314159
> )
> 
> rapply(data, as.character, how = "replace")
> #> $foo
> #> [1] "bar"
> #> 
> #> $baz
> #> [1] "314159"
> ```
> 
> Created on 2022-03-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)

@mrkaye97 Thank you for looking into this! My interpretation is that, if the MLflow server currently accepts string request parameters with numeric contents (e.g. `"100"`) and successfully converts them to the corresponding int / float values for downstream operations, then we should preserve that behavior in this PR. The R client may not be the only source of such requests; for example, third-party tools / user workflows may be making similar requests with "numeric" strings, and we don't want to break them.

Rather than drive-by fixing the R client (I think we can revert that for the scope of this PR), can we instead adjust how we're handling type validation to rely on the existing `parse_dict()` (https://github.com/mlflow/mlflow/blob/2e175e7fd65d86d12d204b648aa5ff73063a7dc9/mlflow/utils/proto_json_utils.py#L150-L153) implementation as the source of truth? If `parse_dict()` succeeds, we should proceed with handling the request. On the other hand, if `parse_dict()` fails, then we can use the expected schema / type information for the request to return an informative error to the user.

Apologies for the additional wrinkle here. Thank you for all your hard work! Let me know if you have any questions.

mrkaye97:
thanks for the feedback @dbczumar -- I'll revert the R changes in favor of your suggestion to preserve existing behavior. Just to make sure I'm understanding how you're thinking about this though: your assumption is that if`parse_dict` succeeds (as in, doesn't throw an error) then the types are good to go? 

Do you have any idea how prevalent this issue is? Is it only for numeric types? or might we break other things like `"true"` --> `True`? I can reliably get the API to throw a `ParseError` if I (e.g.) specify `123` for a string parameter, but I'm concerned that catching _all_ parse errors and treating them as type errors is dangerous.

It seems like maybe a better solution (if we could think of all of the cases) would be to change `_assert_integer` (for example) to `_assert_integerish` and try to first parse the thing that we're asserting on to an int. If the parsing fails, we throw a type error and 400, and if parsing succeeds then the integer type assertion will trivially succeed as well. What do you think of that approach? I'd think that we'd need `_assert_floatish`, `_assert_integerish`, and `_assert_booleanish` and that might be it? It'd keep the logic more concise I think that way ¯\_(ツ)_/¯

dbczumar:
> thanks for the feedback @dbczumar -- I'll revert the R changes in favor of your suggestion to preserve existing behavior. Just to make sure I'm understanding how you're thinking about this though: your assumption is that if`parse_dict` succeeds (as in, doesn't throw an error) then the types are good to go?
> 
> Do you have any idea how prevalent this issue is? Is it only for numeric types? or might we break other things like `"true"` --> `True`? I can reliably get the API to throw a `ParseError` if I (e.g.) specify `123` for a string parameter, but I'm concerned that catching _all_ parse errors and treating them as type errors is dangerous.
> 
> It seems like maybe a better solution (if we could think of all of the cases) would be to change `_assert_integer` (for example) to `_assert_integerish` and try to first parse the thing that we're asserting on to an int. If the parsing fails, we throw a type error and 400, and if parsing succeeds then the integer type assertion will trivially succeed as well. What do you think of that approach? I'd think that we'd need `_assert_floatish`, `_assert_integerish`, and `_assert_booleanish` and that might be it? It'd keep the logic more concise I think that way ¯_(ツ)_/¯

Thanks @mrkaye97 ! Correct, if `parse_dict` succeeds, we can assume the types are good to go. Because I'm not 100% sure of the scope / prevalence here, I'd be wary of trying to implement our own `_assert_integerish`, `_assert_booleanish`, etc logic here. I understand your concern about treating all `parse_dict` failures as type errors, and I think we can resolve it by carefully wording our error message when parse_dict fails. For example, we could say something to the effect of 

```
Failed to parse request body. Please verify that your LogMetric request includes a valid JSON body with fields satisfying the following specifications:

run_id: String, required
key: String, required
value: Double, required
timestamp: Int, required
step: Int
```

This allows us to provide an informative message for most cases while keeping the implementation cheap / simple. If the user can verify that their request looks good and it still isn't working, they can always file an issue.

For more specific cases where we know the cause of the failure - e.g. the field is missing but required or fails to satisfy some numeric constraint - we can use the more specific error messages that have already been implemented in your PR.

Let me know what your thoughts are here.

mrkaye97:
> Thanks @mrkaye97 ! Correct, if `parse_dict` succeeds, we can assume the types are good to go. Because I'm not 100% sure of the scope / prevalence here, I'd be wary of trying to implement our own `_assert_integerish`, `_assert_booleanish`, etc logic here. I understand your concern about treating all `parse_dict` failures as type errors, and I think we can resolve it by carefully wording our error message when parse_dict fails. For example, we could say something to the effect of
> 
> ```
> Failed to parse request body. Please verify that your LogMetric request includes a valid JSON body with fields satisfying the following specifications:
> 
> run_id: String, required
> key: String, required
> value: Double, required
> timestamp: Int, required
> step: Int
> ```
> 
> This allows us to provide an informative message for most cases while keeping the implementation cheap / simple. If the user can verify that their request looks good and it still isn't working, they can always file an issue.
> 
> For more specific cases where we know the cause of the failure - e.g. the field is missing but required or fails to satisfy some numeric constraint - we can use the more specific error messages that have already been implemented in your PR.
> 
> Let me know what your thoughts are here.

Gotcha, that makes sense to me :+1: Agreed that trying to cover all the cases is risky.

I imagine that I'm going to need to rewire some things a little bit, since this conditional logic (only check types on parse failure) is a little more involved than the original logic was. I'll work on this over the next few days and report back!

mrkaye97:
@dbczumar Circling back: I think this approach actually raises some bigger issues buried deeper in the code. For instance: If I make a call to `log_metric` with `name = 31`, `parse_dict` succeeds. As such, the validation for the type of `name` is skipped (even though it should fail, but that's okay in theory).

And then the problem occurs: Downstream, I end up with this:

```python
>       if name is None or not _VALID_PARAM_AND_METRIC_NAMES.match(name):
E       TypeError: expected string or bytes-like object

name       = 31
```

Which I think results in a 500. I actually can't repro this from an R client making the same calls, which I'm a little confused by. In that case, I get back a 400 with the correct validation error. Got any ideas on this? 

Separately, it's weird to me from a design perspective that this validation would happen twice for some calls (but not others?). That seems unnecessary, and it'll result in odd UX for users who will get different error messages for the "same" error, depending on the endpoint they're hitting.

mlflow/server/handlers.py
```diff
@@ -261,12 +262,62 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+        #  see: https://github.com/mlflow/mlflow/pull/5458#issuecomment-1080880870
+        if f in [
+            _assert_int,
+            _assert_string,
+            _assert_bool,
+            _assert_float,
+            _assert_array,
+            _assert_item_type_string
+        ] and parsing_succeeded:
+            continue
```
mrkaye97, line 304: This was the best way I could think of to do this that didn't require a huge refactor

dbczumar:
> @dbczumar Circling back: I think this approach actually raises some bigger issues buried deeper in the code. For instance: If I make a call to `log_metric` with `key = 31`, `parse_dict` succeeds. As such, the validation for the type of `key` is skipped (even though it should fail, but that's okay in theory).
> 
> And then the problem occurs: Downstream, I end up with this:
> 
> ```python
> >       if name is None or not _VALID_PARAM_AND_METRIC_NAMES.match(name):
> E       TypeError: expected string or bytes-like object
> 
> name       = 31
> ```
> 
> Which I think results in a 500. I actually can't repro this from an R client making the same calls, which I'm a little confused by. In that case, I get back a 400 with the correct validation error. Got any ideas on this?
> 
> Separately, it's weird to me from a design perspective that this validation would happen twice for some calls (but not others?). That seems unnecessary, and it'll result in odd UX for users who will get different error messages for the "same" error, depending on the endpoint they're hitting.
> 
> IMO, the bottom line is that it's not clear to me that it's safe to assume that if `parse_dict` succeeds, all is well with the types

@mrkaye97 Thanks for digging into this! I think the behavior you're seeing may be related to client-side validation in Python that we're removing via https://github.com/mlflow/mlflow/pull/5593; we recognize that the backend should be the source of truth regarding what constitutes valid input for MLflow APIs.

I tried checking out your PR branch, starting an MLflow server, and making an HTTP request via Python `requests` using the content you described:

```
import requests

resp = requests.post(
    url="http://localhost:5000/api/2.0/mlflow/runs/log-metric",
    json={
        "run_id": "1234",
        "key": 31,
        "value": 5,
    }
)

print(resp.text)
print(resp.status_code)
```

This yielded a 400:

```
{"error_code": "INVALID_PARAMETER_VALUE", "message": "Invalid value 31 for parameter 'key' supplied. Hint: Value was of type str\nSee the API docs for more information on specifying parameters."}
400
```

In summary, after removing client-side validation logic by merging https://github.com/mlflow/mlflow/pull/5593, the logic you've implemented appears to function as expected. Please let me know if you still have concerns here / are aware of other confusing or problematic behaviors.

mlflow/server/handlers.py
dbczumar: Are these methods invoked at any point? If not, can we remove them from the PR entirely?

dbczumar: Nevermind, misread the logic. This makes sense :)

mlflow/server/handlers.py
```diff
@@ -261,12 +262,62 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+
+def _assert_less_than_or_equal(x, max):
+    assert x <= max
+
+def _assert_item_type_string(x):
+    for item in x:
+        _assert_string(item)
+
+
+def _validate_param_against_schema(schema, param, value, parsing_succeeded = False):
```
dbczumar, line 291: Nit: Can we add a docstring  to this method and document each parameter, particularly `parsing_succeeded`? For clarity, can we also rename this to `proto_parsing_succeeded`?

mlflow/server/handlers.py
```diff
@@ -261,12 +262,62 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+        #  succeeeded, we can safely skip re-checking types
+        #  see: https://github.com/mlflow/mlflow/pull/5458#issuecomment-1080880870
+        if f in [
+            _assert_int,
+            _assert_string,
+            _assert_bool,
+            _assert_float,
+            _assert_array,
+            _assert_item_type_string
+        ] and parsing_succeeded:
```
dbczumar, line 303: Nit: Can we convert this list to a `set()` and pull it out of this method into a global variable so that it's only constructed once? That will make this check cheaper.

mlflow/server/handlers.py
```diff
@@ -773,7 +1125,18 @@ def _log_batch():
...
-    request_message = _get_request_message(LogModel())
+    request_message = _get_request_message(
+        LogModel(),
+        schema = {
+            "run_id": [
+                _assert_string,
+                _assert_required
+            ],
+            "model_json": [
+                _assert_string
```
dbczumar, line 1136: While it's not documented, `model_json` is required too. Can we add `_assert_required`?

mlflow/server/handlers.py
```diff
@@ -756,8 +1053,63 @@ def _get_artifact_repo(run):
...
 @_disable_if_artifacts_only
 def _log_batch():
+
+    def _assert_metrics_correctly_constructed(metrics):
+        for m in metrics:
+            print("This is a metric: ", m)
+            ## Key
+            k = m["key"]
+            _assert_required(k)
+            _assert_string(k)
```
dbczumar, line 1063: We can remove type validations from these methods since ParseDict will handle type validation. `assert_required` is still helpful.

mlflow/server/handlers.py
```diff
@@ -756,8 +1053,63 @@ def _get_artifact_repo(run):
...
+            "metrics": [
+                _assert_array,
+                _assert_metrics_correctly_constructed
+            ],
+            "params": [
+                _assert_array,
+                _assert_params_correctly_constructed
+            ],
+            "tags": [
+                _assert_array
```
dbczumar, line 1109: Can we add `_assert_tags_fields_present`?

mlflow/server/handlers.py
dbczumar: Since we're only going to be checking that required fields are present, perhaps `_assert_metrics_fields_present` is a more descriptive name to use for this method, ditto for params

mlflow/server/handlers.py
```diff
@@ -848,7 +1244,18 @@ def _update_registered_model():
...
-    request_message = _get_request_message(RenameRegisteredModel())
+    request_message = _get_request_message(
+        RenameRegisteredModel(),
+        schema = {
+            "name": [
+                _assert_string,
+                _assert_required
+            ],
+            "new_name": [
+                _assert_string
```
dbczumar, line 1255: While not documented, this is required. Can we add `_assert_required`?

mlflow/server/handlers.py
```diff
@@ -261,12 +262,62 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
     except UnsupportedModelRegistryStoreURIException:
         pass
 
+def _assert_string(x):
+    assert isinstance(x, str)
+
+def _assert_int(x):
+    assert isinstance(x, int)
```
dbczumar, line 269: Can we also include the case where `x` is an int-ish string for consistency with protobuf parsing?

mlflow/server/handlers.py
```diff
@@ -261,12 +262,62 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+
+def _assert_bool(x):
+    assert isinstance(x, bool)
+
+def _assert_float(x):
+    assert isinstance(x, float)
```
dbczumar, line 275: Can we also include the case where `x` is a float-ish string for consistency with protobuf parsing?

mlflow/server/handlers.py
```diff
@@ -883,7 +1310,30 @@ def _list_registered_models():
 @catch_mlflow_exception
 @_disable_if_artifacts_only
 def _search_registered_models():
-    request_message = _get_request_message(SearchRegisteredModels())
+
+    def _assert_item_type_string(x):
+        for item in x:
+            _assert_string(x)
```
dbczumar, line 1316: Already defined above. Can we remove this?

mlflow/server/handlers.py
```diff
@@ -261,12 +262,62 @@ def initialize_backend_stores(backend_store_uri=None, default_artifact_root=None
...
+
+def _assert_required(x):
+    assert x is not None
+
+def _assert_less_than_or_equal(x, max):
+    assert x <= max
+
+def _assert_item_type_string(x):
+    for item in x:
+        _assert_string(item)
```
dbczumar, line 288: Nit:

```suggestion
    assert all([isinstance(item, string) for item in x])
```

tests/tracking/test_rest_tracking.py
```diff
@@ -270,6 +274,25 @@ def test_log_metrics_params_tags(mlflow_client, backend_store_uri):
...
+    with pytest.raises(MlflowException, match=r"Invalid value for 'value' supplied"):
+        mlflow_client.log_metric(run_id, key="foo", value=True, timestamp=31, step=41)
+
+    with pytest.raises(MlflowException, match=r"Invalid value for 'run_id' supplied"):
+        mlflow_client.log_param(31, key="foo", value="bar", timestamp=59, step=26)
+
+    with pytest.raises(MlflowException, match=r"Invalid value for 'key' supplied"):
+        mlflow_client.log_param(run_id, key=31, value="bar", timestamp=59, step=26)
+
+    with pytest.raises(MlflowException, match=r"Invalid value for 'value' supplied"):
```
dbczumar, line 293: Can we add test cases for required fields and thresholds, as well as list-typed fields?

dbczumar (commented):
@mrkaye97 This is fantastic! I left a few small comments, but I think we should be all set to merge once they're addressed. Thanks for all your hard work here!

BenWilson2:
Hi @mrkaye97 #5593 has been merged into master.

mrkaye97:
@BenWilson2 Thanks! I'll rebase on top of `master` :) 

@dbczumar sounds good! give me a day or two to circle back as usual. thanks for all the feedback!

mlflow/server/handlers.py
mrkaye97: Do you want to include this in a separate method? Or just change this (and the float one) to be `_assert_intlike` and try something like this?:

```python
def _assert_int(x):
    try:
        x = int(x)
    except ValueError:
        pass
        
    assert isinstance(x, int)
```

mlflow/server/handlers.py
mrkaye97: does adding an `_assert_params_tags_correctly_constructed` (changed from `_assert_params_correctly_constructed`) work? The `RunTag` and `Param` data types look the same afaict

tests/tracking/test_rest_tracking.py
mrkaye97: Small thing on this: It seems like proto parsing might be automatically converting singletons to lists (e.g. `None` --> `[None]`) which then fails downstream. On one hand, I could always validate on the list types with `_assert_array` to fix this issue, but that'd potentially break downstream clients who are specifying a `experiment_ids` as a single `experiment_id` for example.

Any thoughts on this one? One option would be to just not test it... I can't think of something where wrapping a singleton in a list would break, which is the only way that I can test throwing `400` for something that should be a list

mrkaye97: And another: If I specify something like `metrics="foo"` in `log_batch`, I get an `AttributeError` downstream:

```python
> metric_protos = [metric.to_proto() for metric in metrics]
E   AttributeError: 'str' object has no attribute 'to_proto'
```

Unsure what to do about this one either. I think these are relics of the same top level issue, which is that assuming `parse_dict` not throwing an error meaning that the types are okay isn't actually a safe assumption. 

I also have other tests failing on (e.g.) `create_experiment`, where I specify `name=3141` and it throws a type error downstream. Can you repro that issue? I tried rebasing, but maybe I'm not caught up with `master`

dbczumar: @mrkaye97 These are instances of client-side exceptions thrown before the client ever sends a request to the MLflow server; here's the stack trace for the `create_experiment` scenario:

```
  File "log.py", line 5, in <module>
    mlflow.create_experiment(name=3141)
  File "/Users/corey.zumar/mlflow/mlflow/tracking/fluent.py", line 1023, in create_experiment
    return MlflowClient().create_experiment(name, artifact_location, tags)
  File "/Users/corey.zumar/mlflow/mlflow/tracking/client.py", line 507, in create_experiment
    return self._tracking_client.create_experiment(name, artifact_location, tags)
  File "/Users/corey.zumar/mlflow/mlflow/tracking/_tracking_service/client.py", line 176, in create_experiment
    return self.store.create_experiment(
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/rest_store.py", line 97, in create_experiment
    CreateExperiment(name=name, artifact_location=artifact_location, tags=tag_protos)
TypeError: 3141 has type int, but expected one of: bytes, unicode
```

Similarly, here's the stack trace for the `log_batch` scenario:

```
  File "log.py", line 8, in <module>
    MlflowClient().log_batch("123", metrics="foo")
  File "/Users/corey.zumar/mlflow/mlflow/tracking/client.py", line 918, in log_batch
    self._tracking_client.log_batch(run_id, metrics, params, tags)
  File "/Users/corey.zumar/mlflow/mlflow/tracking/_tracking_service/client.py", line 292, in log_batch
    self.store.log_batch(run_id=run_id, metrics=metrics, params=params, tags=tags)
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/rest_store.py", line 303, in log_batch
    metric_protos = [metric.to_proto() for metric in metrics]
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/rest_store.py", line 303, in <listcomp>
    metric_protos = [metric.to_proto() for metric in metrics]
AttributeError: 'str' object has no attribute 'to_proto'
```

The client is trying & failing to convert the user-specified values into a JSON payload by using proto as an intermediary. If we want to improve the clarity / fidelity of the exceptions in this case, we could apply similar exception handling logic for proto conversion failures to the client. Please let me know if you're interested in taking this on as a follow-up PR.

mlflow/server/handlers.py
dbczumar: I think this is great!

dbczumar (approved):
LGTM! Thanks for all your hard work, @mrkaye97! This is awesome! I'll go ahead and clean up some lint errors and get this merged.

dbczumar (commented):
@mrkaye97 I resolved the lint errors and updated the tests to make use of raw REST requests to bypass client-side validations when testing server-side validations. I also updated the server-side validation logic to check all fields associated with an `_assert_required` function, even if they aren't present in the request body (previously, we only validated a field if it was present in the request, but we should throw if a required field is absent), and I improved the error message for the case where `_assert_required` fails. Finally, I added validations for the `CreateExperiment` `tags` and `artifact_location` fields. I plan to merge tomorrow. Thanks again!

mrkaye97:
awesome, thanks for all the help and doing all of that cleanup to get this over the line @dbczumar! excited to see this go live -- I think it'll improve the UX a lot 🥳

dbczumar:
> awesome, thanks for all the help and doing all of that cleanup to get this over the line @dbczumar! excited to see this go live -- I think it'll improve the UX a lot 🥳

Of course! Thanks so much for all your hard work! Merging! 🎆

amesar:
I did a deep dive into JSON schema back in 2012 and although  a great idea, there were so many issues I ditched it. Glad to see its working now! Schema-less JSON makes me shudder. It's like fingers crossed programming  ;)

Issue #4353, opened 2021-05-13
[BUG] Deadlock issue raised from backend store DB


anjsam1402:
## MLflow Roadmap Item

This is an MLflow Roadmap item that has been prioritized by the MLflow maintainers.

### System information
- **OS Platform and Distribution**: MLflow docker image based on 'python:3.9-slim-buster'. Deployed in k8s cluster.
- **MLflow installed from**: official pip \ pypi
- **MLflow version**: 1.16.0
- **MLflow backend store**: AWS RDS MySQL 8.0.23
- **MLflow artifact root**: AWS S3 bucket
- **Python version**: 3.9
- ** command** : `mlflow server --host 0.0.0.0  --port 5000 --default-artifact-root ${BUCKET}  --backend-store-uri mysql+pymysql://${USERNAME}:${PASSWORD}@${HOST}:${PORT}/${DATABASE}`

### Describe the problem
I have been performing load test on the Mlflow Tracking server with a multiprocessing sample code. For few of the runs the following issue was raised :

> mlflow.exceptions.MlflowException: (pymysql.err.OperationalError) (1213, 'Deadlock found when trying to get lock; try restarting transaction')
[SQL: INSERT INTO latest_metrics (`key`, value, timestamp, step, is_nan, run_uuid) VALUES (%(key)s, %(value)s, %(timestamp)s, %(step)s, %(is_nan)s, %(run_uuid)s)]
[parameters: {'key': 'rmse', 'value': 0.859125957974236, 'timestamp': 1620801352207, 'step': 0, 'is_nan': 0, 'run_uuid': 'eb6d1876d37d4706a97cec4db51937a6'}]

### Code to reproduce issue
```
import time
import multiprocessing
from multiprocessing import Pool
import os
import warnings
import sys
import pandas as pd
import numpy as np
from sklearn.metrics import mean_squared_error, mean_absolute_error, r2_score
from sklearn.model_selection import train_test_split
from sklearn.linear_model import ElasticNet
import mlflow
import mlflow.sklearn

mlflow.set_tracking_uri({REMOTE_TRACKING_SERVER_URI}) 
mlflow.set_experiment({EXPERIMENT_NAME})

def eval_metrics(actual, pred):
    # compute relevant metrics
    rmse = np.sqrt(mean_squared_error(actual, pred))
    mae = mean_absolute_error(actual, pred)
    r2 = r2_score(actual, pred)
    return rmse, mae, r2
    
def load_data(data_path):
[...]

PR #5460, opened 2022-03-06
#4299 log_batch actually writes in batches

### LINKED ISSUE #4299: [FR] SqlAlchemyStore.log_batch should really write in batches
## MLflow Roadmap Item

This is an MLflow Roadmap item that has been prioritized by the MLflow maintainers. We're seeking help with the implementation of roadmap items tagged with the `help wanted` label.

For requirements clarifications and implementation questions, or to request a PR review, please tag @WeichenXu123 in your communications related to this issue.

## Proposal Summary
Since MLFlow can easily get really slow, it would be great if the log_batch method really would log in batches. This method currently only iterates over all metrics, params and tags and then calls the log_metric/param/tag methods on each. In my understanding each of these calls opens a new database connection which is very slow. It seriously slows down my training loop when I log the losses in each batch of each epoch.

## Motivation
See the following snippet from `sqlalchemy_store.py`:

```python
def log_batch(self, run_id, metrics, params, tags):
    _validate_run_id(run_id)
    _validate_batch_log_data(metrics, params, tags)
    _validate_batch_log_limits(metrics, params, tags)
        run = self._get_run(run_uuid=run_id, session=session)
        self._check_run_is_active(run)
    try:
        for param in params:
            self.log_param(run_id, param)
        for metric in metrics:
            self.log_metric(run_id, metric)
        for tag in tags:
            self.set_tag(run_id, tag)
    except MlflowException as e:
        raise e
    except Exception as e:
        raise MlflowException(e, INTERNAL_ERROR)
```

- What is the use case for this feature?
Imagine you are doing a hyperparameter search (e.g. using hyperopt or optuna) and you want to log each of the trials in detail. In my example I am traning 20 models in parallel on 4 GPUs and I want to log the losses. When using the default file storage (mlruns folder), this works very well, but then the UI gets extremely slow. Therefore I am switching to a database backend which makes the UI really fast, but unfortunately the logging very slow.

- Why is this use case valuable to support for MLflow users in general?
Performance is always important and I have already seen many questions about performance issues in MLFlow.

- Why is this use case valuable to support for your project(s) or organization?
Same as above.

- Why is it currently difficult to achieve this use case? (please be as specific as possible about why related MLflow features and components are insufficient)
Logging slows down my training loop a lot. Logging should be highly optimized and the number of DB transactions should be as low as possible.

Components 

Interfaces

Languages 

Integrations

erensahin:

- Relates to https://github.com/mlflow/mlflow/pull/3024
- Fixes https://github.com/mlflow/mlflow/issues/4299 

In this PR, SqlAlchemyStore.log_batch is implemented in a way that it actually writes in batches. Three new public methods are added:

* log_params: log the given params at once. In case of any IntegrityError, rollback the session and call log_param for
each param and let it handle IntegrityError
* log_metrics: log the given metrics at once, and call _update_latest_metric_if_necessary if the metric instance is recently created.
* set_tags: set the given tags at once.

In all of them, logging logic of single param, metric, or tag is preserved.

The main intuition is to share the same session and not to open new session for each atomic operation and commit the session. To maintain the current behavior, log_params actually commits its own session to catch the integrity error and let it handled by `log_param` method, by logging all params one by one by calling `log_param`.

SqlAlchemyStore lacks docstrings and typing, so I did not add them. But I can add if it is desired. I only added it to protected `_get_or_create_many` method to clarify intention of it.

## How is this patch tested?

It passes current unit tests cases of `tests/store/tracking/test_sqlalchemy_store.py` and new test cases are added for `log_params`, `log_metrics`, and `set_tags`.

I also did a small load test:

```python
import os
from timeit import default_timer as timer

import mlflow
from mlflow.tracking import MlflowClient
from mlflow.entities import Param, RunTag, Metric


os.environ["MLFLOW_TRACKING_URI"] = "sqlite:///mock.db"

params = []
metrics = []
tags = []
N = 100 * 100
MAX_BATCH_SIZE = 100

for i in range(N):
    params.append(Param(str(i), "val"))
    metrics.append(Metric(str(i), i, 0, 0))
    tags.append(RunTag(str(i), "val"))


    start = timer()
    for i in range(0, N, MAX_BATCH_SIZE):
        MlflowClient().log_batch(
            run.info.run_id,
            metrics=metrics[i:i+MAX_BATCH_SIZE],
            params=params[i:i+MAX_BATCH_SIZE],
            tags=tags[i:i+MAX_BATCH_SIZE]
        )
    runtime = round(timer() - start, 3)
    print(f"Runtime: {runtime} seconds...")
```

Results of master branch: `Runtime: 246.888 seconds...`
Results of this branch: `Runtime: 4.769 seconds...` 

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

In this change, SqlAlchemyStore.log_batch method is modified and it actually writes in batches now. Also, new public methods are added to SqlAlchemyStore, `log_params`, `log_metrics`, and `set_tags`.

Components 

Interface 

Language 

Integrations

<a name="sqlalchemy-store-log-batch"></a>


erensahin:
@WeichenXu123 hi, you were reviewing https://github.com/mlflow/mlflow/pull/3024 therefore I am tagging you to notify :slightly_smiling_face:

WeichenXu123:
thank you for your contribution! I will review this soon.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -222,6 +223,29 @@ def _get_or_create(self, session, model, **kwargs):
 
         return instance, created
 
+    def _get_or_create_many(self, session, model, list_of_args):
```
dbczumar, line 226: Re-examining the `_get_or_create` code after a few years, I'm wondering if it's truly necessary to have this logic for point writes or for batch writes. The only place where the `created` boolean output of `_get_or_create` is used is https://github.com/mlflow/mlflow/blob/78c8dfad74dbfa06b9707054aeee44be4398bf12/mlflow/store/tracking/sqlalchemy_store.py#L591, where we check it to determine whether or not to update `latest_metrics`.

In all other cases, `_get_or_create` is duplicating the function of an integrity constraint.

Can we remove the `_get_or_create_many` semantics from this change and instead in favor of implementing the the following logic for the following procedures?

**log_params**

1. Create a session
2. Within the session, perform a `_get_run()` operation to fetch all of the params associated with the run
3. Attempt to insert the batch of parameters into the `params` table via `session.add_all()` & `session.commit()`
4. In case of an integrity error, check to see if there are any value differences between the new batch (from 3) and the existing run parameters (from 2). If there are, throw a specific MlflowException. If there aren't, return successfully in order to provide idempotency (`Params: once written, param values cannot be changed (attempting to overwrite a param value will result in an error). However, logging the same param (key, value) is permitted - that is, logging a param is idempotent.` from https://mlflow.org/docs/latest/rest-api.html#log-batch).

This is very similar to https://github.com/mlflow/mlflow/blob/38d9f0efc5d6a6d4ef2b591e95210e0dc3bc5296/mlflow/store/tracking/sqlalchemy_store.py#L757-L767, except we don't `reraise` in the case where there isn't any mismatch between the new parameter values and the existing values in the DB.

**log_metrics**

1. Create a session
2. Attempt to insert the batch of metrics into the `metrics` table via `session.add_all()` & `session.commit()`
3. Unconditionally attempt to update the `latest_metrics` table
4. If any integrity errors are encountered, raise them directly to the caller

**set_tags**

1. Create a session
2. Attempt to upsert the batch of tags into the `tags` table via `session.merge()`
3. If any integrity errors are encountered (unlikely), raise them directly to the caller

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -604,6 +632,52 @@ def log_metric(self, run_id, metric):
             if just_created:
                 self._update_latest_metric_if_necessary(logged_metric, session)
 
+    def log_metrics(self, run_id, metrics, session=None, check_run=True):
```
dbczumar, line 635: Nit: Can we make the new functions introduced by this PR "private" by prepending an underscore, since these methods aren't part of the `AbstractStore` API and don't need to be called by external consumers?

```suggestion
    def _log_metrics(self, run_id, metrics, session=None, check_run=True):
```

mlflow/store/tracking/sqlalchemy_store.py
dbczumar: After implementing these changes, I'd be curious to see whether the performance is even better, given that we'd no longer perform entity lookups prior to insertion.

dbczumar (commented):
@erensahin Thanks so much for picking this up! Awesome PR! I left a comment that should hopefully help us simplify things and, possibly, improve performance further. Let me know what you think!

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -604,6 +632,52 @@ def log_metric(self, run_id, metric):
...
+                run = self._get_run(run_uuid=run_id, session=session)
+                self._check_run_is_active(run)
+            instances = self._get_or_create_many(
+                session, SqlMetric, metric_args)
+            # keep the behavior of updating latest metric table if necessary
+            # for each created or retrieved metric
+            for (sql_metric, just_created) in instances:
+                if just_created:
+                    self._update_latest_metric_if_necessary(
+                        sql_metric, session)
```
dbczumar, line 678: Can we make this a batch operation as well? Given a list of metrics, we can query the `latest_metrics` table for the corresponding metric names to determine which metrics to update.

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Hi @dbczumar , thank you so much for such detailed comment. 

I didn't want to make radical changes and keep things behave as they do in logging one by one, but I myself think these parts should be re-designed as well. I totally agree with your comment on the necessity of `_get_or_create`.

**log_params**

Yeah, it makes sense. You expect to ignore the integrity error if the parameters are same. And even if one parameter does not match, we should rollback the changes. Did I understand correctly?

**log_metrics**

Makes sense. And I should update latest_metrics table in batch afterwards.

**set_tags**

Makes sense. I also would not expect an integrity error since we merge.

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Sure, thanks for the suggestion

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Yeah that's a great suggestion, thanks.

erensahin:
Hi @dbczumar , I think it is ready for another review round, I tried to implement your requests but there is some problems:

1. sqlalchemy does not support batch upsert operations via `session.merge()`. so I didn't implement it.
2. since we change the behavior in _log_params, now some test cases fail. The problem is that we handle the exception (and conditionally raise) for the whole batch, and during handling it, the whole session is rolled back. But the test cases (`test_log_batch_param_overwrite_disallowed_single_req`) still expect that one of the params is logged at all..

mlflow/store/tracking/sqlalchemy_store.py
erensahin: I tried to implement your requests but there is some problems:

sqlalchemy does not support batch upsert operations via session.merge(). so I didn't implement it.
since we change the behavior in _log_params, now some test cases fail. The problem is that we handle the exception (and conditionally raise) for the whole batch, and during handling it, the whole session is rolled back. But the test cases ([test_log_batch_param_overwrite_disallowed_single_req](https://github.com/mlflow/mlflow/blob/2bb754ace70fd72f99a60f6e9bca146f4dba5071/tests/store/tracking/test_sqlalchemy_store.py#L1615)) still expect that one of the params is logged at all..

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -604,6 +609,42 @@ def log_metric(self, run_id, metric):
...
+
+        # use nullcontext approach if a session is given, otherwise obtain
+        # a new session via ManagedSessionMaker
+        context = (
+            self.ManagedSessionMaker() if session is None
+            else contextlib.nullcontext(session)
+        )
+
+        with context as session:
+            self._save_to_db(session=session, objs=metric_instances)
```
dbczumar, line 641: It just occurred to me that we may run into trouble with the idempotence of LogBatch due to this use of `session.add_all()` (`_save_to_db()`) to insert a collection of new `Metric` records into the `metrics` table. If two LogBatch requests with the same metric content are issued sequentially, the second one will fail with a primary key constraint violation.

Can we implement logic similar to what we're using in `_log_params` by catching `IntegrityError` and checking the run's metric history? If the metric history contains all of the metrics being inserted, we can simply return, since we know that the `metrics` table contains the desired values and that they were reflected correctly by `latest_metrics` when they were originally inserted.

dbczumar (commented):
> Hi @dbczumar , I think it is ready for another review round, I tried to implement your requests but there is some problems:
> 
> 1. sqlalchemy does not support batch upsert operations via `session.merge()`. so I didn't implement it.
> 2. since we change the behavior in _log_params, now some test cases fail. The problem is that we handle the exception (and conditionally raise) for the whole batch, and during handling it, the whole session is rolled back. But the test cases ([test_log_batch_param_overwrite_disallowed_single_req](https://github.com/mlflow/mlflow/blob/2bb754ace70fd72f99a60f6e9bca146f4dba5071/tests/store/tracking/test_sqlalchemy_store.py#L1615)) still expect that one of the params is logged at all..

Thanks so much for trying this out and addressing feedback, @erensahin! 

1. My mistake! Apologies for missing this. I've left some suggestions for other approaches to improve performance without using `merge()`.

2. This is okay. We can remove the old conflicting test coverage. Partial writes aren't part of the documented API contract; IMO, removing undocumented partial writes is an improvement.

We should be all good to go once remaining comments are addressed. Let me know if I can provide any additional clarifications / assistance here. Thank you for all your hard work!

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -692,6 +733,61 @@ def log_param(self, run_id, param):
...
+                # using the session. In this case, we re-use the session because the SqlRun, `run`,
+                # is lazily evaluated during the invocation of `run.params`.
+                session.rollback()
+
+                # in case of an integrity error, compare the parameters of the
+                # run. If the parameters match the ones whom being saved, 
+                # ignore the exception since idempotency is reached.
+                non_matching_params = []
+                for param in param_instances:
+                    existing_value = [p.value for p in run.params if p.key == param.key]
```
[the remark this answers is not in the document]
erensahin: Yeah, that's a smart idea. thanks.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -604,6 +609,42 @@ def log_metric(self, run_id, metric):
...
+            self.ManagedSessionMaker() if session is None
+            else contextlib.nullcontext(session)
+        )
+
+        with context as session:
+            self._save_to_db(session=session, objs=metric_instances)
+            # keep the behavior of updating latest metric table if necessary
+            # for each created or retrieved metric
+            for logged_metric in metric_instances:
+                self._update_latest_metric_if_necessary(logged_metric, session)
```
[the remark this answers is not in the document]
erensahin: This looks good to me, and it is quite promising improvement as far as i can foresee right now, thanks for the suggestion :) 4-5 covers an upsert operation here. I am not sure `session.merge` handles the `WHEN NOT MATCHED BY SOURCE THEN DELETE` logic (I dont think that it does). So there's nothing requires a delete operation here, right?

4. I think 3 would return `SqlLatestMetric` objects, and I should not convert them to mlflow entity (which is `Metric`) over here to benefit from updating the attributes.
5. I think I should create `SqlLatestMetric` objects here

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -731,6 +731,95 @@ def test_log_null_metric(self):
             warnings.resetwarnings()
         assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
 
+    def test__log_metrics(self):
```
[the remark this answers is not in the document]
erensahin: I didn't modify any test cases regarding log_batch, but I'll try to add more cases for it :)

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -722,6 +818,29 @@ def set_tag(self, run_id, tag):
...
+        # a new session via ManagedSessionMaker
+        context = (
+            self.ManagedSessionMaker() if session is None
+            else contextlib.nullcontext(session)
+        )
+
+        with context as session:
+            for tag in tags:
+                _validate_tag(tag.key, tag.value)
+                session.merge(SqlTag(run_uuid=run_id, key=tag.key, value=tag.value))
```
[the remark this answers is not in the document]
erensahin: Yeah, it is quite clear. I got your point on using their own sessions in all these methods. 

For the retry case, do you have any suggestions on implementing it on sqlalchemy? Seen some implementations over Query interface, are they ok to use? (https://stackoverflow.com/questions/53287215/retry-failed-sqlalchemy-queries)

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Yeah, it is a good idea. If the metric history does not contain all of the metrics being inserted, I should insert the ones whom not present, and update latest_metrics table, is it correct?

dbczumar: Yep! Exactly :).

mlflow/store/tracking/sqlalchemy_store.py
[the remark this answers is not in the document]
dbczumar: Good question! For retries, I think we can just write some retry logic around the entire `_set_tags` body, without worrying about SqlAlchemy-specific integrations. I noticed the following response on the thread you linked:

```
hi there - SQLAlchemy author here. I've just been pointed to this recipe. I would highly recommend against using a pattern such as the above. "retry" for connectivity should be performed at the top level of a transaction only, and that's what the pool pre_ping feature is for. if you lose a connection in the middle of a transaction, you need to re-run your whole operation. explicit is better than implicit. –
```

tests/store/tracking/test_sqlalchemy_store.py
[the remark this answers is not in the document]
dbczumar: Ah, I think it would be great if we could write test cases for `log_batch()` and remove the test cases for the private methods. I.e. replace private method testing with public method testing, as long as we can get the same level of coverage.

mlflow/store/tracking/sqlalchemy_store.py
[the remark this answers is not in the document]
dbczumar: Fantastic! I agree, we never need to worry about record deletion for `metrics` / `latest_metrics`.

Good catch - apologies for the oversight. 4 and 5 should refer to `SqlLatestMetric`; I'll update the original message.

mlflow/store/tracking/sqlalchemy_store.py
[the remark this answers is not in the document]
erensahin: I tried to implement a retry mechanism, but I was unable to write a comprehensive unit test for it. I did not want to get into multiprocessing/threading. do you have a suggestion?

mlflow/store/tracking/sqlalchemy_store.py
erensahin: I think this one is done

tests/store/tracking/test_sqlalchemy_store.py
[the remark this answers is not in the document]
erensahin: I revisited all unit test cases and converted them to call log_batch instead of private methods

mlflow/store/tracking/sqlalchemy_store.py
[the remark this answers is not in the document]
erensahin: Done, I guess :)

mlflow/store/tracking/sqlalchemy_store.py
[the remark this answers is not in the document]
erensahin: Done

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -696,6 +800,63 @@ def log_param(self, run_id, param):
...
+                    run_uuid=run_id, key=param.key, value=param.value
+                )
+        param_instances = list(param_instances.values())
+
+        # use SessionManager instead of ManagedSessionManager to set
+        # isolation level for individual transaction.
+        # Here, we open a connection with SERIALIZABLE isolation level.
+        # it will be released when session.commit is called.
+        session = self.SessionMaker()
+        session.connection(execution_options={"isolation_level": "SERIALIZABLE"})
```
dbczumar, line 820: We don't need a serializable transaction for log_params and set_tags because we're only writing to a single table and we aren't performing any gap locking (e.g. `select for update`). Can we use `ManagedSessionMaker` instead?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -726,6 +887,76 @@ def set_tag(self, run_id, tag):
...
+        :param tags: List of RunTag instances to log
+        """
+        if not tags:
+            return
+
+        # use SessionManager instead of ManagedSessionManager to set
+        # isolation level for individual transaction.
+        # Here, we open a connection with SERIALIZABLE isolation level.
+        # it will be released when session.commit is called.
+        session = self.SessionMaker()
```
dbczumar, line 904: We don't need a serializable transaction for log_params and set_tags because we're only writing to a single table and we aren't performing any gap locking (e.g. `select for update`). Can we use `ManagedSessionMaker` instead?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -726,6 +887,76 @@ def set_tag(self, run_id, tag):
...
+        session.connection(execution_options={"isolation_level": "SERIALIZABLE"})
+
+        run = self._get_run(run_uuid=run_id, session=session)
+        self._check_run_is_active(run)
+
+        try:
+            current_tags = (
+                session.query(SqlTag)
+                .filter(SqlTag.run_uuid == run_id, SqlTag.key.in_([t.key for t in tags]))
+                .with_for_update()
```
dbczumar, line 914: I think we can loosen the isolation guarantees for `set_tags` because, unlike `log_metrics`, we're only modifying a single table. For concurrent requests setting multiple tags, there's no well-defined behavior. If we want to use `with_for_update`, we'd need to use a serializable transaction to avoid deadlocks due to gap locking (see https://github.com/mlflow/mlflow/issues/4353#issuecomment-859551069).

Can we remove `with_for_update()`?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -726,6 +887,76 @@ def set_tag(self, run_id, tag):
...
+                .all()
+            )
+            current_tags = {t.key: t for t in current_tags}
+
+            new_tag_dict = {}
+            for tag in tags:
+                _validate_tag(tag.key, tag.value)
```
dbczumar, line 921: Nit: Can we move tag validation to the top of the method so that we don't make any DB queries if tags are invalid?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -726,6 +887,76 @@ def set_tag(self, run_id, tag):
...
+                current_tag = current_tags.get(tag.key)
+                new_tag = new_tag_dict.get(tag.key)
+
+                # update the SqlTag if it is already present in DB
+                if current_tag:
+                    current_tag.value = tag.value
+                    continue
+
+                # if the SqlTag instance is recently created,
+                # update with incoming
```
dbczumar, line 931:
```suggestion
                # if a SqlTag instance is already present in `new_tag_dict`, this means that duplicate tags
                # were passed to `set_tags`. In this case, we resolve potential conflicts by updating the 
                # preexisting `SqlTag` instance's `value` with the value of the `new_tag`
```

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -726,6 +887,76 @@ def set_tag(self, run_id, tag):
...
+            # apply retry here.
+            max_retries = 3
+            attempt = (attempt or 0) + 1
+            if attempt > max_retries:
+                raise MlflowException(
+                    "Failed to set tags with given keys={} within {} retries".format(
+                        [t.key for t in tags], max_retries
+                    )
+                )
+            sleep_duration = 0.1 * ((2 ** attempt) - 1)
```
dbczumar, line 955: I think we should increase the sleep duration slightly.

```suggestion
            sleep_duration = ((2 ** attempt) - 1)
```

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -726,6 +887,76 @@ def set_tag(self, run_id, tag):
...
+            session.commit()
+        except sqlalchemy.exc.IntegrityError:
+            session.rollback()
+            # two concurrent operations may try to attempt to insert tags.
+            # apply retry here.
+            max_retries = 3
+            attempt = (attempt or 0) + 1
+            if attempt > max_retries:
+                raise MlflowException(
+                    "Failed to set tags with given keys={} within {} retries".format(
```
dbczumar, line 951: Nit: Can we place the list of keys at the end of the message?

```suggestion
                    "Failed to set tags within {} retries. Keys: {}".format(
```

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -726,6 +887,76 @@ def set_tag(self, run_id, tag):
...
+                    current_tag.value = tag.value
+                    continue
+
+                # if the SqlTag instance is recently created,
+                # update with incoming
+                if new_tag:
+                    new_tag.value = tag.value
+                # otherwise, put it into the dict
+                else:
+                    new_tag = SqlTag(run_uuid=run_id, key=tag.key, value=tag.value)
```
dbczumar, line 936: Instead of checking for duplicates here, could we filter out duplicate tags at the beginning of the method like we do for params here: https://github.com/mlflow/mlflow/pull/5460/files#diff-b2b10e69c9c9134afc376963051160539426069d6ed9ebd6af8252ce2b51de09R808-R812?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -696,6 +800,63 @@ def log_param(self, run_id, param):
...
+        if not params:
+            return
+        # eliminate duplicate Param entities
+        param_instances = {}
+        for param in params:
+            if param.key not in param_instances:
+                param_instances[param.key] = SqlParam(
+                    run_uuid=run_id, key=param.key, value=param.value
+                )
+        param_instances = list(param_instances.values())
```
dbczumar, line 813: I think we can replace this with 

```
params = set(params)
param_instances = [SqlParam(run_uuid=run_id, key=param.key, value=param.value)]
```

mlflow/store/tracking/sqlalchemy_store.py
dbczumar, line 931: Nice! Thanks for handling this case. Small comment suggestion:

```suggestion
                # if a SqlTag instance is already present in `new_tag_dict`, this means that multiple
                # tags with the same key were passed to `set_tags`. In this case, we resolve potential
                # conflicts by updating the value of the existing instance to the value of `tag`
```

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -621,29 +674,80 @@ def _compare_metrics(metric_a, metric_b):
                 metric_b.value,
             )
 
-        # Fetch the latest metric value corresponding to the specified run_id and metric key and
-        # lock its associated row for the remainder of the transaction in order to ensure
+        def _merge_metric(new_metric, old_metric):
```
dbczumar, line 677: Nit: I think `_overwrite_metric` is a more descriptive name here

```suggestion
        def _overwrite_metric(new_metric, old_metric):
```

dbczumar (commented):
@erensahin This is looking great! I left a few small comments, but I think we should be ready to merge once they're addressed! Thank you for all your hard work on this!

It looks like there are a few small test failures in https://github.com/mlflow/mlflow/runs/5825511155?check_suite_focus=true. Please let me know if you'd like me to help debug them.

mlflow/store/tracking/sqlalchemy_store.py
erensahin: I am not sure why I choose this way of doing this, but I think I wanted to make sure that we keep the order of params while making them distinct, since set does not preserve the order

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Thanks, I applied it

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Done

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Good idea, done

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Yeah that's a good idea, done

mlflow/store/tracking/sqlalchemy_store.py
erensahin: I wanted to make sure that the isolation level is guaranteed but if you say that we don't need it for SqlTag entity, it makes sense. I removed it.

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Done

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Done

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Oh, the order matters in tags, so it made me think so :) applied it over here

erensahin:
@dbczumar hey,

I addressed your latest reviews, especially those on undoing isolated transaction usage on _log_params, and _set_tags. I did not want to add new commits like "undo XXX", so I rebased the history a little bit to undo these. I think I addressed all of your comments.

Also, it is a bit cliche but the tests pass on my development environment :)

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Applied your suggestion, that was my other alternative :)

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -727,6 +882,76 @@ def set_tag(self, run_id, tag):
...
+        :param tags: List of RunTag instances to log
+        """
+        if not tags:
+            return
+
+        valid_tags = []
+        for tag in tags:
+            _validate_tag(tag.key, tag.value)
+            valid_tags.append(tag)
+        tags = valid_tags
```
dbczumar, line 899: Nit: I don't think we need to create a new list

```suggestion
        for tag in tags:
            _validate_tag(tag.key, tag.value)
```

dbczumar (approved):
LGTM with one tiny nit! Thanks a bunch, @erensahin! The performance improvement is spectacular!

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -1778,6 +1809,38 @@ def test_log_batch_same_metric_repeated_multiple_reqs(self):
...
+        tkey = "blahmetric2"
+        tval = None
+        metric_2 = entities.Metric(tkey, tval, int(1000 * time.time()), 0)
+
+        metrics = [metric_1, metric_2]
+
+        warnings.simplefilter("ignore")
+        with self.assertRaises(MlflowException) as exception_context, warnings.catch_warnings():
+            self.store.log_batch(run.info.run_id, metrics=metrics, params=[], tags=[])
+            warnings.resetwarnings()
```
harupy, line 1841:
```suggestion
        with self.assertRaises(MlflowException) as exception_context:
            self.store.log_batch(run.info.run_id, metrics=metrics, params=[], tags=[])
```

nit: can we remove unnecessary warning capturing?

harupy: We removed them in https://github.com/mlflow/mlflow/pull/5685

harupy:
Strange. CI got stuck.

mlflow/store/tracking/sqlalchemy_store.py
erensahin: Yeah, absolutely better one.

erensahin:
I applied the latest suggestions, let's see how it goes :)

harupy:
The `database` check is getting stuck:

https://github.com/mlflow/mlflow/runs/6003543193?check_suite_focus=true

https://user-images.githubusercontent.com/17039389/163132092-bfdc1716-dd2d-4889-b5c8-f90c6745ac2d.mov

harupy:
I was able to reproduce the issue by running:

```
./tests/db/compose.sh build --build-arg DEPENDENCIES="$(python setup.py -q dependencies)" mlflow-mssql
./tests/db/compose.sh run --rm --no-TTY mlflow-mssql pytest tests/store/tracking/test_sqlalchemy_store.py tests/db
```

harupy:
Canceled the database check.

erensahin:
Thank you @harupy 

According to the tests in the CI/CD pipeline, there's something wrong on windows side.

https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1741

https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1954

I wonder what makes sqlite give `too many SQL variables error` on windows side but not on linux. Do you have any idea?
Also, I don't have dev environment on windows, any guidance how can I debug this one?

dbczumar:
> Thank you @harupy
> 
> According to the tests in the CI/CD pipeline, there's something wrong on windows side.
> 
> https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1741
> 
> https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1954
> 
> I wonder what makes sqlite give `too many SQL variables error` on windows side but not on linux. Do you have any idea? Also, I don't have dev environment on windows, any guidance how can I debug this one?

@erensahin Thanks for pointing this out! The issue seems to be that we're binding in too many metric keys when querying for latest metrics, >= 1000 for sqlite, which is incompatible:

```
Maximum Number Of Host Parameters In A Single SQL Statement

A host [parameter](https://www.sqlite.org/lang_expr.html#varparam) is a place-holder in an SQL statement that is filled in using one of the [sqlite3_bind_XXXX()](https://www.sqlite.org/c3ref/bind_blob.html) interfaces. Many SQL programmers are familiar with using a question mark ("?") as a host parameter. SQLite also supports named host parameters prefaced by ":", "$", or "@" and numbered host parameters of the form "?123".

Each host parameter in an SQLite statement is assigned a number. The numbers normally begin with 1 and increase by one with each new parameter. However, when the "?123" form is used, the host parameter number is the number that follows the question mark.

SQLite allocates space to hold all host parameters between 1 and the largest host parameter number used. Hence, an SQL statement that contains a host parameter like ?1000000000 would require gigabytes of storage. This could easily overwhelm the resources of the host machine. To prevent excessive memory allocations, the maximum value of a host parameter number is SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.

The maximum host parameter number can be lowered at run-time using the [sqlite3_limit](https://www.sqlite.org/c3ref/limit.html)(db,[SQLITE_LIMIT_VARIABLE_NUMBER](https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber),size) interface.
```

I suggest we split the latest metrics queries into groups of 500 keys; I'll push a commit for this and see if it helps fix the Windows tests. I'll also investigate the hanging database suite, which I suspect may be indicative of deadlock.

erensahin:
> > Thank you @harupy
> > According to the tests in the CI/CD pipeline, there's something wrong on windows side.
> > https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1741
> > https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1954
> > I wonder what makes sqlite give `too many SQL variables error` on windows side but not on linux. Do you have any idea? Also, I don't have dev environment on windows, any guidance how can I debug this one?
> 
> @erensahin Thanks for pointing this out! The issue seems to be that we're binding in too many metric keys when querying for latest metrics, >= 1000 for sqlite, which is incompatible:
> 
> ```
> Maximum Number Of Host Parameters In A Single SQL Statement
> 
> A host [parameter](https://www.sqlite.org/lang_expr.html#varparam) is a place-holder in an SQL statement that is filled in using one of the [sqlite3_bind_XXXX()](https://www.sqlite.org/c3ref/bind_blob.html) interfaces. Many SQL programmers are familiar with using a question mark ("?") as a host parameter. SQLite also supports named host parameters prefaced by ":", "$", or "@" and numbered host parameters of the form "?123".
> 
> Each host parameter in an SQLite statement is assigned a number. The numbers normally begin with 1 and increase by one with each new parameter. However, when the "?123" form is used, the host parameter number is the number that follows the question mark.
> 
> SQLite allocates space to hold all host parameters between 1 and the largest host parameter number used. Hence, an SQL statement that contains a host parameter like ?1000000000 would require gigabytes of storage. This could easily overwhelm the resources of the host machine. To prevent excessive memory allocations, the maximum value of a host parameter number is SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.
> 
> The maximum host parameter number can be lowered at run-time using the [sqlite3_limit](https://www.sqlite.org/c3ref/limit.html)(db,[SQLITE_LIMIT_VARIABLE_NUMBER](https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber),size) interface.
> ```
> 
> I suggest we split the latest metrics queries into groups of 500 keys; I'll push a commit for this and see if it helps fix the Windows tests. I'll also investigate the hanging database suite, which I suspect may be indicative of deadlock.

Thanks @dbczumar,

Sqlite part is quite clear. I still don't get why it does not produce a problem on linux but windows :)
Also, MSSQL have 2100 parameter limit. I think the `MAX_METRICS_PER_BATCH = 1000`should cover it but it should be kept in mind as well

erensahin:
> > > Thank you @harupy
> > > According to the tests in the CI/CD pipeline, there's something wrong on windows side.
> > > https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1741
> > > https://github.com/mlflow/mlflow/runs/6003544271?check_suite_focus=true#step:6:1954
> > > I wonder what makes sqlite give `too many SQL variables error` on windows side but not on linux. Do you have any idea? Also, I don't have dev environment on windows, any guidance how can I debug this one?
> > 
> > 
> > @erensahin Thanks for pointing this out! The issue seems to be that we're binding in too many metric keys when querying for latest metrics, >= 1000 for sqlite, which is incompatible:
> > ```
> > Maximum Number Of Host Parameters In A Single SQL Statement
> > 
> > A host [parameter](https://www.sqlite.org/lang_expr.html#varparam) is a place-holder in an SQL statement that is filled in using one of the [sqlite3_bind_XXXX()](https://www.sqlite.org/c3ref/bind_blob.html) interfaces. Many SQL programmers are familiar with using a question mark ("?") as a host parameter. SQLite also supports named host parameters prefaced by ":", "$", or "@" and numbered host parameters of the form "?123".
> > 
> > Each host parameter in an SQLite statement is assigned a number. The numbers normally begin with 1 and increase by one with each new parameter. However, when the "?123" form is used, the host parameter number is the number that follows the question mark.
> > 
> > SQLite allocates space to hold all host parameters between 1 and the largest host parameter number used. Hence, an SQL statement that contains a host parameter like ?1000000000 would require gigabytes of storage. This could easily overwhelm the resources of the host machine. To prevent excessive memory allocations, the maximum value of a host parameter number is SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.
> > 
> > The maximum host parameter number can be lowered at run-time using the [sqlite3_limit](https://www.sqlite.org/c3ref/limit.html)(db,[SQLITE_LIMIT_VARIABLE_NUMBER](https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber),size) interface.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I suggest we split the latest metrics queries into groups of 500 keys; I'll push a commit for this and see if it helps fix the Windows tests. I'll also investigate the hanging database suite, which I suspect may be indicative of deadlock.
> 
> Thanks @dbczumar,
> 
> Sqlite part is quite clear. I still don't get why it does not produce a problem on linux but windows :) Also, MSSQL have 2100 parameter limit. I think the `MAX_METRICS_PER_BATCH = 1000`should cover it but it should be kept in mind as well

additionally, I think dividing to query to two parts for run_id and for metric keys should be good enough instead of 500 sized batches.

The example query is as below:

```
WHERE latest_metrics.run_uuid = ? AND latest_metrics."key" IN (?, ?, ? ....?)      {1000 params}
```

dbczumar:
My guess regarding Windows is that the sqlite version used for our Windows CI is older than the one used for Linux: 

```
SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.
```

In practice, 2 parts should be equivalent to batches of 500, right? Since the maximum number of keys is 1000. If that max limit was ever raised, we'd want to keep using a smaller batch size.

erensahin:
> My guess regarding Windows is that the sqlite version used for our Windows CI is older than the one used for Linux:
> 
> ```
> SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.
> ```
> 
> In practice, 2 parts should be equivalent to batches of 500, right? Since the maximum number of keys is 1000. If that max limit was ever raised, we'd want to keep using a smaller batch size.

Yes, it is the same in practice

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -589,28 +591,79 @@ def log_metric(self, run_id, metric):
...
+        session.connection(execution_options={"isolation_level": "SERIALIZABLE"})
+
+        run = self._get_run(run_uuid=run_id, session=session)
+        self._check_run_is_active(run)
+
+        # commit the session to make sure that we catch any IntegrityError
+        # and try to handle them.
+        try:
+            self._save_to_db(session=session, objs=metric_instances)
+            self._update_latest_metrics_if_necessary(metric_instances, session)
```
erensahin, line 637: I think that this might be the causing "deadlock", since we operate on metric_instances that are not committed yet? these are another tests that do fail on windows. it might be less performant but I would suggest something like below

```
try:
    self._save_to_db(session=session, objs=metric_instances)
    session.commit()
except:
    session.rollback()
    ...

self._update_latest_metrics_if_necessary(metric_instances, session)
session.commit()
```

dbczumar: I discovered an issue when running locally against sqlite where `self._update_latest_metrics_if_necessary(metric_instances, session)` is failing with a duplicate element and the error handling logic in https://github.com/mlflow/mlflow/pull/5460/files#diff-b2b10e69c9c9134afc376963051160539426069d6ed9ebd6af8252ce2b51de09R645-R664 gets stuck. Investigating further...

dbczumar:
@erensahin I just pushed a commit for the following:

1. Use a chunk size of 500 when querying latest metrics.
2. Use ManagedSessionMaker for creating sessions that use isolated transaction levels, ensuring that sessions are committed and closed properly.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -589,28 +591,76 @@ def log_metric(self, run_id, metric):
...
+                # convert to a set of Metric instance to take advantage of its hashable
+                # and then obtain the metrics that were not logged earlier within this run_id
+                metric_history = {m.to_mlflow_entity() for m in metric_history}
+                non_existing_metrics = [
+                    m for m in metric_instances if m.to_mlflow_entity() not in metric_history
+                ]
+                # if there exist metrics that were tried to be logged & rolled back even though
+                # they were not violating the PK, log them.
+                if non_existing_metrics:
+                    self._log_metrics(run_id, non_existing_metrics)
```
dbczumar, line 661: On second thought, reading up to 1000 metric histories into memory for conflict resolution could be fairly expensive from a latency / memory pressure perspective.

I'm wondering if it might be better to throw in this case instead. For certain sql dialects, we can use a "on duplicate key ignore" construct and otherwise fail open. I'll update this with some more details shortly.

Thoughts @erensahin ?

dbczumar: The more I think about it, the more I favor the current approach. The data volume concern can be addressed via further chunking

erensahin: Thank you so much for handling the problem @dbczumar :)

In my opinion, throwing an error when duplicate metric entry (key-timestamp-step-value) is given in a batch is perfectly fine without even trying to insert. It is the way how log_params treat. And I agree on "on duplicate key ignore" approach on certain dialects, but for the sake of stability raising an early error makes more sense for me.

dbczumar: Understood! I was tempted to go that direction, but I don't want to break any existing user workloads that might rely on idempotent logging of duplicate metric keys. I've just made a minor adjustment here to chunk the metrics into batches of 100 to avoid loading too much data into memory at once; I think the data volume is acceptable since this case is unlikely to occur frequently. I also refactored slightly so that we call an internal method when handling the IntegrityError, rather than retrying the whole `_log_metric` method, which would create a nested session. I've gone ahead and pushed the changes; let me know how they look! If CI passes, I think we're all set to merge! Thanks for all your hard work @erensahin !

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -727,6 +895,74 @@ def set_tag(self, run_id, tag):
...
+            return
+
+        for tag in tags:
+            _validate_tag(tag.key, tag.value)
+
+        with self.ManagedSessionMaker() as session:
+            run = self._get_run(run_uuid=run_id, session=session)
+            self._check_run_is_active(run)
+
+            def _try_insert_tags(attempt_number, max_retries=3):
```
dbczumar, line 915: @erensahin I also made a slight tweak here to call an internal method when retrying tag setting; this helps us avoid creating a new session nested within the previous session. Let me know if you have any concerns!

dbczumar (approved):
@erensahin I believe that this also addresses https://github.com/mlflow/mlflow/issues/4353 by serializing metric write operations. I've added a test case to confirm. Awesome work! I'll plan to merge tomorrow once tests pass! Thank you for your contribution!

erensahin:
> @erensahin I believe that this also addresses #4353 by serializing metric write operations. I've added a test case to confirm. Awesome work! I'll plan to merge tomorrow once tests pass! Thank you for your contribution!

thats great news! thank you for your support. Do you have any planned release time? when we can get these new changes?
I would also appreciate if you can apply these changes to Databricks Managed Mlflow as well. We are using it and we hope to see a major speed-up :)

erensahin:
I also updated the MR description, we started with 240 secs --> 25 secs improvement and we end up 240 secs --> 5 secs

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -622,29 +678,108 @@ def _compare_metrics(metric_a, metric_b):
...
+                latest_metrics_batch = (
+                    session.query(SqlLatestMetric)
+                    .filter(
+                        SqlLatestMetric.run_uuid == logged_metrics[0].run_uuid,
+                        SqlLatestMetric.key.in_(latest_metric_keys_from_db),
+                    )
+                    # Order by the metric run ID and key to ensure a consistent locking order
+                    # across transactions, reducing deadlock likelihood
+                    .order_by(SqlLatestMetric.run_uuid, SqlLatestMetric.key)
+                    .with_for_update()
```
dbczumar, line 734: @erensahin After adding a concurrency test, I found that using a serializable transaction leads to deadlock due to range locking. I've instead reverted back to the previous `with_for_update()` row locking procedure and derisked it by ensuring that we only lock preexisting rows in the database, resolving the issue described in https://github.com/mlflow/mlflow/issues/4353#issue-890814942. While it's possible that two separate transactions may simultaneously insert a brand new metric, this case is highly unlikely to occur in practice, and it wasn't handled before. The slightly more common case where two threads are logging values for a preexisting metric simultaneously is safe from a concurrency perspective with row locking.

dbczumar (commented):
@erensahin Thanks again for all your hard work! This will be released with MLflow 1.26.0, which should roll out some time in May. The Databricks-managed MLflow Tracking service implements fast, scalable, and threadsafe batch metadata logging; feel free to check it out!

Merging!

PR #5461, opened 2022-03-06
Adds input example and model signature to autologging for tf.keras for certain x types on tf.keras.Model.fit

bali0019:

- Adds autologging input example/model signature support for certain x types: ndarray, dict mapping of input names to ndarrays, tf.data.Dataset and tf.keras.utils.Sequence. 
- Extracts either ndarray  or dic(str -> ndarray) for autologging input example/model signature
- Unit tests for all x types support added
## How is this patch tested?

Unit tests

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Add example and model signature for autologging of tensorflow.keras flavor.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/tensorflow/__init__.py
+++ b/mlflow/tensorflow/__init__.py
@@ -55,11 +55,11 @@ from mlflow.utils.autologging_utils import (
     PatchFunction,
     log_fn_args_as_params,
     batch_metrics_logger,
+    get_autologging_config,
 )
 from mlflow.entities import Metric
 from mlflow.tracking._model_registry import DEFAULT_AWAIT_MAX_SLEEP_SECONDS
 from mlflow.models import infer_signature
-from mlflow.utils.autologging_utils import get_autologging_config
 
 FLAVOR_NAME = "tensorflow"
 
@@ -908,38 +908,12 @@ def autolog(
                 if log_models:
                     import numpy as np
 
-                    def _is_x_type_supported_for_autologging(input_data):
-                        is_x_supported = False
-                        if isinstance(input_data, np.ndarray):
-                            is_x_supported = True
-                        elif isinstance(input_data, pandas.Series):
-                            is_x_supported = True
-                        elif isinstance(input_data, tensorflow.data.Dataset):
-                            is_x_supported = True
-                        elif isinstance(input_data, dict):
-                            is_x_supported = True
-                            for name in input_data.keys():
-                                data_type = input_data[name]
-                                if not isinstance(data_type, np.ndarray):
-                                    is_x_supported = False
-                        elif isinstance(input_data, tensorflow.keras.utils.Sequence):
-                            if isinstance(input_data[:][0], np.ndarray):
-                                is_x_supported = True
-
-                        if not is_x_supported:
-                            warnings.warn(
-                                "Tensorflow keras autologging only "
-                                "supports one of these types: numpy array, "
-                                "dictionary of (name -> numpy.ndarray), "
-                                "tf.data dataset or tf.keras.utils.Sequence"
-                            )
-                        return is_x_supported
-
-                    def _get_input_data_slice(input_example, history):
+                    def _get_input_data_slice():
+                        input_training_data = args[0]
                         input_example_slice = None
-                        if isinstance(input_example, np.ndarray):
-                            input_example_slice = input_example[:INPUT_EXAMPLE_SAMPLE_ROWS]
-                        elif isinstance(input_example, tensorflow.data.Dataset):
+                        if isinstance(input_training_data, np.ndarray):
+                            input_example_slice = input_training_data[:INPUT_EXAMPLE_SAMPLE_ROWS]
+                        elif isinstance(input_training_data, tensorflow.data.Dataset):
                             steps = 1
                             if history.params is not None and "steps" in history.params:
                                 steps = history.params["steps"]
@@ -963,36 +937,41 @@ def autolog(
                                     ]
                                 )
 
-                            return _extract_n_steps(input_example)
-                        elif isinstance(input_example, dict):
+                            return _extract_n_steps(input_training_data)
+                        elif isinstance(input_training_data, dict):
                             input_example_slice = {
                                 k: np.take(v, range(0, INPUT_EXAMPLE_SAMPLE_ROWS))
-                                for k, v in input_example.items()
+                                for k, v in input_training_data.items()
                             }
-                        elif isinstance(input_example, tensorflow.keras.utils.Sequence):
-                            input_example_slice = input_example[:][0][:INPUT_EXAMPLE_SAMPLE_ROWS]
+                        elif isinstance(input_training_data, tensorflow.keras.utils.Sequence):
+                            input_example_slice = input_training_data[:][0][:INPUT_EXAMPLE_SAMPLE_ROWS]
+
+                        else:
+                            warnings.warn(
+                                "Tensorflow keras autologging only "
+                                "supports input types of: numpy.ndarray, "
+                                "dict(<key> -> numpy.ndarray), tensorflow.data.Dataset, "
+                                "or tensorflow.keras.utils.Sequence"
+                            )
 
                         return input_example_slice
 
-                    def _infer_model_signature(input_example):
+                    def _infer_model_signature(input_data_slice):
                         try:
-                            model_output = history.model.predict(input_example)
-                            model_signature = infer_signature(input_example, model_output)
+                            model_output = history.model.predict(input_data_slice)
+                            model_signature = infer_signature(input_data_slice, model_output)
                         except TypeError as te:
                             warnings.warn(str(te))
                             model_signature = None
                         return model_signature
 
-                    signature = None
-                    input_example = None
-                    if _is_x_type_supported_for_autologging(args[0]):
-                        input_example, signature = resolve_input_example_and_signature(
-                            lambda: _get_input_data_slice(args[0], history),
-                            _infer_model_signature,
-                            log_input_examples,
-                            log_model_signatures,
-                            _logger,
-                        )
+                    input_example, signature = resolve_input_example_and_signature(
+                        lambda: _get_input_data_slice(),
+                        _infer_model_signature,
+                        log_input_examples,
+                        log_model_signatures,
+                        _logger,
+                    )
 
                     mlflow.keras.log_model(
                         keras_model=history.model,
--- a/mlflow/tensorflow/__init__.py
+++ b/mlflow/tensorflow/__init__.py
@@ -944,7 +944,9 @@ def autolog(
                                 for k, v in input_training_data.items()
                             }
                         elif isinstance(input_training_data, tensorflow.keras.utils.Sequence):
-                            input_example_slice = input_training_data[:][0][:INPUT_EXAMPLE_SAMPLE_ROWS]
+                            input_example_slice = input_training_data[:][0][
+                                :INPUT_EXAMPLE_SAMPLE_ROWS
+                            ]
 
                         else:
                             warnings.warn(
@@ -966,7 +968,7 @@ def autolog(
                         return model_signature
 
                     input_example, signature = resolve_input_example_and_signature(
-                        lambda: _get_input_data_slice(),
+                        _get_input_data_slice,
                         _infer_model_signature,
                         log_input_examples,
                         log_model_signatures,
--- a/mlflow/tensorflow/__init__.py
+++ b/mlflow/tensorflow/__init__.py
@@ -21,7 +21,6 @@ import pandas
 from packaging.version import Version
 from threading import RLock
 
-
 import mlflow
 import mlflow.keras
 from mlflow import pyfunc
@@ -50,6 +49,8 @@ from mlflow.utils.model_utils import _get_flavor_configuration
 from mlflow.utils.autologging_utils import (
     autologging_integration,
     safe_patch,
+    INPUT_EXAMPLE_SAMPLE_ROWS,
+    resolve_input_example_and_signature,
     picklable_exception_safe_function,
     PatchFunction,
     log_fn_args_as_params,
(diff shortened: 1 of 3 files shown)


bali0019:
@BenWilson2  
Questions/Concerns:
1) Do I need to add this for the new tests?
`@pytest.mark.skipif(
    Version(tf.__version__) < Version("2.6.0"),
    reason=("TensorFlow autologging is not used for vanilla Keras models in Keras < 2.6.0"),
)`

2) I added an exception catcher step inside ` mlflow/tensorflow/_autolog.py`  because `self.model.summary(print_fn=sum_list.append)` throws an exception for the new test model i added in unit tests. Need to investigate the cause of this further.

3) I disabled the log_model that happens inside keras callback (set it to False), registered inside `mlflow/tensorflow/__init__.py` since I added the `log_model` inside fit patch function that now logs input_example/model_signature as well for certain types, but I am not 100% sure if that is the right place to add the logic since we have that callback for `on_train_end`, where i didn't add the same logic, since even an input data slice can be big in certain cases. 

4) two new unit tests for x as tf.Data.Dataset are failing due to the shape (i believe), I am looking into this, but x as ndarray, dict and Sequence are good to go.

mlflow/tensorflow/__init__.py
```diff
@@ -50,14 +49,17 @@
...
     picklable_exception_safe_function,
     PatchFunction,
     log_fn_args_as_params,
     batch_metrics_logger,
 )
 from mlflow.entities import Metric
 from mlflow.tracking._model_registry import DEFAULT_AWAIT_MAX_SLEEP_SECONDS
-
+from mlflow.models import infer_signature
+from mlflow.utils.autologging_utils import get_autologging_config
```
BenWilson2, line 62: can add this to the block lines 49:57

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,113 @@ def _patch_implementation(
...
-                        callbacks, log_models, metrics_logger
+                        callbacks, False, metrics_logger
                     )
 
                 early_stop_callback = _get_early_stop_callback(callbacks)
                 _log_early_stop_callback_params(early_stop_callback)
 
                 history = original(inst, *args, **kwargs)
 
+                if log_models:
```
BenWilson2, line 908:
```suggestion
TF_SUPPORTED_AUTOLOGGING_TYPES = (np.ndarray, 
                  pd.Series, 
                  tensorflow.data.Dataset, 
                  dict, 
                  tensorflow.keras.utils.Sequence)
def _is_supported_for_autologging(input_data):
    support = isinstance(input_data, TF_SUPPORTED_AUTOLOGGING_TYPES)
    if support:
        if isinstance(input_data, dict):
            support = all({isinstance(value, np.ndarray) for value in input_data.values()})
        if isinstance(input_data, tensorflow.keras.utils.Sequence):
            support = isinstance(input_data[:][0], np.ndarray)
    if not support:
        warnings.warn("Tensorflow keras autologging only supports input types of: "
                      "numpy.ndarray, dict(<key> -> numpy.ndarray), tensorflow.data.Dataset, "
                      "or tensorflow.keras.utils.Sequence")
    return support
```
might want to simplify this logic a bit.

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,113 @@ def _patch_implementation(
...
+                        if not is_x_supported:
+                            warnings.warn(
+                                "Tensorflow keras autologging only "
+                                "supports one of these types: numpy array, "
+                                "dictionary of (name -> numpy.ndarray), "
+                                "tf.data dataset or tf.keras.utils.Sequence"
+                            )
+                        return is_x_supported
+
+                    def _get_input_data_slice(input_example, history):
```
BenWilson2, line 938: does this shadow the higher level `history` variable? couldn't we just refer to it direclty with a history.params.get("steps", None) and then do a validate on the existence of that integer from within this private scoped function?

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,113 @@ def _patch_implementation(
...
+                                for k, v in input_example.items()
+                            }
+                        elif isinstance(input_example, tensorflow.keras.utils.Sequence):
+                            input_example_slice = input_example[:][0][:INPUT_EXAMPLE_SAMPLE_ROWS]
+
+                        return input_example_slice
+
+                    def _infer_model_signature(input_example):
+                        try:
+                            model_output = history.model.predict(input_example)
```
BenWilson2, line 979: re: above comment, here you're referring to the encapsulated variable from the parent reference directly. Should be good to do above as well.

BenWilson2:
> 3\. I disabled the log_model that happens inside keras callback (set it to False), registered inside `mlflow/tensorflow/__init__.py` since I added the `log_model` inside fit patch function that now logs input_example/model_signature as well for certain types, but I am not 100% sure if that is the right place to add the logic since we have that callback for `on_train_end`, where i didn't add the same logic, since even an input data slice can be big in certain cases.

I think that logic will have to be propagated (or moved) to ensure backwards compatibility with tf 2.6.3 and older versions (see cross version test results with "model not built yet" and other per-version specific issues where a validation of the instance __dict__() is looking for the values to be present. 
Seems to be just some minor things to address! Good job!

BenWilson2:
autoformat

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,113 @@ def _patch_implementation(
...
 
                 early_stop_callback = _get_early_stop_callback(callbacks)
                 _log_early_stop_callback_params(early_stop_callback)
 
                 history = original(inst, *args, **kwargs)
 
+                if log_models:
+                    import numpy as np
+
+                    def _is_x_type_supported_for_autologging(input_data):
```
dbczumar, line 911: Do we need this function, or can we modify the behavior of `_get_input_data_slice` to return `None` if the input data is of an unsupported / unrecognized type for input example extraction?

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,113 @@ def _patch_implementation(
...
+                        except TypeError as te:
+                            warnings.warn(str(te))
+                            model_signature = None
+                        return model_signature
+
+                    signature = None
+                    input_example = None
+                    if _is_x_type_supported_for_autologging(args[0]):
+                        input_example, signature = resolve_input_example_and_signature(
+                            lambda: _get_input_data_slice(args[0], history),
```
dbczumar, line 990: What happens if this throws? Should we try / catch and print a warning?

dbczumar: Edit: Nevermind - `resolve_input_example_and_signature` handles this.

mlflow/tensorflow/__init__.py
bali0019: done. thanks!

mlflow/tensorflow/__init__.py
bali0019: Incorporated Corey's suggestion below and removed this method. thanks.

mlflow/tensorflow/__init__.py
bali0019: good point, fixed. thanks.

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,109 @@ def _patch_implementation(
...
+                                "supports input types of: numpy.ndarray, "
+                                "dict(<key> -> numpy.ndarray), tensorflow.data.Dataset, "
+                                "or tensorflow.keras.utils.Sequence"
+                            )
+
+                        return input_example_slice
+
+                    def _infer_model_signature(input_data_slice):
+                        try:
+                            original_stop_training = history.model.stop_training
```
BenWilson2, line 963: no need to define this as a local variable. The conditional below could just use:
`if(Version(tensorflow.__version__) <= Version("2.1.4") and history.model.stop_training):`

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,109 @@ def _patch_implementation(
...
+                            model_output = history.model.predict(input_data_slice)
+
+                            if (
+                                Version(tensorflow.__version__) <= Version("2.1.4")
+                                and original_stop_training
+                            ):
+                                # For these versions, `stop_training` flag on Model is set to False
+                                # This flag is used by the callback
+                                # (inside `_log_early_stop_callback_metrics`)
```
BenWilson2, line 972: nit: double backticks for method ref [970, 972, 975]

tests/tensorflow/test_tensorflow2_autolog.py
```diff
@@ -987,3 +1063,181 @@ def test_import_keras_with_fluent_autolog_enables_tensorflow_autologging():
...
+    model = _create_model_for_dict_mapping()
+    with mlflow.start_run() as run:
+        model.fit(random_train_dict_mapping, random_one_hot_labels)
+        model_path = os.path.join(run.info.artifact_uri, "model")
+        model_conf = Model.load(os.path.join(model_path, "MLmodel"))
+        input_example = _read_example(model_conf, model_path)
+        for k, v in random_train_dict_mapping.items():
+            np.testing.assert_array_almost_equal(input_example[k], np.take(v, range(0, 5)))
+        pyfunc_model = mlflow.pyfunc.load_model(os.path.join(run.info.artifact_uri, "model"))
+        # Ensure input data slice works and is valid for the model
```
BenWilson2, line 1184: nit: remove note

tests/tensorflow/test_tensorflow2_autolog.py
```diff
@@ -49,6 +52,80 @@ def random_one_hot_labels():
...
+    )
+    return model
+
+
+@pytest.fixture
+def keras_data_gen_sequence(random_train_data, random_one_hot_labels):
+    class DataGenerator(tf.keras.utils.Sequence):
+        # def __init__(self):
+        # self.batch_size = 32
+        # self.output_shape = (6, 12)
```
BenWilson2, line 116: remove commented out lines

tests/tensorflow/test_tensorflow2_autolog.py
```diff
@@ -987,3 +1063,181 @@ def test_import_keras_with_fluent_autolog_enables_tensorflow_autologging():
...
+        assert json.loads(signature["outputs"]) == output_sig_spec
+
+
+def _assert_keras_autolog_input_example_load_and_predict_with_nparray(run, random_train_data):
+    model_path = os.path.join(run.info.artifact_uri, "model")
+    model_conf = Model.load(os.path.join(model_path, "MLmodel"))
+    input_example = _read_example(model_conf, model_path)
+    np.testing.assert_array_almost_equal(input_example, random_train_data[:5])
+    pyfunc_model = mlflow.pyfunc.load_model(os.path.join(run.info.artifact_uri, "model"))
+    # Ensure input data slice works and is valid for the model
```
BenWilson2, line 1093: nit: remove note

tests/tensorflow/test_tensorflow2_autolog.py
```diff
@@ -987,3 +1063,181 @@ def test_import_keras_with_fluent_autolog_enables_tensorflow_autologging():
...
+def test_keras_autolog_input_example_load_and_predict_with_tf_dataset(fashion_mnist_tf_dataset):
+    mlflow.tensorflow.autolog()
+    fashion_mnist_model = _create_fashion_mnist_model()
+    with mlflow.start_run() as run:
+        fashion_mnist_model.fit(fashion_mnist_tf_dataset)
+        model_path = os.path.join(run.info.artifact_uri, "model")
+        model_conf = Model.load(os.path.join(model_path, "MLmodel"))
+        input_example = _read_example(model_conf, model_path)
+        pyfunc_model = mlflow.pyfunc.load_model(os.path.join(run.info.artifact_uri, "model"))
+        # Ensure input data slice works and is valid for the model
```
BenWilson2, line 1145: nit: remove note

BenWilson2 (approved):
Great job! After a few nits are fixed LGTM!

mlflow/tensorflow/__init__.py
bali0019: done, thx.

tests/tensorflow/test_tensorflow2_autolog.py
bali0019: done, thx.

tests/tensorflow/test_tensorflow2_autolog.py
bali0019: done, thx.

tests/tensorflow/test_tensorflow2_autolog.py
bali0019: done, thx.

tests/tensorflow/test_tensorflow2_autolog.py
bali0019: done, thx.

mlflow/tensorflow/__init__.py
```diff
@@ -887,14 +897,109 @@ def _patch_implementation(
                     # & tf.keras callbacks if necessary
                     callbacks = list(kwargs.get("callbacks") or [])
                     kwargs["callbacks"], self.log_dir = _setup_callbacks(
-                        callbacks, log_models, metrics_logger
+                        callbacks, False, metrics_logger
```
dbczumar, line 900: @bali0019 @BenWilson2 Does this imply that there's some cleanup we can do within `_setup_callbacks` to remove model logging code, since we're always passing `False` now? Can we file a follow-up PR for a cleanup here?

mlflow/tensorflow/__init__.py
```diff
@@ -627,6 +629,8 @@ def autolog(
     disable_for_unsupported_versions=False,
     silent=False,
     registered_model_name=None,
+    log_input_examples=True,
```
dbczumar, line 632: @bali0019 @BenWilson2 This should default to `False` because input examples may contain sensitive data, and we want users to have to opt-in. Other autologging integrations default to `False` for this param.

dbczumar:
@bali0019 Thanks so much!

mlflow/tensorflow/__init__.py
BenWilson2: for TF <2.1.0 the fit_generator() method patch still uses this flag. Should we mark this with a remove note when we no longer support those older TF versions?

dbczumar: Does that imply that our autologging for signatures / input examples doesn't work with `fit_generator()`. If so, why not?

PR #5470, opened 2022-03-08
Restoring virtual python environment for mlflow pyfunc spark_udf

WeichenXu123:

Restoring virtual python environment for mlflow pyfunc spark_udf

## How is this patch tested?

(Details)

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mlflow/pyfunc/__init__.py
```diff
@@ -939,7 +1132,53 @@ def predict(*args):
...
+        gunicorn_bin_path = os.path.join(os.path.dirname(virtual_env_python_bin_on_executor), "gunicorn")
+        # launch scoring server
+        scoring_server_proc = subprocess.Popen(
+            [
+                gunicorn_bin_path, "-b", f"127.0.0.1:{server_port}", "-w", "1", "--timeout=500", "--",
+                "mlflow.pyfunc.scoring_server.wsgi:app"
+            ],
+            stdout=sys.stdout, stderr=sys.stderr,
+            env=server_proc_env
+        )
```
dbczumar, line 1161: Could we use `PyFuncBackend.serve()` instead? https://github.com/mlflow/mlflow/blob/5e2c868d69aad968dd6eaebf8c8f447e72366140/mlflow/pyfunc/backend.py#L67.

This way, we can support both conda and virtualenv / pyenv. We should use conda by default in MLflow 1.x. Can you work with @harupy to add pyenv / virtualenv support to `PyFuncBackend`? @harupy perhaps it would be helpful to have @WeichenXu123 implement this, since it directly relates to `spark_udf` environment restoration.

mlflow/pyfunc/__init__.py
```diff
@@ -939,7 +1132,53 @@ def predict(*args):
...
-    udf = pandas_udf(predict, result_type)
+    def predict(iterator):
+        import requests
+        from mlflow.pyfunc.scoring_server.client import ScoringServerClient
+
+        # Note: this is a pandas udf function in iteration style, which takes an iterator of
+        # tuple of pandas.Series and outputs an iterator of pandas.Series.
+        if not use_nfs and not spark_in_local_mode:
+            model_on_executor, local_model_path_on_executor = SparkModelCache.get_or_load(model_archive_path)
+            virtual_env_python_bin_on_executor = _setup_restored_python_env(model_py_version, req_file_path)
```
dbczumar, line 1143: We should make sure to support conda as well.

mlflow/pyfunc/__init__.py
```diff
@@ -760,6 +902,40 @@ def _warn_potentially_incompatible_py_version_if_necessary(model_py_version=None
...
+    except:
+        pass
+
+    try:
+        os.symlink(src, dst)
+    except:
+        pass
 
 
 def spark_udf(spark, model_uri, result_type="double"):
```
dbczumar, line 941: We should introduce an argument (e.g. `env=...`) for controlling the environment that is used for UDF inference

1. `local` (default)
2. `conda`
3. `virtualenv`

For MLflow 1.x, `local` should remain the default. For MLflow 2.0, we can make `virtualenv` the default.

mlflow/pyfunc/__init__.py
```diff
@@ -656,12 +675,154 @@ def _warn_dependency_requirement_mismatches(model_path):
...
+            _VIRTUAL_ENV_DIR = tempfile.mkdtemp()
+            atexit.register(shutil.rmtree, _VIRTUAL_ENV_DIR, ignore_errors=True)
+
+    return _VIRTUAL_ENV_DIR
+
+
+_PICKLE_PROTOCOL_FOR_RESTORE_PY_ENV = 3
+
+
+def _setup_restored_python_env(py_version, requirements_file_path):
```
dbczumar, line 747: Following from https://github.com/mlflow/mlflow/pull/5470/files#r822326583, we should introduce this to `PyFuncBackend` so that it can be used for `mlflow models serve` and Spark UDF.

mlflow/pyfunc/__init__.py
```diff
@@ -656,12 +675,154 @@ def _warn_dependency_requirement_mismatches(model_path):
...
+        else:
+            import atexit
+            import shutil
+            _VIRTUAL_ENV_DIR = tempfile.mkdtemp()
+            atexit.register(shutil.rmtree, _VIRTUAL_ENV_DIR, ignore_errors=True)
+
+    return _VIRTUAL_ENV_DIR
+
+
+_PICKLE_PROTOCOL_FOR_RESTORE_PY_ENV = 3
```
dbczumar, line 744: What is this used for, and why 3 instead of the default protocol?

mlflow/pyfunc/__init__.py
```diff
@@ -656,12 +675,154 @@ def _warn_dependency_requirement_mismatches(model_path):
...
+
+    virtual_env_python_bin = os.path.join(
+        virtual_env_dir, 'bin/python'
+    )
+    return virtual_env_python_bin
+
+
+def _load_model_from_local_path(
+    local_path: str, suppress_warnings: bool = False
+):
```
dbczumar, line 805: Do we need to make these changes? If the path already exists on the local FS, `download_artifact_from_uri` just returns it without doing any copying.

mlflow/utils/__init__.py
```diff
@@ -172,3 +174,40 @@ def _inspect_original_var_name(var, fallback_name):
...
+def write_data_to_stream(data_in_bytes, stream):
+    write_long_to_stream(len(data_in_bytes), stream)
+    stream.write(data_in_bytes)
+
+
+def get_file_sha_hexdigest(file_path):
+    with open(file_path, 'rb') as f:
+        file_data = f.read()
+
+    return hashlib.sha1(file_data).hexdigest()
```
dbczumar, line 204: `file_utils.py` is a good location for this

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,62 @@
+import requests
+import time
+from mlflow.pyfunc import scoring_server
+
+
+class ScoringServerClient:
```
dbczumar, line 6: Love this! Awesome!

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,62 @@
...
+    def ping(self):
+        ping_status = requests.get(url=self.url_prefix + "/ping")
+        print("connection attempt", i, "server is up! ping status", ping_status)
+        if ping_status.status_code != 200:
+            raise Exception(f"ping failed (error code {ping_status.status_code})")
+
+    def wait_server_ready(self, timeout=30):
+        begin_time = time.time()
+        while True:
+            time.sleep(0.3)
```
dbczumar, line 20: Perhaps pinging with exponential backoff would be good

mlflow/pyfunc/__init__.py
WeichenXu123: In my test , No. `_download_artifact_from_uri` always creates a temporary directory and copy files into the directory.

mlflow/pyfunc/__init__.py
WeichenXu123: Useless. will remove it

mlflow/pyfunc/__init__.py
WeichenXu123: I considered this, I don't like do setup restored python env inside "PyFuncBackend.serve()", I hope in databricks we can apply the optimization: in driver side directly in the NFS directory setup restored python env (by virtualenv/conda) , and then all executors can use the restored python env. The optimization improve performance very much.

For reusing code, we can do:
* provide set-up virtualenv/conda env helper function
* in serve we can call the helper function above to setup env (if specified argument)
* in spark_udf and in databricks, we call the helper function above to setup env in Driver side in NFS directory. then in `PyFuncBackend.serve()` we skip setting up restored env.

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,62 @@
...
+                content_type == scoring_server.CONTENT_TYPE_JSON
+                or content_type == scoring_server.CONTENT_TYPE_JSON_SPLIT_ORIENTED
+            ):
+                data = data.to_json(orient="split")
+            elif content_type == scoring_server.CONTENT_TYPE_CSV:
+                data = data.to_csv(index=False)
+            else:
+                raise Exception(
+                    "Unexpected content type for Pandas dataframe input %s" % content_type
+                )
```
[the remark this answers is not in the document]
WeichenXu123: The server also support CSV type input but the `_get_jsonable_obj` does not support it.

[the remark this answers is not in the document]
dbczumar: We're the ones starting the server though, so we don't need to worry about setting it up to support CSV inputs. CSV and JSON are identical to us, since we're the ones converting the DataFrame to string. No need to duplicate logic here.

mlflow/pyfunc/__init__.py
dbczumar: I don't think so.

I tried running the following code from my `/private/tmp` directory:

```
import mlflow

with mlflow.start_run():
    mlflow.log_dict({"a": "b"}, "mydict.json")
    rid = mlflow.active_run().info.run_id
    print(rid)


from mlflow.tracking.artifact_utils import _download_artifact_from_uri

print(_download_artifact_from_uri("runs:/" + rid + "/mydict.json"))
print(_download_artifact_from_uri("/private/tmp/mlruns/0/" + rid + "/artifacts/mydict.json"))
```

```
(base2) corey.zumar@C02F17H3MD6R /private/tmp $ python test.py
f27285ff6207401bb2f5b597c75c60ec
/private/tmp/mlruns/0/f27285ff6207401bb2f5b597c75c60ec/artifacts/mydict.json
/private/tmp/mlruns/0/f27285ff6207401bb2f5b597c75c60ec/artifacts/mydict.json
```

The original file path is being returned.

mlflow/pyfunc/__init__.py
dbczumar: Can't we set up the environment using NFS and then call `PyFuncBackend.serve()` after? `PyFuncBackend.serve()` should then see that the environment is already there. We use hashes of the conda environment file to accomplish this for `mlflow models serve`: if a conda environment already exists with a matching hash, we just activate it. We can implement the same thing for virtualenv / pyenv.

mlflow/pyfunc/__init__.py
WeichenXu123: got it . my fault. You are right.

mlflow/pyfunc/__init__.py
WeichenXu123: > We use hashes of the conda environment file to accomplish this for mlflow models serve: if a conda environment already exists with a matching hash, we just activate it

good idea. let me try it.

PR #5476, opened 2022-03-09
Fixed EvaluationResult not saving or loading artifacts with proper file names

### LINKED ISSUE #5475: [BUG] EvaluationResult from DefaultEvaluator throws ValueError (unknown file extension) when calling .save()
### Willingness to contribute

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: Added one additional line.
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS Big Sur
- **MLflow version (run ``mlflow --version``)**: 1.24.0
- **Python version**: 3.9.7
- **Exact command to reproduce**: Add `result.save("output")` to the end of [`evaluate_on_regressor.py`](https://github.com/mlflow/mlflow/blob/master/examples%2Fevaluation%2Fevaluate_on_regressor.py). Then `python mlflow/examples/evaluation/evaluate_on_regressor.py`

### Describe the problem
This error occurs because `ImageEvaluationArtifact` relies on [pillow's save() method](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save). It infers the image format based on the provided path's file extension. However, the provided path to [`EvaluationResult.save`](https://github.com/mlflow/mlflow/blob/78df312065d73cbd117767157a72cc68ae176d35/mlflow/models/evaluation/base.py#L135) by `DefaultEvaluator` from [here](https://github.com/mlflow/mlflow/blob/4c38389e15e44b8ac2bc00ab89d7b288d2f1e9da/mlflow/models/evaluation/default_evaluator.py#L381) consists of only the artifact's name (without extension).

### Code to reproduce issue
```python
import mlflow
from sklearn.datasets import fetch_california_housing

mlflow.sklearn.autolog()

california_housing_data = fetch_california_housing()

X_train, X_test, y_train, y_test = train_test_split(
    california_housing_data.data, california_housing_data.target, test_size=0.33, random_state=42
)

    model = LinearRegression().fit(X_train, y_train)
    model_uri = mlflow.get_artifact_uri("model")

    result = mlflow.evaluate(
        model_uri,
        X_test,
        targets=y_test,
        model_type="regressor",
        dataset_name="california_housing",
        evaluators="default",
        feature_names=california_housing_data.feature_names,
        evaluator_config={"explainability_nsamples": 1000},
    )
result.save("output")
```

### Other info / logs
```console
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/usr/local/anaconda3/lib/python3.9/site-packages/PIL/Image.py in save(self, fp, format, **params)
   2219             try:
-> 2220                 format = EXTENSION[ext]
   2221             except KeyError as e:

KeyError: ''

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
/var/folders/5d/lcq9fgm918l8mg8vlbcq4d0c0000gp/T/ipykernel_77385/2290762558.py in <module>
     26         evaluator_config={"explainability_nsamples": 1000},
     27     )
---> 28 result.save("output")

/usr/local/anaconda3/lib/python3.9/site-packages/mlflow/models/evaluation/base.py in save(self, path)
    133 
    134         for artifact_name, artifact in self.artifacts.items():
--> 135             artifact._save(os.path.join(artifacts_dir, artifact_name))
    136 
    137     @property

/usr/local/anaconda3/lib/python3.9/site-packages/mlflow/models/evaluation/artifacts.py in _save(self, output_artifact_path)
      6 class ImageEvaluationArtifact(EvaluationArtifact):
      7     def _save(self, output_artifact_path):
----> 8         self._content.save(output_artifact_path)
      9 
     10     def _load_content_from_file(self, local_artifact_path):

/usr/local/anaconda3/lib/python3.9/site-packages/PIL/Image.py in save(self, fp, format, **params)
   2220                 format = EXTENSION[ext]
   2221             except KeyError as e:
-> 2222                 raise ValueError(f"unknown file extension: {ext}") from e
   2223 
   2224         if format.upper() not in SAVE:

ValueError: unknown file extension: 
```

Components 

Interface 

Language 

Integrations

MarkYHZhang:
Signed-off-by: Mark Zhang <markzhang.inbox@gmail.com>

Fixes https://github.com/mlflow/mlflow/issues/5475

Updated `EvaluationResult` to now use artifacts' uri to extract the proper file name to be saved locally. Whereas, before it was using the name of the artifacts directly without file extensions.

Also added tests in `test_evaluation` to test the behavior reported in https://github.com/mlflow/mlflow/issues/5475 with an image artifact.

## How is this patch tested?

Updated unit tests to reflect the addition of file extensions

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -16,7 +16,7 @@ import struct
 import sys
 import math
 import urllib
-import posixpath
+import pathlib
 from collections import OrderedDict
 from abc import ABCMeta, abstractmethod
 
@@ -109,7 +109,7 @@ class EvaluationResult:
             uri = meta["uri"]
             ArtifactCls = _get_class_from_string(meta["class_name"])
             artifact = ArtifactCls(uri=uri)
-            filename = posixpath.basename(urllib.parse.urlparse(uri).path)
+            filename = pathlib.Path(urllib.parse.urlparse(uri).path).name
             artifact._load(os.path.join(artifacts_dir, filename))
             artifacts[artifact_name] = artifact
 
@@ -135,7 +135,7 @@ class EvaluationResult:
         os.mkdir(artifacts_dir)
 
         for artifact in self.artifacts.values():
-            filename = posixpath.basename(urllib.parse.urlparse(artifact.uri).path)
+            filename = pathlib.Path(urllib.parse.urlparse(artifact.uri).path).name
             artifact._save(os.path.join(artifacts_dir, filename))
 
     @property
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -15,6 +15,7 @@ import logging
 import struct
 import sys
 import math
+import pathlib
 from collections import OrderedDict
 from abc import ABCMeta, abstractmethod
 
@@ -107,7 +108,7 @@ class EvaluationResult:
             uri = meta["uri"]
             ArtifactCls = _get_class_from_string(meta["class_name"])
             artifact = ArtifactCls(uri=uri)
-            artifact._load(os.path.join(artifacts_dir, artifact_name))
+            artifact._load(os.path.join(artifacts_dir, pathlib.Path(uri).name))
             artifacts[artifact_name] = artifact
 
         return EvaluationResult(metrics=metrics, artifacts=artifacts)
@@ -131,8 +132,8 @@ class EvaluationResult:
         artifacts_dir = os.path.join(path, "artifacts")
         os.mkdir(artifacts_dir)
 
-        for artifact_name, artifact in self.artifacts.items():
-            artifact._save(os.path.join(artifacts_dir, artifact_name))
+        for artifact in self.artifacts.values():
+            artifact._save(os.path.join(artifacts_dir, pathlib.Path(artifact.uri).name))
 
     @property
     def metrics(self) -> Dict[str, Any]:
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -15,7 +15,8 @@ import logging
 import struct
 import sys
 import math
-import pathlib
+import urllib
+import posixpath
 from collections import OrderedDict
 from abc import ABCMeta, abstractmethod
 
@@ -108,7 +109,8 @@ class EvaluationResult:
             uri = meta["uri"]
             ArtifactCls = _get_class_from_string(meta["class_name"])
             artifact = ArtifactCls(uri=uri)
-            artifact._load(os.path.join(artifacts_dir, pathlib.Path(uri).name))
+            filename = posixpath.basename(urllib.parse.urlparse(uri).path)
+            artifact._load(os.path.join(artifacts_dir, filename))
             artifacts[artifact_name] = artifact
 
         return EvaluationResult(metrics=metrics, artifacts=artifacts)
@@ -133,7 +135,8 @@ class EvaluationResult:
         os.mkdir(artifacts_dir)
 
         for artifact in self.artifacts.values():
-            artifact._save(os.path.join(artifacts_dir, pathlib.Path(artifact.uri).name))
+            filename = posixpath.basename(urllib.parse.urlparse(artifact.uri).path)
+            artifact._save(os.path.join(artifacts_dir, filename))
 
     @property
     def metrics(self) -> Dict[str, Any]:
--- a/tests/models/test_evaluation.py
+++ b/tests/models/test_evaluation.py
@@ -7,6 +7,7 @@ from mlflow.models.evaluation import (
     ModelEvaluator,
     EvaluationArtifact,
 )
+from mlflow.models.evaluation.artifacts import ImageEvaluationArtifact
 from mlflow.models.evaluation.base import (
     EvaluationDataset,
     _normalize_evaluators_and_evaluator_config_args as _normalize_config,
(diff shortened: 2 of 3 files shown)


mlflow/models/evaluation/artifacts.py
```diff
@@ -4,6 +4,8 @@
 
 
 class ImageEvaluationArtifact(EvaluationArtifact):
+    ext = "png"
```
dbczumar, line 7: Can we make this an `@abstract` `@property` method on `EvaluationArtifact`?

MarkYHZhang: But I think there are cases where I want to access the extension without having an instance of the artifact.

dbczumar: @staticmethod ?

MarkYHZhang: Ohh I see, yeah that works. I was trying to stick with `ext` being a property rather than a function.

dbczumar: `@property` makes `ext` appear to be a property. See https://www.programiz.com/python-programming/property for more info.

MarkYHZhang: Yes, I am aware of that. This is why I mentioned that I want to access it without having an instance. Because `@property` can only be accessed through an instance. So we have to pick between `@staticmethod` or `@property`.

dbczumar: @MarkYHZhang Ah, I missed that. Apologies. How about https://stackoverflow.com/a/7864317/11952869?

MarkYHZhang: IMO I think this is a bit overly complicated and as the comment mentioned, not a great use of singleton design pattern. I think we should either stick with what I have right now, or change it to the static method way. Opinions?

apurva-koti: Personally I'm not a fan of the singleton pattern in general, and I also agree it seems like overkill for this use case.
Ultimately we just want to associate an EvaluationArtifact with its associated file extension with an easy-to-access class property. @staticmethod seems best to me.

MarkYHZhang: Updated the code with `@staticmethod` and `@abstractmethod`. I also added `raise NotImplementedError` inside the abstract method. Otherwise, no exception will be thrown when child class provides no implementations to `file_extension()` when called without class instance. Concretely:

```python
from abc import abstractmethod, ABCMeta

class Parent(metaclass=ABCMeta):
    def __init__(self):
        pass
    
    @staticmethod
    @abstractmethod
    def file_extension():
        pass

class Child(Parent):
    def __init__(self):
        pass

Child().file_extension() # will throw TypeError: Can't instantiate abstract class Child with abstract methods ext
Child.file_extension() # will not throw anything, will simply return None.
```

mlflow/models/evaluation/base.py
```diff
@@ -31,6 +31,14 @@ def __init__(self, uri, content=None):
         self._uri = uri
         self._content = content
 
+    @staticmethod
+    @abstractmethod
+    def file_extension():
+        """File extension for this artifact type (e.g. "png", "csv", etc.)"""
+        raise NotImplementedError(
+            "File extension must be provided for implementations of EvaluationArtifact"
```
jinzhang21, line 39: Though it's generally a good idea to have file extension in file name, I'm not fully convinced all artifacts have to come with an extension. If I want to store some byte string in base64 encoded format, it could be literally anything, e.g. a protobuf, a PNG, a serialized JSON etc.

mlflow/models/evaluation/artifacts.py
```diff
@@ -4,6 +4,10 @@
 
 
 class ImageEvaluationArtifact(EvaluationArtifact):
+    @staticmethod
+    def file_extension():
+        return "png"
```
jinzhang21, line 9: Why does an image must be in PNG format?

mlflow/models/evaluation/base.py
MarkYHZhang: I guess my original reasoning was that since every artifact corresponds to a file, it should have a file extension. For example, in the case of a byte string in base64 artifact, its file extension will be whatever file type it is being stored as. Like .proto, .png, .json. Does that make sense?

dbczumar (approved):
LGTM! Thanks @MarkYHZhang !

mlflow/models/evaluation/base.py
```diff
@@ -107,7 +109,8 @@ def load(cls, path):
             uri = meta["uri"]
             ArtifactCls = _get_class_from_string(meta["class_name"])
             artifact = ArtifactCls(uri=uri)
-            artifact._load(os.path.join(artifacts_dir, artifact_name))
+            filename = posixpath.basename(urllib.parse.urlparse(uri).path)
```
jinzhang21, line 112: The OSS team agreed to gradually transition the use of path access to PathLib. It handles paths, especially Windows paths better. Does it work here? https://docs.python.org/3/library/pathlib.html

jinzhang21 (approved):

mlflow/models/evaluation/base.py
MarkYHZhang: Sounds good!

PR #5487, opened 2022-03-14
Restoring virtual python environment for mlflow pyfunc spark_udf

WeichenXu123:

Restoring virtual python environment for `mlflow pyfunc spark_udf`. Add an `env_manager` argument:
```
The environment manager to use in order to create the
software environment for model inference. Default value is ``local``,
The following values are supported:
- ``conda``: (Recommended) Use Conda to restore the software environment
  that was used to train the model. Note that environment is only restored
  in the context of the PySpark UDF; the software environment outside of
  the UDF is unaffected.
- ``local``: Use the current Python environment for model inference, which
  may differ from the environment used to train the model and may lead to
  errors or invalid predictions.
```

This is a new PR which takes over https://github.com/mlflow/mlflow/pull/5470

## How is this patch tested?

* Test on spark local-cluster mode, unit tests added.
* Test on spark cluster mode. https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/478391771932729/command/478391771932784

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mlflow/pyfunc/__init__.py
```diff
@@ -941,6 +972,58 @@ def predict(*args):
...
+        scoring_server_proc = None
+        if env_type == "conda":
+            server_port = find_free_port()
+
+            # launch scoring server
+            # TODO: set timeout for server requests handler.
+            scoring_server_proc = start_proc(
+                [
+                    "mlflow", "models", "serve", "-m", local_model_path,
+                    "-p", str(server_port), "-w", "1", "--install-mlflow", "true",
```
harupy, line 997:
```suggestion
                    "-p", str(server_port), "-w", "1", "--install-mlflow",
```

tests/pyfunc/test_spark.py
```diff
@@ -77,7 +77,8 @@ def model_path(tmpdir):
 
 
 @pytest.mark.large
-def test_spark_udf(spark, model_path):
+@pytest.mark.parametrize("env_manager", ["local", "conda"])
```
harupy, line 80: Why do we need to parametrize every single test with `env_manager`?

mlflow/pyfunc/__init__.py
```diff
@@ -764,7 +766,28 @@ def _warn_potentially_incompatible_py_version_if_necessary(model_py_version=None
         )
 
 
-def spark_udf(spark, model_uri, result_type="double"):
+def _get_or_create_model_cache_dir():
+    from mlflow.utils.nfs_on_spark import get_nfs_cache_root_dir
+    import tempfile
```
harupy, line 771: Can we import packages at the top if we don't need to lazily import them?

tests/pyfunc/test_spark.py
WeichenXu123: I add a new test for restoring conda env case, and keep old tests unchanged.

mlflow/pyfunc/__init__.py
harupy: Ok, can we move built-in package?

tests/pyfunc/test_scoring_server.py
```diff
@@ -613,3 +614,29 @@ def test_get_cmd(args: dict, expected: str):
...
+
+    port = find_free_port()
+
+    try:
+        server_proc = start_server(port, model_path, host='127.0.0.1', no_conda=True)
+
+        client = ScoringServerClient(host='127.0.0.1', port=port)
+        client.wait_server_ready()
+
+        assert client.get_module_version("sklearn") == sklearn.__version__
```
harupy, line 635: Curious why we need to check the module version.

tests/pyfunc/test_spark.py
```diff
@@ -21,6 +22,11 @@
 import tests
 from mlflow.types import Schema, ColSpec
 
+import sklearn.datasets as datasets
+from collections import namedtuple, OrderedDict
+import sklearn.neighbors as knn
```
harupy, line 27:
```suggestion
import sklearn.neighbors import KNeighborsClassifier
```

`sklearn.neighbors` is not k-nearest neighbors.

tests/pyfunc/test_spark.py
```diff
@@ -124,6 +143,33 @@ def test_spark_udf(spark, model_path):
...
+    )
+    infer_data = pd.DataFrame(sklearn_model.inference_data, columns=['a', 'b'])
+
+    infer_spark_df = spark.createDataFrame(infer_data).repartition(1)
+
+    check_dict = {"sklearn": sklearn_log_version}
+    new_env = {
+        **os.environ.copy(),
+        'MLFLOW_SPARK_UDF_RESTORED_ENV_MODULE_VERSION_CHECK_DICT': json.dumps(check_dict)
+    }
```
harupy, line 163: `monkeypatch.setenv` is useful to temporarily set an environment variable:
https://docs.pytest.org/en/6.2.x/monkeypatch.html#monkeypatching-environment-variables

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,143 @@
...
+        status = requests.get(url=self.url_prefix + f"/version/{module_name}")
+        if status.status_code != 200:
+            raise Exception(f"get_module_version failed (error code {status.status_code})")
+        return status.text.strip()
+
+    def invoke(self, data, pandas_orient="records"):
+        import pandas as pd
+
+        content_type_list = []
+        if type(data) == pd.DataFrame:
```
harupy, line 48:
```suggestion
        if isintance(data, pd.DataFrame):
```

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,143 @@
...
+def start_server(
+        server_port, local_model_path,
+        host='127.0.0.1', num_workers=1,
+        no_conda=False,
+        env=None, stdout=sys.stdout, stderr=sys.stderr,
+):
+    cmd = [
+        "mlflow", "models", "serve", "-m", local_model_path,
+        "-h", host,
+        "-p", str(server_port), "-w", str(num_workers), "--install-mlflow",
```
harupy, line 100: Can we remove `--install-mlflow` or not?

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,143 @@
...
+                    "Unexpected pandas_orient for Pandas dataframe input %s" % pandas_orient
+                )
+        else:
+            raise RuntimeError("Unsupported data type.")
+
+        post_data = json.dumps(scoring_server._get_jsonable_obj(data, pandas_orient=pandas_orient))
+
+        response = requests.post(
+            url=self.url_prefix + "/invocations",
+            data=post_data,
```
harupy, line 65:
```suggestion
        post_data = scoring_server._get_jsonable_obj(data, pandas_orient=pandas_orient)

        response = requests.post(
            url=self.url_prefix + "/invocations",
            data=post_data,
```

Can we directly pass a dict to `requests.post`

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -188,6 +189,21 @@ def predictions_to_json(raw_predictions, output):
     json.dump(predictions, output, cls=NumpyEncoder)
 
 
+def load_predictions_from_json_str(json_str):
+    raw_predictions = json.loads(json_str)  # TODO: Implement numpy decoder and set cls to be numpy decoder ?
```
harupy, line 193: Why do we need to implemente numpy?

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -227,6 +243,11 @@ def ping():  # pylint: disable=unused-variable
         status = 200 if health else 404
         return flask.Response(response="\n", status=status, mimetype="application/json")
 
+    @app.route("/version/<module>", methods=["GET"])
+    def get_module_version(module):
+        ver = importlib.import_module(module).__version__
+        return flask.Response(response=ver, status=200, mimetype="text/plain")
```
harupy, line 249: Another way to get a module version is to create a custom model whose predict method returns the version of a specified module.

mlflow/pyfunc/__init__.py
```diff
@@ -941,7 +984,99 @@ def predict(*args):
...
+            client = ScoringServerClient("127.0.0.1", server_port)
+            client.wait_server_ready(timeout=30)
+
+            # If env variable "MLFLOW_SPARK_UDF_RESTORED_ENV_MODULE_VERSION_CHECK_DICT" is set,
+            # in spark UDF task, check the restored env modules has the same version with the
+            # provided version in the given dict.
+            # This is for testing purpose.
+            if restored_env_module_version_check_dict is not None:
+                for module, expected_version in restored_env_module_version_check_dict.items():
+                    assert client.get_module_version(module) == expected_version
```
harupy, line 1054: Is this for testing purposes?

mlflow/pyfunc/scoring_server/client.py
harupy: Or are we using `type(data) == pd.DataFrame` on purpose?

mlflow/pyfunc/scoring_server/client.py
WeichenXu123: No. We need install the current PR version mlflow in the created environment. Because I add some code in the scoring server module.

The conda yaml includes a mlflow dependency which results in installing the latest released version mlflow.

mlflow/pyfunc/scoring_server/client.py
WeichenXu123: No. In my test it failed.

mlflow/pyfunc/scoring_server/__init__.py
WeichenXu123: Because in `predictions_to_json` function, it does `json.dump(predictions, output, cls=NumpyEncoder)` which using `NumpyEncoder`, so when decoding we need a corresponding decoder ? But I am not very sure whether we need it.

mlflow/pyfunc/__init__.py
WeichenXu123: Yes.

mlflow/pyfunc/scoring_server/client.py
harupy: Ah, I thought `post_data` is a dictionary but it's not, right?

mlflow/pyfunc/scoring_server/__init__.py
WeichenXu123: > create a custom model

in order to load custom model in the "created python env", we need load the module which defines the custom model class. For databricks/universe integration test, this is hard. So this helper method helps tests in databricks/universe side.

mlflow/pyfunc/scoring_server/client.py
WeichenXu123: no. changing to isintance is fine

mlflow/pyfunc/scoring_server/__init__.py
harupy: We don't because we don't need to convert a list to a numpy array.

mlflow/pyfunc/scoring_server/client.py
harupy: This seems related to the content type.

mlflow/pyfunc/scoring_server/client.py
harupy: > We need install the current PR version mlflow in the created environment.

But users don't need to, right?

tests/pyfunc/test_scoring_server.py
WeichenXu123: Just for testing `get_module_version` helper method provided by the client class.

mlflow/pyfunc/scoring_server/client.py
WeichenXu123: Oh yes user don't need to. let me update code

tests/pyfunc/test_spark.py
```diff
@@ -76,6 +83,35 @@ def model_path(tmpdir):
...
+    y = iris.target
+    knn_model = KNeighborsClassifier()
+    knn_model.fit(X, y)
+    return ModelWithData(model=knn_model, inference_data=X)
+
+
+def _log_sklearn_model_and_save_preditions(model_path, predictions_output_path):
+    """
+    This method is used in `test_spark_udf_with_conda_env_restored`.
+    """
```
harupy, line 101: Is this a method?

dbczumar (commented):
@WeichenXu123 Awesome stuff! Left a few comments.

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,197 @@
...
+        If a user cancels the PySpark job, the worker process gets killed, regardless of
+        PySpark daemon and worker reuse settings.
+        We use prctl to ensure the command process receives SIGTERM after spark job cancellation.
+        The command process itself should handle SIGTERM properly.
+        This is a no-op on macOS because prctl is not supported.
+
+        Note: we cannot use `atexit` registering "kill_server" handler to replace `prctl` because:
+        The functions registered via "atexit" are not called when the program is killed
+        by a signal not handled by Python, when a Python fatal internal error is detected, or
+        when os._exit().
```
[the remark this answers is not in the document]
WeichenXu123: When a pyspark job canceled, the UDF python process are killed by signal "SIGKILL",
This case neither "atexit" nor signal handler can capture SIGKILL signal.
prctl is the only way to capture SIGKILL signal.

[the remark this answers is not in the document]
WeichenXu123: For MacOS system, I haven't find approach to handle SIGKILL.
For Windows system, there're some discussion here, https://stackoverflow.com/questions/53208/how-do-i-automatically-destroy-child-processes-in-windows

tests/pyfunc/test_spark.py
WeichenXu123: removed.

tests/pyfunc/test_scoring_server.py
WeichenXu123: removed.

tests/pyfunc/test_spark.py
```diff
@@ -124,6 +160,60 @@ def test_spark_udf(spark, model_path):
...
+    conda_path = get_conda_bin_executable("conda")
+    tmp_preds_output_path = os.path.join(tmpdir.strpath, "preds.json")
+    log_sklearn_model_commands = [
+        f"conda env create -f {tmp_conda_file_path} --force",
+        f"source {os.path.dirname(conda_path)}/../etc/profile.d/conda.sh",
+        "conda activate test_spark_udf_log_sklearn_model_env",
+        f"python -c \""
+        "from tests.pyfunc.test_spark import _log_sklearn_model_and_save_preditions;"
+        f"_log_sklearn_model_and_save_preditions('{model_path}', '{tmp_preds_output_path}')\""
+    ]
```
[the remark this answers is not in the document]
WeichenXu123: removed. use python_model instead.

mlflow/models/flavor_backend.py
```diff
@@ -35,7 +35,7 @@ def predict(self, model_uri, input_path, output_path, content_type, json_format)
         pass
 
     @abstractmethod
-    def serve(self, model_uri, port, host, enable_mlserver):
+    def serve(self, model_uri, port, host, enable_mlserver, blocking=True):
```
dbczumar, line 38:
```suggestion
    def serve(self, model_uri, port, host, enable_mlserver, synchronous=True):
```

mlflow/pyfunc/__init__.py
```diff
@@ -837,6 +862,11 @@ def spark_udf(spark, model_uri, result_type="double"):
 
         - ``ArrayType(StringType)``: All columns converted to ``string``.
 
+    :param env_manager: Available values "local", "conda", "virtualenv" (default "local").
```
dbczumar, line 865: Can we exclude `virtualenv` for now until we implement support for it?

mlflow/pyfunc/__init__.py
```diff
@@ -866,18 +910,38 @@ def spark_udf(spark, model_uri, result_type="double"):
...
         _warn_dependency_requirement_mismatches(local_model_path)
+    else:
+        if not sys.platform.startswith("linux"):
+            _logger.warning(
+                "In order to run inference code in restored python environment, pyspark UDF "
+                "processes spawns mlflow model server as child processes to run inference, but "
+                "due to system limitation, if spark job being canceled, these UDF processes "
+                "are killed by SIGKILL and the mlflow model server child processes cannot be "
+                "killed."
+            )
```
dbczumar, line 928:
```suggestion
                "In order to run inference code in restored python environment, PySpark UDF "
                "processes spawn MLflow Model servers as child processes. Due to system limitations, "
                " with handling SIGKILL signals, these MLflow Model server child processes cannot be " 
                " cleaned up if the Spark Job is canceled."
            )
```

mlflow/pyfunc/__init__.py
```diff
@@ -941,7 +1005,91 @@ def predict(*args):
...
+        #      help saving disk space.
+        #   2. We have conda package cache dir and pip cache dir which are shared across all
+        #      python processes which help reducing downloading time.
+        #   3. Avoid race conditions related issues.
+        #
+        # TODO:
+        #   For NFS available case, set conda env dir / virtualenv env dir in sub-directory under
+        #   NFS directory, and in spark driver side prepare restored env once, and then all
+        #   spark UDF tasks running on spark workers can skip re-creating the restored env.
+        if env_manager == "conda":
```
dbczumar, line 1035: Nit: Can we break these into separate functions?

mlflow/pyfunc/__init__.py
```diff
@@ -941,7 +1005,91 @@ def predict(*args):
...
+            ).serve(
+                model_uri=local_model_path_on_executor,
+                port=server_port,
+                host='127.0.0.1',
+                enable_mlserver=False,
+                blocking=False
+            )
+
+            client = ScoringServerClient("127.0.0.1", server_port)
+            client.wait_server_ready(timeout=30)
```
dbczumar, line 1064: Is 30 seconds enough? I think we should wait for at least 90 seconds in case the model is large and takes a long time to load.

mlflow/pyfunc/backend.py
```diff
@@ -137,7 +188,9 @@ def copy_model_into_container(dockerfile_context_dir):
         )
 
 
-def _execute_in_conda_env(conda_env_path, command, install_mlflow, command_env=None):
+def _execute_in_conda_env(
+        conda_env_path, command, install_mlflow, command_env=None, blocking=True, preexec_fn=None
```
dbczumar, line 192:
```suggestion
        conda_env_path, command, install_mlflow, command_env=None, synchronous=True, preexec_fn=None
```

mlflow/pyfunc/scoring_server/__init__.py
dbczumar: @WeichenXu123 Can we remove this `get_module_version()` method?

mlflow/pyfunc/scoring_server/client.py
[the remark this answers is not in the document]
dbczumar: Thanks!

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,76 @@
...
+                raise Exception(
+                    "Unexpected pandas_orient for Pandas dataframe input %s" % pandas_orient
+                )
+        post_data = json.dumps(scoring_server._get_jsonable_obj(data, pandas_orient=pandas_orient))
+
+        response = requests.post(
+            url=self.url_prefix + "/invocations",
+            data=post_data,
+            headers={"Content-Type": "; ".join(content_type_list)},
+        )
```
dbczumar, line 58: Leaving this here again. It was marked `resolved` but doesn't seem resolved.

For general-purpose model scoring, `data` doesn't have to be a Pandas DataFrame. It could also be a list or a numpy array. Because this class isn't specific to Spark UDF, I think we should make this a generic interface that can accept any pyfunc-compatible input.

`scoring_server._get_jsonable_obj` already handles the conversion of pyfunc-compatible inputs to JSON. All we need to do is specify the `CONTENT_TYPE_JSON` content type. For an example, see https://github.com/mlflow/mlflow/blob/9781af9c0898827bf616a8f159168477a69036dd/mlflow/sagemaker/__init__.py#L2519-L2523.

mlflow/rfunc/backend.py
```diff
@@ -37,7 +37,7 @@ def predict(self, model_uri, input_path, output_path, content_type, json_format)
         )
         _execute(command)
 
-    def serve(self, model_uri, port, host, enable_mlserver):
+    def serve(self, model_uri, port, host, enable_mlserver, blocking=True):
```
dbczumar, line 40:
```suggestion
    def serve(self, model_uri, port, host, enable_mlserver, synchronous=True):
```

mlflow/utils/process.py
```diff
@@ -1,5 +1,7 @@
 import os
 import subprocess
+import signal
+import sys
```
dbczumar, line 4: Do we need these imports?

mlflow/pyfunc/scoring_server/client.py
dbczumar: We probably don't need the `pandas_orient` argument here, since the server can handle both and the user likely doesn't care which format is used for the request.

tests/pyfunc/test_spark.py
```diff
@@ -124,6 +145,61 @@ def test_spark_udf(spark, model_path):
...
+
+        def predict(self, context, model_input):
+            from packaging.version import Version
+            import sklearn
+            if Version(sklearn.__version__) < Version('0.23'):
+                self.is_low_version = True
+            else:
+                self.is_low_version = False
+
+            return model_input.apply(lambda row: (1.0 if self.is_low_version else 2.0), axis=1)
```
dbczumar, line 163: Can we make this more accurate by returning 1.0 if the version matches the expected version exactly and 0 if it doesn't?

mlflow/pyfunc/__init__.py
WeichenXu123: Shall we need to split them ? local / conda mode share many parts of code.

WeichenXu123: We can refactor it in better way, I think we can consider doing that after we add vitualenv support ?

dbczumar: Sounds good!

mlflow/pyfunc/__init__.py
```diff
@@ -866,18 +913,42 @@ def spark_udf(spark, model_uri, result_type="double"):
...
+        artifact_uri=model_uri, output_path=_get_or_create_model_cache_dir()
+    )
+
+    if env_manager == "local":
         # Assume spark executor python environment is the same with spark driver side.
         _warn_dependency_requirement_mismatches(local_model_path)
+        _logger.warning(
+            "You can set `mlflow.pyfunc.spark_udf` function argument `env_manager` to "
+            "be 'conda' to run inference in restored python environment."
+        )
```
dbczumar, line 926:
```suggestion
        _logger.warning(
            "Calling `spark_udf()` with `env_manager="local"` does not recreate the same environment that was used"
            " during training, which may lead to errors or inaccurate predictions. We recommend specifying"
            " `env_manager="conda"`, which automatically recreates the environment that was used to train the"
            " model and performs inference in the recreated environment."
        )
```

mlflow/pyfunc/__init__.py
```diff
@@ -837,6 +862,11 @@ def spark_udf(spark, model_uri, result_type="double"):
 
         - ``ArrayType(StringType)``: All columns converted to ``string``.
 
+    :param env_manager: Available values "local", "conda" (default "local").
+        Specify the environment manage used to restore the model required python environment.
+        If set "local", then use current python environment.
+        If set "conda", then use conda to restore model required python environment.
```
dbczumar, line 868:
```suggestion
    :param env_manager: Specifies the environment manager that is used to create the
                        software environment for model inference. The following values are supported:
                        
                        - ``conda``: (Recommended) Use Conda to restore the software environment that was used to
                                               train the model. Note that environment is only restored in the context of the
                                               PySpark UDF; the software environment outside of the UDF is unaffected.
                                               
                        - ``local``: Use the current Python environment for model inference, which may differ from
                                            the environment used to train the model and may lead to errors or invalid predictions.
```

mlflow/pyfunc/__init__.py
```diff
@@ -845,10 +875,27 @@ def spark_udf(spark, model_uri, result_type="double"):
...
     from pyspark.sql.types import ArrayType, DataType as SparkDataType
     from pyspark.sql.types import DoubleType, IntegerType, FloatType, LongType, StringType
+    from pyspark.sql.functions import PandasUDFType
+    from mlflow.models.cli import _get_flavor_backend
+
+    # importing here to prevent circular import
+    from mlflow.pyfunc.scoring_server.client import prepare_env
+
+    if env_manager not in ["local", "conda"]:
+        raise ValueError("Illegal env_manager setting.")
```
dbczumar, line 891: Can we raise `MlflowException` with error code `INVALID_PARAMETER_VALUE`?

mlflow/pyfunc/__init__.py
```diff
@@ -866,18 +913,42 @@ def spark_udf(spark, model_uri, result_type="double"):
...
+    )
+
+    if env_manager == "local":
         # Assume spark executor python environment is the same with spark driver side.
         _warn_dependency_requirement_mismatches(local_model_path)
+        _logger.warning(
+            "You can set `mlflow.pyfunc.spark_udf` function argument `env_manager` to "
+            "be 'conda' to run inference in restored python environment."
+        )
+    else:
```
dbczumar, line 927: If `env_manager` is `conda`, can we also print a `_logger.info()` statement indicating that Conda will be used to recreate the software environment for inference and that this may take extra time?

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def invoke(self, data):
+        """
+        Invoke inference on input data. The input data must be pandas dataframe or json instance.
+        """
+
+        import pandas as pd
+
+        if isinstance(data, pd.DataFrame):
+            content_type = scoring_server.CONTENT_TYPE_JSON_RECORDS_ORIENTED
```
dbczumar, line 43: Can we use the `split` format by default, since the records format might reorder prediction columns (dictionaries don't have a well-defined ordering)?

dbczumar: Also filed https://github.com/mlflow/mlflow/pull/5522 to fix this for the sagemaker deployment plugin's `predict()` method. Can you review?

dbczumar: The `application/json` content type should be fine for all Pandas DataFrame orientations because the scoring server uses the data type of the payload (dict or list) to determine whether to parse the Pandas DataFrame from records or split format. See https://github.com/mlflow/mlflow/blob/01182994ec31966dbe8528603d9e1a0ed2188234/mlflow/pyfunc/scoring_server/__init__.py#L93-L105.

Including this special case for added safety seems fine though :)

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -0,0 +1,71 @@
...
+
+def prepare_env(local_model_path, stdout=sys.stdout, stderr=sys.stderr):
+    cmd = [
+        "mlflow",
+        "models",
+        "prepare-env",
+        "-m",
+        local_model_path,
+    ]
+    return subprocess.run(cmd, stdout=stdout, stderr=stderr, universal_newlines=True, check=True)
```
dbczumar, line 71: This failed when I tried it on Databricks with the following stacktrace:

```
UnsupportedOperation                      Traceback (most recent call last)
<[command-478391771928007]()> in <module>
----> 1 udf = mlflow.pyfunc.spark_udf(spark, logged_model.model_uri, env_manager="conda")

/local_disk0/.ephemeral_nfs/envs/pythonEnv-bec7f63d-3c4f-4a99-94e4-ee6b73674b94/lib/python3.9/site-packages/mlflow/pyfunc/__init__.py in spark_udf(spark, model_uri, result_type, env_manager)
    944         # Prepare restored environment in driver side if possible.
    945         if env_manager == "conda":
--> 946             prepare_env(local_model_path)
    947 
    948     model_metadata = Model.load(os.path.join(local_model_path, MLMODEL_FILE_NAME))

/local_disk0/.ephemeral_nfs/envs/pythonEnv-bec7f63d-3c4f-4a99-94e4-ee6b73674b94/lib/python3.9/site-packages/mlflow/pyfunc/scoring_server/client.py in prepare_env(local_model_path, stdout, stderr)
     69         local_model_path,
     70     ]
---> 71     return subprocess.run(cmd, stdout=stdout, stderr=stderr, universal_newlines=True, check=True)

/usr/lib/python3.9/subprocess.py in run(input, capture_output, timeout, check, *popenargs, **kwargs)
    503         kwargs['stderr'] = PIPE
    504 
--> 505     with Popen(*popenargs, **kwargs) as process:
    506         try:
    507             stdout, stderr = process.communicate(input, timeout=timeout)

/usr/lib/python3.9/subprocess.py in __init__(self, args, bufsize, executable, stdin, stdout, stderr, preexec_fn, close_fds, shell, cwd, env, universal_newlines, startupinfo, creationflags, restore_signals, start_new_session, pass_fds, user, group, extra_groups, encoding, errors, text, umask)
    827         (p2cread, p2cwrite,
    828          c2pread, c2pwrite,
--> 829          errread, errwrite) = self._get_handles(stdin, stdout, stderr)
    830 
    831         # We wrap OS handles *before* launching the child, otherwise a

/usr/lib/python3.9/subprocess.py in _get_handles(self, stdin, stdout, stderr)
   1596             else:
   1597                 # Assuming file-like object
-> 1598                 c2pwrite = stdout.fileno()
   1599 
   1600             if stderr is None:

/databricks/python/lib/python3.9/site-packages/ipykernel/iostream.py in fileno(self)
    306             return self._original_stdstream_copy
    307         else:
--> 308             raise io.UnsupportedOperation("fileno")
    309 
    310     def _watch_pipe_fd(self):

UnsupportedOperation: fileno
```

Do we need to run this in a subprocess with `check=True`, or can we call `PyFuncBackend.prepare_env()` directly?

mlflow/pyfunc/scoring_server/__init__.py
dbczumar: @WeichenXu123 This method is still here. Can we remove it?

mlflow/pyfunc/scoring_server/client.py
WeichenXu123: Good catch!

Starting from python 3.6, the builtin dict preserve key order, but version lower than py3.6 it might cause column order changed.

WeichenXu123: I saw "predictions_to_json" also use  pandas_orient="records", we need to also change it to split orient

WeichenXu123: Change predictions_to_json format will change the returned value format, it is a breaking change, let's discuss it in follow-up task.
Most cases prediction result is one column data so it might be a minor issue.

mlflow/pyfunc/scoring_server/client.py
WeichenXu123: Replaced with:
```
            _get_flavor_backend(local_model_path, no_conda=False, install_mlflow=False).prepare_env(
                model_uri=local_model_path
            )
```

mlflow/pyfunc/__init__.py
```diff
@@ -941,7 +1029,91 @@ def predict(*args):
         else:
             return result[result.columns[0]]
 
-    udf = pandas_udf(predict, result_type)
+    result_type_hint = pd.DataFrame if isinstance(result_type, SparkStructType) else pd.Series
+
+    @pandas_udf(result_type)
+    def udf(
+        iterator: Iterator[Tuple[Union[pd.Series, pd.DataFrame], ...]]
+    ) -> Iterator[result_type_hint]:
```
WeichenXu123, line 1037: CC @HyukjinKwon

The `pandas_udf` "functionType" argument will be deprecated and removed soon.
So defining pandas iterator style UDF we need define type hint like this.

mlflow/pyfunc/scoring_server/__init__.py
WeichenXu123: removed!

WeichenXu123:
My testing notebook:
https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/478391771932729/command/478391771932780

One issue:

'Conda env create' command might hit some race conditions if running in parallel.
in spark UDF multiple UDF tasks might running in parallel in one executor.
Each spark UDF task will run "conda env create" command,
So UDF task might raise error like:
"InvalidArchiveError("Error with archive /databricks/conda/pkgs/setuptools-60.10.0-py38h578d9bd_0.tar.bz2. "
I will address this in follow-up PRs.
For current testing, I set partition nubmer to be 1 (so that only one UDF task running) to avoid it.

To address this race condition issue, in follow-up PRs I will do:
* For databricks user, use NFS so the env only created once in driver side.
* For apache/spark user, we need create a file lock when calling "conda env create" in UDF task side.

tests/pyfunc/test_spark.py
```diff
@@ -273,3 +363,56 @@ def get_model(_):
...
     results2 = spark.sparkContext.parallelize(range(100), 30).map(get_model).collect()
     assert min(results2) > 0
+
+
+@pytest.mark.skipif(
+    not sys.platform.startswith("linux"),
+    reason="Only Linux system support setting  parent process death signal via prctl lib.",
+)
+@pytest.mark.large
+def test_spark_udf_embedded_model_server_killed_when_job_canceled(spark, sklearn_model, model_path):
```
WeichenXu123, line 373: Note: a new added test, only available on linux system.

dbczumar:
> My testing notebook: https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/478391771932729/command/478391771932780
> 
> One issue:
> 
> 'Conda env create' command might hit some race conditions if running in parallel. in spark UDF multiple UDF tasks might running in parallel in one executor. Each spark UDF task will run "conda env create" command, So UDF task might raise error like:
> 
> > InvalidArchiveError("Error with archive /databricks/conda/pkgs/setuptools-60.10.0-py38h578d9bd_0.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message from libarchive was:\n\nFailed to create dir 'lib'")
> 
> See [conda/conda#8870](https://github.com/conda/conda/issues/8870)
> 
> I will address this in follow-up PRs. For current testing, I set partition nubmer to be 1 (so that only one UDF task running) to avoid it.
> 
> To address this race condition issue, in follow-up PRs I will do:
> 
> * For databricks user, use NFS so the env only created once in driver side.
> * For apache/spark user, we need create a file lock when calling "conda env create" in UDF task side.

Awesome, thanks Weichen! How difficult is it to implement the file lock? Will we have it done before the next MLflow release, which will occur on or before 4/11 US? If not, perhaps we should make sure that this feature isn't included in the 1.25.0 release branch.

mlflow/pyfunc/__init__.py
```diff
@@ -866,18 +924,44 @@ def spark_udf(spark, model_uri, result_type="double"):
...
+            'Calling `spark_udf()` with `env_manager="local"` does not recreate the same '
+            "environment that was used during training, which may lead to errors or inaccurate "
+            'predictions. We recommend specifying `env_manager="conda"`, which automatically '
+            "recreates the environment that was used to train the model and performs inference "
+            "in the recreated environment."
+        )
+    else:
+        _logger.info(
+            "Conda will be used to recreate the software environment for inference so that this "
+            "may take extra time."
```
dbczumar, line 944:
```suggestion
            "Conda will be used to recreate the model's software environment for inference. This "
            "may take extra time."
```

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -20,6 +20,10 @@
 import pandas as pd
 import sys
 import traceback
+import importlib
+import subprocess
+import warnings
+import signal
```
dbczumar, line 26: We can remove these unused imports

mlflow/pyfunc/backend.py
```diff
@@ -137,9 +195,14 @@ def copy_model_into_container(dockerfile_context_dir):
...
 
 
-def _execute_in_conda_env(conda_env_path, command, install_mlflow, command_env=None):
+def _execute_in_conda_env(
+    conda_env_path, command, install_mlflow, command_env=None, synchronous=True, preexec_fn=None
+):
     if command_env is None:
         command_env = os.environ
+
+    command_env['PIP_NO_INPUT'] = '1'
```
WeichenXu123, line 204: In databricks,
the driver side we set env variable `PIP_NO_INPUT` to be 1, but in executor side it does not set the env variable.
To keep consistency, I set it here.

mlflow/pyfunc/__init__.py
```diff
@@ -964,11 +1150,11 @@ def udf_with_default_cols(*args):
                         error_code=INVALID_PARAMETER_VALUE,
                     )
             else:
-                _logger.warning(
+                raise MlflowException(
                     "Attempting to apply udf on zero columns because no column names were "
-                    "specified as arguments or inferred from the model signature."
+                    "specified as arguments or inferred from the model signature.",
+                    error_code=INVALID_PARAMETER_VALUE
                 )
```
WeichenXu123, line 1157: Note: Raise error here instead of log warning.

UDF on zero columns make no sense.
and in this PR we have to use "pandas iteration UDF" which requires at least one column input.

mlflow/models/flavor_backend.py
```diff
@@ -35,14 +35,19 @@ def predict(self, model_uri, input_path, output_path, content_type, json_format)
...
         Serve the specified MLflow model locally.
 
         :param model_uri: URI pointing to the MLflow model to be used for scoring.
         :param port: Port to use for the model deployment.
         :param host: Host to use for the model deployment. Defaults to ``localhost``.
         :param enable_mlserver: Whether to use MLServer or the local scoring server.
+        :param synchronous: Whether to wait until server exit or return the server process
+                            `Popen` instance immediately.
+        :param stdout: Redirect srever stdout
+        :param stderr: Redirect srever stderr
```
dbczumar, line 50:
```suggestion
        :param stdout: Redirect server stdout
        :param stderr: Redirect server stderr
```

mlflow/pyfunc/__init__.py
dbczumar: Thank you!

mlflow/pyfunc/__init__.py
```diff
@@ -941,7 +1040,127 @@ def predict(*args):
...
+            )
+            server_redirect_log_thread.setDaemon(True)
+            server_redirect_log_thread.start()
+
+            client = ScoringServerClient("127.0.0.1", server_port)
+
+            try:
+                client.wait_server_ready(timeout=90, scoring_server_proc=scoring_server_proc)
+            except Exception:
+                err_msg = f"During spark UDF task execution, mlflow model server launching failed. "
```
dbczumar, line 1127: Nit:

```suggestion
                err_msg = f"During Spark UDF task execution, the MLflow model server failed to launch. "
```

mlflow/pyfunc/__init__.py
```diff
@@ -941,7 +1040,127 @@ def predict(*args):
...
+                if len(server_tail_logs) == _MLFLOW_SERVER_OUTPUT_TAIL_LINES_TO_KEEP:
+                    err_msg += f"Launching mlflow server command last {_MLFLOW_SERVER_OUTPUT_TAIL_LINES_TO_KEEP} lines output are:\n"
```
dbczumar, line 1129:
```suggestion
                    err_msg += f"Last {_MLFLOW_SERVER_OUTPUT_TAIL_LINES_TO_KEEP} lines of MLflow model server output:\n"
```

mlflow/pyfunc/__init__.py
```diff
@@ -941,7 +1040,127 @@ def predict(*args):
...
+                else:
+                    err_msg += "Launching mlflow server command output are:\n"
```
dbczumar, line 1131:
```suggestion
                    err_msg += "MLflow model server output:\n"
```

tests/pyfunc/test_spark.py
```diff
@@ -273,3 +396,56 @@ def get_model(_):
...
+
+    job_thread = threading.Thread(target=run_job)
+    job_thread.start()
+
+    client = ScoringServerClient("127.0.0.1", server_port)
+    client.wait_server_ready(timeout=20)
+    spark.sparkContext.cancelAllJobs()
+    job_thread.join()
+
+    time.sleep(5)  # waiting server to exit and release the port.
```
dbczumar, line 447: Is this always enough time to wait to avoid flakiness?

mlflow/pyfunc/backend.py
```diff
@@ -137,9 +198,15 @@ def copy_model_into_container(dockerfile_context_dir):
         )
 
 
-def _execute_in_conda_env(conda_env_path, command, install_mlflow, command_env=None):
+def _execute_in_conda_env(
+    conda_env_path, command, install_mlflow, command_env=None, synchronous=True, preexec_fn=None,
+    stdout=None, stderr=None,
+):
```
dbczumar, line 204: Can we add docstrings for the params here?

tests/pyfunc/test_spark.py
```diff
@@ -247,21 +363,28 @@ def test_model_cache(spark, model_path):
...
+    check_get_or_load_return_value(local_model, local_model_path)
+
     # Request the model on all executors, and see how many times we got cache hits.
     def get_model(_):
-        model = SparkModelCache.get_or_load(archive_path)
-        assert isinstance(model, PyFuncModel)
-        # NB: Can not use instanceof test as remote does not know about ConstantPyfuncWrapper class.
-        assert type(model._model_impl).__name__ == constant_model_name
+        _model, _local_model_path = SparkModelCache.get_or_load(archive_path)
+        check_get_or_load_return_value(_model, _local_model_path)
```
dbczumar, line 387:
```suggestion
        executor_model, executor_model_path = SparkModelCache.get_or_load(archive_path)
        check_get_or_load_return_value(executor_model, executor_model_path)
```

tests/pyfunc/test_spark.py
```diff
@@ -247,21 +363,28 @@ def test_model_cache(spark, model_path):
...
-    # Ensure we can use the model locally.
-    local_model = SparkModelCache.get_or_load(archive_path)
-    assert isinstance(local_model, PyFuncModel)
-    assert isinstance(local_model._model_impl, ConstantPyfuncWrapper)
-
     # Define the model class name as a string so that each Spark executor can reference it
     # without attempting to resolve ConstantPyfuncWrapper, which is only available on the driver.
     constant_model_name = ConstantPyfuncWrapper.__name__
 
+    def check_get_or_load_return_value(_model, _path):
```
dbczumar, line 370:
```suggestion
    def check_get_or_load_return_value(model_from_cache, model_path_from_cache):
```

dbczumar (approved):
@WeichenXu123 LGTM once a few nits are addressed. Awesome work!

mlflow/pyfunc/backend.py
WeichenXu123: It force PIP to ignore console input, (otherwise pip might prompt "yes or no" and ask stdin input)

dbczumar: Nice! Let's add a small inline comment explaining this

tests/pyfunc/test_spark.py
WeichenXu123: increase to 10s. should be long enough. When killed, the server exits quickly.

Issue #4235, opened 2021-04-09
[BUG] mlflow.autolog() for PyTorch Lightning logs `_step` metrics for epochs instead of steps


danieltremblay:
### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [x] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [ ] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**: Reproducible with example from [official docs](https://www.mlflow.org/docs/latest/python_api/mlflow.pytorch.html)
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Docker image: `nvidia/cuda:10.2-base-ubuntu16.04`
- **MLflow installed from (source or binary)**: Python package
- **MLflow version (run ``mlflow --version``)**: 1.15.0
- **Python version**: 3.8.8
- **npm version, if running the dev UI**: No
- **Exact command to reproduce**: See code snippet

### Describe the problem
Expected behavior:
If I'm using `mlflow.autolog()` with a `pl.LightningModule` and use `self.log("train_loss", loss, on_epoch=True, on_step=True, logger=True)` inside its `training_step` function, MLflow should log `train_loss_step` metric for every logged **step** and `train_loss_epoch` for every logged **epoch**

Actual behavior:
`train_loss_step` and `train_loss_epoch` are recorded for every logged **epoch** only.

After running code in below sample, `train_loss_epoch`  contains 1 entry per epoch:
```
1617996945494 1.3506861925125122 0
1617996966448 1.201533317565918 1
1617996986735 1.1901590824127197 2
1617997007487 1.190608024597168 3
1617997028088 1.1838353872299194 4
1617997048012 1.1807078123092651 5
1617997068702 1.18239426612854 6
1617997089547 1.1766492128372192 7
1617997110542 1.175412654876709 8
1617997130747 1.1763886213302612 9
1617997150977 1.1739486455917358 10
1617997171487 1.1722205877304077 11
1617997191321 1.170091152191162 12
1617997211502 1.1701279878616333 13
1617997232425 1.1694077253341675 14
1617997252996 1.1664137840270996 15
1617997273641 1.1662592887878418 16
1617997294056 1.1691796779632568 17
[...]

PR #5497, opened 2022-03-16
Support logging PyTorch metrics on training batch steps

adamreeve:
Signed-off-by: Adam Reeve <adreeve@gmail.com>

Add a new `log_every_n_step` parameter to `mlflow.pytorch.autolog` that enables logging step based metrics using the step count instead of epoch. By default, this parameter is `False` and these metrics will be logged using epoch numbers to keep backwards compatibility.

This addresses #4235. @danieltremblay, do you have any thoughts on this approach given you created this issue?

## How is this patch tested?

New unit test added.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

`mlflow.pytorch.autolog` now accepts a `log_every_n_step` parameter that can be used to log step based metrics against the step number.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pytorch/__init__.py
+++ b/mlflow/pytorch/__init__.py
@@ -876,6 +876,7 @@ def load_state_dict(state_dict_uri, **kwargs):
 @autologging_integration(FLAVOR_NAME)
 def autolog(
     log_every_n_epoch=1,
+    log_every_n_step=False,
     log_models=True,
     disable=False,
     exclusive=False,
@@ -905,6 +906,9 @@ def autolog(
 
     :param log_every_n_epoch: If specified, logs metrics once every `n` epochs. By default, metrics
                        are logged after every epoch.
+    :param log_every_n_step: If specified, logs batch metrics once every `n` global step. By default,
+                       metrics are not logged for steps. Note that setting this to 1 can cause
+                       performance issues and is not recommended.
     :param log_models: If ``True``, trained models are logged as MLflow model artifacts.
                        If ``False``, trained models are not logged.
     :param disable: If ``True``, disables the PyTorch Lightning autologging integration.
--- a/mlflow/pytorch/__init__.py
+++ b/mlflow/pytorch/__init__.py
@@ -876,7 +876,7 @@ def load_state_dict(state_dict_uri, **kwargs):
 @autologging_integration(FLAVOR_NAME)
 def autolog(
     log_every_n_epoch=1,
-    log_every_n_step=False,
+    log_every_n_step=None,
     log_models=True,
     disable=False,
     exclusive=False,
@@ -906,9 +906,9 @@ def autolog(
 
     :param log_every_n_epoch: If specified, logs metrics once every `n` epochs. By default, metrics
                        are logged after every epoch.
-    :param log_every_n_step: If specified, logs batch metrics once every `n` global step. By default,
-                       metrics are not logged for steps. Note that setting this to 1 can cause
-                       performance issues and is not recommended.
+    :param log_every_n_step: If specified, logs batch metrics once every `n` global step.
+                       By default, metrics are not logged for steps. Note that setting this to 1 can
+                       cause performance issues and is not recommended.
     :param log_models: If ``True``, trained models are logged as MLflow model artifacts.
                        If ``False``, trained models are not logged.
     :param disable: If ``True``, disables the PyTorch Lightning autologging integration.
--- a/tests/pytorch/test_pytorch_autolog.py
+++ b/tests/pytorch/test_pytorch_autolog.py
@@ -43,6 +43,19 @@ def pytorch_model_without_validation():
     return trainer, run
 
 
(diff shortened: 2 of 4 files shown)


mlflow/pytorch/__init__.py
```diff
@@ -876,6 +876,7 @@ def load_state_dict(state_dict_uri, **kwargs):
 @autologging_integration(FLAVOR_NAME)
 def autolog(
     log_every_n_epoch=1,
+    log_every_n_step=False,
```
harupy, line 879: @adamreeve Thanks for the PR! Curious why the default value is False. Can we set it to `None`?

adamreeve: Not sure why I went with `False` to be honest, `None` is probably more conventional, I'll change it.

harupy:
@adamreeve If it's challenging to enable this feature in older versions of pytorch-lightning (PL), we can conditionally enable it depending the version of pytorch-lightning.

adamreeve:
OK yes I hadn't tested with older versions, I'll see if I can get this working in all the supported versions.

adamreeve:
I've now updated this to work with PyTorch-Lightning versions >= 1.1.0, I couldn't see an obvious way to support versions earlier than that.

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -73,9 +87,19 @@ def test_pytorch_autolog_logs_expected_data(pytorch_model):
...
     client = mlflow.tracking.MlflowClient()
-    for metric_key in ["loss", "train_acc", "val_loss", "val_acc"]:
+    for metric_key in [
+        "loss",
+        "train_acc",
+        "val_loss",
+        "val_acc",
+        "loss_forked",
+        "loss_forked_step",
+        "loss_forked_epoch",
```
harupy, line 101: Q. What is the difference between `loss_forked` and `loss_forked_epoch`?

harupy: They seem identical:

image

tests/pytorch/test_pytorch_autolog.py
```diff
@@ -43,6 +44,19 @@ def pytorch_model_without_validation():
     return trainer, run
 
 
+@pytest.fixture
+def pytorch_model_with_every_step_logged():
+    mlflow.pytorch.autolog(log_every_n_step=1)
```
harupy, line 49: @adamreeve Can we add a test case for `log_every_n_step != 1` (say 10)?

tests/pytorch/test_pytorch_autolog.py
adamreeve: Yes they should be exactly the same. PyTorch-Lightning records the same value using the original name (`loss_forked`) as well as the "forked" names (`loss_forked_epoch` and `loss_forked_step`)  in the callback metrics. The non-forked name is used when recording both the step values and epoch values, but I exclude the non-forked metric when logging against the step number as otherwise there would be some values logged against step numbers and some against epoch numbers under the same key so the data wouldn't make any sense.

I didn't also exclude the non-forked name when logging at the epoch end as I didn't want to change the existing behaviour there.

harupy: > I didn't also exclude the non-forked name when logging at the epoch end as I didn't want to change the existing behaviour there.

Makes sense, thanks for the explanation!

harupy (approved):
Looks good to me once https://github.com/mlflow/mlflow/pull/5497#discussion_r828751442 is addressed!

tests/pytorch/test_pytorch_autolog.py
adamreeve: Yep good idea, I've added that now

harupy:
Hi @adamreeve, I found https://github.com/PyTorchLightning/pytorch-lightning/pull/12195 which marks the `logger_connector` attribute as protected. Our daily cron job caught this change:

https://github.com/mlflow/mlflow/runs/5596868465?check_suite_focus=true#step:12:733

```python
    def _get_step_metrics(trainer):
        # logger_connector.metrics provides only on_step or on_epoch metrics depending on whether
        # we're at the end of an epoch, as opposed to logger_connector.callback_metrics which caches
        # all seen metrics and so will also contain epoch logged metrics after an epoch has
        # completed.
>       return trainer.logger_connector.metrics["callback"]
E       AttributeError: 'Trainer' object has no attribute 'logger_connector'
```

adamreeve:
Ah, that's unfortunate. I'll look into whether there's another way to support this feature with the existing public API, or I might need to ask for a new public API method in PyTorch-Lightning. In the meantime, would it be best to revert this PR merge?

harupy:
Thanks for the quick reply, yeah let's revert it )

harupy:
I should've applied the `enable-dev-tests` label to run tests against pytorch-lightning installed from the master branch.

PR #5503, opened 2022-03-17
Implement function to get MLflow pyfunc model dependencies

WeichenXu123:

Implement function to get MLflow pyfunc model dependencies

Context: when `pyfunc.load_model` called, the model might require different python env with current python env. So it is useful to provide an API to get model dependencies for user and tell user instructions for how to recover the dependencies.

For now we don't provide "env_manager" argument for `pyfunc.load_model` because `pyfunc_model.predict` allow many kinds of input data , transferring input data between different python process is hard.

The following API added in this PR:

```
def get_model_dependencies(model_uri, format="pip"):
    """
    Given a model URL, and format ("pip" or "conda", default value "pip"),
    return the downloaded dependency file path.
    If "pip" format specified but model does not have "requirements.txt" file,
    fallback to parse the pip section of the model's "conda.yaml" and ignore other
    non-pip dependencies in the "conda.yaml" file.
    """
```

## How is this patch tested?

Unit tests

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +717,24 @@ def load_model(
     return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
 
 
+def print_model_dependencies(model_uri):
+    """
+    Given a model URL, print model dependencies provided by model’s requirements.txt file.
+    And for databricks runtime, also print an instruction to update python environment with
+    the model dependencies.
+    """
+    local_path = _download_artifact_from_uri(artifact_uri=model_uri, output_path=None)
```
harupy, line 726: Looks like we only need `requirements.txt`. Is it possible to only download it rather than the entire model artifacts?

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +717,24 @@ def load_model(
     return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
 
 
+def print_model_dependencies(model_uri):
```
harupy, line 720: Should this be a fluent API (e.g. `mlflow.print_model_dependencies`)?

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +717,24 @@ def load_model(
...
+    local_path = _download_artifact_from_uri(artifact_uri=model_uri, output_path=None)
+    req_file_path = os.path.join(local_path, _REQUIREMENTS_FILE_NAME)
+    print(f"model {model_uri} dependencies:")
+    with open(req_file_path, "r") as f:
+        print(f.read())
+    print("\n")
+    if is_in_databricks_runtime():
+        print(f"On databricks notebook, you can run command '%pip install -r {req_file_path}' "
+              "to update current python environment with model’s requirements.txt, and rerun "
+              "inference commands.")
```
harupy, line 735: Instead of printing out dependencies, can we return them as a list of strings?

## Example use cases

```python
deps = list_model_dependencies(mdoel_uri)

# Check dependencies
print(deps)

# Construct pip install command
print("pip install", " ".join(deps))

# Copy dependencies
mlflow.*.log_model(..., pip_requirements=deps)
```

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +717,24 @@ def load_model(
...
+
+
```
harupy, line 737: Can you add a test if this is a public API?

mlflow/pyfunc/__init__.py
WeichenXu123: I have another idea, for each model URI, we can maintain a dict { model_uri --> downloaded_model_local_path } in memory, once we download a model to a temp local dir from model_uri, update the dict in memory. So that we can avoid download every time.

mlflow/pyfunc/__init__.py
WeichenXu123: we can.
and according to @mengxr 's suggestion, we should also print an instruction to tell user how to update python environment with these dependencies.

mlflow/pyfunc/__init__.py
harupy: > I have another idea, for each model URI, we can maintain a dict { model_uri --> downloaded_model_local_path } in memory, once we download a model to a temp local dir from model_uri, update the dict in memory. So that we can avoid download every time.

Do we really need this optimization? I don't think we do for now.

harupy:
I personally don't like this approach because a lot of actions are required to complete the installation:

1. Call `print_model_dependencies`
2. Find where `pip install -r ...` is located in the printed-out text.
3. Select the pip install command.
4. Copy
5. Move to a different cell.
6. Paste the command.
7. Run it

(I'm considering a user journey on Databricks).

Can we provide an API (e.g. `install_model_dependencies`) that calls `%pip install`?

mlflow/pyfunc/__init__.py
WeichenXu123: I hope we can do this optimization. It is easy. And note some model might be large. This optimization is very meaningful on databricks. Because:

current code by default it downloaded into a temporary directory created by tempfile, the temp directory is not delete until OS reboot. For databricks users, their all notebook processes are running on driver machine, and OS does not reboot until cluster restart.

Some model might be very big., e.g. some deep-learning model, it might be several GBs size.

WeichenXu123:
> I'm not a big fan of the current approach because a lot of actions are required to complete the installation:
> 
> 1. Call `print_model_dependencies`
> 2. Find where `pip install -r ...` is located in the printed-out text.
> 3. Select `pip install -r ...`.
> 4. Copy
> 5. Move to a different cell.
> 6. Paste the command.
> 7. Run it
> 
> (I'm considering a user journey on Databricks).
> 
> Can we provide a function (e.g. `install_model_dependencies`) that installs model dependencies using `%pip install` (I'm not sure if we can programmatically call `%pip` though)? I think that's more straightfoward.

one issue is, suppose we using conda to do that:
we need create a conda env first, and then activate the created env.
the tricky part is, seems we cannot trigger it in current process (except current process is a "bash",) we need activate it in a sub process and sub process must be a "bash", and after sub process exit, the activated env are also deactivated.
On databricks,
"%pip command" also require restarting current python process. (then all your previous generated variables in notebook are destroyed)

mlflow/pyfunc/__init__.py
dbczumar: I don't think the fluent API is the right place for this. The `mlflow.models` namespace seems better.

mlflow/pyfunc/__init__.py
dbczumar: I think it's best to download individual files, rather than the whole model. For pip requirements, we should download `requirements.txt`. For conda, we should download MLModel and use the `env` attribute of the pyfunc flavor spec to find the `conda.yaml` file & download it.

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +724,18 @@ def load_model(
     return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
 
 
+def print_model_dependencies(model_uri):
+    """
+    Given a model URL, print model dependencies provided by model’s requirements.txt file.
+    """
+    local_path = _download_artifact_from_uri(artifact_uri=model_uri, output_path=None)
+    req_file_path = os.path.join(local_path, _REQUIREMENTS_FILE_NAME)
```
dbczumar, line 732: What happens if the model doesn't have a `requirements.txt` file? Also, what happens if the user wants to get conda dependencies instead of just pip dependencies? We should support this.

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +724,18 @@ def load_model(
     return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
 
 
+def print_model_dependencies(model_uri):
```
dbczumar, line 727: As discussed in https://github.com/mlflow/mlflow/pull/5503#issuecomment-1070513180, printing the dependencies isn't the end goal for most users (apologies for naming the task this way). Instead, users want to be able to install the dependencies in their current environment.

Can we adjust this method to instead download the dependencies to the local filesystem and return a path to the downloaded dependencies? This should be as simple as downloading `requirements.txt` and / or `conda.yaml` (depending on what the user wants).

We should let the user specify the dependencies format (e.g. `format="conda"`, `format="pip"`) that we can use to determine whether we resolve `requirements.txt` or `conda.yaml`. This should default to `format="pip"`.

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
     return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
 
 
+def _download_model_conda_env(model_uri):
+    model_meta_path = _download_artifact_from_uri(os.path.join(model_uri, MLMODEL_FILE_NAME))
+    model_meta = Model.load(model_meta_path)
+    conda_yml_file_name = model_meta.flavors[FLAVOR_NAME][ENV]
```
dbczumar, line 725: `mlflow.utils.model_utils._get_flavor_configuration_from_uri` (https://github.com/mlflow/mlflow/blob/01182994ec31966dbe8528603d9e1a0ed2188234/mlflow/utils/model_utils.py#L45) will take care of getting the flavor configuration for us. It will also throw a descriptive error if the model doesn't have the pyfunc flavor. Let's use it.

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
+                    f"Parse conda.yaml file in the model directory failed, error: {repr(e)}."
+                )
+
+            if pip_deps is not None:
+                tmp_dir = tempfile.mkdtemp()
+                pip_file_path = os.path.join(tmp_dir, _REQUIREMENTS_FILE_NAME)
+                with open(pip_file_path, 'w') as f:
+                    f.write("\n".join(dep['pip']) + "\n")
+            else:
+                raise ValueError(
```
dbczumar, line 762: Can we raise `MlflowException` with error code `RESOURCE_DOOES_NOT_EXIST`

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
+            try:
+                with open(conda_yml_path, 'r') as yf:
+                    conda_yml = yaml.safe_load(yf)
+
+                for dep in conda_yml['dependencies']:
+                    if isinstance(dep, dict) and 'pip' in dep:
+                        pip_deps = dep['pip']
+                        break
+            except Exception as e:
+                raise RuntimeError(
```
dbczumar, line 752: Can we raise `MlflowException`?

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
+                tmp_dir = tempfile.mkdtemp()
+                pip_file_path = os.path.join(tmp_dir, _REQUIREMENTS_FILE_NAME)
+                with open(pip_file_path, 'w') as f:
+                    f.write("\n".join(dep['pip']) + "\n")
+            else:
+                raise ValueError(
+                    "No pip section found in conda.yaml file in the model directory."
+                )
+        if is_in_databricks_runtime():
+            print(f"On databricks notebook, you can run command '%pip install -r {pip_file_path}' "
```
dbczumar, line 766: We should use `_logger.info()` instead of print.

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
+                  "to update current python environment with model's dependencies.")
+        return pip_file_path
+
+    elif format == "conda":
+        conda_yml_path = _download_model_conda_env(model_uri)
+        if is_in_databricks_runtime():
+            print(f"On databricks notebook, you can run command '%conda env update -f {conda_yml_path}' "
```
dbczumar, line 773: We should use `_logger.info()` instead of print.

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
+                  "to update current python environment with model's dependencies.")
```
dbczumar, line 774: `%conda` has been removed in recent versions of the Databricks Machine Learning Runtime. I don't think we should print this.

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
+        return conda_yml_path
+    else:
+        raise ValueError(f"Illegal format argument '{format}'.")
```
dbczumar, line 777: Can we raise `MlflowException` with error code `INVALID_PARAMETER_VALUE`?

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
+    Given a model URL, and format, return the downloaded dependency file path.
+    Available dependency file format includes "pip" and "conda".
+    """
+    req_file_uri = os.path.join(model_uri, _REQUIREMENTS_FILE_NAME)
+
+    if format == "pip":
+        try:
+            pip_file_path = _download_artifact_from_uri(req_file_uri)
+        except Exception as e:
+            # fallback to download conda.yaml file and parse the "pip" section from it.
```
dbczumar, line 740: Can we print (`_logger.info()`) a message in this case telling the user that we're going to fetch pip requirements from the conda.yaml file and that other conda dependencies will be ignored?

dbczumar: In general, if the user specifies `format=pip`, I think it would be useful to download the `conda.yaml` and check if there are non-pip requirements. If there are non-pip requirements in `conda.yaml`, we should print a warning to the user indicating that these requirements will be excluded.

mlflow/pyfunc/__init__.py
dbczumar, line 766:
```suggestion
            print(f"To install these model dependencies in your Databricks notebook, run the following command: '%pip install -r {pip_file_path}' "
```

mlflow/pyfunc/__init__.py
```diff
@@ -716,6 +719,64 @@ def load_model(
...
 
 
+def _download_model_conda_env(model_uri):
+    model_meta_path = _download_artifact_from_uri(os.path.join(model_uri, MLMODEL_FILE_NAME))
+    model_meta = Model.load(model_meta_path)
+    conda_yml_file_name = model_meta.flavors[FLAVOR_NAME][ENV]
+    return _download_artifact_from_uri(os.path.join(model_uri, conda_yml_file_name))
+
+
+def get_model_dependencies(model_uri, format="pip"):
```
dbczumar, line 729: @harupy @WeichenXu123 Note that this doesn't handle the case where `conda.yaml`'s `pip` section includes `-r requirements.txt` references. It also doesn't handle the case where a user manually constructs a `requirements.txt` file in the MLflow Model that references other `requirements.txt` files.

I don't think we need to handle either of these cases. In MLflow 2.0, I think we should implement some additional restrictions on the accepted pip requirements / conda + pip requirements to disallowed `-r`. In MLflow 2.0, we should also expand the pyfunc flavor's `env` field to include multiple environment specs: pip `requirements.txt` and Conda `conda.yaml`; it's currently a bit strange that `env` doesn't include `requirements.txt`.

dbczumar (commented):
@WeichenXu123 Looks great! Just a few comments.

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+
+def get_model_dependencies(model_uri, format="pip"):  # pylint: disable=redefined-builtin
+    """
+    Given a model URL, and format ("pip" or "conda", default value "pip"),
+    return the downloaded dependency file path.
+    If "pip" format specified but model does not have "requirements.txt" file,
+    fallback to parse the pip section of the model's "conda.yaml" and ignore other
+    non-pip dependencies in the "conda.yaml" file.
+    """
+    req_file_uri = os.path.join(model_uri, _REQUIREMENTS_FILE_NAME)
```
harupy, line 743:
```suggestion
    req_file_uri = os.path.join(model_uri, _REQUIREMENTS_FILE_NAME)
```

I think this causes an issue in Windows.

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+    if format == "pip":
+        try:
+            pip_file_path = _download_artifact_from_uri(req_file_uri)
+        except Exception:
+            # fallback to download conda.yaml file and parse the "pip" section from it.
+            _logger.info(
+                f"Download model '{_REQUIREMENTS_FILE_NAME}' file failed. Fallback to fetch "
+                "pip requirements from the model's 'conda.yaml' file and other conda "
+                "dependencies will be ignored."
+            )
```
harupy, line 754: I think the error information should be included.

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+    return the downloaded dependency file path.
+    If "pip" format specified but model does not have "requirements.txt" file,
+    fallback to parse the pip section of the model's "conda.yaml" and ignore other
+    non-pip dependencies in the "conda.yaml" file.
+    """
+    req_file_uri = os.path.join(model_uri, _REQUIREMENTS_FILE_NAME)
+
+    if format == "pip":
+        try:
+            pip_file_path = _download_artifact_from_uri(req_file_uri)
```
harupy, line 747:
```suggestion
            return _download_artifact_from_uri(req_file_uri)
```

Can we exit-fast if this `_download_artifact_from_uri` call succeeds?

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+            conda_deps = []
+            try:
+                with open(conda_yml_path, "r") as yf:
+                    conda_yml = yaml.safe_load(yf)
+
+                for dep in conda_yml["dependencies"]:
+                    if (
+                        isinstance(dep, str)
+                        and not dep.startswith("python=")
+                        and not dep.startswith("pip=")
```
harupy, line 766:
```suggestion
                        and not dep.startswith("pip<=")
```

I think `<=` shoulbe be used since we don't use `=`:

https://github.com/mlflow/mlflow/blob/2e175e7fd65d86d12d204b648aa5ff73063a7dc9/mlflow/utils/environment.py#L57

tests/pyfunc/test_dependencies_functions.py
```diff
@@ -110,3 +111,81 @@ def mock_get_installed_version_fn(package, module=None):
                     "detecting model dependency mismatches"
                 )
             )
+
+
+@pytest.mark.large
+def test_get_model_dependencies(tmpdir):
```
harupy, line 117:
```suggestion
def test_get_model_dependencies(tmp_path):
```

Let's use the `tmp_path` fixture.

tests/pyfunc/test_dependencies_functions.py
```diff
@@ -110,3 +111,81 @@ def mock_get_installed_version_fn(package, module=None):
...
+    # Test getting pip dependencies
+    assert get_model_dependencies(model_path, format="pip") == req_file_path
+
+    # Test getting pip dependencies will print instructions on databricks
+    with mock.patch("mlflow.pyfunc._logger.info") as mock_log_info:
+        get_model_dependencies(model_path, format="pip")
+        mock_log_info.assert_not_called()
+
+        mock_log_info.reset_mock()
+        with mock.patch("mlflow.pyfunc.is_in_databricks_runtime", lambda: True):
```
harupy, line 170:
```suggestion
        with mock.patch("mlflow.pyfunc.is_in_databricks_runtime", return_value=True):
```

Can we use `return_value` ([reason](https://github.com/mlflow/mlflow/pull/5386#discussion_r808718952))?

mlflow/pyfunc/__init__.py
dbczumar, line 766: Should we be excluding these from the printed list of `conda_deps`? They'll still be excluded from the `requirements.txt` file, so I think it's valid to include them in the warning message.

tests/pyfunc/test_dependencies_functions.py
```diff
@@ -110,3 +111,81 @@ def mock_get_installed_version_fn(package, module=None):
...
+  - mlflow
+  - cloudpickle==2.0.0
+  - scikit-learn==1.0.1
+name: mlflow-env"""
+        )
+
+    model_path = tmpdir.strpath
+
+    # Test getting pip dependencies
+    assert get_model_dependencies(model_path, format="pip") == req_file_path
```
harupy, line 162: Can we check the file content is correct?

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
     return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
 
 
+def _download_model_conda_env(model_uri):
+    conda_yml_file_name = _get_flavor_configuration_from_uri(model_uri, FLAVOR_NAME)[ENV]
```
harupy, line 731:
```suggestion
    conda_yml_file_name = _get_flavor_configuration_from_uri(model_uri, FLAVOR_NAME)[ENV]
```

Does the `ENV` key already exist?

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+            if pip_deps is not None:
+                tmp_dir = tempfile.mkdtemp()
+                pip_file_path = os.path.join(tmp_dir, _REQUIREMENTS_FILE_NAME)
+                with open(pip_file_path, "w") as f:
+                    f.write("\n".join(pip_deps) + "\n")
+            else:
+                raise MlflowException(
+                    "No pip section found in conda.yaml file in the model directory.",
+                    error_code=RESOURCE_DOES_NOT_EXIST,
+                )
```
harupy, line 786: Can you add test for this exception?

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+
+        return pip_file_path
+
+    elif format == "conda":
+        conda_yml_path = _download_model_conda_env(model_uri)
+        return conda_yml_path
+    else:
+        raise MlflowException(
+            f"Illegal format argument '{format}'.", error_code=INVALID_PARAMETER_VALUE
+        )
```
harupy, line 806: Can you add a test for this exception?

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+                for dep in conda_yml["dependencies"]:
+                    if (
+                        isinstance(dep, str)
+                        and not dep.startswith("python=")
+                        and not dep.startswith("pip=")
+                    ):
+                        conda_deps.append(dep)
+                    if isinstance(dep, dict) and "pip" in dep:
+                        pip_deps = dep["pip"]
+                        break
```
harupy, line 771:
```suggestion
```

I don't think we should break here because `conda.yaml` might look like this:

```yaml
channels:
- conda-forge
dependencies:
- pip:
  - mlflow
  - cloudpickle==2.0.0
  - scikit-learn==1.0.1
- python=3.7.12
- pip=22.0.3
- scikit-learn=0.22.0
- tensorflow=2.0.0
name: mlflow-env
```

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
     return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
 
 
+def _download_model_conda_env(model_uri):
+    conda_yml_file_name = _get_flavor_configuration_from_uri(model_uri, FLAVOR_NAME)[ENV]
+    return _download_artifact_from_uri(os.path.join(model_uri, conda_yml_file_name))
```
harupy, line 732: Same here: https://github.com/mlflow/mlflow/pull/5503/files#r836022727

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +727,85 @@ def load_model(
...
+
+
+def get_model_dependencies(model_uri, format="pip"):  # pylint: disable=redefined-builtin
+    """
+    Given a model URL, and format ("pip" or "conda", default value "pip"),
```
harupy, line 737:
```suggestion
    Given a model URI, and format ("pip" or "conda", default value "pip"),
```

nit

mlflow/pyfunc/__init__.py
harupy: Looks like it does.

tests/pyfunc/test_dependencies_functions.py
harupy, line 117: Can we split the test?

mlflow/pyfunc/__init__.py
WeichenXu123: we need to also print 
```
        if is_in_databricks_runtime():
            _logger.info(
                "To install these model dependencies in your Databricks notebook, run the "
                f"following command: '%pip install -r {pip_file_path}'."
            )
``
before return.

File: mlflow/pyfunc/__init__.py (hunk 19 above)
Comment (WeichenXu123) [reply mlflow/pyfunc/__init__.py:766]: ok.

File: mlflow/pyfunc/__init__.py (hunk 23 above)
Comment (WeichenXu123) [reply mlflow/pyfunc/__init__.py:732]: Not sure. What's the case ENV key does not exists ? In old version ?

File: mlflow/pyfunc/__init__.py [hunk 29]
```diff
@@ -724,6 +728,80 @@ def load_model(
...
+            # fallback to download conda.yaml file and parse the "pip" section from it.
+            _logger.info(
+                f"Download model '{_REQUIREMENTS_FILE_NAME}' file failed, error is {repr(e)}. "
+                "Fallback to fetch pip requirements from the model's 'conda.yaml' file and "
+                "other conda dependencies will be ignored."
+            )
+            conda_yml_path = _download_model_conda_env(model_uri)
+            pip_deps = None
+            conda_deps = []
+            try:
```
harupy, line 759:
```suggestion
        conda_yml_path = _download_model_conda_env(model_uri)
        pip_deps = None
        conda_deps = []
        try:
```

I think we can unindent this block to avoid nested try-except.

tests/pyfunc/test_dependencies_functions.py
```diff
@@ -110,3 +113,100 @@ def mock_get_installed_version_fn(package, module=None):
...
+    with mock.patch("mlflow.pyfunc._logger.info") as mock_log_info:
+        get_model_dependencies(model_path, format="pip")
+        mock_log_info.assert_not_called()
+
+        mock_log_info.reset_mock()
+        with mock.patch("mlflow.pyfunc.is_in_databricks_runtime", return_value=True):
+            get_model_dependencies(model_path, format="pip")
+            mock_log_info.assert_called_once_with(
+                "To install these model dependencies in your Databricks notebook, run the "
+                f"following command: '%pip install -r {str(req_file)}'."
```
harupy, line 140:
```suggestion
                f"following command: '%pip install -r {req_file}'."
```

`str` can be removed.

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +728,80 @@ def load_model(
...
+                with open(conda_yml_path, "r") as yf:
```
harupy, line 760:
```suggestion
        conda_yml_path = _download_model_conda_env(model_uri)
        pip_deps = None
        conda_deps = []
        try:
            ...
```

Can we pull this block out from try-catch?

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +728,80 @@ def load_model(
...
+
+                for dep in conda_yml["dependencies"]:
+                    if isinstance(dep, str):
+                        conda_deps.append(dep)
+                    if isinstance(dep, dict) and "pip" in dep:
+                        pip_deps = dep["pip"]
+            except Exception as e:
+                raise MlflowException(
+                    f"Parse conda.yaml file in the model directory failed, error: {repr(e)}."
+                )
```
harupy, line 771: Do we need this try-catch?

mlflow/pyfunc/__init__.py
harupy: I missed that line.

mlflow/pyfunc/__init__.py
harupy: Never mind this comment.

mlflow/pyfunc/__init__.py
WeichenXu123: I preferred add this try catch, so we can wrap the original exception with more clear error message.

harupy: This is how the error message looks like:

```python
2022/03/29 11:25:20 INFO mlflow.pyfunc: Download model 'requirements.txt' file failed, error is OSError("No such file or directory: '/var/folders/33/kkm875tn7cxb89b8mmcykrx00000gn/T/tmpu3scvbl2/requirements.txt'"). Fallback to fetch pip requirements from the model's 'conda.yaml' file and other conda dependencies will be ignored.
  File "/Users/harutakakawamura/Desktop/repositories/mlflow/mlflow/pyfunc/__init__.py", line 749, in get_model_dependencies
    pip_file_path = _download_artifact_from_uri(req_file_uri)
  File "/Users/harutakakawamura/Desktop/repositories/mlflow/mlflow/tracking/artifact_utils.py", line 96, in _download_artifact_from_uri
    artifact_path=artifact_path, dst_path=output_path
  File "/Users/harutakakawamura/Desktop/repositories/mlflow/mlflow/store/artifact/local_artifact_repo.py", line 79, in download_artifacts
    raise IOError("No such file or directory: '{}'".format(local_artifact_path))
OSError: No such file or directory: '/var/folders/33/kkm875tn7cxb89b8mmcykrx00000gn/T/tmpu3scvbl2/requirements.txt'

During handling of the above exception, another exception occurred:

  File "/Users/harutakakawamura/Desktop/repositories/mlflow/mlflow/pyfunc/__init__.py", line 764, in get_model_dependencies
    for dep in conda_yml["dependencies"]:
KeyError: 'dependencies'

During handling of the above exception, another exception occurred:

  File "a.py", line 39, in <module>
    test_get_model_dependencies_read_conda_file(Path(tmp_path))
  File "a.py", line 33, in test_get_model_dependencies_read_conda_file
    Path(mlflow.pyfunc.get_model_dependencies(model_path, format="pip")).read_text()
  File "/Users/harutakakawamura/Desktop/repositories/mlflow/mlflow/pyfunc/__init__.py", line 771, in get_model_dependencies
    f"Parse conda.yaml file in the model directory failed, error: {repr(e)}."
mlflow.exceptions.MlflowException: Parse conda.yaml file in the model directory failed, error: KeyError('dependencies').
```

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +729,80 @@ def load_model(
...
+    return _download_artifact_from_uri(append_to_uri_path(model_uri, conda_yml_file_name))
+
+
+def get_model_dependencies(model_uri, format="pip"):  # pylint: disable=redefined-builtin
+    """
+    Given a model URI, and format ("pip" or "conda", default value "pip"),
+    return the downloaded dependency file path.
+    If "pip" format specified but model does not have "requirements.txt" file,
+    fallback to parse the pip section of the model's "conda.yaml" and ignore other
+    non-pip dependencies in the "conda.yaml" file.
```
harupy, line 743:
```suggestion
    non-pip dependencies in the "conda.yaml" file.
    
    :param: ...
```

Can you add parame descriptions if this is a public API?

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +729,80 @@ def load_model(
...
+            conda_deps = []
+            try:
+                with open(conda_yml_path, "r") as yf:
+                    conda_yml = yaml.safe_load(yf)
+
+                for dep in conda_yml["dependencies"]:
+                    if isinstance(dep, str):
+                        conda_deps.append(dep)
+                    if isinstance(dep, dict) and "pip" in dep:
+                        pip_deps = dep["pip"]
```
harupy, line 768: I think this block can be improved using `for-else`:

```python
        pip_deps_index = None
        conda_deps = conda_yml.get("dependencies", [])
        for index, dep in enumerate(conda_deps):
            if _is_pip_deps(dep):
                pip_deps_index = index
                break
        else:
            raise MlflowException(
                "No pip section found in conda.yaml file in the model directory.",
                error_code=RESOURCE_DOES_NOT_EXIST,
            )
        pip_deps = conda_deps.pop[pip_deps_index]
```

mlflow/pyfunc/__init__.py
```diff
@@ -664,9 +667,11 @@ def _warn_dependency_requirement_mismatches(model_path):
             mismatch_str = " - " + "\n - ".join(mismatch_infos)
             warning_msg = (
                 "Detected one or more mismatches between the model's dependencies and the current "
-                f"Python environment:\n{mismatch_str}"
+                f"Python environment:\n{mismatch_str}\nTo get full model dependencies, you can "
+                "call `mlflow.pyfunc.get_model_dependencies(model_uri)`."
```
harupy, line 671: I'd fix this warning message to:

```suggestion
                f"Python environment:\n{mismatch_str}\nTo fix the mismatches,
                "call `mlflow.pyfunc.get_model_dependencies(model_uri)` and run the printed out pip install command"
```

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +729,85 @@ def load_model(
...
+
+def _get_model_dependencies(model_uri, format="pip"):  # pylint: disable=redefined-builtin
+    if format == "pip":
+        req_file_uri = append_to_uri_path(model_uri, _REQUIREMENTS_FILE_NAME)
+        try:
+            return _download_artifact_from_uri(req_file_uri)
+        except Exception as e:
+            # fallback to download conda.yaml file and parse the "pip" section from it.
+            _logger.info(
+                f"Download model '{_REQUIREMENTS_FILE_NAME}' file failed, error is {repr(e)}. "
```
harupy, line 745:
```suggestion
                f"Encountered unexpected error while downloading {req_file_uri}: {repr(e)}."
```

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +729,85 @@ def load_model(
...
+        )
+
+
+def get_model_dependencies(model_uri, format="pip"):  # pylint: disable=redefined-builtin
+    """
+    Given a model URI, and format ("pip" or "conda", default value "pip"),
+    return the downloaded dependency file path.
+    If "pip" format specified but model does not have "requirements.txt" file,
+    fallback to parse the pip section of the model's "conda.yaml" and ignore other
+    non-pip dependencies in the "conda.yaml" file.
```
harupy, line 794: I think this can be included in the parameter description of `format`.

mlflow/pyfunc/__init__.py
```diff
@@ -664,9 +667,12 @@ def _warn_dependency_requirement_mismatches(model_path):
             mismatch_str = " - " + "\n - ".join(mismatch_infos)
             warning_msg = (
                 "Detected one or more mismatches between the model's dependencies and the current "
-                f"Python environment:\n{mismatch_str}"
+                f"Python environment:\n{mismatch_str}\n"
+                "To fix the mismatches, call `mlflow.pyfunc.get_model_dependencies(model_uri)` "
+                "and run the printed out pip install command."
```
dbczumar, line 672: If the user calls `get_model_dependencies(format="conda")`, they won't see a command.

```suggestion
                "To fix the mismatches, call `mlflow.pyfunc.get_model_dependencies(model_uri)` "
                "to fetch the model's environment and install dependencies from the resulting file."
```

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +730,85 @@ def load_model(
...
+        req_file_uri = append_to_uri_path(model_uri, _REQUIREMENTS_FILE_NAME)
+        try:
+            return _download_artifact_from_uri(req_file_uri)
+        except Exception as e:
+            # fallback to download conda.yaml file and parse the "pip" section from it.
+            _logger.info(
+                f"Download model '{_REQUIREMENTS_FILE_NAME}' file failed, error is {repr(e)}. "
+                "Fallback to fetch pip requirements from the model's 'conda.yaml' file and "
+                "other conda dependencies will be ignored."
+            )
```
dbczumar, line 749:
```suggestion
                f"Downloading model '{_REQUIREMENTS_FILE_NAME}' file failed, error is {repr(e)}. "
                "Falling back to fetching pip requirements from the model's 'conda.yaml' file. "
                "Other conda dependencies will be ignored."
            )
```

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +730,85 @@ def load_model(
...
+
+        pip_deps = conda_deps.pop(pip_deps_index)["pip"]
+        tmp_dir = tempfile.mkdtemp()
+        pip_file_path = os.path.join(tmp_dir, _REQUIREMENTS_FILE_NAME)
+        with open(pip_file_path, "w") as f:
+            f.write("\n".join(pip_deps) + "\n")
+
+        if len(conda_deps) > 0:
+            _logger.warning(
+                f"The following conda dependencies are excluded: {', '.join(conda_deps)}."
```
dbczumar, line 775:
```suggestion
                f"The following conda dependencies have been excluded from the environment file: {', '.join(conda_deps)}."
```

mlflow/pyfunc/__init__.py
```diff
@@ -724,6 +730,85 @@ def load_model(
...
+    :param model_uri: The uri of the model to get dependencies from.
+    :param format: The format of the returned dependency file. If "pip" specified,
+                   return path of "requirements.txt" file which contains pip dependencies,
+                   if "conda" specified, return path of "conda.yaml" file which contains
+                   conda environment config. If "pip" format specified but model does not
+                   have "requirements.txt" file, fallback to parse the pip section of the
+                   model's "conda.yaml" and ignore other non-pip dependencies in the
+                   "conda.yaml" file. Default value is "pip".
+
+    :return: return the path of the model's pip dependency file or the path of "conda.yaml" file
```
dbczumar, line 800:
```suggestion
    :param format: The format of the returned dependency file. If the ``"pip"`` format is specified,
                   the path to a pip ``requirements.txt`` file which is returned. If the ``"conda"`` format
                   is specified, the path to a ``"conda.yaml"`` file is returned . If the ``"pip"`` format is
                   specified but the model was not saved with a ``requirements.txt`` file, the ``pip`` section
                   of the model's ``conda.yaml`` file is extracted instead, and any additional conda dependencies
                   are ignored. Default value is ``"pip"``.
    :return: The local filesystem path to either a pip ``requirements.txt`` file (if ``format="pip"``)
             or a ``conda.yaml`` file (if ``format="conda"``) specifying the model's dependencies.
```

dbczumar (approved):
LGTM with a couple small docs / message nits! Thanks @WeichenXu123 !

PR #5510, opened 2022-03-17
Added autologging of input example and signature for tf.estimator and keras fit_generator

bali0019:

Autologging of input example/signature for tf.estimator
Autologging of input example/signature for fit_generator for keras
Cleanup of log models inside keras callback

## How is this patch tested?

Unit tests

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Added autologging of input example and signature for tf.estimator and keras fit_generator

Components 

Interface 

Language 

Integrations


mlflow/tensorflow/__init__.py
```diff
@@ -1087,10 +1177,11 @@ def _patch_implementation(
                     # & tf.keras callbacks if necessary
                     callbacks = list(kwargs.get("callbacks") or [])
                     kwargs["callbacks"], self.log_dir = _setup_callbacks(
-                        callbacks, log_models, metrics_logger
+                        callbacks, False, metrics_logger
```
BenWilson2, line 1180: along with line 1107, both calls to `_setup_callbacks()` have the 2nd argument as 'False'. Can we remove this dead code since all 4 usages of that function are now set to a bool False constant?

bali0019: cleaned up in the last commit. thanks!

examples/tensorflow/tf2/train_predict_2.py
```diff
@@ -63,7 +63,7 @@ def eval_input_fn(features, labels, batch_size):
 
 
 # Enable auto-logging to MLflow to capture TensorBoard metrics.
-mlflow.tensorflow.autolog()
+mlflow.tensorflow.autolog(log_model_signatures=False)
```
dbczumar, line 66: Why do we need to disable signature logging for this example?

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
         if metric_key is not None:
             metrics_logger.record_metrics(restored_metrics, stopped_epoch + 1)
 
+    def _log_keras_model_with_warning(history, args):
```
dbczumar, line 971: Nit: I don't think we need to include `with_warning` in the method name.

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+        try:
+            if log_models:
```
dbczumar, line 973: If `log_models` is `False`, the rest of this method is a no-op. Can we move the `log_models` check out of this method and check it at the call sites?

I.e.

```
if log_models:
    _log_keras_model(history, args)
```

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                            "Tensorflow keras autologging only "
+                            "supports input types of: numpy.ndarray, "
+                            "dict(<key> -> numpy.ndarray), tensorflow.data.Dataset, "
+                            "or tensorflow.keras.utils.Sequence"
+                        )
+
+                    return input_example_slice
+
+                def _infer_model_signature(input_data_slice):
+                    try:
```
dbczumar, line 1032: `resolve_input_example_and_signature` already does error handling for signature inference failures (see https://github.com/mlflow/mlflow/blob/01182994ec31966dbe8528603d9e1a0ed2188234/mlflow/utils/autologging_utils/__init__.py#L168-L175), so we don't need to try / catch in this method.

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                    keras_model=history.model,
+                    artifact_path="model",
+                    input_example=input_example,
+                    signature=signature,
+                    registered_model_name=get_autologging_config(
+                        FLAVOR_NAME, "registered_model_name", None
+                    ),
+                )
+        except Exception as e:
+            warnings.warn(f"Failed to autolog keras model: {str(e)}")
```
dbczumar, line 1073: We shouldn't try / catch the model logging procedure. If overall model logging fails, the run should be marked as "failed". The autologging safe patch framework handles this automatically. We can just remove the outer try/catch here.

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                            input_training_data = input_training_data_original[:][0]
+                        return input_training_data
+
+                    input_training_data = _extract_from_sequence_or_generator()
+                    if isinstance(input_training_data, np.ndarray):
+                        input_example_slice = input_training_data[:INPUT_EXAMPLE_SAMPLE_ROWS]
+                    elif isinstance(input_training_data, tensorflow.data.Dataset):
+                        steps = 1
+                        if history.params is not None and "steps" in history.params:
+                            steps = history.params["steps"]
```
dbczumar, line 996: Nit: `steps = (history.params or {}).get("steps", 1)` should work as a simplification

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                        input_training_data = input_training_data_original
+                        if isinstance(input_training_data_original, types.GeneratorType):
+                            input_training_data = next(input_training_data_original)[0]
+                        elif isinstance(
+                            input_training_data_original, tensorflow.keras.utils.Sequence
+                        ):
+                            input_training_data = input_training_data_original[:][0]
+                        return input_training_data
+
+                    input_training_data = _extract_from_sequence_or_generator()
```
dbczumar, line 990: Nit: I don't think we need to define a separate function here since it only references nonlocal state and is only ever called once. The following is easier to maintain:

```suggestion
                    input_training_data = args[0]
                    input_example_slice = None

                      if isinstance(input_training_data, types.GeneratorType):
                          input_training_data = next(input_training_data)[0]
                      elif isinstance(
                          input_training_data, tensorflow.keras.utils.Sequence
                      ):
                          input_training_data = input_training_data[:][0]
```

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                import types
+
+                def _get_input_data_slice():
+                    input_training_data_original = args[0]
+                    input_example_slice = None
+
+                    def _extract_from_sequence_or_generator():
+                        input_training_data = input_training_data_original
+                        if isinstance(input_training_data_original, types.GeneratorType):
+                            input_training_data = next(input_training_data_original)[0]
```
dbczumar, line 983: Does calling `next()` on the data affect training execution at all? I.e. by consuming part of this iterator, are we preventing the model from consuming it during training? If so, we should make a copy of the object before mutating it and add test coverage to make sure we aren't mutating the training dataset.

mlflow/tensorflow/__init__.py
```diff
@@ -805,7 +856,37 @@ def log_model_without_starting_new_run():
                         tf_meta_graph_tags=[tag_constants.SERVING],
                         tf_signature_def_key="predict",
                     )
-                    save_model(path=local_path, mlflow_model=mlflow_model, **save_model_kwargs)
+
+                    input_example = None
+                    signature = None
+                    if log_models:
```
dbczumar, line 862: Nice! If `log_models` is `False`, we shouldn't be logging the model at all. Looks like the previous version of the code didn't respect this flag.

Can we move the `save_model` and `MlflowClient.log_artifacts()` calls under this `if log_models` condition as well?

mlflow/tensorflow/__init__.py
```diff
@@ -805,7 +856,37 @@ def log_model_without_starting_new_run():
...
+                                    k: tensorflow.convert_to_tensor(v)
+                                    for k, v in input_example_slice.items()
+                                }
+                            return tensorflow.data.Dataset.from_tensor_slices(input_slices).batch(1)
+
+                        predicted_values = [v for v in tf_model.predict(predict_input_fn)]
+
+                        input_example, signature = resolve_input_example_and_signature(
+                            lambda: input_example_slice,
+                            lambda in_ex: infer_signature(input_example_slice, predicted_values[0]),
```
dbczumar, line 877:
```suggestion
                            lambda in_ex: infer_signature(in_ex, predicted_values[0]),
```

mlflow/tensorflow/__init__.py
```diff
@@ -769,6 +787,39 @@ def train(original, self, *args, **kwargs):
...
+                if isinstance(next_input, dict):
+                    input_ex_dict_tensor = {
+                        k: _extract_tensor_as_numpy_input_example_slice(v)
+                        for k, v in next_input.items()
+                    }
+                elif isinstance(next_input, tensorflow.Tensor):
+                    input_ex_dict_tensor = _extract_tensor_as_numpy_input_example_slice(next_input)
+                return input_ex_dict_tensor
+
+            input_training_data = kwargs.get("input_fn")()
```
dbczumar, line 802: Just making sure - is it always safe to call this method multiple times?

mlflow/tensorflow/__init__.py
```diff
@@ -752,6 +754,22 @@ def autolog(
         warnings.warn("Could not log to MLflow. TensorFlow versions below 1.12 are not supported.")
         return
 
+    def _extract_tensor_as_numpy_input_example_slice(next_input):
```
dbczumar, line 757: For maintainability, can we add docstrings to the internal methods that we're defining throughout this PR?

mlflow/tensorflow/__init__.py
```diff
@@ -805,7 +856,37 @@ def log_model_without_starting_new_run():
...
+                        def predict_input_fn():
+                            input_slices = input_example_slice
+                            if isinstance(input_example_slice, dict):
+                                input_slices = {
+                                    k: tensorflow.convert_to_tensor(v)
+                                    for k, v in input_example_slice.items()
+                                }
+                            return tensorflow.data.Dataset.from_tensor_slices(input_slices).batch(1)
+
+                        predicted_values = [v for v in tf_model.predict(predict_input_fn)]
```
dbczumar, line 873: What does this accomplish?

mlflow/tensorflow/__init__.py
```diff
@@ -769,6 +787,39 @@ def train(original, self, *args, **kwargs):
...
+        def _extract_data_from_input_fn():
+            def _extract_dict_or_tensor(next_input):
+                input_ex_dict_tensor = None
+                if isinstance(next_input, dict):
+                    input_ex_dict_tensor = {
+                        k: _extract_tensor_as_numpy_input_example_slice(v)
+                        for k, v in next_input.items()
+                    }
+                elif isinstance(next_input, tensorflow.Tensor):
+                    input_ex_dict_tensor = _extract_tensor_as_numpy_input_example_slice(next_input)
```
dbczumar, line 799: Can we add some inline comments that explain what the data looks like for each type, including a small example?

mlflow/tensorflow/__init__.py
```diff
@@ -769,6 +787,39 @@ def train(original, self, *args, **kwargs):
...
+                return input_ex_dict_tensor
+
+            input_training_data = kwargs.get("input_fn")()
+            input_features = None
+            if isinstance(input_training_data, tuple):
+                input_features = _extract_dict_or_tensor(input_training_data[0])
+            elif isinstance(input_training_data, tensorflow.data.Dataset):
+                train_iter = iter(input_training_data)
+                next_input = next(train_iter)[0]
+                input_features = _extract_dict_or_tensor(next_input)
```
dbczumar, line 809: Can we add some inline comments that explain what the data looks like for each type, including a small example?

This will help with maintainability; for example, it will make it easier to understand why `next_input = next(train_iter)[0]`. For my own understanding, why do we take the first element of the resulting output and discard the others?

mlflow/tensorflow/__init__.py
```diff
@@ -769,6 +787,39 @@ def train(original, self, *args, **kwargs):
...
+                    input_ex_dict_tensor = _extract_tensor_as_numpy_input_example_slice(next_input)
+                return input_ex_dict_tensor
+
+            input_training_data = kwargs.get("input_fn")()
+            input_features = None
+            if isinstance(input_training_data, tuple):
+                input_features = _extract_dict_or_tensor(input_training_data[0])
+            elif isinstance(input_training_data, tensorflow.data.Dataset):
+                train_iter = iter(input_training_data)
+                next_input = next(train_iter)[0]
```
dbczumar, line 808: For my own understanding, why do we take the first element of the resulting output and discard the others?

From the docs for https://www.tensorflow.org/api_docs/python/tf/data/Dataset, I see the following:

```
Element: A single output from calling next() on a dataset iterator. Elements may be nested structures containing multiple components. For example, the element (1, (3, "apple")) has one tuple nested in another tuple. The components are 1, 3, and "apple".
```

This seems to imply that `next()` may yield tuple or another non-scalar result. By only extracting the first element, are we missing some of the dataset's structure?

mlflow/tensorflow/__init__.py
```diff
@@ -887,6 +968,110 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                            return np.array(
+                                [
+                                    v[0]
+                                    for v in input_example_n_steps.take(
+                                        INPUT_EXAMPLE_SAMPLE_ROWS
+                                    ).as_numpy_iterator()
+                                ]
+                            )
+
+                        input_example_slice = _extract_n_steps(input_training_data)
```
dbczumar, line 1015: Not fully following what's going on here. Why does extracting the first `INPUT_EXAMPLE_SAMPLE_ROWS` from a `tf.Dataset` depend on `history.params["steps"]`? Is this step-dependent logic necessary? If it is, can we add a clear docstring to this method to explain why?

dbczumar (commented):
@bali0019 Awesome stuff! Thanks for contributing this! I left a few comments, mostly code structure / readability nits and some questions for my own understanding.

examples/tensorflow/tf2/train_predict_2.py
bali0019: Good point. When `pyfunc_model` is loaded in this example, it is predicted with `pd.DataFrame` and the predictions are validated against that type as well, but if `log_model_signatures` is set to `True`, then the output is always of predict is always of type numpy.ndarray (or dict ) and so the example logic breaks. I thought it was just simpler to turn `log_model_signatures` off for this example, but let me know if you would like the example to be adjusted rather?!

dbczumar: Interesting! In that case, the fact that automatically including a signature changes the output type of `predict()` constitutes a backwards-incompatible change. If a user was running the workflow from this example in production using MLflow 1.24 and then upgraded to a version of MLflow that contains the changes made in this PR, their workflow would break.

Can we root cause why the output type is changing and see if we can fix or work around it?

dbczumar: In particular, it's a bit curious to me that the output of `predict()` with a Pandas DataFrame input (as provided by this example - https://github.com/mlflow/mlflow/pull/5510/files#diff-3ca06392b52de110cff4856f4cd144a4fdc801fba93dce851aaa7f0d66993c79R153-R160) would produce anything other than a DataFrame, given the following logic: https://github.com/mlflow/mlflow/blob/2e175e7fd65d86d12d204b648aa5ff73063a7dc9/mlflow/tensorflow/__init__.py#L533

bali0019: With`log_model_signatures` set to `True`, the data is enforced as 
```
{'SepalLength': array([5.1, 5.9, 6.9]), 'SepalWidth': array([3.3, 3. , 3.1]), 'PetalLength': array([1.7, 4.2, 5.4]), 'PetalWidth': array([0.5, 1.5, 2.1])}```

by [_enforce_schema](https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/__init__.py#L622)
to be of type dict (str -> ndarray), which actually hits [this](https://github.com/mlflow/mlflow/blob/2e175e7fd65d86d12d204b648aa5ff73063a7dc9/mlflow/tensorflow/__init__.py#L531)...i think to be able to put a backward compatibility fix we need to know the original type passed to the pyfunc model predict, which is lost when it gets to the TF flavor's [predict](https://github.com/mlflow/mlflow/blob/2e175e7fd65d86d12d204b648aa5ff73063a7dc9/mlflow/tensorflow/__init__.py#L503).

Comment (BenWilson2) [reply examples/tensorflow/tf2/train_predict_2.py:66]: Isn't the original data submitted to that predict method in the TF Wrapper? that explicit dict casting of {column_class: <values>} is performed within that predict method (line 524::528). 
Why is the type validation in line 530 returning a dict and not a pandas DataFrame from line 533? Shouldn't it be re-casting to a DF?

Comment (bali0019) [reply examples/tensorflow/tf2/train_predict_2.py:66]: Thanks @dbczumar @BenWilson2 ! Will put those fixes shortly including all open comments below.

Comment (bali0019) [reply examples/tensorflow/tf2/train_predict_2.py]: Committed, please let me know if you would like further changes to it.

File: mlflow/tensorflow/__init__.py (hunk 3 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: removed. thx

File: mlflow/tensorflow/__init__.py (hunk 4 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: done. thx

File: mlflow/tensorflow/__init__.py (hunk 5 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: done. thx

File: mlflow/tensorflow/__init__.py (hunk 6 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: done. thx

File: mlflow/tensorflow/__init__.py (hunk 8 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: done. thx

File: mlflow/tensorflow/__init__.py (hunk 9 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: As discussed with you offline, `GeneratorType` is going to be a separate PR that I will raise soon.

File: mlflow/tensorflow/__init__.py (hunk 10 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py:885]: done. thx

File: mlflow/tensorflow/__init__.py (hunk 12 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py:825]: the output of `input_fn' can be of type [tf.data.Dataset](https://www.tensorflow.org/api_docs/python/tf/data/Dataset) or [A tuple (features, labels)](https://www.tensorflow.org/api_docs/python/tf/estimator/Estimator#train). So it should be safe as there should be no side-effects for getting the return of those repeatedly.

File: mlflow/tensorflow/__init__.py (hunk 13 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py:766]: added to all of them.

File: mlflow/tensorflow/__init__.py (hunk 17 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py:831]: As seen [here](https://www.tensorflow.org/api_docs/python/tf/estimator/Estimator#train): "A [tf.data.Dataset](https://www.tensorflow.org/api_docs/python/tf/data/Dataset) object: Outputs of Dataset object must be a tuple (features, labels) with same constraints as below."

The first item of that tuple is 'x' entry which we need to extract as an input example.

File: mlflow/tensorflow/__init__.py (hunk 18 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: will add an example for it (please refer to this (unit test)[https://github.com/mlflow/mlflow/blob/2d57dd8ae6a12368b118dcaa25b7d925a8582956/tests/tensorflow/test_tensorflow2_autolog.py#L1388]), but if the dataset is batched, then this just extracts the first batch and then next extracts the first 5 samples from that first batch. Refer to: (steps_per_epoch)[https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit]

File: mlflow/tensorflow/__init__.py (hunk 12 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py:825]: That being said, if those datasets are generated with some custom code with transient side-effects inside `input_fn` , that could be an issue. Just to be safe, we could do call that function once and wrap the results inside another function (with no side-effects) that can be passed down to the train, but we will have the same [issue](https://github.com/mlflow/mlflow/blob/master/mlflow/utils/autologging_utils/safety.py#L869) that GeneratorType change was throwing: AssertionError around the original input_fn being mutated.

File: mlflow/tensorflow/__init__.py (hunk 14 above)
Comment (bali0019) [reply mlflow/tensorflow/__init__.py]: done and added a docstring to the predict input function. As discussed will add examples to the methods to make it easier to understand, but there are unit tests for all 'x' types that document the behavior

File: mlflow/tensorflow/__init__.py [hunk 19]
```diff
@@ -888,6 +996,99 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+
+            if Version(tensorflow.__version__) <= Version("2.1.4") and original_stop_training:
+                # For these versions, `stop_training` flag on Model is set to False
+                # This flag is used by the callback
+                # (inside ``_log_early_stop_callback_metrics``)
+                # for logging of early stop metrics. In order for
+                # that to work, need to force that flag to be True again since doing
+                # predict on that model sets `stop_training` to false for
+                # those TF versions
+                history.model.stop_training = True
```
dbczumar, line 1069: I think it would be simpler and version-agnostic to do the following:

```
original_stop_training = history.model.stop_training
model_output = history.model.predict(input_data_slice)
history.model.stop_training = original_stop_training
```

dbczumar, line 1069: I think it would be simpler and version-agnostic to do the following:

```suggestion
            original_stop_training = history.model.stop_training
            model_output = history.model.predict(input_data_slice)
            history.model.stop_training = original_stop_training
```

mlflow/tensorflow/__init__.py
```diff
@@ -888,6 +996,99 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                input_example_slice = {
+                    k: np.take(v, range(0, INPUT_EXAMPLE_SAMPLE_ROWS))
+                    for k, v in input_training_data.items()
+                }
+            else:
+                warnings.warn(
+                    "Tensorflow keras autologging only "
+                    "supports input types of: numpy.ndarray, "
+                    "dict(<key> -> numpy.ndarray), tensorflow.data.Dataset, "
+                    "or tensorflow.keras.utils.Sequence"
```
dbczumar, line 1052: Can we adjust this to make it clearer that* input example logging* is only supported for these types? The rest of the autologging functionality still works.

mlflow/tensorflow/__init__.py
```diff
@@ -888,6 +996,99 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+                input_example_slice = input_training_data[:INPUT_EXAMPLE_SAMPLE_ROWS]
+            elif isinstance(input_training_data, tensorflow.data.Dataset):
+                steps = 1
+                if history.params is not None and "steps" in history.params:
+                    steps = history.params["steps"]
+
+                def _extract_n_steps(input_example_n_steps):
+                    # Example:
+                    # when the tensorflow.data.Dataset is batched
+                    # (Total number of steps (batches of samples)),
```
dbczumar, line 1023: Still having some trouble interpreting what this means. Can we give a concrete example in the inline comment?

mlflow/tensorflow/__init__.py
```diff
@@ -888,6 +996,99 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
...
+            if isinstance(input_training_data, np.ndarray):
+                input_example_slice = input_training_data[:INPUT_EXAMPLE_SAMPLE_ROWS]
+            elif isinstance(input_training_data, tensorflow.data.Dataset):
+                steps = 1
+                if history.params is not None and "steps" in history.params:
+                    steps = history.params["steps"]
+
+                def _extract_n_steps(input_example_n_steps):
+                    # Example:
+                    # when the tensorflow.data.Dataset is batched
```
dbczumar, line 1022: Nit:

```suggestion
                    # When the tensorflow.data.Dataset is batched
```

mlflow/tensorflow/__init__.py
dbczumar: This seems okay to me for now :)

mlflow/tensorflow/__init__.py
```diff
@@ -770,6 +803,46 @@ def train(original, self, *args, **kwargs):
...
+                input_features = _extract_dict_or_tensor(input_training_data[0])
+            elif isinstance(input_training_data, tensorflow.data.Dataset):
+                train_iter = iter(input_training_data)
+                next_input = next(train_iter)[0]
+                input_features = _extract_dict_or_tensor(next_input)
+            else:
+                warnings.warn(
+                    "Tensorflow estimator doesn't support this "
+                    f"type for training: {type(input_training_data)}."
+                )
```
dbczumar, line 837: Can we update this message to make it clear that this type isn't support for *input example logging*. The rest of TensorFlow autologging capabilities will continue to work, so we don't want to confuse users.

mlflow/tensorflow/__init__.py
```diff
@@ -770,6 +803,46 @@ def train(original, self, *args, **kwargs):
...
+            return input_features
+
+        if log_models:
+            nonlocal input_example_slice
+            nonlocal tf_model
+            input_example_slice = _extract_data_from_input_fn()
+            tf_model = result
```
dbczumar, line 844: Can we update the logic here such that we only collect an input example slice if `log_input_examples` is `True`? That way, we aren't performing extra computations in most cases, and any issues associated with input example collection are only encountered when users explicitly call autolog with `log_input_examples=True`.

mlflow/tensorflow/__init__.py
dbczumar: Thanks! Can we leave an inline comment explaining this?

mlflow/tensorflow/__init__.py
dbczumar: Thanks!

mlflow/tensorflow/__init__.py
```diff
@@ -753,6 +763,29 @@ def autolog(
         warnings.warn("Could not log to MLflow. TensorFlow versions below 1.12 are not supported.")
         return
 
+    def _extract_tensor_as_numpy_input_example_slice(next_input):
```
dbczumar, line 766: Can we move these utility methods into a separate `mlflow/tensorflow/utils.py` module for readability?

mlflow/tensorflow/__init__.py
```diff
@@ -740,6 +740,16 @@ def autolog(
...
     atexit.register(_flush_queue)
 
+    if log_model_signatures:
+        warnings.warn(
+            "When 'log_model_signatures' is set to True, "
+            "it can result in output types that differ from the input type"
+            " for predict using PyFunc implementation. "
+            "For example, if pandas DF is used for prediction input, "
+            "the output could be ndarray or dict of (str -> ndarray) based "
+            "on what the signature of the model was inferred."
```
dbczumar, line 750: Can we move this warning to `mlflow.tensorflow.log_model()`? This applies beyond autologging. I think the following text may be clearer:

```
            " The pyfunc inference behavior of TensorFlow models logged with signatures differs"
            " from the behavior of TensorFlow models logged without signatures. Specifically, when a"
            " signature is present, passing a Pandas DataFrame as input to the pyfunc `predict()` API"
            " produces an `ndarray` (for single-output models) or a dictionary of `str -> ndarray`:"
            " (for multi-output models). In contrast, when a signature is *not* present, `predict()` produces"
            " a Pandas DataFrame output in response to a Pandas DataFrame input."
```
            
            
Can we also use `_logger.warning()` rather than `warnings.warn()`?

dbczumar: This applies to `keras` models as well, right? In that case, we should also add the warning to `mlflow.keras.log_model()`

mlflow/tensorflow/__init__.py
```diff
@@ -753,6 +763,29 @@ def autolog(
...
+
+        input_feature_slice = None
+        if isinstance(next_input, tensorflow.Tensor):
+            input_feature_slice = np.take(next_input.numpy(), range(0, INPUT_EXAMPLE_SAMPLE_ROWS))
+        elif isinstance(next_input, np.ndarray):
+            input_feature_slice = np.take(next_input, range(0, INPUT_EXAMPLE_SAMPLE_ROWS))
+        else:
+            warnings.warn(
+                f"Tensorflow estimator doesn't support '{type(next_input)}' type for features"
+            )
```
dbczumar, line 783: Can we update this message to make it clear that this type isn't support for *input example logging*? The rest of TensorFlow autologging capabilities will continue to work, so we don't want to confuse users.

mlflow/tracking/fluent.py
```diff
@@ -1632,7 +1660,11 @@ def setup_autologging(module):
             if prev_config and not prev_config.get(CONF_KEY_IS_GLOBALLY_CONFIGURED, False):
                 return
 
-            autologging_params = get_autologging_params(autolog_fn)
+            autologging_params = (
+                _get_tensorflow_autologging_params()
+                if module_name == "tensorflow"
+                else get_autologging_params(autolog_fn)
+            )
```
dbczumar, line 1667: For maintainability, it would be nice if we could scope this special-case logic inside the `mlflow.tensorflow.autolog()` module. One way to do this would be to leverage the value of `AUTOLOGGING_INTEGRATIONS["tensorflow"].get(CONF_KEY_IS_GLOBALLY_CONFIGURED)`, similar to https://github.com/mlflow/mlflow/blob/855881f93703b15ffe643003fb4d7c84f0ec2502/mlflow/tracking/fluent.py#L1631-L1632.

During TF autologging, we can check to see if autologging for TensorFlow is globally configured and, if so, default `log_model_signatures` to `False`.

dbczumar: Implemented this here: https://github.com/mlflow/mlflow/pull/5652. Feel free to rebase / merge master once #5652 is merged.

BenWilson2:
#5652 is merged. Please pull those test conf changes in :)

mlflow/tensorflow/__init__.py
BenWilson2: this change is in master now.

bali0019:
@dbczumar @BenWilson2  I pushed fixes in the latest commit that addresses all comments. please let me know if there are additional changes needed.

mlflow/tensorflow/__init__.py
```diff
@@ -752,6 +751,19 @@ def autolog(
...
+            " The pyfunc inference behavior of TensorFlow models logged "
+            "with signatures differs from the behavior of TensorFlow "
+            "models logged without signatures. Specifically, when a "
+            "signature is present, passing a Pandas DataFrame as "
+            "input to the pyfunc `predict()` API produces an `ndarray` "
+            "(for single-output models) or a dictionary of `str -> ndarray`:"
+            " (for multi-output models). In contrast, when a signature "
+            "is *not* present, `predict()` produces"
+            " a Pandas DataFrame output in response to a Pandas DataFrame input."
+        )
```
dbczumar, line 765: @bali0019 Can we move this warning to `mlflow.keras.log_model()` and `mlflow.tensorflow.log_model()`? This is a property of model logging with signatures, not just autologging.

mlflow/tensorflow/__init__.py
```diff
@@ -818,9 +866,45 @@ def log_model_without_starting_new_run():
...
+                            """
+                            input_slices = input_example_slice
+                            if isinstance(input_example_slice, dict):
+                                input_slices = {
+                                    k: tensorflow.convert_to_tensor(v)
+                                    for k, v in input_example_slice.items()
+                                }
+                            return tensorflow.data.Dataset.from_tensor_slices(input_slices).batch(1)
+
+                        predicted_values = list(tf_model.predict(predict_input_fn))
```
dbczumar, line 888: Do we need to save `tf_model` as a global reference, or can we call `self.predict()`?

mlflow/tensorflow/__init__.py
```diff
@@ -818,9 +866,45 @@ def log_model_without_starting_new_run():
...
+
+                        input_example, signature = resolve_input_example_and_signature(
+                            lambda: input_example_slice,
+                            lambda in_ex: infer_signature(input_example_slice, predicted_values[0]),
+                            log_input_examples,
+                            log_model_signatures,
```
dbczumar, line 894: We also need to apply the logic from https://github.com/mlflow/mlflow/pull/5510/files#diff-e8adca6c94982c5c5bf934b8f9b2226c374a1c97f563eba8b7d1894b1af3a7bbR1044-R1060 so that `log_model_signatures=True` only if `mlflow.tensorflow.autolog(log_model_signatures=True)` was called; it should still be `False` here if `mlflow.autolog(log_model_signatures=True)` was called for backwards compatibility.

Can we extract the logic from https://github.com/mlflow/mlflow/pull/5510/files#diff-e8adca6c94982c5c5bf934b8f9b2226c374a1c97f563eba8b7d1894b1af3a7bbR1044-R1060 into a common function that we can call?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +57,110 @@ def on_epoch_end(self, epoch, logs=None):
...
-        if self.log_models:
-            registered_model_name = get_autologging_config(
-                mlflow.tensorflow.FLAVOR_NAME, "registered_model_name", None
-            )
-            mlflow.keras.log_model(
-                self.model, artifact_path="model", registered_model_name=registered_model_name
-            )
+
+def extract_sample_tensor_as_numpy_input_example_slice(
+    next_input: Union[tensorflow.Tensor, np.ndarray]
```
dbczumar, line 62:
```suggestion
    input_features: Union[tensorflow.Tensor, np.ndarray]
```

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +57,110 @@ def on_epoch_end(self, epoch, logs=None):
...
+        input_feature_slice = next_input[0:INPUT_EXAMPLE_SAMPLE_ROWS]
+    else:
+        warnings.warn(
+            f"Tensorflow estimator doesn't support '{type(next_input)}' type for features"
+        )
+    return input_feature_slice
+
+
+def extract_sample_dict_or_tensor(
+    next_input: Union[Dict[str, Union[tensorflow.Tensor, np.ndarray]], tensorflow.Tensor]
```
dbczumar, line 98: Why do we include `tensorflow.Tensor` as a valid input here if there's a separate function that covers this case? It looks like we delegate to that method in the implementation of this method anyway: https://github.com/mlflow/mlflow/pull/5510/files#diff-5cf5c34a96d4329cbe33c8f4e8ab769cbaa5d176042838a9385554f0f36ba03cR126. Can the calling code call `extract_sample_tensor_as_numpy_input_example_slice` instead?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +57,110 @@ def on_epoch_end(self, epoch, logs=None):
...
+) -> Union[Dict[str, np.ndarray], np.ndarray]:
```
dbczumar, line 99: Why do we include `tensorflow.Tensor` as a valid input here if there's a separate function that covers this case? It looks like we delegate to that method in the implementation of this method anyway: https://github.com/mlflow/mlflow/pull/5510/files#diff-5cf5c34a96d4329cbe33c8f4e8ab769cbaa5d176042838a9385554f0f36ba03cR126. Can the calling code call `extract_sample_tensor_as_numpy_input_example_slice` instead?

```suggestion
def extract_input_example_from_tensor_dict(
    input_features: Dict[str, Union[tensorflow.Tensor, np.ndarray]]
) -> Dict[str, np.ndarray]:
```

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +57,110 @@ def on_epoch_end(self, epoch, logs=None):
...
-            registered_model_name = get_autologging_config(
-                mlflow.tensorflow.FLAVOR_NAME, "registered_model_name", None
-            )
-            mlflow.keras.log_model(
-                self.model, artifact_path="model", registered_model_name=registered_model_name
-            )
+
+def extract_sample_tensor_as_numpy_input_example_slice(
+    next_input: Union[tensorflow.Tensor, np.ndarray]
+) -> np.ndarray:
```
dbczumar, line 63:
```suggestion
def extract_input_example_from_tensor_or_ndarray(
    input_features: Union[tensorflow.Tensor, np.ndarray]
) -> np.ndarray:
```

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +57,110 @@ def on_epoch_end(self, epoch, logs=None):
...
+            k: extract_sample_tensor_as_numpy_input_example_slice(v) for k, v in next_input.items()
+        }
+    elif isinstance(next_input, tensorflow.Tensor):
+        sliced_data_as_numpy = extract_sample_tensor_as_numpy_input_example_slice(next_input)
+    return sliced_data_as_numpy
+
+
+def extract_sample_features_from_batched_tf_dataset(
+    dataset: tensorflow.data.Dataset,
+) -> Union[np.ndarray, Dict[str, np.ndarray]]:
```
dbczumar, line 132:
```suggestion
def extract_input_example_from_batched_tf_dataset(
    dataset: tensorflow.data.Dataset,
) -> Union[np.ndarray, Dict[str, np.ndarray]]:
```

mlflow/tensorflow/__init__.py
```diff
@@ -782,6 +797,39 @@ def train(original, self, *args, **kwargs):
...
+            elif isinstance(input_training_data, tensorflow.data.Dataset):
+                input_features = extract_sample_features_from_batched_tf_dataset(
+                    input_training_data
+                )
+            else:
+                warnings.warn(
+                    "Tensorflow autologging doesn't support this "
+                    f"type for input example: {type(input_training_data)}."
+                )
+            return input_features
```
dbczumar, line 825: Can we move this function into `_autolog.py` and have it accept training data input as an argument? Then, can we make the methods it relies on private by prefixing with `_`?  E.g. `extract_sample_dict_or_tensor` -> `_extract_sample_dict_or_tensor`.

mlflow/tensorflow/__init__.py
```diff
@@ -818,9 +866,45 @@ def log_model_without_starting_new_run():
...
+                                    k: tensorflow.convert_to_tensor(v)
+                                    for k, v in input_example_slice.items()
+                                }
+                            return tensorflow.data.Dataset.from_tensor_slices(input_slices).batch(1)
+
+                        predicted_values = list(tf_model.predict(predict_input_fn))
+
+                        input_example, signature = resolve_input_example_and_signature(
+                            lambda: input_example_slice,
+                            lambda in_ex: infer_signature(input_example_slice, predicted_values[0]),
```
dbczumar, line 892: If `input_example_slice` is `None`, I think this will print a strange error - e.g. "cannot concatenate `NoneType` and `string`." This can be `None` if `_extract_data_from_input_fn` doesn't know how to handle the data type.

Instead of passing `lambda: input_example_slice`, can we pass a function that either returns `input_example_slice` if it is `not None` or throws?

E.g.

```
def get_input_example_slice():
    if input_example_slice is None:
        raise MlflowException(
            "TensorFlow autologging can only log input examples for inputs of the following types: ..."
        )
    return input_example_slice

```

mlflow/tensorflow/__init__.py
```diff
@@ -782,6 +797,39 @@ def train(original, self, *args, **kwargs):
...
+            if isinstance(input_training_data, tuple):
+                input_features = extract_sample_dict_or_tensor(input_training_data[0])
+            elif isinstance(input_training_data, tensorflow.data.Dataset):
+                input_features = extract_sample_features_from_batched_tf_dataset(
+                    input_training_data
+                )
+            else:
+                warnings.warn(
+                    "Tensorflow autologging doesn't support this "
+                    f"type for input example: {type(input_training_data)}."
```
dbczumar, line 823: Can we remove this warning and instead rely on the function passed to `resolve_input_example_and_signature` to properly print a warning? See https://github.com/mlflow/mlflow/pull/5510/files#r851523221 for more context.

mlflow/tensorflow/__init__.py
```diff
@@ -782,6 +797,39 @@ def train(original, self, *args, **kwargs):
...
+                )
+            return input_features
+
+        if log_input_examples:
+            nonlocal input_example_slice
+            nonlocal tf_model
+            input_example_slice = _extract_data_from_input_fn()
+            tf_model = result
```
dbczumar, line 831: We may not need to save this to a global var; see https://github.com/mlflow/mlflow/pull/5510/files#r851515415

mlflow/tensorflow/__init__.py
```diff
@@ -782,6 +797,39 @@ def train(original, self, *args, **kwargs):
 
         result = original(self, *args, **kwargs)
 
+        def _extract_data_from_input_fn():
```
dbczumar, line 800: Can we move this function into `_autolog.py` and have it accept training data input as an argument? Then, can we make the methods it relies on private by prefixing with `_`?  E.g. `extract_sample_dict_or_tensor` -> `_extract_sample_dict_or_tensor`.

mlflow/tensorflow/__init__.py
```diff
@@ -900,6 +984,94 @@ def _log_early_stop_callback_metrics(callback, history, metrics_logger):
         if metric_key is not None:
             metrics_logger.record_metrics(restored_metrics, stopped_epoch + 1)
 
+    def _log_keras_model(history, args):
+        def _get_input_data_slice():
```
dbczumar, line 988: Similar to https://github.com/mlflow/mlflow/pull/5510/files#r851523774, can we move this to `_autolog.py`?

mlflow/keras.py
```diff
@@ -401,6 +401,18 @@ def log_model(
         with mlflow.start_run() as run:
             mlflow.keras.log_model(keras_model, "models")
     """
+    if signature is not None:
+        warnings.warn(
+            " The pyfunc inference behavior of keras model logged "
```
dbczumar, line 406:
```suggestion
            " The pyfunc inference behavior of Keras models logged "
```

mlflow/keras.py
```diff
@@ -401,6 +401,18 @@ def log_model(
...
+            "with signatures differs from the behavior of keras "
+            "model logged without signatures. Specifically, when a "
```
dbczumar, line 408:
```suggestion
            "with signatures differs from the behavior of Keras "
            "models logged without signatures. Specifically, when a "
```

mlflow/keras.py
```diff
@@ -401,6 +401,18 @@ def log_model(
...
+            "signature is present, passing a Pandas DataFrame as "
+            "input to the pyfunc `predict()` API produces an `ndarray` "
+            "(for single-output models) or a dictionary of `str -> ndarray`:"
+            " (for multi-output models). In contrast, when a signature "
```
dbczumar, line 412:
```suggestion
            "(for single-output models) or a dictionary of `str -> ndarray`: "
            "(for multi-output models). In contrast, when a signature "
```

mlflow/keras.py
```diff
@@ -401,6 +401,18 @@ def log_model(
...
+            "is *not* present, `predict()` produces"
+            " a Pandas DataFrame output in response to a Pandas DataFrame input."
+        )
```
dbczumar, line 415:
```suggestion
            " The pyfunc inference behavior of Keras models logged "
            "with signatures differs from the behavior of Keras "
            "models logged without signatures. Specifically, when a "
            "signature is present, passing a Pandas DataFrame as "
            "input to the pyfunc `predict()` API produces an `ndarray` "
            "(for single-output models) or a dictionary of `str -> ndarray`: "
            "(for multi-output models). In contrast, when a signature "
            "is *not* present, `predict()` produces "
            "a Pandas DataFrame output in response to a Pandas DataFrame input."
        )
```

mlflow/tensorflow/__init__.py
```diff
@@ -195,6 +193,18 @@ def log_model(
...
+        warnings.warn(
+            " The pyfunc inference behavior of TensorFlow models logged "
+            "with signatures differs from the behavior of TensorFlow "
+            "models logged without signatures. Specifically, when a "
+            "signature is present, passing a Pandas DataFrame as "
+            "input to the pyfunc `predict()` API produces an `ndarray` "
+            "(for single-output models) or a dictionary of `str -> ndarray`:"
+            " (for multi-output models). In contrast, when a signature "
+            "is *not* present, `predict()` produces"
+            " a Pandas DataFrame output in response to a Pandas DataFrame input."
```
dbczumar, line 206:
```suggestion
            " The pyfunc inference behavior of TensorFlow models logged "
            "with signatures differs from the behavior of TensorFlow "
            "models logged without signatures. Specifically, when a "
            "signature is present, passing a Pandas DataFrame as "
            "input to the pyfunc `predict()` API produces an `ndarray` "
            "(for single-output models) or a dictionary of `str -> ndarray`: "
            "(for multi-output models). In contrast, when a signature "
            "is *not* present, `predict()` produces "
            "a Pandas DataFrame output in response to a Pandas DataFrame input."
```

mlflow/tensorflow/__init__.py
```diff
@@ -765,6 +775,29 @@ def autolog(
         warnings.warn("Could not log to MLflow. TensorFlow versions below 1.12 are not supported.")
         return
 
+    input_example_slice = None
+
+    def _log_model_signatures():
```
dbczumar, line 780:
```suggestion
    def _should_log_model_signatures():
```

mlflow/tensorflow/__init__.py
dbczumar: @bali0019 Can we address this comment?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,171 @@ def on_epoch_end(self, epoch, logs=None):
...
+            f" {type(input_training_data)}. TensorFlow autologging can"
+            " only log input examples and model signatures for the following"
+            " input types: tuple, tensorflow.data.Dataset (TensorFlow >= 2.1.0 required)"
+            " or tensorflow.Tensor",
+            INVALID_PARAMETER_VALUE,
+        )
+    return input_features
+
+
+def get_input_data_slice(input_training_data):
```
dbczumar, line 197:
```suggestion
def get_tf_keras_input_data_slice(input_training_data):
```

mlflow/tensorflow/__init__.py
```diff
@@ -782,6 +815,12 @@ def train(original, self, *args, **kwargs):
 
         result = original(self, *args, **kwargs)
 
+        if log_input_examples:
+            nonlocal input_example_slice
+            from mlflow.tensorflow._autolog import extract_data_from_tf_input_fn
+
+            input_example_slice = extract_data_from_tf_input_fn(kwargs.get("input_fn"))
```
dbczumar, line 822: Looking at the implementation of this function, it seems to throw if the data type is unsupported. We should only throw when the example slice is fetched within `resolve_input_example_and_signature`. See my comment on https://github.com/mlflow/mlflow/pull/5510/files#r851523221. Can we update this based on my recommendation in https://github.com/mlflow/mlflow/pull/5510/files#r851523221?

dbczumar (commented):
@bali0019 This is looking fantastic! After the remaining small comments are addressed, this will be ready for merge!

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+    limited_df_iter = list(dataset.take(INPUT_EXAMPLE_SAMPLE_ROWS))
+    first_batch = limited_df_iter[0]
+    if isinstance(first_batch, tuple):
+        features = first_batch[0]
+        if isinstance(features, dict):
+            return _extract_sample_numpy_dict(features)
+        elif isinstance(features, (np.ndarray, tensorflow.Tensor)):
+            return _extract_input_example_from_tensor_or_ndarray(features)
+        raise TypeError(f"Unsupported type for features of type: {type(features)}")
+    raise TypeError(f"Unsupported type for dataset batch slice of type: {type(first_batch)}")
```
dbczumar, line 162: @bali0019 We shouldn't throw here; this will cause autologging to short-circuit.

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+
+def _extract_input_example_from_tensor_or_ndarray(
+    input_features: Union[tensorflow.Tensor, np.ndarray]
+) -> np.ndarray:
+    """
+    Extracts first `INPUT_EXAMPLE_SAMPLE_ROWS` from the next_input, which can either be of
+    numpy array or tensor type.
+
+    :param input_features: an input of type `np.ndarray` or `tensorflow.Tensor`
+    :return: a slice (of limit `INPUT_EXAMPLE_SAMPLE_ROWS`)  of the input of type `np.ndarray`
```
dbczumar, line 71: Can we update this to also say that we return `None` if the type of `input_features` is unsupported?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+
+    input_feature_slice = None
+    if isinstance(input_features, tensorflow.Tensor):
+        input_feature_slice = input_features.numpy()[0:INPUT_EXAMPLE_SAMPLE_ROWS]
+    elif isinstance(input_features, np.ndarray):
+        input_feature_slice = input_features[0:INPUT_EXAMPLE_SAMPLE_ROWS]
+    else:
+        warnings.warn(
+            f"Tensorflow estimator doesn't support '{type(input_features)}' type for features"
+        )
```
dbczumar, line 95: No need to warn here; we'll warn during `resolve_input_example_and_signature`

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+
+def _extract_sample_numpy_dict(
+    input_numpy_features_dict: Dict[str, np.ndarray]
+) -> Union[Dict[str, np.ndarray], np.ndarray]:
+    """
+    Extracts `INPUT_EXAMPLE_SAMPLE_ROWS` sample from next_input
+    as numpy array of dict(str -> ndarray) type.
+
+    :param input_numpy_features_dict: A tensor or numpy array
+    :return:a slice (limit `INPUT_EXAMPLE_SAMPLE_ROWS`)  of the input of same type as next_input
```
dbczumar, line 107: Can we update this to also say that we return `None` if the type of `input_numpy_features_dict` is unsupported?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+    elif isinstance(input_features, np.ndarray):
+        input_feature_slice = input_features[0:INPUT_EXAMPLE_SAMPLE_ROWS]
+    else:
+        warnings.warn(
+            f"Tensorflow estimator doesn't support '{type(input_features)}' type for features"
+        )
+    return input_feature_slice
+
+
+def _extract_sample_numpy_dict(
```
dbczumar, line 99:
```suggestion
def _extract_input_example_from_numpy_dict(
```

dbczumar: @bali0019 Previously, we also had code to support `Dict[str, Tensor]`, right? Why did we remove it? Is it unnecessary?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+    if isinstance(input_numpy_features_dict, dict):
+        sliced_data_as_numpy = {
+            k: _extract_input_example_from_tensor_or_ndarray(v)
+            for k, v in input_numpy_features_dict.items()
+        }
+    return sliced_data_as_numpy
+
+
+def _extract_input_example_from_batched_tf_dataset(
+    dataset: tensorflow.data.Dataset,
```
dbczumar, line 127: Can we update this to also say that we return `None` if the contents of `dataset` are unsupported?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+        features = first_batch[0]
+        if isinstance(features, dict):
+            return _extract_sample_numpy_dict(features)
+        elif isinstance(features, (np.ndarray, tensorflow.Tensor)):
+            return _extract_input_example_from_tensor_or_ndarray(features)
+        raise TypeError(f"Unsupported type for features of type: {type(features)}")
+    raise TypeError(f"Unsupported type for dataset batch slice of type: {type(first_batch)}")
+
+
+def extract_data_from_tf_input_fn(input_fn):
```
dbczumar, line 165:
```suggestion
def extract_input_example_from_tf_input_fn(input_fn):
```

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+    dataset: tensorflow.data.Dataset,
+) -> Union[np.ndarray, Dict[str, np.ndarray]]:
+    """
+    Extracts sample feature tensors from the input dataset as numpy array.
+    Input Dataset's tensors must contain tuple of (features, labels) that are
+    used for tensorflow/keras train or fit methods
+
+
+    :param dataset: a tensorflow batched/unbatched dataset representing tuple of (features, labels)
+    :return: a numpy array of length `INPUT_EXAMPLE_SAMPLE_ROWS`
```
dbczumar, line 136: Can we update this to also say that we return `None` if the contents of `dataset` are unsupported?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+    raise TypeError(f"Unsupported type for dataset batch slice of type: {type(first_batch)}")
+
+
+def extract_data_from_tf_input_fn(input_fn):
+    """
+    Extracts sample data from dict (str -> ndarray),
+    ``tensorflow.Tensor`` or ``tensorflow.data.Dataset`` type.
+
+    :param input_fn: Tensorflow's input function used for train method
+    :return: a slice (of limit ``mlflow.utils.autologging_utils.INPUT_EXAMPLE_SAMPLE_ROWS``)
```
dbczumar, line 171: Can we update this to also say that we return `None` if the contents of the dataset retrieved by calling `input_fn` are unsupported?

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+        if isinstance(features, dict):
+            input_features = _extract_sample_numpy_dict(features)
+        elif isinstance(features, (np.ndarray, tensorflow.Tensor)):
+            input_features = _extract_input_example_from_tensor_or_ndarray(features)
+    elif isinstance(input_training_data, tensorflow.data.Dataset):
+        input_features = _extract_input_example_from_batched_tf_dataset(input_training_data)
+    return input_features
+
+
+def get_tf_keras_input_data_slice(input_training_data):
```
dbczumar, line 188:
```suggestion
def get_tf_keras_input_example(input_training_data):
```

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+    else:
+        raise MlflowException(
+            "Cannot log input example or model signature for input with type"
+            f" {type(input_training_data)}. TensorFlow Keras autologging can"
+            " only log input examples and model signatures for the following"
+            " input types: numpy.ndarray, dict[string -> numpy.ndarray],"
+            " tensorflow.keras.utils.Sequence, and"
+            " tensorflow.data.Dataset (TensorFlow >= 2.1.0 required)",
+            INVALID_PARAMETER_VALUE,
+        )
```
dbczumar, line 218: Can we move this exception into the part of the code in `mlflow.tensorflow.autolog()` where the input example / model signature is resolved? This utility method is just responsible for trying to get the data for the caller. It's up to the caller how to handle the case where the data is `None`, so the caller should throw.

mlflow/tensorflow/_autolog.py
```diff
@@ -52,11 +59,162 @@ def on_epoch_end(self, epoch, logs=None):
...
+
+
+def get_tf_keras_input_data_slice(input_training_data):
+    """
+    Generates a sample ndarray or dict (str -> ndarray)
+    from the input type 'x' for keras ``fit`` or ``fit_generator``
+
+    :param input_training_data: Keras input function used for ``fit`` or
+                                ``fit_generator`` methods
+    :return: a slice of type ndarray or
```
dbczumar, line 195: Following from https://github.com/mlflow/mlflow/pull/5510/files#r852379015, can we update this to also say that we return `None` if the contents of `input_training_data` are unsupported?

dbczumar: @bali0019 Can we also add a simple test case for unsupported inputs to make sure that we return `None` and don't throw?

mlflow/tensorflow/_autolog.py
dbczumar: @bali0019 Can we also add a simple test case for unsupported inputs to make sure that we return `None` and don't throw?

mlflow/keras.py
```diff
@@ -401,6 +401,18 @@ def log_model(
         with mlflow.start_run() as run:
             mlflow.keras.log_model(keras_model, "models")
     """
+    if signature is not None:
+        warnings.warn(
+            " The pyfunc inference behavior of Keras models logged "
```
dbczumar, line 406:
```suggestion
            "The pyfunc inference behavior of Keras models logged "
```

mlflow/tensorflow/__init__.py
```diff
@@ -195,6 +193,18 @@ def log_model(
     :return: A :py:class:`ModelInfo <mlflow.models.model.ModelInfo>` instance that contains the
              metadata of the logged model.
     """
+    if signature is not None:
+        warnings.warn(
+            " The pyfunc inference behavior of TensorFlow models logged "
```
dbczumar, line 198:
```suggestion
            "The pyfunc inference behavior of TensorFlow models logged "
```

dbczumar (approved):
LGTM! Thanks so much, @bali0019 ! Let's resolve two small spacing comments, then I'll go ahead and merge :)

Issue #5501, opened 2022-03-16
[FR] Add MLflow version to the MLModel spec


dbczumar:
Thank you for submitting a feature request. **Before proceeding, please review MLflow's [Issue Policy for feature requests](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md#feature-requests) and the [MLflow Contributing Guide](https://github.com/mlflow/mlflow/blob/master/CONTRIBUTING.rst)**.

**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [ ] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [X] No. I cannot contribute this feature at this time.

## Proposal Summary

Add an `mlflow_version` field to the MLModel spec that is automatically populated with the current MLflow version when an MLflow Model is saved / logged.

## Motivation
- What is the use case for this feature? When saving an MLflow Model, it is helpful to know what version of MLflow was used to save the model for backwards compatibility & validation purposes.
- Why is this use case valuable to support for MLflow users in general? ^
- Why is this use case valuable to support for your project(s) or organization? ^
- Why is it currently difficult to achieve this use case? MLflow version information isn't available anywhere in the serialized MLflow Model format.

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [X] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interfaces
- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
[...]

PR #5515, opened 2022-03-20
Implement issue-5501 - add mlflow_version info to logged model

r3stl355:
Signed-off-by: Nikolay Ulmasov <ulmasov@hotmail.com>

Implement issue #5501 - Add an mlflow_version field to the MLModel spec that is automatically populated with the current MLflow version when an MLflow Model is saved / logged.

Models logged with previous versions will have `mlflow_version=None` when loaded from a file.

## How is this patch tested?

Extended existing tests and added a test to load previously logged models without mlflow_version info

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Users can check the version of MLFlow the model with logged with. 

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/models/model.py
+++ b/mlflow/models/model.py
@@ -26,6 +26,7 @@ _LOG_MODEL_METADATA_WARNING_TEMPLATE = (
     "mlflow server via REST, consider upgrading the server version to MLflow "
     "1.7.0 or above."
 )
+_MLFLOW_VERSION_KEY = "mlflow_version"
 
 
 class ModelInfo(NamedTuple):
@@ -94,7 +95,7 @@ class Model:
         signature=None,  # ModelSignature
         saved_input_example_info: Dict[str, Any] = None,
         model_uuid: Union[str, Callable, None] = lambda: uuid.uuid4().hex,
-        mlflow_version: Union[str, Callable, None] = lambda: mlflow.version.VERSION,
+        mlflow_version: Union[str, None] = mlflow.version.VERSION,
         **kwargs,
     ):
         # store model id instead of run_id and path to avoid confusion when model gets exported
@@ -107,7 +108,7 @@ class Model:
         self.signature = signature
         self.saved_input_example_info = saved_input_example_info
         self.model_uuid = model_uuid() if callable(model_uuid) else model_uuid
-        self.mlflow_version = mlflow_version() if callable(mlflow_version) else mlflow_version
+        self.mlflow_version = mlflow_version
         self.__dict__.update(kwargs)
 
     def __eq__(self, other):
@@ -200,9 +201,8 @@ class Model:
 
         # Do not save mlflow_version attribute if it's not defined
         info = self.to_dict()
-        mlflow_version_key = "mlflow_version"
-        if not self.mlflow_version and mlflow_version_key in info:
-            info.pop(mlflow_version_key)
+        if not self.mlflow_version and _MLFLOW_VERSION_KEY in info:
+            info.pop(_MLFLOW_VERSION_KEY)
 
         return yaml.safe_dump(info, stream=stream, default_flow_style=False)
 
@@ -239,8 +239,8 @@ class Model:
         if "model_uuid" not in model_dict:
             model_dict["model_uuid"] = None
 
-        if "mlflow_version" not in model_dict:
-            model_dict["mlflow_version"] = None
+        if _MLFLOW_VERSION_KEY not in model_dict:
+            model_dict[_MLFLOW_VERSION_KEY] = None
 
         return cls(**model_dict)
 
--- a/mlflow/models/model.py
+++ b/mlflow/models/model.py
@@ -194,17 +194,13 @@ class Model:
             res["signature"] = self.signature.to_dict()
         if self.saved_input_example_info is not None:
             res["saved_input_example_info"] = self.saved_input_example_info
+        if self.mlflow_version is None and _MLFLOW_VERSION_KEY in res:
+            res.pop(_MLFLOW_VERSION_KEY)
         return res
 
     def to_yaml(self, stream=None):
         """Write the model as yaml string."""
-
-        # Do not save mlflow_version attribute if it's not defined
-        info = self.to_dict()
-        if not self.mlflow_version and _MLFLOW_VERSION_KEY in info:
-            info.pop(_MLFLOW_VERSION_KEY)
-
-        return yaml.safe_dump(info, stream=stream, default_flow_style=False)
+        return yaml.safe_dump(self.to_dict(), stream=stream, default_flow_style=False)
 
     def __str__(self):
         return self.to_yaml()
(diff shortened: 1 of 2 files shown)


mlflow/models/model.py
```diff
@@ -202,6 +208,11 @@ def to_json(self):
 
     def save(self, path):
         """Write the model as a local YAML file."""
+
+        # Saved model must have mlflow_version set
```
BenWilson2, line 212: This is set in the constructor. Any reason to validate here?

mlflow/models/model.py
```diff
@@ -175,6 +180,7 @@ def get_model_info(self):
             saved_input_example_info=self.saved_input_example_info,
             signature_dict=self.signature.to_dict() if self.signature else None,
             utc_time_created=self.utc_time_created,
+            mlflow_version=self.mlflow_version,
```
BenWilson2, line 183: to make this backward compatible, perhaps add in a conditional check?

```suggestion
            mlflow_version=self.mlflow_version if self.mlflow_version else None,
```

tests/models/test_model.py
```diff
@@ -110,6 +110,8 @@ def test_model_log():
         assert isinstance(loaded_example, pd.DataFrame)
         assert loaded_example.to_dict(orient="records")[0] == input_example
 
+        assert loaded_model.mlflow_version == mlflow.version.VERSION
```
BenWilson2, line 113: `from packaging.version import Version`

```suggestion
        assert Version(loaded_model.mlflow_version) == Version(mlflow.version.VERSION)
```

mlflow/models/model.py
r3stl355: This is code that runs within the instance that will always have the mlflow_version property, even if it's loaded from an older file in which case it will be just None so no real need to add that conditional check. Happy to add it though if you think it's really necessary

mlflow/models/model.py
r3stl355: but I don't like that option either. My view is that model loaded from an older file must have mlflow_version set to None but what happens when it saved - that's the decision that is needed. Setting a value on save creates a side effect but not doing so is against the new save approach. I thoughts?

tests/models/test_model.py
```diff
@@ -110,6 +110,8 @@ def test_model_log():
         assert isinstance(loaded_example, pd.DataFrame)
         assert loaded_example.to_dict(orient="records")[0] == input_example
```
BenWilson2, line 112: Import statement to add:
``` py
from packaging.version import Version
```

mlflow/models/model.py
r3stl355: Yes, this business of loading a model from an older version file then saving it isn't straightforward. Arguably, if the model is saved into the same file as It was loaded from then I would probably not expect a new attribute to appear there (e.g. not to break some other tooling that may depend on that location). In contrast, if the model is saved to a different file that it was loaded from then perhaps adding a new attribute would be appropriate. Unfortunately, we don't capture the information about the file location on load so implementing this would not be possible without adding that info on load which sounds more like a different ticket/issue. 

How about leaving it to the user to explicitly set the mlflow_version if they load the model from an older version? And if they don't set the value then the attribute is not saved at all?

BenWilson2: we don't persist the loaded location on load, but the classmethod `load()` uses the uri as it's singular argument. Setting an instance param of that location could give you the information needed to enforce that backwards compatibility. 
I'll bring it up at standup and see what the team thinks about this behavior for older versions. Stay tuned!

r3stl355: Cool, thanks, let me know what the team thinks. In the meantime, I'm going to submit a version that does not change the attribute when model is loaded from the older file

BenWilson2: We have a unanimous decision: "Let's not apply any additional information to a pre-mlflow-version-attribute to models that are loaded (or saved). So long as all new models receive this and handling a missing value in the artifact to maintain backwards compatibility are maintained, we're good to go."

mlflow/models/model.py
```diff
@@ -191,7 +197,14 @@ def to_dict(self):
 
     def to_yaml(self, stream=None):
         """Write the model as yaml string."""
-        return yaml.safe_dump(self.to_dict(), stream=stream, default_flow_style=False)
+
+        # Do not save mlflow_version attribute if it's not defined
+        info = self.to_dict()
+        mlflow_version_key = "mlflow_version"
```
BenWilson2, line 203: Since this key name is referenced in more than one place(ref: line 243), perhaps we could create a module private constant?
`_MLFLOW_VERSION_KEY = "mlflow_version"` towards the top and reference that throughout the module.

dbczumar: +1

mlflow/models/model.py
```diff
@@ -191,7 +197,14 @@ def to_dict(self):
 
     def to_yaml(self, stream=None):
         """Write the model as yaml string."""
-        return yaml.safe_dump(self.to_dict(), stream=stream, default_flow_style=False)
+
+        # Do not save mlflow_version attribute if it's not defined
+        info = self.to_dict()
```
dbczumar, line 202: Can we move the logic for discarding `mlflow_version` if it's `None` into `to_dict()`?

mlflow/models/model.py
```diff
@@ -91,6 +94,7 @@ def __init__(
         signature=None,  # ModelSignature
         saved_input_example_info: Dict[str, Any] = None,
         model_uuid: Union[str, Callable, None] = lambda: uuid.uuid4().hex,
+        mlflow_version: Union[str, Callable, None] = lambda: mlflow.version.VERSION,
```
dbczumar, line 97: No need for a callable here since the version doesn't need to be dynamically generated (unlike UUID).

```suggestion
        mlflow_version: Union[str, None] = lambda: mlflow.version.VERSION,
```

dbczumar (approved):
@r3stl355 Looks great! LGTM once remaining comments have been addressed!

mlflow/models/model.py
r3stl355: Will change

mlflow/models/model.py
r3stl355: Surely can do but my thinking was that a new version of model code (or some external code using it) "knows" object has a property named `mlflow_version` so logically may also expect the same property in the model info and call it without checking if it exists. Also, `to_dict` could be called outside of the `save` scope. Let me know if you think moving it to `to_dict` is still a good idea and I'll do it.

mlflow/models/model.py
r3stl355: Just to verify, sounds like this design matches my current implementation, please confirm or correct my understanding. Currently, I store `mlflow_version=None` if the model is loaded from the pre-mlflow-version-attribute file and don't save into file the `mlflow_version` attribute at all if a save is called on that model.

mlflow/models/model.py
r3stl355: sure, will do

BenWilson2 (approved):
Thanks for the addition! LGTM!

Issue #4750, opened 2021-08-27
[FR] Display max/min value of a metric in the experiments UI


dannyfriar:
## Proposal Summary
Allow displaying the max/min value of a metric for each run in the experiments UI. E.g. if logging a validation metric every epoch for NN training it would be useful to easily compare runs by the "best" value of the metric rather than the latest one.

## Motivation
Latest value of a metric for a specific run may not accurately represent how well the experiment run is doing. This would allow easily comparing experiment runs in the experiments UI when training NNs or boosted trees with early stopping.

Similar to: https://github.com/mlflow/mlflow/issues/1422

niksbansode:
Hi,
I'm willing to contribute to this enhancement. I'll contribute to both UI and python server implementation. Below is the mocked screen for this enhancement. Min/Max values are calculated across the epochs for any particular run. 

![MinMaxFR](https://user-images.githubusercontent.com/19736733/131241756-391dea21-4cd0-4ed0-8095-2d65a7f1461f.png)

@dannyfriar please confirm if this enhancement is correctly aligned with the proposed solution.

niksbansode:
Seems API changes are required, @harupy for that, changes might be needed in MlflowMessages.js which looks like proto file. How to generate that file? I'm not able to find necessary resources.
Here I'm facing difficulty:
For this enhancement, API response may look like this:
API: ajax-api/2.0/preview/mlflow/runs/get?run_id=3deb45d5572c4801ab3ee67a3395dd25
```
{
  "run": {
    "info": {
	....
	....
	....
    },
    "data": {
      "metrics": [
        {
          "key": "val_accuracy",
          "value": 0.8064516186714172,
          "timestamp": "1630175704448",
          "step": "4",
          "min_value": 0.7619577050209045,
          "max_value": 0.8109009861946106
        },
	....
	....
	....
        {
          "key": "loss",
          "value": 0.328249454498291,
          "timestamp": "1630175704448",
          "step": "4",
          "min_value": 0.328249454498291,
          "max_value": 1.4314649105072021
        }
      ],
      "params": [
        ....
	....
	....
      ],
      "tags": [
        ....
	....
	....
      ]
    }
  }
}
```

Here I have added min_value & max_value as additional parameters. Now I want to render them on UI but Metric prototype only return key value pair and we need max_value & min_value parameter. Could you please suggest an approach for this problem?

dannyfriar:
Hi @niksbansode ,

[...]

PR #5517, opened 2022-03-21
Show minimum and maximum metric values in the run view UI

adamreeve:

Adds "Min" and "Max" columns to the metrics table in the run view, and renames the "Value" column to "Latest". The minimum and maximum values are computed using the GetMetricHistory API, as discussed in #4750

min-max-run

## How is this patch tested?

Unit tests added for the new reduce functions and for the new component behaviour.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

The run view in the UI now displays the minimum and maximum metric values for a run.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/server/js/src/experiment-tracking/components/RunView.js
+++ b/mlflow/server/js/src/experiment-tracking/components/RunView.js
@@ -8,7 +8,7 @@ import { HtmlTableView } from './HtmlTableView';
 import { Link } from 'react-router-dom';
 import Routes from '../routes';
 import ArtifactPage from './ArtifactPage';
-import { getLatestMetrics } from '../reducers/MetricReducer';
+import { getLatestMetrics, getMinMetrics, getMaxMetrics } from '../reducers/MetricReducer';
 import { Experiment } from '../sdk/MlflowMessages';
 import Utils from '../../common/utils/Utils';
 import { NOTE_CONTENT_TAG, NoteInfo } from '../utils/NoteUtils';
@@ -17,7 +17,7 @@ import { EditableTagsTableView } from '../../common/components/EditableTagsTable
 import { Button, Descriptions, message } from 'antd';
 import { CollapsibleSection } from '../../common/components/CollapsibleSection';
 import { EditableNote } from '../../common/components/EditableNote';
-import { setTagApi, deleteTagApi } from '../actions';
+import { getMetricHistoryApi, setTagApi, deleteTagApi } from '../actions';
 import { PageHeader, OverflowMenu } from '../../shared/building_blocks/PageHeader';
 
 export class RunViewImpl extends Component {
@@ -29,10 +29,13 @@ export class RunViewImpl extends Component {
     params: PropTypes.object.isRequired,
     tags: PropTypes.object.isRequired,
     latestMetrics: PropTypes.object.isRequired,
+    minMetrics: PropTypes.object,
+    maxMetrics: PropTypes.object,
     getMetricPagePath: PropTypes.func.isRequired,
     runDisplayName: PropTypes.string.isRequired,
     runName: PropTypes.string.isRequired,
     handleSetRunTag: PropTypes.func.isRequired,
+    getMetricHistoryApi: PropTypes.func.isRequired,
     setTagApi: PropTypes.func.isRequired,
     deleteTagApi: PropTypes.func.isRequired,
     modelVersions: PropTypes.arrayOf(PropTypes.object),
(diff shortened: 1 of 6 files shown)


mlflow/server/js/src/experiment-tracking/components/RunView.js
```diff
@@ -51,6 +54,13 @@ export class RunViewImpl extends Component {
   componentDidMount() {
     const pageTitle = `${this.props.runDisplayName} - MLflow Run`;
     Utils.updatePageTitle(pageTitle);
+    // Request metric history for all the run metrics so that minimum
+    // and maximum values will be computed.
+    // Because the RunView is wrapped in a RequestStateWrapper, it won't be
+    // mounted until after latestMetrics data is available.
+    Object.values(this.props.latestMetrics).forEach((metric) => {
+      this.props.getMetricHistoryApi(this.props.runUuid, metric.key);
+    });
```
sunishsheth2009, line 63: Do you think it would be a heavy call if there are a lot of metrics? @dbczumar, is there a way to include the max and min in the latest metric itself where the BE already gets the history and computes it rather than doing this?
I see your comment [here](https://github.com/mlflow/mlflow/issues/4750#issuecomment-1068725414), but would like to understand the pros and cons of your suggestion.

sunishsheth2009 (commented):
Can you share a screenshot of how the new UI would look like? I see the mock in the discussion but just wanted to add one from your UI as part of this PR for reference as well.

mlflow/server/js/src/experiment-tracking/reducers/MetricReducer.js
```diff
@@ -72,6 +81,39 @@ export const latestMetricsByRunUuid = (state = {}, action) => {
...
+  switch (action.type) {
+    case fulfilled(GET_METRIC_HISTORY_API): {
+      const newState = { ...state };
+      const { runUuid, key } = action.meta;
+      const { metrics } = action.payload;
+      if (metrics && metrics.length > 0) {
+        const reducedMetric = Metric.fromJs(reducer(metrics));
+        if (newState[runUuid]) {
+          newState[runUuid][key] = reducedMetric;
+        } else {
```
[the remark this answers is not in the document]
adamreeve: The different reduced metrics stored under the same runUuid will be for different metric names/keys. So eg. for one run we might store both the minimum training_loss and minimum validation_loss.

I think overriding makes sense here, eg. if we were to later re-fetch the metric history the min and max might have changed for a run that is still ongoing, so we'd want to recompute the min and max (this isn't used in this PR though, I only fetch the data once after the run view component mounts).

This code is mostly actually copied from the existing code above under the `fulfilled(GET_METRIC_HISTORY_API)` case in `latestMetricsByRunUuid`.

mlflow/server/js/src/experiment-tracking/components/RunView.js
adamreeve: Yes this is a bit of a concern, I've made sure that fetching this data doesn't block the rest of the UI though. If the metric history is slow to query then a user might just see empty min and max values for a while until they're populated.

I have started prototyping support for storing the min and max values in the database, and am hoping that change might eventually be considered too. This is a much larger and more disruptive change though, needing DB schema and API changes (the current changes on top of this branch can bee seen at https://github.com/adamreeve/mlflow/compare/run_min_max...metric_min_max).

mlflow/server/js/src/experiment-tracking/reducers/MetricReducer.js
```diff
@@ -1,3 +1,4 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
adamreeve: Good point thanks, I've fixed that now.

mlflow/server/js/src/experiment-tracking/components/RunView.js
dbczumar: @adamreeve @sunishsheth2009 Apologies for failing to consider the performance implications here earlier. As noted by @adamreeve , adjusting the service API and backend storage layer to accommodate this change is quite costly (at least a couple months of coordinated effort across multiple third-party service implementers). However, there's also a performance concern if we make too many repeated calls to GetMetricHistory on the same page. While I think this is likely fine if, e.g. a run has < 10 metrics, it may be insufficiently performant for runs with >= 10 metrics. 

One solution to this problem would be to require the user to "click to load" this information for a particular metric. 

Another solution would be to move this summary information into the individual Metric pages, which may still provide some UX benefit because, right now, the user needs to interact with the plot to fetch this information. We could also integrate this with the Metric page for run comparison: when multiple runs are being compared based on a single metric, we can include the min / max / latest metric values for each run in a table. This provides a much simpler UX relative to current one where users need to manually inspect each plot curve.

My preference would be to implement the second workflow. While it doesn't help us surface this information higher up in the MLflow UI, it still makes it significantly easier to interpret data from a single metric across one or more runs. I realize that this is substantially different from the table implemented in this PR, and I apologize again for the lack of clarity up front. Let me know what your thoughts are @adamreeve @sunishsheth2009.

adamreeve: Hi @dbczumar, yes I think having this information on the individual metric pages would still be useful, and being able to show them on the comparison page would probably be more useful than just in the run page here and allow loading the history on-demand (I'm imagining a select drop-down to select from latest/min/max next to the metric name for example), so I'd be happy to take a look at implementing that.

adamreeve: Ah sorry I misread your comment, I was thinking of the metric table on the run comparison page, but yes just displaying the min and max values in the single metric comparison page would make sense.

adamreeve:
I've created a new PR (#5574) that shows the latest/min/max values in the metric view instead, so will close this.

PR #5531, opened 2022-03-25
Use schema when parsing csv

vvijay-bolt:

Use the input schema when parsing csv input during model serving. This ensures that the resulting dataframe has the correct types.

## How is this patch tested?

Train and serve the modified logistic regression example below.
Save the following in test_csv.csv:
x
1
-1

Then try:
curl -d '{"columns":["x"],"index":[0,1],"data":[[1],[-1]]}' -H 'Content-Type: application/json' localhost:5000/invocations
curl --data-binary '@test_csv.csv' -H 'Content-Type: text/csv' localhost:5000/invocations

The json request returns [1, 0].
On master, the csv request will produce: {"error_code": "BAD_REQUEST", "message": "Incompatible input types for column x. Can not safely convert int64 to int32."}
After applying the fix, the csv request will also product [1, 0].

Modified logistic regression example:

import numpy as np
import pandas as pd
from mlflow.models.signature import infer_signature

import mlflow
import mlflow.sklearn

if __name__ == "__main__":
    X = pd.DataFrame({'x': [-2, -1, 0, 1, 2, 1]}, dtype='int32')
    y = np.array([0, 0, 1, 1, 1, 0])
    lr = LogisticRegression()
    lr.fit(X, y)
    score = lr.score(X, y)
    print("Score: %s" % score)

    class Model(mlflow.pyfunc.PythonModel):
        def __init__(self, m):
            self.m = m

        def predict(self, context, x):
            return self.m.predict(x)

    model = Model(lr)
    signature = infer_signature(X, model.predict(None, X))
    print('Signature')
    print(signature)
    mlflow.pyfunc.log_model('model', python_model=model, signature=signature)
    print("Model saved in run %s" % mlflow.active_run().info.run_uuid)

No

No
Components 

- [- ] `rn/bug-fix` - A user-facing bug fix worth mentioning in the release notes


BenWilson2:
Hi @vvijay-bolt could you add a test suite to validate this change to ensure that the parsing of a .csv with different submitted data type orientations functions correctly? 
It can be added here: https://github.com/mlflow/mlflow/blob/master/tests/pyfunc/test_scoring_server.py

vvijay-bolt:
@BenWilson2 I added a test (modeled after a similar json test) to verify that the datatypes were set correctly after parsing. The pandas functions to read and write csv files were throwing errors on the datetime fields, so I left them out of the test.

mlflow/utils/proto_json_utils.py
```diff
@@ -193,6 +193,20 @@ def default(self, o):  # pylint: disable=E0202
             return super().default(o)
 
 
+def _extract_pandas_input_types_from_schema(schema):
```
arjundc-db, line 196: If the input is of type csv, it has to be a pandas input. Since a pandas schema will always be a ColSpec, I dont think this entire function is necessary.
Simply using `dtypes = dict(zip(schema.input_names(), schema.pandas_types()))` should suffice.

vvijay-bolt: Made the change.

mlflow/utils/proto_json_utils.py
```diff
@@ -210,17 +210,7 @@ def _dataframe_from_json(
...
-            #  - a single tensor: attempt to parse all columns with the same dtype
-            #  - a dictionary of tensors: each column gets the type from an equally named tensor
-            if len(schema.inputs) == 1:
-                dtypes = schema.numpy_types()[0]
-            else:
-                dtypes = dict(zip(schema.input_names(), schema.numpy_types()))
-        else:
-            dtypes = dict(zip(schema.input_names(), schema.pandas_types()))
-
+        dtypes = _extract_pandas_input_types_from_schema(schema)
```
arjundc-db, line 213: Can you also revert this change please.

vvijay-bolt: Done.

vvijay-bolt:
autoformat

arjundc-db (approved):
LGTM.
Thank you @vvijay-bolt !

tomasatdatabricks (approved):
lgtm!

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -136,14 +136,20 @@ def parse_json_input(json_input, orient="split", schema: Schema = None):
...
+def parse_csv_input(csv_input, schema: Schema = None):
     """
     :param csv_input: A CSV-formatted string representation of a Pandas DataFrame, or a stream
                       containing such a string representation.
+    :param schema: Optional schema specification to be used during parsing.
     """
 
     try:
-        return pd.read_csv(csv_input)
+        csv_input = StringIO(csv_input)
```
arjundc-db, line 147: Let's move this line back to  https://github.com/mlflow/mlflow/pull/5531/files#diff-e607d3e6aefa036b1bfe9416fd1aa8b0934120f458cf263ddebab224f832e4a5L264 since the input to parse_csv can also be a path to a csv file. We need to fix our documentation for this.

tests/pyfunc/test_scoring_server.py
```diff
@@ -429,6 +444,14 @@ def test_split_oriented_json_to_df():
     assert set(str(dt) for dt in df.dtypes) == {"object", "float64", "int64"}
 
 
+def test_parse_with_schema_csv(pandas_df_with_csv_types):
+    schema = Schema([ColSpec(c, c) for c in pandas_df_with_csv_types.columns])
+    df = _shuffle_pdf(pandas_df_with_csv_types)
+    csv_str = df.to_csv(index=False)
+    df = pyfunc_scoring_server.parse_csv_input(csv_str, schema=schema)
```
arjundc-db, line 451: We need StringIO here:
```suggestion
    df = pyfunc_scoring_server.parse_csv_input(csv_str, schema=schema)
```

Please import it using 
```
try:
    from StringIO import StringIO
except ImportError:
    from io import StringIO
```
to maintain compatibility.

arjundc-db (commented):
Sorry, two more comments to fix test failures and we should we good to go.

test_csv.csv
```diff
@@ -0,0 +1,3 @@
+x
```
BenWilson2, line 1: Where is this used?

examples/sklearn_logistic_mod/train.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
```
BenWilson2: Any new examples need to be registered to execute as part of the Examples CI here: https://github.com/mlflow/mlflow/blob/101ad6e8eb383c769178df0df83d1d2a1cea6b4a/tests/examples/test_examples.py#L92-L94
. Do we need this additional test or can a signature example be added to examples/sklearn_logistic_regression/train.py? 
It might also be worthwhile to add a load from the logged (saved) location to run the score inference again to demonstrate the loading of the pyfunc flavor with the type-validated signature for a .csv file.

BenWilson2 (approved):
LGTM!

PR #5540, opened 2022-03-28
Run `test_sqlalchemy_store.py` against PostgreSQL, MySQL, and MSSQL

harupy:

Run `test_sqlalchemy_store.py` against PostgreSQL, MySQL, and MSSQL to ensure MLflow tracking operations work in different databases and prevent issues / regressions.

## How is this patch tested?

Fixed tests

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 
- [x] `area/windows`: Windows support

Language 

Integrations

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -168,11 +168,12 @@ jobs:
         ./tests/db/compose.sh build --build-arg DEPENDENCIES="$(python setup.py -q dependencies)"
     - name: Database tests - run
       run: |
+        set +e
         for service in $(./tests/db/compose.sh config --services | grep '^mlflow-')
         do
           # Set `--no-TTY` to show container logs on GitHub Actions:
           # https://github.com/actions/virtual-environments/issues/5022
-          ./tests/db/compose.sh run --rm --no-TTY $service
+          ./tests/db/compose.sh run --rm --no-TTY $service pytest tests/store/tracking/test_sqlalchemy_store.py tests/db
         done
     - name: Database tests - clean up
       run: |
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -168,7 +168,6 @@ jobs:
         ./tests/db/compose.sh build --build-arg DEPENDENCIES="$(python setup.py -q dependencies)"
     - name: Database tests - run
       run: |
-        set +e
         for service in $(./tests/db/compose.sh config --services | grep '^mlflow-')
         do
           # Set `--no-TTY` to show container logs on GitHub Actions:
--- a/mlflow/store/tracking/sqlalchemy_store.py
+++ b/mlflow/store/tracking/sqlalchemy_store.py
@@ -987,9 +987,9 @@ def _get_orderby_clauses(order_by_list, session):
             observed_order_by_clauses.add((key_type, key))
 
             if ascending:
-                clauses.append(order_value)
+                clauses.append(sqlalchemy.nulls_last(order_value))
             else:
-                clauses.append(order_value.desc())
+                clauses.append(sqlalchemy.nulls_last(order_value.desc()))
 
     if (SearchUtils._ATTRIBUTE_IDENTIFIER, SqlRun.start_time.key) not in observed_order_by_clauses:
         clauses.append(SqlRun.start_time.desc())
--- a/mlflow/store/tracking/sqlalchemy_store.py
+++ b/mlflow/store/tracking/sqlalchemy_store.py
@@ -147,6 +147,9 @@ class SqlAlchemyStore(AbstractStore):
             with self.ManagedSessionMaker() as session:
                 self._create_default_experiment(session)
 
+    def _get_dialect(self):
+        return self.engine.dialect.name
+
     def _set_zero_value_insertion_for_autoincrement_column(self, session):
         if self.db_type == MYSQL:
             # config letting MySQL override default
@@ -777,7 +780,7 @@ class SqlAlchemyStore(AbstractStore):
             cases_orderby, parsed_orderby, sorting_joins = _get_orderby_clauses(order_by, session)
 
             stmt = select(SqlRun, *cases_orderby)
-            for j in _get_sqlalchemy_filter_clauses(parsed_filters, session):
+            for j in _get_sqlalchemy_filter_clauses(parsed_filters, session, self._get_dialect()):
                 stmt = stmt.join(j)
             # using an outer join is necessary here because we want to be able to sort
             # on a column (tag, metric or param) without removing the lines that
@@ -792,7 +795,7 @@ class SqlAlchemyStore(AbstractStore):
                 .filter(
                     SqlRun.experiment_id.in_(experiment_ids),
                     SqlRun.lifecycle_stage.in_(stages),
-                    *_get_attributes_filtering_clauses(parsed_filters),
+                    *_get_attributes_filtering_clauses(parsed_filters, self._get_dialect()),
                 )
                 .order_by(*parsed_orderby)
                 .offset(offset)
@@ -846,7 +849,7 @@ class SqlAlchemyStore(AbstractStore):
             session.merge(SqlTag(key=MLFLOW_LOGGED_MODELS, value=value, run_uuid=run_id))
 
 
-def _get_attributes_filtering_clauses(parsed):
+def _get_attributes_filtering_clauses(parsed, dialect):
     clauses = []
     for sql_statement in parsed:
         key_type = sql_statement.get("type")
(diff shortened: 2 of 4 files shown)


.github/workflows/master.yml
```diff
@@ -168,11 +168,12 @@ jobs:
         ./tests/db/compose.sh build --build-arg DEPENDENCIES="$(python setup.py -q dependencies)"
     - name: Database tests - run
       run: |
+        set +e
```
harupy, line 171: I'll remove this

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -967,7 +970,7 @@ def _get_orderby_clauses(order_by_list, session):
                         # the column (is_nan) is not nullable. However it could become an issue
                         # if this precondition changes in the future.
                         (subquery.c.is_nan == sqlalchemy.true(), 1),
-                        (order_value.is_(None), 1),
+                        (order_value.is_(None), 2),
```
harupy, line 973: Let's say we run code like this:

```python
with mlflow.start_run():
    log_metric("m", 1)

with mlflow.start_run():
    log_metric("m", np.nan)

with mlflow.start_run():
    pass

mlflow.search_runs(["0"], order_by=["metrics.m"])
```

This code creates a table looks like this, then sorts rows using the `metrics.m` and `clause_1` columns.

| run_id | metrics.m | is_nan | clause_1 |
| :----- | :-------- | :----- | :------- |
| 1      | 1         | False  | 0        |
| 2      | 0         | True   | 1        |
| 3      | NULL      | False  | 1        |

```sql
SELECT ... ORDER BY clause_1, metrics.m
```

In PostgreSQL, the result looks like this because NULLs are sorted AFTER non-null values:

| run_id | metrics.m | is_nan | clause_1 |
| :----- | :-------- | :----- | :------- |
| 1      | 1         | False  | 0        |
| 2      | 0         | True   | 1        |
| 3      | NULL      | False  | 1        |

In other databases (SQLite, MySQL, SQL server), the result looks like this because NULLs are sorted BEFORE non-null values:

| run_id | metrics.m | is_nan | clause_1 |
| :----- | :-------- | :----- | :------- |
| 1      | 1         | False  | 0        |
| 3      | NULL      | False  | 1        |
| 2      | 0         | True   | 1        |

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -1943,24 +1981,19 @@ def test_sqlalchemy_store_can_be_initialized_when_default_experiment_has_been_de
...
-        os.close(fd)
-        self.db_url = "%s%s" % (DB_URI, self.temp_dbfile)
+        super()._setup_db_uri()
         engine = sqlalchemy.create_engine(self.db_url)
         InitialBase.metadata.create_all(engine)
         invoke_cli_runner(mlflow.db.commands, ["upgrade", self.db_url])
         self.store = SqlAlchemyStore(self.db_url, ARTIFACT_URI)
 
-    def tearDown(self):
-        os.remove(self.temp_dbfile)
```
harupy, line 1962: We don't need to override the tearDown method.

mlflow/utils/search_utils.py
```diff
@@ -89,8 +90,14 @@ class SearchUtils:
     }
 
     @classmethod
-    def get_sql_filter_ops(cls, column, operator):
-        sql_filter_ops = {"LIKE": column.like, "ILIKE": column.ilike}
+    def get_sql_filter_ops(cls, column, operator, dialect):
+        if dialect == "mysql":
+            like_op = column.op("LIKE BINARY")
+        elif dialect == "mssql":
+            like_op = sa.collate(column, "SQL_Latin1_General_CP1_CS_AS").like
```
harupy, line 97: `LIKE` is case-insensitive in MySQL and SQL server.

harupy: This block enables `case_sensitive_like` for SQLite:

https://github.com/mlflow/mlflow/blob/d9489d0c61f75d763a567ade17b7168c47b4da74/mlflow/store/db/utils.py#L76-L78

mlflow/utils/search_utils.py
```diff
@@ -89,8 +89,14 @@ class SearchUtils:
     }
 
     @classmethod
-    def get_sql_filter_ops(cls, column, operator):
-        sql_filter_ops = {"LIKE": column.like, "ILIKE": column.ilike}
+    def get_sql_filter_ops(cls, column, operator, dialect):
+        if dialect == "mysql":
+            like_op = column.op("LIKE BINARY")
+        elif dialect == "mssql":
+            like_op = column.collate("SQL_Latin1_General_CP1_CS_AS").like
```
BenWilson2, line 96: Do we want to include Kana sensitive as well? 
"SQL_Latin1_General_CP1_CS_AS_KS" or does sorting based on characters make the ordering confusing?

tests/store/tracking/test_sqlalchemy_store.py
BenWilson2: Was this originally for Windows?

harupy: I'm not sure, `TestSqlAlchemyStore.tearDown` also contains `os.remove(self.temp_dbfile)`.

mlflow/utils/search_utils.py
harupy: I think we do :)

harupy: Updated!

harupy: `SQL_Latin1_General_CP1_CS_AS_KS` didn't work:

https://github.com/mlflow/mlflow/runs/5728726867?check_suite_focus=true#step:11:10259

```
E           mlflow.exceptions.MlflowException: (pyodbc.ProgrammingError) ('42000', "[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Invalid collation 'SQL_Latin1_General_CP1_CS_AS_KS'. (448) (SQLExecDirectW)")
```

harupy: Is `_KS` only available for Japanese collations such as `Japanese_Bushu_Kakusu_100_CS_AS_KS_WS_UTF8`?

harupy: I don't think kana-insensitivity is a big issue though.

harupy: Offline discussion: we should also support kana sensitivity if there is no downside.

BenWilson2: I'm wondering if the `_UTF8` addition as suffix is required. It's pretty hard to tell from their docs... https://docs.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support?view=sql-server-ver15#Collation_Defn
I wonder if people using MSSQL as a tracking backend end up writing a T-SQL script when they setup MLflow to enable faster filtering and apply specific column collations to speed it up?

I agree, not that big of a deal to enable this (it would just be nice to enable functionality for as many users as possible, but if the config is broken, then it's broken...)

mlflow/utils/search_utils.py
```diff
@@ -89,8 +90,14 @@ class SearchUtils:
     }
 
     @classmethod
-    def get_sql_filter_ops(cls, column, operator):
-        sql_filter_ops = {"LIKE": column.like, "ILIKE": column.ilike}
+    def get_sql_filter_ops(cls, column, operator, dialect):
+        if dialect == MYSQL:
+            like_op = column.op("LIKE BINARY")
```
harupy, line 95: Offline discussion: insert a non-binary `LIKE` to improve the performance.

```sql
LIKE :pattern OR LIKE BINARY :pattern
```

harupy:
Code to check kana sensitivity in SQL server:

```python
import mlflow
import uuid

name = uuid.uuid4().hex
experiment_id = mlflow.create_experiment(name)
mlflow.set_experiment(name)

for val in ["りんご", "リンゴ"]:
    with mlflow.start_run():
        mlflow.log_param("p", val)

print(mlflow.search_runs([experiment_id], filter_string="parameters.p LIKE 'リンゴ'"))
```

mlflow/utils/search_utils.py
```diff
@@ -89,8 +91,20 @@ class SearchUtils:
...
+        if dialect == MYSQL:
+
+            def like_op(value):
+                # Filter by `LIKE` ahead of `LIKE BINARY` for runtime performance
+                return sa.and_(
+                    column.like(value), column.op("LIKE BINARY", is_comparison=True)(value)
+                )
+
+        elif dialect == MSSQL:
+            like_op = column.collate("Japanese_Bushu_Kakusu_100_CS_AS_KS_WS").like
```
harupy, line 104:
```suggestion
            like_op = column.collate("Japanese_Bushu_Kakusu_100_CS_AS_KS_WS_UTF8").like
```

I tried `Japanese_Bushu_Kakusu_100_CS_AS_KS_WS_UTF8` on local but got an invalid collation error.

BenWilson2 (approved):
This is great stuff, Haru! Improving usability for users utilizing different backend stores and helping to notify us if any changes will create issues for them is wonderful. 
LGTM!

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -1518,6 +1567,10 @@ def test_search_with_max_results(self):
 
         assert runs[:1000] == self._search(exp)
         for n in [0, 1, 2, 4, 8, 10, 20, 50, 100, 500, 1000, 1200, 2000]:
+            if n == 0 and self.store._get_dialect() == MSSQL:
+                # In SQL server, `max_results = 0` results in the following error:
```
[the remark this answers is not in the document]
harupy: I believe it doesn't because UI doesn't allow users to set the number of runs to fetch.

PR #5543, opened 2022-03-28
Make mlflow.pyfunc.spark_udf driver side raise exception contains conda env create command stdout/stderr output.

WeichenXu123:

Make mlflow.pyfunc.spark_udf driver side raise exception contains conda env create command stdout/stderr output
This make it easier for user to debug when conda env creation failed in UDF side.

## How is this patch tested?

https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/478391771932729/command/2674920562632835

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mlflow/pyfunc/backend.py
```diff
@@ -231,6 +235,7 @@ def _execute_in_conda_env(
                         If False, return the server process `Popen` instance immediately.
     :param stdout: Redirect server stdout
     :param stderr: Redirect server stderr
+    :param stream_output_during_creating_env: stream output during running "conda env create" command.
```
dbczumar, line 238: Can we clarify what happens when `stream_output` is `True` and what happens when it's `False`? It's not very obvious from reading the code right now. Specifically, are we relying on the fact that passing `stream_output=False` to `mlflow.utils.process.exec_cmd()` captures and returns the stdout / stderr streams? If so, the naming of the variable is also pretty confusing, since both modes are streaming output; the only difference is the destination streams: sys.stderr / sys.stdout vs subprocess.PIPEs.

mlflow/pyfunc/__init__.py
```diff
@@ -963,7 +963,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
         # Prepare restored environment in driver side if possible.
         if env_manager == "conda":
             _get_flavor_backend(local_model_path, no_conda=False, install_mlflow=False).prepare_env(
-                model_uri=local_model_path
+                model_uri=local_model_path, stream_output=False
```
dbczumar, line 966: Are we relying on the fact that passing `stream_output=False` to `mlflow.utils.process.exec_cmd()` captures and returns the stdout / stderr streams? I don't see those streams being used anywhere.

The naming is a bit confusing. `stream_output=False` would seem to imply that the command output shouldn't be sent anywhere at all.

harupy: `capture_output` might make more sense.

dbczumar (approved):
LGTM once we change to `capture_output` and replace `%s` with `"".format(...)`

mlflow/pyfunc/backend.py
```diff
@@ -220,6 +225,7 @@ def _execute_in_conda_env(
     preexec_fn=None,
     stdout=None,
     stderr=None,
+    capture_output_during_creating_env=False,
```
harupy, line 228:
```suggestion
    capture_output_while_creating_env=False,
```

Can we rename this argument?

harupy:
```suggestion
    capture_output_during_env_creation=False,
```

mlflow/pyfunc/backend.py
```diff
@@ -31,13 +31,18 @@ def __init__(self, config, workers=1, no_conda=False, install_mlflow=False, **kw
...
 
-    def prepare_env(self, model_uri):
+    def prepare_env(self, model_uri, capture_output=False):
         local_path = _download_artifact_from_uri(model_uri)
         if self._no_conda or ENV not in self._config:
             return 0
         conda_env_path = os.path.join(local_path, self._config[ENV])
         command = 'python -c ""'
-        return _execute_in_conda_env(conda_env_path, command, self._install_mlflow)
+        return _execute_in_conda_env(
```
harupy, line 40: On second thought, can we split `_execute_in_conda_env` into two functions, one that creates an env and returns the env name, one that takes an env name and a command to run?

mlflow/pyfunc/backend.py
WeichenXu123: what about `capture_conda_failure_output_and_raise` which is more meaningful ?

mlflow/pyfunc/backend.py
WeichenXu123: We already have the function `get_or_create_conda_env`
but "prepare_env" still calls `_execute_in_conda_env` I think it is for running `python -c ""` immediately after conda env created in order to validate the env works ?

mlflow/pyfunc/backend.py
WeichenXu123: what about `capture_conda_failure_output_and_raise` which is more meaningful here ?

mlflow/utils/conda.py
```diff
@@ -162,25 +165,29 @@ def get_or_create_conda_env(conda_env_path, env_id=None):
...
+                            "--yes",
+                            "--name",
+                            project_env_name,
+                            "--all",
+                        ],
+                        capture_output=False,
+                    )
+            except Exception:
+                # swallow exception raised during cleaning env.
+                pass
```
harupy, line 190: Can we print out the exception using `_logger.warning`?

mlflow/pyfunc/backend.py
harupy: Can we pull `get_or_create_conda_env` out of `_execute_in_conda_env`?

mlflow/pyfunc/backend.py
harupy: I think `capture_conda_failure_output_and_raise` is very confusing and difficult to guess its behavior.

WeichenXu123: removed the param after pull out "get_or_create_conda_env"

mlflow/pyfunc/backend.py
```diff
@@ -30,18 +30,23 @@ def __init__(self, config, workers=1, no_conda=False, install_mlflow=False, **kw
...
         local_path = _download_artifact_from_uri(model_uri)
         if self._no_conda or ENV not in self._config:
             return 0
         conda_env_path = os.path.join(local_path, self._config[ENV])
+
+        conda_env_name = get_or_create_conda_env(
+            conda_env_path, env_id=self._env_id, capture_output=capture_output
+        )
+
         command = 'python -c ""'
```
harupy, line 45: Do we really need to run `python -c ""`?

mlflow/utils/conda.py
```diff
@@ -186,7 +186,7 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
                         capture_output=False,
                     )
             except Exception:
-                # swallow exception raised during cleaning env.
+                _logger.warning(f"Removing conda env '{project_env_name}' failed.", exc_info=True)
```
harupy, line 189: Can we include the exception in the warning message?

mlflow/pyfunc/backend.py
WeichenXu123:
```
    if install_mlflow:
        if "MLFLOW_HOME" in os.environ:  # dev version
            install_mlflow = "pip install -e {} 1>&2".format(os.environ["MLFLOW_HOME"])
        else:
            install_mlflow = "pip install mlflow=={} 1>&2".format(VERSION)
```
this part is in `_execute_in_conda_env`, I think 'python -c ""' is for triggering it.

mlflow/utils/conda.py
```diff
@@ -162,25 +165,29 @@ def get_or_create_conda_env(conda_env_path, env_id=None):
...
+                            "--yes",
+                            "--name",
+                            project_env_name,
+                            "--all",
+                        ],
+                        capture_output=False,
+                    )
+            except Exception:
+                _logger.warning(f"Removing conda env '{project_env_name}' failed.", exc_info=True)
+                pass
```
harupy, line 190:
```suggestion
```

mlflow/utils/process.py
```diff
@@ -54,6 +54,6 @@ def exec_cmd(
         exit_code = child.wait()
         if throw_on_error and exit_code != 0:
             raise ShellCommandException(
-                "Non-zero exit code: %s\n\nSTDOUT:\n%s\n\nSTDERR:%s" % (exit_code, stdout, stderr)
+                f"Non-zero exit code: {exit_code}\n\nSTDOUT:\n{stdout}\n\nSTDERR:\n{stderr}"
```
harupy, line 57: Are `stdout` and `stderr` binary strings?

mlflow/utils/process.py
```diff
@@ -18,7 +18,7 @@ def exec_cmd(
     throw_on_error -- if true, raises an Exception if the exit code of the program is nonzero
     env -- additional environment variables to be defined when running the child process
     cwd -- working directory for child process
-    stream_output -- if true, does not capture standard output and error; if false, captures these
+    capture_output -- if False, does not capture standard output and error; if True, captures these
```
harupy, line 21:
```suggestion
    capture_output -- if True, captures standard output and error; if False, does not captures these
```

nit: can we swap `True` and `False`?

mlflow/utils/process.py
WeichenXu123: No. because the argument `universal_newlines` is True. which ensures they are str type.

mlflow/utils/conda.py
harupy: I missed line 186. `capture_output` is set to False. I think we don't need `exc_info=True`.

```suggestion
                _logger.warning(f"Removing conda env '{project_env_name}' failed.")
```

mlflow/utils/process.py
harupy: Got it!

mlflow/utils/conda.py
```diff
@@ -162,24 +165,31 @@ def get_or_create_conda_env(conda_env_path, env_id=None):
...
+                            "--name",
+                            project_env_name,
+                            "--all",
+                        ],
+                        capture_output=False,
+                    )
+            except Exception as e:
+                _logger.warning(
+                    f"Removing conda env '{project_env_name}' failed (error: {repr(e)}).",
+                    exc_info=True,
```
harupy, line 192:
```suggestion
                _logger.warning(
                    f"Removing conda env '{project_env_name}' failed (error: {repr(e)})."
```

mlflow/utils/process.py
```diff
@@ -18,8 +18,8 @@ def exec_cmd(
     throw_on_error -- if true, raises an Exception if the exit code of the program is nonzero
     env -- additional environment variables to be defined when running the child process
     cwd -- working directory for child process
-    stream_output -- if true, does not capture standard output and error; if false, captures these
-      streams and returns them
+    capture_output -- if True, captures standard output and error; if False, does not captures
+                      these.
```
harupy, line 22:
```suggestion
    capture_output -- if True, captures standard output and error; if False, does not capture
                      these.
```

mlflow/utils/conda.py
```diff
@@ -99,6 +99,9 @@ def get_or_create_conda_env(conda_env_path, env_id=None):
                    same conda dependencies but are supposed to be different based on the context.
                    For example, when serving the model we may install additional dependencies to the
                    environment after the environment has been activated.
+    :param capture_output: If True, captures these streams and returns them, and when failed,
+                           raise error contains captured stdout/stderr output, if False, does not
+                           capture standard output and error.
```
harupy, line 104: This sounds like an exception is not thrown when `capture_output` is False.

mlflow/utils/process.py
```diff
@@ -18,8 +18,8 @@ def exec_cmd(
     throw_on_error -- if true, raises an Exception if the exit code of the program is nonzero
     env -- additional environment variables to be defined when running the child process
     cwd -- working directory for child process
-    stream_output -- if true, does not capture standard output and error; if false, captures these
-      streams and returns them
+    capture_output -- if True, captures standard output and error; if False, does not capture
+                      these.
```
harupy, line 22:
```suggestion
    capture_output -- if True, captures standard output and error and returns them; if False, does not capture these streams.
```

I think we should indicate standard output and error will be returned.

harupy (approved):
LGTM once the remaining comments are addressed

mlflow/utils/conda.py
```diff
@@ -99,6 +99,11 @@ def get_or_create_conda_env(conda_env_path, env_id=None):
                    same conda dependencies but are supposed to be different based on the context.
                    For example, when serving the model we may install additional dependencies to the
                    environment after the environment has been activated.
+    :param capture_output: If True, captures conda env creation command stdout/stderr output and
+                           when conda command failed, attach captured stdout/stderr output to
+                           exception message and raise exception; if False, does not capture
+                           conda env creation command stdout/stderr output and when conda command
+                           failed, raise exception without stdout/stderr output attached.
```
harupy, line 106: To me, this parameter description sounds a bit confusing because `get_or_create_conda_env` doesn't attach stdout/stderr to an exception but `process.exec_cmd` does.

WeichenXu123: updated.

mlflow/utils/process.py
```diff
@@ -18,8 +18,11 @@ def exec_cmd(
     throw_on_error -- if true, raises an Exception if the exit code of the program is nonzero
     env -- additional environment variables to be defined when running the child process
     cwd -- working directory for child process
-    stream_output -- if true, does not capture standard output and error; if false, captures these
-      streams and returns them
+    capture_output -- If True, captures command stdout/stderr output and when conda command
+                      failed, attach captured stdout/stderr output to exception message and
+                      raise exception; if False, does not capture command stdout/stderr output
+                      and when command failed, raise exception without stdout/stderr
+                      output attached.
```
harupy, line 25: Why does the parameter description of `capture_output` contain conda?

WeichenXu123: removed conda words

harupy:
@dbczumar Could you also take another look since the code has been updated after your approval?

mlflow/utils/process.py
```diff
@@ -18,8 +18,11 @@ def exec_cmd(
     throw_on_error -- if true, raises an Exception if the exit code of the program is nonzero
     env -- additional environment variables to be defined when running the child process
     cwd -- working directory for child process
-    stream_output -- if true, does not capture standard output and error; if false, captures these
-      streams and returns them
+    capture_output -- If True, captures command stdout/stderr output and when command
+                      failed, attach captured stdout/stderr output to exception message and
+                      raise exception; if False, does not capture command stdout/stderr output
+                      and when command failed, raise exception without stdout/stderr
+                      output attached.
```
harupy, line 25: @WeichenXu123 I think I'd simplify this to the following:

```suggestion
    capture_output -- If True, stdout and stderr will be captured and included in an exception message on failure; if False, these streams won't be captured.
```

dbczumar (approved):
LGTM! Thanks @WeichenXu123 !

PR #5561, opened 2022-03-30
pyfunc.spark_udf: conda env dir / cache dir isolation and NFS optimization 

WeichenXu123:

For pyfunc.spark_udf:
1. Make conda env dir / cache dir isolated for different python processes.
2. NFS optimization: Create conda env once in driver side (write env dir into NFS mounted directory), and executor side directly read the env set up in driver side. 

## How is this patch tested?

Manually:
1. NFS enabled: https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/2674920562695789/command/2674920562695795

2. NFS disabled: https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/2674920562695847/command/2674920562695848

   next step, otherwise fix it.
4. Click `Details` on the right to open the job page of CircleCI.
5. Click the `Artifacts` tab.
6. Click `docs/build/html/index.html`.
7. Find the changed pages / sections and make sure they render correctly.

For pyfunc.spark_udf:

Make conda env dir / cache dir isolated for different python processes.
NFS optimization: Create conda env once in driver side (write env dir into NFS mounted directory), and executor side directly read the env set up in driver side.

Components 

Interface 

Language 

Integrations


dbczumar (commented):
@WeichenXu123 Looks great! Tried it out on Databricks & it seemed to work well! Left a few small comments.

mlflow/pyfunc/__init__.py
```diff
@@ -879,6 +879,37 @@ def _get_or_create_model_cache_dir():
...
+            # TODO: register deleting tmp_model_dir handler when exit
+        else:
+            import atexit
+            import shutil
+
+            conda_env_root_dir = tempfile.mkdtemp()
+            atexit.register(shutil.rmtree, conda_env_root_dir, ignore_errors=True)
+
+        # Create individual package cache dir "pkgs" under the conda_env_root_dir
+        # for each python process.
```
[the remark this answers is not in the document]
WeichenXu123: Good question... This might be hard to address.
But what's the case user need to fork a process ? Is it a common case ?

mlflow/pyfunc/__init__.py
```diff
@@ -991,15 +1022,25 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
...
 
+    if env_manager != "local" and os.name == "nt":
+        # TODO:
+        #  support windows system.
+        #   1. Address the issue killing mlflow server process when pyspark UDF task process died
+        #      (spark job canceled)
+        #   2. Addresss potential pyenv race condition issues on windows.
+        raise MlflowException(
+            f"`mlflow.pyfunc.spark_udf` env_manager value '{env_manager}' dose not support "
+            "windows system."
```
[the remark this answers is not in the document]
WeichenXu123: I don't find proper approach for handling windows case (mainly the killing child process issue when parent process died by receiving SIGKILL). What about creating a FR ticket and ask microsoft guys to help ?

mlflow/pyfunc/__init__.py
```diff
@@ -879,6 +879,37 @@ def _get_or_create_model_cache_dir():
...
+def _get_or_create_conda_env_root_dir(nfs_root_dir):
+    global _CONDA_ENV_ROOT_DIR
+    if _CONDA_ENV_ROOT_DIR is None:
+        if nfs_root_dir is not None:
+            # In databricks, the '/local_disk0/.ephemeral_nfs' is mounted as NFS disk
+            # the data stored in the disk is shared with all remote nodes.
+            root_dir = os.path.join(nfs_root_dir, "conda_envs")
+            os.makedirs(root_dir, exist_ok=True)
+            conda_env_root_dir = tempfile.mkdtemp(dir=root_dir)
+            # TODO: register deleting tmp_model_dir handler when exit
```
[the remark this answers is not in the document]
WeichenXu123: The size depends on the driver instance type.
I think I can address it before next DBR release.

WeichenXu123:
Note the conda env creation also calls "pip" to install packages, we need also consider whether to isolate pip cache directory within each process. CC @MrBago Does current databricks `%pip` do isolation for pip cache directory ?

mlflow/pyfunc/__init__.py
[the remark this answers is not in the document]
dbczumar: Could be! Can't we check the process id? https://www.geeksforgeeks.org/python-os-getpid-method/

mlflow/pyfunc/__init__.py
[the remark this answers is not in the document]
dbczumar: Sounds great! :D. Can we update the exception content too and also add `error_code=INVALID_PARAMETER_VALUE`?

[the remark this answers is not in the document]
dbczumar: Actually, how bad is it if we run anyway on Windows and just warn the user that there may be race conditions (are race conditions still a problem if we isolate the pip package cache?) and that the model server may continue running after executors terminate? I think this would be a better uX.

mlflow/pyfunc/mlserver.py
```diff
@@ -7,8 +8,15 @@
...
 def get_cmd(
-    model_uri: str, port: int = None, host: str = None, nworkers: int = None
+    model_uri: str, port: int = None, host: str = None, nworkers: int = None,
+    conda_env_root_dir: str = None
 ) -> Tuple[str, Dict[str, str]]:
+    if conda_env_root_dir is not None:
+        raise MlflowException(
+            "MLServer does not support setting `conda_env_root_dir`",
+            error_code=INVALID_PARAMETER_VALUE,
+        )
```
dbczumar, line 18: How hard is it to support this? Is https://github.com/mlflow/mlflow/pull/5561/files#diff-e607d3e6aefa036b1bfe9416fd1aa8b0934120f458cf263ddebab224f832e4a5R389-R390 sufficient?

mlflow/utils/conda.py
```diff
@@ -89,7 +90,19 @@ def _list_conda_environments():
     return list(map(os.path.basename, json.loads(prc.stdout).get("envs", [])))
 
 
-def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
+def _get_conda_env_root_dir_env(conda_env_root_dir):
```
dbczumar, line 93:
```suggestion
def _get_conda_env_root_dir_env(conda_env_root_dir=None):
```

mlflow/utils/conda.py
```diff
@@ -133,10 +149,20 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
+        # Append a suffix "-isolated" because if a conda env name exist in
+        # default conda env root dir, then if we set new "CONDA_ENVS_PATH" and run "conda env create"
+        # with the same conda env name, "CondaValueError: prefix already exists" error will
+        # be raised.
+        project_env_name = project_env_name + "-isolated"
+
+    if project_env_name not in _list_conda_environments():
         _logger.info("=== Creating conda environment %s ===", project_env_name)
+        if conda_env_root_dir is not None:
+            _logger.info("Use isolated conda environment root directory: %s", conda_env_root_dir)
```
dbczumar, line 165:
```suggestion
            _logger.info("Using isolated conda environment root directory: %s", conda_env_root_dir)
```

mlflow/utils/conda.py
```diff
@@ -190,5 +218,11 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
                     f"Removing conda env '{project_env_name}' failed (error: {repr(e)})."
                 )
             raise
+    else:
+        if conda_env_root_dir is not None:
+            _logger.info(
+                "Reuse cached conda environment at %s",
```
dbczumar, line 224:
```suggestion
                "Reusing cached conda environment at %s",
```

dbczumar (commented):
@WeichenXu123 Nearly LGTM! A couple questions about Windows (can we support it and warn the user about possible issues?) and MLServer.

mlflow/pyfunc/__init__.py
```diff
@@ -989,15 +1020,25 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
 
     env_manager = _EnvManager.from_string(env_manager)
 
+    if env_manager != "local" and os.name == "nt":
```
harupy, line 1023:
```suggestion
    if env_manager is _EnvManager.LOCAL  and os.name == "nt":
```

mlflow/pyfunc/__init__.py
```diff
@@ -1047,10 +1088,18 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
...
         # Prepare restored environment in driver side if possible.
+        # Note: In databricks runtime, because databricks notebook cell output cannot capture
+        # child process output, so that set capture_output to be True so that when conda prepare
+        # env command failed, the exception message will include command stdout/stderr output.
+        # otherwise user have to check cluster driver log to find command stdout/stderr output.
         if env_manager is _EnvManager.CONDA:
             _get_flavor_backend(
-                local_model_path, env_manager=_EnvManager.CONDA, install_mlflow=False
-            ).prepare_env(model_uri=local_model_path, capture_output=False)
+                local_model_path, no_conda=False, install_mlflow=False,
```
harupy, line 1097:
```suggestion
                local_model_path, env_manager=_EnvManager.CONDA, install_mlflow=False,
```

mlflow/pyfunc/__init__.py
```diff
@@ -1047,10 +1088,18 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
...
+                conda_env_root_dir=conda_env_root_dir
+            ).prepare_env(
+                model_uri=local_model_path,
+                capture_output=is_in_databricks_runtime()
```
harupy, line 1101: Curious why we only capture output on databricks.

mlflow/utils/conda.py
```diff
@@ -149,6 +175,7 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
                         "--file",
                         conda_env_path,
                     ],
+                    env=additional_env,
```
harupy, line 178:
```suggestion
                    extra_env=additional_env,
```

I think we should use `extra_env` here.

mlflow/utils/conda.py
```diff
@@ -164,6 +191,7 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
                         project_env_name,
                         "python",
                     ],
+                    env=additional_env,
```
harupy, line 194: Same here: https://github.com/mlflow/mlflow/pull/5561/files#r842400517

mlflow/pyfunc/__init__.py
WeichenXu123: The benefit of "do not capture output" is the output will be printed immediately, otherwise you have to wait conda command fail and suddenly get all output printed (included in error message).

I prefer only enable capture output in databricks because databricks notebook cell output cannot print out output generated by child process, if without "capture_output", user have to open cluster driver log to see the child process output.

mlflow/pyfunc/__init__.py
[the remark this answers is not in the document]
WeichenXu123: I enable it under windows.
the related warning are printed here:
https://github.com/mlflow/mlflow/blob/e7caa0a1ffd47255f52b2818fad49d34ed332d99/mlflow/pyfunc/__init__.py#L1042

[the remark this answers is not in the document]
WeichenXu123: > are race conditions still a problem if we isolate the pip package cache?

No.

mlflow/pyfunc/mlserver.py
WeichenXu123: The "get_cmd" function does not need the "conda_env_root_dir" argument, because in `backend.serve` it calls `_execute_in_conda_env` with env_root_dir setting. Removed the argument for both mlserver and scoring_server.

mlflow/pyfunc/backend.py
```diff
@@ -27,13 +28,27 @@ class PyFuncBackend(FlavorBackend):
...
+        config,
+        workers=1,
+        env_manager=_EnvManager.CONDA,
+        install_mlflow=False,
+        env_root_dir=None,
+        **kwargs,
     ):
+        """
+        :param: env_root_dir: Root path for conda env. If None, use default one. Note if this is
+                              set, conda package cache path becomes "env_root_path/conda_pkgs"
```
harupy, line 41: What is `env_root_path`?

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -349,7 +349,10 @@ def _serve(model_uri, port, host):
 
 
 def get_cmd(
-    model_uri: str, port: int = None, host: int = None, nworkers: int = None
+    model_uri: str,
+    port: int = None,
```
harupy, line 353:
```suggestion
    port: Optional[int] = None,
```

mlflow/utils/conda.py
```diff
@@ -84,12 +85,43 @@ def _get_conda_executable_for_create_env():
...
+    if env_root_dir is not None:
+        # Create isolated conda package cache dir "conda_pkgs" under the env_root_dir
+        # for each python process.
+        # Note: shared conda package cache dir causes race condition issues:
+        # See https://github.com/conda/conda/issues/8870
+        # See https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#specify-environment-directories-envs-dirs
+        # and https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#specify-package-directories-pkgs-dirs
+
+        conda_env_path = os.path.join(env_root_dir, "conda_envs")
+        conda_pkgs_path = os.path.join(env_root_dir, "conda_pkgs")
```
harupy, line 105:
```suggestion
        conda_pkgs_path = os.path.join(env_root_dir, "conda_pkgs")
```

Can we include `cache` in the directory name?

mlflow/utils/file_utils.py
```diff
@@ -483,3 +485,54 @@ def _handle_readonly_on_windows(func, path, exc_info):
         raise exc_value
     os.chmod(path, stat.S_IWRITE)
     func(path)
+
+
+_TMP_DIR = None
```
harupy, line 490: Can we use a more descriptive name?

mlflow/utils/conda.py
```diff
@@ -84,12 +85,43 @@ def _get_conda_executable_for_create_env():
...
+        pip_cache_dir = os.path.join(env_root_dir, "pip_pkgs")
```
harupy, line 106: I think we should be careful with terminology here. "pip cache" sounds like [pip's wheel cache](https://pip.pypa.io/en/stable/cli/pip_cache/).

mlflow/utils/conda.py
```diff
@@ -84,12 +85,43 @@ def _get_conda_executable_for_create_env():
...
+    prc = process._exec_cmd(
+        [get_conda_bin_executable("conda"), "env", "list", "--json"], extra_env=extra_env
+    )
     return list(map(os.path.basename, json.loads(prc.stdout).get("envs", [])))
 
 
-def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
+def _get_conda_extra_envs(env_root_dir=None):
+    if env_root_dir is not None:
+        # Create isolated conda package cache dir "conda_pkgs" under the env_root_dir
```
harupy, line 97:
```suggestion
    if env_root_dir is None:
        return None
    # Create isolated conda package cache dir "conda_pkgs" under the env_root_dir
```

nit

mlflow/utils/conda.py
```diff
@@ -84,12 +85,43 @@ def _get_conda_executable_for_create_env():
...
+        pip_cache_dir = os.path.join(env_root_dir, "pip_pkgs")
+
+        os.makedirs(conda_env_path, exist_ok=True)
+        os.makedirs(conda_pkgs_path, exist_ok=True)
+        os.makedirs(pip_cache_dir, exist_ok=True)
+
+        return {
+            "CONDA_ENVS_PATH": conda_env_path,
+            "CONDA_PKGS_DIRS": conda_pkgs_path,
+            "PIP_CACHE_DIR": pip_cache_dir,
```
harupy, line 115: Is this environment variable documented somewhere in pip's doc?

mlflow/utils/conda.py
```diff
@@ -183,12 +235,19 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
             except Exception as e:
                 _logger.warning(
                     f"Removing conda env '{project_env_name}' failed (error: {repr(e)})."
                 )
             raise
+    else:
+        if env_root_dir is not None:
+            _logger.info(
+                "Reusing cached conda environment at %s",
+                os.path.join(os.path.join(env_root_dir, "conda_envs", project_env_name)),
```
harupy, line 250:
```suggestion
               os.path.join(env_root_dir, "conda_envs", project_env_name),
```

mlflow/pyfunc/backend.py
```diff
@@ -250,13 +277,17 @@ def _execute_in_conda_env(
...
     :param stdout: Redirect server stdout
     :param stderr: Redirect server stderr
+    :param: env_root_dir: Root path for conda env. If None, use default one. Note if this is
+                          set, conda package cache path becomes "env_root_path/conda_pkgs"
+                          instead of the global package cache path, and pip package cache path
+                          becomes "env_root_path/pip_pkgs" instead of the global package cache
+                          path.
     """
     if command_env is None:
         command_env = os.environ
```
harupy, line 287: Can we add `.copy` here?

mlflow/utils/conda.py
WeichenXu123: See https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

> pip’s command line options can be set with environment variables using the format PIP_<UPPER_LONG_NAME> . Dashes (-) have to be replaced with underscores (_).

Because pip has a command argument of "--cache-dir" , so setting env "PIP_CACHE_DIR" is equivalent to that argument according to the doc.

mlflow/utils/conda.py
WeichenXu123: I choose the same name with the pip --cache-dir argument

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -349,7 +349,10 @@ def _serve(model_uri, port, host):
 
 
 def get_cmd(
-    model_uri: str, port: int = None, host: int = None, nworkers: int = None
+    model_uri: str,
+    port: Optional[int] = None,
+    host: Optional[int] = None,
+    nworkers: Optional[int] = None,
```
harupy, line 355: btw do we need this change? It looks like the function signature is not changed at all.

mlflow/utils/conda.py
harupy: ok

mlflow/utils/conda.py
```diff
@@ -133,10 +176,27 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
+    if env_root_dir is not None:
+        # Generate the full path of the conda env
+        # Checking whether env was created using full path instead of env name,
+        # otherwise it might match existing env under other different env_root_dir.
+        project_env_path = os.path.normpath(
+            os.path.join(env_root_dir, _CONDA_ENVS_DIR, project_env_name)
+        )
+    else:
+        project_env_path = project_env_name
+    if project_env_path not in _list_conda_environments(conda_extra_envs):
```
harupy, line 198: Does `_list_conda_environments` return a list of paths?

mlflow/utils/conda.py
harupy: got it!

mlflow/utils/conda.py
```diff
@@ -183,12 +254,19 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
                 _logger.warning(
-                    f"Removing conda env '{project_env_name}' failed (error: {repr(e)})."
+                    f"Removing conda env '{project_env_path}' failed (error: {repr(e)})."
                 )
             raise
+    else:
+        if env_root_dir is not None:
+            _logger.info(
+                "Reusing cached conda environment at %s",
+                os.path.join(os.path.join(env_root_dir, _CONDA_ENVS_DIR, project_env_name)),
```
harupy, line 269:
```suggestion
                os.path.join(env_root_dir, _CONDA_ENVS_DIR, project_env_name),
```

mlflow/utils/conda.py
WeichenXu123: I already updated `_list_conda_environments` to support this

mlflow/pyfunc/backend.py
```diff
@@ -27,13 +28,27 @@ class PyFuncBackend(FlavorBackend):
...
+        self,
+        config,
+        workers=1,
+        env_manager=_EnvManager.CONDA,
+        install_mlflow=False,
+        env_root_dir=None,
+        **kwargs,
     ):
+        """
+        :param env_root_dir: Root path for conda env. If None, use default config. Note if this is
```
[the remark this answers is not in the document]
WeichenXu123: i.e using config in "conda info" listed.

mlflow/utils/conda.py
```diff
@@ -133,10 +181,29 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
+
+    if env_root_dir is not None:
+        # Generate the full path of the conda env
+        # Checking whether env was created using full path instead of env name,
+        # otherwise it might match existing env under other different env_root_dir.
+        project_env_path = os.path.normpath(
+            os.path.join(env_root_dir, _CONDA_ENVS_DIR, project_env_name)
+        )
+    else:
+        project_env_path = project_env_name
```
harupy, line 202: Can we process everything using a path? Using both name and path is a bit confusing and complicates the code.

WeichenXu123: If so , the default case (i.e. without CONDA_ENVS_PATH setting), we need run `conda info --json` and parse the result "envs_dirs" entry (It might includes multiple dirs and we need to check the first dir which is writable). The logic is also a bit complex.

mlflow/utils/conda.py
```diff
@@ -133,10 +181,29 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
+
     project_env_name = _get_conda_env_name(conda_env_path, env_id)
-    if project_env_name not in env_names:
-        _logger.info("=== Creating conda environment %s ===", project_env_name)
+    if env_root_dir is not None:
+        # Append a suffix "-isolated" because if a conda env name exist in
+        # default conda env root dir, then if we set new "CONDA_ENVS_PATH" and run "conda env create"
+        # with the same conda env name, "CondaValueError: prefix already exists" error will
+        # be raised.
+        project_env_name = project_env_name + "-isolated"
```
harupy, line 192: What would happen if an environment named `project_env_name + "-isolated"` exists in the default conda env root?

mlflow/pyfunc/mlserver.py
```diff
@@ -1,13 +1,17 @@
...
 MLServerDefaultModelName = "mlflow-model"
 
 
 def get_cmd(
-    model_uri: str, port: int = None, host: str = None, nworkers: int = None
+    model_uri: str,
+    port: Optional[int] = None,
+    host: Optional[str] = None,
+    nworkers: Optional[int] = None,
 ) -> Tuple[str, Dict[str, str]]:
```
harupy, line 15: Does this change relate to this PR?

mlflow/pyfunc/__init__.py
```diff
@@ -1047,10 +1058,17 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
...
+        # otherwise user have to check cluster driver log to find command stdout/stderr output.
         if env_manager is _EnvManager.CONDA:
             _get_flavor_backend(
-                local_model_path, env_manager=_EnvManager.CONDA, install_mlflow=False
-            ).prepare_env(model_uri=local_model_path, capture_output=False)
+                local_model_path,
+                env_manager=_EnvManager.CONDA,
+                install_mlflow=False,
+                env_root_dir=conda_env_root_dir,
+            ).prepare_env(model_uri=local_model_path, capture_output=is_in_databricks_runtime())
```
dbczumar, line 1071: Is `capture_output=True` useful for non-Databricks use cases?

mlflow/pyfunc/backend.py
dbczumar, line 40:
```suggestion
        :param env_root_dir: Root path for conda env. If None, use Conda's default environments directory. Note if this is
```

mlflow/pyfunc/backend.py
```diff
@@ -250,13 +275,17 @@ def _execute_in_conda_env(
                         If False, return the server process `Popen` instance immediately.
     :param stdout: Redirect server stdout
     :param stderr: Redirect server stderr
+    :param env_root_dir: Root path for conda env. If None, use default config. Note if this is
```
dbczumar, line 278:
```suggestion
    :param env_root_dir: Root path for conda env. If None, use Conda's default environments directory. Note if this is
```

mlflow/pyfunc/mlserver.py
dbczumar: Doesn't seem like it, but it doesn't hurt.

mlflow/pyfunc/scoring_server/__init__.py
dbczumar: +1, Let's put this in a follow-up PR if we want to adjust the type hints.

mlflow/utils/conda.py
```diff
@@ -84,12 +85,54 @@ def _get_conda_executable_for_create_env():
     return conda_env_create_path
 
 
-def _list_conda_environments():
-    prc = process._exec_cmd([get_conda_bin_executable("conda"), "env", "list", "--json"])
-    return list(map(os.path.basename, json.loads(prc.stdout).get("envs", [])))
+def _list_conda_environments(conda_extra_envs=None, return_full_path=False):
```
dbczumar, line 88: Can we add a docstring for this to explain the input parameters and their behaviors?

mlflow/utils/conda.py
```diff
@@ -84,12 +85,54 @@ def _get_conda_executable_for_create_env():
...
+    else:
+        return list(map(os.path.basename, result))
+
+
+_CONDA_ENVS_DIR = "conda_envs"
+_CONDA_CACHE_PKGS_DIR = "conda_cache_pkgs"
+_PIP_CACHE_DIR = "pip_cache_pkgs"
+
+
+def _get_conda_extra_envs(env_root_dir=None):
```
dbczumar, line 106: Can we add a docstring for this method?

mlflow/utils/conda.py
```diff
@@ -101,6 +144,11 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
                    environment after the environment has been activated.
     :param capture_output: Specify the capture_output argument while executing the
                            "conda env create" command.
+    :param env_root_dir Root path for conda env. If None, use default one. Note if this is
```
dbczumar, line 147:
```suggestion
    :param env_root_dir Root path for conda env. If None, use Conda's default environments directory. Note if this is
```

mlflow/utils/conda.py
dbczumar, line 106:
```suggestion
def _get_conda_extra_env_vars(env_root_dir=None):
```

mlflow/utils/conda.py
```diff
@@ -133,10 +181,29 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
             )
         )
 
-    env_names = _list_conda_environments()
+    conda_extra_envs = _get_conda_extra_envs(env_root_dir)
```
dbczumar, line 184:
```suggestion
    conda_extra_env_vars = _get_conda_extra_env_vars(env_root_dir)
```

mlflow/utils/conda.py
```diff
@@ -133,10 +181,29 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
+
     project_env_name = _get_conda_env_name(conda_env_path, env_id)
-    if project_env_name not in env_names:
-        _logger.info("=== Creating conda environment %s ===", project_env_name)
+    if env_root_dir is not None:
+        # Append a suffix "-isolated" because if a conda env name exist in
+        # default conda env root dir, then if we set new "CONDA_ENVS_PATH" and run "conda env create"
+        # with the same conda env name, "CondaValueError: prefix already exists" error will
+        # be raised.
```
dbczumar, line 191: Are you sure this happens? Is there a way to exclude the default conda env root dir? The docs (https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#specify-environment-directories-envs-dirs) seem to imply that the default location should no longer be checked if `CONDA_ENVS_PATH` is specified:

```
Specify environment directories (envs_dirs)[](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#specify-environment-directories-envs-dirs)
Specify directories in which environments are located. If this key is set, the root prefix envs_dir is not used unless explicitly included. This key also determines where the package caches are located.
```

mlflow/utils/conda.py
dbczumar: Can we instead do `conda info --envs --json`, look at the entries under the `"envs"` key, and then find the entry whose `os.path.basename() == project_env_name`? 

Here's an example of the `envs` key content:

```
"envs": [
    "/Users/corey.zumar/opt/anaconda3",
    "/Users/corey.zumar/opt/anaconda3/envs/base2",
    "/Users/corey.zumar/opt/anaconda3/envs/foo",
    "/Users/corey.zumar/opt/anaconda3/envs/javapydiff",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-1b55d4c51bfa467f642492d4eb4474cda2c85ee1",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-2db1ebf9ea6f978512f24e8516507b38fb7ad1fc",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-515e205e1ab623cdd02144de8e4d25085796e193",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-54ade87f6825f3ab12eac103b777d6cda0d13852",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-795b1deee44b2fe066ea40822019bf3c00b1cfb0",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-7ef5a6c33fa6db6e310779594c9919d72af7ea28",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-9acaa695f270d7ef672ecf4e5b6beeb77cd423b1",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-a86033905519765fe1d87442150b5284ca965398",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-af9fca66a859c021675ec43c9c75c12080cfc17e",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-b6ecf73bc09f0108678380d09092e88bc31bbf3e",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-cbb4f40ff67c4094fa0737599b864e21d2ff364c",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-d080e661670d8d0df6823d2cc5af762cf782e60b",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-e9c06b6c2e9a28c4146182cd65ccda34e27be6af",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-env",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflow-env2",
    "/Users/corey.zumar/opt/anaconda3/envs/mlflowskinny",
    "/Users/corey.zumar/opt/anaconda3/envs/pipdeps",
    "/Users/corey.zumar/opt/anaconda3/envs/py36_failtest",
    "/Users/corey.zumar/opt/anaconda3/envs/python=3.8",
    "/Users/corey.zumar/opt/anaconda3/envs/skinny",
    "/Users/corey.zumar/opt/anaconda3/envs/skinny2",
    "/Users/corey.zumar/opt/anaconda3/envs/tf1",
    "/Users/corey.zumar/opt/anaconda3/envs/tf204",
    "/Users/corey.zumar/opt/anaconda3/envs/tf2042",
    "/Users/corey.zumar/opt/anaconda3/envs/tf24",
    "/Users/corey.zumar/opt/anaconda3/envs/tf25",
    "/Users/corey.zumar/opt/anaconda3/envs/xgbrecent"
  ],
```

This way, like @harupy suggested, we don't need any special casing.

mlflow/utils/conda.py
dbczumar: Instead of appending `isolated`, can we hash the environment root directory and append it for uniqueness?

mlflow/utils/conda.py
dbczumar: @WeichenXu123 Is it true that conda environment paths must have unique basenames?

E.g. the following would be invalid?

```
"/Users/corey.zumar/opt/anaconda3/envs/tf24",
"/Users/corey.zumar/tmp/my_other_envs/tf24",
```

This seems to be implied by your comment here: https://github.com/mlflow/mlflow/pull/5561/files#diff-69f4364af03d692af1d39ecf262500cf91306991c5a2dc52dde9844f1a3f7f1bR188-R191.

If this is the case, can't we list the environment basenames (i.e. take `os.path.basename()` for each of the outputs of `_list_conda_environments()`) and just check if `project_env_name` is in the list?

mlflow/utils/conda.py
dbczumar: Please see https://github.com/mlflow/mlflow/pull/5561/files#r844817217 first to determine whether the changes to this method are necessary.

mlflow/utils/file_utils.py
dbczumar: +1

dbczumar: I think an inline comment above the variable name is a better choice. It makes sense that this is called `_TMP_DIR` because it's referenced by `get_or_create_tmp_dir`.

mlflow/utils/file_utils.py
```diff
@@ -483,3 +485,54 @@ def _handle_readonly_on_windows(func, path, exc_info):
...
+    """
+    from mlflow.utils.databricks_utils import is_in_databricks_runtime, get_repl_id
+    from mlflow.utils.nfs_on_spark import get_nfs_cache_root_dir
+
+    global _TMP_NFS_DIR
+
+    nfs_root_dir = get_nfs_cache_root_dir()
+
+    if _TMP_NFS_DIR is None:
+        if is_in_databricks_runtime():
```
dbczumar, line 529: Shall we also check that `get_repl_id() is not None` like we do in `get_or_create_tmp_dir()`?

mlflow/utils/file_utils.py
```diff
@@ -483,3 +485,54 @@ def _handle_readonly_on_windows(func, path, exc_info):
...
+            # Note: In databricks, atexit hook does not work.
+            # The /nfs_root_dir/repl_tmp_data/{repl_id} directory will be removed once databricks notebook detached.
+            _TMP_NFS_DIR = os.path.join(nfs_root_dir, "repl_tmp_data", get_repl_id())
```
dbczumar, line 532: Nit: Why the different construction of the path in this method compared to `get_or_create_tmp_dir()` (https://github.com/mlflow/mlflow/pull/5561/files#diff-6f28a867435ec4c1fc06706110a3b515fe096be3af1fb20ab70f3db5315f828fR505). I think both should use `os.path.join`.

dbczumar (commented):
@WeichenXu123 This is almost ready to merge! One major question: https://github.com/mlflow/mlflow/pull/5561/files#r844817217 and a few minor comments.

mlflow/utils/conda.py
WeichenXu123: > Instead of appending isolated, can we hash the environment root directory path and append it for uniqueness?

Good idea

mlflow/utils/conda.py
WeichenXu123: > Is it true that conda environment paths must have unique basenames?

That's the hidden bug I found yesterday.

(1) Between default conda env root path and custom conda env root path, basenames  must be unique.
(2) Between multiple custom conda env root paths, basenames can be the same.

e.g. You can try run following commands one by one

```
conda env create -f conda.yaml -n name_1
CONDA_ENVS_PATH=/tmp/a1 conda env create -f conda.yaml -n name_1  # conflict name failure
CONDA_ENVS_PATH=/tmp/a1 conda env create -f conda.yaml -n name_1-isolated # works
CONDA_ENVS_PATH=/tmp/a2 conda env create -f conda.yaml -n name_1-isolated # still works, although name conflict with the one in `/tmp/a1`

conda env list --json  # NOTE: it will list all the envs (under all custom root paths) created in above....
```

WeichenXu123: But, I adopt your suggestion:
https://github.com/mlflow/mlflow/pull/5561/files#r844813894

> Instead of appending isolated, hash the environment root directory path and append it for uniqueness

This address the issue in a simpler way

mlflow/pyfunc/__init__.py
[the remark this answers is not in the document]
WeichenXu123: Addressed! See `cache_return_value_per_process` decorator and it is applied on `get_or_create_nfs_tmp_dir`, `get_or_create_tmp_dir`, `_get_or_create_env_root_dir`

mlflow/pyfunc/__init__.py
WeichenXu123: See my explanation here.
https://github.com/mlflow/mlflow/pull/5561#discussion_r842723083

mlflow/utils/process.py
```diff
@@ -68,3 +69,29 @@ def _exec_cmd(
...
+            prev_value, prev_pid = _per_process_value_cache_map.get((fn, args))
+            if os.getpid() == prev_pid:
+                return prev_value
+
+        new_value = fn(*args)
+        new_pid = os.getpid()
+        _per_process_value_cache_map[(fn, args)] = (new_value, new_pid)
+        return new_value
+
+    return wrapped_fn
```
WeichenXu123, line 97: This addressed the issue mentioned in https://github.com/mlflow/mlflow/pull/5561#discussion_r842384536

mlflow/pyfunc/__init__.py
dbczumar: Can we update the inline comment to explain why we prefer `False` outside of Databricks?

tests/utils/test_process_utils.py
```diff
@@ -0,0 +1,46 @@
...
+import tempfile
+import pytest
+
+from mlflow.utils.process import cache_return_value_per_process
+
+
+@pytest.mark.skipif(
+    os.name == "nt",
+    reason="Windows does not support fork",
+)
```
dbczumar, line 11: Can we use Python's `multiprocessing` module instead of os.fork() in order to test Windows? https://docs.python.org/3/library/multiprocessing.html

dbczumar (approved):
LGTM with 2 small comments (https://github.com/mlflow/mlflow/pull/5561/files#r845622185, https://github.com/mlflow/mlflow/pull/5561#discussion_r845625674). Thanks @WeichenXu123 ! Feel free to merge once they're addressed in order to make it into the next release :).

mlflow/utils/process.py
```diff
@@ -68,3 +69,29 @@ def _exec_cmd(
...
+
+def cache_return_value_per_process(fn):
+    """
+    A decorator which globally cache the return value of the decorated function.
+    But if current process forked out a new child process, in child process,
+    old cache values are invalidated.
+    """
+
+    @functools.wraps(fn)
+    def wrapped_fn(*args):
```
harupy, line 86: Do we need to add kwargs here?
 
```suggestion
    def wrapped_fn(*args, **kwargs):
```

tests/utils/test_process_utils.py
```diff
@@ -0,0 +1,48 @@
...
+import tempfile
+import pytest
+
+from mlflow.utils.process import cache_return_value_per_process
+from multiprocessing import Process, Queue
+
+
+@cache_return_value_per_process
+def _gen_temp_dir1(_):
+    return tempfile.mkdtemp()
```
harupy, line 11: I think we should return a random string here because `tempfile.mkdtemp` doesn't clean up the created experiment.

mlflow/utils/process.py
WeichenXu123: I prefer to force the decorated functions calls with only positional arguments., otherwise,
suppose:
```
def f1(a, b):
   return a + b
```
if allow using kwargs, then we might calls with `f1(1, b=3)` or `f1(a=1, b=3)` or `f1(1, 3)`, which generates different args/kwargs, but they should be treated as the same call.

tests/utils/test_process_utils.py
harupy: Can we add a test the case where a function to decorate has no arguments?

mlflow/utils/process.py
WeichenXu123: I can add a checking to forbid the decorated function called with key-word arguments.

mlflow/utils/process.py
```diff
@@ -68,3 +69,29 @@ def _exec_cmd(
...
+        if (fn, args) in _per_process_value_cache_map:
+            prev_value, prev_pid = _per_process_value_cache_map.get((fn, args))
+            if os.getpid() == prev_pid:
+                return prev_value
+
+        new_value = fn(*args)
+        new_pid = os.getpid()
+        _per_process_value_cache_map[(fn, args)] = (new_value, new_pid)
```
harupy, line 94: Probably not a big issue, but what would happen if `args` contains unhashable objects (e.g. `list`)?

WeichenXu123: It is not supported. I will document it about the restriction.

mlflow/utils/conda.py
```diff
@@ -61,11 +61,22 @@ def get_conda_bin_executable(executable_name):
...
+        conda_env_contents = ""
+
     if env_id:
         conda_env_contents += env_id
-    return "mlflow-%s" % hashlib.sha1(conda_env_contents.encode("utf-8")).hexdigest()
+
+    env_name = "mlflow-%s" % hashlib.sha1(conda_env_contents.encode("utf-8")).hexdigest()
+    if env_root_dir:
+        env_root_dir = os.path.normpath(env_root_dir)
+        env_name += "-%s" % hashlib.sha1(env_root_dir.encode("utf-8")).hexdigest()
```
harupy, line 77:
```suggestion
    hash_input = conda_env_contents + (os.path.normpath(env_root_dir) if env_root_dir else "")
    env_name = "mlflow-%s" % hashlib.sha1(hash_input.encode("utf-8")).hexdigest()
```

Can we create the env suffix like above?

mlflow/utils/process.py
harupy: It might be helpful to indicate this function cannot be applied to a function that takes unhashable objects.

mlflow/utils/conda.py
```diff
@@ -84,12 +95,58 @@ def _get_conda_executable_for_create_env():
     return conda_env_create_path
 
 
-def _list_conda_environments():
-    prc = process._exec_cmd([get_conda_bin_executable("conda"), "env", "list", "--json"])
+def _list_conda_environments(conda_extra_envs=None):
+    """
+    Return a list of names of conda environments.
+
+    :param conda_extra_envs: extra environment variables generated by `_get_conda_extra_envs`.
```
harupy, line 102: `_get_conda_extra_envs` doesn't seem to exist.

harupy: Does `conda_extra_envs` has to be created by `_get_conda_extra_envs`?

mlflow/utils/conda.py
harupy: Or are we intentionally splitting the suffixes?

mlflow/utils/conda.py
```diff
@@ -133,10 +191,18 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
+    # Include the env_root_dir hash in the project_env_name,
+    # this is for avoid conda env name conflicts between different CONDA_ENVS_PATH.
+    project_env_name = _get_conda_env_name(conda_env_path, env_id=env_id, env_root_dir=env_root_dir)
+    if env_root_dir is not None:
+        project_env_path = os.path.join(env_root_dir, _CONDA_ENVS_DIR, project_env_name)
+    else:
+        project_env_path = project_env_name
+
+    if project_env_name not in _list_conda_environments(conda_extra_env_vars):
+        _logger.info("=== Creating conda environment %s ===", project_env_path)
```
harupy, line 205:
```suggestion
        _logger.info("=== Creating conda environment %s ===", project_env_name)
```

Can we use `project_env_name`?

mlflow/utils/conda.py
WeichenXu123: I prefer to hash the "conda_env_contents" part and "env_root_dir" part separately, it is easier for debug, i.e,
"mlflow-{conda_env_contents_hash}-{env_root_dir_hash}"
If "conda_env_contents_hash" part is the same, it means it is conda env based on the same conda.yaml file.
This is useful for debugging.
But if we connect them and hash together, then according to the env name we cannot distinguish whether they are based on the same  conda.yaml file or not

mlflow/utils/conda.py
WeichenXu123: I prefer print full path because print full path helps debugging.

mlflow/utils/file_utils.py
```diff
@@ -483,3 +486,46 @@ def _handle_readonly_on_windows(func, path, exc_info):
...
+    Get or create a temporary NFS directory which will be removed once python process exit.
+    """
+    from mlflow.utils.databricks_utils import is_in_databricks_runtime, get_repl_id
+    from mlflow.utils.nfs_on_spark import get_nfs_cache_root_dir
+
+    nfs_root_dir = get_nfs_cache_root_dir()
+
+    if is_in_databricks_runtime() and get_repl_id() is not None:
+        # Note: In databricks, atexit hook does not work.
+        # The /nfs_root_dir/repl_tmp_data/{repl_id} directory will be removed once databricks
```
harupy, line 523:
```suggestion
        # The {nfs_root_dir}/repl_tmp_data/{repl_id} directory will be removed once databricks
```

nit: can we indicate the `nfs_root_dir` part can vary?

mlflow/utils/conda.py
harupy: ok, can you add a comment to indicate that?

tests/models/test_cli.py
```diff
@@ -535,3 +539,54 @@ def test_env_manager_unsupported_value():
...
+        (
+            env_root2_path,
+            model1_path,
+            "1.0.1",
+        ),  # test the same env created in different env root path.
+        (
+            env_root1_path,
+            model2_path,
+            "1.0.2",
+        ),  # test different env created in the same env root path.
```
harupy, line 568: Do these case create different conda environments?

harupy, line 568: Curious what should happen after running these? 3 conda environments should be created?

tests/utils/test_process_utils.py
```diff
@@ -0,0 +1,63 @@
...
+def test_cache_return_value_per_process():
+
+    path1 = _gen_random_str1(True)
+    path2 = _gen_random_str1(True)
+
+    no_arg_path1 = _gen_random_no_arg()
+    no_arg_path2 = _gen_random_no_arg()
+    assert no_arg_path1 == no_arg_path2
+
+    assert path1 == path2
```
harupy, line 40: I think this line should be located after line 34.

tests/models/test_cli.py
harupy: If so, does the current test code guarantee that?

mlflow/pyfunc/backend.py
```diff
@@ -27,13 +28,27 @@ class PyFuncBackend(FlavorBackend):
...
+        workers=1,
+        env_manager=_EnvManager.CONDA,
+        install_mlflow=False,
+        env_root_dir=None,
+        **kwargs,
     ):
+        """
+        :param env_root_dir: Root path for conda env. If None, use Conda's default environments
+                             directory. Note if this is set, conda package cache path becomes
+                             "env_root_dir/conda_cache_pkgs" instead of the global package cache
```
harupy, line 42:
```suggestion
                             "{env_root_dir}/conda_cache_pkgs" instead of the global package cache
```

mlflow/utils/conda.py
harupy: Ok

harupy, line 205:
```suggestion
    if project_env_name in _list_conda_environments(conda_extra_env_vars):
        ...
        return projece_env_name

    _logger.info("=== Creating conda environment %s ===", project_env_path)
```

Let's exit here if the environment already exists.

mlflow/pyfunc/backend.py
```diff
@@ -27,13 +28,27 @@ class PyFuncBackend(FlavorBackend):
...
+                             path, and pip package cache path becomes "env_root_dir/pip_cache_pkgs"
```
harupy, line 43:
```suggestion
                             path, and pip package cache path becomes "{env_root_dir}/pip_cache_pkgs"
```

same here https://github.com/mlflow/mlflow/pull/5561/files#r845689485

tests/models/test_cli.py
WeichenXu123: added test

mlflow/utils/conda.py
```diff
@@ -133,62 +193,77 @@ def get_or_create_conda_env(conda_env_path, env_id=None, capture_output=False):
...
+    # Include the env_root_dir hash in the project_env_name,
+    # this is for avoid conda env name conflicts between different CONDA_ENVS_PATH.
+    project_env_name = _get_conda_env_name(conda_env_path, env_id=env_id, env_root_dir=env_root_dir)
+    if env_root_dir is not None:
+        project_env_path = os.path.join(env_root_dir, _CONDA_ENVS_DIR, project_env_name)
+    else:
+        project_env_path = project_env_name
+
+    if project_env_name in _list_conda_environments(conda_extra_env_vars):
+        _logger.info("Reusing cached conda environment %s", project_env_path)
```
harupy, line 207:
```suggestion
        _logger.info("%s already exists", project_env_path)
```

harupy (approved):
LGTM once the remaining comments are addressed.

mlflow/utils/conda.py
WeichenXu123: no. updated code

mlflow/utils/conda.py
```diff
@@ -61,11 +61,24 @@ def get_conda_bin_executable(executable_name):
...
+
     if env_id:
         conda_env_contents += env_id
-    return "mlflow-%s" % hashlib.sha1(conda_env_contents.encode("utf-8")).hexdigest()
+
+    env_name = "mlflow-%s" % hashlib.sha1(conda_env_contents.encode("utf-8")).hexdigest()
+    if env_root_dir:
+        env_root_dir = os.path.normpath(env_root_dir)
+        # Generate env name with format "mlflow-{conda_env_contents_hash}-{env_root_dir_hash}"
+        # hash `conda_env_contents` and `env_root_dir` separately helps debugging
```
harupy, line 78:
```suggestion
        # Hashing `conda_env_contents` and `env_root_dir` separately helps debugging
```

mlflow/utils/conda.py
```diff
@@ -84,12 +97,58 @@ def _get_conda_executable_for_create_env():
     return conda_env_create_path
 
 
-def _list_conda_environments():
-    prc = process._exec_cmd([get_conda_bin_executable("conda"), "env", "list", "--json"])
+def _list_conda_environments(conda_extra_envs=None):
```
harupy, line 100:
```suggestion
def _list_conda_environments(extra_envs=None):
```

mlflow/utils/conda.py
```diff
@@ -84,12 +97,58 @@ def _get_conda_executable_for_create_env():
...
+    """
+    Return a list of names of conda environments.
+
+    :param conda_extra_envs: extra environment variables for running "conda env list" command.
```
harupy, line 104:
```suggestion
    :param extra_envs: extra environment variables for running "conda env list" command.
```

mlflow/utils/process.py
```diff
@@ -68,3 +69,37 @@ def _exec_cmd(
...
     return prc
+
+
+# A global map storing (function, args_tuple) --> (value, pid)
+_per_process_value_cache_map = {}
+
+
+def cache_return_value_per_process(fn):
+    """
+    A decorator which globally cache the return value of the decorated function.
```
harupy, line 80:
```suggestion
    A decorator which globally caches the return value of the decorated function.
```

mlflow/utils/file_utils.py
```diff
@@ -483,3 +486,46 @@ def _handle_readonly_on_windows(func, path, exc_info):
...
+    """
+    from mlflow.utils.databricks_utils import is_in_databricks_runtime, get_repl_id
+    from mlflow.utils.nfs_on_spark import get_nfs_cache_root_dir
+
+    nfs_root_dir = get_nfs_cache_root_dir()
+
+    if is_in_databricks_runtime() and get_repl_id() is not None:
+        # Note: In databricks, atexit hook does not work.
+        # The {nfs_root_dir}/repl_tmp_data/{repl_id} directory will be removed once databricks
+        # notebook detached.
```
harupy, line 524:
```suggestion
        # notebook detaches.
```

mlflow/utils/file_utils.py
```diff
@@ -483,3 +486,46 @@ def _handle_readonly_on_windows(func, path, exc_info):
...
+def get_or_create_tmp_dir():
+    """
+    Get or create a temporary directory which will be removed once python process exit.
+    """
+    from mlflow.utils.databricks_utils import is_in_databricks_runtime, get_repl_id
+
+    if is_in_databricks_runtime() and get_repl_id() is not None:
+        # Note: For python process attached to databricks notebook, atexit does not work.
+        # The /tmp/repl_tmp_data/{repl_id} directory will be removed once databricks notebook
+        # detached.
```
harupy, line 501:
```suggestion
        # detaches.
```

mlflow/pyfunc/__init__.py
```diff
@@ -1047,10 +1054,21 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
 
     if not should_use_spark_to_broadcast_file:
         # Prepare restored environment in driver side if possible.
+        # Note: In databricks runtime, because databricks notebook cell output cannot capture
+        # child process output, so that set capture_output to be True so that when conda prepare
+        # env command failed, the exception message will include command stdout/stderr output.
```
harupy, line 1059:
```suggestion
        # child process output, so that set capture_output to be True so that when `conda prepare
        # env` command failed, the exception message will include command stdout/stderr output.
```

mlflow/pyfunc/__init__.py
```diff
@@ -1047,10 +1054,21 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
...
+        # otherwise user have to check cluster driver log to find command stdout/stderr output.
+        # In non-databricks runtime, set capture_output to be False, because the benefit of
+        # "do not capture output" is the output will be printed immediately, otherwise you have
```
harupy, line 1062:
```suggestion
        # "capture_output=False" is the output will be printed immediately, otherwise you have
```

PR #5562, opened 2022-03-30
turn sphinx nitpicky mode on and force a failure

BenWilson2:

Force circleci sphinx build to be in nitpicky mode

## How is this patch tested?

forced failure and CI validation

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Enforce strict formatting of docs build (warnings will convert to errors)

Components 

Interface 

Language 

Integrations

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,13 +62,8 @@ jobs:
           environment:
             JAVA_HOME: /usr/lib/jvm/default-java
           command: |
-            make rsthtml SPHINXOPTS="-b html -W --keep-going -n -T"
-            pydocs=$?
+            make rsthtml
             make javadocs
-            javadocs=$?
-            err=$((pydocs + javadocs))
-            echo "Build test result: $err"
-            test $err = 0
       - store_artifacts:
           path: docs/build/html
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,7 @@ jobs:
             make javadocs
             javadocs=$?
             err=$((pydocs + javadocs))
+            echo 'Build test result: $err'
             test $err = 0
       - store_artifacts:
           path: docs/build/html
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ jobs:
             make javadocs
             javadocs=$?
             err=$((pydocs + javadocs))
-            echo 'Build test result: $err'
+            echo "Build test result: $err"
             test $err = 0
       - store_artifacts:
           path: docs/build/html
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -62,8 +62,13 @@ jobs:
           environment:
             JAVA_HOME: /usr/lib/jvm/default-java
           command: |
-            make rsthtml
+            make rsthtml SPHINXOPTS="-b html -W --keep-going -n -T"
```
BenWilson2, line 65: notes:
rsthtml default is `-b html` but it doesn't hurt to be explicit with the full opts passing
`-W` turns warnings into errors for 'most' serious warnings (not all)
`--keep-going` is default for rsthtml, but might as well be explicit in case sphinx changes defaults (which it has)
`-n` turns nitpicky mode on to catch all types of warnings (strict mode)
`-T` is full stack trace to aid in debugging any warning or errors (helpful on the circleci build page and doesn't hurt the build time)

harupy, line 65: Can we include these options in `Makefile`?

https://github.com/mlflow/mlflow/blob/d6d4d63fbaf1a98b70c450c49658a56fcb3f55b8/docs/Makefile#L5

.circleci/config.yml
```diff
@@ -62,8 +62,13 @@ jobs:
...
+            pydocs=$?
             make javadocs
+            javadocs=$?
+            err=$((pydocs + javadocs))
+            echo "Build test result: $err"
+            test $err = 0
```
harupy, line 71: Why do we need these lines?

BenWilson2: great point. Setting +e would be required, but messing around with inducing failures in local I don't see any benefit. Reverting these changes and I'll push the modifications to the Makefile :)

.circleci/config.yml
BenWilson2: definitely. I completely forgot about that config setup.

harupy (approved):
LGTM!

docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -W --keep-going
+SPHINXOPTS    = -b html -W --keep-going -n -T
```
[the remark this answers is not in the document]
BenWilson2: It does exactly what it should do... force generation of .html files for each of the source directories (ignores the settings of the subclass of dirhtml and generates the same files as a 'build html' (which is a wrapper around the entry point but the -b arg overrides the syntax-sugar exec command). 
Great catch, @harupy ! 
Removed that override and just sticking with the -n -T values:

Local: 

``` sh
➜ make dirhtml
sphinx-build -b dirhtml -d build/doctrees  -W --keep-going -n -T source build/dirhtml
```

harupy (approved):
Approved again :)

PR #5585, opened 2022-04-01
Introduce an mlflow.artifacts.download_artifacts() API, equivalent to the preexisting CLI

dbczumar:

Introduce an mlflow.artifacts.download API, which is an equivalent to the preexisting CLI. Also introduce a `dst_path` argument to this API to control the artifact output destination.

## How is this patch tested?

Unit tests

- [X] Yes. Make sure the changed pages / sections render correctly by following the steps below.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

## Release Notes

Introduce an `mlflow.artifacts.download_artifacts()` API for downloading MLflow artifacts for parity with the existing `mlflow artifacts download` CLI

Components 
- [X] `area/artifacts`: Artifact stores and artifact logging

Interface 

Language 

Integrations


mlflow/store/artifact/cli.py
```diff
@@ -107,27 +107,36 @@ def _file_infos_to_json(file_infos):
...
+    "--dst-path",
+    "-d",
+    help=(
+        "Path of the local filesystem destination directory to which to download the"
+        " specified artifacts. If the directory does not exist, it is created. If unspecified"
+        " the artifacts are downloaded to a new uniquely-named directory on the local filesystem,"
+        " unless the artifacts already exist on the local filesystem, in which case their local"
+        " path is returned directly"
+    ),
+)
```
dbczumar, line 120: Useful addition for CLI users

mlflow/artifacts/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+from mlflow.exceptions import MlflowException
+from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
+from mlflow.tracking import _get_store
+from mlflow.tracking.artifact_utils import _download_artifact_from_uri, get_artifact_repository
+
+
+def download_artifacts(
+    artifact_uri: Optional[str] = None,
+    run_id: Optional[str] = None,
+    artifact_path: Optional[str] = None,
```
dbczumar, line 16: While `run_id` and `artifact_path` are not strictly necessary, I've included them for parity with the existing `mlflow artifacts download` CLI.

docs/source/python_api/mlflow.artifacts.rst
```diff
@@ -0,0 +1,7 @@
+mlflow.artifacts
```
dbczumar, line 1: Document the new module. We opt to introduce `mlflow.artifacts` for parity / consistency with the existing `mlflow artifacts` CLI

mlflow/store/artifact/cli.py
```diff
@@ -107,27 +107,36 @@ def _file_infos_to_json(file_infos):
...
-        return
+        run_id = None
+        artifact_path = None
 
-    artifact_path = artifact_path if artifact_path is not None else ""
-    store = _get_store()
-    artifact_uri = store.get_run(run_id).info.artifact_uri
-    artifact_repo = get_artifact_repository(artifact_uri)
-    artifact_location = artifact_repo.download_artifacts(artifact_path)
-    click.echo(artifact_location)
```
dbczumar, line 130: Artifact download logic moved into `mlflow.artifacts.download_artifacts()`, which this API now depends on.

tests/store/artifact/test_cli.py
```diff
@@ -71,21 +82,61 @@ def test_get_artifact_repository(artifact_uri):
...
+def test_download_artifacts_with_dst_path(run_with_artifact, tmp_path, dst_subdir_path):
+    run, artifact_path, _ = run_with_artifact
+    artifact_uri = f"runs:/{run.info.run_id}/{artifact_path}"
+    dst_path = tmp_path / dst_subdir_path if dst_subdir_path else tmp_path
+
+    command = ["mlflow", "artifacts", "download", "-u", artifact_uri, "-d", dst_path]
+    p = Popen(command, stdout=PIPE, stderr=STDOUT)
+    output = p.stdout.readlines()
+    downloaded_file_path = output[-1].strip().decode("utf-8")
+    assert downloaded_file_path.startswith(str(dst_path))
```
dbczumar, line 142: New test cases for improved coverage

tests/store/artifact/test_cli.py
```diff
@@ -1,16 +1,27 @@
 import json
-import os
```
dbczumar, line 2: Swap os for pathlib

tests/store/artifact/test_cli.py
```diff
@@ -1,16 +1,27 @@
 import json
```
dbczumar, line 1: Added some additional test coverage for the existing CLI. Ideally, it would be sufficient to have a single test asserting that this CLI relies on `mlflow.artifacts.download_artifacts()`, but there are some small behavior differences (e.g. for backcompat, the CLI shouldn't throw if run id *and* artifact URI are specified, the CLI also prints the output path). Because these tests are cheap and it's somewhat difficult to test that the CLI calls this method due to the fact that the CLI runs in a subprocess, I've opted to retain existing test coverage and supplement it slightly.

mlflow/__init__.py
```diff
@@ -43,6 +43,7 @@
 import mlflow.projects as projects
 import mlflow.tracking as tracking
 import mlflow.models
+import mlflow.artifacts
```
dbczumar, line 46: Include `mlflow.artifacts` in the init file so that users can call its functions simply by importing `mlflow` first (no need to import `mlflow.artifacts` explicitly)

tests/store/artifact/test_cli.py
```diff
@@ -71,21 +82,61 @@ def test_get_artifact_repository(artifact_uri):
...
-    run_uri = "runs:/{run_id}/test".format(run_id=run.info.run_id)
-    actual_uri = posixpath.join(run.info.artifact_uri, "test")
+def _run_download_artifact_command(command):
+    """
+    :param command: An `mlflow artifacts` command list.
+    :return: The downloaded artifact content.
+    """
+    p = Popen(command, stdout=PIPE, stderr=STDOUT)
+    output = p.stdout.readlines()
+    download_output_path = pathlib.Path(output[-1].strip().decode("utf-8"))
```
harupy, line 92:
```suggestion
    p = Popen(command, stdout=PIPE, stderr=STDOUT, text=True)
    output = p.stdout.readlines()
    download_output_path = pathlib.Path(output[-1].strip())
```

Nit: can we use `text=True`?

tests/store/artifact/test_cli.py
```diff
@@ -71,21 +82,61 @@ def test_get_artifact_repository(artifact_uri):
...
+    downloaded_file = list(download_output_path.iterdir())[0]
```
harupy, line 93:
```suggestion
    downloaded_file = next(download_output_path.iterdir())
```

nit: Is `next` simpler than `ilst(...)[0]`?

mlflow/artifacts/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+            error_code=INVALID_PARAMETER_VALUE,
+        )
+    elif artifact_uri is not None and artifact_path is not None:
+        raise MlflowException(
+            message="`artifact_path` cannot be specified if `artifact_uri` is specified",
+            error_code=INVALID_PARAMETER_VALUE,
+        )
+
+    if dst_path is not None:
+        pathlib.Path(dst_path).mkdir(exist_ok=True)
```
harupy, line 51:
```suggestion
        pathlib.Path(dst_path).mkdir(exist_ok=True, parenets=True)
```

Do we need `parents=here` (https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir)?

tests/artifacts/test_artifacts.py
```diff
@@ -0,0 +1,71 @@
...
+    download_output_path = mlflow.artifacts.download_artifacts(
+        run_id=run.info.run_id, artifact_path=artifact_path
+    )
+    downloaded_artifact_path = list(pathlib.Path(download_output_path).iterdir())[0]
+    assert downloaded_artifact_path.read_text() == artifact_content
+
+
+def test_download_artifacts_with_run_id_no_path(run_with_artifact):
+    run, artifact_path, _ = run_with_artifact
+    artifact_relative_path_top_level_dir = pathlib.Path(artifact_path).parts[0]
```
harupy, line 42:
```suggestion
    artifact_relative_path_top_level_dir = pathlib.PurePosixPath(artifact_path).parts[0]
```

Probably not a big issue, but should this be `PurePosixPath` because `artifact_path` is a posix path?

tests/artifacts/test_artifacts.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.mark.parametrize("dst_subdir_path", [None, "doesnt_exist_yet"])
+def test_download_artifacts_with_dst_path(run_with_artifact, tmp_path, dst_subdir_path):
+    run, artifact_path, _ = run_with_artifact
+    dst_path = tmp_path / dst_subdir_path if dst_subdir_path else tmp_path
+
+    download_output_path = mlflow.artifacts.download_artifacts(
+        run_id=run.info.run_id, artifact_path=artifact_path, dst_path=dst_path
+    )
+    assert download_output_path.startswith(str(download_output_path))
```
harupy, line 58:
```suggestion
    assert download_output_path.startswith(str(dst_path))
```

Can we fix this line?

harupy:
```suggestion
    assert dst_path.samefile(download_output_path)
```

Can we use `samefile` ([doc](https://docs.python.org/3/library/pathlib.html#pathlib.Path.samefile))?

mlflow/artifacts/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    :param dst_path: Path of the local filesystem destination directory to which to download the
+                     specified artifacts. If the directory does not exist, it is created. If
+                     unspecified, the artifacts are downloaded to a new uniquely-named directory on
+                     the local filesystem, unless the artifacts already exist on the local
+                     filesystem, in which case their local path is returned directly.
+    :return: The location of the artifact file or directory on the local filesystem.
+    """
+    if (run_id is None and artifact_uri is None) or (
+        run_id is not None and artifact_uri is not None
+    ):
```
harupy, line 39:
```suggestion
    if (run_id, artifact_uri).count(None) != 1:
```

nit: can we simplify this line like this?

harupy: Hmm, this approach doesn't seem to align with other if branches.

tests/store/artifact/test_cli.py
```diff
@@ -71,21 +82,61 @@ def test_get_artifact_repository(artifact_uri):
...
-    command = ["mlflow", "artifacts", "download", "-u"]
-    # Test with run uri
-    run_uri = "runs:/{run_id}/test".format(run_id=run.info.run_id)
-    actual_uri = posixpath.join(run.info.artifact_uri, "test")
+def _run_download_artifact_command(command):
+    """
+    :param command: An `mlflow artifacts` command list.
+    :return: The downloaded artifact content.
+    """
+    p = Popen(command, stdout=PIPE, stderr=STDOUT)
```
harupy, line 90: Can we check if this process successfully finished?

harupy:
@dbczumar Left some comments (most of them are minor)!

tests/store/artifact/test_cli.py
```diff
@@ -71,21 +82,61 @@ def test_get_artifact_repository(artifact_uri):
...
+    assert downloaded_content == artifact_content
+
+
+@pytest.mark.parametrize("dst_subdir_path", [None, "doesnt_exist_yet"])
+def test_download_artifacts_with_dst_path(run_with_artifact, tmp_path, dst_subdir_path):
+    run, artifact_path, _ = run_with_artifact
+    artifact_uri = f"runs:/{run.info.run_id}/{artifact_path}"
+    dst_path = tmp_path / dst_subdir_path if dst_subdir_path else tmp_path
+
+    command = ["mlflow", "artifacts", "download", "-u", artifact_uri, "-d", dst_path]
```
harupy, line 138: It appears Popen doesn't accept a command args containing a path-like object in Python 3.7:

https://github.com/mlflow/mlflow/runs/5796393850?check_suite_focus=true#step:6:2081

```
>           needquote = (" " in arg) or ("\t" in arg) or not arg
E           TypeError: argument of type 'WindowsPath' is not iterable
```

https://docs.python.org/3/library/subprocess.html#subprocess.Popen says:

> Changed in version 3.8: args parameter accepts a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object) if shell is False and a sequence containing bytes and path-like objects on Windows.

tests/artifacts/test_artifacts.py
BenWilson2: From what I understand, the "Pure" variant is to limit the access of code to not permit write capabilities to the underlying file system. For reference access to paths, Path and the Pure variants behave identically. I suppose this is a design consideration (do we want to be explicit in path manipulations that don't create directories, write files, remove files, etc) to enforce the 'Pure' variants when reading or constructing representations of paths and only use the abstract top-level Path class when interfacing with the OS? (Interested to hear everyone's thoughts on this one)

tests/store/artifact/test_cli.py
BenWilson2: will it work with a `str(dst_path)` ?

harupy: Yes, that should work

mlflow/artifacts/__init__.py
dbczumar: Great catch! Adjusted test case to capture this.

tests/artifacts/test_artifacts.py
dbczumar: Good call, @harupy ! This should be PurePosixPath because MLflow artifact paths always use the POSIX-style construction (`/` separator). Updated!

tests/artifacts/test_artifacts.py
dbczumar: Awesome! Done!

tests/store/artifact/test_cli.py
dbczumar: Done!

tests/store/artifact/test_cli.py
dbczumar: Done!

tests/store/artifact/test_cli.py
dbczumar: Heck yeah! Done!

tests/store/artifact/test_cli.py
dbczumar: Thanks folks! Fixed!

dbczumar (commented):
@harupy @BenWilson2 Thanks for the reviews! I've addressed your comments! Can you take another look?

tests/store/artifact/test_cli.py
BenWilson2: Marked for unified logic with 2.0 release? :)

dbczumar: Yep! :D

BenWilson2 (approved):
LGTM! Elegant tests too

harupy (approved):
LGTM!

hagerf:
Could this PR have broken the CLI somehow? I'm running some scripts for downloading files like so:
```
downloaded_path=$(mlflow artifacts download --run-id "$MLFLOW_MODEL_RUN_ID" --artifact-path model/"$model_file")
```
and `downloaded_path` is suddenly empty and my scripts are failing :/

hagerf:
It seems it failed because I was running the downloads as shell background tasks. Running them one after the other solves the issue. Sorry for alarming

PR #5593, opened 2022-04-04
Remove client name validation from metrics params and tags

BenWilson2:

Remove the client validation checks for params, metrics, and tags key names. 

## How is this patch tested?

Unit test with previously illegal characters and utf extended set names.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

The Mlflow client will no longer perform name validation checking on parameter names, metric names, or tag names. The underlying tracking backend's restrictions will instead be used to validate names. This does not change pathing validations, however, since the existence of pathing collision detection is distinct from prior valid name checking.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/store/tracking/file_store.py
+++ b/mlflow/store/tracking/file_store.py
@@ -39,6 +39,7 @@ from mlflow.utils.validation import (
     _validate_batch_log_data,
     _validate_list_experiments_max_results,
     _validate_param_keys_unique,
+    _validate_experiment_name as validate_experiment_name,
 )
 from mlflow.utils.env import get_env
 from mlflow.utils.file_utils import (
@@ -295,11 +296,8 @@ class FileStore(AbstractStore):
         return experiment_id
 
     def _validate_experiment_name(self, name):
-        """Check the validity of an experiment name."""
-        if name is None or name == "":
-            raise MlflowException(
-                "Invalid experiment name '%s'" % name, databricks_pb2.INVALID_PARAMETER_VALUE
-            )
+
+        validate_experiment_name(name)
         experiment = self.get_experiment_by_name(name)
         if experiment is not None:
             if experiment.lifecycle_stage == LifecycleStage.DELETED:
@@ -395,7 +393,7 @@ class FileStore(AbstractStore):
         conflict_experiment = self._get_experiment_path(experiment_id, ViewType.ACTIVE_ONLY)
         if conflict_experiment is not None:
             raise MlflowException(
-                "Cannot restore eperiment with ID %d. "
+                "Cannot restore experiment with ID %d. "
                 "An experiment with same ID already exists." % experiment_id,
                 databricks_pb2.RESOURCE_ALREADY_EXISTS,
             )
--- a/mlflow/store/tracking/file_store.py
+++ b/mlflow/store/tracking/file_store.py
@@ -39,7 +39,7 @@ from mlflow.utils.validation import (
     _validate_batch_log_data,
     _validate_list_experiments_max_results,
     _validate_param_keys_unique,
-    _validate_experiment_name as validate_experiment_name,
+    _validate_experiment_name,
 )
 from mlflow.utils.env import get_env
 from mlflow.utils.file_utils import (
@@ -295,9 +295,7 @@ class FileStore(AbstractStore):
                 self.set_experiment_tag(experiment_id, tag)
         return experiment_id
 
-    def _validate_experiment_name(self, name):
-
-        validate_experiment_name(name)
+    def _validate_experiment_does_not_exist(self, name):
         experiment = self.get_experiment_by_name(name)
         if experiment is not None:
             if experiment.lifecycle_stage == LifecycleStage.DELETED:
@@ -316,7 +314,8 @@ class FileStore(AbstractStore):
 
     def create_experiment(self, name, artifact_location=None, tags=None):
         self._check_root_dir()
-        self._validate_experiment_name(name)
+        _validate_experiment_name(name)
+        self._validate_experiment_does_not_exist(name)
         # Get all existing experiments and find the one with largest numerical ID.
         # len(list_all(..)) would not work when experiments are deleted.
         experiments_ids = [
@@ -400,6 +399,7 @@ class FileStore(AbstractStore):
         mv(experiment_dir, self.root_directory)
 
     def rename_experiment(self, experiment_id, new_name):
+        _validate_experiment_name(new_name)
         meta_dir = os.path.join(self.root_directory, experiment_id)
         # if experiment is malformed, will raise error
         experiment = self._get_experiment(experiment_id)
@@ -408,7 +408,7 @@ class FileStore(AbstractStore):
                 "Experiment '%s' does not exist." % experiment_id,
                 databricks_pb2.RESOURCE_DOES_NOT_EXIST,
             )
-        self._validate_experiment_name(new_name)
+        self._validate_experiment_does_not_exist(new_name)
         experiment._set_name(new_name)
         if experiment.lifecycle_stage != LifecycleStage.ACTIVE:
             raise Exception(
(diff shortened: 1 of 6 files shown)


mlflow/utils/validation.py
```diff
@@ -234,12 +232,8 @@ def _validate_param_name(name):
 
 
 def _validate_tag_name(name):
```
dbczumar, line 234: These validation APIs are used by storage backends (e.g. `validate_param_name` is used by FileStore, `validate_tag_name` is used by FileStore & SqlAlchemyStore via `validate_tag`). The backends should still be validating the names, but the clients should not be. Rather than changing the validation util function implementations, we should be changing where these functions are called (i.e. remove the function calls from the MlflowClient layer).

dbczumar: Specifically, we should update https://github.com/mlflow/mlflow/blob/HEAD/mlflow/tracking/_tracking_service/client.py

mlflow/tracking/_tracking_service/client.py
```diff
@@ -301,12 +294,6 @@ def log_batch(self, run_id, metrics=(), params=(), tags=()):
             return
         if len(params) > 1:
```
BenWilson2, line 295: I left this validation in here so that we don't have a partial commit. Is this ok?

mlflow/utils/validation.py
```diff
@@ -80,7 +82,12 @@ def path_not_unique(name):
 
 def _validate_metric_name(name):
     """Check that `name` is a valid metric name and raise an exception if it isn't."""
-    if name is None or not _VALID_PARAM_AND_METRIC_NAMES.match(name):
+    if name is None:
```
BenWilson2, line 85: I split this out since the Exception thrown for a `None` key name threw the same for an invalid name string and I could see how a user might find this confusing. Is this ok or should I revert this?

mlflow/utils/validation.py
BenWilson2: totally read that ticket wrong. Moved the logic to correct location.

tests/tracking/test_tracking.py
```diff
@@ -534,6 +541,13 @@ def test_log_batch_validates_entity_names_and_values():
             tracking.MlflowClient().log_batch(run_id, tags=tags)
         assert e.value.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
 
+        metrics = [Metric(key=None, value=42.0, timestamp=4, step=1)]
```
BenWilson2, line 544: Added this after splitting out the `None` check on the key name. The original exception message was a little confusing for this case.

tests/tracking/test_tracking.py
```diff
@@ -319,9 +319,9 @@ def test_log_batch():
     for key, value in finished_run.data.metrics.items():
         assert expected_metrics[key] == value
     metric_history0 = client.get_metric_history(run_id, "metric-key0")
-    assert set([(m.value, m.timestamp, m.step) for m in metric_history0]) == set([(1.0, t, 0)])
+    assert set([(m.value, m.timestamp, m.step) for m in metric_history0]) == {(1.0, t, 0)}
```
harupy, line 322: Do we need `set` here?

```suggestion
    assert [(m.value, m.timestamp, m.step) for m in metric_history0] == [(1.0, t, 0)]
```

harupy: If `metric_history0` contains multiple elements, `set` would make sense.

harupy: I guess this is for parity with other assertions.

mlflow/utils/validation.py
```diff
@@ -80,7 +82,12 @@ def path_not_unique(name):
...
+        raise MlflowException(
+            f"Invalid metric name: '{name}'. {_MISSING_KEY_NAME_MESSAGE}",
```
harupy, line 87: This might look like `mlflow.log_metrics({"None": 0})` is disallowed.

tests/tracking/test_tracking.py
BenWilson2: an excellent catch. Changed all but the locations where dealing with a `set` makes sense.

mlflow/utils/validation.py
BenWilson2: great observation. changed!

tests/tracking/test_tracking.py
BenWilson2: Had to revert a handful of them due to no ordering guarantees and the simplified runtime of set equivalency compared to recursive list comparisons.

tests/tracking/test_tracking.py
```diff
@@ -142,13 +142,12 @@ def test_set_experiment_with_deleted_experiment():
...
-                for exp in client.list_experiments(view_type=view_type_arg)
-            ]
-        )
-        assert result == set([(exp_id, stage) for exp_id, stage in ids_to_lifecycle_stage.items()])
+        result = [
+            (exp.experiment_id, exp.lifecycle_stage)
+            for exp in client.list_experiments(view_type=view_type_arg)
+        ]
+
+        assert result == [(exp_id, stage) for exp_id, stage in ids_to_lifecycle_stage.items()]
```
harupy, line 150:
```suggestion
        assert result == list(ids_to_lifecycle_stage.items())
```

nit: can we simplify this line?

mlflow/tracking/_tracking_service/client.py
harupy: `FileStore` performs the same validation.

https://github.com/mlflow/mlflow/blob/0fa849ad75e5733bf76cc14a4455657c5c32f107/mlflow/store/tracking/file_store.py#L885-L889

harupy: `SqlAlchemyStore` doesn't:

https://github.com/mlflow/mlflow/blob/d748f559b4122bdfba1283ccf3b4847e73f0d602/mlflow/store/tracking/sqlalchemy_store.py#L811-L815

harupy: `RestStore` performs no validation (is this intentional?):

https://github.com/mlflow/mlflow/blob/e7caa0a1ffd47255f52b2818fad49d34ed332d99/mlflow/store/tracking/rest_store.py#L302-L307

harupy: Are there any downsides to performing this validation in a backend store?

BenWilson2: With the SqlAlchemyStore validation not being done, the first of a duplicate key will be written, raising an underlying SQLAlchemy primary key index collision exception. I think the proper thing to do here is to enforce the unique check on SqlAlchemy and remove from the client. I'll push that change soon.

BenWilson2: (specifically: `sqlite3.IntegrityError: UNIQUE constraint failed: params.key, params.run_uuid`)

mlflow/store/tracking/file_store.py
```diff
@@ -39,6 +39,7 @@
     _validate_batch_log_data,
     _validate_list_experiments_max_results,
     _validate_param_keys_unique,
+    _validate_experiment_name as validate_experiment_name,
```
harupy, line 42: Do we need the as part? I think `_validate_experiment_name` and `FileStore._validate_experiment_name` don't conflict.

harupy (approved):
Left one more comment, otherwise LGTM!

dbczumar (approved):
LGTM! Thanks @BenWilson2 !

mlflow/store/tracking/file_store.py
BenWilson2: that was my intention. Do you think I should refactor the internal module name instead?

harupy (approved):

mlflow/store/tracking/file_store.py
harupy: Offline discussion:
- remove `as validate_experiment_name `
- rename `FileStore._validate_experiment_name` to `FileStore._validate_experiment_does_not_exist` and apply it and `_validate_experiment_name` separately.

PR #5571, opened 2022-03-31
Add cli validation checks for backend-store-uri in artifacts-only mode

BenWilson2:
Signed-off-by: Ben Wilson <benjamin.wilson@databricks.com>

## What changes are proposed in this pull request?

Introduces a validation step when using a proxy artifact server to ensure that a configuration issue doesn't cause the inability to list artifacts if the backend store isn't locally available on the proxy artifact server.

## How is this patch tested?

Unit test

## Does this PR change the documentation?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Make sure the changed pages / sections render correctly by following the steps below.

1. Check the status of the `ci/circleci: build_doc` check. If it's successful, proceed to the
   next step, otherwise fix it.
2. Click `Details` on the right to open the job page of CircleCI.
3. Click the `Artifacts` tab.
4. Click `docs/build/html/index.html`.
5. Find the changed pages / sections and make sure they render correctly.

### Is this a user-facing change?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [x] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [x] `area/server-infra`: MLflow Tracking server backend
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface 
- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

[...]

PR #5650, opened 2022-04-09
Add configuration warning for --artifacts-only server initialization

BenWilson2:

Adds a cli warning if an mlflow server is started in --artifacts-only mode and the default value of --backend-store-uri is not the default value. The warning will print to stdout that any other servers that are started should not define the --backend-store-uri to this external location to prevent issues with artifact listing (namely in the UI). 
This is a corrected change to the reverted PR #5571 which incorrectly raised an exception.
This PR also moves server cli utility functions into its own module since the function `resolve_default_artifact_root()` was the sole importer of several modules and didn't really belong in `mlflow.utils.uri`.  

## How is this patch tested?

Unit test added to validate stdout printing of the proper warning raised and to validate that the server is still capable of executing a start command.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/server_cli_utils.py
+++ b/mlflow/utils/server_cli_utils.py
@@ -1,0 +1,47 @@
+"""
+Utilities for MLflow cli server config validation and resolving.
+NOTE: these functions are intended to be used as utilities for the cli click-based interface.
+Do not use for any other purpose as the potential Exceptions being raised will be misleading
+for users.
+"""
+
+import click
+
+from mlflow.store.tracking import DEFAULT_ARTIFACTS_URI, DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
+from mlflow.utils.logging_utils import eprint
+from mlflow.utils.uri import is_local_uri
+
+
+def resolve_default_artifact_root(
+    serve_artifacts: bool,
+    default_artifact_root: str,
+    backend_store_uri: str,
+    resolve_to_local: bool = False,
+) -> str:
+    if serve_artifacts and not default_artifact_root:
+        default_artifact_root = DEFAULT_ARTIFACTS_URI
+    elif not serve_artifacts and not default_artifact_root:
+        if is_local_uri(backend_store_uri):
+            default_artifact_root = backend_store_uri
+        elif resolve_to_local:
+            default_artifact_root = DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
+        else:
+            msg = (
+                "Option 'default-artifact-root' is required when backend store is not "
+                "local file based."
+            )
+            eprint(msg)
+            raise click.UsageError(message=msg)
+    return default_artifact_root
+
+
+def artifacts_only_config_warning(artifacts_only: bool, backend_store_uri: str) -> None:
+
+    if artifacts_only and not is_local_uri(backend_store_uri):
+        msg = (
+            "You are starting a tracking server in `--artifacts-only` mode with a non-default "
+            f"`--backend_store_uri`: '{backend_store_uri}'. To prevent errors in listing "
+            "artifacts, please ensure that any other tracking servers that are started do not "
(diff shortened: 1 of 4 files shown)


mlflow/utils/server_cli_utils.py
```diff
@@ -0,0 +1,47 @@
...
+                "local file based."
+            )
+            eprint(msg)
+            raise click.UsageError(message=msg)
+    return default_artifact_root
+
+
+def artifacts_only_config_warning(artifacts_only: bool, backend_store_uri: str) -> None:
+
+    if artifacts_only and not is_local_uri(backend_store_uri):
```
dbczumar, line 40: Can we specifically check if it's the default `./mlruns` URI?

mlflow/utils/server_cli_utils.py
```diff
@@ -0,0 +1,47 @@
...
+        msg = (
+            "You are starting a tracking server in `--artifacts-only` mode with a non-default "
+            f"`--backend_store_uri`: '{backend_store_uri}'. To prevent errors in listing "
```
dbczumar, line 43: Rather than saying "non-default --backend-store-uri", I think it's clearer to users if we say "you've specified --backend-store-uri and --artifacts-only together" and then explain why that doesn't make much sense.

mlflow/utils/server_cli_utils.py
BenWilson2: changed to explicitly check if the non-default value is specified.

mlflow/utils/server_cli_utils.py
BenWilson2: Agreed. They really shouldn't be specified together. Changed to a Click.UsageError exception and adjusted the test to pattern match check on the exception message to stderr

mlflow/utils/server_cli_utils.py
```diff
@@ -0,0 +1,61 @@
...
+def artifacts_only_config_validation(artifacts_only: bool, backend_store_uri: str) -> None:
+
+    if artifacts_only and not _is_default_backend_store_uri(backend_store_uri):
+        msg = (
+            "You are starting a tracking server in `--artifacts-only` mode and have provided a "
+            f"value for `--backend_store_uri`: '{backend_store_uri}'. A tracking server in "
+            "`--artifacts-only` mode cannot have a custom value set for `--backend_store_uri` to "
+            "properly proxy access to the artifact storage location. Remove the "
+            "`--backend_store_uri` argument in your configuration arguments for this server and "
+            "any other tracking servers when operating in proxy artifact access mode."
```
dbczumar, line 59: `--backend-store-uri` should still be specified for the `mlflow server` instance responsible for metadata operations (the one not running in `--artifacts-only` mode). Can we remove this last sentence?

mlflow/utils/server_cli_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    return backend_store_uri == DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH
+
+
+def artifacts_only_config_validation(artifacts_only: bool, backend_store_uri: str) -> None:
+
+    if artifacts_only and not _is_default_backend_store_uri(backend_store_uri):
+        msg = (
+            "You are starting a tracking server in `--artifacts-only` mode and have provided a "
+            f"value for `--backend_store_uri`: '{backend_store_uri}'. A tracking server in "
+            "`--artifacts-only` mode cannot have a custom value set for `--backend_store_uri` to "
```
dbczumar, line 56: Can we avoid referring to `custom value`? The point is that `--backend-store-uri` shouldn't be specified at all because it's meaningless for a server running in `--artifacts-only` mode.

dbczumar (approved):
Awesome stuff, @BenWilson2 ! Nailed it!

PR #5657, opened 2022-04-11
Fixes #5582

### LINKED ISSUE #5582: `mlflow.pyfunc.PythonModel`s referencing multiple artifacts with distinct uris but common sub-paths do not work as expected

### Willingness to contribute

### Describe the problem
Actual behaviour: 
A common pattern I have been working with is to use a pyfunc.PythonModel to wrap up a sequence of inference steps from separate models. The idea is illustrated below. The PythonModel references multiple artifacts that are distinguishable through their uris, but share a common artifact sub-path. When these are downloaded as part of a call to `mlflow.pyfunc.save_model` the artifacts are sequentially overwritten, and the result is that multiple distinct references are resolved to the same underlying artifacts. This all happens quietly, and could cause a real headache that might be hard to catch for practitioners.

```python
import mlflow

class TwoStepPyfuncModel(mlflow.pyfunc.PythonModel):
  def load_context(self, context):
    self._first_step_model = mlflow.keras.load_model(context.artifacts["first_step_model"])
    self._second_step_model = mlflow.keras.load_model(context.artifacts["second_step_model"])
    # Issue: context.artifacts["first_step_model"] = context.artifacts["second_step_model"]
    # Naive way to catch:
    if context.artifacts["first_step_model"] == context.artifacts["second_step_model"]:
      raise Exception("oops")

  def predict(self, context, model_input):    
    first_step_result = self._first_step_model(model_input, training=False)
    if condition(first_step_result):
      second_step_result = self.second_step_model(model_input, training=False)
      return {"first_step_result": first_step_result, "second_step_result": second_step_result}
    else:
      return {"first_step_result": first_step_result, "second_step_result": None}


two_step_model = TwoStepPyfuncModel()

mlflow.pyfunc.save_model(
  "./pyfunc_model", 
  python_model = two_step_model, 
  artifacts = {
    "first_step_model": 'runs:/abc/model', # Default artifact path for saving a model.
    "second_step_model": 'runs:/cde/model' # Default artifact path for saving a model.
  }
)

two_step_model_reloaded = mlflow.pyfunc.load_model("./pyfunc_model")

two_step_model_reloaded.predict(some_input)
# Exception: Oops
```
Desired behaviour:
It should not be possible for artifacts specified in `mlflow.pyfunc.save_model` to be quietly overwritten. I imagine there are reasons why it may be desirable for these artifacts to retain the same relative path as their originating sources, but at the very least we should tell the user that this has happened and raise an exception. An alternative would be to handle this situation by injecting a differentiator in the sub-paths of the artifacts downloaded for the new PythonModel, and to alert the user to this, though I don't have much visibility on whether this would cause side-effects elsewhere. I note that this is only an issue if the source artifacts in their separate runs happen to have the same relative path - but this is often the case, for the first model logged to an mlflow run, in my experience. 

Components 

Interface 

Language 

Integrations

kyle-jarvis:
Signed-off-by: Kyle Jarvis <kyle.jarvis1905@gmail.com>

The PR pertains to issue #5582 . It proposes to include a sub-directory to index artifacts as they are downloaded when saving an `mlflow.pyfunc.PythonModel`. This helps to avoid cases where artifacts may overwrite one-another due to common relative paths from distinct base uris, resulting in possibly unexpected behaviour.

## How is this patch tested?

A new test has been introduced to illustrate the problem that this patch addresses:

test_model_export_with_class_and_artifacts.py::test_multiple_artifacts_with_common_sub_paths_resolve_correctly_when_saving_model

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyfunc/model.py
+++ b/mlflow/pyfunc/model.py
@@ -175,9 +175,14 @@ def _save_model_with_class_artifacts_params(
         with TempDir() as tmp_artifacts_dir:
             tmp_artifacts_config = {}
             saved_artifacts_dir_subpath = "artifacts"
-            for artifact_name, artifact_uri in artifacts.items():
+            for artifact_index, (artifact_name, artifact_uri) in enumerate(artifacts.items()):
+                artifact_download_output_path = (
+                    os.path.join(tmp_artifacts_dir.path(), str(artifact_index)) if len(artifacts) > 1
+                    else tmp_artifacts_dir.path()
+                )
+                os.makedirs(artifact_download_output_path, exist_ok=True)
                 tmp_artifact_path = _download_artifact_from_uri(
-                    artifact_uri=artifact_uri, output_path=tmp_artifacts_dir.path()
+                    artifact_uri=artifact_uri, output_path=artifact_download_output_path
                 )
                 tmp_artifacts_config[artifact_name] = tmp_artifact_path
                 saved_artifact_subpath = posixpath.join(
--- a/mlflow/pyfunc/model.py
+++ b/mlflow/pyfunc/model.py
@@ -177,7 +177,8 @@ def _save_model_with_class_artifacts_params(
             saved_artifacts_dir_subpath = "artifacts"
             for artifact_index, (artifact_name, artifact_uri) in enumerate(artifacts.items()):
                 artifact_download_output_path = (
-                    os.path.join(tmp_artifacts_dir.path(), str(artifact_index)) if len(artifacts) > 1
+                    tmp_artifacts_dir.path(str(artifact_index))
+                    if len(artifacts) > 1
                     else tmp_artifacts_dir.path()
                 )
                 os.makedirs(artifact_download_output_path, exist_ok=True)
--- a/tests/pyfunc/test_model_export_with_class_and_artifacts.py
+++ b/tests/pyfunc/test_model_export_with_class_and_artifacts.py
@@ -847,6 +847,40 @@ def test_save_model_correctly_resolves_directory_artifact_with_nested_contents(
 
 
 @pytest.mark.large
+def test_multiple_artifacts_with_common_sub_paths_resolve_correctly_when_saving_model(
+    sklearn_knn_model, model_path, iris_data
+):
+    default_model_path = "model"
+    with mlflow.start_run():
+        mlflow.sklearn.log_model(sk_model=sklearn_knn_model, artifact_path=default_model_path)
+        first_model_uri = "runs:/{run_id}/{artifact_path}".format(
+            run_id=mlflow.active_run().info.run_id, artifact_path=default_model_path
+        )
+
+    with mlflow.start_run():
+        mlflow.sklearn.log_model(sk_model=sklearn_knn_model, artifact_path=default_model_path)
+        second_model_uri = "runs:/{run_id}/{artifact_path}".format(
+            run_id=mlflow.active_run().info.run_id, artifact_path=default_model_path
+        )
+
+    class ArtifactValidationModel(mlflow.pyfunc.PythonModel):
+        def predict(self, context, model_input):
+            first_model_resolved_path = context.artifacts["first_model"]
+            second_model_resolved_path = context.artifacts["second_model"]
+            return first_model_resolved_path != second_model_resolved_path
+
+    mlflow.pyfunc.save_model(
+        path=model_path,
+        artifacts={"first_model": first_model_uri, "second_model": second_model_uri},
+        python_model=ArtifactValidationModel(),
+        conda_env=_conda_env(),
+    )
+
+    loaded_model = mlflow.pyfunc.load_model(model_uri=model_path)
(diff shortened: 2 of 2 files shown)


tests/pyfunc/test_model_export_with_class_and_artifacts.py
```diff
@@ -846,6 +846,40 @@ def predict(self, context, model_input):
...
+@pytest.mark.large
+def test_multiple_artifacts_with_common_sub_paths_resolve_correctly_when_saving_model(
+    sklearn_knn_model, model_path, iris_data
+):
+    default_model_path = "model"
+    with mlflow.start_run():
+        mlflow.sklearn.log_model(sk_model=sklearn_knn_model, artifact_path=default_model_path)
+        first_model_uri = "runs:/{run_id}/{artifact_path}".format(
+            run_id=mlflow.active_run().info.run_id, artifact_path=default_model_path
+        )
```
harupy, line 858:
```suggestion
        model_info = mlflow.sklearn.log_model(sk_model=sklearn_knn_model, artifact_path=default_model_path)
        first_model_uri = model_info.model_uri
```

This block can be simplified by using the return value of `log_model`.

tests/pyfunc/test_model_export_with_class_and_artifacts.py
```diff
@@ -846,6 +846,40 @@ def predict(self, context, model_input):
...
+
+    with mlflow.start_run():
+        mlflow.sklearn.log_model(sk_model=sklearn_knn_model, artifact_path=default_model_path)
+        second_model_uri = "runs:/{run_id}/{artifact_path}".format(
+            run_id=mlflow.active_run().info.run_id, artifact_path=default_model_path
+        )
```
harupy, line 864: Same here https://github.com/mlflow/mlflow/pull/5657/files#r847259182

mlflow/pyfunc/model.py
```diff
@@ -175,9 +175,14 @@ def _save_model_with_class_artifacts_params(
         with TempDir() as tmp_artifacts_dir:
             tmp_artifacts_config = {}
             saved_artifacts_dir_subpath = "artifacts"
-            for artifact_name, artifact_uri in artifacts.items():
+            for artifact_index, (artifact_name, artifact_uri) in enumerate(artifacts.items()):
+                artifact_download_output_path = (
+                    os.path.join(tmp_artifacts_dir.path(), str(artifact_index)) if len(artifacts) > 1
```
harupy, line 180:
```suggestion
                    tmp_artifacts_dir.path(str(artifact_index)) if len(artifacts) > 1
```

Can we `str(artifact_index)` to `tmp_artifacts_dir.path`?

harupy:
@kyle-jarvis Thanks for the PR, left a few comments!

kyle-jarvis:
@harupy Thanks for tidying that up, incorporated your suggestions!

harupy (approved):
LGTM!

vilmar-hillow:
This seems to be a minor breaking change for us. Is there a more elegant way to handle it?
We are downloading specific artifacts from the run, for example a Pipfile to initialize an environment. Previously we would reference it via something like `mlflow artifacts download --artifact-uri=${MODEL_URI}/artifacts/Pipfile`. Now we have to make sure we are referencing the correct subdirectory, i.e. `artifacts/0/Pipfile`. It seems possible by ensuring the order of artifact mapping passed to `save_model()`, but seems kind of unintuitive and prone to introducing errors in the future.

MajerMartin:
Hi, this is a breaking change for us as well.

We sometimes use “multi-models”, which are basically models that encapsulates different weights for different use-cases but under single common interface. Imagine sentiment analysis model for different languages - all languages share the same code but use different weights.

The weights used to be stored in separate subdirectories as follows:
```
artifact/
  en/
    weights.pt
  cs/
    weights.pt
  …      
```
This allowed us to simply download only the weights we need and run the model for english only. Or for english and czech at the same time. Or for all languages. All of this with just one model in registry and correctly downloaded weights.

Now the subdirectories are prefixed with index:
```
artifact/
  0/
    en/
      weights.pt
  1/
    cs/
      weights.pt
  …    
```
so I have to agree with @vilmar-hillow that this change makes it unintuitive and more difficult to work with. I do not now internal of MLflow that well, but could we consider different way how the handle the original issue?

Furthermore, I do not think that such changes should be under patch release. It is presented as a bug fix, but it only fixes behavior for one particular use-case and as you can see, it actually is a breaking change.

kyle-jarvis:
@vilmar-hillow @MajerMartin Sorry about that! Interesting to see how you folks are working with these kinds of things. I have raised #5891 to revert these changes and will look back at the original issue my use case was experiencing with the points you raised in mind.
@harupy thanks for your original input on this.

MajerMartin:
Hi @kyle-jarvis, thank you for your response and for taking an initiative to fix this. I really appreciate it and hope someone suggests another way to make your use-case work too.

PR #5676, opened 2022-04-12
Make `pyfunc.spark_udf` support virtualenv

WeichenXu123:

Make `pyfunc.spark_udf` support using virtualenv to restore python environment.

## How is this patch tested?

Unit tests.

Manually tests on databricks spark cluster:

(enable nfs)
https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/4182715196395444/command/4182715196395447

(disable nfs)
https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/2161901739220879/command/2161901739220880

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Make `pyfunc.spark_udf` support using virtualenv to restore python environment.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -1178,7 +1178,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
                 local_model_path_on_executor,
                 workers=1,
                 install_mlflow=False,
-                env_manager=_EnvManager.CONDA,
+                env_manager=env_manager,
                 env_root_dir=env_root_dir_on_executor,
             )
 
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -969,15 +969,17 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
 
         - ``ArrayType(StringType)``: All columns converted to ``string``.
 
-    :param env_manager: The environment manager to use in order to create the
-                        software environment for model inference. Default value is ``local``,
-                        The following values are supported:
+    :param env_manager: The environment manager to use in order to create the software environment
+                        for model inference. Note that environment is only restored in the context
+                        of the PySpark UDF; the software environment outside of the UDF is
+                        unaffected. Default value is ``local``, and the following values are
+                        supported:
 
                          - ``conda``: (Recommended) Use Conda to restore the software environment
                            that was used to train the model. Note that environment is only restored
                            in the context of the PySpark UDF; the software environment outside of
                            the UDF is unaffected.
-                        - ``virtualenv``: Use virtualenv to restore the software environment
+                         - ``virtualenv``: Use virtualenv to restore the software environment.
                          - ``local``: Use the current Python environment for model inference, which
                            may differ from the environment used to train the model and may lead to
                            errors or invalid predictions.
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -969,7 +969,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
 
         - ``ArrayType(StringType)``: All columns converted to ``string``.
 
-    :param env_manager: The environment manager to use in order to create the software environment
+    :param env_manager: The environment manager to use in order to create the python environment
                         for model inference. Note that environment is only restored in the context
                         of the PySpark UDF; the software environment outside of the UDF is
                         unaffected. Default value is ``local``, and the following values are
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -977,6 +977,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
                            that was used to train the model. Note that environment is only restored
                            in the context of the PySpark UDF; the software environment outside of
                            the UDF is unaffected.
+                        - ``virtualenv``: Use virtualenv to restore the software environment
                          - ``local``: Use the current Python environment for model inference, which
                            may differ from the environment used to train the model and may lead to
                            errors or invalid predictions.
@@ -1009,7 +1010,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
     nfs_root_dir = get_nfs_cache_root_dir()
     should_use_nfs = nfs_root_dir is not None
     should_use_spark_to_broadcast_file = not (is_spark_in_local_mode or should_use_nfs)
-    conda_env_root_dir = _get_or_create_env_root_dir(should_use_nfs)
+    env_root_dir = _get_or_create_env_root_dir(should_use_nfs)
 
     if not isinstance(result_type, SparkDataType):
         result_type = _parse_datatype_string(result_type)
@@ -1067,12 +1068,12 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
         # "capture_output=False" is the output will be printed immediately, otherwise you have
         # to wait conda command fail and suddenly get all output printed (included in error
         # message).
-        if env_manager is _EnvManager.CONDA:
+        if env_manager is not _EnvManager.LOCAL:
             _get_flavor_backend(
                 local_model_path,
-                env_manager=_EnvManager.CONDA,
+                env_manager=env_manager,
                 install_mlflow=False,
-                env_root_dir=conda_env_root_dir,
+                env_root_dir=env_root_dir,
             ).prepare_env(model_uri=local_model_path, capture_output=is_in_databricks_runtime())
 
     # Broadcast local model directory to remote worker if needed.
@@ -1162,24 +1163,23 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
 
         scoring_server_proc = None
 
-        # TODO: Support virtual env.
-        if env_manager is _EnvManager.CONDA:
+        if env_manager is not _EnvManager.LOCAL:
             if should_use_spark_to_broadcast_file:
                 local_model_path_on_executor = _SparkDirectoryDistributor.get_or_extract(
                     archive_path
                 )
                 # Create individual conda_env_root_dir for each spark UDF task process.
-                conda_env_root_dir_on_executor = _get_or_create_env_root_dir(should_use_nfs)
+                env_root_dir_on_executor = _get_or_create_env_root_dir(should_use_nfs)
             else:
                 local_model_path_on_executor = local_model_path
-                conda_env_root_dir_on_executor = conda_env_root_dir
+                env_root_dir_on_executor = env_root_dir
 
             pyfunc_backend = _get_flavor_backend(
                 local_model_path_on_executor,
                 workers=1,
                 install_mlflow=False,
                 env_manager=_EnvManager.CONDA,
-                env_root_dir=conda_env_root_dir_on_executor,
+                env_root_dir=env_root_dir_on_executor,
             )
 
             if should_use_spark_to_broadcast_file:
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -977,7 +977,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
 
                          - ``conda``: (Recommended) Use Conda to restore the software environment
                            that was used to train the model.
-                         - ``virtualenv``: Use virtualenv to restore the software environment that
+                         - ``virtualenv``: Use virtualenv to restore the python environment that
                            was used to train the model.
                          - ``local``: Use the current Python environment for model inference, which
                            may differ from the environment used to train the model and may lead to
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -976,10 +976,9 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
                         supported:
 
                          - ``conda``: (Recommended) Use Conda to restore the software environment
-                           that was used to train the model. Note that environment is only restored
-                           in the context of the PySpark UDF; the software environment outside of
-                           the UDF is unaffected.
-                         - ``virtualenv``: Use virtualenv to restore the software environment.
+                           that was used to train the model.
+                         - ``virtualenv``: Use virtualenv to restore the software environment that
+                           was used to train the model.
                          - ``local``: Use the current Python environment for model inference, which
                            may differ from the environment used to train the model and may lead to
                            errors or invalid predictions.
--- a/mlflow/utils/file_utils.py
+++ b/mlflow/utils/file_utils.py
@@ -499,7 +499,9 @@ def get_or_create_tmp_dir():
         # Note: For python process attached to databricks notebook, atexit does not work.
         # The /tmp/repl_tmp_data/{repl_id} directory will be removed once databricks notebook
         # detaches.
-        tmp_dir = os.path.join("/tmp", "repl_tmp_data", get_repl_id())
+        # The repl_tmp_data directory is designed to be used by all kinds of applications,
+        # so create a child directory "mlflow" for storing mlflow temp data.
+        tmp_dir = os.path.join("/tmp", "repl_tmp_data", get_repl_id(), "mlflow")
         os.makedirs(tmp_dir, exist_ok=True)
     else:
         tmp_dir = tempfile.mkdtemp()
@@ -525,7 +527,9 @@ def get_or_create_nfs_tmp_dir():
         # Note: In databricks, atexit hook does not work.
         # The {nfs_root_dir}/repl_tmp_data/{repl_id} directory will be removed once databricks
         # notebook detaches.
-        tmp_nfs_dir = os.path.join(nfs_root_dir, "repl_tmp_data", get_repl_id())
+        # The repl_tmp_data directory is designed to be used by all kinds of applications,
+        # so create a child directory "mlflow" for storing mlflow temp data.
+        tmp_nfs_dir = os.path.join(nfs_root_dir, "repl_tmp_data", get_repl_id(), "mlflow")
         os.makedirs(tmp_nfs_dir, exist_ok=True)
     else:
         tmp_nfs_dir = tempfile.mkdtemp(dir=nfs_root_dir)
--- a/mlflow/utils/process.py
+++ b/mlflow/utils/process.py
@@ -47,7 +47,7 @@ def _exec_cmd(
                             message on failure; if False, these streams won't be captured.
     :param: synchronous: If True, wait process complete and return a CompletedProcess instance,
                          If False, does not wait process complete and return a Popen instance,
-                         and ignore the `throw_on_error`, `check`, `capture_output` argument.
+                         and ignore the `throw_on_error`, `check`, argument.
     :param kwargs: Keyword arguments (except `check` and `text`) passed to `subprocess.run` or
                    `subproces.Popen`.
     :return:  If synchronous is True, return a `subprocess.CompletedProcess` instance,
@@ -81,6 +81,10 @@ def _exec_cmd(
             raise ShellCommandException.from_completed_process(prc)
         return prc
     else:
+        if capture_output:
+            kwargs['stdout'] = subprocess.PIPE
+            kwargs['stderr'] = subprocess.PIPE
+
         return subprocess.Popen(
             cmd,
             env=env,
--- a/mlflow/utils/process.py
+++ b/mlflow/utils/process.py
@@ -33,6 +33,7 @@ def _exec_cmd(
     throw_on_error=True,
     extra_env=None,
     capture_output=True,
+    synchronous=True,
     **kwargs,
 ):
     """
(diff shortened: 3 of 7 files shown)


mlflow/utils/virtualenv.py
```diff
@@ -32,11 +32,17 @@ def _get_mlflow_virtualenv_root():
...
+
 def _is_pyenv_available():
     """
     Returns True if pyenv is available, otherwise False.
     """
-    return shutil.which("pyenv") is not None
+    if is_in_databricks_runtime():
+        return os.path.exists(_DATABRICKS_PYENV_BIN_PATH)
+    else:
+        return shutil.which("pyenv") is not None
```
harupy, line 45: Just because `os.path.exists(_DATABRICKS_PYENV_BIN_PATH)` returns True doesn't necessarily mean pyenv is available, right?

harupy: Curious why we need this special logic for Databricks.

mlflow/pyfunc/__init__.py
```diff
@@ -977,6 +977,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
                            that was used to train the model. Note that environment is only restored
                            in the context of the PySpark UDF; the software environment outside of
                            the UDF is unaffected.
+                        - ``virtualenv``: Use virtualenv to restore the software environment
```
dbczumar, line 980:
```suggestion
                           in the context of the PySpark UDF; the software environment outside of
                           the UDF is unaffected.
                         - ``virtualenv``: Use virtualenv to restore the software environment.
```

mlflow/pyfunc/__init__.py
```diff
@@ -977,6 +977,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
                            that was used to train the model. Note that environment is only restored
                            in the context of the PySpark UDF; the software environment outside of
                            the UDF is unaffected.
```
dbczumar, line 979: This note about the environment only being restored for UDF also applies to virtualenv. Instead of repeating it for `conda` and `virtualenv`, can we move it above the list?

```
    :param env_manager: The environment manager to use in order to create the
                        software environment for model inference. Note that environment is only restored
                        in the context of the PySpark UDF; the software environment outside of the UDF is 
                         unaffected. Default value is ``local``, and the following values are supported:
```

mlflow/utils/virtualenv.py
dbczumar: I think we need this because the system path used by subprocesses is not identical to the system path in the driver REPL. We've seen this in the past where `which python` in a subprocess doesn't resolve to the same python binary as the one made available to the REPL.

mlflow/utils/virtualenv.py
```diff
@@ -108,21 +114,26 @@ def _install_python(version):
...
+            pyenv_root = _exec_cmd([pyenv_bin_path, "root"], capture_output=True).stdout.strip()
         path_to_bin = ("bin", "python")
     else:
-        # pyenv-win doesn't provide the `pyenv root` command
-        pyenv_root = os.getenv("PYENV_ROOT")
         if pyenv_root is None:
-            raise MlflowException("Environment variable 'PYENV_ROOT' must be set")
+            # pyenv-win doesn't provide the `pyenv root` command
+            pyenv_root = os.getenv("PYENV_ROOT")
+            if pyenv_root is None:
```
harupy, line 135: Do we need the second `if pyenv_root is None:`?

dbczumar: We need to throw, but the following is probably cleaner:

```
pyenv_root = pyenv_root or os.getenv("PYENV_ROOT")
if pyenv_root is None:
    raise MlflowException("Environment variable 'PYENV_ROOT' must be set")
```

harupy: I missed `pyenv_root = os.getenv("PYENV_ROOT")`. Yeah, that's cleaner.

mlflow/utils/virtualenv.py
harupy: Got it, let's keep this.

mlflow/utils/process.py
```diff
@@ -56,21 +62,31 @@ def _exec_cmd(
...
-    )
-
-    if throw_on_error and prc.returncode != 0:
-        raise ShellCommandException.from_completed_process(prc)
-    return prc
+
+    # In Python < 3.8, `subprocess.Popen` doesn't accpet a command containing path-like
+    # objects (e.g. `["ls", pathlib.Path("abc")]`) on Windows. To avoid this issue,
+    # stringify all elements in `cmd`. Note `str(pathlib.Path("abc"))` returns 'abc'.
+    cmd = list(map(str, cmd))
```
WeichenXu123, line 69: `ShellCommandException.from_completed_process` print log includes the "cmd", the `map` object cmd is not readable.

mlflow/pyfunc/__init__.py
```diff
@@ -969,14 +969,16 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
...
+                        of the PySpark UDF; the software environment outside of the UDF is
+                        unaffected. Default value is ``local``, and the following values are
+                        supported:
 
                          - ``conda``: (Recommended) Use Conda to restore the software environment
-                           that was used to train the model. Note that environment is only restored
-                           in the context of the PySpark UDF; the software environment outside of
-                           the UDF is unaffected.
+                           that was used to train the model.
+                         - ``virtualenv``: Use virtualenv to restore the software environment that
```
harupy, line 980:
```suggestion
                         - ``virtualenv``: Use virtualenv to restore the python environment that
```

mlflow/utils/process.py
```diff
@@ -56,21 +62,31 @@ def _exec_cmd(
...
+        )
+        if throw_on_error and prc.returncode != 0:
+            raise ShellCommandException.from_completed_process(prc)
+        return prc
+    else:
+        return subprocess.Popen(
+            cmd,
+            env=env,
+            text=True,
+            **kwargs,
```
harupy, line 88: Can we set `stdout` and `stderr` if `capture_output` is True?

mlflow/utils/file_utils.py
```diff
@@ -499,7 +499,9 @@ def get_or_create_tmp_dir():
         # Note: For python process attached to databricks notebook, atexit does not work.
         # The /tmp/repl_tmp_data/{repl_id} directory will be removed once databricks notebook
         # detaches.
-        tmp_dir = os.path.join("/tmp", "repl_tmp_data", get_repl_id())
+        # The repl_tmp_data directory is designed to be used by all kinds of applications,
```
harupy, line 502: Can you elaborate on `designed to be used by all kinds of applications`?

WeichenXu123: e.g., on databricks, any applications can write data into `/tmp/repl_tmp_data/{repl_id}` directory, to prevent path conflict, we'd better assign a child directory for each application.
and when notebook detached, the root directory `/tmp/repl_tmp_data/{repl_id}` is automatically deleted.

mlflow/pyfunc/__init__.py
```diff
@@ -969,14 +969,16 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
 
         - ``ArrayType(StringType)``: All columns converted to ``string``.
 
-    :param env_manager: The environment manager to use in order to create the
-                        software environment for model inference. Default value is ``local``,
-                        The following values are supported:
+    :param env_manager: The environment manager to use in order to create the software environment
```
harupy, line 972:
```suggestion
    :param env_manager: The environment manager to use in order to create the python environment
```

mlflow/utils/process.py
harupy: Nice!

mlflow/utils/process.py
```diff
@@ -44,8 +45,13 @@ def _exec_cmd(
                       If this argument is specified, `kwargs` cannot contain `env`.
     :param: capture_output: If True, stdout and stderr will be captured and included in an exception
                             message on failure; if False, these streams won't be captured.
-    :param kwargs: Keyword arguments (except `check` and `text`) passed to `subprocess.run`.
-    :return: A `subprocess.CompletedProcess` instance.
+    :param: synchronous: If True, wait process complete and return a CompletedProcess instance,
```
harupy, line 48: I'm not sure if this is a good idea, can we provide two methods (`_exec_cmd_sync` and `_exec_cmd_async`) to simplify the behavior?

mlflow/utils/virtualenv.py
```diff
@@ -192,14 +204,27 @@ def _create_virtualenv(local_model_path, python_bin_path, env_dir, python_env):
...
 
-def _get_or_create_virtualenv(local_model_path, env_id=None):
+def _get_virtualenv_extra_env_vars(env_root_dir=None):
+    extra_env = {
+        # PIP_NO_INPUT=1 make pip run in non-interactive mode,
+        # otherwise pip might prompt "yes or no" and ask stdin input
+        "PIP_NO_INPUT": "1",
+    }
+    if env_root_dir is not None:
+        extra_env["PIP_CACHE_DIR"] = os.path.join(env_root_dir, "pip_cache_pkgs")
```
harupy, line 221:
```suggestion
        extra_env["PIP_CACHE_DIR"] = os.path.join(env_root_dir, PIP_CACHE_PACKAGES_DIR)
```

Can we de-hardcode `"pip_cache_pkgs"`?

mlflow/utils/virtualenv.py
```diff
@@ -245,7 +287,16 @@ def _get_or_create_virtualenv(local_model_path, env_id=None):
...
 
-def _execute_in_virtualenv(activate_cmd, command, install_mlflow, command_env=None):
+def _execute_in_virtualenv(
+    activate_cmd,
+    command,
+    install_mlflow,
+    command_env=None,
+    synchronous=True,
+    capture_output=False,
+    env_root_dir=None,
```
harupy, line 297: Do we need this argument? Can we include `env_root_dir` in `command_env` when we call this function?

mlflow/utils/virtualenv.py
```diff
@@ -220,10 +245,20 @@ def _get_or_create_virtualenv(local_model_path, env_id=None):
     local_model_path = Path(local_model_path)
     python_env = _get_python_env(local_model_path)
 
+    extra_env = _get_virtualenv_extra_env_vars(env_root_dir)
+    if env_root_dir is not None:
+        virtual_envs_root_path = Path(env_root_dir) / "virtualenv_envs"
+        pyenv_root_dir = os.path.join(env_root_dir, "pyenv")
```
harupy, line 251: This line looks like we assume `env_root_dir` contains `pyenv`.

harupy, line 251: This line assumes `env_root_dir` contains `pyenv`. Do we copy `pyenv` in `env_root_dir` somewhere?

mlflow/utils/virtualenv.py
```diff
@@ -254,10 +305,35 @@ def _execute_in_virtualenv(activate_cmd, command, install_mlflow, command_env=No
...
+        # Add "exec" before the starting scoring server command, so that the scoring server
+        # process replaces the bash process, otherwise the scoring server process is created
+        # as a child process of the bash process.
+        # Note we in `mlflow.pyfunc.spark_udf`, use prctl PR_SET_PDEATHSIG to ensure scoring
+        # server process being killed when UDF process exit. The PR_SET_PDEATHSIG can only
+        # send signal to the bash process, if the scoring server process is created as a
+        # child process of the bash process, then it cannot receive the signal sent by prctl.
+        # TODO: For Windows, there's no equivalent things of Unix shell's exec. Windows also
+        #  does not support prctl. We need to find an approach to address it.
+        command = "exec " + command
```
harupy, line 333: Can we pull this out from this function?

mlflow/utils/process.py
dbczumar: @harupy Why isn't this a good idea? We do it elsewhere.

mlflow/utils/process.py
```diff
@@ -56,21 +62,35 @@ def _exec_cmd(
...
-        capture_output=capture_output,
-        text=True,
-        **kwargs,
-    )
-
-    if throw_on_error and prc.returncode != 0:
-        raise ShellCommandException.from_completed_process(prc)
-    return prc
+
+    # In Python < 3.8, `subprocess.Popen` doesn't accpet a command containing path-like
```
dbczumar, line 66:
```suggestion
    # In Python < 3.8, `subprocess.Popen` doesn't accept a command containing path-like
```

mlflow/utils/virtualenv.py
dbczumar: I think `env_root_dir` is better as a standalone parameter because it's well-documented and corresponds to the `PyFuncBackend` API.

mlflow/utils/virtualenv.py
dbczumar: @WeichenXu123 Does `_install_python` create this directory? Doesn't seem like we're creating it here.

dbczumar (approved):
LGTM once remaining comments are addressed. Thanks a bunch, @WeichenXu123 !

mlflow/utils/process.py
harupy: I just don't prefer a function that returns the union of multiple objects (`Union[obj1, obj2, ...]`). We currently don't use a type checker like mypy, but if we did, such a function might cause an issue like below:

```python
import typing as t


def get_str_or_int(flag: bool) -> t.Union[str, int]:
    return "foo" if flag else 0


r = get_str_or_int(True)
# this works fine but mypy complains `Item "int" of "Union[str, int]" has no attribute "upper"`
r.upper()

# this also works and mypy is happy now but every time we want to call string methods on `r`,
# we need `isinstance`
if isinstance(r, str):
    r.upper()
```

> We do it elsewhere.

True, let's keep the current code.

mlflow/utils/process.py
```diff
@@ -56,21 +62,35 @@ def _exec_cmd(
...
+            text=True,
+            **kwargs,
+        )
+        if throw_on_error and prc.returncode != 0:
+            raise ShellCommandException.from_completed_process(prc)
+        return prc
+    else:
+        if capture_output:
+            kwargs["stdout"] = subprocess.PIPE
+            kwargs["stderr"] = subprocess.PIPE
```
harupy, line 86:
```suggestion
        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs["stdout"] = subprocess.PIPE
            kwargs["stderr"] = subprocess.PIPE
```

Let's use the same validation as `subprocess`

https://github.com/python/cpython/blob/0897a0bf9c75bacf428b8f6f08114fffc9f9c542/Lib/subprocess.py#L495

mlflow/utils/process.py
harupy: Can we only `Popen` instead of using both `Popen` and `subprocess.run`?

```python
import subprocess


def _exec_cmd(
    cmd,
    *,
    throw_on_error=True,
    extra_env=None,
    capture_output=True,
    synchronous=True,
    **kwargs,
):
    ...
    if capture_output:
        kwargs["stdout"] = subprocess.PIPE
        kwargs["stderr"] = subprocess.PIPE

    process = subprocess.Popen(
        cmd,
        env=env,
        text=True,
        **kwargs,
    )
    if not synchronous:
        return process

    stdout, stderr = process.communicate()
    returncode = process.poll()
    comp_process = subprocess.CompletedProcess(
        process.args,
        returncode=returncode,
        stdout=stdout,
        stderr=stderr,
    )
    if throw_on_error and returncode != 0:
        raise ShellCommandException.from_completed_process(comp_process)
    return comp_process

```

mlflow/utils/virtualenv.py
harupy: Agree!

mlflow/utils/virtualenv.py
WeichenXu123: To clarify, the "pyenv_root_dir" is used to holds the python binaries downloaded by pyenv. We can still use "pyenv" installed in other directory .

> Does _install_python create this directory? Doesn't seem like we're creating it here.
I added `pyenv_root_dir.mkdir(parents=True, exist_ok=True)` to create it. (Don't create it is also fine, pyenv will automatically create it if it does not exist. But create it in advance is safer)

harupy: Makes sense now. Can you add a comment above https://github.com/mlflow/mlflow/pull/5676/files#diff-addd2cf35dec6d4143b239c1a5b26d2a24d2a93540c3d4555ab289c1667b4017R117?

mlflow/utils/virtualenv.py
WeichenXu123: What's the function name for the pull out part do you suggest ?

mlflow/utils/process.py
```diff
@@ -44,8 +45,13 @@ def _exec_cmd(
...
-    :return: A `subprocess.CompletedProcess` instance.
+    :param: synchronous: If True, wait process complete and return a CompletedProcess instance,
+                         If False, does not wait process complete and return a Popen instance,
+                         and ignore the `throw_on_error` argument.
+    :param kwargs: Keyword arguments (except `check` and `text`) passed to `subprocess.run` or
+                   `subproces.Popen`.
+    :return:  If synchronous is True, return a `subprocess.CompletedProcess` instance,
+              otherwise return a Popen instance.
     """
     illegal_kwargs = set(kwargs.keys()).intersection(("check", "text"))
```
harupy, line 56: I think we can remove `check` now

mlflow/utils/process.py
```diff
@@ -44,8 +45,13 @@ def _exec_cmd(
                       If this argument is specified, `kwargs` cannot contain `env`.
     :param: capture_output: If True, stdout and stderr will be captured and included in an exception
                             message on failure; if False, these streams won't be captured.
-    :param kwargs: Keyword arguments (except `check` and `text`) passed to `subprocess.run`.
-    :return: A `subprocess.CompletedProcess` instance.
+    :param: synchronous: If True, wait process complete and return a CompletedProcess instance,
+                         If False, does not wait process complete and return a Popen instance,
+                         and ignore the `throw_on_error` argument.
+    :param kwargs: Keyword arguments (except `check` and `text`) passed to `subprocess.run` or
+                   `subproces.Popen`.
```
harupy, line 52:
```suggestion
    :param kwargs: Keyword arguments (except `text`) passed to 
 `subproces.Popen`.
```

Let's remove `subprocess.run` and `check`.

mlflow/utils/virtualenv.py
WeichenXu123: add it in `def _install_python(version, pyenv_root=None, capture_output=False)` arugment doc

mlflow/utils/virtualenv.py
```diff
@@ -91,12 +97,16 @@ def _find_latest_installable_python_version(version_prefix):
...
-def _install_python(version):
+def _install_python(version, pyenv_root=None, capture_output=False):
     """
     Installs a specified version of python with pyenv and returns a path to the installed python
     binary.
 
     :param version: Python version to install.
+    :param pyenv_root: PYENV_ROOT path string. If None, use default PYENV_ROOT. The path is used
+                       to hold the python binaries downloaded by pyenv. We can still use "pyenv"
+                       command installed in other directory.
```
harupy, line 108:
```suggestion
    :param pyenv_root: The value of the "PYENV_ROOT" environment variable used when installing python using `pyenv install`.
```

mlflow/utils/process.py
```diff
@@ -33,21 +33,26 @@ def _exec_cmd(
...
 
     :param cmd: The command to run, as a list of strings.
     :param throw_on_error: If True, raises an Exception if the exit code of the program is nonzero.
     :param extra_env: Extra environment variables to be defined when running the child process.
                       If this argument is specified, `kwargs` cannot contain `env`.
     :param: capture_output: If True, stdout and stderr will be captured and included in an exception
                             message on failure; if False, these streams won't be captured.
-    :param kwargs: Keyword arguments (except `check` and `text`) passed to `subprocess.run`.
-    :return: A `subprocess.CompletedProcess` instance.
+    :param: synchronous: If True, wait process complete and return a CompletedProcess instance,
```
harupy, line 48:
```suggestion
    :param: synchronous: If True, wait for process to complete and return a CompletedProcess instance,
```

mlflow/utils/process.py
```diff
@@ -33,21 +33,26 @@ def _exec_cmd(
...
+                         If False, does not wait process complete and return a Popen instance,
```
harupy, line 49:
```suggestion
                         If False, does not wait for process to complete and return a Popen instance,
```

mlflow/utils/virtualenv.py
```diff
@@ -192,14 +207,32 @@ def _create_virtualenv(local_model_path, python_bin_path, env_dir, python_env):
...
         finally:
             local_model_path.joinpath(tmp_req_file).unlink()
 
     return activate_cmd
 
 
-def _get_or_create_virtualenv(local_model_path, env_id=None):
+def _get_virtualenv_extra_env_vars(env_root_dir=None):
+    extra_env = {
+        # PIP_NO_INPUT=1 make pip run in non-interactive mode,
```
harupy, line 219:
```suggestion
        # PIP_NO_INPUT=1 makes pip run in non-interactive mode,
```

harupy (approved):
Looks good to me!

PR #5683, opened 2022-04-12
Added autologging safety validation exemptions, and fixed autolog does not log implicit batch_size for tf keras fit method

### LINKED ISSUE #5339: [BUG] tensorflow autolog doesn't capture batch size correct for tf Dataset input
### Willingness to contribute

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:
- **MLflow installed from (source or binary)**: pip?
- **MLflow version (run ``mlflow --version``)**: 1.23.1
- **Python version**: 3.8
- **npm version, if running the dev UI**: n/a
- **Exact command to reproduce**: see code below

### Describe the problem

When calling `fit` with a tensorflow dataset with a batch operation applied, the mlflow tracking ui records batch size of None, not the correct batch size. Note that [the fit docs](https://keras.io/api/models/model_training_apis/#fit-method) instruct to not supply `batch_size` when the input data is a dataset:
> batch_size: ... Do not specify the batch_size if your data is in the form of datasets, generators, or keras.utils.Sequence instances

The cause of this appears to be [this line](https://github.com/mlflow/mlflow/blob/1335164d20a6521dd4ebfe3f361d19fe9672efcf/mlflow/tensorflow/__init__.py#L860) in the patched fit function for auto-logging, which logs the batch_size param, regardless of whether the input to fit was a batched dataset.

### Code to reproduce issue

```python
import mlflow
import tensorflow as tf


def gen():
    for i in range(10):
        yield i, -i

dataset = tf.data.Dataset.from_generator(
    gen,
    output_signature=(
        tf.TensorSpec(shape=(), dtype=tf.float32),
        tf.TensorSpec(shape=(), dtype=tf.float32),
    )
)
dataset = dataset.batch(2)

mlflow.tensorflow.autolog()

model = tf.keras.Sequential()
model.add(tf.keras.layers.Dense(1, input_shape=(1,)))
model.compile(loss="mse")
model.fit(dataset, epochs=10)
```
### Other info / logs

N/A

Components 

Interface 

Language 

Integrations

MarkYHZhang:

Fixes https://github.com/mlflow/mlflow/issues/5339 by providing implicit `batch_size` extraction for `tf.data.Dataset`, generator, and `tf.keras.utils.Sequence`. As part of this PR, I also added a way to add autologging safety validation exemptions.

## How is this patch tested?

Unit tests.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Autolog will now log implicit `batch_size` for Tensorflow Keras models with input data of types: `tf.data.Dataset`, generator, or `tf.keras.utils.Sequence`. 

Components 

Interface 

Language 

Integrations


mlflow/tensorflow/__init__.py
```diff
@@ -897,6 +898,26 @@ def _patch_implementation(
         ):  # pylint: disable=arguments-differ
             unlogged_params = ["self", "x", "y", "callbacks", "validation_data", "verbose"]
 
+            batch_size = None
+            if isinstance(args[0], tensorflow.data.Dataset):
```
dbczumar, line 902: For readability, can we name `args[0]`? e.g. `training_data = args[0]`?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -24,6 +25,13 @@
 
 _AUTOLOGGING_PATCHES = {}
 
+_MUTABLE_ARGS_ALLOWLIST = {
+    # This mutation is allowed since when patching the tensorflow keras fit function for autolog,
+    # we need to infer batch_size from 'x' if 'x' is a generator. This is because the documentation
+    # for the fit function instructs user not to supply batch_size explicitly if 'x' is a generator
+    "tensorflow": {"fit": (types.GeneratorType,)}
```
dbczumar, line 32: Can we scope this to a particular index too?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -24,6 +25,13 @@
 
 _AUTOLOGGING_PATCHES = {}
 
+_MUTABLE_ARGS_ALLOWLIST = {
```
dbczumar, line 28: Can we add a comment above this list explaining what it's for? I.e for specifying a list of function arguments that can be mutated by autologging code. We should also warn developers that this list shouldn't be modified unless absolutely necessary and that, if modified, a clear reason needs to be provided as an inline comment + thorough integration tests need to be introduced.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -834,6 +854,8 @@ def _validate(autologging_call_input, user_call_input=None):
             - for all other input types, `autologging_call_input` and `user_call_input`
               must be equivalent by reference equality or by object equality
         """
+        nonlocal autologging_integration, function_name
```
dbczumar, line 857: I don't think we need `nonlocal` here since we're not mutating these vars, right?

mlflow/tensorflow/__init__.py
```diff
@@ -909,6 +910,27 @@ def _patch_implementation(
         ):  # pylint: disable=arguments-differ
             unlogged_params = ["self", "x", "y", "callbacks", "validation_data", "verbose"]
 
+            batch_size = None
+            training_data = args[0]
```
apurva-koti, line 914: What if someone passes in `training_data` via `kwargs`?

mlflow/tensorflow/__init__.py
```diff
@@ -909,6 +910,27 @@ def _patch_implementation(
...
+            if isinstance(training_data, tensorflow.data.Dataset):
+                batch_size = training_data._batch_size.numpy()
+            elif isinstance(training_data, tensorflow.keras.utils.Sequence):
+                first_batch = training_data[0]
+                first_batch_inputs, _ = first_batch
```
apurva-koti, line 919: Nit: don't need `first_batch`

mlflow/tensorflow/__init__.py
```diff
@@ -909,6 +910,27 @@ def _patch_implementation(
...
+                batch_size = len(first_batch_inputs)
+            elif is_generator(args[0]):
```
apurva-koti, line 921: Shouldn't we refer to `training_data` instead of `args[0]`?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +780,50 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    """
+
+    # When extracting implicitly defined `batch_size` in the case that `x` is a generator or a
+    # generator class, we need to consume and restore the first element back to the generator to
+    # calculate the `batch_size`. This means that:
+    # 1. The type of `x` will become 'generator' regardless if user provided `x` as a generator or a
+    #    custom generator class.
+    # 2. The instance of `x` will be different, since we reconstructed the generator after consuming
+    #    the first element.
+    if autologging_integration == "tensorflow":
```
apurva-koti, line 812: can we use `mlflow.tensorflow.FLAVOR_NAME`

mlflow/utils/autologging_utils/__init__.py
```diff
@@ -591,3 +591,11 @@ def get_method_call_arg_value(arg_index, arg_name, default_value, call_pos_args,
         return call_pos_args[arg_index]
     else:
         return default_value
+
+
+def is_generator(obj):
```
dbczumar, line 596: Nit: I think this should be named `is_iterator`, since the checks implemented are related to iterator attributes. This *is* the right thing to check, but we should adjust the naming.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +780,50 @@ def _validate_autologging_run(autologging_integration, run_id):
     ), "Autologging run with id {} has a non-terminal status '{}'".format(run_id, run.info.status)
 
 
+def __is_arg_exempt_from_validation(
```
dbczumar, line 783:
```suggestion
def _is_arg_exempt_from_validation(
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +780,50 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    safety validations. This includes both type checking and immutable checking.
+
+    WARNING: Exemptions should NOT be introduced unless absolutely necessary. If deemed necessary,
+             clear reasons must be provided as comment in addition to thorough integration tests.
+
+    :param autologging_integration: the name of the autologging integration
+    :param function_name: the name of the function that is being validated
+    :param arg: the actual argument
+    :param pos_arg_index: the index of the argument, if it is a positional argument
+    :param kwarg_name: the name of the argument, if it is a keyword argument
```
dbczumar, line 801:
```suggestion
    :param pos_arg_index: the index of the argument, if it is passed as a positional argument
    :param kwarg_name: the name of the argument, if it is passed as a keyword argument
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -874,8 +951,8 @@ def _validate(autologging_call_input, user_call_input=None):
                 " Original: '{}'. Expected: '{}'".format(autologging_call_input, user_call_input)
             )
 
-    _validate(autologging_call_args, user_call_args)
-    _validate(autologging_call_kwargs, user_call_kwargs)
+    _validate(0, autologging_call_args, user_call_args)
+    _validate(0, autologging_call_kwargs, user_call_kwargs)
```
dbczumar, line 955: Rather than tracking the recursive validation stack depth and performing validation in the case where `depth==1`, I'd rather we pull the depth=0 validation procedure out of the `_validate` function, even if we have to repeat some code.

For example:

```
length_difference = len(autologging_call_args) - len(user_call_args)
assert (
    length_difference >= 0
), "{} expected inputs are missing from the call to the original function.".format(
    length_difference
)

for autologging_call_arg, user_call_arg in itertools.zip_longest(autologging_call_args, user_call_args):
    if not _is_arg_exempt_from_validation(...):
        _validate(autologging_call_arg, user_call_arg)

assert set(user_call_kwargs.keys()).issubset(set(autologging_call_kwargs.keys())), (
    "Keyword or dictionary arguments to original function omit"
    " one or more expected keys: '{}'".format(
        set(user_call_kwargs.keys()) - set(autologging_call_kwargs.keys())
    )
)

for key in autologging_call_kwargs.keys():
    if not _is_arg_exempt_from_validation(...):
        _validate(
            autologging_call_kwargs[key],
            user_call_kwargs.get(key, None),
        )
```

We can pull the logic for list / dict length and key subset validation into separate methods and reuse them here & in `_validate`.

dbczumar: Additionally, when validating positional args, can we pass `autologging_call_kwargs` to `_is_arg_exempt_from_validation`? This will tell us how to interpret the positional arg index, which can change based on whether or not an argument is specified as a kwarg.

For example, consider the following function accepting integer arguments:

```
def func(foo, bar):
```

if arg `foo` is exempt from validation and the method is called as `func(bar=1, foo=2)`, the current logic in `_is_arg_exempt_from_validation` will exempt `bar` *and* `foo`.

We should only exempt a positional arg if its name is not found in the set of kwargs.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +780,50 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    # When extracting implicitly defined `batch_size` in the case that `x` is a generator or a
+    # generator class, we need to consume and restore the first element back to the generator to
+    # calculate the `batch_size`. This means that:
+    # 1. The type of `x` will become 'generator' regardless if user provided `x` as a generator or a
+    #    custom generator class.
+    # 2. The instance of `x` will be different, since we reconstructed the generator after consuming
+    #    the first element.
+    if autologging_integration == mlflow.tensorflow.FLAVOR_NAME:
+        if function_name == "fit":
+            if pos_arg_index == 1 or kwarg_name == "x":
```
dbczumar, line 814: I think this check will lead us to over-exempting positional args in the unlikely scenario where multiple adjacent args have the same type. See https://github.com/mlflow/mlflow/pull/5683/files#r851456203.

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -1141,16 +1141,16 @@ def test_validate_args_succeeds_when_arg_sets_are_equivalent_or_identical():
         "biz": {"baz": 5},
     }
 
-    _validate_args(args, kwargs, args, kwargs)
-    _validate_args(args, None, args, None)
-    _validate_args(None, kwargs, None, kwargs)
+    _validate_args("", "", args, kwargs, args, kwargs)
```
dbczumar, line 1144: Nit: Can we pass the strings `"autologging_integration_name"` and `"function_name"` as arguments, rather than empty strings, so that maintainers know how to interpret the purpose / meaning of these args?

tests/tensorflow/test_tensorflow2_autolog.py
```diff
@@ -260,6 +261,111 @@ def test_tf_keras_autolog_logs_expected_data(tf_keras_random_data_run):
...
+    model.fit(x=generate_data(batch_size), verbose=0)
+    assert mlflow.last_active_run().data.params["batch_size"] == str(batch_size)
+
+
+@pytest.mark.large
+@pytest.mark.skipif(
+    Version(tf.__version__) < Version("2.1.4"),
+    reason="Does not support passing of generator classes as `x` in `fit`",
+)
+@pytest.mark.parametrize("generator", [__generator, __GeneratorClass])
```
dbczumar, line 338: Are there any other types of generators used with tf.keras that we should test?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +780,50 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    This function is responsible for determining whether or not an argument is exempt from autolog
+    safety validations. This includes both type checking and immutable checking.
+
+    WARNING: Exemptions should NOT be introduced unless absolutely necessary. If deemed necessary,
+             clear reasons must be provided as comment in addition to thorough integration tests.
+
+    :param autologging_integration: the name of the autologging integration
+    :param function_name: the name of the function that is being validated
+    :param arg: the actual argument
+    :param pos_arg_index: the index of the argument, if it is a positional argument
```
dbczumar, line 800: Nit: In a function signature, an argument always has exactly one index. At call time, the function may be passed positionally at that index or may be passed via kwargs. This function refers to an argument's characteristics in the context of the function signature, so I think `arg_index` is a more appropriate name.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +780,50 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    # 1. The type of `x` will become 'generator' regardless if user provided `x` as a generator or a
+    #    custom generator class.
+    # 2. The instance of `x` will be different, since we reconstructed the generator after consuming
+    #    the first element.
+    if autologging_integration == mlflow.tensorflow.FLAVOR_NAME:
+        if function_name == "fit":
+            if pos_arg_index == 1 or kwarg_name == "x":
+                from mlflow.utils.autologging_utils import is_generator
+
+                return is_generator(arg)
```
dbczumar, line 817: For testing purposes, can we move this to a more generic allowlist form like we had in the previous commits? I propose a list of named tuples with the following attributes:

1. autologging integration
2. function name
3. argument name
4. argument index
5. type_fn (a function that takes an object and returns True if the object matches the argument type, False otherwise)

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -1320,15 +1382,55 @@ def test_validate_args_throws_when_arg_types_or_values_are_changed():
...
+            "",
+            user_call_args,
+            user_call_kwargs,
+            user_call_args,
+            invalid_autologging_call_kwargs_2,
         )
 
 
+@pytest.mark.usefixtures(test_mode_on.__name__)
+def test_validate_args_exempts_tensorflow_keras_fit_x_generator():
```
dbczumar, line 1405: I think we're overfitting to the tensorflow case for unit testing argument exemption. In https://github.com/mlflow/mlflow/pull/5683/files#r851466125, I've proposed that we instead define exemptions as allowlist entries. Then, we can mock the allowlist here and verify that arguments are exempted properly for different cases (e.g. try a couple of different functions & types).

This way, we can also test that, e.g. if the type doesn't match or the function name doesn't match or the integration name doesn't match, arguments aren't exempted (negative testing).

mlflow/utils/autologging_utils/safety.py
dbczumar: Nevermind, this refers to the index at which the argument was passed at call time. Can we explain that it's `None` if it's a keyword arg? I think `pos_arg_index` is a good name.

tests/tensorflow/test_tensorflow2_autolog.py
MarkYHZhang: There exist tf.keras built-in generators like: [ImageDataGenerator](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/image/ImageDataGenerator?version=nightly) and [TimeseriesGenerator](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/sequence/TimeseriesGenerator?version=nightly). But they all inherits from [Sequence](https://www.tensorflow.org/api_docs/python/tf/keras/utils/Sequence) or [Iterator](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/image/Iterator?version=nightly) (which also itself inherits from [Sequence](https://www.tensorflow.org/api_docs/python/tf/keras/utils/Sequence)) which are already being tested. Additionally, they will soon be deprecated in future versions:

> Prefer loading data with either [tf.keras.utils.text_dataset_from_directory](https://www.tensorflow.org/api_docs/python/tf/keras/utils/text_dataset_from_directory) or [tf.keras.utils.image_dataset_from_directory](https://www.tensorflow.org/api_docs/python/tf/keras/utils/image_dataset_from_directory), and then transforming the output [tf.data.Dataset](https://www.tensorflow.org/api_docs/python/tf/data/Dataset) with preprocessing layers.

[Source](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing?version=nightly)

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,98 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+class ValidationExemptArgument(typing.NamedTuple):
+    """
+    A NamedTuple representing the properties of an argument that is exempt from validation
+
+    autologging_integration: the name of the autologging integration
+    function_name: the name of the function that is being validated
+    type_function: a Callable that accepts an object and returns True if the given object matches
+                   the argument type. Returns False otherwise.
+    positional_argument_index: the index of the argument, if it is passed as a positional argument.
+                               Otherwise, it is None.
```
dbczumar, line 794: For the purpose of defining an allowlist, this should always be defined. It's the index in the function signature.

```suggestion
    argument_index: the index of the argument in the function signature.
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,98 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    keyword_argument_name: the name of the argument, if it is passed as a keyword argument.
+                           Otherwise, it is None.
```
dbczumar, line 796:
```suggestion
    argument_name: the name of the argument in the function signature.
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,98 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    keyword_argument_name: str = None
+
+    def is_exempt(
+        self,
+        autologging_integration,
+        function_name,
+        argument,
+        positional_argument_index=None,
+        keyword_argument_name=None,
+    ):
```
dbczumar, line 812: Can we document this method?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -820,6 +919,37 @@ def _validate_new_input(inp):
...
             )
 
+    def _assert_same_type(autologging_call_input, user_call_input):
+        assert type(autologging_call_input) == type(
+            user_call_input
+        ), "Type of input to original function '{}' does not match expected type '{}'".format(
+            type(autologging_call_input), type(user_call_input)
+        )
+
+    def _assert_same_input_length(autologging_call_input, user_call_input):
```
dbczumar, line 929: This method name isn't quite accurate. It's asserting that the autologging call input length is `>=` the user call input length to make sure that we didn't delete user arguments.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -820,6 +919,37 @@ def _validate_new_input(inp):
...
+        length_difference = len(autologging_call_input) - len(user_call_input)
+        assert (
+            length_difference >= 0
+        ), "{} expected inputs are missing from the call to the original function.".format(
+            length_difference
+        )
+
+    def _assert_same_key_set(autologging_call_input, user_call_input):
```
dbczumar, line 937: This method name isn't quite accurate. The method is asserting that the autologging call input kwarg keys are a superset of the user call input kwarg keys to make sure that we didn't delete user keyword arguments.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -820,6 +919,37 @@ def _validate_new_input(inp):
...
+        assert set(user_call_input.keys()).issubset(set(autologging_call_input.keys())), (
+            "Keyword or dictionary arguments to original function omit"
+            " one or more expected keys: '{}'".format(
+                set(user_call_input.keys()) - set(autologging_call_input.keys())
+            )
+        )
+
+    def _assert_equivalence(autologging_call_input, user_call_input):
```
dbczumar, line 945:
```suggestion
    def _assert_equivalent(autologging_call_input, user_call_input):
```

tests/autologging/test_autologging_safety_unit.py
```diff
@@ -1168,12 +1172,22 @@ def test_validate_args_throws_when_extra_args_are_not_functions_classes_or_lists
 
     with pytest.raises(Exception, match="Invalid new input"):
         _validate_args(
-            user_call_args, user_call_kwargs, invalid_type_autologging_call_args, user_call_kwargs
+            "",
+            "",
```
dbczumar, line 1176: @MarkYHZhang Can we apply https://github.com/mlflow/mlflow/pull/5683/files#diff-c18d334f6497b607d53017ccc716f80d1f2ff2cd989358c5bca64ae877f545ccR1146-R1148 to the rest of this file as well? I.e. instead of passing empty strings, include content that helps maintainers interpret the meaning of these args.

mlflow/utils/autologging_utils/safety.py
```diff
@@ -833,49 +963,73 @@ def _validate(autologging_call_input, user_call_input=None):
...
+            for index, autologging_call_arg, user_call_arg in itertools.zip_longest(
+                range(len(user_call_args)), autologging_call_args, user_call_args
+            ):
+                if not _is_arg_exempt_from_validation(
+                    autologging_integration,
+                    function_name,
+                    user_call_arg,
+                    positional_argument_index=index,
+                ):
+                    _validate(autologging_call_arg, user_call_arg)
```
dbczumar, line 1010: Positional arguments of a function call are always tuples. For simplicity, can we rewrite this logic with that assumption?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -833,49 +963,73 @@ def _validate(autologging_call_input, user_call_input=None):
...
+                if not _is_arg_exempt_from_validation(
+                    autologging_integration,
+                    function_name,
+                    user_call_kwargs.get(key, None),
+                    keyword_argument_name=key,
+                ):
+                    _validate(
+                        autologging_call_kwargs[key],
+                        user_call_kwargs.get(key, None),
+                    )
```
dbczumar, line 1032: The set of keyword arguments of a function call is always a dictionary. For simplicity, can we rewrite this logic with that assumption?

mlflow/utils/autologging_utils/safety.py
dbczumar, line 1010: The set of positional arguments of a function is always a tuple. For simplicity, can we rewrite this logic with that assumption?

tests/tensorflow/test_tensorflow2_autolog.py
dbczumar: Thanks!

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,107 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    keyword_argument_name: The name of the argument in the function signature.
+    """
+
+    autologging_integration: str
+    function_name: str
+    type_function: typing.Callable
+    positional_argument_index: int = None
+    keyword_argument_name: str = None
+
+    def match(
```
dbczumar, line 803:
```suggestion
    def matches(
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,107 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+        self,
+        autologging_integration,
+        function_name,
+        argument,
```
dbczumar, line 807:
```suggestion
        value,
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,107 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    type_function: A Callable that accepts an object and returns True if the given object matches
+                   the argument type. Returns False otherwise.
+    positional_argument_index: The index of the argument in the function signature.
+    keyword_argument_name: The name of the argument in the function signature.
+    """
+
+    autologging_integration: str
+    function_name: str
+    type_function: typing.Callable
+    positional_argument_index: int = None
```
dbczumar, line 800:
```suggestion
    argument_index: int = None
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,107 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    keyword_argument_name: str = None
```
dbczumar, line 801:
```suggestion
    argument_name: str = None
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,107 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+
+    def match(
+        self,
+        autologging_integration,
+        function_name,
+        argument,
+        positional_argument_index=None,
+        keyword_argument_name=None,
+    ):
+        """
```
dbczumar, line 811: In the docstring, can we explain what this method does?

mlflow/utils/autologging_utils/safety.py
```diff
@@ -773,8 +782,107 @@ def _validate_autologging_run(autologging_integration, run_id):
...
+    return any(
+        exemption.match(
+            autologging_integration,
+            function_name,
+            argument,
+            positional_argument_index,
+            keyword_argument_name,
+        )
+        for exemption in _VALIDATION_EXEMPT_ARGUMENTS
+    )
```
dbczumar, line 876: Nice! This reads much better now :)

mlflow/utils/autologging_utils/safety.py
```diff
@@ -820,6 +928,24 @@ def _validate_new_input(inp):
...
+        autologging_call_input, user_call_input
+    ):
+        length_difference = len(autologging_call_input) - len(user_call_input)
+        assert (
+            length_difference >= 0
+        ), "{} expected inputs are missing from the call to the original function.".format(
+            length_difference
+        )
+
+    def _assert_autologging_call_input_is_superset(autologging_call_input, user_call_input):
```
dbczumar, line 941:
```suggestion
    def _assert_autologging_input_kwargs_are_superset(autologging_call_input, user_call_input):
```

mlflow/utils/autologging_utils/safety.py
```diff
@@ -820,6 +928,24 @@ def _validate_new_input(inp):
                 "such exception safe functions / classes.".format(inp)
             )
 
+    def _assert_autologging_input_length_greater_than_or_equal_to(
```
dbczumar, line 931:
```suggestion
    def _assert_autologging_input_positional_args_are_superset(
```

dbczumar (approved):
LGTM once a few nitty naming comments are addressed. Thanks a bunch, @MarkYHZhang !

PR #5684, opened 2022-04-13
Fix temp directory permission issue on worker side

WeichenXu123:

## How is this patch tested?

(Details)

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/utils/file_utils.py
+++ b/mlflow/utils/file_utils.py
@@ -503,6 +503,7 @@ def get_or_create_tmp_dir():
         os.makedirs(tmp_dir, exist_ok=True)
     else:
         tmp_dir = tempfile.mkdtemp()
+        os.chmod(tmp_dir, 0o777)
         atexit.register(shutil.rmtree, tmp_dir, ignore_errors=True)
 
     return tmp_dir
@@ -526,6 +527,7 @@ def get_or_create_nfs_tmp_dir():
         os.makedirs(tmp_nfs_dir, exist_ok=True)
     else:
         tmp_nfs_dir = tempfile.mkdtemp(dir=nfs_root_dir)
+        os.chmod(tmp_nfs_dir, 0o777)
         atexit.register(shutil.rmtree, tmp_nfs_dir, ignore_errors=True)
 
     return tmp_nfs_dir
--- a/mlflow/utils/file_utils.py
+++ b/mlflow/utils/file_utils.py
@@ -503,7 +503,7 @@ def get_or_create_tmp_dir():
         os.makedirs(tmp_dir, exist_ok=True)
     else:
         tmp_dir = tempfile.mkdtemp()
-        # mkdtemp create directory with permission 0o700
+        # mkdtemp creates a directory with permission 0o700
         # change it to be 0o777 to ensure it can be seen in spark UDF
         os.chmod(tmp_dir, 0o777)
         atexit.register(shutil.rmtree, tmp_dir, ignore_errors=True)
@@ -529,7 +529,7 @@ def get_or_create_nfs_tmp_dir():
         os.makedirs(tmp_nfs_dir, exist_ok=True)
     else:
         tmp_nfs_dir = tempfile.mkdtemp(dir=nfs_root_dir)
-        # mkdtemp create directory with permission 0o700
+        # mkdtemp creates a directory with permission 0o700
         # change it to be 0o777 to ensure it can be seen in spark UDF
         os.chmod(tmp_nfs_dir, 0o777)
         atexit.register(shutil.rmtree, tmp_nfs_dir, ignore_errors=True)
--- a/mlflow/utils/file_utils.py
+++ b/mlflow/utils/file_utils.py
@@ -503,6 +503,8 @@ def get_or_create_tmp_dir():
         os.makedirs(tmp_dir, exist_ok=True)
     else:
         tmp_dir = tempfile.mkdtemp()
+        # mkdtemp create directory with permission 0o700
+        # change it to be 0o777 to ensure it can be seen in spark UDF
         os.chmod(tmp_dir, 0o777)
         atexit.register(shutil.rmtree, tmp_dir, ignore_errors=True)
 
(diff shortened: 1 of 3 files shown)


mlflow/utils/file_utils.py
```diff
@@ -503,6 +503,7 @@ def get_or_create_tmp_dir():
         os.makedirs(tmp_dir, exist_ok=True)
     else:
         tmp_dir = tempfile.mkdtemp()
+        os.chmod(tmp_dir, 0o777)
```
BenWilson2, line 506: do we need the workers to have write authority over this location? Could we do 0o755 instead?

BenWilson2: nevermind. I just checked the docs on `os.makedirs()` and the *default* behavior there is 0o777. 
tempfile.mkdtemp sets 0o700 by default. 
This looks good to me!

BenWilson2 (approved):
LGTM!

WeichenXu123:
Let me add some comments before merging

mlflow/utils/file_utils.py
```diff
@@ -503,6 +503,9 @@ def get_or_create_tmp_dir():
         os.makedirs(tmp_dir, exist_ok=True)
     else:
         tmp_dir = tempfile.mkdtemp()
+        # mkdtemp create directory with permission 0o700
```
harupy, line 506:
```suggestion
        # mkdtemp creates a directory with permission 0o700
        # https://github.com/python/cpython/blob/0859368335d470b9ff33fc53ed9a85ec2654b278/Lib/tempfile.py#L368
        # Change it to 0o777 to ensure it can be seen in spark UDF
```

mlflow/utils/file_utils.py
```diff
@@ -526,6 +529,9 @@ def get_or_create_nfs_tmp_dir():
         os.makedirs(tmp_nfs_dir, exist_ok=True)
     else:
         tmp_nfs_dir = tempfile.mkdtemp(dir=nfs_root_dir)
+        # mkdtemp creates a directory with permission 0o700
+        # change it to be 0o777 to ensure it can be seen in spark UDF
```
harupy, line 533: Does the temporary directory have to be writable?

WeichenXu123: Because "shutils" makedirs also use "0o777", I keep it the same with it

harupy: So it has to be writable?

dbczumar (approved):
LGTM! Thanks @WeichenXu123 !

mlflow/utils/file_utils.py
dbczumar: I don't think it strictly needs to be writeable, but this was the preexisting behavior for the Spark model cache. To avoid regressions, I think it's best to use the same permissions level.

harupy: Makes sense! I asked this because `can be seen` sounds like `should be readable but doesn't have to be writable`.

dbczumar (changes requested):
@WeichenXu123 Before merging, can we resolve `os.makedirs()` permissions exceptions thrown when operating on Table ACL clusters?

mlflow/utils/nfs_on_spark.py
```diff
@@ -25,7 +29,18 @@ def get_nfs_cache_root_dir():
             == "true"
         )
         if nfs_enabled:
-            return "/local_disk0/.ephemeral_nfs"
+            nfs_root_dir = "/local_disk0/.ephemeral_nfs"
+            # Test whether the NFS directory is writable.
+            test_path = os.path.join(nfs_root_dir, uuid.uuid4().hex)
```
harupy, line 34: Do we remove this line before the merge?

harupy: Ah ok

dbczumar: We'll keep this line because we're now testing whether or not we can write to NFS by attempting to create a directory at `test_path`

harupy: Can we use `os.access(nfs_root_dir, os.W_OK)`?

ref: https://stackoverflow.com/a/2113511/6943581

dbczumar: I think we should try to write, as is done in the current PR. The best way to test for write access is to try to write.

From the comments of https://stackoverflow.com/a/2113511/6943581:

```
Testing a directory for just the write bit isn't enough if you want to write files to the directory. You will need to test for the execute bit as well if you want to write into the directory. os.access('/path/to/folder', os.W_OK | os.X_OK) With os.W_OK by itself you can only delete the directory (and only if that directory is empty
```

If we're not careful, we might make a mistake.

dbczumar (approved):
LGTM! Thanks a bunch, @WeichenXu123 !

mlflow/utils/nfs_on_spark.py
harupy: Sounds good!

mlflow/utils/nfs_on_spark.py
```diff
@@ -25,7 +29,18 @@ def get_nfs_cache_root_dir():
...
+            try:
+                os.makedirs(test_path)
+                return nfs_root_dir
+            except Exception:
+                # For databricks cluster enabled Table ACL, we have no permission to access NFS
+                # directory, in this case, return None representing NFS is not available.
+                return None
+            finally:
+                shutil.rmtree(test_path, ignore_errors=True)
```
harupy, line 43: Does this line fail when `test_path` doesn't exist?

dbczumar: Good catch! Fortunately, it swallows:

```
In [11]: shutil.rmtree("/tmp/asfsadfsdf", ignore_errors=True)

In [12]: shutil.rmtree("/tmp/asfsadfsdf")
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-12-8bc109b15d91> in <module>
----> 1 shutil.rmtree("/tmp/asfsadfsdf")

~/opt/anaconda3/lib/python3.8/shutil.py in rmtree(path, ignore_errors, onerror)
    704             orig_st = os.lstat(path)
    705         except Exception:
--> 706             onerror(os.lstat, path, sys.exc_info())
    707             return
    708         try:

~/opt/anaconda3/lib/python3.8/shutil.py in rmtree(path, ignore_errors, onerror)
    702         # lstat()/open()/fstat() trick.
    703         try:
--> 704             orig_st = os.lstat(path)
    705         except Exception:
    706             onerror(os.lstat, path, sys.exc_info())

FileNotFoundError: [Errno 2] No such file or directory: '/tmp/asfsadfsdf'
```

harupy: Confirmed it doesn't

harupy (approved):

Issue #5704, opened 2022-04-15
[BUG] DB IntegrityError is not caught causing the server to hang for over 60 seconds


grybeck:
Thank you for submitting an issue. Please refer to our [issue policy](https://www.github.com/mlflow/mlflow/blob/master/ISSUE_POLICY.md) for additional information about bug reports. For help with debugging your code, please refer to [Stack Overflow](https://stackoverflow.com/questions/tagged/mlflow).

**Please fill in this bug report template to ensure a timely and thorough response.**

### Willingness to contribute
The MLflow Community encourages bug fix contributions. Would you or another member of your organization be willing to contribute a fix for this bug to the MLflow code base?

- [ ] Yes. I can contribute a fix for this bug independently.
- [ ] Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.
- [x] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: CentOS Linux 7
- **MLflow installed from (source or binary)**: binary
- **MLflow version (run ``mlflow --version``)**: mlflow, version 1.23.1
- **Python version**: Python 3.9.5
- **npm version, if running the dev UI**: N/A
- **Exact command to reproduce**: 

```
# this gives an expected error RestException: RESOURCE_ALREADY_EXISTS
mlflow.create_experiment("Default")
# this gives an unexpected error and makes the server unresponsive for over 60 seconds
mlflow.create_experiment("Default", artifact_location="local")
```

### Describe the problem
Describe the problem clearly here. Include descriptions of the expected behavior and the actual behavior.

When I run `mlflow.create_experiment("HelloWorld", artifact_location="local")` where HelloWorld is an experiment that already exists, instead of getting the expected error `RestException: RESOURCE_ALREADY_EXISTS`, the request hangs for over 60 seconds effectively making the server go unresponsive, and finally returns the error: 
```
MlflowException: API request to http://localhost:5000/api/2.0/mlflow/experiments/create failed with exception HTTPConnectionPool(host='localhost', port=5000): Max retries exceeded with url: /api/2.0/mlflow/experiments/create (Caused by ResponseError('too many 500 error responses'))
```

This does not happen without artifact_location argument. 

Server Error: 
```
Traceback (most recent call last):
[...]

PR #5715, opened 2022-04-18
Mlflow db store raise mlflow exception with correct message argument

WeichenXu123:

Mlflow db store raise mlflow exception with correct message argument

## How is this patch tested?

Manually test:
(1) start server first: `mlflow server --backend-store-uri sqlite:///mydb.sqlite --default-artifact-root=local-artifacts --host 0.0.0.0 -w 1`
(2) run mlflow code:
```
import mlflow
mlflow.set_tracking_uri('http://localhost:5000/')
mlflow.create_experiment("HelloWorld", artifact_location="local")
mlflow.create_experiment("HelloWorld", artifact_location="local")
```
the second run "create_experiment" will fail immediately and raise error like: 
```
  File "/Users/weichen.xu/work/mlflow/mlflow/tracking/fluent.py", line 1023, in create_experiment
    return MlflowClient().create_experiment(name, artifact_location, tags)
  File "/Users/weichen.xu/work/mlflow/mlflow/tracking/client.py", line 507, in create_experiment
    return self._tracking_client.create_experiment(name, artifact_location, tags)
  File "/Users/weichen.xu/work/mlflow/mlflow/tracking/_tracking_service/client.py", line 176, in create_experiment
    return self.store.create_experiment(
  File "/Users/weichen.xu/work/mlflow/mlflow/store/tracking/rest_store.py", line 99, in create_experiment
    response_proto = self._call_endpoint(CreateExperiment, req_body)
  File "/Users/weichen.xu/work/mlflow/mlflow/store/tracking/rest_store.py", line 56, in _call_endpoint
  File "/Users/weichen.xu/work/mlflow/mlflow/utils/rest_utils.py", line 256, in call_endpoint
  File "/Users/weichen.xu/work/mlflow/mlflow/utils/rest_utils.py", line 185, in verify_rest_response
mlflow.exceptions.RestException: RESOURCE_ALREADY_EXISTS: (sqlite3.IntegrityError) UNIQUE constraint failed: experiments.name
[SQL: INSERT INTO experiments (name, artifact_location, lifecycle_stage) VALUES (?, ?, ?)]
[parameters: ('HelloWorld', 'local', 'active')]
(Background on this error at: https://sqlalche.me/e/14/gkpj)
```

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components 

Interface 

Language 

Integrations


mlflow/store/db/utils.py
```diff
@@ -83,7 +83,7 @@ def make_managed_session():
             raise
         except Exception as e:
             session.rollback()
-            raise MlflowException(message=e, error_code=INTERNAL_ERROR)
+            raise MlflowException(message=str(e), error_code=INTERNAL_ERROR)
```
WeichenXu123, line 86: The `e` might be database `IntegrityError` which cannot be dumped as JSON, this might cause error such the one reported by https://github.com/mlflow/mlflow/issues/5704

mlflow/store/db/utils.py
```diff
@@ -81,9 +82,12 @@ def make_managed_session():
         except MlflowException:
             session.rollback()
             raise
+        except sqlite3.IntegrityError as e:
+            session.rollback()
+            raise MlflowException(message=str(e), error_code=RESOURCE_ALREADY_EXISTS)
```
WeichenXu123, line 87: `RESOURCE_ALREADY_EXISTS` http error respond code is 400, it should prevent client retries, but in my testing it still retries. What's the reason ?

harupy: Can you share the code you used?

mlflow/store/db/utils.py
```diff
@@ -81,9 +82,12 @@ def make_managed_session():
         except MlflowException:
             session.rollback()
             raise
+        except sqlite3.IntegrityError as e:
```
dbczumar, line 85: This should be `sqlalchemy.exc.IntegrityError`

mlflow/store/db/utils.py
WeichenXu123: is 400 responding code preventing retries ?

mlflow/store/db/utils.py
WeichenXu123: it works!

mlflow/store/db/utils.py
harupy: Can we stringify `message` in `MlflowException`?

mlflow/store/db/utils.py
```diff
@@ -81,9 +82,12 @@ def make_managed_session():
         except MlflowException:
             session.rollback()
             raise
+        except sqlalchemy.exc.IntegrityError as e:
+            session.rollback()
+            raise MlflowException(message=str(e), error_code=RESOURCE_ALREADY_EXISTS)
```
dbczumar, line 87: `RESOURCE_DOES_NOT_EXIST` may not be accurate for certain integrity errors, such as foreign key constraint violations. Can we use `BAD_REQUEST` instead?

```suggestion
            raise MlflowException(message=str(e), error_code=BAD_REQUEST)
```

WeichenXu123: I think we should use RESOURCE_ALREADY_EXISTS, because we do converting here:
https://github.com/mlflow/mlflow/blob/56a733af68fde4c454d7ba193e65dec61bac649f/mlflow/server/handlers.py#L447
and the converter map is:
https://github.com/mlflow/mlflow/blob/56a733af68fde4c454d7ba193e65dec61bac649f/mlflow/exceptions.py#L23

so, RESOURCE_ALREADY_EXISTS will be converted into "400" code when return error code in http responding result.

mlflow/store/db/utils.py
WeichenXu123: > Can we stringify message in the MlflowException constructor?
this sounds a little strange. The argument name is "message" which means it should be a text.

harupy: It does, but I think it's better than replace all `MlflowException(error_instance)` to `MlflowException(str(e))`.

harupy: If we used a type checker like mypy, we could prevent passing a non-string value to `MlflowException` but we currently don't use mypy.

mlflow/store/db/utils.py
harupy: https://github.com/mlflow/mlflow/blob/352d6ae9c9adeff59a2a12bc7a0b35f18a992af7/mlflow/exceptions.py#L25

(If I'm not mistaken) `BAD_REQUEST` also will be converted to 400.

WeichenXu123: oh. My link wrong. I am using "RESOURCE_ALREADY_EXISTS" in my code, not "RESOURCE_DOES_NOT_EXIST" .
"RESOURCE_ALREADY_EXISTS" should be more accurate than "BAD_REQUEST"

harupy:
```python
>>> MlflowException("foo", RESOURCE_ALREADY_EXISTS).get_http_status_code()
400
>>> MlflowException("foo", BAD_REQUEST).get_http_status_code()
400
```

harupy: > "RESOURCE_ALREADY_EXISTS" should be more accurate than "BAD_REQUEST"

Can you tell us why?

mlflow/exceptions.py
```diff
@@ -38,8 +38,8 @@ class MlflowException(Exception):
 
     def __init__(self, message, error_code=INTERNAL_ERROR, **kwargs):
```
harupy, line 39: Just a second thought, can we create a class method name `from_exception`?

harupy: Never mind. I think `str(message)` should suffice for now.

mlflow/store/db/utils.py
WeichenXu123: I rethink on this, `RESOURCE_ALREADY_EXISTS` might be confusing, the experiment is not a "resource", BAD_REQUEST code is better.

mlflow/exceptions.py
```diff
@@ -50,6 +50,10 @@ def __init__(self, message, error_code=INTERNAL_ERROR, **kwargs):
             self.error_code = ErrorCode.Name(error_code)
         except (ValueError, TypeError):
             self.error_code = ErrorCode.Name(INTERNAL_ERROR)
+        if isinstance(message, BaseException):
+            message = f"{message.__module__}.{message.__class__}: {str(message)}"
```
harupy, line 54: If we do this, I'd define `from_exception` class method.

tests/store/tracking/test_sqlalchemy_store.py
```diff
@@ -542,7 +542,8 @@ def test_run_needs_uuid(self):
                 run = models.SqlRun()
                 session.add(run)
                 warnings.resetwarnings()
-        assert exception_context.exception.error_code == ErrorCode.Name(INTERNAL_ERROR)
+        assert exception_context.exception.error_code == ErrorCode.Name(BAD_REQUEST), \
+            f"Wrong exception {exception_context.exception.message} raised."
```
harupy, line 546: why do we need `f"Wrong exception {exception_context.exception.message} raised."`? How does the pytest error report look like when this line fails?

mlflow/exceptions.py
WeichenXu123: Some tests failed in CI but passed in my local machine. I added this for debugging for now.

WeichenXu123:
I am thinking we should catch all `SQLAlchemyError` and mark it as "BAD_REQUEST".
Thoughts ? @harupy @dbczumar 

I checked the sub-classes of `SQLAlchemyError`, when these errors happens we should not retry.

harupy (approved):

PR #5717, opened 2022-04-18
Dev env setup script

BenWilson2:

Add a development environment setup script for Python dev 

## How is this patch tested?

Manually

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Add a Development Environment setup script for automated construction of a CI-friendly Python virtual environment.

Components 

Interface 

Language 

Integrations


dev/dev-env-setup.sh
```diff
@@ -0,0 +1,158 @@
...
+pyenv_exist=$(command -v pyenv)
+
+if [ -z "$pyenv_exist" ]; then
+  read -p "pyenv is required to be installed to manage python versions. Would you like to install it? $(tput bold)(y/n)$(tput sgr0): " -n 1 -r
+  echo
+  if [[ $REPLY =~ ^[Yy]$ ]]; then
+    echo "Updating brew and installing pyenv..."
+    echo "Note: this will probably take a considerable amount of time."
+    brew update
+    brew install pyenv
```
harupy, line 65: Does this script only support macOS?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,158 @@
...
+  else
+    exit 1
+  fi
+fi
+
+MLFLOW_HOME=$(pwd)
+
+# Get the minimum supported version from MLflow to ensure any feature development adheres to legacy Python versions
+min_py_version=$(grep "python_requires=" "$MLFLOW_HOME/setup.py" | grep -E -o "([0-9]{1,}\.)+[0-9]{1,}")
```
harupy, line 74: We might want to add a command to print out the minimum supported python version in `setup.py`.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,158 @@
...
+      - If replacing, delete old virtual environment.
+  - Create a virtual environment using the minimum required Python version based on previous step logic
+  - Activate the environment
+  - Install required dependencies for the dev envrionment
+
+  Example usage:
+
+  From root of MLflow repository on local with a destination virtualenv path of <root>/.venvs/mlflow-dev:
+
+  dev/dev-env-setup.sh -d ~/.venvs/mlflow-dev
```
harupy, line 24: I think the environment location should be `{path_to_mlflow_repository}/.venv` to make it easier to remove it.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,158 @@
...
+echo "Installing pip dependencies for development environment."
+# Install required dependencies for Prophet
+tmp_dir=$(mktemp -d)
+pip download "$( (( $quiet == 1 && $verbose == 0 )) && printf %s '-q' )" --no-deps --dest "$tmp_dir" --no-cache-dir prophet
+tar -zxvf "$tmp_dir"/*.tar.gz -C "$tmp_dir"
+pip install "$( (( $quiet == 1 && $verbose == 0 )) && printf %s '-q' )" -r "$(find "$tmp_dir" -name requirements.txt)"
+rm -rf "$tmp_dir"
+
+# Install dev requirements and test plugin
+pip install "$( (( $quiet == 1 && $verbose == 0 )) && printf %s '-q' )" -r "$MLFLOW_HOME/requirements/dev-requirements.txt"
```
harupy, line 123: Core maintainers like us need all the extra dependencies, but most contributors don't, and installing all of them is difficult to succeed. Can we install a minimum set of dev dependencies by default?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env bash
```
harupy, line 1: Can we run this scrip somewhere to ensure it works?

dev/dev-env-setup.sh
BenWilson2: good idea! 
changed to 
`dev/dev-env-setup.sh -d $(pwd)/.venvs/mlflow-dev`

dev/dev-env-setup.sh
BenWilson2: great point. I'll add the pyenv setup for a few OS's.

dev/dev-env-setup.sh
BenWilson2: added!

dev/dev-env-setup.sh
BenWilson2: Should we make the [extras] optional? Or:

Options:
--packages 
"light" - skinny, small, lint
"full" - light + large, doc, extra-ml
"all" - dev-requirements + [extras]

WDYT?

dev/dev-env-setup.sh
harupy: Can we just provide the pyenv installation instructions and let contributors install pyenv by themselves?

harupy: I'm concerned about maintenance costs. If we support a few OS, I think we need to test the script on the supported operating systems.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+    read -p "Enter your email address for your commit signature: " -r git_user_email
+    git config --global user.email "$git_user_email"
+    echo "Git user email set as: $(git config user.email)"
+  else
+    echo "Failing to set git 'user.name' and 'user.email' will result in unsigned commits. Ensure that you sign commits manually for CI checks to pass."
+  fi
+fi
+
+# setup pre-commit hooks
+git config core.hooksPath "$MLFLOW_HOME/hooks"
```
harupy, line 228: Does this line work on windows?

harupy, line 228: Does this line work on windows?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+  rm -rf "$tmp_dir"
+  # Install dev requirements and test plugin
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -r "$MLFLOW_HOME/requirements/dev-requirements.txt"
+  # Install current checked out version of MLflow (local)
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -e .[extras]
+  # Install test plugin
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -e "$MLFLOW_HOME/tests/resources//mlflow-test-plugin"
+  echo "Finished installing pip dependencies."
+else
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -r "$MLFLOW_HOME/requirements/skinny-requirements.txt"
```
harupy, line 198: I think we can remove this line because `skinny-requirements.txt` is a subset of `small-requirements.txt`.

harupy, line 198: I think we can remove this line because `skinny-requirements.txt` is a subset of `small-requirements.txt`.

.github/workflows/dev-setup.yml
```diff
@@ -0,0 +1,90 @@
+name: Dev environment setup
```
harupy, line 1: Do we need to run this workflow in any PRs? or only when we change `dev/dev-env-setup.sh`?

harupy, line 1: Do we need to run this workflow in any PRs? or only when we change `dev/dev-env-setup.sh`?

.github/workflows/dev-setup.yml
```diff
@@ -0,0 +1,90 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash --noprofile --norc -exo pipefail {0}
+
+jobs:
```
harupy, line 30: Let's use `strategy.matrix` to DRY the code: https://github.com/mlflow/mlflow/blob/8b0bfc5e9ce46545d452181dc181896043eb8088/.github/workflows/js.yml#L23

.github/workflows/dev-setup.yml
```diff
@@ -0,0 +1,90 @@
...
+          git config --global user.email "test@mlflow.org"
+      - name: Install development environment
+        run: |
+          ./dev/dev-env-setup.sh -q -d ./.venvs/mlflow-dev
+      - name: Run Environment tests
+        run: |
+          ./dev/test-dev-env-setup.sh
+
+
+
```
harupy, line 90:
```suggestion
```

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+if [[ -d "$directory"  ]]; then
+  if [ -z "$MLFLOW_DEV_ENV_REPLACE_ENV" ]; then
+    read -p "A virtual environment is already located at $directory. Do you wish to replace it? $(tput bold; tput setaf 2)(y/n) $(tput sgr0)" -n 1 -r
+    echo
+  fi
+  if [[ $REPLY =~ ^[Yy]$ || $MLFLOW_DEV_ENV_REPLACE_ENV == 1 ]]; then
+    deactivate
+    rm -rf "$directory"
+    echo "Virtual environment removed from '$directory'. Installing new instance."
+    pyenv exec virtualenv "$directory"
```
harupy, line 167: Does virtualenv automatically pick up the current python version?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+# Check if the virtualenv already exists at the specified path
+if [[ -d "$directory"  ]]; then
+  if [ -z "$MLFLOW_DEV_ENV_REPLACE_ENV" ]; then
+    read -p "A virtual environment is already located at $directory. Do you wish to replace it? $(tput bold; tput setaf 2)(y/n) $(tput sgr0)" -n 1 -r
+    echo
+  fi
+  if [[ $REPLY =~ ^[Yy]$ || $MLFLOW_DEV_ENV_REPLACE_ENV == 1 ]]; then
+    deactivate
+    rm -rf "$directory"
+    echo "Virtual environment removed from '$directory'. Installing new instance."
```
harupy, line 166: Can we use could use `--clear` flag:

```
$ virtualenv --help
...
  --clear                       remove the destination directory if exist before starting (will overwrite files otherwise) (default: False)
```

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+      sudo apt-get update -y
+      sudo apt-get install -y make build-essential libssl-dev zlib1g-dev \
+      libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm \
+      libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev
+      # Install pyenv from source
+      git clone --depth 1 https://github.com/pyenv/pyenv.git "$HOME/.pyenv"
+      PYENV_ROOT="$HOME/.pyenv"
+      PYENV_BIN="$PYENV_ROOT/bin"
+      echo "$PYENV_BIN" >> "$GITHUB_PATH"
+      echo "PYENV_ROOT=$PYENV_ROOT" >> "$GITHUB_ENV"
```
harupy, line 102: I think these two lines should be executed only in GitHub Actions workflows.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+
+echo "$(tput setaf 2)Current Python version: $(tput bold)$(python --version)$(tput sgr0)"
+echo "$(tput setaf 3)Activated environment is located: $(tput bold) $directory/bin/activate$(tput sgr0)"
+
+echo "Installing pip dependencies for development environment."
+
+if [[ $full_install == 1 ]]; then
+  # Install required dependencies for Prophet
+  tmp_dir=$(mktemp -d)
+  pip download $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) --no-deps --dest "$tmp_dir" --no-cache-dir prophet
```
harupy, line 186: Can we put `(( quiet == 1 && verbose == 0 )) && printf %s '-q'` in a variable and reuse it?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+  else
+    echo "Failing to set git 'user.name' and 'user.email' will result in unsigned commits. Ensure that you sign commits manually for CI checks to pass."
+  fi
+fi
+
+# setup pre-commit hooks
+git config core.hooksPath "$MLFLOW_HOME/hooks"
+
+# Install pytest
+pip install $( (( $quiet == 1 && $verbose == 0 )) && printf %s '-q' ) pytest
```
harupy, line 231: I think pytest should be already be installed at this point.

dev/dev-env-setup.sh
harupy: Sorry I wasn't clear. I meant a command like `ListDependencies`:

https://github.com/mlflow/mlflow/blob/f7a42be1e29ec1ff1f6086cdb79d75248f877715/setup.py#L90

.github/workflows/dev-setup.yml
BenWilson2: I think we'd likely be running this on a weekly schedule. I really don't want to run this on PRs or during nightly builds since it's going to impact our PR feedback heavily for something that really shouldn't be changing very often at all.

.github/workflows/dev-setup.yml
BenWilson2: great idea!

.github/workflows/dev-setup.yml
harupy: I think we can do something similar to the cross-version test. Run this job weekly and when we update `dev/dev-env-setup.sh`.

harupy:
@BenWilson2 https://github.com/mlflow/mlflow/pull/5748 has been merged. Now we can get the minimum python version mlflow supports with `python setup.py -q min_python_version`.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
...
+  echo "Finished installing pip dependencies."
+else
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -r "$MLFLOW_HOME/requirements/skinny-requirements.txt"
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -r "$MLFLOW_HOME/requirements/small-requirements.txt"
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -r "$MLFLOW_HOME/requirements/lint-requirements.txt"
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -r "$MLFLOW_HOME/requirements/large-requirements.txt"
+  pip install $( (( quiet == 1 && verbose == 0 )) && printf %s '-q' ) -r "$MLFLOW_HOME/requirements/doc-requirements.txt"
+fi
+
+command -v docker >/dev/null 2>&1 || echo "$(tput bold; tput setaf 1)A docker installation cannot be found. Please ensure that docker is installed to run all tests locally.$(tput sgr0)"
```
WeichenXu123, line 205: Can we also support installing docker in the script ?

WeichenXu123: We also have some tests rely on kubernetes. Shall we also support installing and set-up kubernetes in the script ?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env bash
```
WeichenXu123, line 1: Does this shebang `#!/usr/bin/env bash` support windows system ?
Looks like windows path is something like `C:\....`

WeichenXu123: Oh maybe I am wrong, I think the `bash` part means the one used in windows ？

dev/dev-env-setup.sh
WeichenXu123: It is fine to start with support docker/kubernetes installation only in mac.

WeichenXu123:
and shall we support installing "node.js" in the env setup script ?
node.js is used in mflow web server.

WeichenXu123:
sorry click closing button unexpected.

dev/dev-env-setup.sh
harupy: I don't think we need to setup docket or kubenetes because most of contributors don't need to touch them.

harupy: The more things we try to setup, the harder it is for this script to succeed.

dev/dev-env-setup.sh
BenWilson2: it goes through a discovery process that, if the version is found on the local system, will by default create a symlink. Do you think we should disable this and do a copy mode to force a replica?

dev/dev-env-setup.sh
BenWilson2: good idea!

dev/dev-env-setup.sh
BenWilson2: added (but not pushed yet)

.github/workflows/dev-setup.yml
```diff
@@ -0,0 +1,60 @@
+name: Dev environment setup
+
+on:
+  schedule:
```
harupy, line 4: I think we should run this workflow in PRs that change `dev/dev-env-setup.sh`.

.github/workflows/dev-setup.yml
```diff
@@ -0,0 +1,60 @@
...
+    shell: bash --noprofile --norc -exo pipefail {0}
+
+jobs:
+  env_setup:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - os: ubuntu-latest
+            shell: bash
```
harupy, line 37: I think we can remove this because we set the workflow-level default shell.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,254 @@
...
+    elif [[ "$machine" == linux ]]; then
+      sudo apt-get update -y
+      sudo apt-get install -y make build-essential libssl-dev zlib1g-dev \
+      libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm \
+      libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev
+      # Install pyenv from source
+      git clone --depth 1 https://github.com/pyenv/pyenv.git "$HOME/.pyenv"
+      PYENV_ROOT="$HOME/.pyenv"
+      PYENV_BIN="$PYENV_ROOT/bin"
+      if [ "$MLFLOW_DEV_ENV_CI_RUN" == 1 ]; then
```
harupy, line 105:
```suggestion
      if [ ! -z "$GITHUB_ACTIONS" ]; then
```

Can we use the `GITHUB_ACTIONS` environment variable instead?

harupy:
> and shall we support installing "node.js" in the env setup script ? node.js is used in mflow web server.

I'd say no because the number of contributions to MLflow UI is not large. We can this later if there is a request.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,254 @@
...
+    else
+      echo "The current OS is unknown. Please visit: https://github.com/pyenv/pyenv#installation for instructions. "
+      exit 1
+    fi
+    echo "pyenv is required to use this environment setup script. Please install by following instructions here: $PYENV_README"
+    exit 1
+  fi
+fi
+
+MLFLOW_HOME=$(pwd)
```
harupy, line 142:
```suggestion
REPO_ROOT=$(git rev-parse --show-toplevel)
```

might make more sense.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,254 @@
...
+    echo
+  fi
+  if [[ $REPLY =~ ^[Yy]$ || $MLFLOW_DEV_ENV_PYENV_INSTALL == 1 ]]; then
+    if [[ "$machine" == mac ]]; then
+      # Check if brew is installed and install it if it isn't present
+      # Note: if xcode isn't installed, this will fail.
+      if [ -z "$(command -v brew)" ]; then
+        echo "Brew is required to install pyenv on MacOS. Installing in your home directory."
+        wget -O ~/brew_install.sh https://raw.githubusercontent.com/Homebrew/install/master/install.sh
+        bash ~/brew_install.sh
```
harupy, line 90:
```suggestion
        bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```

to avoid leaving `brew_install.sh`.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,254 @@
...
+      fi
+      echo "Updating brew and installing pyenv..."
+      echo "Note: this will probably take a considerable amount of time."
+      brew update
+      brew install pyenv
```
harupy, line 95:
```suggestion
      brew update
      brew install pyenv
      brew install openssl readline sqlite3 xz zlib
```

https://github.com/pyenv/pyenv/wiki#suggested-build-environment

dev/dev-env-setup.sh
WeichenXu123: We can provide an option to install docker / kubernets and by default the option is off.
We can do it in follow-up PR.

harupy: Sounds good. Do you want to work on that?

WeichenXu123: I would help to work on that when I have free time.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,254 @@
...
+# Resolve a minor version to the latest micro version
+case $min_py_version in
+  "3.7") PY_INSTALL_VERSION="3.7.13" ;;
+  "3.8") PY_INSTALL_VERSION="3.8.13" ;;
+  "3.9") PY_INSTALL_VERSION="3.9.11" ;;
+  "3.10") PY_INSTALL_VERSION="3.10.3" ;;
+esac
+
+# Install the Python version if it cannot be found
+pyenv install -s $PY_INSTALL_VERSION
```
WeichenXu123, line 172: I suggest use `pyenv install -s -v $PY_INSTALL_VERSION` 
the `-v` make pyenv print compilation status to stdout which helps debugging  when error happens.

and shall we add a "pyenv_root_dir" argument to allow user to specify where pyenv download the python tarballs to ?

harupy: > the -v make pyenv print compilation status to stdout which helps debugging when error happens.

Have you tried running `pyenv install -v`? It's too verbose?

harupy: > shall we add a "pyenv_root_dir" argument to allow user to specify where pyenv download the python tarballs to ?

I don't see a need for that argument . In what situation would that argument be useful?

WeichenXu123: > Have you tried running pyenv install -v? It's too verbose
Yes it might be too verbose but it helps provides more clues when compiling failed. 

> I don't see a need for that argument . In what situation would that argument be useful?
E.g the case the system disk remaining capacity is not sufficient, we can specify the path to be on other disk. Just a minor suggestion.

harupy: See https://github.com/pyenv/pyenv/blob/fab0082bd5cdda07f0bfdd69a9c676bc2d2906b3/plugins/python-build/bin/python-build#L135. pyenv stores build logs in a temporary file.

harupy: > E.g the case the system disk remaining capacity is not sufficient, we can specify the path to be on other disk. Just a minor suggestion.

Can we just remove some files to free disk space?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,232 @@
...
+  - Check if the virtual environment already exists
+    - If it does, prompt for replacement
+      - If replacing, delete old virtual environment.
+  - Create a virtual environment using the minimum required Python version based on previous step logic
+  - Activate the environment
+  - Install required dependencies for the dev envrionment
+
+  Example usage:
+
+  From root of MLflow repository on local with a destination virtualenv path of <MLFLOW_HOME>/.venvs/mlflow-dev:
```
harupy, line 23:
```suggestion
  From root of MLflow repository on local with a destination virtualenv path of <MLFLOW_HOME>/.venv:
```

Can we use `.venv` instead of `.venvs/mlflow-dev-env`?

.github/workflows/dev-setup.yml
```diff
@@ -0,0 +1,45 @@
...
+          repository: ${{ github.event.inputs.repository }}
+          ref: ${{ github.event.inputs.ref }}
+      - uses: ./.github/actions/setup-pyenv
+      - name: Setup environment
+        run: |
+          git config --global user.name "test"
+          git config --global user.email "test@mlflow.org"
+      - name: Run Environment tests
+        run: |
+          TERM=xterm bash ./dev/test-dev-env-setup.sh
```
harupy, line 45: Is `TERM=xterm` only required on github?

dev/dev-env-setup.sh
harupy: I saw https://github.com/mlflow/mlflow/pull/5717/files#diff-76e7ca2f2843ac5f659b45382768c1b64bf8704e8d3c6189c4e3623f5d465bc2R162-R170. Now the current approach makes sense.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,232 @@
...
+  if [ -z "$GITHUB_ACTIONS" ]; then
+    read -p "pyenv is required to be installed to manage python versions. Would you like to install it? $(tput bold)(y/n)$(tput sgr0): " -n 1 -r
+    echo
+  fi
+  if [[ $REPLY =~ ^[Yy]$ || -n "$GITHUB_ACTIONS" ]]; then
+    if [[ "$machine" == mac ]]; then
+      # Check if brew is installed and install it if it isn't present
+      # Note: if xcode isn't installed, this will fail.
+      if [ -z "$(command -v brew)" ]; then
+        echo "Brew is required to install pyenv on MacOS. Installing in your home directory."
```
harupy, line 88:
```suggestion
        echo "Homebrew is required to install pyenv on MacOS. Installing in your home directory."
```

nit

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,232 @@
...
+      brew install openssl readline sqlite3 xz zlib
+    elif [[ "$machine" == linux ]]; then
+      sudo apt-get update -y
+      sudo apt-get install -y make build-essential libssl-dev zlib1g-dev \
+      libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm \
+      libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev
+      # Install pyenv from source
+      git clone --depth 1 https://github.com/pyenv/pyenv.git "$HOME/.pyenv"
+      PYENV_ROOT="$HOME/.pyenv"
+      PYENV_BIN="$PYENV_ROOT/bin"
```
harupy, line 104:
```suggestion
      PYENV_ROOT="$HOME/.pyenv"
      PYENV_BIN="$PYENV_ROOT/bin"
      PATH="$PYENV_BIN:$PATH"
```

I think we need to add `PYENV_BIN` to `PATH`.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,232 @@
...
+    echo "pyenv is required to use this environment setup script. Please install by following instructions here: $PYENV_README"
+    exit 1
+  fi
+fi
+
+MLFLOW_HOME=$(pwd)
+rd="$MLFLOW_HOME/requirements"
+
+# Get the minimum supported version from MLflow to ensure any feature development adheres to legacy Python versions
+min_py_version=$(grep "python_requires=" "$MLFLOW_HOME/setup.py" | grep -E -o "([0-9]{1,}\.)+[0-9]{1,}")
```
harupy, line 124:
```suggestion
min_py_version=$(python setup.py -q min_python_version)
```

Let's use the `min_python_version` command.

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,232 @@
...
+
+  Note: it is recommended to preface virtualenv locations with a directory name prefaced by '.' (i.e., ".venvs").
+
+  The default environment setup is for basic functionality, installing the minimum development requirements dependencies.
+  To install the full development environment that supports working on all flavors and running all tests locally, set
+  the flag '-f' or '--full'
+
+-h, -help,        --help        Display help
+
+-d, -directory    --directory   The path to install the virtual environment into
```
harupy, line 35: Does `./dev/dev-env-setup.sh -directory` or `./dev/dev-env-setup.sh --directory` work?

dev/test-dev-env-setup.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Test that the dev-env-setup.sh script installs appropriate versions
+set -x
+
+err=0
+trap 'err=1' ERR
```
harupy, line 7: Can we fail-fast?

.github/workflows/dev-setup.yml
BenWilson2: added

dev/dev-env-setup.sh
BenWilson2: thanks! Added!

.github/workflows/dev-setup.yml
BenWilson2: yes. In a local terminal client there will be some defined terminal value to allow for syntactic formatting. I took this tip directly from a Github dev on an issue that they responded on that recommended specifying this if using any sort of tput commands within the script.

dev/dev-env-setup.sh
BenWilson2: updated!

dev/dev-env-setup.sh
BenWilson2: it does now!

harupy: btw do we need `-directory`?

dev/dev-env-setup.sh
```diff
@@ -0,0 +1,268 @@
...
+  "3.7") PY_INSTALL_VERSION="3.7.13" ;;
+  "3.8") PY_INSTALL_VERSION="3.8.13" ;;
+  "3.9") PY_INSTALL_VERSION="3.9.11" ;;
+  "3.10") PY_INSTALL_VERSION="3.10.3" ;;
+esac
+
+microver=$(grep -o '\.' <<< "$override_py_ver" | wc -l)
+if [[ $microver -gt 1 ]]; then
+  PY_INSTALL_VERSION=$override_py_ver
+fi
```
harupy, line 169: The logic here is a bit difficult to understand to me:

```sh
minor_to_micro() {
  case $min_py_version in
    "3.7") echo "3.7.13" ;;
    "3.8") echo "3.8.13" ;;
    "3.9") echo "3.9.11" ;;
    "3.10") echo"3.10.3" ;;
  esac
}

if override_py_ver is specified
  if override_py_ver looks like minor version
      PY_INSTALL_VERSION=$(minor_to_micro $override_py_ver)
  elif override_py_ver looks like micro version
      PY_INSTALL_VERSION=$override_py_ver
  else
      # Invalid version
      exit 1
  fi
else
  PY_INSTALL_VERSION=$(minor_to_micro $min_py_version)
fi
```

.github/workflows/dev-setup.yml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  linux-env-setup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: ${{ github.event.inputs.repository }}
+          ref: ${{ github.event.inputs.ref }}
+      - uses: ./.github/actions/setup-pyenv
```
harupy, line 43:
```suggestion
```

Can we remove this line and test pyenv installation?

CONTRIBUTING.rst
```diff
@@ -114,8 +114,66 @@ The majority of the MLflow codebase is developed in Python. This includes the CL
...
+.. code-block:: bash
+
+    dev/dev-env-setup.sh -h
+
+An example usage of this script that will build a development environment using ``virtualenv`` and the minimum supported
+Python version (to ensure compatibility) is:
+
+.. code-block:: bash
+
+    dev/dev-env-setup.sh -d ~/.venvs/mlflow-dev -q
```
harupy, line 145:
```suggestion
    dev/dev-env-setup.sh -d .venvs/mlflow-dev -q
```

Can we create an environment in the current working directory?

harupy (approved):
LGTM once the remaining comments are addressed!

dev/dev-env-setup.sh
BenWilson2: great idea! This logic is much easier to follow.

.github/workflows/dev-setup.yml
BenWilson2: definitely! Removing and validating CI build.

dev/dev-env-setup.sh
BenWilson2: removed those redundant arg keys

PR #5719, opened 2022-04-19
Autologging of input example and signature for pyspark ml

bali0019:

Autologging of input example/signature for pyspark ml pipelines and estimators.
Unit tests to validate the functionality.

## How is this patch tested?

Unit tests

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Adds autologging of input example and signature for pyspark ml pipeline/estimators

Components 

Interface 

Language 

Integrations

### DIFF
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -938,46 +938,33 @@ def autolog(
 
         if log_models:
             if _should_log_model(spark_model):
-                # TODO: support model signature
-                from mlflow.models.signature import infer_signature
+                from mlflow.models import infer_signature
+                from mlflow.pyspark.ml._autolog import (
+                    cast_spark_df_with_vector_to_array,
+                    get_feature_cols,
+                )
+                from mlflow.spark import find_and_set_features_col_as_vector_if_needed
+                from pyspark.sql import SparkSession
 
-                def _cast_spark_df_with_vector_to_array(input_spark_df):
-                    """
-                    Finds columns of vector type in a spark dataframe and
-                    casts them to array<double> type.
+                spark = SparkSession.builder.getOrCreate()
 
-                    :param input_spark_df:
-                    :return: a spark dataframe with vector columns transformed to array<double> type
-                    """
-                    from functools import reduce
-                    from pyspark.ml.functions import vector_to_array
-                    from pyspark.ml.linalg import VectorUDT
-
-                    vector_type_columns = [
-                        _field.name
-                        for _field in input_spark_df.schema
-                        if isinstance(_field.dataType, VectorUDT)
-                    ]
-                    return reduce(
-                        lambda df, vector_col: df.withColumn(
-                            vector_col, vector_to_array(vector_col)
-                        ),
-                        vector_type_columns,
-                        input_spark_df,
+                def _get_input_example_as_pd_df():
+                    feature_cols = list(get_feature_cols(input_df.schema, spark_model))
+                    limited_input_df = input_df.select(feature_cols).limit(
+                        INPUT_EXAMPLE_SAMPLE_ROWS
                     )
+                    return cast_spark_df_with_vector_to_array(limited_input_df).toPandas()
 
-                limited_input_df = input_df.limit(INPUT_EXAMPLE_SAMPLE_ROWS)
-
-                def _get_input_example():
-                    return _cast_spark_df_with_vector_to_array(limited_input_df).toPandas()
-
-                def _infer_model_signature(input_data_slice):
-                    model_output = spark_model.transform(input_data_slice)
-                    return infer_signature(input_data_slice.toPandas(), model_output.toPandas())
+                def _infer_model_signature(input_example_slice):
+                    input_slice_df = find_and_set_features_col_as_vector_if_needed(
+                        spark.createDataFrame(input_example_slice), spark_model
+                    )
+                    model_output = spark_model.transform(input_slice_df)
+                    return infer_signature(input_example_slice, model_output.toPandas())
 
                 input_example, signature = resolve_input_example_and_signature(
-                    _get_input_example,
-                    lambda _: _infer_model_signature(limited_input_df),
+                    _get_input_example_as_pd_df,
+                    _infer_model_signature,
                     log_input_examples,
                     log_model_signatures,
                     _logger,
@@ -1013,10 +1000,14 @@ def autolog(
         else:
             # we need generate estimator param map so we call `self.copy(params)` to construct
             # an estimator with the extra params.
+            from pyspark.storagelevel import StorageLevel
+
             estimator = self.copy(params) if params is not None else self
             _log_pretraining_metadata(estimator, params)
+            input_training_df = args[0].persist(StorageLevel.MEMORY_AND_DISK)
             spark_model = original(self, *args, **kwargs)
-            _log_posttraining_metadata(estimator, spark_model, params, args[0])
+            _log_posttraining_metadata(estimator, spark_model, params, input_training_df)
+            input_training_df.unpersist()
 
             return spark_model
 
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -943,7 +943,7 @@ def autolog(
                     cast_spark_df_with_vector_to_array,
                     get_feature_cols,
                 )
-                from mlflow.spark import find_and_set_features_col_as_vector_if_needed
+                from mlflow.spark import _find_and_set_features_col_as_vector_if_needed
                 from pyspark.sql import SparkSession
 
                 spark = SparkSession.builder.getOrCreate()
@@ -956,7 +956,7 @@ def autolog(
                     return cast_spark_df_with_vector_to_array(limited_input_df).toPandas()
 
                 def _infer_model_signature(input_example_slice):
-                    input_slice_df = find_and_set_features_col_as_vector_if_needed(
+                    input_slice_df = _find_and_set_features_col_as_vector_if_needed(
                         spark.createDataFrame(input_example_slice), spark_model
                     )
                     model_output = spark_model.transform(input_slice_df).drop(
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -835,14 +835,8 @@ def autolog(
     :param log_model_signatures: If ``True``,
                                  :py:class:`ModelSignatures <mlflow.models.ModelSignature>`
                                  describing model inputs and outputs are collected and logged along
-                                 with tf/keras model artifacts during training. If ``False``,
-                                 signatures are not logged. ``False`` by default because
-                                 logging Spark ML models with signatures changes their pyfunc
-                                 inference behavior when Spark DataFrames are passed to
-                                 ``predict()``: when a signature is present, an ``np.ndarray``
-                                 (for single-output models) or a mapping from
-                                 ``str`` -> ``np.ndarray`` (for multi-output models) is returned;
-                                 when a signature is not present, a Spark DataFrame is returned.
+                                 with spark ml pipeline/estimator artifacts during training. If ``False``,
+                                 signatures are not logged.
     **The default log model allowlist in mlflow**
         .. literalinclude:: ../../../mlflow/pyspark/ml/log_model_allowlist.txt
            :language: text
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -959,7 +959,9 @@ def autolog(
                     input_slice_df = find_and_set_features_col_as_vector_if_needed(
                         spark.createDataFrame(input_example_slice), spark_model
                     )
-                    model_output = spark_model.transform(input_slice_df)
+                    model_output = spark_model.transform(input_slice_df).drop(
+                        *input_slice_df.columns
+                    )
                     return infer_signature(input_example_slice, model_output.toPandas())
 
                 input_example, signature = resolve_input_example_and_signature(
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -835,8 +835,9 @@ def autolog(
     :param log_model_signatures: If ``True``,
                                  :py:class:`ModelSignatures <mlflow.models.ModelSignature>`
                                  describing model inputs and outputs are collected and logged along
-                                 with spark ml pipeline/estimator artifacts during training. If ``False``,
-                                 signatures are not logged.
+                                 with spark ml pipeline/estimator artifacts during training.
+                                 If ``False`` signatures are not logged.
+
     **The default log model allowlist in mlflow**
         .. literalinclude:: ../../../mlflow/pyspark/ml/log_model_allowlist.txt
            :language: text
--- a/mlflow/pyspark/ml/__init__.py
+++ b/mlflow/pyspark/ml/__init__.py
@@ -18,6 +18,7 @@ from mlflow.utils.autologging_utils import (
     _get_new_training_session_class,
     autologging_integration,
     safe_patch,
+    resolve_input_example_and_signature,
 )
 from mlflow.utils.autologging_utils import get_method_call_arg_value
 from mlflow.utils.file_utils import TempDir
@@ -26,6 +27,9 @@ from mlflow.utils.validation import (
     MAX_PARAMS_TAGS_PER_BATCH,
     MAX_PARAM_VAL_LENGTH,
     MAX_ENTITY_KEY_LENGTH,
+)
+from mlflow.utils.autologging_utils import (
+    INPUT_EXAMPLE_SAMPLE_ROWS,
 )
 
 _logger = logging.getLogger(__name__)
@@ -706,6 +710,8 @@ def autolog(
     silent=False,
     log_post_training_metrics=True,
     registered_model_name=None,
+    log_input_examples=False,
+    log_model_signatures=True,
 ):  # pylint: disable=unused-argument
     """
     Enables (or disables) and configures autologging for pyspark ml estimators.
@@ -823,7 +829,20 @@ def autolog(
     :param registered_model_name: If given, each time a model is trained, it is registered as a
                                   new model version of the registered model with this name.
                                   The registered model is created if it does not already exist.
-
+    :param log_input_examples: If ``True``, input examples from training datasets are collected and
+                               logged along with pyspark ml model artifacts during training. If
+                               ``False``, input examples are not logged.
+    :param log_model_signatures: If ``True``,
+                                 :py:class:`ModelSignatures <mlflow.models.ModelSignature>`
+                                 describing model inputs and outputs are collected and logged along
+                                 with tf/keras model artifacts during training. If ``False``,
+                                 signatures are not logged. ``False`` by default because
+                                 logging Spark ML models with signatures changes their pyfunc
+                                 inference behavior when Spark DataFrames are passed to
+                                 ``predict()``: when a signature is present, an ``np.ndarray``
+                                 (for single-output models) or a mapping from
+                                 ``str`` -> ``np.ndarray`` (for multi-output models) is returned;
+                                 when a signature is not present, a Spark DataFrame is returned.
     **The default log model allowlist in mlflow**
         .. literalinclude:: ../../../mlflow/pyspark/ml/log_model_allowlist.txt
            :language: text
@@ -876,7 +895,7 @@ def autolog(
 
         mlflow.set_tags(_get_estimator_info_tags(estimator))
 
-    def _log_posttraining_metadata(estimator, spark_model, params):
+    def _log_posttraining_metadata(estimator, spark_model, params, input_df):
 
         if _is_parameter_search_estimator(estimator):
             try:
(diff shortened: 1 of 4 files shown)


mlflow/pyspark/ml/__init__.py
```diff
@@ -925,8 +939,56 @@ def _log_posttraining_metadata(estimator, spark_model, params):
...
+                    ]
+                    return reduce(
+                        lambda df, vector_col: df.withColumn(
+                            vector_col, vector_to_array(vector_col)
+                        ),
+                        vector_type_columns,
+                        input_spark_df,
+                    )
+
+                limited_input_df = input_df.limit(INPUT_EXAMPLE_SAMPLE_ROWS)
```
BenWilson2, line 969: What if the input data already contains a Sparse or Dense Vector? Should there be an optimistic cast to capture that signature?

mlflow/spark.py
```diff
@@ -760,11 +760,34 @@ def predict(self, pandas_df):
...
+                ml_model_estimator.featuresCol
+            )
+            features_col_type = [
+                g
+                for g in spark_df.schema.fields
+                if g.name == features_col_name and g.dataType == t.ArrayType(t.DoubleType())
+            ]
+            if len(features_col_type) == 1:
+                spark_df = spark_df.withColumn(
+                    features_col_name, array_to_vector(f.col(features_col_name))
```
BenWilson2, line 786: Does array_to_vector do a functionally safe cast to DenseVector?

tests/spark/autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -876,3 +897,95 @@ def test_autolog_registering_model(spark_session, dataset_binomial):
...
         registered_model = MlflowClient().get_registered_model(registered_model_name)
         assert registered_model.name == registered_model_name
+
+
+def _assert_autolog_infers_model_signature_correctly(run, input_sig_spec, output_sig_spec):
+    artifacts_dir = run.info.artifact_uri.replace("file://", "")
+    client = mlflow.tracking.MlflowClient()
+    artifacts = [x.path for x in client.list_artifacts(run.info.run_id, "model")]
+    ml_model_filename = "MLmodel"
+    assert str(os.path.join("model", ml_model_filename)) in artifacts
```
BenWilson2, line 907: switch to pathlib syntax

tests/spark/autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -276,7 +279,29 @@ def test_fit_with_a_list_of_params(dataset_binomial):
             mock_set_tags.assert_not_called()
 
 
-def test_should_log_model(dataset_binomial, dataset_multinomial, dataset_text):
+@pytest.fixture()
```
BenWilson2, line 282: Should we test for the StringIndexer -> IndexToString for multiclass classification use cases?
https://spark.apache.org/docs/latest/ml-features.html#indextostring 
or is that going to complicate things?

mlflow/spark.py
BenWilson2: confirmed. Always casts.

mlflow/pyspark/ml/__init__.py
```diff
@@ -925,8 +939,56 @@ def _log_posttraining_metadata(estimator, spark_model, params):
...
                 # TODO: support model signature
+                from mlflow.models.signature import infer_signature
+
+                def _cast_spark_df_with_vector_to_array(input_spark_df):
+                    """
+                    Finds columns of vector type in a spark dataframe and
+                    casts them to array<double> type.
+
+                    :param input_spark_df:
+                    :return: a spark dataframe with vector columns transformed to array<double> type
```
dbczumar, line 950: @bali0019 @BenWilson2 Are columns of type `VectorUDT` the only non-scalar columns we have to special case? What about other `ArrayType` columns, e.g. string arrays?

mlflow/pyspark/ml/__init__.py
WeichenXu123, line 969: My concern is the collecting example rows will trigger one or more spark job.
Suppose the spark dataframe haven't been cached and the spark dataframe contains a long lineage, then it might generate a bunch of spark jobs.
and when we do training, these spark jobs might run again.

mlflow/pyspark/ml/__init__.py
WeichenXu123: Other types (primitive type or complex type constructed from primitive type) can convert to pandas dataframe directly. I don't think they need special handling.

mlflow/pyspark/ml/__init__.py
```diff
@@ -925,8 +939,56 @@ def _log_posttraining_metadata(estimator, spark_model, params):
...
+                            vector_col, vector_to_array(vector_col)
+                        ),
+                        vector_type_columns,
+                        input_spark_df,
+                    )
+
+                limited_input_df = input_df.limit(INPUT_EXAMPLE_SAMPLE_ROWS)
+
+                def _get_input_example():
+                    return _cast_spark_df_with_vector_to_array(limited_input_df).toPandas()
```
WeichenXu123, line 972: Note spark dataframe might includes some column which is unused in spark model training. These columns we are hard to exclude them, no simple API can filter them out.

mlflow/pyspark/ml/__init__.py
WeichenXu123: so shall we make the signature logging enabled by default ?

mlflow/pyspark/ml/__init__.py
bali0019: Just to confirm, I am going to add a few more unit tests that will cover all DataTypes in spark to make sure auto-logging doesn't fail for some types.

mlflow/pyspark/ml/__init__.py
```diff
@@ -925,8 +939,56 @@ def _log_posttraining_metadata(estimator, spark_model, params):
...
+
+                def _infer_model_signature(input_data_slice):
+                    model_output = spark_model.transform(input_data_slice)
+                    return infer_signature(input_data_slice.toPandas(), model_output.toPandas())
```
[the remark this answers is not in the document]
BenWilson2: Would a cast to `set` and a documentation callout about the signature for Spark models work?

mlflow/pyspark/ml/__init__.py
bali0019: @WeichenXu123, that is a great point! I noticed many of the spark estimators handle persistence internally by persisting before the fit and unpersisting afterward. Is it OK if we do it upfront (`input_df.persist(StorageLevel.MEMORY_AND_DISK)`) before the original fit is called and then unpersist it after autologging ?

mlflow/pyspark/ml/__init__.py
bali0019: @WeichenXu123 Another great point :). I had to do something similar in the past for automl-toolkit ( cc @BenWilson2) in the past.
To get the true input columns for the training pipeline, we can just step through each stage and extract inputCol. And then do a select on the input_df to get all those columns needed for all the stages in a pipeline. 
But if it is just an estimator (or a pipeline with a single stage), then we can just extract the single features column. 
what do you think of that approach?

mlflow/pyspark/ml/__init__.py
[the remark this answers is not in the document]
bali0019: that makes sense, please see my above comment [here](https://github.com/mlflow/mlflow/pull/5719#discussion_r853107051). thoughts?

tests/spark/autologging/ml/test_pyspark_ml_autologging.py
bali0019: Good point, working on it.

mlflow/pyspark/ml/__init__.py
WeichenXu123: > handle persistence internally by persisting before the fit and unpersisting afterward

This sounds a good idea.

mlflow/pyspark/ml/__init__.py
WeichenXu123: > it is just an estimator (or a pipeline with a single stage), then we can just extract the single features column.

Seems some model input is not features column, e.g. the ALSModel

> we can just step through each stage and extract inputCol

Some pipeline , follow-on stages might take previous stage output column as input column, we need special handling.

And for some meta-model, e.g. OneVsRestModel, its input column depends on its nested classification model input.

WeichenXu123: I think a safer way is, for pyspark Transformer/Model type, add APIs "getInputColumns" and "getOutputColumns", and implement the API for each specific model. Note many models inherits Predictor Model / Classification Model, which we can implement "getInputColumns" and "getOutputColumns" in their base class.

bali0019: thanks, @WeichenXu123 !
What are your thoughts about this approach (a quick dirty script to showcase the idea):

`
from pyspark.sql import SparkSession, types as t
from pyspark.ml.feature import IndexToString, StringIndexer, VectorAssembler
from pyspark.ml import Pipeline
from pyspark.ml.classification import LogisticRegression
import re
from pyspark.sql.utils import IllegalArgumentException

spark = SparkSession.builder.getOrCreate()

training_df = spark.createDataFrame(
    [
        (0, "a", "a", "b"),
        (1, "b", "a", "b"),
        (2, "c", "a", "b"),
        (3, "a", "a", "b"),
        (4, "a", "a", "b"),
        (5, "c", "a", "b"),
    ]
).toDF("id", "category", "not_a_feature_1", "not_a_feature_2")

string_indexer = StringIndexer(inputCol="category", outputCol="label").fit(training_df)
pipeline = Pipeline(
    stages=[
        string_indexer,
        VectorAssembler(inputCols=["id"], outputCol="features"),
        LogisticRegression(),
        IndexToString(
            inputCol="prediction", outputCol="originalLabel", labels=string_indexer.labels
        ),
    ]
)

pipeline_model = pipeline.fit(training_df)
params = pipeline_model


def do_pipeline_transform(spark_df):
    return pipeline_model.transform(spark_df)


def get_struct_type_by_cols(cols):
    if len(cols) > 0:
        return t.StructType([_field for _field in training_df.schema.fields if _field.name in cols])
    return []


def get_feature_cols(input_fields=None):
    if input_fields is None:
        input_fields = set({})
        df = spark.createDataFrame([()])
    else:
        df = spark.createDataFrame(
            spark.sparkContext.emptyRDD(), get_struct_type_by_cols(input_fields)
        )
    try:
        do_pipeline_transform(df)
    except IllegalArgumentException as iae:
        col_name_search = re.search("(.*) does not exist.", iae.desc, re.IGNORECASE)
        if col_name_search:
            col_name = col_name_search.group(1)
            input_fields.add(col_name)
            get_feature_cols(input_fields)
    return input_fields


required_cols = get_feature_cols()

print(
    t.StructType(
        [_field for _field in training_df.schema.fields if _field.name in required_cols]
    ).json()
)
`
Output:
`
{"fields":[{"metadata":{},"name":"id","nullable":true,"type":"long"}],"type":"struct"}
`

We can add unit tests around that message that spark pipeline/transformer internals throw to make sure it doesn't change or keep up with it. 
Thoughts?

dbczumar: Very interesting idea! Are there any significant performance implications? One drawback is that this doesn't solve the "useless columns" issue identified by @WeichenXu123: just because `do_pipeline_transform()` fails for a set of input columns doesn't mean that all of those input columns are required; it just means that one or more required columns, which may be explored later, are missing from the set.

dbczumar: > I think a safer way is, for pyspark Transformer/Model type, add APIs "getInputColumns" and "getOutputColumns", and implement the API for each specific model. Note many models inherits Predictor Model / Classification Model, which we can implement "getInputColumns" and "getOutputColumns" in their base class.

@WeichenXu123 Is the implementation going to be standard for `Predictor Model` / `Classification Model`? If so, can we implement the same logic for these models in MLflow right now and contribute it to PySpark ML after? It's okay to begin with a limited set of models.

bali0019: @dbczumar 
that last print statement above does that check to filter out columns that aren't needed for any of the pipelines' stages. I just edited that line above to make it part of required_cols. thoughts?

bali0019: Also, I am doing some local testing to see if there could be performance issues with this approach, which is one of the reasons I am using an empty DataFrame in the example above with the right schema,  but the way all the internal spark transformers are written, they [validate schema](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/Pipeline.scala#L134) before any fit/transform:

bali0019: Option 2 is to do something like this:

```
from pyspark.sql import SparkSession, types as t
from pyspark.ml.feature import IndexToString, StringIndexer, VectorAssembler
from pyspark.ml import Pipeline
from pyspark.ml.classification import LogisticRegression
spark = SparkSession.builder.getOrCreate()

training_df = spark.createDataFrame(
    [
        (0, "a", "b", "c", "d", 1.0),
        (1, "a", "b", "c", "d", 0.0),
        (2, "a", "b", "c", "d", 2.0),
        (3, "a", "b", "c", "d", 3.0),
        (4, "a", "b", "c", "d", 1.0),
        (5, "a", "b", "c", "d", 1.0),
    ]
).toDF("id", "feature_a", "feature_b", "not_a_feature_1", "not_a_feature_2", "label")

pipeline = Pipeline(
    stages=[
        StringIndexer(inputCols=["feature_a", "feature_b"], outputCols=["feature_a_indexed", "feature_b_indexed"]),
        VectorAssembler(inputCols=["id", "feature_a_indexed", "feature_b_indexed"], outputCol="features"),
        LogisticRegression()
    ]
)

pipeline_model = pipeline.fit(training_df)

pipeline_input_cols = set({})
label_col = None

pipeline_model.transform(training_df.drop("categories")).show()

for stage in pipeline_model.stages:
    if stage.hasParam("inputCol"):
        pipeline_input_cols.add(stage.extractParamMap().get(stage.inputCol))
    if stage.hasParam("inputCols"):
        input_cols = stage.extractParamMap().get(stage.inputCols)
        if input_cols is not None:
            pipeline_input_cols.update(input_cols)
    if stage.hasParam("featuresCol"):
        pipeline_input_cols.add(stage.extractParamMap().get(stage.featuresCol))
    if stage.hasParam("labelCol"):
        pipeline_input_cols.add(stage.extractParamMap().get(stage.featuresCol))

required_cols = set(training_df.columns).intersection(pipeline_input_cols)
print(required_cols)
```
output:
```
{'feature_a', 'feature_b', 'id'}
```

Checking [here](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/param/shared/sharedParams.scala) in the shared params, we could have: `inputCol`, `inputCols`, `labelCol`, `outputCol`, `outputCols`, `featuresCol`. We should only need to look at `inputCol`,  `inputCols` to be the input columns needed for pipeline stages, but ALS seems like an exception with `itemCol` and `userCol`. 

The first option example above will cover all stages (subject to further unit tests) and the second option just needs extra steps to cover an outlier pipeline stage like ALS (to handle custom params other than `inputCol`/`inputCols` that refer to the input columns from the training df).

BenWilson2: Could you add a presence check for 'itemCol' and 'userCol' (both must be present) for ALS to the logic you have in Option 2? 
Also, nit: `pipeline_input_cols = set()`

dbczumar: > @dbczumar that last print statement above does that check to filter out columns that aren't needed for any of the pipelines' stages. I just edited that line above to make it part of input_required_struct. thoughts?

Ah, my mistake. I missed the `re.search("(.*) does not exist.", iae.desc, re.IGNORECASE)` condition. Seems like this will work. @WeichenXu123 thoughts?

WeichenXu123: I incline to use this approach https://github.com/mlflow/mlflow/pull/5719#discussion_r854314443
which looks like a clever way.

One concern is, some model might raise error when hit empty input dataframe. So we should wrap try ... catch around it and if it failed, we log warning and skip logging signature.

WeichenXu123: > > I think a safer way is, for pyspark Transformer/Model type, add APIs "getInputColumns" and "getOutputColumns", and implement the API for each specific model. Note many models inherits Predictor Model / Classification Model, which we can implement "getInputColumns" and "getOutputColumns" in their base class.
> 
> @WeichenXu123 Is the implementation going to be standard for all subclasses of `Predictor Model` / `Classification Model`? If so, can we implement the same logic for these models in MLflow right now and contribute it to PySpark ML after? It's okay to begin with a limited set of models.

"implement the same logic for these models in MLflow side", we might need to implement many class including many transformers. Currently I incline to use https://github.com/mlflow/mlflow/pull/5719#discussion_r854314443 approach but eventually we should contribute to pyspark ML and implementing the safe getInputColumns/getOutputColumns logic

mlflow/pyspark/ml/__init__.py
```diff
@@ -924,9 +938,44 @@ def _log_posttraining_metadata(estimator, spark_model, params):
...
+                    limited_input_df = input_df.select(feature_cols).limit(
+                        INPUT_EXAMPLE_SAMPLE_ROWS
+                    )
+                    return cast_spark_df_with_vector_to_array(limited_input_df).toPandas()
+
+                def _infer_model_signature(input_example_slice):
+                    input_slice_df = find_and_set_features_col_as_vector_if_needed(
+                        spark.createDataFrame(input_example_slice), spark_model
+                    )
+                    model_output = spark_model.transform(input_slice_df)
```
WeichenXu123, line 962: The `spark_model.transform` return a new dataframe which appends new columns (e.g. prediction column) to input dataframe 
i.e., the real output is only the "appended" columns (i.e. prediction column), but the `model_output` here contains both the original input columns and appended columns.

I think here we should extract these appended columns as the real output column. @dbczumar what do you think ?

mlflow/pyspark/ml/_autolog.py
```diff
@@ -0,0 +1,95 @@
...
+        spark = SparkSession.builder.getOrCreate()
+    if input_fields is None:
+        input_fields = set()
+        # Using an empty dataset doesn't work for all estimators
+        # such as: pyspark.ml.classification.OneVsRest, so set
+        # a single row and column of double type
+        df = spark.createDataFrame([(1.0,)])
+    else:
+        df = spark.createDataFrame(
+            spark.sparkContext.emptyRDD(),
```
WeichenXu123, line 84: the empty input might not work in the case of "CrossValidator" or "TrainValidaationSplit", because they need to split the input dataset first.
I remember if input dataset contains too few rows it might fail.

Shall we do some special handling for them ?

WeichenXu123:
LGTM once final 2 questions addressed.

mlflow/pyspark/ml/__init__.py
dbczumar: @WeichenXu123 Yes, I agree! @bali0019 can we make this change?

tests/spark/autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -876,3 +898,237 @@ def test_autolog_registering_model(spark_session, dataset_binomial):
...
+    df_with_array_features = cast_spark_df_with_vector_to_array(dataset_binomial)
+    df_with_vector_features = find_and_set_features_col_as_vector_if_needed(
+        df_with_array_features, pipeline_model
+    )
+    assert isinstance(
+        [_field for _field in df_with_vector_features.schema.fields if _field.name == "features"][
+            0
+        ].dataType,
+        VectorUDT,
+    ), "'features' column wasn't cast to vector type"
```
harupy, line 1129: Can we split this into two lines?

```suggestion
    features_col = next(
        filter(lambda f: f.name == "features", df_with_vector_features.schema.fields)
    )
    assert isinstance(features_col.dataType, VectorUDT)

```

tests/spark/autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -876,3 +898,237 @@ def test_autolog_registering_model(spark_session, dataset_binomial):
...
+@pytest.mark.large
+def test_spark_df_with_vector_to_array_casts_successfully(dataset_multinomial):
+    from pyspark.sql.types import ArrayType, DoubleType
+
+    output_df = cast_spark_df_with_vector_to_array(dataset_multinomial)
+    assert [_field for _field in output_df.schema.fields if _field.name == "features"][
+        0
+    ].dataType == ArrayType(
+        DoubleType(), False
+    ), "'features' column isn't of expected type array<double>"
```
harupy, line 1102: Same here https://github.com/mlflow/mlflow/pull/5719/files#r858280087

WeichenXu123 (approved):
LGTM after addressing haru's new comments.

and the CI failure https://github.com/mlflow/mlflow/runs/6167872512?check_suite_focus=true should be not related, it does not block merging.

mlflow/spark.py
```diff
@@ -741,6 +741,56 @@ def _load_pyfunc(path):
     return _PyFuncModelWrapper(spark, _load_model(model_uri=path))
 
 
+def find_and_set_features_col_as_vector_if_needed(spark_df, spark_model):
```
dbczumar, line 744: Can we make this an internal method so that it doesn't appear in the docs?

```suggestion
def _find_and_set_features_col_as_vector_if_needed(spark_df, spark_model):
```

mlflow/spark.py
```diff
@@ -759,15 +809,26 @@ def predict(self, pandas_df):
...
-            # make sure predict work by default for Transformers
-            self.spark_model.stages[-1].setOutputCol("prediction")
+            from pyspark.sql import SparkSession
+
+            spark = SparkSession.builder.getOrCreate()
+            # do a transform with an empty input DataFrame
+            # to get the schema of the transformed DataFrame
+            transformed_df = self.spark_model.transform(spark.createDataFrame([], spark_df.schema))
+            # Ensure prediction column doesn't already exist
+            if prediction_column not in transformed_df.columns:
```
dbczumar, line 826: What was the previous behavior when the `"prediction"` column already existed? Did we throw? We should make sure to preserve the previous behavior to avoid breaking changes.

dbczumar (commented):
@bali0019 Awesome stuff! Left a couple small comments. Autologging test failures for Spark 3.0.0 do seem related to these code changes, but it's unclear why; as evidence, I independently kicked off testing from the master branch, and all tests pass: https://github.com/mlflow/mlflow/runs/6181935162?check_suite_focus=true. I'll work to root cause this today.

mlflow/spark.py
bali0019: yes, it threw `java.lang.IllegalArgumentException` exception, but it can only occur if the transformer is a pipeline type and if the last stage of that pipeline is not a Model transformer such as an `IndexToString` stage, I added unit tests fot it [here](https://github.com/mlflow/mlflow/pull/5719/files#diff-317be50f891ddc0e3a113301f7db84008e7d35818b097bfc4668958465661005R1009) and [here](https://github.com/mlflow/mlflow/pull/5719/files#diff-317be50f891ddc0e3a113301f7db84008e7d35818b097bfc4668958465661005R1023).
Considering that and knowing it threw an exception, this method was basically of no use currently if someone had a pipeline with an `IndexToString` stage in it. Do you still think we should continue to throw an exception and add this change in the next major version ?

dbczumar: Ah, got it. If the previous throwing behavior was purely a limitation and we're removing the limitation here by expanding the supported set of pipelines, I think this is great. Thanks Jas!

dbczumar (approved):
LGTM! Thanks @bali0019 ! I just pushed a change to convert `pytest.fixture()` decorators to `pytest.fixture`, which seems to resolve the autologging issue with pyspark 3.0.0 in testing here: https://github.com/mlflow/mlflow/runs/6202091822. Let's see if tests pass.

PR #5725, opened 2022-04-20
Update ag-grid and implement getRowId to improve runs table performance

### LINKED ISSUE #5653: [BUG] Loading more runs in the experiment UI becomes very slow with a large number of rows
### Willingness to contribute

### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Fedora Linux 35
- **MLflow installed from (source or binary)**: Source
- **MLflow version (run ``mlflow --version``)**: 1.24.1.dev0 (at commit 3984c2c9a536505eabd183a04467ab52a80127db)
- **Python version**: 3.9.12
- **npm version, if running the dev UI**: 8.3.1
- **Exact command to reproduce**: Create > 1000 runs in an experiment. Then in the experiment view, keep clicking "Load more"

### Describe the problem

As the number of runs in the run table increases, clicking "Load more" to load more rows takes longer and performance degrades significantly.

The table below shows the time taken to load 100 more rows, measured as the time from clicking "Load more" until the new rows are fully rendered and the load more button is visible again.

| Starting number of rows | Time taken to load 100 more rows |
|-------------------------|----------------------------------|
| 100 | 3 s |
| 200 | 5 s |
| 300 | 8 s |
| 400 | 11 s |
| 500 | 15 s |
| 600 | 21 s |
| 700 | 26 s |
| 800 | 37 s |
| 900 | 45 s |
| 1000 | 59 s |

Only around 100 ms of each loading time is actually fetching the data from the backend, plus two queries to the model-versions/search endpoint that take less than 10 ms each (they are returning empty responses in my test case).

These numbers are all based on running a production build of the UI from commit 3984c2c9a536505eabd183a04467ab52a80127db, using Firefox 98 on Fedora Linux, and using a local SQLite store, on a machine with an AMD 5900X and 64 GB of RAM. I see the same behaviour under Chrome on Linux and Windows too, although performance does seem a bit better with Chrome.

I've tested a series of changes that all help to improve performance significantly:
* Updating to ag-grid 27.1.0, which reduces the time to go from 1000 to 1100 runs to 33 s
* Removing the use of the full width cell renderer and moving the LoadMoreBar out of the table: down to 11 s
* Implementing `getRowId`, which is supposed to help prevent re-rendering rows that haven't changed when updating the data (https://www.ag-grid.com/react-data-grid/row-ids/): 3 s

Just using `getRowId` without removing the full width cell renderer didn't work, this caused some errors internally within ag-grid (`this.cellComp is undefined`).

And making these two changes while sticking to version 25.3.0 didn't work well either, removing the full width cell renderer improved performance but implementing `getRowNodeId` (which was deprecated in 27.1.0 and replaced by `getRowId`) seemed to make performance worse.

There are a couple of new warnings that appear with a dev build after the update to 27.1.0:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
AgGridReactUi@http://localhost:3000/static-files/static/js/bundle.js:117332:20
AgGridReact@http://localhost:3000/static-files/static/js/bundle.js:116686:20
div
ExperimentRunsTableMultiColumnView2@http://localhost:3000/static-files/static/js/bundle.js:17195:193
div
Spacer@http://localhost:3000/static-files/static/js/bundle.js:45732:193
div
div
ExperimentView@http://localhost:3000/static-files/static/js/bundle.js:18230:193
injectIntl(ExperimentView)
ConnectFunction@http://localhost:3000/static-files/static/js/bundle.js:423772:68
C@http://localhost:3000/static-files/static/js/bundle.js:426980:31
```

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
GridBodyComp@http://localhost:3000/static-files/static/js/bundle.js:118705:37
TabGuardCompRef@http://localhost:3000/static-files/static/js/bundle.js:120742:18
div
div
GridComp@http://localhost:3000/static-files/static/js/bundle.js:118950:17
div
AgGridReactUi@http://localhost:3000/static-files/static/js/bundle.js:117332:20
AgGridReact@http://localhost:3000/static-files/static/js/bundle.js:116686:20
div
ExperimentRunsTableMultiColumnView2@http://localhost:3000/static-files/static/js/bundle.js:17195:193
div
Spacer@http://localhost:3000/static-files/static/js/bundle.js:45732:193
div
div
ExperimentView@http://localhost:3000/static-files/static/js/bundle.js:18230:193
injectIntl(ExperimentView)
ConnectFunction@http://localhost:3000/static-files/static/js/bundle.js:423772:68
C@http://localhost:3000/static-files/static/js/bundle.js:426980:31 react-dom.development.js:67
```

It looks like this has been opened as a bug in ag-grid but closed without any activity: https://github.com/ag-grid/ag-grid/issues/4817 (if Databricks have a support contract with ag-grid, maybe you could get them to take a look?)

There's also currently a problem with nested runs where the parent run's row doesn't re-render when it is expanded or contracted, so the plus icon stays as a plus. Hopefully I can find a workaround for that but I thought I should open an issue before doing too much work on this.

Edit: Also, the start time and model columns don't get re-rendered for in-progress runs when clicking "Refresh" (and other columns with custom renderers are probably affected too). I'm guessing this is because eg. the model column refers to a non-existent `models` field, and the `startTime` field value doesn't actually change, so ag-grid doesn't think these need re-rendering. It should be possible to make this work correctly by using fields that actually change when the rendered result should change.

Would you be happy to consider a PR with these changes?

### Code to reproduce issue

NA

### Other info / logs

Profiling data from Firefox when going from 500 to 600 rows, building with `yarn run build --profile`: https://share.firefox.dev/3jjD11S

Components 

Interface 

Language 

Integrations

adamreeve:

Upgrades to ag-grid 27.2.0 and makes the following changes to improve the performance of the runs table when loading more rows:
* Move the "Load more" button out of the grid and just render it below the grid in a separate element
* Implement `getRowId` using the run uuid so that previously rendered rows aren't re-rendered (see https://www.ag-grid.com/react-data-grid/row-ids/)
  * In order to correctly re-render cells when data has changed, this required making sure each column definition had a field specified that corresponded to an actual field present in the row data. Eg. previously the `models` column referred to the `models` field that didn't actually exist, so models were always considered equal and weren't re-rendered when they changed. Similarly, the `date` column actually used many different fields that affected how the cell should be rendered, so just comparing `startTime` values wasn't sufficient to decide whether the cell should be re-rendered.

Fixes #5653 (see that issue for some performance numbers)

## How is this patch tested?

Running existing unit tests, manual testing of the UI to check that behaviour seems correct.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Improve the performance of the runs table when loading a large number of runs.

Components 

Interface 

Language 

Integrations

(I'm not sure about this classification, feel free to change it)


dbczumar:
@harupy @sunishsheth2009 Can you take a look?

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
```diff
@@ -179,10 +176,26 @@ export class ExperimentRunsTableMultiColumnView2 extends React.Component {
         },
         {
           headerName: ATTRIBUTE_COLUMN_LABELS.DATE,
-          field: 'startTime',
+          field: 'runDateInfo',
```
sunishsheth2009, line 179: I am curious on this, Are we getting the same data from a new field now?

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
```diff
@@ -179,10 +176,26 @@ export class ExperimentRunsTableMultiColumnView2 extends React.Component {
...
+              dateInfo1.referenceTime === dateInfo2.referenceTime &&
+              dateInfo1.startTime === dateInfo2.startTime &&
+              dateInfo1.experimentId === dateInfo2.experimentId &&
+              dateInfo1.runUuid === dateInfo2.runUuid &&
+              dateInfo1.runStatus === dateInfo2.runStatus &&
+              dateInfo1.isParent === dateInfo2.isParent &&
+              dateInfo1.hasExpander === dateInfo2.hasExpander &&
+              dateInfo1.expanderOpen === dateInfo2.expanderOpen &&
+              _.isEqual(dateInfo1.childrenIds, dateInfo2.childrenIds)
+            );
```
sunishsheth2009, line 197: Why do we need to check the equals here. Where is this function evoked from and how do we use it?

sunishsheth2009 (commented):
I have some questions on how do we improve performance here? 

I thought setting the `getRowId` to runId is enough. What are all the other changes needed for?

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
adamreeve: Sort of. I've added a corresponding new `runDateInfo` property in the data returned from `getRowData`, which is an object aggregating all of the data needed to render the date cell rather than just the start time. And `DateCellRenderer` now accesses these values through its `value` prop instead of the `data` prop that contains data for the whole row. This is needed so that we can correctly implement the equals property to handle deciding whether the cell needs to be re-rendered.

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
adamreeve: It isn't used directly within mlflow code, but is used by ag-grid to decide whether a cell needs to be re-rendered. After the data is updated, ag-grid can decide not to re-render a cell if the row id is the same and the cell values are considered to be equal. If we didn't implement this then a lot of cells would unnecessarily be re-rendered because by default reference equality is used.

This wasn't needed previously when row ids were assigned by ag-grid as a new set of ids is assigned after data is updated.

adamreeve:
> I thought setting the `getRowId` to runId is enough. What are all the other changes needed for?

Implementing `getRowId` is the main change that improves performance, and most of the other changes are needed for the grid to work correctly with application assigned row ids and re-render cells when needed.

Eg. previously for the "Start Time" column, the value of the cell was set to the `startTime` field, but the `DateCellRenderer` actually used a bunch of other properties from the row data to render this cell. This meant that when data was updated, the "Start Time" cells were never re-rendered even if things like `expanderOpen` had changed, because ag-grid only tested for equality of `startTime` to decide whether to re-render this cell. I've also had to add a `referenceDate` field here so that cells are re-rendered when time passes due to start time being rendered as something like "x seconds ago".

Similarly for the "Models" column, this was configured to use a non-existent `models` field, which would always evaluate to undefined so values for this column were always considered equal and not re-rendered, so I've had to add a new `models` field to fix the rendering behaviour for this column.

There are similar changes to other columns so that every column is mapped to a value field that can be tested for equality to decide whether cells need to be re-rendered when data is updated. This wasn't a problem previously when using grid assigned row ids as new ids were assigned whenever data was updated and every row was completely re-rendered.

The two changes that aren't directly related to implementing `getRowId` are the change to the `runInfosByUuid` reducer, which is a minor performance improvement I added after seeing this take a fairly long time when profiling, and moving the `LoadMore` button outside of the grid.

Moving the load more button out gave a performance improvement on its own, as I guess ag-grid then didn't have to test each row to see if it needed to use the `FullWidthCellRenderer` and could simplify its rendering logic. I'm not sure if this would still show as big a performance improvement when done after the `getRowId` change. When I tried just implementing `getRowId` without moving the load more button and using ag-grid 27.1.0, I was getting errors within ag-grid. Possibly these have been fixed in 27.2.0 but I'd argue it's still better to keep the load more button outside of the grid as rendering it within a row just seems to add unnecessary complication without any obvious benefit.

sunishsheth2009 (approved):
Thank for the detailed explanation. It makes sense. :) 
Also thank you for making these changes and the upgrade. Appreciate your help 

@xanderwebs can you take a look at it as well?

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
xanderwebs: Since we're using _.equals here already, what's stopping us from doing `_.isEqual(dateInfo1, dateInfo2)`?

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
```diff
@@ -673,27 +693,32 @@ function SourceCellRenderer(props) {
...
+  return Utils.renderSourceVersion(version, name, type) || EMPTY_CELL_PLACEHOLDER;
 }
 
-ExperimentNameRenderer.propTypes = { data: PropTypes.object };
+VersionCellRenderer.propTypes = { value: PropTypes.object };
+
 function ExperimentNameRenderer(props) {
-  const { experimentId, experimentName, experimentBasename } = props.data;
+  const { experimentId } = props.data;
+  const { name, basename } = props.value;
```
xanderwebs, line 707: What's the difference between `data` and `value` here? ==> The question behind this question is really, why are we reading `experimentId` in `DateCellRenderer` off of `value` there, but off of `data` here?

xanderwebs (changes requested):
Thanks for the contribution, code generally looks good, the only thing from my end is that it would be good to keep the function signature of `Utils.renderSource` the same (even if the params are unused here).

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
adamreeve: I was concerned there might be more performance overhead of using `_.isEqual` but I haven't tested that to verify. Using `_.isEqual(dateInfo1, dateInfo2)` would definitely simplify things so I'll switch to that if I don't see a big performance difference.

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
adamreeve: `data` contains the data for the whole row, whereas `value` is the value for this cell (selected from the row data using the `field` configured for the column). The `value` here doesn't include the `experimentId` because I'm using the experiment name values directly from the map returned by `Utils.getExperimentNameMap` in `getRowData`.

I could probably create new value objects that also include the `experimentId` which might be tidier, but that seems unnecessary as the experiment id for a row is never going to change (if it could change it would be important to include it so that the equality comparison was correct).

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
adamreeve: Yeah using `_.isEqual` for the top-level comparisons is slower but it's pretty insignificant compared to the time everything else takes. Eg. when loading 100 more rows with 1000 already loaded I get 25 ms in the value comparisons with the current code and then 100 ms when changing all the comparisons to just use `_.isEqual`, but the whole loading operation takes multiple seconds so I'll make this change.

mlflow/server/js/src/experiment-tracking/components/ExperimentRunsTableMultiColumnView2.js
adamreeve: I've added a comment in the function to explain this too.

adamreeve:
Thanks for the feedback @xanderwebs, I think I've addressed all of your comments now.

xanderwebs (approved):
LGTM!

Issue #5739, opened 2022-04-21
[FR] Better support for ndarrays and tensors


ntakouris:
**Please fill in this feature request template to ensure a timely and thorough response.**

## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [ ] Yes. I can contribute this feature independently.
- [  x ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

Many times, you log metrics such as loss per step (tensors with gradients) or other numpy arrays (ndarrays). It would be very useful if there was built-in support to:

**For params**
- Extract single (`.item()`) or multiple values from ndarrays ([tolist](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tolist.html) ?)

**For metrics**
- [Detach torch tensors from the gradient graph](https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html) - pytorch lightning does this too
- Move tensors from gpu to cpu, convert to numpy if needed (ex. to convert to list)
- Extract values from single-element tensors or ndarrays, raise if resulting value is not supported (not float?)

Adding those few simple features, can make the usage much more intuitive and simpler, for deep learning engineers. Here's an example:

```py
anchors = np.array([[16], [32], [64], [128], [256]], dtype=np.int32)
x = np.array([1.0])

# in your train on batch func
loss = ...
```

- Before
```py
mlflow.log_params({'anchors': anchors.tolist(), 'x': x.item(), 'loss': loss.detach().item()/.tolist()})
````

- After
```py
mlflow.log_params({'anchors': anchors, 'x': x, 'loss': loss})
````

### What component(s), interfaces, languages, and integrations does this feature affect?
Components
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
[...]

PR #5756, opened 2022-04-22
Tracking: Better support for ndarrays and tensors

ntakouris:

See #5739 

Many times, you log metrics such as loss per step (tensors with gradients) or other numpy arrays (ndarrays). It would be very useful if there was built-in support to:

**For params**
- Extract single (`.item()`) or multiple values from ndarrays ([tolist](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tolist.html) ?)

**For metrics**
- [Detach torch tensors from the gradient graph](https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html) - pytorch lightning does this too
- Move tensors from gpu to cpu, convert to numpy if needed (ex. to convert to list)
- Extract values from single-element tensors or ndarrays, raise if resulting value is not supported (not float?)

Adding those few simple features, can make the usage much more intuitive and simpler, for deep learning engineers. Here's an example:

```py
anchors = np.array([[16], [32], [64], [128], [256]], dtype=np.int32)
x = np.array([1.0])

# in your train on batch func
loss = ...
```

- Before
```py
mlflow.log_params({'anchors': anchors.tolist(), 'x': x.item(), 'loss': loss.detach().item()/.tolist()})
````

- After
```py
mlflow.log_params({'anchors': anchors, 'x': x, 'loss': loss})
````
## How is this patch tested?

- WIP: Will add tests once some initial code review is performed
- 

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

WIP

Components

### CODE REVIEW

File: mlflow/entities/conversion_utils.py [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+
+try:
+    import tensorflow as tf
+    __import_np = True
+except ImportError:
+    __import_tf = False
+
+try:
+    import torch
+    __import_torch = True
```
WeichenXu123, line 22: Can we avoid importing pyspark/torch/tensorflow when we importing this `conversion_utils` module ? We can move these lines into function.

mlflow/entities/param.py
```diff
@@ -1,21 +1,17 @@
...
 class Param(_MLflowObject):
     """
     Parameter object.
     """
 
     def __init__(self, key, value):
-        if "pyspark.ml" in sys.modules:
-            import pyspark.ml.param
+        key = convert_to_str_if_possible(key)
+        value = convert_to_str_if_possible(value)
```
WeichenXu123, line 13: I would like to keep `if isinstance(key, pyspark.ml.param.Param):` and `key = key.name` logic inside `Param.__init__`, because convert_to_str_if_possible is a generic function but the `Param.name` logic only apply to "Param key" case.

WeichenXu123, line 13: I would like to keep `if isinstance(key, pyspark.ml.param.Param):` and `key = key.name` logic inside `Param.__init__`, because convert_to_str_if_possible is a generic function but the `Param.name` logic only apply to "Param key" case.

WeichenXu123 (commented):
We need to document log_metric / log_param related APIs, to say we now support logging these types values.

mlflow/entities/param.py
ntakouris: The refactor which was performed leads to equivalent behaviour. Is there a specific reason that you want this branching inside the constructor even if it leads to the exact same result?

I have not checked if tests cover that because I am waiting for the code review to end.

mlflow/entities/conversion_utils.py
```diff
@@ -0,0 +1,117 @@
...
+    converter_fns_to_try = [convert_to_str_if_pyspark_mlparam, convert_to_str_if_ndarray, convert_to_str_if_tensorflow_tensor, convert_to_str_if_torch_tensor]
+
+    for converter_fn in converter_fns_to_try:
+        possible_str = converter_fn(x)
+        if type(possible_str) == str:
+            return possible_str
+
+    return str(x)
+
+def convert_to_float_if_possible(x) -> float:
```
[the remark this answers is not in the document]
ntakouris: +1, will do.

I think we need to also introduce a `MetricKey` and `MetricValue` type, so that:

```py
MetricKey = Union[str, pyspark.ml.param.Param]
MetricValue = Union[float, np.ndarray, tf.Tensor, torch.Tensor]
```

... but I am not sure how to do that without eagerly importing all these modules (which I guess we want to avoid)

Would just adding a doc, naming and tests should be sufficient?

mlflow/entities/conversion_utils.py
WeichenXu123: You don't need to use flag like `__np_module_imported`

if we execute `import xxx` command twice, the second execution do nothing, because when we import one module, it is cached in `sys.modules`.

mlflow/entities/param.py
WeichenXu123: ok it's fine.

mlflow/entities/conversion_utils.py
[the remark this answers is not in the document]
WeichenXu123: Don't need to add type hint for now. Avoid importing unnecessary modules is more important. The type hint requires importing modules in top level it will cause the issue.

WeichenXu123:
@ntakouris Coud you rerun failed CI and see whether they passed ?

ntakouris:
@WeichenXu123 can give this a final review? tests pass locally.

examples/h2o/random_forest.ipynb
```diff
@@ -94,12 +94,13 @@
     "import mlflow\n",
```
dbczumar, line 94: @ntakouris Do we need to make these formatting changes to the example notebooks? If not, can we revert them? Let me know if I'm missing something, but these don't appear to be related to the PR.

mlflow/entities/metric.py
```diff
@@ -1,15 +1,33 @@
...
+    convert_metric_value_to_str_if_possible,
+)
+
 
 class Metric(_MLflowObject):
     """
     Metric object.
+
+    Args:
+        key: `str`, `pyspark.ml.param.Param`, `numpy.ndarray`, `tensorflow.Tensor` or `torch.Tensor`
```
dbczumar, line 15: @ntakouris Apologies if I'm asking a question that's already been answered, but can you provide some information about the use case for having numpy arrays / tensors as metric names? This seem useful as metric values, though MLflow only supports scalar metrics; however, I don't see a clear use case for these as metric names (`key` == `name`). Also, as far as I'm aware, MLflow's Python APIs will cast metric names to strings automatically during logging, so I don't think this is strictly necessary.

mlflow/entities/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,141 @@
...
+
+    return x
+
+
+@__converter_requires("numpy")
+def convert_metric_value_to_float_if_ndarray(x):
+    import numpy as np
+
+    if isinstance(x, np.ndarray):
+        return float(x.item())
```
dbczumar, line 88: This will throw if the array has multiple elements. What does the user-facing exception look like in that case? Should we catch the exception and throw something more readable?

mlflow/entities/metric.py
```diff
@@ -1,15 +1,33 @@
...
+from .metic_value_conversion_utils import (
+    convert_metric_value_to_float_if_possible,
+    convert_metric_value_to_str_if_possible,
+)
+
 
 class Metric(_MLflowObject):
     """
     Metric object.
+
```
dbczumar, line 13: @ntakouris Rather than making changes to the low-level metric entity, I think we should update the fluent `mlflow.log_metric` and `mlflow.log_metrics()` APIs to support these additional types (https://github.com/mlflow/mlflow/blob/37c678320fbf82b93d2614dd14c0bee784d95175/mlflow/tracking/fluent.py#L559-L587). The fundamental definition of a metric should have a string key and a scalar numeric value, but upstream user-facing APIs can definitely accept multiple types and coerce them to this format. Please let me know if you have questions here.

dbczumar (changes requested):
@ntakouris Thank you for raising this PR! Supporting scalar numpy arrays / tensors for metric logging seems very convenient for users. I left some feedback about the level of abstraction where we're introducing this capability as well as some questions about the use case for arrays as keys, which seems a bit less clear. Please let me know if you have questions! Looking forward to getting this merged!

cc @WeichenXu123

mlflow/entities/metic_value_conversion_utils.py
ntakouris: This will just raise a `ValueError`, because it expects this to be a singular value. I can maybe catch that value error and re-raise a `ValueError` with a more helpful message?

examples/h2o/random_forest.ipynb
ntakouris: My apologies, I thought I reverted that, but it certainly did not take effect. It's was a sideffect of using `black . `

mlflow/entities/metric.py
ntakouris: Agreed. Will migrate changes to the high level api.

mlflow/entities/metric.py
ntakouris: I think we shouldn't do that for metric names. For param values though, we can convert ndarrays/tensors to str though.

mlflow/entities/metic_value_conversion_utils.py
WeichenXu123: Shall we catch the original exception and wrap it to be a `MlflowException` ?

ftsamis:
LGTM

ntakouris:
@dbczumar most changes are done, do you want to include them into `mlflow/tracking/_tracking_service/client.py` , `mlflow/tracking/fluent.py` or `mlflow/tracking/client.py` ?

examples/h2o/random_forest.ipynb
dbczumar: Got it! Can we revert these by checking out the notebooks from `master`?

dbczumar:
> @dbczumar most changes are done, do you want to include them into `mlflow/tracking/_tracking_service/client.py` , `mlflow/tracking/fluent.py` or `mlflow/tracking/client.py` ?

@ntakouris Apologies for the delay. Can we add the changes to `mlflow/tracking/_tracking_service/client.py` and remove them from the `Param` and `Metric` entity definitions?

mlflow/entities/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        convert_metric_value_to_str_if_tensorflow_tensor,
+        convert_metric_value_to_str_if_torch_tensor,
+    ]
+
+    for converter_fn in converter_fns_to_try:
+        possible_str = converter_fn(x)
+        if type(possible_str) == str:
+            return possible_str
+
+    return str(x)
```
dbczumar, line 39: @ntakouris I'm still not fully following why we need to special case string conversion for the use of tensors / pyspark mlparams as metric keys (names). I think it should be sufficient to convert all `key` inputs to string using `str()`. Is there a particular use case you have for using a tensor as a metric name?

examples/h2o/random_forest.ipynb
ntakouris: Yes, I have not yet pushed all the changes.

mlflow/entities/metic_value_conversion_utils.py
ntakouris: I've not pushed my local changes.

ntakouris:
Are those the api changes you expect?

mlflow/tracking/_tracking_service/client.py
```diff
@@ -222,14 +227,18 @@ def log_metric(self, run_id, key, value, timestamp=None, step=None):
...
     def log_param(self, run_id, key, value):
         """
-        Log a parameter against the run ID. Value is converted to a string.
+        Log a parameter against the run ID.
+
+        :param value: `str`, or ndarray or tensor, which will be converted to
+                      a list and then stringified.
         """
-        param = Param(key, str(value))
+        param = Param(key, convert_metric_value_to_str_if_possible(value))
```
dbczumar, line 241: @ntakouris Do we need to make changes here? The string representation of an ndarray is already human-readable:

```
In [4]: k = np.array([1, 2, 3])

In [5]: str(k)
Out[5]: '[1 2 3]'
```

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
+import sys
```
dbczumar, line 1: Nit: Looks like there's a small typo in the naming of this file (missing an `r` in metric)

Can we rename to `metric_value_conversion_utils.py`?

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
...
+def is_module_imported(module_name: str) -> bool:
+    return module_name in sys.modules
+
+
+def __try_get_item(x):
+    try:
+        xval = x.item()
+    except ValueError:
+        raise MlflowException(
+            f"Expected metric value to contain a single element, got shape {x.shape} instead"
```
dbczumar, line 15: Can we add the INVALID_PARAMETER_VALUE `error_code` as well? Example: https://github.com/mlflow/mlflow/blob/1539d9e467bcd865c335e800dad597517b0a9bcd/mlflow/pyfunc/__init__.py#L791-L793

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
+import sys
+
+from mlflow.exceptions import MlflowException
+
+
+def is_module_imported(module_name: str) -> bool:
+    return module_name in sys.modules
+
+
+def __try_get_item(x):
```
dbczumar, line 10: Nit: Can we use single underscore prefixes for these method names?

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    try:
+        xval = x.item()
+    except ValueError:
+        raise MlflowException(
+            f"Expected metric value to contain a single element, got shape {x.shape} instead"
+        )
+
+    return xval
```
dbczumar, line 18:
```suggestion
    try:
        return x.item()
    except ValueError:
        raise MlflowException(
            f"Expected metric value to contain a single element, got shape {x.shape} instead"
        )
```

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def __converter_requires(module_name: str):
```
dbczumar, line 21: Very neat decorator! :D

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
...
+        convert_metric_value_to_str_if_tensorflow_tensor,
+        convert_metric_value_to_str_if_torch_tensor,
+    ]
+
+    for converter_fn in converter_fns_to_try:
+        possible_str = converter_fn(x)
+        if type(possible_str) == str:
+            return possible_str
+
+    return str(x)
```
dbczumar, line 52: Following from https://github.com/mlflow/mlflow/pull/5756/files#r866392045, I don't think we strictly need this method at this time. I also see that it's only called for parameter values (not metrics), so, at minimum, we should adjust the naming

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def convert_metric_value_to_float_if_possible(x, convert_int: bool = True) -> float:
```
dbczumar, line 55: Looks like `convert_int = False` is never specified by any callers. Shall we remove it?

Looks like we always convert to float when logging metrics anyway:

```
import mlflow

from mlflow.tracking.client import MlflowClient

client = MlflowClient()

run = mlflow.start_run()
mlflow.log_metric("M", 1)

run_fetched = client.get_run(run.info.run_id)
print(run_fetched.data.metrics)
```

yields

```
1.0
```

mlflow/tracking/metic_value_conversion_utils.py
```diff
@@ -0,0 +1,163 @@
+import sys
+
+from mlflow.exceptions import MlflowException
+
+
+def is_module_imported(module_name: str) -> bool:
```
dbczumar, line 6:
```suggestion
def _is_module_imported(module_name: str) -> bool:
```

tests/tracking/fluent/test_metric_value_conversion_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+import numpy as np
+import tensorflow as tf
+import torch
+
+import pyspark.ml.param
+import pyspark.ml.util
+
+
+class TestMetricValueConversionUtils(unittest.TestCase):
```
dbczumar, line 15: Nit: Can we remove `unittest.TestCase` in favor of pytest-style test cases? For example, see https://github.com/mlflow/mlflow/blob/master/tests/models/test_model_input_examples.py.

tests/tracking/test_tracking.py
```diff
@@ -451,9 +451,9 @@ def test_set_tags():
 def test_log_metric_validation():
     with start_run() as active_run:
         run_id = active_run.info.run_id
-        with pytest.raises(MlflowException, match="Got invalid value apple for metric") as e:
+        with pytest.raises(MlflowException):
```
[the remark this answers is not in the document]
dbczumar: Ideally, it should be up to the backend to determine the validity of a metric value. If our attempted conversion fails, instead of throwing, can we propagate the original metric value as part of the request?

dbczumar (commented):
@ntakouris This is looking great! I left a few more comments; we should be close to merge-ready once they're addressed! Thanks a bunch!

mlflow/tracking/_tracking_service/client.py
ntakouris: Agreed, will remove.

mlflow/tracking/metic_value_conversion_utils.py
ntakouris: Sure thing!

ntakouris:
If latest changes are ok, I can proceed to add a few sanity tests for the high level client logging api.

dbczumar:
> If latest changes are ok, I can proceed to add a few sanity tests for the high level client logging api.

Hi @ntakouris, apologies for the delay! The latest changes look great; some test coverage would be fantastic! Thank you for your contribution!

ntakouris:
@dbczumar @WeichenXu123 Bump, this should be ready to merge

tests/tracking/fluent/test_metric_value_conversion_utils.py
```diff
@@ -0,0 +1,47 @@
...
+        time_mock.side_effect = [123 for _ in range(100)]
+        run_id = active_run.info.run_id
+        mlflow.log_metric("name_numpy", ndarray_val)
+
+    finished_run = tracking.MlflowClient().get_run(run_id)
+    # Validate metrics
+    assert len(finished_run.data.metrics) == 1
+    expected_pairs = {"name_numpy": ndarray_float_val}
+    for key, value in finished_run.data.metrics.items():
+        assert expected_pairs[key] == value
```
harupy, line 47: Can we simplify the code?

```suggestion
    with start_run() as run:
        mlflow.log_metric("name_numpy", ndarray_val)

    finished_run = tracking.MlflowClient().get_run(run.info.run_id)
    assert len(finished_run.data.metrics) == {"name_numpy": ndarray_float_val}
```

tests/tracking/test_tracking.py
```diff
@@ -451,8 +451,9 @@ def test_set_tags():
 def test_log_metric_validation():
     with start_run() as active_run:
         run_id = active_run.info.run_id
-        with pytest.raises(MlflowException, match="Got invalid value apple for metric") as e:
+        with pytest.raises(MlflowException) as e:
             mlflow.log_metric("name_1", "apple")
+
```
harupy, line 456: Why do we need this change? `match` should not be removed.

ntakouris:
@harupy Addressed both comments.

Also rebased and force-pushed on master => CI got fixed.

tests/pytorch/test_pytorch_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+from unittest import mock
+
+import mlflow
+from mlflow import tracking
+from mlflow.tracking.fluent import start_run
+from mlflow.exceptions import MlflowException, INVALID_PARAMETER_VALUE, ErrorCode
+from mlflow.tracking.metric_value_conversion_utils import *
```
harupy, line 8: Can we avoid using wild car import?

tests/pytorch/test_pytorch_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+from unittest import mock
```
harupy, line 2: Let's remove unused `mock.`

tests/pytorch/test_pytorch_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,35 @@
...
+from mlflow.exceptions import MlflowException, INVALID_PARAMETER_VALUE, ErrorCode
+from mlflow.tracking.metric_value_conversion_utils import *
+
+import torch
+
+
+def test_reraised_value_errors():
+    multi_item_torch_tensor = torch.rand((2, 2))
+
+    with pytest.raises(MlflowException) as e:
```
harupy, line 16:
```suggestion
    with pytest.raises(MlflowException, match="<replace me>") as e:
```

Can we add `match`?

tests/pytorch/test_pytorch_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+def test_log_torch_tensor_as_metric():
+    torch_tensor_val = torch.rand(1)
+    torch_tensor_float_val = float(torch_tensor_val[0])
+
+    with start_run() as run:
+        mlflow.log_metric("name_torch", torch_tensor_val)
+
+    finished_run = tracking.MlflowClient().get_run(run.info.run_id)
+    expected_pairs = {"name_torch": torch_tensor_float_val}
```
harupy, line 35:
```suggestion
    assert finished_run.data.metrics == {"name_torch": torch_tensor_float_val}
```

tests/tensorflow/test_tensorflow2_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+def test_log_tf_tensor_as_metric():
+    tf_tensor_val = tf.random.uniform([])
+    tf_tensor_float_val = float(tf_tensor_val.numpy())
+
+    with start_run() as run:
+        mlflow.log_metric("name_tf", tf_tensor_val)
+
+    finished_run = tracking.MlflowClient().get_run(run.info.run_id)
+    expected_pairs = {"name_tf": tf_tensor_float_val}
```
harupy, line 35:
```suggestion
    assert finished_run.data.metrics == {"name_tf": tf_tensor_float_val}
```

tests/tracking/fluent/test_metric_value_conversion_utils.py
```diff
@@ -0,0 +1,41 @@
...
+from mlflow.tracking.metric_value_conversion_utils import *
+from tests.helper_functions import random_int
+
+import numpy as np
+
+
+def test_reraised_value_errors():
+    multi_item_array = np.random.rand(2, 2)
+
+    with pytest.raises(MlflowException) as e:
```
harupy, line 18: same here https://github.com/mlflow/mlflow/pull/5756/files#r889763618

tests/tracking/test_tracking.py
```diff
@@ -453,6 +453,7 @@ def test_log_metric_validation():
         run_id = active_run.info.run_id
         with pytest.raises(MlflowException, match="Got invalid value apple for metric") as e:
             mlflow.log_metric("name_1", "apple")
+
```
harupy, line 456: Let's revert this change.

tests/pytorch/test_pytorch_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+    with pytest.raises(MlflowException) as e:
+        convert_metric_value_to_float_if_possible(multi_item_torch_tensor)
+
+    assert e.value.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+
+
+def test_convert_metric_value_to_float():
+    torch_tensor_val = torch.rand(1)
+    assert float(torch_tensor_val[0]) == convert_metric_value_to_float_if_possible(torch_tensor_val)
```
harupy, line 24:
```suggestion
    assert convert_metric_value_to_float_if_possible(torch_tensor_val) == float(torch_tensor_val[0])
```

Can we move the actual value to the left?

harupy: Can we also apply this change to other assert statements?

harupy:
@ntakouris Left a few more comments. LGTM once they are addressed.

tests/pytorch/test_pytorch_metric_value_convertion_utils.py
ntakouris: Of course! will also update other 2 tests that use this.

ntakouris:
Addressed the changes.

tests/tensorflow/test_tensorflow2_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    with pytest.raises(
+        MlflowException, match=r"Expected metric value to contain a single element"
+    ) as e:
+        convert_metric_value_to_float_if_possible(multi_item_tf_tensor)
+
+    assert e.value.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+
+
+def test_convert_metric_value_to_float():
+    tf_tensor_val = tf.random.uniform([])
```
harupy, line 24:
```suggestion
    tf_tensor_val = tf.random.uniform([])  # pytlin: disable=no-value-for-parameter
```

Let's add a `# pytlin: disable=no-value-for-parameter` here to ignore the pylint error below.

tests/tensorflow/test_tensorflow2_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    with pytest.raises(
+        MlflowException, match=r"Expected metric value to contain a single element"
+    ) as e:
+        convert_metric_value_to_float_if_possible(multi_item_tf_tensor)
+
+    assert e.value.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+
+
+def test_convert_metric_value_to_float():
+    tf_tensor_val = tf.random.uniform([])  # pytlin: disable=no-value-for-parameter
```
harupy, line 24:
```suggestion
    tf_tensor_val = tf.random.uniform([])  # pylint: disable=no-value-for-parameter
```

Sorry for the typo.

tests/tensorflow/test_tensorflow2_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    assert convert_metric_value_to_float_if_possible(tf_tensor_val) == float(tf_tensor_val.numpy())
+
+
+def test_log_tf_tensor_as_metric():
+    tf_tensor_val = tf.random.uniform([])
```
harupy, line 29:
```suggestion
    tf_tensor_val = tf.random.uniform([])  # pylint: disable=no-value-for-parameter
```

tests/tensorflow/test_tensorflow2_metric_value_convertion_utils.py
```diff
@@ -0,0 +1,36 @@
...
+from mlflow import tracking
+from mlflow.tracking.fluent import start_run
+from mlflow.exceptions import MlflowException, INVALID_PARAMETER_VALUE, ErrorCode
+from mlflow.tracking.metric_value_conversion_utils import convert_metric_value_to_float_if_possible
+
+import tensorflow as tf
+
+
+def test_reraised_value_errors():
+    multi_item_tf_tensor = tf.random.uniform([2, 2])
```
harupy, line 13:
```suggestion
    multi_item_tf_tensor = tf.random.uniform([2, 2])  # pylint: disable=no-value-for-parameter
```

ntakouris:
@harupy No need to skip pylint. I just added `dtype=tf.float32`.

ntakouris:
@harupy pylint passes locally. Mind letting this CI go one more time?

mlflow/tracking/metric_value_conversion_utils.py
```diff
@@ -0,0 +1,91 @@
...
+        return x.item()
+    except ValueError:
+        raise MlflowException(
+            f"Expected metric value to contain a single element, got shape {x.shape} instead",
+            error_code=INVALID_PARAMETER_VALUE,
+        )
+
+
+def _converter_requires(module_name: str):
+    """Wrapper function that checks if specified `module_name` is already imported before invoking wrapped function."""
```
harupy, line 21: Let's fix this line.

ntakouris:
@harupy  I reduced the length of that line. It wasn't caught by `black`.

ntakouris:
Looks good now?

mlflow/tracking/metric_value_conversion_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    import tensorflow as tf
+
+    if isinstance(x, tf.Tensor):
+        try:
+            return float(x)
+        except TypeError:
+            raise MlflowException(
+                f"Expected metric value to contain a single element, got shape {x.shape} instead",
+                error_code=INVALID_PARAMETER_VALUE,
+            )
```
harupy, line 90: same here https://github.com/mlflow/mlflow/pull/5756/files#r890268871

ntakouris:
@harupy We now retain the original exception message too.

harupy (approved):
@ntakouris LGTM, thanks for the contribution!

PR #5759, opened 2022-04-22
SparkML model uses mlflowdbfs to load and log models in the artifact store

liangz1:
Signed-off-by: Liang Zhang <liang.zhang@databricks.com>

SparkML model uses mlflowdbfs to load and save models in the artifact store if mlflow is running on a databricks cluster and mlflowdbfs is available.
Allow users to choose not to use mlflowdbfs via an environment variable: `"DISABLE_MLFLOWDBFS"`. Setting this environment variable to any `not None and not "" (empty string)` value will disable using mlflowdbfs for sparkml model loading and logging.

## How is this patch tested?

TODO.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components

Interface

Language

Integrations


harupy:
@liangz1 I filed https://github.com/mlflow/mlflow/pull/5760 to fix this error:

https://github.com/mlflow/mlflow/runs/6136633995?check_suite_focus=true#step:5:16

```
invalid command name 'min-python-version'
```

mlflow/spark.py
```diff
@@ -235,6 +239,9 @@ def log_model(
     try:
         spark_model.save(posixpath.join(model_dir, _SPARK_MODEL_PATH_SUB))
     except Py4JError:
+        if databricks_utils.is_in_cluster() and databricks_utils.is_mlflowdbfs_available():
+            return _log_model_via_mlflowdbfs(run_root_artifact_uri, artifact_path, spark_model)
```
jinzhang21, line 243: Should we return here w/o calling `Model.log()` below?

mlflow/spark.py
```diff
@@ -235,6 +239,9 @@ def log_model(
     try:
         spark_model.save(posixpath.join(model_dir, _SPARK_MODEL_PATH_SUB))
     except Py4JError:
+        if databricks_utils.is_in_cluster() and databricks_utils.is_mlflowdbfs_available():
```
jinzhang21, line 242: Will MLflowDBFS ever be available outside of Cluster? That is, can we move `databricks_utils.is_in_cluster()` inside `databricks_utils.is_mlflowdbfs_available()`?

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -316,6 +316,24 @@ def _get_run_relative_artifact_path_for_upload(self, src_file_path, dst_artifact
...
+            db_creds.username,
+            db_creds.password,
+        )
+
+    def _clear_credential_context(self):
+        dbutils = _get_dbutils()
+        dbutils.endpoint.clearMlflowProperties()
+
+    def _get_mlflowdbfs_path(self, artifact_path):
+        return "mlflowdbfs:///artifacts?run_id={}&path=/{}".format(self.run_id, artifact_path)
```
jinzhang21, line 335: Should we strip the leading "/" for the input `artifact_path` before using it? Could a user-error be a concern here? How does MLflowDBFS handle paths starting with "//" or "///"?

mlflow/spark.py
liangz1: We still need `Model.log()` if `databricks_utils.is_mlflowdbfs_available() == False`.

mlflow/store/artifact/databricks_artifact_repo.py
liangz1: mlflowdbfs can handle multiple slashes: https://github.com/databricks/runtime/blob/1ba6a8ee9e63226ecf6e5697b16ab43c34eb3831/sql/core/src/test/scala/com/databricks/mlflowdbfs/MlflowdbfsClientSuite.scala#L102-L103
However, I agree that we should check for leading slashes in `artifact_path` since that is not allowed by the definition of this parameter.

tests/spark/autologging/ml/test_pyspark_ml_autologging.py
```diff
@@ -368,7 +368,7 @@ def setModel(self, model: Model):
         def setEvaluator(self, evaluator: Evaluator):
             return self._set(evaluator=evaluator)
 
-        def _fit(self, dataset):
+        def _fit(self, dataset):  # pylint: disable=unused-argument
```
liangz1, line 371: fix pylint for merging master

jinzhang21 (approved):

mlflow/spark.py
dbczumar: `Model.log()` also makes a call to the MLflow Tracking `LogModel` API automatically; https://github.com/mlflow/mlflow/blob/398345c435421ee2f4ab8d8b3237bc845f25a2c6/mlflow/protos/service.proto#L477-L492.

We should still make sure that this API is called, which means we should fall through to https://github.com/mlflow/mlflow/pull/5759/files#diff-c01988c33f2da6ab63864c712c6258450f6dc8fcedf0566507d0ce3b4ac9ed2bR261-R283 after this portion of the logic executes.

We can accomplish this via:

```
    try:
        spark_model.save(posixpath.join(model_dir, _SPARK_MODEL_PATH_SUB))
    except Py4JError:
        if databricks_utils.is_mlflowdbfs_available():
            _log_model_via_mlflowdbfs(run_root_artifact_uri, artifact_path, spark_model)
        else:
            return Model.log(
                artifact_path=artifact_path,
                flavor=mlflow.spark,
                spark_model=spark_model,
                conda_env=conda_env,
                code_paths=code_paths,
                dfs_tmpdir=dfs_tmpdir,
                sample_input=sample_input,
                registered_model_name=registered_model_name,
                signature=signature,
                input_example=input_example,
                await_registration_for=await_registration_for,
                pip_requirements=pip_requirements,
                extra_pip_requirements=extra_pip_requirements,
            )
```

It would also be great to test that we're saving model metadata when mlflowdbfs is available.

dbczumar (changes requested):
@liangz1 Awesome stuff! Still reviewing, but requesting changes now until we address https://github.com/mlflow/mlflow/pull/5759/files#r859320474.

mlflow/spark.py
```diff
@@ -705,6 +740,10 @@ def load_model(model_uri, dfs_tmpdir=None):
         _logger.info("'%s' resolved as '%s'", runs_uri, model_uri)
     flavor_conf = _get_flavor_configuration_from_uri(model_uri, FLAVOR_NAME)
     model_uri = append_to_uri_path(model_uri, flavor_conf["model_data"])
+
+    if databricks_utils.is_in_cluster() and databricks_utils.is_mlflowdbfs_available():
+        return _load_model_via_mlflowdbfs(model_uri)
```
dbczumar, line 745: Looks like we don't add the model's code paths to the Python system path because we're returning before we reach https://github.com/mlflow/mlflow/pull/5759/files#diff-c01988c33f2da6ab63864c712c6258450f6dc8fcedf0566507d0ce3b4ac9ed2bR748.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -316,6 +316,28 @@ def _get_run_relative_artifact_path_for_upload(self, src_file_path, dst_artifact
...
+    def _clear_credential_context(self):
+        dbutils = _get_dbutils()
+        dbutils.endpoint.clearMlflowProperties()
+
+    def _get_mlflowdbfs_path(self, artifact_path):
+        if artifact_path.startswith("/"):
+            raise MlflowException(
+                "artifact_path should be relative, found: {}".format(artifact_path)
+            )
+        return "mlflowdbfs:///artifacts?run_id={}&path=/{}".format(self.run_id, artifact_path)
```
dbczumar, line 339: I don't think these methods should be added to `DatabricksArtifactRepository`, since they're orthogonal (`DatabricksArtifactRepository` is not used for file uploads / downloads when mlflowdbfs is available). Can we put them in `mlflow.utils.databricks_utils` instead?

mlflow/spark.py
```diff
@@ -660,6 +667,34 @@ def _load_model(model_uri, dfs_tmpdir_base=None):
     return PipelineModel.load(model_uri)
 
 
+def _log_model_via_mlflowdbfs(run_root_artifact_uri, artifact_path, model):
+    artifact_repo = get_artifact_repository(artifact_uri=run_root_artifact_uri)
+    mlflowdbfs_path = artifact_repo._get_mlflowdbfs_path(artifact_path)
+    artifact_repo._set_databricks_host_creds_to_credential_context()
```
dbczumar, line 673: See https://github.com/mlflow/mlflow/pull/5759/files#r859321751. I don't think these methods should be included in the artifact repo. This is also technically dangerous if the user has set a different artifact root URI for the MLflow experiment, as these methods wouldn't exist. In fact, we shouldn't be using `mlflowdbfs` at all if the artifact root URI isn't of the form `dbfs:/databricks/mlflow-tracking`.

mlflow/spark.py
```diff
@@ -235,6 +239,9 @@ def log_model(
     try:
         spark_model.save(posixpath.join(model_dir, _SPARK_MODEL_PATH_SUB))
     except Py4JError:
+        if databricks_utils.is_mlflowdbfs_available():
```
dbczumar, line 242: We also need to check that the run's artifact root URI is of the form `dbfs:/databricks/mlflow-tracking`, which we can accomplish by fetching the artifact repo associated with the run and verifying that it's an instance of `DatabricksArtifactRepository`.

If we don't perform this check, model logging will fail in MLR / DBR for experiments with custom artifact root locations, such as `s3://...`, `dbfs:/my/custom/root/locaton`, etc.

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,38 @@ def get_workspace_url():
         return None
 
 
+def is_mlflowdbfs_available():
+    if not is_in_cluster():
+        return False
+
+    # allow users to choose not to use mlflowdbfs via an environment variable
+    disable_mlflowdbfs = os.environ.get("DISABLE_MLFLOWDBFS")
+    if disable_mlflowdbfs is not None and disable_mlflowdbfs != "":
```
dbczumar, line 344:
```suggestion
    if disable_mlflowdbfs:
```

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,38 @@ def get_workspace_url():
...
+
+    from py4j.protocol import Py4JJavaError
+
+    try:
+        spark_session.read.load("mlflowdbfs:///123")
+    except Py4JJavaError as e:
+        if (
+            str(e.java_exception)
+            == "org.apache.hadoop.fs.UnsupportedFileSystemException:"
+            + ' No FileSystem for scheme "mlflowdbfs"'
```
dbczumar, line 363: Comparing against this exact message seems brittle. Perhaps we want to check that the message contains `No FileSystem for scheme "mlflowdbfs"`?

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,38 @@ def get_workspace_url():
         return None
 
 
+def is_mlflowdbfs_available():
+    if not is_in_cluster():
```
dbczumar, line 339: I think `is_in_databricks_runtime` is a better check here.

mlflow/tracking/artifact_utils.py
```diff
@@ -55,11 +55,9 @@ def get_artifact_uri(run_id, artifact_path=None, tracking_uri=None):
 
 # TODO: This would be much simpler if artifact_repo.download_artifacts could take the absolute path
 # or no path.
-def _download_artifact_from_uri(artifact_uri, output_path=None):
+def _get_root_uri_and_artifact_path(artifact_uri):
```
dbczumar, line 58: Nit: Can we add a docstring explaining what this method does?

mlflow/spark.py
liangz1: But we don't have a `local_model_path` if using mlflowdbfs. What path do we want to add here?

mlflow/spark.py
liangz1: Sorry, I don't get the idea. The link above seems to be broken.
The code snippet you suggested does not seem to be calling `Model.log()` if `databricks_utils.is_mlflowdbfs_available()` is true.

dbczumar: Right. We shouldn't call `Model.log()` if `databricls_utils.is_mflowdbfs_available()` is `True`. We should instead fall through to the logic at the bottom of the `log_model()` function that replicates the rest of the operations performed by `Model.log()`.

mlflow/spark.py
dbczumar: We want to load any files referenced by the `code` attribute of the `pyfunc` flavor in the `MLmodel` file. These should be downloaded separately and imported before the Spark model is resolved with `mlflowdbfs`.

mlflow/spark.py
liangz1: Oh, I see! Thanks!

mlflow/spark.py
liangz1: I moved
```python
    local_model_path = _download_artifact_from_uri(model_uri)
    _add_code_from_conf_to_system_path(local_model_path, flavor_conf)
```
up. Is this what you mean?

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,57 @@ def get_workspace_url():
...
+
+    try:
+        spark_session = _get_active_spark_session()
+    except Exception:
+        return False
+
+    if spark_session is None:
+        return False
+
+    from py4j.protocol import Py4JJavaError
```
dbczumar, line 355: Py4J import should be moved inside the `try/catch` below; we should also catch `ImportError` just in case.

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,57 @@ def get_workspace_url():
...
+
+    try:
+        spark_session.read.load("mlflowdbfs:///123")
+    except Py4JJavaError as e:
+        if str(e.java_exception) == 'No FileSystem for scheme "mlflowdbfs"':
+            return False
+
+        return True
```
dbczumar, line 363: We should determine whether or not there's a higher-fidelity way to check for the presence of `mlflowdbfs`. This approach where we perform case-sensitive string matching on exception content is a bit brittle.

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,57 @@ def get_workspace_url():
         return None
```
dbczumar, line 335: New methods in this module need docstrings.

mlflow/spark.py
```diff
@@ -660,6 +673,34 @@ def _load_model(model_uri, dfs_tmpdir_base=None):
     return PipelineModel.load(model_uri)
 
 
+def _log_model_via_mlflowdbfs(artifact_repo, artifact_path, model):
+    mlflowdbfs_path = databricks_utils.get_mlflowdbfs_path(artifact_repo.run_id, artifact_path)
+    databricks_utils.set_databricks_host_creds_to_credential_context(
```
dbczumar, line 678: Nit: `set_databricks_host_creds_to_credential_context` phrasing is a bit awkward. Let's rename this method. Perhaps `add_databricks_host_creds_to_credential_context`.

mlflow/spark.py
```diff
@@ -660,6 +673,34 @@ def _load_model(model_uri, dfs_tmpdir_base=None):
...
+        artifact_repo.databricks_profile_uri
+    )
+
+    try:
+        model = PipelineModel.load(mlflowdbfs_path)
+    finally:
+        databricks_utils.clear_credential_context()
+
+    return model
```
dbczumar, line 701: These methods need docstrings, and we should document the assumptions that these methods make (i.e. mlflowdbfs is available *and* the artifact location associated with the Run lies within `dbfs:/databricks/mlflow-tracking`).

mlflow/spark.py
```diff
@@ -235,21 +240,29 @@ def log_model(
...
-            pip_requirements=pip_requirements,
-            extra_pip_requirements=extra_pip_requirements,
-        )
+        artifact_repo = get_artifact_repository(artifact_uri=run_root_artifact_uri)
+        if (
+            isinstance(artifact_repo, DatabricksArtifactRepository)
+            and databricks_utils.is_mlflowdbfs_available()
+        ):
+            # TODO: test we're saving model metadata when mlflowdbfs is available.
+            _log_model_via_mlflowdbfs(artifact_repo, artifact_path, spark_model)
```
dbczumar, line 249: If this procedure fails, we should raise a clear and actionable failure message. Rather than nesting try / catch blocks, I propose that we move this code out of the exception handler, since, `isinstance(artifact_repo, DatabricksArtifactRepository) == True` implies that `run_root_artifact_uri` is of the form `dbfs:/databricks/mlflow-tracking/...`, which can't be written to via `spark_model.save(model_dir)` anyway. The current try / except should be moved to an `else` block.

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,57 @@ def get_workspace_url():
...
+
+
+def clear_credential_context():
+    dbutils = _get_dbutils()
+    dbutils.endpoint.clearMlflowProperties()
+
+
+def get_mlflowdbfs_path(run_id, artifact_path):
+    if artifact_path.startswith("/"):
+        raise MlflowException("artifact_path should be relative, found: {}".format(artifact_path))
```
dbczumar, line 385: `error_code=INVALID_PARAMETER_VALUE` should be specified.

mlflow/spark.py
```diff
@@ -235,21 +240,29 @@ def log_model(
...
-            await_registration_for=await_registration_for,
-            pip_requirements=pip_requirements,
-            extra_pip_requirements=extra_pip_requirements,
-        )
+        artifact_repo = get_artifact_repository(artifact_uri=run_root_artifact_uri)
+        if (
+            isinstance(artifact_repo, DatabricksArtifactRepository)
+            and databricks_utils.is_mlflowdbfs_available()
+        ):
+            # TODO: test we're saving model metadata when mlflowdbfs is available.
```
dbczumar, line 248: Let's make sure to add this test coverage by mocking the presence / behavior of `mlflowdbfs`.

mlflow/utils/databricks_utils.py
```diff
@@ -335,6 +335,57 @@ def get_workspace_url():
...
 
 
+def is_mlflowdbfs_available():
+    if not is_in_databricks_runtime():
+        return False
+
+    # allow users to choose not to use mlflowdbfs via an environment variable
+    disable_mlflowdbfs = os.environ.get("DISABLE_MLFLOWDBFS")
+    if disable_mlflowdbfs:
+        return False
```
dbczumar, line 345: We should make sure to test that this environment variable is respected.

mlflow/tracking/artifact_utils.py
```diff
@@ -55,11 +55,10 @@ def get_artifact_uri(run_id, artifact_path=None, tracking_uri=None):
...
     """
+    Parse the artifact_uri to get the root_uri and artifact_path.
     :param artifact_uri: The *absolute* URI of the artifact to download.
```
dbczumar, line 61: param docstring is out of date - this method doesn't download anything

apurva-koti:
Merged master to enable iteration post protobuf-fix.

PR #5826, opened 2022-05-07
Spark Datasource autologging: Fix REPL ID propagation from datasource listener to publisher

dbczumar:

This PR addresses two issues with `ReplAwareSparkDataSourceListener`:

1. When multiple tables are read as part of the same SparkSQL operation, REPL ID information is only passed to the first call to `publisher.publishEvent` corresponding to the first table.

2. When the REPL ID is missing, `publisher.publishEvent` is still called and subscribers are still notified of the event, even though we emit a warning log indicating that the event will be dropped

## How is this patch tested?

Improved integration test coverage

   next step, otherwise fix it.
3. Click `Details` on the right to open the job page of CircleCI.
4. Click the `Artifacts` tab.
5. Click `docs/build/html/index.html`.
6. Find the changed pages / sections and make sure they render correctly.

## Release Notes

Fix a bug in Spark Datasource autologging on Databricks that caused table reads initiated by other users on the same cluster to be included in datasource tags.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/ReplAwareSparkDataSourceListener.scala
+++ b/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/ReplAwareSparkDataSourceListener.scala
@@ -50,7 +50,7 @@ class ReplAwareSparkDataSourceListener(
     val replIdOpt = popReplIdOpt(event)
     if (replIdOpt.isDefined) {
       tableInfos.foreach { tableInfo =>
-        publisher.publishEvent(replIdOpt, sparkTableInfo = tableInfo)
+        publisher.publishEvent(replIdOpt = replIdOpt, sparkTableInfo = tableInfo)
       }
     }
   }
--- a/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/ReplAwareSparkDataSourceListener.scala
+++ b/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/ReplAwareSparkDataSourceListener.scala
@@ -44,7 +44,7 @@ class ReplAwareSparkDataSourceListener(
     executionIdToReplId.put(executionId, replIdOpt.get)
   }
 
-  override protected def getReplIdOpt(event: SparkListenerSQLExecutionEnd): Option[String] = {
+  override protected def popReplIdOpt(event: SparkListenerSQLExecutionEnd): Option[String] = {
     executionIdToReplId.remove(event.executionId)
   }
 }
--- a/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/ReplAwareSparkDataSourceListener.scala
+++ b/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/ReplAwareSparkDataSourceListener.scala
@@ -44,7 +44,18 @@ class ReplAwareSparkDataSourceListener(
     executionIdToReplId.put(executionId, replIdOpt.get)
   }
 
-  override protected def popReplIdOpt(event: SparkListenerSQLExecutionEnd): Option[String] = {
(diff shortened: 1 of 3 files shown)


mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/SparkDataSourceListener.scala
```diff
@@ -17,14 +17,12 @@ class SparkDataSourceListener(
     DatasourceAttributeExtractor
   }
 
-  protected def getReplIdOpt(event: SparkListenerSQLExecutionEnd): Option[String] = None
```
dbczumar, line 20: Removing `getRepldOpt` from `SparkDataSourceListener` because 1. REPL ID fetching only applies to `ReplAwareSparkDataSourceListener`, so `SparkDatasourceListener` should *never* have a REPL ID to propagate and 2. `ReplAwareSparkDataSourceListener` should *always* have a REPL ID to propagate.

Previously, the logic in `ReplAwareSparkDatasourceListener.onSQLExecutionEnd`, which was directly inherited from `SparkDatasourceListener.onSQLExecutionEnd` erroneously published events when the REPL ID was missing, and these events were actually delivered to subscribers because of logic here: https://github.com/mlflow/mlflow/blob/0a88eabb031b4b7083476dd2de617356c9c35e81/mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/MlflowAutologEventPublisher.scala#L154.

Accordingly, we now override `onSQLExecutionEnd` in `ReplAwareSparkDataSourceListener` to handle the uncommon (shouldn't ever happen?) case where REPL ID is missing.

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/ReplAwareSparkDataSourceListener.scala
```diff
@@ -44,7 +44,18 @@ class ReplAwareSparkDataSourceListener(
...
   }
 
-  override protected def getReplIdOpt(event: SparkListenerSQLExecutionEnd): Option[String] = {
+  protected[autologging] override def onSQLExecutionEnd(event: SparkListenerSQLExecutionEnd): Unit = {
+    val extractor = getDatasourceAttributeExtractor
+    val tableInfos = extractor.getTableInfos(event)
+    val replIdOpt = popReplIdOpt(event)
+    if (replIdOpt.isDefined) {
+      tableInfos.foreach { tableInfo =>
+        publisher.publishEvent(replIdOpt = replIdOpt, sparkTableInfo = tableInfo)
```
dbczumar, line 53: Fetch the REPL ID once and, only if it is defined, publish table read events to the subscriber(s) associated with the REPL

dbczumar:
@tleyden Can you take a look as well?

mlflow/java/spark/src/main/scala/org/mlflow/spark/autologging/SparkDataSourceListener.scala
smurching: Thanks for the detailed explanation! And 

> Accordingly, we now override onSQLExecutionEnd in ReplAwareSparkDataSourceListener to handle the uncommon (shouldn't ever happen?) case where REPL ID is missing.

This sounds good :) IIRC it makes sense to not throw in the case where REPL ID is missing as that can actually break Spark job execution (which we don't want autologging to do). We could log a warning to facilitate debugging but that doesn't seem quite as critical to add in this PR, since it looks like we do already log a warning if the repl ID is missing in `onJobStart`

smurching (approved):
LGTM!

tleyden (commented):
LGTM and +1 to the rename to `popReplIdOpt()` - that is much more accurate w.r.t the method's behavior.

PR #5867, opened 2022-05-13
Update CONTRIBUTING.rst

timcharper:
Add a helpful note about --legacy-peer-devs

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -245,7 +245,7 @@ dependencies via:
 .. code-block:: bash
 
    cd mlflow/server/js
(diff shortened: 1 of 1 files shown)


harupy:
@timcharper Thanks for the PR, but we no longer use `npm`. We currently use `yarn` so `CONTRIBUTING.rst` needs to be updated.

harupy:
https://stackoverflow.com/questions/66239691/what-does-npm-install-legacy-peer-deps-do-exactly-when-is-it-recommended-wh#comment126876429_66620869 says we don't need this flag for yarn.

PR #5920, opened 2022-05-20
Add Shap kernel explainer support for DefaultEvaluator

WeichenXu123:
## Related Issues/PRs

#xxx

Add Shap kernel explainer support for DefaultEvaluator.

The kernel explainer determine the impact of a feature by using "background" data to replace original feature data. So that it can support categorical feature data without encoder.

**In mlflow pipeline evaluation step, I suggest to use the "kernel" explainer which can handle categorical features easily**

## How is this patch tested?

Unit tests.

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Add Shap kernel explainer support for DefaultEvaluator.

Components

Interface

Language

Integrations


mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +499,49 @@ def _log_model_explainability(self):
...
-                self.predict_fn,
-                sampled_X,
-                feature_names=truncated_feature_names,
-                algorithm=algorithm,
-            )
+            if algorithm == "kernel":
+                # predict_fn = lambda x: self.predict_fn(pd.DataFrame(x, columns=sampled_X.columns))
+                mode = sampled_X.mode().iloc[0]
+                sampled_X = sampled_X.fillna(mode)
+                background_X = background_X.fillna(mode)
```
WeichenXu123, line 530: Shall we replace NA values to be most frequent values ?
https://github.com/databricks/mlflow-private/blob/25b605e498046982004b3bcda1828d444184989c/mlflow/pipelines/regression/v1/steps/evaluate.py#L70
The code snippet here does this. Is this a common way ?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +499,49 @@ def _log_model_explainability(self):
...
+                explainer = shap.KernelExplainer(self.predict_fn, background_X, link="identity")
```
WeichenXu123, line 531: The link function has other options, shall we add config for this ?

mlflow/models/evaluation/default_evaluator.py
BenWilson2: For ordinal data the `mode` approach definitely makes sense (it's what I would do). For continuous data, I would use the `mean` iif the data is normally distributed. I would fill the `mean` after scaling / normalizing (e.g. log scaling) for any distribution that isn't normally distributed. 
That being said, I've never tried to use SHAP with missing feature data. Imputing data is effectively what SHAP is doing in the kernel explainer but only for the mutation vector being used to get the differencing value between real vs. synthetic vectors. Because of this, the probability of having a sparsely populated feature within a dataset be populated with a limited number of identical values can render the evaluation grossly biased. 

TL&DR; I think `mode` is safe for `integer` but I would use `mean` for `float`.

mlflow/models/evaluation/default_evaluator.py
BenWilson2: I would definitely add the ability to configure this to use `logit`. `identity` is useless with LogisticRegression or any other model that relies on probability thresholds as their output.

tests/models/test_default_evaluator.py
```diff
@@ -294,6 +296,28 @@ def test_svm_classifier_evaluation(svm_model_uri, breast_cancer_dataset):
...
+            evaluators="default",
+            evaluator_config={"explainability_algorithm": "kernel"},
+        )
+    _, _, _, artifacts = get_run_data(run.info.run_id)
+    for shap_plot in [
+        "shap_beeswarm_plot_on_data_pipeline_model_dataset.png",
+        "shap_feature_importance_plot_on_data_pipeline_model_dataset.png",
+        "shap_summary_plot_on_data_pipeline_model_dataset.png",
+    ]:
+        assert shap_plot in artifacts
```
harupy, line 318:
```suggestion
    assert [
        "shap_beeswarm_plot_on_data_pipeline_model_dataset.png",
        "shap_feature_importance_plot_on_data_pipeline_model_dataset.png",
        "shap_summary_plot_on_data_pipeline_model_dataset.png",
    ] == artifacts
```

harupy: or

```suggestion
    assert {
        "shap_beeswarm_plot_on_data_pipeline_model_dataset.png",
        "shap_feature_importance_plot_on_data_pipeline_model_dataset.png",
        "shap_summary_plot_on_data_pipeline_model_dataset.png",
    }.issubset(artifacts)
```

tests/models/test_evaluation.py
```diff
@@ -136,6 +138,57 @@ def breast_cancer_dataset():
...
+    return data, "y"
+
+
+@pytest.fixture(scope="module")
+def pipeline_model_dataset():
+    data, target_col = get_pipeline_model_dataset()
+    constructor_args = {"data": data[0::3], "targets": target_col, "name": "pipeline_model_dataset"}
+    ds = EvaluationDataset(**constructor_args)
+    ds._constructor_args = constructor_args
+    return ds
```
harupy, line 168: Can we remove this fixture and directly use `get_pipeline_model_dataset` in `test_pipeline_model_kernel_explainer_on_categorical_features`?

tests/models/test_evaluation.py
```diff
@@ -136,6 +138,57 @@ def breast_cancer_dataset():
...
+            ("clf", sklearn.linear_model.LogisticRegression()),
+        ]
+    )
+    pipeline.fit(X, y)
+
+    with mlflow.start_run() as run:
+        mlflow.sklearn.log_model(pipeline, "pipeline_model")
+        pipeline_model_uri = get_artifact_uri(run.info.run_id, "pipeline_model")
+
+    return pipeline_model_uri
```
harupy, line 189:
```suggestion
    with mlflow.start_run() as run:
        model_info = mlflow.sklearn.log_model(pipeline, "pipeline_model")
        return model_info.model_uri
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +499,51 @@ def _log_model_explainability(self):
             f: f2 for f, f2 in zip(self.feature_names, truncated_feature_names)
         }
 
-        sampled_X = shap.sample(self.X, sample_rows)
+        sampled_X = shap.sample(self.X, sample_rows, random_state=0)
+        if algorithm == "kernel":
+            background_X = shap.sample(self.X, sample_rows, random_state=3)
```
harupy, line 504: `sampled_x` and `background_X` may have overlapping rows. Is that fine?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +499,51 @@ def _log_model_explainability(self):
...
+        else:
+            # If sample_X is numpy array, convert to pandas dataframe.
+            sampled_X = pd.DataFrame(sampled_X, columns=truncated_feature_name_map)
+            if background_X is not None:
+                background_X = pd.DataFrame(background_X, columns=truncated_feature_name_map)
 
         if algorithm:
-            supported_algos = ["exact", "permutation", "partition"]
+            supported_algos = ["exact", "permutation", "partition", "kernel"]
             if algorithm not in supported_algos:
```
harupy, line 522: We might want to do this validation right after `self.evaluator_config.get("explainability_algorithm", None)`.

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: I think it is fine.

mlflow/models/evaluation/default_evaluator.py
jinzhang21: +1. supported_algos can be a module level constant like `_SUPPORTED_SHAP_ALGORITHMS = ("exact", "permutation", "partition", "kernel")`

jinzhang21 (approved):
LGTM!

dbczumar (approved):
LGTM once https://github.com/mlflow/mlflow/pull/5920/files#r879598221 is addressed. Thanks @WeichenXu123 !

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -370,6 +370,33 @@ def __validate_artifacts(artifacts):
...
+            float_columns.append(col_name)
+        else:
+            non_float_column.append(col_name)
+
+    means = df[float_columns].mean().to_dict()
+    modes = df[non_float_column].mode().loc[0].to_dict()
+    return {**means, **modes}
+
+
+_SUPPORTED_SHAP_ALGORITHMS = ["exact", "permutation", "partition", "kernel"]
```
harupy, line 397:
```suggestion
_SUPPORTED_SHAP_ALGORITHMS = ("exact", "permutation", "partition", "kernel")
```

Let's use a tuple here.

tests/models/test_evaluation.py
```diff
@@ -136,6 +138,46 @@ def breast_cancer_dataset():
...
+
+    pipeline = sklearn.pipeline.Pipeline(
+        [
+            ("indexer", sklearn.preprocessing.OrdinalEncoder()),
+            ("clf", sklearn.linear_model.LogisticRegression()),
+        ]
+    )
+    pipeline.fit(X, y)
+
+    with mlflow.start_run() as run:
```
harupy, line 176:
```suggestion
    with mlflow.start_run():
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +531,50 @@ def _log_model_explainability(self):
...
+                if kernel_link not in ["identity", "logit"]:
+                    raise ValueError(
+                        f"explainability_kernel_link config can only be set to 'identity' or 'logit'."
+                    )
+                mode_or_mean_dict = _compute_df_mode_or_mean(self.X)
+                mode_or_mean_dict = {truncated_feature_name_map[k]: v for k, v in mode_or_mean_dict}
+                sampled_X = sampled_X.fillna(mode_or_mean_dict)
+                background_X = background_X.fillna(mode_or_mean_dict)
+                predict_fn = lambda x: self.predict_fn(pd.DataFrame(x, columns=sampled_X.columns))
+                explainer = shap.KernelExplainer(predict_fn, background_X, link=kernel_link)
```
harupy, line 564:
```suggestion
                explainer = shap.KernelExplainer(self.predict_fn, background_X, link=kernel_link)
```

What would happen if we directly pass `self.predict_fn`?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +370,33 @@ def __validate_artifacts(artifacts):
...
+    float_columns = []
+    non_float_column = []
+    for col_name in columns:
+        if df.dtypes[col_name].name.lower().startswith("float"):
+            float_columns.append(col_name)
+        else:
+            non_float_column.append(col_name)
+
+    means = df[float_columns].mean().to_dict()
+    modes = df[non_float_column].mode().loc[0].to_dict()
```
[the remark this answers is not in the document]
WeichenXu123: Sounds a good idea.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +370,33 @@ def __validate_artifacts(artifacts):
     )
 
 
+def _compute_df_mode_or_mean(df):
+    """
+    Compute mode (for integer or non-numeric columns or mean (for float columns) for the
+    input dataframe, return a dict, key is column name, value is the corresponding mode or
+    mean value.
+    """
+    # convert pandas dataframe columns to be the best type.
```
harupy, line 379:
```suggestion
    # convert pandas dataframe columns to best possible dtypes.
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +522,50 @@ def _log_model_explainability(self):
...
-            supported_algos = ["exact", "permutation", "partition"]
-            if algorithm not in supported_algos:
-                raise ValueError(
-                    f"Specified explainer algorithm {algorithm} is unsupported. Currently only "
-                    f"support {','.join(supported_algos)} algorithms."
+            if algorithm == "kernel":
+                kernel_link = self.evaluator_config.get("explainability_kernel_link", "identity")
+                if kernel_link not in ["identity", "logit"]:
+                    raise ValueError(
+                        f"explainability_kernel_link config can only be set to 'identity' or 'logit'."
```
harupy, line 548: Can we include the value of `kernel_link` in the error message?

tests/models/test_evaluation.py
```diff
@@ -136,6 +138,46 @@ def breast_cancer_dataset():
...
+    """
+    X, y = get_iris()
+
+    def convert_num_to_label(x):
+        return f"v_{round(x)}"
+
+    data = pd.DataFrame(
+        {
+            "f1": np.array(map(convert_num_to_label, X[:, 0])),
+            "f2": np.array(map(convert_num_to_label, X[:, 1])),
```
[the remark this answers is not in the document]
WeichenXu123: The `OrdinalEncoder` in the pipeline force to convert all columns values into index.
Do you know a better transformer to replace it ?

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: If so,
On the `mlflow pipeline evaluate` example test data, one error will be raised:
```
  File "/Users/weichen.xu/work/mlflow-private/mlflow/models/evaluation/default_evaluator.py", line 555, in _log_model_explainability
    explainer = shap.KernelExplainer(self.predict_fn, background_X, link=kernel_link)
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/shap/explainers/_kernel.py", line 69, in __init__
    model_null = match_model_to_data(self.model, self.data)
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/shap/utils/_legacy.py", line 112, in match_model_to_data
    out_val = model.f(data.data)
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/sklearn/utils/metaestimators.py", line 113, in <lambda>
    out = lambda *args, **kwargs: self.fn(obj, *args, **kwargs)  # noqa
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/sklearn/pipeline.py", line 469, in predict
    Xt = transform.transform(Xt)
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/sklearn/compose/_column_transformer.py", line 748, in transform
    Xs = self._fit_transform(
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/sklearn/compose/_column_transformer.py", line 606, in _fit_transform
    return Parallel(n_jobs=self.n_jobs)(
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/joblib/parallel.py", line 1043, in __call__
    if self.dispatch_one_batch(iterator):
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/joblib/parallel.py", line 833, in dispatch_one_batch
    islice = list(itertools.islice(iterator, big_batch_size))
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/sklearn/compose/_column_transformer.py", line 609, in <genexpr>
    X=_safe_indexing(X, column, axis=1),
  File "/Users/weichen.xu/opt/miniconda3/envs/py39/lib/python3.9/site-packages/sklearn/utils/__init__.py", line 370, in _safe_indexing
    raise ValueError(
ValueError: Specifying the columns using strings is only supported for pandas DataFrames
```

The error is caused by: the `predict_fn` passed to "shap.KernelExpainer" was called with a "Numpy array" type argument.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +522,51 @@ def _log_model_explainability(self):
...
 
         if isinstance(sampled_X, pd.DataFrame):
             # For some shap explainer, the plot will use the DataFrame column names instead of
             # using feature_names argument value. So rename the dataframe column names.
             sampled_X = sampled_X.rename(columns=truncated_feature_name_map, copy=False)
+            if background_X is not None:
+                background_X = background_X.rename(columns=truncated_feature_name_map, copy=False)
+        else:
+            # If sample_X is numpy array, convert to pandas dataframe.
+            sampled_X = pd.DataFrame(sampled_X, columns=truncated_feature_name_map)
```
harupy, line 539: `truncated_feature_name_map` is a dict, right? What does `sampled_X.columns` look like if `truncated_feature_name_map` is `{"a": "b", "c": "d"}`?

harupy: Quick test:


```python
>>> pd.DataFrame([[1,2], [3,4]], columns={"a": "b", "c": "d"}).columns
Index(['a', 'c'], dtype='object')
```

Is this the desired result?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +522,51 @@ def _log_model_explainability(self):
...
-            if algorithm not in supported_algos:
-                raise ValueError(
-                    f"Specified explainer algorithm {algorithm} is unsupported. Currently only "
-                    f"support {','.join(supported_algos)} algorithms."
+            if algorithm == "kernel":
+                kernel_link = self.evaluator_config.get("explainability_kernel_link", "identity")
+                if kernel_link not in ["identity", "logit"]:
+                    raise ValueError(
+                        "explainability_kernel_link config can only be set to 'identity' or "
+                        f"'logit', but get value '{kernel_link}'."
```
harupy, line 549:
```suggestion
                        f"'logit', but got '{kernel_link}'."
```

tests/models/test_evaluation.py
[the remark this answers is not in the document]
harupy: https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html#sklearn.compose.ColumnTransformer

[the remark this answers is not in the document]
harupy: https://scikit-learn.org/stable/modules/generated/sklearn.compose.make_column_transformer.html

mlflow/models/evaluation/default_evaluator.py
harupy: got it 👍

tests/models/test_default_evaluator.py
```diff
@@ -294,6 +297,42 @@ def test_svm_classifier_evaluation(svm_model_uri, breast_cancer_dataset):
...
+            dataset_name="pipeline_model_dataset",
+            evaluators="default",
+            evaluator_config={"explainability_algorithm": "kernel"},
+        )
+    _, _, _, artifacts = get_run_data(run.info.run_id)
+    assert {
+        "shap_beeswarm_plot_on_data_pipeline_model_dataset.png",
+        "shap_feature_importance_plot_on_data_pipeline_model_dataset.png",
+        "shap_summary_plot_on_data_pipeline_model_dataset.png",
+    }.issubset(artifacts)
```
harupy, line 317:
```suggestion
    run_data = get_run_data(run.info.run_id)
    assert {
        "shap_beeswarm_plot_on_data_pipeline_model_dataset.png",
        "shap_feature_importance_plot_on_data_pipeline_model_dataset.png",
        "shap_summary_plot_on_data_pipeline_model_dataset.png",
    }.issubset(run_data.artifacts)
```

nit: this might be cleaner

mlflow/models/evaluation/base.py
```diff
@@ -719,11 +719,14 @@ def evaluate(
...
         - **explainability_algorithm**: A string to specify the SHAP Explainer algorithm for model
-          explainability. Supported algorithm includes: 'exact', 'permutation', 'partition'.
+          explainability. Supported algorithm includes: 'exact', 'permutation', 'partition',
+          'kernel'.
           If not set, ``shap.Explainer`` is used with the "auto" algorithm, which chooses the best
           Explainer based on the model.
         - **explainability_nsamples**: The number of sample rows to use for computing model
           explainability insights. Default value is 2000.
+        - **explainability_kernel_link**: The kernel link function used by shap kernal explainer.
+          Avalable values are "identity" and "logit". Default value is "identity".
```
harupy, line 729:
```suggestion
          Available values are "identity" and "logit". Default value is "identity".
```

typo

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +370,24 @@ def __validate_artifacts(artifacts):
...
+def _compute_df_mode_or_mean(df):
+    """
+    Compute mode (for integer or non-numeric columns or mean (for float columns) for the
+    input dataframe, return a dict, key is column name, value is the corresponding mode or
+    mean value.
+    """
+    # convert pandas dataframe columns to best possible dtypes.
+    # e.g. if one column has float type values, but they are actually all integers,
+    # it will be converted to integer column
+    df = df.convert_dtypes()
```
harupy, line 382: Do we really need this conversion? Let's say we have a float feature and it only contains integers (e.g. `[1.0, 2.0, 5.0]`)) by coincidence. Should we really use mode for this feature?

WeichenXu123: This case I think we should treat it as "integer" feature. The model might use this feature as categorical feature, but if we compute "mean", then we might generate a number which is not integer, then it might break the model prediction.

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: My fault. Fix it. Should use `truncated_feature_names` here.

mlflow/models/evaluation/default_evaluator.py
BenWilson2: I think most model infrastructure that does dynamic casting internally uses a cardinality check, right? Would it be better to handle the mean / mode validation for numeric types by using a similar approach?

BenWilson2: I think the one that you implemented for determining if a model is a classifier or a regressor for the evaluator API will work great!

harupy (approved):

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +380,31 @@ def __validate_artifacts(artifacts):
...
+    input dataframe, return a dict, key is column name, value is the corresponding mode or
+    mean value.
+    """
+    # convert pandas dataframe columns to best possible dtypes.
+    # e.g. if one column has float type values, but they are actually all integers,
+    # it will be converted to integer column
+    categorical_cols = []
+    continuous_cols = []
+    for col in df.columns:
+        if _is_categorical_values(df[col].tolist()):
```
harupy, line 395:
```suggestion
        if _is_categorical_values(df[col]):
```

Do we need `tolist()`?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -41,20 +42,30 @@
 _DEFAULT_SAMPLE_ROWS_FOR_SHAP = 2000
 
 
-def _infer_model_type_by_labels(labels):
+def _is_categorical_values(labels):
```
harupy, line 45: Can we call this function `_is_categorical_feature`?

```suggestion
def _is_categorical_feature(values):
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +380,31 @@ def __validate_artifacts(artifacts):
     )
 
 
+def _compute_df_mode_or_mean(df):
+    """
+    Compute mode (for integer or non-numeric columns or mean (for float columns) for the
+    input dataframe, return a dict, key is column name, value is the corresponding mode or
+    mean value.
+    """
```
harupy, line 388: Can we update the docstring?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +380,31 @@ def __validate_artifacts(artifacts):
...
+    # convert pandas dataframe columns to best possible dtypes.
+    # e.g. if one column has float type values, but they are actually all integers,
+    # it will be converted to integer column
```
harupy, line 391: Do we need this comment?

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: This method is also called from `_infer_model_type_by_labels`, used to check labels type.

harupy: ok, can we just call `_is_categorical`?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +380,29 @@ def __validate_artifacts(artifacts):
     )
 
 
+def _compute_df_mode_or_mean(df):
+    """
+    Compute mode (for categorical columns or mean (for continuous columns) for the
```
harupy, line 385:
```suggestion
    Compute mode (for categorical columns) or mean (for continuous columns) for the
```

tests/models/test_evaluation.py
```diff
@@ -136,6 +138,46 @@ def breast_cancer_dataset():
     return ds
 
 
+def get_pipeline_model_dataset():
+    """
+    The dataset tweaks the IRIS dataset by changing its features into categorical features.
+    The dataset is prepared for a pipeline model
+    (contains an indexer on feature "f1" and a classifier).
```
harupy, line 145:
```suggestion
    The dataset is prepared for a pipeline model that consists of a transformer indexing feature f1 and f2, and a classifier.
```

tests/models/test_evaluation.py
```diff
@@ -136,6 +140,80 @@ def breast_cancer_dataset():
...
+            "y": y,
+        }
+    )
+    return data, "y"
+
+
+@pytest.fixture
+def pipeline_model_uri():
+    """
+    Create a pipeline model that transforming and training on dataset returned by
```
harupy, line 179:
```suggestion
    Create a pipeline model that transforms and trains on dataset returned by
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -41,20 +42,30 @@
...
-    if len(distinct_labels) > 1000 and len(distinct_labels) / len(labels) > 0.7:
-        return "regressor"
-    if len(distinct_labels) <= 20:
-        return "classifier"
+            return False
+    if len(distinct_values) > 1000 and len(distinct_values) / len(values) > 0.7:
+        return False
+    if len(distinct_values) <= 20:
+        return True
     return None  # unknown
```
jinzhang21, line 56: What does None mean here? Can we just return either True or False?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +380,29 @@ def __validate_artifacts(artifacts):
     )
 
 
+def _compute_df_mode_or_mean(df):
+    """
+    Compute mode (for categorical columns) or mean (for continuous columns) for the
+    input dataframe, return a dict, key is column name, value is the corresponding mode or
+    mean value, this function calls `_is_categorical_values` to determine whether the
```
jinzhang21, line 387: calls `_is_categorical` ...

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -41,20 +42,30 @@
 _DEFAULT_SAMPLE_ROWS_FOR_SHAP = 2000
 
 
-def _infer_model_type_by_labels(labels):
-    distinct_labels = set(labels)
-    for v in distinct_labels:
+def _is_categorical(values):
+    distinct_values = set(values)
```
jinzhang21, line 46: Instead of doing manual inspection, can we use Pandas to determine whether it's categorical first? https://pandas.pydata.org/docs/user_guide/categorical.html#categorical-data

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -41,20 +42,30 @@
...
+            return True
         if not float(v).is_integer():
-            return "regressor"
-    if len(distinct_labels) > 1000 and len(distinct_labels) / len(labels) > 0.7:
-        return "regressor"
-    if len(distinct_labels) <= 20:
-        return "classifier"
+            return False
+    if len(distinct_values) > 1000 and len(distinct_values) / len(values) > 0.7:
+        return False
```
jinzhang21, line 53: How is this criteria determined? If I have a word embedding could it be classified as continuous?

mlflow/models/evaluation/default_evaluator.py
harupy: +1

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -494,28 +537,57 @@ def _log_model_explainability(self):
...
+                if kernel_link not in ["identity", "logit"]:
+                    raise ValueError(
+                        "explainability_kernel_link config can only be set to 'identity' or "
+                        f"'logit', but got '{kernel_link}'."
+                    )
+                mode_or_mean_dict = _compute_df_mode_or_mean(self.X)
+                mode_or_mean_dict = {
+                    truncated_feature_name_map[k]: v for k, v in mode_or_mean_dict.items()
+                }
+                sampled_X = sampled_X.fillna(mode_or_mean_dict)
```
harupy, line 570: What would happen if the imputation here is different from the imputation the model performs?

harupy: For example, let's say we have a pipeline model that contains an imputer. What would happen if the imputer just fills NaN with `0`? Would this lead to incorrect feature importance?

harupy: btw do we need imputation when the model contains an imputer?

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: None means we cannot decide whether it is categorical or continuous.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -41,20 +42,30 @@
...
 
 
+def _infer_model_type_by_labels(labels):
+    result = _is_categorical(labels)
+    if result is None:
+        return None  # unknown
+    elif result:
+        return "classifier"
+    else:
+        return "regressor"
```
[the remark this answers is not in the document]
WeichenXu123: Yes, it is used here. https://github.com/mlflow/mlflow/blob/b2687ccb5af1ed898a5bc2098008bb730b7d3119/mlflow/models/evaluation/default_evaluator.py#L917

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: It is empirical.

> If I have a word embedding could it be classified as continuous

If a component column of a word embedding contains values all integers, it might be regarded as categorical here.

The logic is not ideal, for uncertain cases, we will treat it as categorical and compute "mode" instead of "mean".

But, I suggest, shall we let user provide information whether the feature column is categorical or continuous ?
There's no good rules to precisely infer this from feature values.

CC @BenWilson2 WDYT ?

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: Yes, it might affect feature importance, But, I think this is a minor issue. Because the kernel explainer runs in approximate way, and we sample rows as the "background" data, the different sampling will also affect the result.

harupy: @WeichenXu123 do we need to impute sampled_X when the model contains an imputer? SHAP doesn't work when sampled_X contains missing values?

WeichenXu123: For KernelExplainer, shap can still work with NA values if we don't impute sampled_X and background_X.
Because KernelExplainer it won't transform the provided feature values before it feeds feature values into the model.
Rethink: do we still need imputation preprocessing for shap KernelExplainer ?

I link @BenWilson2 's old comment about this.
https://github.com/mlflow/mlflow/pull/5920#discussion_r879598221

harupy: btw why did you think imputation is required?

BenWilson2: I think this brings up 3 separate discussion points...
1. Does the model handle missing data? 
    For models like XGBoost / LightGBM which are quite prevalent in usage for supervised learning tasks, an assumption by many is made that data is imputed. It is not. NaN / Null / NA values are all tagged as a 'special internal value' that can be split on (i.e. at a decision branch node, the logic may be 'col x is NA -> Left ; col x is not NA -> Right'. It doesn't fill in zeros or use mean values. 
    For linear models, imputation is absolutely needed and, if not explicitly handled by an imputer stage transformer (which is definitely a best practice but also one that is not universally performed by practitioners), will default to the modeling package implementation (consistency in behavior is not guaranteed).
2. SHAP handles missing values in a rather complex way. If the column under evaluation is evaluated as 'missing at random' (low counts of missing feature vector positions in a global sense), the local evaluation being performed will toggle that feature, resulting in a SHAP value of effectively 0 for that feature of that vector evaluation. Global missing throughout a vector position will essentially behave as a universal feature toggle and SHAP will report effectively 0 importance for the feature. Imputing sparse data will, naturally, change this behavior (but not by an appreciable amount). [I think it's low-risk on the accuracy side]
3. If the eventual goal is to support a broader application of efficient evaluation execution (supporting the full range of model-family-specific-implementations (i.e., TreeSHAP for tree-based models)) and to support arbitrary ML packages that may have dramatically different internal handling behavior for missing value elements, exposing a flexible config-driven logic for how to handle imputation and missing values to produce relevant feature importance results might be worthwhile.

mlflow/models/evaluation/default_evaluator.py
BenWilson2: This is an excellent point and one that isn't universally supported in OSS ML packages but which proprietary statistical modeling packages support. 
For any 'machine data type' (e.g., int, double, string) each of these, in different scenarios and applications, could be representing data that is ordinal, nominal, or continuous. Having a hybrid typing system that represents these compound types through explicit user declaration could probably go a long way in preventing issues or perceived bugs due to complex inference logic. 
I feel like this is a broader discussion point that opens up a Pandora's box of a deep design (and incredibly out of scope for this PR), but something worth bringing up. 

The generally accepted best practice for handling data types and any inferred processing performed on them would be something akin to:

Nominal (non-ordered) [either categorical or numeric]: validate that an encoding has been performed (binary encoding, byte encoding, binning, clustering, etc.) and that is the appropriate type of encoding for the model being used based on the cardinality of a raw feature (i.e. don't do one hot encoding for a nominal value being sent to a tree-based algorithm if the count of unique values is > 30%  (placeholder; this is use-case dependent) of the data)

Ordinal (ordered) [either categorical or numeric]: generally handled by an encoder that preserves relationships through an enumeration or binning based on related characteristics of the underlying data (e.g., shirt sizes XXXS, XXS, XS -> 0 ; S, M -> 1; L -> 2; XL, XXL, XXXL -> 3....). 

I think implementing any of these validations or handling of inferred decisions based on the data without user input will be a truly daunting level of complexity. 
Can we do a 'best effort' with inference but leave it up to the users to explicitly define handling of feature types if they see that the inference isn't handling their data in the way that they expect it to?

jinzhang21: Agree with Ben best effort detection probably suffice here. I'd recommend to remove L50-L55. Only return True if we for sure knows it is categorical. Returns False otherwise. But we need to add Pandas type detection at the beginning L46, to make sure users can always tell us it's categorical by changing the column to "category" if that's what they want.

mlflow/models/evaluation/default_evaluator.py
jinzhang21: If we don't know if a column is categorical, just return False. Real numbers contains all discrete set of numbers (superset).

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: Updated.
Now newest code for "category", "string", "boolean" types, `_is_categorical` return True, otherwise return False.

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: Updated, newest code changes:

1. Make imputation preprocessing for all kinds of Shap explainers, the imputation rule is: if we can confirm it is continuous values, use mean imputation, otherwise use mode imputation.

2. Add exception catching wrapper during running shap evaluation.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -41,18 +42,38 @@
...
 
+def _is_categorical(values):
+    """
+    Infer whether input values are categorical on best effort.
+    Return True represent they are categorical, return False represent we cannot determine result.
+    """
+    dtype_name = pd.Series(values).convert_dtypes().dtype.name.lower()
+    if dtype_name in ["category", "string", "boolean"]:
+        return True
+    return False
```
harupy, line 53:
```suggestion
    return dtype_name in ["category", "string", "boolean"]:
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -41,18 +42,38 @@
...
+
+def _is_continuous(values):
+    """
+    Infer whether input values is continuous on best effort.
+    Return True represent they are continous, return False represent we cannot determine result.
+    """
+    dtype_name = pd.Series(values).convert_dtypes().dtype.name.lower()
+    if dtype_name.startswith("float"):
+        return True
+    return False
```
harupy, line 64:
```suggestion
    return dtype_name.startswith("float")
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +390,29 @@ def __validate_artifacts(artifacts):
...
+    continuous_cols = []
+    other_cols = []  # contains columns which are categorical columns or unknown type columns.
+    for col in df.columns:
+        if _is_continuous(df[col]):
+            continuous_cols.append(col)
+        else:
+            other_cols.append(col)
+
+    means = df[continuous_cols].mean().to_dict()
+    modes = df[other_cols].mode().loc[0].to_dict()
```
harupy, line 409:
```suggestion
    continuous_cols = []
    for col in df.columns:
        if _is_continuous(df[col]):
            continuous_cols.append(col)

    means = df[continuous_cols].mean().to_dict()
    modes = df.drop(continuous_cols, axis=1).mode().loc[0].to_dict()
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,7 +641,11 @@ def _log_model_explainability(self):
             # TODO: The explainer saver is buggy, if `get_underlying_model_flavor` return "unknown",
             #   then fallback to shap explainer saver, and shap explainer will call `model.save`
             #   for sklearn model, there is no `.save` method, so error will happen.
-            _logger.warning(f"Log explainer failed. Reason: {str(e)}")
+            _logger.warning(
+                f"Log explainer failed. Reason: {repr(e)}."
```
harupy, line 645:
```suggestion
                f"Logging explainer failed. Reason: {repr(e)}."
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +386,26 @@ def __validate_artifacts(artifacts):
...
+    Compute mean (for continuous columns) and compute mode (for other columns) for the
+    input dataframe, return a dict, key is column name, value is the corresponding mode or
+    mean value, this function calls `_is_continuous` to determine whether the
+    column is continuous column.
+    """
+    df_cols = df.columns.tolist()
+    continuous_cols = [c for c in df_cols if _is_continuous(df[c])]
+    means = df[continuous_cols].mean().to_dict()
+    if continuous_cols == df_cols:
+        return means
```
WeichenXu123, line 400: This case `df.drop(continuous_cols)` results is a empty dataframe which causes `.loc[0]` raises key error.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -369,6 +386,26 @@ def __validate_artifacts(artifacts):
...
+    else:
+        modes = df.drop(continuous_cols, axis=1).mode().loc[0].to_dict()
+        return {**means, **modes}
```
harupy, line 403: `df[continuous_cols]` could be empty. I would do something like this:

```suggestion
    a = df[continuous_cols]
    b = df.drop(continuous_cols, axis=1)
    
    means = {} if a.empty else a.mean().to_dict()
    modes = {} if b.empty else b.loc[0].mean().to_dict()
    return {**means, **modes}
```

tests/models/test_default_evaluator.py
```diff
@@ -305,12 +308,67 @@ def test_svm_classifier_evaluation(svm_model_uri, breast_cancer_dataset):
...
+
+
 def test_infer_model_type_by_labels():
     assert _infer_model_type_by_labels(["a", "b"]) == "classifier"
+    assert _infer_model_type_by_labels([True, False]) == "classifier"
     assert _infer_model_type_by_labels([1, 2.5]) == "regressor"
-    assert _infer_model_type_by_labels(list(range(2000))) == "regressor"
-    assert _infer_model_type_by_labels([1, 2, 3]) == "classifier"
-    assert _infer_model_type_by_labels(list(range(30))) is None
+    assert _infer_model_type_by_labels(pd.Series(["a", "b"], dtype="category")) is "classifier"
```
harupy, line 370:
```suggestion
    assert _infer_model_type_by_labels(pd.Series(["a", "b"], dtype="category")) == "classifier"
```

mlflow/models/evaluation/default_evaluator.py
harupy: or you can use:

```
df.mode().head(1).squeeze().to_dict()
```

this works even if `df` is empty.

BenWilson2 (approved):
LGTM with one small nit about a test validating the categorical assignment on floats.

Issue #6105, opened 2022-06-18
[BUG] Nested `MlflowException` when calling `mlflow.set_experiment()` on a RestStore endpoint that requires authentication


balvisio:
### Willingness to contribute

Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.

### MLflow version

mlflow, version 1.26.1

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: CentOS 8
- **Python version**: Python 3.9.12
- **yarn version, if running the dev UI**:

### Describe the problem

**Note**: This might be not considered a bug but more of an improvement or optimization.

When setting a RestStore that requires authentication but the credentials are NOT set:
```python
mlflow.set_tracking_uri(mlflow_params.tracking_uri) # This endpoint requires auth
try:
    mlflow.set_experiment(mlflow_params.experiment)
except MlflowException as e:
    log.exception(msg="An error occurred while setting MLflow experiment")
```

The following nested exceptions occur:
```python-traceback
ERROR:root:An error occurred while setting MLflow experiment
Traceback (most recent call last):
  File "/home/bruno/anaconda3/envs/mlflow-env/lib/python3.9/site-packages/mlflow/store/tracking/rest_store.py", line 285, in get_experiment_by_name
    response_proto = self._call_endpoint(GetExperimentByName, req_body)
  File "/home/bruno/anaconda3/envs/mlflow-env/lib/python3.9/site-packages/mlflow/store/tracking/rest_store.py", line 56, in _call_endpoint
    return call_endpoint(self.get_host_creds(), endpoint, method, json_body, response_proto)
  File "/home/bruno/anaconda3/envs/mlflow-env/lib/python3.9/site-packages/mlflow/utils/rest_utils.py", line 256, in call_endpoint
    response = verify_rest_response(response, endpoint)
  File "/home/bruno/anaconda3/envs/mlflow-env/lib/python3.9/site-packages/mlflow/utils/rest_utils.py", line 191, in verify_rest_response
    raise MlflowException("%s. Response body: '%s'" % (base_msg, response.text))
mlflow.exceptions.MlflowException: API request to endpoint /api/2.0/mlflow/experiments/get-by-name failed with error code 401 != 200. Response body: '<html>
<head><title>401 Authorization Required</title></head>
<body>
<center><h1>401 Authorization Required</h1></center>
<hr><center>nginx/1.20.0</center>
</body>
</html>
'

[...]

PR #6106, opened 2022-06-18
Handle auth required http status code to prevent nested exception

balvisio:
## Related Issues/PRs

Related Issue: https://github.com/mlflow/mlflow/issues/6105

The fix to the issue is to:
1. Populate `MlflowException` with the appropriate `error_code` derived from the returned status code of the request.
2. When the exception is being handled in `get_experiment_by_name()` look at the `error_code` and if this error is not temporal raise the exception to the caller.

## How is this patch tested?

I have built a new dev package and tested on my local. With the proposed change, I see that only one exception is raised:

```python-traceback
INFO:root:An error occurred while setting MLflow experiment
  File "/home/bruno/repos/NIH-NEI/mouse-image-segmentation/unet/model/training.py", line 297, in train_model
    mlflow.set_experiment(mlflow_params.experiment)
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/tracking/fluent.py", line 110, in set_experiment
    experiment = client.get_experiment_by_name(experiment_name)
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/tracking/client.py", line 462, in get_experiment_by_name
    return self._tracking_client.get_experiment_by_name(name)
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/tracking/_tracking_service/client.py", line 163, in get_experiment_by_name
    return self.store.get_experiment_by_name(name)
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/store/tracking/rest_store.py", line 297, in get_experiment_by_name
    raise e
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/store/tracking/rest_store.py", line 286, in get_experiment_by_name
    response_proto = self._call_endpoint(GetExperimentByName, req_body)
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/store/tracking/rest_store.py", line 57, in _call_endpoint
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/utils/rest_utils.py", line 260, in call_endpoint
  File "/home/bruno/anaconda3/envs/oct-seg/lib/python3.9/site-packages/mlflow/utils/rest_utils.py", line 192, in verify_rest_response
    raise MlflowException(
mlflow.exceptions.MlflowException: API request to endpoint /api/2.0/mlflow/experiments/get-by-name failed with error code 401 != 200. Response body: '<html>
<head><title>401 Authorization Required</title></head>
<body>
<center><h1>401 Authorization Required</h1></center>
<hr><center>nginx/1.20.0</center>
</body>
</html>
'
```

I would need some guidance to write appropriate tests.

   next step, otherwise fix it.
3. Click `Details` on the right to open the job page of CircleCI.
4. Click the `Artifacts` tab.
5. Click `docs/build/html/index.html`.
6. Find the changed pages / sections and make sure they render correctly.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/exceptions.py
+++ b/mlflow/exceptions.py
@@ -5,6 +5,7 @@ from mlflow.protos.databricks_pb2 import (
     TEMPORARILY_UNAVAILABLE,
     ENDPOINT_NOT_FOUND,
     PERMISSION_DENIED,
+    CUSTOMER_UNAUTHORIZED,
     REQUEST_LIMIT_EXCEEDED,
     BAD_REQUEST,
     INVALID_PARAMETER_VALUE,
@@ -22,10 +23,22 @@ ERROR_CODE_TO_HTTP_STATUS = {
     ErrorCode.Name(ENDPOINT_NOT_FOUND): 404,
     ErrorCode.Name(RESOURCE_DOES_NOT_EXIST): 404,
     ErrorCode.Name(PERMISSION_DENIED): 403,
+    ErrorCode.Name(CUSTOMER_UNAUTHORIZED): 401,
     ErrorCode.Name(BAD_REQUEST): 400,
     ErrorCode.Name(RESOURCE_ALREADY_EXISTS): 400,
     ErrorCode.Name(INVALID_PARAMETER_VALUE): 400,
 }
+
+HTTP_STATUS_TO_ERROR_CODE = dict((v, k) for k, v in ERROR_CODE_TO_HTTP_STATUS.items())
+HTTP_STATUS_TO_ERROR_CODE[400] = ErrorCode.Name(BAD_REQUEST)
+HTTP_STATUS_TO_ERROR_CODE[404] = ErrorCode.Name(ENDPOINT_NOT_FOUND)
+HTTP_STATUS_TO_ERROR_CODE[500] = ErrorCode.Name(INTERNAL_ERROR)
+
+
+def get_error_code(http_status):
+    return ErrorCode.Value(
+        HTTP_STATUS_TO_ERROR_CODE.get(http_status, ErrorCode.Name(INTERNAL_ERROR))
+    )
 
 
 class MlflowException(Exception):
--- a/mlflow/store/tracking/rest_store.py
+++ b/mlflow/store/tracking/rest_store.py
@@ -289,15 +289,13 @@ class RestStore(AbstractStore):
                 databricks_pb2.RESOURCE_DOES_NOT_EXIST
             ):
                 return None
-            elif e.error_code == databricks_pb2.ErrorCode.Name(
-                databricks_pb2.REQUEST_LIMIT_EXCEEDED
-            ):
-                raise e
-            # Fall back to using ListExperiments-based implementation.
-            for experiment in self.list_experiments(ViewType.ALL):
-                if experiment.name == experiment_name:
-                    return experiment
-            return None
+            elif e.error_code == databricks_pb2.ErrorCode.Name(databricks_pb2.ENDPOINT_NOT_FOUND):
+                # Fall back to using ListExperiments-based implementation.
+                for experiment in self.list_experiments(ViewType.ALL):
+                    if experiment.name == experiment_name:
+                        return experiment
+                return None
+            raise e
 
     def log_batch(self, run_id, metrics, params, tags):
         metric_protos = [metric.to_proto() for metric in metrics]
--- a/mlflow/utils/rest_utils.py
+++ b/mlflow/utils/rest_utils.py
@@ -14,7 +14,7 @@ from mlflow.protos import databricks_pb2
 from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE, ENDPOINT_NOT_FOUND, ErrorCode
 from mlflow.utils.proto_json_utils import parse_dict
 from mlflow.utils.string_utils import strip_suffix
-from mlflow.exceptions import MlflowException, RestException
+from mlflow.exceptions import get_error_code, MlflowException, RestException
 
 RESOURCE_DOES_NOT_EXIST = "RESOURCE_DOES_NOT_EXIST"
 _REST_API_PATH_PREFIX = "/api/2.0"
@@ -188,7 +188,10 @@ def verify_rest_response(response, endpoint):
                 endpoint,
                 response.status_code,
             )
-            raise MlflowException("%s. Response body: '%s'" % (base_msg, response.text))
+            raise MlflowException(
+                "%s. Response body: '%s'" % (base_msg, response.text),
+                error_code=get_error_code(response.status_code),
+            )
 
     # Skip validation for endpoints (e.g. DBFS file-download API) which may return a non-JSON
     # response


mlflow/protos/databricks.proto
```diff
@@ -229,6 +229,9 @@ enum ErrorCode {
   // If the API request is rejected due to throttling
   REQUEST_LIMIT_EXCEEDED = 1007;
 
+  // Indicates that the given API endpoint requires auth credentials
+  AUTHORIZATION_REQUIRED = 1008;
+
```
dbczumar, line 234: @balvisio Looks like https://github.com/mlflow/mlflow/pull/6106/files#diff-1bf3cdd39aa480c64e304f8e7075e7592f7b7dcc2e2bdf80ae0e242914eb23ecR227 (`CUSTOMER_UNAUTHORIZED = 1006;`) covers this case already. Can we use that error status value instead?

mlflow/store/tracking/rest_store.py
```diff
@@ -289,8 +290,9 @@ def get_experiment_by_name(self, experiment_name):
                 databricks_pb2.RESOURCE_DOES_NOT_EXIST
             ):
                 return None
-            elif e.error_code == databricks_pb2.ErrorCode.Name(
-                databricks_pb2.REQUEST_LIMIT_EXCEEDED
+            elif (
+                e.error_code == databricks_pb2.ErrorCode.Name(databricks_pb2.REQUEST_LIMIT_EXCEEDED)
+                or e.get_http_status_code() not in _TRANSIENT_FAILURE_RESPONSE_CODES
```
dbczumar, line 295: Nice catch!

dbczumar (commented):
@balvisio LGTM once https://github.com/mlflow/mlflow/pull/6106/files#r902022618 is addressed and proto changes are reverted; thanks so much for your contribution!

mlflow/protos/databricks.proto
balvisio: @dbczumar Sure. I missed it. I will revert the change here.

balvisio:
Thanks @dbczumar . I have reverted the proto.

dbczumar (approved):
LGTM! Thanks @balvisio !

mlflow/store/tracking/rest_store.py
dbczumar: @balvisio Looks like an associated test (`TestRestStore.test_get_experiment_by_name[RestStore] `, https://github.com/mlflow/mlflow/runs/6990695520?check_suite_focus=true) is failing with

```
defresponse_fn(*args, **kwargs):
# pylint: disable=unused-argument
if kwargs.get("endpoint") == "/api/2.0/mlflow/experiments/get-by-name":
raise MlflowException(
>               "GetExperimentByName is not implemented", ENDPOINT_NOT_FOUND
            )
E           mlflow.exceptions.MlflowException: GetExperimentByName is not implemented
args       = ()
kwargs     = {'endpoint': '/api/2.0/mlflow/experiments/get-by-name', 'host_creds': <mlflow.utils.rest_utils.MlflowHostCreds object at 0x7f749f6a16d0>, 'method': 'GET', 'params': {'experiment_name': 'abc'}}
list_exp_response = <MagicMock id='140138615968208'>
```

Can we special case `ENDPOINT_NOT_FOUND` here as well & fall back to the `ListExperiments`-based implementation?

balvisio: @dbczumar yeah I saw the failure. What if the function is written like this:

```python
def get_experiment_by_name(self, experiment_name):
        try:
            req_body = message_to_json(GetExperimentByName(experiment_name=experiment_name))
            response_proto = self._call_endpoint(GetExperimentByName, req_body)
            return Experiment.from_proto(response_proto.experiment)
        except MlflowException as e:
            if e.error_code == databricks_pb2.ErrorCode.Name(
                databricks_pb2.RESOURCE_DOES_NOT_EXIST
            ):
                return None
            elif e.error_code == databricks_pb2.ErrorCode.Name(
                databricks_pb2.ENDPOINT_NOT_FOUND
            ):
                # Fall back to using ListExperiments-based implementation.
                for experiment in self.list_experiments(ViewType.ALL):
                    if experiment.name == experiment_name:
                        return experiment
                return None 
            raise e
```

I run the rest_store tests locally and they pass

balvisio: I think the implementation above is not fully correct though because we are populating the error code from the HTTP status but since 404 maps both to ENDPOINT_NOT_FOUND and RESOURCE_DOES_NOT_EXIST the function could be returning `None` when it actually needs to do the listing method. A solution could be to add :

```python
HTTP_STATUS_TO_ERROR_CODE = dict((v, k) for k, v in ERROR_CODE_TO_HTTP_STATUS.items())
HTTP_STATUS_TO_ERROR_CODE[404] = ErrorCode.Name(ENDPOINT_NOT_FOUND)
```

to error on the safe side

balvisio: Also a question I have is: why if the ENDPOINT_NOT_FOUND then `list_experiment` help? Doesn't this function also calls the same endpoint?

dbczumar: @balvisio Thanks for calling this out! Can we update `mlflow.utils.rest_utils.get_error_code` to check the response *text* if the error code is 404 and, if the text contains `"ENDPOINT_NOT_FOUND"`, return the "endpoint not found" error code representation; otherwise, the "resource does not exist" error code representation should be returned.

balvisio: @dbczumar I think that removing `RESOURCE_DOES_NOT_EXIST` from the mapping is the safe option because we will be defaulting to `ENDPOINT_NOT_FOUND` which it will cause the code to try the `list_experiment()` path. 

From looking at the tests, I thought that the MLFlow server specifically returned a JSON struct containing: `RESOURCE_DOES_NOT_EXIST`: 

```
{
    "error_code": "RESOURCE_DOES_NOT_EXIST", 
    "message": "Node type not supported"
}
```

which is handled in this branch of `verify_rest_response`:

```
if _can_parse_as_json_object(response.text):
else:
...
```

The reason I am reluctant to look at the text message is that if there is a proxy server between the MLflow server and the client (for example, nginx) the string "API not found" might or might be there depending on the proxy.

Let me know if you still want to go ahead with https://github.com/mlflow/mlflow/pull/6106#discussion_r904376705 and I can update the PR.

dbczumar: @balvisio Apologies for the delay. I think that defaulting to `RESOURCE_DOES_NOT_EXIST` for 404 is a better option, since `ENDPOINT_NOT_FOUND` is a more specific status that is not commonly used throughout MLflow. Rather than checking for message text to parse `ENDPOINT_NOT_FOUND`, which I agree is brittle, could we instead rely on the fact that `ENDPOINT_NOT_FOUND` is parsed as a `RestException` in the `get_experiment_by_name()` exception handler?

If that approach doesn't work, another option is to un-indent the preexisting `list_experiments()` fallback logic. This way, we won't encounter the nested exception. I realize that we may end up falling back to `list_experiments()` for irrecoverable errors (e.g. auth errors) in this case, but that's consistent with preexisting behavior.

balvisio: @dbczumar thanks for the reply. I got the point of `ENDPOINT_NOT_FOUND` not being used frequently. 

However, by injecting `ENDPOINT_NOT_FOUND` in `rest_utils.py:191` (using the `get_error_code()` function) and then calling `list_experiments()` in the `ENDPOINT_NOT_FOUND` case (`rest_store.py:333`), we are effectively "falling back" to calling `list_experiments()` path which is what we want.
I also just noticed the implementation `get_experiment_by_name()` of the `DatabricksRestStore` (`rest_store.py:325`) uses the same logic.

dbczumar: Right, we want to effectively fall back. It seems like, if the `ENDPOINT_NOT_FOUND` error is parsed as a `RestException`, which I think is the case, we can achieve this fallback logic without mapping the 404 status code to `ENDPOINT_NOT_FOUND`. I.e. tests should still pass and the behavior should still be correct if we change


```
# Removing 'RESOURCE_DOES_NOT_EXIST' from dict to disambiguate between RESOURCE_DOES_NOT_EXIST and
# ENDPOINT_NOT_FOUND. When 'get_error_code()' is used it is the ENDPOINT_NOT_FOUND case since when
# RESOURCE_DOES_NOT_EXIST the response content is (and parsed as) a JSON object.
HTTP_STATUS_TO_ERROR_CODE = dict(
    (v, k)
    for k, v in ERROR_CODE_TO_HTTP_STATUS.items()
    if k != ErrorCode.Name(RESOURCE_DOES_NOT_EXIST)
)
```

to 

```
HTTP_STATUS_TO_ERROR_CODE = dict(
    (v, k)
    for k, v in ERROR_CODE_TO_HTTP_STATUS.items()
)
HTTP_STATUS_TO_ERROR_CODE[404] = ErrorCode.Name(RESOURCE_DOES_NOT_EXIST)
```

balvisio: @dbczumar Great. Updated the PR. Not sure why in the last run some pipelines failed at the beginning; looked like some spurious error.

mlflow/exceptions.py
```diff
@@ -22,11 +23,21 @@
...
     ErrorCode.Name(RESOURCE_DOES_NOT_EXIST): 404,
     ErrorCode.Name(PERMISSION_DENIED): 403,
+    ErrorCode.Name(CUSTOMER_UNAUTHORIZED): 401,
     ErrorCode.Name(BAD_REQUEST): 400,
     ErrorCode.Name(RESOURCE_ALREADY_EXISTS): 400,
     ErrorCode.Name(INVALID_PARAMETER_VALUE): 400,
 }
 
+HTTP_STATUS_TO_ERROR_CODE = dict((v, k) for k, v in ERROR_CODE_TO_HTTP_STATUS.items())
+HTTP_STATUS_TO_ERROR_CODE[404] = ErrorCode.Name(RESOURCE_DOES_NOT_EXIST)
```
dbczumar, line 33: @balvisio Looks great! One more small ask: could we explicitly map `500` to `INTERNAL_ERROR` and `400` to `BAD_REQUEST`?

balvisio: @dbczumar Updated the PR.

dbczumar (approved):
@balvisio Thanks for all your hard work on this! It looks great! Merging :)

PR #6117, opened 2022-06-23
Add `pos_label` parameter for `mlflow.sklearn.autolog`

wenfeiy-db:
## Related Issues/PRs

Related PR https://github.com/mlflow/mlflow/issues/5512

The previous PR https://github.com/mlflow/mlflow/issues/5512 exposes the same parameter `pos_label` in `mlflow.sklearn.eval_and_log_metrics` to allow user to optionally fill in the `pos_label` if they are doing binary classification. This PR adds `pos_label` parameter for `mlflow.sklearn.autolog`. If set, the `pos_label` will be used for the autologged training metrics.

## How is this patch tested?
New unit test

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Allow user passing in `pos_label` in sklearn autologging.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -1184,11 +1184,10 @@ def autolog(
                                   new model version of the registered model with this name.
                                   The registered model is created if it does not already exist.
     :param pos_label: If given, used as the positive label to compute binary classification
-                      training metrics such as precision, recall, f1, etc. This parameter
-                      should only be set for binary classification model
-                      - if used on multi-label model, the training metrics calculation will fail
-                        and the training metrics won't be logged.
-                      - if used for regression model, the parameter will be ignored.
+                      training metrics such as precision, recall, f1, etc. This parameter should
+                      only be set for binary classification model. If used for multi-label model,
+                      the training metrics calculation will fail and the training metrics won't
+                      be logged. If used for regression model, the parameter will be ignored.
     """
     _autolog(
         flavor_name=FLAVOR_NAME,
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -1183,9 +1183,9 @@ def autolog(
     :param registered_model_name: If given, each time a model is trained, it is registered as a
                                   new model version of the registered model with this name.
                                   The registered model is created if it does not already exist.
-    :param pos_label: If given, used as the positive label to compute binary classification metrics such as
+    :param pos_label: If given, used as the positive label to compute binary classification training metrics such as
                       precision, recall, f1, etc. This parameter should only be set for binary classification model
-                      - if used on multi-label model, the evaluation will fail;
+                      - if used on multi-label model, the training metrics calculation will fail and won't be logged.
                       - if used for regression model, the parameter will be ignored.
     """
     _autolog(
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -1183,9 +1183,11 @@ def autolog(
     :param registered_model_name: If given, each time a model is trained, it is registered as a
                                   new model version of the registered model with this name.
                                   The registered model is created if it does not already exist.
-    :param pos_label: If given, used as the positive label to compute binary classification training metrics such as
-                      precision, recall, f1, etc. This parameter should only be set for binary classification model
-                      - if used on multi-label model, the training metrics calculation will fail and won't be logged.
+    :param pos_label: If given, used as the positive label to compute binary classification
+                      training metrics such as precision, recall, f1, etc. This parameter
+                      should only be set for binary classification model
+                      - if used on multi-label model, the training metrics calculation will fail
+                        and the training metrics won't be logged.
                       - if used for regression model, the parameter will be ignored.
     """
     _autolog(
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -925,6 +925,7 @@ def autolog(
     log_post_training_metrics=True,
     serialization_format=SERIALIZATION_FORMAT_CLOUDPICKLE,
     registered_model_name=None,
+    pos_label=None,
 ):  # pylint: disable=unused-argument
     """
     Enables (or disables) and configures autologging for scikit-learn estimators.
@@ -1182,6 +1183,10 @@ def autolog(
     :param registered_model_name: If given, each time a model is trained, it is registered as a
                                   new model version of the registered model with this name.
                                   The registered model is created if it does not already exist.
+    :param pos_label: If given, used as the positive label to compute binary classification metrics such as
+                      precision, recall, f1, etc. This parameter should only be set for binary classification model
+                      - if used on multi-label model, the evaluation will fail;
+                      - if used for regression model, the parameter will be ignored.
     """
     _autolog(
         flavor_name=FLAVOR_NAME,
@@ -1195,6 +1200,7 @@ def autolog(
         max_tuning_runs=max_tuning_runs,
         log_post_training_metrics=log_post_training_metrics,
         serialization_format=serialization_format,
+        pos_label=pos_label,
     )
 
 
@@ -1210,6 +1216,7 @@ def _autolog(
     max_tuning_runs=5,
     log_post_training_metrics=True,
     serialization_format=SERIALIZATION_FORMAT_CLOUDPICKLE,
+    pos_label=None,
 ):  # pylint: disable=unused-argument
     """
     Internal autologging function for scikit-learn models.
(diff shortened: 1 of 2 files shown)


mlflow/sklearn/__init__.py
```diff
@@ -925,6 +925,7 @@ def autolog(
     log_post_training_metrics=True,
     serialization_format=SERIALIZATION_FORMAT_CLOUDPICKLE,
     registered_model_name=None,
+    pos_label=None,
```
WeichenXu123, line 928: One concern:
```
mlflow.sklearn.autolog(pos_label=X)
```
then it means all sklearn estimator training it must be binary classifier with pos_label=X or regressor (regressor ignore the pos_label), in other cases autologging will fail or generate wrong metric result. @dbczumar Do you think this is a risk ?

dbczumar: @WeichenXu123 @wenfeiy-db Yes, this is a concern. We should only make use of `pos_label` when the model is of a relevant type (classifier).

wenfeiy-db: @WeichenXu123 I thought for "other cases" (i.e. setting pos_label=X for multi-label question) autolog won't fail, but we'll just have a  warning that it fails to compute the metrics, which is covered by the try-catch [block](https://github.com/mlflow/mlflow/blob/2bf69e5c9a8910d2c1003e128ac4d20ee03fcfce/mlflow/sklearn/utils.py#L489) in `_log_specialized_estimator_content`.  Did I get it wrong?
@dbczumar I believe it's only applied for classifiers right now. Because this PR just passes the `pos_label` to `_log_estimator_content`, which again passes to `_log_specialized_estimator_content`, which only uses the `pos_label` if the model is a classifier ([code link](https://github.com/mlflow/mlflow/blob/f033eae3f95520bcdabbf9bce854fffb63af5c09/mlflow/sklearn/utils.py#L490-L493)). Please correct me if I'm wrong.

harupy: `pos_label` is ignored for regressors:

https://github.com/mlflow/mlflow/blob/83b778bb4a098e963c1fbb748c93af64ec258fb3/mlflow/sklearn/utils.py#L481-L495

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html#sklearn.metrics.precision_score says:

> pos_labelstr or int, default=1
The class to report if average='binary' and the data is binary. **If the data are multiclass or multilabel, this will be ignored**; setting labels=[pos_label] and average != 'binary' will report scores for that label only.

wenfeiy-db: To make it clear, these are different cases can happen when `pos_label` is set and not None:
* If it's a binary classification, autologging uses the given `pos_label` to calculate training metrics and log them
* If it's a multi-label classification, the training metrics calculation will fail (but not the overall autolog). The calculation fails instead of `pos_label` being ignored because we explicitly set `average` to `binary` when having `pos_label` in [_get_classifier_metrics](https://github.com/mlflow/mlflow/blob/816e035786245ff42723b9c53eb9407885e7cd75/mlflow/sklearn/utils.py#L150). In this case autolog just doesn't log any training metrics.
* If it's a regression, pos_label got ignored. The training metrics are calculated as normal.

The above cases sound okay to me. The only doubtful point is that whether we want to log the normal weighted training metrics instead of not logging anything in the 2nd case.

dbczumar: @wenfeiy-db Thank you for the clarification. I think that failing the metric computation in the case where `pos_label` is specified for a multi-class classifier is acceptable; can we make sure that we raise a clear error message?

wenfeiy-db: Hi @dbczumar I added a test which makes sure that autolog emits warning when the `pos_label` is specified for multi-class. I think the current warning message is pretty clear. Let me know if it looks good.

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1992,3 +1992,39 @@ def test_autolog_registering_model():
...
+    model = fit_model(model, X, y, "fit")
+
+    run_id = mlflow.active_run().info.run_id
+    _, training_metrics, _, _ = get_run_data(run_id)
+    expected_training_metrics = mlflow.sklearn.eval_and_log_metrics(
+        model=model, X=X, y_true=y, prefix="training_", pos_label=1
+    )
+    assert training_metrics == expected_training_metrics
+
+    mlflow.end_run()
```
harupy, line 2015:
```suggestion
    with mlflow.start_run() as run:
        model.fit(model, X, y, "fit")

    _, training_metrics, _, _ = get_run_data(run.info.run_id)
    expected_training_metrics = mlflow.sklearn.eval_and_log_metrics(
        model=model, X=X, y_true=y, prefix="training_", pos_label=1
    )
    assert training_metrics == expected_training_metrics
```

tests/sklearn/test_sklearn_autolog.py
```diff
@@ -1992,3 +1992,39 @@ def test_autolog_registering_model():
...
+
+def test_autolog_emits_warning_message_when_pos_label_used_for_multilabel():
+    mlflow.sklearn.autolog(pos_label=1)
+
+    model = sklearn.svm.SVC()
+    X, y = get_iris()
+
+    with mlflow.start_run(), mock.patch("mlflow.sklearn.utils._logger.warning") as mock_warning:
+        model.fit(X, y)
+        mock_warning.called_once_with(
```
harupy, line 2026:
```suggestion
        mock_warning.assert_called_once_with(
```

mlflow/sklearn/__init__.py
```diff
@@ -1182,6 +1183,12 @@ def fetch_logged_data(run_id):
     :param registered_model_name: If given, each time a model is trained, it is registered as a
                                   new model version of the registered model with this name.
                                   The registered model is created if it does not already exist.
+    :param pos_label: If given, used as the positive label to compute binary classification
+                      training metrics such as precision, recall, f1, etc. This parameter
+                      should only be set for binary classification model
+                      - if used on multi-label model, the training metrics calculation will fail
+                        and the training metrics won't be logged.
+                      - if used for regression model, the parameter will be ignored.
```
harupy, line 1191: @wenfeiy-db Can you fix this doc build error: https://app.circleci.com/pipelines/github/mlflow/mlflow/12552/workflows/cb9ac64c-2368-4844-907e-37547ffef633/jobs/21422?invite=true#step-104-62?

harupy (approved):
LGTM once remaining comments are addressed :)

WeichenXu123:
@wenfeiy-db 
Shall we also emit warning when `pos_label` is not `1` or `True` but the binary dataset there's no values equal to the specified pos_label ?

e.g., user set `pos_label` in autologging to be `2`, but then user accidentally run a model training over dataset with normal binary target values (0/1 or True/False), then user will get unexpected result.

harupy:
@WeichenXu123 scikit-learn throws an exception in that case:

```python
from sklearn.metrics import precision_score

y_true = [0, 1, 2, 0, 1, 2]
y_pred = [0, 2, 1, 0, 0, 1]
precision_score(y_true, y_pred, average='macro')
```

```
  File "a.py", line 5, in <module>
    precision_score(y_true, y_pred, pos_label=2)
  File "/home/haru/miniconda3/envs/mlflow-dev-env/lib/python3.7/site-packages/sklearn/metrics/_classification.py", line 1765, in precision_score
    zero_division=zero_division,
  File "/home/haru/miniconda3/envs/mlflow-dev-env/lib/python3.7/site-packages/sklearn/metrics/_classification.py", line 1544, in precision_recall_fscore_support
    labels = _check_set_wise_labels(y_true, y_pred, average, labels, pos_label)
  File "/home/haru/miniconda3/envs/mlflow-dev-env/lib/python3.7/site-packages/sklearn/metrics/_classification.py", line 1357, in _check_set_wise_labels
    f"pos_label={pos_label} is not a valid label. It "
ValueError: pos_label=2 is not a valid label. It should be one of [0, 1]
```

tests/sklearn/test_sklearn_autolog.py
wenfeiy-db: There're actually 3 warnings, each for precision, recall and f1. I changed the test accordingly. 
I also found `called_once_with` actually doesn't do any assert, while it's used in several other tests, such as `test_autolog_max_tuning_runs_logs_info_correctly`. I'm not able to quickly fix the other tests to use `assert_called_once_with` in this PR, but we'll probably want to fix it some time.

mlflow/sklearn/__init__.py
wenfeiy-db: The doc build now passes and the page LGTM.
image

WeichenXu123 (approved):

PR #6141, opened 2022-06-25
Fix default evaluator behavior for regression models that mutate input datasets

dbczumar:

Fix default evaluator behavior for regression models that mutate input datasets by ensuring that metrics & explanations are only computed on fresh dataset copies.

## How is this patch tested?

Added integration test

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

## Release Notes

Fix a bug in ``mlflow.evaluate()`` that caused kernel explanations to fail for models that mutate input data during inference.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/models/evaluation/default_evaluator.py
+++ b/mlflow/models/evaluation/default_evaluator.py
@@ -1063,7 +1063,7 @@ class DefaultEvaluator(ModelEvaluator):
         """
         Wrapper around a data object that only provides copies of the data object, guarding against
         accidental reuse of the data object after it is mutated, e.g. during inference with
-        scikit-learn models
+        scikit-learn models.
         """
 
         def __init__(self, data):
--- a/mlflow/models/evaluation/default_evaluator.py
+++ b/mlflow/models/evaluation/default_evaluator.py
@@ -94,12 +94,12 @@ def _extract_raw_model(model):
 
 
 def _extract_predict_fn(model, raw_model):
-    og_predict_fn = model.predict
-    og_predict_proba_fn = None
+    predict_fn = model.predict
+    predict_proba_fn = None
 
     if raw_model is not None:
-        og_predict_fn = raw_model.predict
-        og_predict_proba_fn = getattr(raw_model, "predict_proba", None)
+        predict_fn = raw_model.predict
+        predict_proba_fn = getattr(raw_model, "predict_proba", None)
 
         try:
             import xgboost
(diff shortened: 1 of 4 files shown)


mlflow/models/evaluation/default_evaluator.py
```diff
@@ -94,40 +94,25 @@ def _extract_raw_model(model):
...
-    def predict_fn(input_df):
-        return og_predict_fn(input_df.copy())
-
-    if og_predict_proba_fn is not None:
-
-        def predict_proba_fn(input_df):
-            return og_predict_proba_fn(input_df.copy())
-
-    else:
-        predict_proba_fn = og_predict_proba_fn
```
dbczumar, line 129: Remove one-of `pd.DataFrame.copy()` statements in favor of introducing an intermediate `_MutationGuardedData` class that *only* provides access to data by way of copying (see below).

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -1049,3 +1042,44 @@ def evaluate(
...
+        scikit-learn models.
+        """
+
+        def __init__(self, data):
+            """
+            :param data: A data object, such as a Pandas DataFrame, numPy array, or list.
+            """
+            self._data = data
+
+        def copy_in_case_of_mutation(self):
```
dbczumar, line 1075: Prefer `copy_in_case_of_mutation()` over `copy()` so that, when reading the code, it becomes clear why we're obtaining a new copy of the dataset.

jinzhang21: It might be my English, but the name `copy_in_case_of_mutation` made me think it may not perform a copy operation if there's no mutation on the DataFrame. But the way it's implemented is actually always-copy. Shall we rename it to avoid such confusion? e.g. `copy_to_avoid_mutation`(). IMO simple `copy()` also reads pretty clear to me what it's trying to achieve, or something like `pass_by_copy()`, `get_copy()`

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -506,7 +491,8 @@ def _log_model_explainability(self):
             )
             return
 
-        if not (np.issubdtype(self.y.dtype, np.number) or self.y.dtype == np.bool_):
+        labels_dtype = self.y.copy_in_case_of_mutation().dtype
```
jinzhang21, line 494: Do we need to make a copy every single time `self.X` or `self.y` is accessed? It feels excessive and not as efficient.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -693,7 +679,9 @@ def plot_feature_importance():
     def _evaluate_sklearn_model_score_if_scorable(self):
         if self.model_loader_module == "mlflow.sklearn":
             try:
-                score = self.raw_model.score(self.X, self.y)
+                score = self.raw_model.score(
+                    self.X.copy_in_case_of_mutation(), self.y.copy_in_case_of_mutation()
```
jinzhang21, line 683: It might be alright to just make a copy when the model is involved.

jinzhang21 (approved):
Thanks for the fix, Corey!

mlflow/models/evaluation/default_evaluator.py
dbczumar: For current usages of `self.X`, we need a new copy every time with the exception of the `feature_dtypes` check in https://github.com/mlflow/mlflow/pull/6141/files#diff-1b3ab8e1152a42ab0bd3d4e01c7621b37a95aa02fff19642a5abf23f75b6e688R511. I'd rather err on the side of caution for `self.X` because we've seen multiple issues here.

There's no evidence that `self.y` needs to be copied, so I've gone ahead and reverted that overly-conservative change.

mlflow/models/evaluation/default_evaluator.py
dbczumar: Agreed! I've changed this to `copy_in_case_of_mutation()`, which more clearly gets at the motivation. Naively, a reader might wonder why we're calling `copy()`; I think `copy_in_case_of_mutation()` makes things slightly clearer.

mlflow/models/evaluation/default_evaluator.py
dbczumar: Addressed by adding a `get_original()` method to `_MutationGuardedDataset`; The benefit of this class is that it requires explicit API calls to obtain either the original dataset or a copy thereof, requiring the developer to think about whether their workflow might mutate the object.

PR #6147, opened 2022-06-27
Add Leader board for MLP train step card

WeichenXu123:
## Related Issues/PRs

#xxx

FR:
Leaderboard of top 2 models of previous MLflow pipeline runs, plus the latest pipeline run, show in train step with validation data, ranked with primary metric.

This should be a new tab called “Model Leaderboard”.

The board should be a table containing the top 3 models (or fewer if there is not enough previous runs) and the latest model. Columns: run_uuid, run_time, primary_metric, and other relevant metrics if it makes sense, sorted by primary metric.

## How is this patch tested?

Manually

first run:
image

second run:
image

third run:
image

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Mlflow pipeline train step card: Leaderboard of top 2 models of previous MLflow pipeline runs, plus the latest pipeline run, show in train step with validation data, ranked with primary metric.

Components

Interface

Language

Integrations


mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +171,69 @@ def _run(self, output_directory):
...
 
+        from mlflow.entities import ViewType
+
+        mlflow_client = MlflowClient()
+        exp_id = _get_experiment_id()
+        search_result = mlflow_client.search_runs(
+            experiment_ids=exp_id,
+            run_view_type=ViewType.ALL,
+            max_results=10000,
+            order_by=["attribute.start_time DESC", f"metrics.{self.primary_metric}"],
```
harupy, line 182: Can you add sorting order for `primary_metric`?

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +171,69 @@ def _run(self, output_directory):
...
+                }
+                leader_board_items.append(
+                    {
+                        "Run ID": old_run.info.run_id,
+                        "Run Time": datetime.datetime.fromtimestamp(
+                            old_run.info.start_time // 1000
+                        ),
+                        **metric_values,
+                    }
+                )
```
harupy, line 207:
```suggestion
        for old_run in search_result:
            if all(metric_key in old_run.data.metrics for metric_key in metric_keys):
                metric_values = {
                    metric_name: "{:.6g}".format(old_run.data.metrics[metric_key])
                    for metric_name, metric_key in zip(metric_names, metric_keys)
                }
                leader_board_items.append(
                    {
                        "Run ID": old_run.info.run_id,
                        "Run Time": datetime.datetime.fromtimestamp(
                            old_run.info.start_time // 1000
                        ),
                        **metric_values,
                    }
                )
                if len(leader_board_items) >= 2:
                    break
```

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +171,69 @@ def _run(self, output_directory):
...
+
+        leader_board_items.sort(
+            key=lambda x: x[primary_metric_name],
+            reverse=metric_greater_is_better[primary_metric_name],
+        )
```
harupy, line 212:
```suggestion
```

`leader_board_items` is already sorted by the primary metric.

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +171,69 @@ def _run(self, output_directory):
...
+            run.info.run_id
+        )
+        latest_model_item = leader_board_items[latest_model_index]
+
+        # metric columns order: primary metric, then custom metrics, then builtin metrics.
+        metric_columns = (
+            [primary_metric_name]
+            + sorted(list(set(custom_metric_names) - {primary_metric_name}))
+            + sorted(list(set(metric_names) - set(custom_metric_names) - {primary_metric_name}))
+        )
```
harupy, line 225:
```suggestion
        # metric columns order: primary metric, then custom metrics, then builtin metrics.
        def sorter(m):
            if m == primary_metric_name:
                return 0
            elif m in custom_metric_names:
                return 1
            else:
                return 2
         
         metric_columns = sorted(metric_names, key=sorter)
```

mlflow/pipelines/steps/train.py
harupy, line 207:
```suggestion
                )
                if len(leader_board_items) >= 2:
                    break
```

We don't need to iterate through all the runs.

mlflow/pipelines/steps/train.py
jinzhang21: Not sure if the suggested approach will work within each category by sorting them alphabetically.

jinzhang21 (approved):
Looks good overall! Thanks, Weichen!

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,77 @@ def _run(self, output_directory):
...
+
+        primary_metric_order = (
+            "DESC" if self.metric_greater_is_better[self.primary_metric] else "ASC"
+        )
+        search_result = mlflow_client.search_runs(
+            experiment_ids=exp_id,
+            run_view_type=ViewType.ALL,
+            max_results=10000,
+            order_by=[
+                "attribute.start_time DESC",
```
harupy, line 188: Why do we sort by start time?

mlflow/pipelines/steps/train.py
harupy: If we want to sort the items within each category, we can return `(num, m)` in `sorter`.

WeichenXu123:
@jinzhang21 

Can we skip displaying the "total_run_number" in model rank ?
If so we can improve the performance by only search_runs for top K runs with optimal primary metric value.
and then we can compute the latest model rank to be (1, K) or larger than K, 
and we can define K to be a smaller value (e.g. 100),

so that we don't need to fetch 100000 historical runs which is quite expensive.

jinzhang21:
> @jinzhang21
> 
> Can we skip displaying the "total_run_number" in model rank ? If so we can improve the performance by only search_runs for top K runs with optimal primary metric value. and then we can compute the latest model rank to be (1, K) or larger than K, and we can define K to be a smaller value (e.g. 100),
> 
> so that we don't need to fetch 100000 historical runs which is quite expensive.

Sounds great to me!

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,103 @@ def _run(self, output_directory):
...
+            order_by=[f"metrics.{self.primary_metric} {primary_metric_order}"],
+        )
+
+        primary_metric_name = _get_primary_metric(self.step_config)
+        custom_metric_names = [cm["name"] for cm in _get_custom_metrics(self.step_config)]
+        metric_names = self.metric_greater_is_better.keys()
+        metric_keys = [f"{metric_name}_on_data_validation" for metric_name in metric_names]
+
+        leaderboard_items = []
+        for old_run in search_result:
```
harupy, line 198: Note the search result contains the latest run.

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,103 @@ def _run(self, output_directory):
...
+            ):
+                break
+
+        if i < len(leaderboard_items):
+            latest_model_item["Model Rank"] = f"{i + 1}"
+        else:
+            latest_model_item["Model Rank"] = f"> {i + 1}"
+
+        # metric columns order: primary metric, then custom metrics, then builtin metrics.
+        # metric columns order: primary metric, then custom metrics, then builtin metrics.
```
harupy, line 248:
```suggestion
        # metric columns order: primary metric, then custom metrics, then builtin metrics.
```

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,103 @@ def _run(self, output_directory):
...
+
+        for i, leaderboard_item in enumerate(leaderboard_items):
+            latest_value = latest_model_item[self.primary_metric]
+            historical_value = leaderboard_item[self.primary_metric]
+            if (primary_metric_greater_is_better and latest_value >= historical_value) or (
+                not primary_metric_greater_is_better and latest_value <= historical_value
+            ):
+                break
+
+        if i < len(leaderboard_items):
```
harupy, line 242: Looks like this condition is always false.

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,103 @@ def _run(self, output_directory):
...
+            latest_model_item["Model Rank"] = f"{i + 1}"
+        else:
+            latest_model_item["Model Rank"] = f"> {i + 1}"
```
harupy, line 245: You can use the for-else syntax here.

harupy, line 245: What does `> {i + 1}` mean?

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,103 @@ def _run(self, output_directory):
...
+        exp_id = _get_experiment_id()
+
+        primary_metric_greater_is_better = self.metric_greater_is_better[self.primary_metric]
+
+        primary_metric_order = "DESC" if primary_metric_greater_is_better else "ASC"
+
+        search_max_results = 100
+        search_result = mlflow_client.search_runs(
+            experiment_ids=exp_id,
+            run_view_type=ViewType.ALL,
```
harupy, line 187: Just to confirm, we also want to include deleted runs?

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,103 @@ def _run(self, output_directory):
...
+                        **{
+                            metric_name: old_run.data.metrics[metric_key]
+                            for metric_name, metric_key in zip(metric_names, metric_keys)
+                        },
+                    }
+                )
+
+        top_leaderboard_items = leaderboard_items[:2]
+        for i, top_leaderboard_item in enumerate(top_leaderboard_items):
+            top_leaderboard_item["Model Rank"] = f"{i + 1}"
```
harupy, line 215:
```suggestion
        top_leaderboard_items = {"Model Rank": i + 1, **top_leaderboard_item} in enumerate(top_leaderboard_items}
```

mlflow/pipelines/steps/train.py
WeichenXu123: I don't think so. If the routine does not run into L240 "break", then the i will be ==  `len(leaderboard_items)`

mlflow/pipelines/steps/train.py
WeichenXu123: Means the model rank is "larger than {i+1}" but we cannot get the precise number.

mlflow/pipelines/steps/train.py
WeichenXu123: How to set view type to exclude deleted runs but keep all other runs ?

harupy: See https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.search_runs

jinzhang21:
I did a quick test on the tab generated. It appears the rank is incorrect
Screen Shot 2022-06-28 at 2 04 26 PM
.

mlflow/pipelines/steps/train.py
jinzhang21: Agree with Haru that we should just search `ACTIVE_ONLY` runs.

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,104 @@ def _run(self, output_directory):
             raw_validation_df, prediction_result, self.target_col
         )
 
+        from mlflow.entities import ViewType
```
jinzhang21, line 175: Can we pull this section out as a separate function? It's getting really long.

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,104 @@ def _run(self, output_directory):
...
+            experiment_ids=exp_id,
+            run_view_type=ViewType.ALL,
+            max_results=search_max_results,
+            order_by=[f"metrics.{self.primary_metric} {primary_metric_order}"],
+        )
+
+        primary_metric_name = _get_primary_metric(self.step_config)
+        custom_metric_names = [cm["name"] for cm in _get_custom_metrics(self.step_config)]
+        metric_names = self.metric_greater_is_better.keys()
+        metric_keys = [f"{metric_name}_on_data_validation" for metric_name in metric_names]
```
jinzhang21, line 195: I'm not sure `metric_keys` simplifies the code much.

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,104 @@ def _run(self, output_directory):
...
+                            old_run.info.start_time // 1000
+                        ),
+                        **{
+                            metric_name: old_run.data.metrics[metric_key]
+                            for metric_name, metric_key in zip(metric_names, metric_keys)
+                        },
+                    }
+                )
+
+        top_leaderboard_items = leaderboard_items[:2]
```
jinzhang21, line 213: Can we inline this in L214?

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,104 @@ def _run(self, output_directory):
...
+
+        leaderboard_items_contains_latest_model = False
+        for i, leaderboard_item in enumerate(leaderboard_items):
+            latest_value = latest_model_item[self.primary_metric]
+            historical_value = leaderboard_item[self.primary_metric]
+            if (primary_metric_greater_is_better and latest_value >= historical_value) or (
+                not primary_metric_greater_is_better and latest_value <= historical_value
+            ):
+                leaderboard_items_contains_latest_model = True
+                break
```
jinzhang21, line 243: I don't understand what this section is doing, will take more time to process it later. Have to go.

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,104 @@ def _run(self, output_directory):
...
+        primary_metric_greater_is_better = self.metric_greater_is_better[self.primary_metric]
+
+        primary_metric_order = "DESC" if primary_metric_greater_is_better else "ASC"
+
+        search_max_results = 100
+        search_result = mlflow_client.search_runs(
+            experiment_ids=exp_id,
+            run_view_type=ViewType.ALL,
+            max_results=search_max_results,
+            order_by=[f"metrics.{self.primary_metric} {primary_metric_order}"],
```
harupy, line 189:
```suggestion
            order_by=[f"metrics.{self.primary_metric}_on_data_validation {primary_metric_order}"],
```

I think the order-by expression is missing a `_on_data_xxx` suffix.

mlflow/pipelines/steps/train.py
dbczumar: Done!

mlflow/pipelines/steps/train.py
dbczumar: Rewrote it slightly to make things clear. It determines whether or not the latest model is on the leaderboard and computes its rank accordingly; if it's not on the leaderboard, the rank is "something greater than the length of the leaderboard"

mlflow/pipelines/steps/train.py
dbczumar: Done

mlflow/pipelines/steps/train.py
dbczumar: Done!

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,104 @@ def _run(self, output_directory):
...
+        )
+
+        primary_metric_name = _get_primary_metric(self.step_config)
```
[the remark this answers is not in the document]
dbczumar: Yep, fixed!

mlflow/pipelines/steps/train.py
dbczumar: `metric_keys` is used in multiple places below, which is why it's helpful to define it up here

mlflow/pipelines/steps/train.py
```diff
@@ -158,6 +172,104 @@ def _run(self, output_directory):
...
+        custom_metric_names = [cm["name"] for cm in _get_custom_metrics(self.step_config)]
+        metric_names = self.metric_greater_is_better.keys()
+        metric_keys = [f"{metric_name}_on_data_validation" for metric_name in metric_names]
+
+        leaderboard_items = []
+        for old_run in search_result:
+            if all(metric_key in old_run.data.metrics for metric_key in metric_keys):
```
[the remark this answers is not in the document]
dbczumar: This is our solution for handling missing metrics. We only render leaderboard items if they share the same metrics. While imperfect, I don't think we have time to thoroughly handle the case where some metrics are missing from some runs given proximity to release.

dbczumar (commented):
@jinzhang21 I've addressed your comments and cleaned up the code a bit. Please take another look.

mlflow/pipelines/steps/train.py
dbczumar: Good catch! Fixed!

mlflow/pipelines/steps/train.py
```diff
@@ -166,13 +193,111 @@ def _run(self, output_directory):
...
+        for i, leaderboard_item in enumerate(leaderboard_items):
+            latest_value = latest_model_item[self.primary_metric]
+            historical_value = leaderboard_item[self.primary_metric]
+            if (primary_metric_greater_is_better and latest_value >= historical_value) or (
+                not primary_metric_greater_is_better and latest_value <= historical_value
+            ):
+                latest_model_item["Model Rank"] = str(i + 1)
+                break
+        else:
+            latest_model_item["Model Rank"] = f"> {len(leaderboard_items)}"
```
harupy, line 272:
```suggestion
        for i, leaderboard_item in enumerate(leaderboard_items):
            if latest_model_item["Run ID"] == leaderboard_item["Run ID"]:
                latest_model_item["Model Rank"] = str(i + 1)
                break
        else:
            latest_model_item["Model Rank"] = f"> {len(leaderboard_items)}"
```

Can we just compare run IDs here?

WeichenXu123: I don't think so.

We only query the top 100 historical runs, 
and the latest run ID might not be in the top 100 historical runs,
and even if the latest run ID might not be in the top 100 historical runs, the latest run rank might still be in the range of (0, 100).

mlflow/pipelines/steps/train.py
WeichenXu123: Just to confirm, if a historical run has been terminated, but it haven't been deleted, is it an active run ?

dbczumar: Yep! :)

mlflow/pipelines/steps/train.py
dbczumar: Makes sense!

mlflow/pipelines/steps/train.py
WeichenXu123: I was confused because `mlflow.active_run()` return the current active run which haven't ended. But the search_runs API "active_run" type also includes runs which has finished. @harupy

mlflow/pipelines/steps/train.py
harupy: Makes sense now. My approach doesn't work when `leaderboard_items` contains ties:

|run_id|primary metric value|rank (current implementation)|rank (suggested implementation)|
|:-|:-|:-|:-|
|1|1|1|1
|...|...|...|...
|98|98|98|98|
|99|99|99|99|
|100|99|99|99|
|101 (latest run)|100|99|> 100|

PR #6148, opened 2022-06-27
Add test to ensure MLP dataframe split is deterministic

WeichenXu123:

## Related Issues/PRs

#xxx

MLP Split is strictly deterministic across different python versions and different runs.
Add test to ensure MLP dataframe split is deterministic

## How is this patch tested?

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components

Interface

Language

Integrations

### DIFF
--- a/tests/pipelines/test_split_step.py
+++ b/tests/pipelines/test_split_step.py
@@ -6,7 +6,12 @@ import pandas as pd
 
 from mlflow.exceptions import MlflowException
 from mlflow.pipelines.utils.execution import _MLFLOW_PIPELINES_EXECUTION_DIRECTORY_ENV_VAR
-from mlflow.pipelines.steps.split import _get_split_df, _make_elem_hashable, SplitStep
+from mlflow.pipelines.steps.split import (
+    _get_split_df,
+    _hash_pandas_dataframe,
+    _make_elem_hashable,
+    SplitStep,
+)
 from unittest import mock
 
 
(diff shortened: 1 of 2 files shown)


tests/pipelines/test_split_step.py
```diff
@@ -42,9 +47,9 @@ def test_split_step_run(tmp_path):
     output_validation_df = pd.read_parquet(str(split_output_dir / "validation.parquet"))
     output_test_df = pd.read_parquet(str(split_output_dir / "test.parquet"))
 
-    assert np.isclose(len(output_train_df) / num_good_rows, split_ratios[0], rtol=0.1)
-    assert np.isclose(len(output_validation_df) / num_good_rows, split_ratios[1], rtol=0.1)
-    assert np.isclose(len(output_test_df) / num_good_rows, split_ratios[2], rtol=0.1)
+    assert len(output_train_df) == 551
+    assert len(output_validation_df) == 266
+    assert len(output_test_df) == 83
```
jinzhang21, line 52: Though this is obviously an improvement, but `551 + 266 + 83 = 900`. Since `split_ratios = [0.6, 0.3, 0.1]`, I would expect exact numbers from each split are like `540 + 270 + 90` instead. I think we can improve on the floating point calculation in converting ratios to dataset sizes to make it exact.

WeichenXu123: Because we generate hash for each row first, then we assign hash to buckets, we cannot ensure results number to be exact [540 + 270 + 90], but only can ensure the split deterministic.

jinzhang21: OK. Let's think about it a bit more. I'm OK to merge this PR as is.

jinzhang21 (approved):

tests/pipelines/test_split_step.py
WeichenXu123: @jinzhang21 

I think out an approach which can generate exact 540 + 270 + 90 split results:

1. generate hash for each row first (The same step with current code)
2. sort all rows by the hash value
3. cut the dataframe into 3 parts according to the exact split ratio.

To achieve the exact 540 + 270 + 90 behavior, the critical part is the sort step, but the main concern is it is a bit expensive, it requires shuffling when dataset is large or distributed.

jinzhang21: This is an interesting idea. It should work with intermediate data size but you are correct that when distributed, sorting is not a great solution in general. On the other hand, non-exact bucketing is mostly an issue when dataset is small. For large dataset, non-exact solution does not make a major difference in validation/evaluation outcome.

Issue #6096, opened 2022-06-17
[FR] [Roadmap] Support filtering experiments by tags


BenWilson2:
## MLflow Roadmap Item

This is an MLflow Roadmap item that has been prioritized by the MLflow maintainers. We’ve identified this feature as a highly requested addition to the MLflow package based on community feedback. 
We're seeking a community contribution for the implementation of this feature and will enthusiastically support the development and review of a submitted PR for this.

## Contribution Note

As with other roadmap items, there may be a desire for multiple contributors to work on an issue. While we don’t discourage collaboration, we strongly encourage that a primary contributor is assigned to roadmap issues to simplify the merging process. 
The items on the roadmap are of a high priority. Due to the wide-spread demand of roadmap features, we encourage potential contributors to only agree to take on the work of creating a PR, making changes, and ensuring that test coverage is adequately created for the feature if they are willing and able to see the implementation through to a merged state. 

## Feature scope

This roadmap feature’s complexity is classified as:

- [ ] `good-first-issue`: This feature is limited in complexity and effort required to implement.
- [ ] `simple`: This feature does not require a large amount of effort to implement and / or is clear enough to not need a design discussion with maintainers.
- [ ] `involved`: This feature will require a substantial amount of development effort but does not require an agreed-upon design from the maintainers. The feedback given during the PR phase may be involved and necessitate multiple iterations before approval. (Please bear with us as we collaborate with you to make a great contribution)
- [X] `design-recommended`: This is a substantial feature that should have a design document approved prior to working on an implementation (to save your time, not ours). After agreeing to work on this feature, a maintainer will be assigned to support you throughout the development process.

## Proposal Summary

This FR is for adding support for filtering search queries of experiments to include the ability to use tags as a query term arg (optional dictionary to search by a key's value OR optional set to search by key presence)

This implementation will require adding support for:
- the fluent API
- the client API (supporting both FileStore and SqlAlchemyStore)
- the REST API

Note: this change will be substantial and, as such, a RFC submission is not only recommended, but required.

## Motivation
> #### What is the use case for this feature? 
  Many requests have been made to add an abstracted layer above that of experiments and to provide the ability to use tags as a search filtering device for experiments with multi-team deployments of MLflow. 

> #### Why is this use case valuable to support for MLflow users in general? ^

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
[...]

PR #6171, opened 2022-06-30
Implement `SqlAlchemyStore.search_experiments`

harupy:

## Related Issues/PRs

#6096

This PR implements:
  - `SqlAlchemyStore.search_experiments`
  - `MlflowClient.search_experiments`
  - `mlflow.search_experiments`

## How is this patch tested?

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Add a new API searhc_expeirments that enables users to search for experiments.

Components

Interface

Language

Integrations


mlflow/store/tracking/rest_store.py
```diff
@@ -350,3 +360,13 @@ def _paginate_list_experiments(self, view_type):
...
+
+    def search_experiments(
+        self,
+        view_type=ViewType.ACTIVE_ONLY,
+        max_results=None,
+        filter_string=None,
+        order_by=None,
+        page_token=None,
+    ):
+        raise NotImplementedError("Not implemented yet")
```
dbczumar, line 372: Do we need to define `search_experiments` again in the `DatabricksRestStore` if the intended implementation will be identical to `RestStore`?

mlflow/tracking/client.py
```diff
@@ -396,6 +396,39 @@ def print_experiment_info(experiments):
...
+        view_type: int = ViewType.ACTIVE_ONLY,
+        max_results: Optional[int] = None,
+        filter_string: Optional[str] = None,
+        order_by: Optional[List[str]] = None,
+        page_token=None,
+    ) -> PagedList[Experiment]:
+        """
+        Search for experiments that match the specified search criteria.
+
+        :param view_type: Qualify requested type of experiments.
```
dbczumar, line 410: Can we be a bit more descriptive / specific about what this means? For example, here's the `view_type` docstring for the `search_runs` fluent API:

```
one of enum values ACTIVE_ONLY, DELETED_ONLY, or ALL runs defined in [mlflow.entities.ViewType](https://mlflow.org/docs/latest/python_api/mlflow.entities.html#mlflow.entities.ViewType).
```

Can we make this docstring change throughout the PR?

mlflow/tracking/client.py
```diff
@@ -396,6 +396,39 @@ def print_experiment_info(experiments):
...
+        :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                            passed, all experiments will be returned for the File and SQL backends.
+                            For the REST backend, the server will pick a maximum number of results
+                            to return.
+        :param filter_string: Filter query string, defaults to searching for all experiments.
```
dbczumar, line 415: Can we add some examples of valid `filter_strings`? Can we also add these to the fluent API docs?

mlflow/tracking/fluent.py
```diff
@@ -1041,6 +1041,44 @@ def pagination_wrapper_func(number_to_get, next_page_token):
...
+    :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                        passed, all experiments will be returned for the File and SQL backends.
+                        For the REST backend, the server will pick a maximum number of results
+                        to return.
+    :param filter_string: Filter query string, defaults to searching for all experiments.
+    :param order_by: List of column names with ASC|DESC annotation, to be used for ordering
+                     matching search results.
+    :return: A :py:class:`PagedList <mlflow.store.entities.PagedList>` of
+             :py:class:`Experiment <mlflow.entities.Experiment>` objects. The pagination token
+             for the next page can be obtained via the ``token`` attribute of the object.
```
dbczumar, line 1063: I don't think this return type is accurate - should just be a regular list of experiments.

mlflow/utils/search_utils.py
```diff
@@ -819,3 +819,69 @@ def parse_filter_for_registered_models(cls, filter_string):
         return cls._parse_filter_for_model_registry(
             filter_string, cls.VALID_SEARCH_KEYS_FOR_REGISTERED_MODELS
         )
+
+
+class SearchExperimentsUtils(SearchUtils):
+    VALID_SEARCH_ATTRIBUTE_KEYS = ("name",)
+    VALID_ORDER_BY_ATTRIBUTE_KEYS = ("name",)
```
dbczumar, line 826: We should also support ordering by ID and creation time

tests/tracking/fluent/test_fluent.py
```diff
@@ -357,6 +357,91 @@ def test_list_experiments(view_type, tmpdir):
...
+            for name in deleted_experiment_names
+        ]
+        session.add_all(active_experiments + deleted_experiments)
+        # Create some tags
+        experiment_tags = [
+            SqlExperimentTag(key="tag", value="x", experiment_id=1),
+            SqlExperimentTag(key="tag", value="x", experiment_id=2),
+            SqlExperimentTag(key="tag", value="y", experiment_id=3),
+        ]
+        session.add_all(experiment_tags)
```
dbczumar, line 396: Rather than using `store` directly to populate experiment content, can we set up the test by calling existing MLflow fluent / MLflow client APIs?

mlflow/utils/search_utils.py
dbczumar: Can we add tests for these too?

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -1271,3 +1331,61 @@ def _get_orderby_clauses(order_by_list, session):
...
+            elif comparator == "=":
+                f = attr == value
+            elif comparator == "!=":
+                f = attr != value
+            else:
+                raise MlflowException.invalid_parameter_value(
+                    f"Invalid comparator for attribute: {comparator}"
+                )
+            attribute_filters.append(f)
+        elif typ == "tag":
```
dbczumar, line 1357: Can we support `LIKE` / `ILIKE` comparators for tag-based search as well?

mlflow/utils/search_utils.py
```diff
@@ -819,3 +819,69 @@ def parse_filter_for_registered_models(cls, filter_string):
...
+                    f"Valid entity types are {valid_entity_types}"
+                )
+            identifier = cls._valid_entity_type(entity_type)
+
+        key = cls._trim_backticks(cls._strip_quotes(key))
+        if identifier == cls._ATTRIBUTE_IDENTIFIER and key not in valid_attributes:
+            raise MlflowException(
+                "Invalid attribute key '{}' specified. Valid keys "
+                "are '{}'".format(key, valid_attributes)
+            )
```
dbczumar, line 870: Can we add the `INVALID_PARAMETER_VALUE` error code?

mlflow/store/tracking/sqlalchemy_store.py
harupy: Sure!

mlflow/utils/search_utils.py
harupy: Thanks for the catch! Added.

tests/tracking/fluent/test_fluent.py
harupy: Compared the performance between the current approach and using `mlflow.create_experiment`:

- The existing approach: took 2.23s
- Using `mlflow.create_experiment`: took 5.89s

The difference is small. I'll update the code to use `mlflow.create_experiment` :)

harupy: Updated :)

mlflow/store/tracking/rest_store.py
harupy: I added this to silence the following pylint error:

```
> pylint mlflow/store/tracking/rest_store.py
************* Module mlflow.store.tracking.rest_store
mlflow/store/tracking/rest_store.py:326:0: W0223: Method 'search_experiments' is abstract in class 'RestStore' but is not overridden (abstract-method)
```

pylint still thinks `RestStore.search_experiments` must be overridden.

harupy: Ref: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/abstract-method.html

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -1271,3 +1331,61 @@ def _get_orderby_clauses(order_by_list, session):
         clauses.append(SqlRun.start_time.desc())
     clauses.append(SqlRun.run_uuid)
     return select_clauses, clauses, ordering_joins
+
+
+def _get_search_experiments_filter_clauses(parsed_filters, dialect):
```
WeichenXu123, line 1336: Q: Shall we resue the method `_get_attributes_filtering_clauses` to get attribute filters ?

mlflow/store/tracking/abstract_store.py
```diff
@@ -36,6 +36,34 @@ def list_experiments(self, view_type=ViewType.ACTIVE_ONLY, max_results=None, pag
...
+    ):
+        """
+        Search for experiments that match the specified search criteria.
+
+        :param view_type: Qualify requested type of experiments.
+        :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                            passed, all experiments will be returned for the File and SQL backends.
+                            For the REST backend, the server will pick a maximum number of results
+                            to return.
+        :param filter_string: Filter query string, defaults to searching for all experiments.
```
WeichenXu123, line 56: Let's describe which fields (attribute.*, tag.*) can be used in filter and which comparators are allowed.

mlflow/tracking/fluent.py
harupy: Thanks for the catch! It's not accurate. Fixed!

mlflow/tracking/client.py
harupy: Got it, I'll some examples and also describe supported attributes and operators.

mlflow/utils/search_utils.py
harupy: > We should also support ordering by ID and creation time

In OSS MLflow, the experiments table doesn't have a `creation_time` column.

https://github.com/mlflow/mlflow/blob/8693a26bf2acd418aa93aa5ea2e5536caa199fc3/tests/db/schemas/sqlite.sql#L8-L15

harupy: Experiment ID can be supported. I'll add it.

mlflow/store/tracking/sqlalchemy_store.py
harupy: Added support for `LIKE` and `ILIKE`.

mlflow/store/tracking/rest_store.py
harupy: Reported this issue in the pylint repository: https://github.com/PyCQA/pylint/issues/7124

mlflow/store/tracking/abstract_store.py
```diff
@@ -36,6 +36,53 @@ def list_experiments(self, view_type=ViewType.ACTIVE_ONLY, max_results=None, pag
...
+              - ``ILIKE``: Case-insensitive sensitive pattern match.
+
+            Logical operators
+              - ``AND``: Combines two sub-queries and returns True if both of them are True.
+
+        :param order_by: List of columns to order by. The ``order_by`` column can contain an
+                         optional ``DESC`` or ``ASC`` value (e.g., "name DESC"). The default is
+                         ``ASC`` so ``"name"`` is equivalent to ``"name ASC"``.
+        :param page_token: Token specifying the next page of results. It should be obtained from
+                            a ``search_experiments`` call.
```
harupy, line 79:
```suggestion
                           a ``search_experiments`` call.
```

mlflow/store/tracking/abstract_store.py
harupy: Done

mlflow/store/tracking/rest_store.py
harupy: One of the pylint maintaners replied to my comment: https://github.com/PyCQA/pylint/issues/7124#issuecomment-1174461952

`RestStore.search_experiments` is still an abstract method since it only contains `raise NotImplementedError`.

mlflow/tracking/client.py
harupy: Updated!

mlflow/utils/search_utils.py
harupy: Could we add a creation_time column?

mlflow/tracking/_tracking_service/client.py
```diff
@@ -155,6 +155,58 @@ def list_experiments(self, view_type=ViewType.ACTIVE_ONLY, max_results=None, pag
...
+        :param max_results: If passed, specifies the maximum number of experiments desired. If not
+                            passed, all experiments will be returned for the File and SQL backends.
+                            For the REST backend, the server will pick a maximum number of results
+                            to return.
+        :param filter_string:
+            Filter query string (e.g., ``"name = 'my_experiment'"``), defaults to searching for all
+            experiments. The following fields, comparators, and logical operators are supported.
+
+            Fields
+              - ``name``: Experiment name.
```
WeichenXu123, line 180: Do we need to use `attribute.name` ?
and also support experiment_id field ?

harupy: > Do we need to use attribute.name?

We don't need to. `name` is equivalent to `attribute.name`.

mlflow/store/tracking/rest_store.py
harupy: Can we raise `MlflowException`?

mlflow/tracking/_tracking_service/client.py
WeichenXu123: ok.
and `experiment_id` should be also a valid attribute field in filter, add it in doc

mlflow/utils/search_utils.py
```diff
@@ -97,14 +97,20 @@ def get_sql_filter_ops(cls, column, operator, dialect):
...
+
+        def case_sensitive_mysql_like(value):
             return sa.and_(column.like(value), column.op("LIKE BINARY", is_comparison=True)(value))
 
         sql_filter_ops = {
-            "=": column.__eq__,
-            "!=": column.__ne__,
-            "LIKE": mysql_like if dialect == MYSQL else column.like,
+            "=": case_sensitive_mysql_eq if dialect == MYSQL else column.__eq__,
+            "!=": case_sensitive_mysql_ne if dialect == MYSQL else column.__ne__,
```
WeichenXu123, line 112: good fix!

WeichenXu123 (approved):
LGTM

mlflow/tracking/_tracking_service/client.py
harupy: The current implementation doesn't support filtering by experiment ID.

WeichenXu123: ok

harupy: We can support filtering experiments with experiment ID if necessary though.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -338,6 +339,65 @@ def list_experiments(
...
+                next_token = SearchExperimentsUtils.create_page_token(final_offset)
+
+            return next_token
+
+        if max_results and max_results > SEARCH_MAX_RESULTS_THRESHOLD:
+            raise MlflowException(
+                "Invalid value for request parameter max_results. It must be at "
+                "most {}, but got value {}".format(SEARCH_MAX_RESULTS_THRESHOLD, max_results),
+                INVALID_PARAMETER_VALUE,
+            )
```
dbczumar, line 363: This check seems to imply that `max_results` can be `None`; in that case, what happens when we pass it to `limit()` on line 380?

The docstring for `AbstractStore.search_experiments()` indicates that, if max results is unspecified, the SqlAlchemyStore will return all results. If that's true, it's a bit strange that we throw an exception if `max_results` is specified and is too big, since we clearly support returning all available experiments by leaving `max_results` unspecified.

Shall we change the documentation of the AbstractStore method?

dbczumar: Looks like the MlflowClient also has the same message; shall we change it?

dbczumar (approved):
LGTM with a couple small comments. Awesome work, @harupy !

mlflow/store/tracking/sqlalchemy_store.py
harupy: > what happens when we pass it to limit() on line 380?

`limit(None)` would fetch all experiments ([related SO question](https://stackoverflow.com/questions/12836194/how-do-i-remove-a-limit-that-is-already-applied-to-a-query-object))

> it's a bit strange that we throw an exception if max_results is specified and is too big, since we clearly support returning all available experiments by leaving max_results unspecified.

Yeah it's strange. I'll update the MlflowClient and AbstractStore docstrings.

harupy: Can we

- Set the default value of `max_results` to `SEARCH_MAX_RESULTS_DEFAULT`
- Update the validation like this:

```python
        if not isinstance(max_results, int) and max_results < 1:
            raise MlflowException(
                "max_results must be a positive integer, but got {}".format(max_results),
                INVALID_PARAMETER_VALUE,
            )
       
        if max_results > SEARCH_MAX_RESULTS_THRESHOLD:
            raise MlflowException(
                "Invalid value for request parameter max_results. It must be at"
                " most {}, but got value {}".format(SEARCH_MAX_RESULTS_THRESHOLD, max_results),
                INVALID_PARAMETER_VALUE,
            )
```

- Update the doc like this:

```
        :param max_results: The maximum number of experiments desired. For the REST backend,
                            the server may apply its own limit.
```

?

dbczumar: Sounds good!

PR #6227, opened 2022-07-11
Implement `FileStore.search_experiments`

harupy:
## Related Issues/PRs

#6096

Implement `FileStore.search_experiments`.

## How is this patch tested?

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/store/tracking/file_store.py
+++ b/mlflow/store/tracking/file_store.py
@@ -1009,10 +1009,8 @@ def _get_search_experiments_filters(parsed_filters):
         comparator = f["comparator"]
         value = f["value"]
         if type_ == "attribute":
-            f = _comparator_to_func(comparator, value)
             filters.append(filter_by_attribute(key, comparator, value))
         elif type_ == "tag":
-            f = _comparator_to_func(comparator, value)
             filters.append(filter_by_tags(key, comparator, value))
         else:
             raise MlflowException.invalid_parameter_value(f"Invalid token type: {type_}")
--- a/mlflow/store/tracking/file_store.py
+++ b/mlflow/store/tracking/file_store.py
@@ -3,6 +3,7 @@ import logging
 import os
 import sys
 import shutil
+import re
 
 import uuid
 
@@ -23,12 +24,18 @@ from mlflow.entities.lifecycle_stage import LifecycleStage
 from mlflow.entities.run_info import check_run_is_active, check_run_is_deleted
 from mlflow.exceptions import MlflowException, MissingConfigException
 import mlflow.protos.databricks_pb2 as databricks_pb2
-from mlflow.protos.databricks_pb2 import INTERNAL_ERROR, RESOURCE_DOES_NOT_EXIST
+from mlflow.protos.databricks_pb2 import (
+    INTERNAL_ERROR,
+    RESOURCE_DOES_NOT_EXIST,
+    INVALID_PARAMETER_VALUE,
+)
 from mlflow.store.tracking import (
     DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH,
+    SEARCH_MAX_RESULTS_DEFAULT,
     SEARCH_MAX_RESULTS_THRESHOLD,
 )
 from mlflow.store.tracking.abstract_store import AbstractStore
+from mlflow.store.entities.paged_list import PagedList
 from mlflow.utils.validation import (
     _validate_metric_name,
     _validate_param_name,
@@ -61,6 +68,7 @@ from mlflow.utils.file_utils import (
     local_file_uri_to_path,
     path_to_local_file_uri,
 )
+from mlflow.utils.search_utils import SearchUtils, SearchExperimentsUtils
 from mlflow.utils.string_utils import is_string_type
 from mlflow.utils.uri import append_to_uri_path
 from mlflow.utils.mlflow_tags import MLFLOW_LOGGED_MODELS
@@ -244,9 +252,6 @@ class FileStore(AbstractStore):
                  :py:class:`Experiment <mlflow.entities.Experiment>` objects. The pagination token
                  for the next page can be obtained via the ``token`` attribute of the object.
         """
-        from mlflow.utils.search_utils import SearchUtils
-        from mlflow.store.entities.paged_list import PagedList
-
         _validate_list_experiments_max_results(max_results)
         self._check_root_dir()
         rsl = []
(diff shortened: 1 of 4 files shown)


mlflow/store/tracking/file_store.py
```diff
@@ -931,3 +970,81 @@ def record_logged_model(self, run_id, mlflow_model):
...
+
+def _get_search_experiments_filters(parsed_filters):
+    filters = []
+    for f in parsed_filters:
+        type_ = f["type"]
+        key = f["key"]
+        comparator = f["comparator"]
+        value = f["value"]
+        if type_ == "attribute":
+            f = _comparator_to_func(comparator, value)
```
WeichenXu123, line 1012: this line is useless.

mlflow/store/tracking/file_store.py
```diff
@@ -931,3 +970,81 @@ def record_logged_model(self, run_id, mlflow_model):
...
+            filters.append(filter_by_attribute(key, comparator, value))
+        elif type_ == "tag":
+            f = _comparator_to_func(comparator, value)
```
WeichenXu123, line 1015: ditto

mlflow/store/tracking/file_store.py
```diff
@@ -931,3 +970,81 @@ def record_logged_model(self, run_id, mlflow_model):
...
+
+
+def filter_by_attribute(key, comparator, value):
+    f = _comparator_to_func(comparator, value)
+    return lambda exp: f(getattr(exp, key))
+
+
+def filter_by_tags(key, comparator, value):
+    f = _comparator_to_func(comparator, value)
+    return lambda exp: any(f(tag[1]) for tag in filter(lambda tag: tag[0] == key, exp.tags.items()))
```
WeichenXu123, line 1001: Do we need a loop here?
`exp.tags` should be a dict,
and we can do:
```
if key in exp.tags:
  return f(exp.tags[key])
return False
```

mlflow/store/tracking/file_store.py
harupy: good catch, forgot to remove it. removed.

mlflow/store/tracking/file_store.py
```diff
@@ -931,3 +970,81 @@ def record_logged_model(self, run_id, mlflow_model):
...
+        elif type_ == "tag":
+            f = _comparator_to_func(comparator, value)
+            filters.append(filter_by_tags(key, comparator, value))
+        else:
+            raise MlflowException.invalid_parameter_value(f"Invalid token type: {type_}")
+    return filters
+
+
+# https://stackoverflow.com/a/56842689
+class _Reversor:
```
WeichenXu123, line 1023: Do we need to also implement `__le__`, `__gt__`, `__ge__` method ?

harupy: We don't: https://stackoverflow.com/questions/37693373/how-to-sort-a-list-with-two-keys-but-one-in-reverse-order/56842689#comment105013524_56842689

WeichenXu123: Nice. We can add the link in doc: https://docs.python.org/3/library/functools.html#functools.total_ordering

mlflow/store/tracking/file_store.py
harupy: Good suggestion, fixed

mlflow/store/tracking/file_store.py
harupy: > Total ordering is unnecessary for use as a key parameter in the sorted function. You only need == and <

mlflow/store/tracking/file_store.py
```diff
@@ -931,3 +970,91 @@ def record_logged_model(self, run_id, mlflow_model):
...
+        key = f["key"]
+        comparator = f["comparator"]
+        value = f["value"]
+        if type_ == "attribute":
+            filters.append(filter_by_attribute(key, comparator, value))
+        elif type_ == "tag":
+            filters.append(filter_by_tags(key, comparator, value))
+        else:
+            raise MlflowException.invalid_parameter_value(f"Invalid token type: {type_}")
+    return filters
```
dbczumar, line 1028: Shall we place methods like this in `SearchExperimentsUtils` (https://github.com/mlflow/mlflow/blob/e1f493b8de58a2ebbc4d8a3ef27ee2166f748586/mlflow/utils/search_utils.py#L831)?

mlflow/store/tracking/file_store.py
```diff
@@ -931,3 +970,91 @@ def record_logged_model(self, run_id, mlflow_model):
...
+def _create_matcher(comparator, value):
+    if comparator == "=":
+        return value.__eq__
+    elif comparator == "!=":
+        return value.__ne__
+    elif comparator in ("LIKE", "ILIKE"):
+        flag = re.IGNORECASE if comparator == "ILIKE" else 0
+        regex = re.compile(_like_pattern_to_regex(value), flag)
+        return regex.match
+    raise MlflowException.invalid_parameter_value(f"Invalid comparator: {comparator}")
```
dbczumar, line 991: Perhaps we want to place this in `SearchUtils`? The existing `FileStore.search_runs()` API just calls out to `SearchUtils`. Would be nice to encapsulate the code in a consistent location.

mlflow/store/tracking/file_store.py
```diff
@@ -931,3 +970,91 @@ def record_logged_model(self, run_id, mlflow_model):
...
+def filter_by_tags(key, comparator, value):
+    matcher = _create_matcher(comparator, value)
+
+    def has_tag_and_matched(experiment):
+        val = experiment.tags.get(key)
+        if val is not None:
+            return matcher(val)
+        return False
+
+    return has_tag_and_matched
```
dbczumar, line 1012: A lot of similar logic seems to exist in https://github.com/mlflow/mlflow/blob/e1f493b8de58a2ebbc4d8a3ef27ee2166f748586/mlflow/utils/search_utils.py#L395-L435. Anything we can reuse from there instead?

harupy: Updated the code to reuse the existing logic :)

mlflow/store/tracking/file_store.py
harupy: Removed this function while addressing https://github.com/mlflow/mlflow/pull/6227#discussion_r919155658.

mlflow/store/tracking/file_store.py
harupy: Removed this function while addressing https://github.com/mlflow/mlflow/pull/6227#discussion_r919155658.

mlflow/utils/search_utils.py
```diff
@@ -892,3 +899,91 @@ def parse_order_by_for_search_experiments(cls, order_by):
...
+        comparator = sed.get("comparator").upper()
+
+        if cls.is_attribute(key_type, comparator):
+            lhs = getattr(experiment, key)
+        elif cls.is_tag(key_type, comparator):
+            if key not in experiment.tags:
+                return False
+            lhs = experiment.tags.get(key, None)
+            if lhs is None:
+                return experiment
```
WeichenXu123, line 928: If the tag value is None, why return `experiment` ?
this method should return True or False.
Do I miss something ?

mlflow/utils/search_utils.py
```diff
@@ -892,3 +899,91 @@ def parse_order_by_for_search_experiments(cls, order_by):
...
         identifier = cls._get_identifier(token_value.strip(), cls.VALID_ORDER_BY_ATTRIBUTE_KEYS)
         return identifier["type"], identifier["key"], is_ascending
+
+    @classmethod
+    def is_attribute(cls, key_type, comparator):
+        if key_type == cls._ATTRIBUTE_IDENTIFIER:
+            if comparator not in cls.VALID_STRING_ATTRIBUTE_COMPARATORS:
+                raise MlflowException(
+                    "Invalid comparator '{}' not one of "
+                    "'{}".format(comparator, cls.VALID_STRING_ATTRIBUTE_COMPARATORS)
```
WeichenXu123, line 909: set error_code=INVALID_PARAMETER_VALUE

tests/store/tracking/test_file_store.py
```diff
@@ -162,6 +172,141 @@ def test_list_experiments_paginated(self):
...
+        experiment_names = ["a", "b"]
+        experiment_ids = self.create_experiments(experiment_names)
+        self.store.delete_experiment(experiment_ids[1])
+
+        experiments = self.store.search_experiments(view_type=ViewType.ACTIVE_ONLY)
+        assert [e.name for e in experiments] == ["a", "Default"]
+        experiments = self.store.search_experiments(view_type=ViewType.DELETED_ONLY)
+        assert [e.name for e in experiments] == ["b"]
+        experiments = self.store.search_experiments(view_type=ViewType.ALL)
+        assert [e.name for e in experiments] == ["b", "a", "Default"]
```
WeichenXu123, line 186: If we don't specify the ordering, is the returned experiment order undefined ? If so we should sort the result and then compare with expected result, to make the test deterministic.

mlflow/utils/search_utils.py
harupy: I think I forgot to remove it.

tests/store/tracking/test_file_store.py
harupy: We always sort experiments by experiment_id to break ties so the result is deterministic.

WeichenXu123: Suggestion:

Shall we  break ties  by sorting by "create_time desc" ?
So that when no order clause specified, it will return latest created experiments. This is more reasonable when there're many experiments.

WeichenXu123: ok. can we use `experiment_id desc` as break ties sorting ? New experiment ids are incremental.

dbczumar (approved):
LGTM! Thanks @harupy !

PR #6228, opened 2022-07-11
Jasperhg90feat pipeline yaml arg

### LINKED ISSUE #6209: Train and batch inference pipelines in same directory
Thanks for adding this feature to the MLFlow API!

Our situation is as follows:

- We want to share steps between train and inference pipelines
- To do this, we would need to be able to create multiple pipelines in a single directory
- However, we see [here](https://github.com/mlflow/mlflow/blob/0ab45a6e11f17484589a6195bd1fccb2e0a3c389/mlflow/pipelines/utils/__init__.py#L11) that the pipeline name is hard-coded. 

Is there a specific reason why we cannot simply pass the name of the pipeline ('pipeline_train.yaml', 'pipeline_inference.yaml') to the `Pipeline` mlflow class? How otherwise would you propose that users share steps between their jobs?

Happy to make a PR to propose implementing the above.

JasperHG90:
## Related Issues/PRs

This PR closes #6209. 

This PR adds a 'pipeline_yaml_file' input argument for MLFlow pipelines such that the user can specify which pipeline should be used in case that there are multiple pipelines in the same directory. The input argument defaults to 'pipeline.yaml'. We also added this to the CLI.

## How is this patch tested?

We added the default pipeline name to an [existing test](https://github.com/JasperHG90/mlflow/blob/master/tests/pipelines/test_pipeline.py#L58-L66) and added a test that should fail because it is looking for a pipeline that does not exist.

No.

Yes. Users can pass the name of their pipeline to the `Pipeline` class. Previously, this was not possible.

Components

### DIFF
--- a/mlflow/pipelines/cli.py
+++ b/mlflow/pipelines/cli.py
@@ -19,6 +19,18 @@ _CLI_ARG_PIPELINE_PROFILE = click.option(
 )
 
 
+_CLI_ARG_PIPELINE_FILE_NAME = click.option(
+    "--pipeline_file_name",
+    type=click.STRING,
+    default="pipeline.yaml",
+    required=False,
+    help=(
+        "The name of the pipeline YAML file to use. If not provided, will default to" +
+        " 'pipeline.yaml'"
+    ),    
+)
+
+
 @click.group("pipelines")
 def commands():
     """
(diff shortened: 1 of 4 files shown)


dbczumar:
Hi @JasperHG90. Thank you for filing this pull request. At this time, we recommend that each pipeline have its own directory. If you want to reference different pipelines, you can enter their corresponding directories before performing execution. We are also exploring more graceful support for specifying the path to a pipeline, which will alleviate this burden. At this time, we unfortunately cannot commit to introducing extra runtime flags for specifying the pipeline.yaml file. However, we should have an update to simplify these multi-pipeline workflows soon. Thank you for using MLflow Pipelines!

PR #6263, opened 2022-07-18
Allow logging of NA parameter values

nathaneastwood:

Some parameters for models can be `NA`, for example `period` within
`stats::arima()`. This PR allows for that.

## How is this patch tested?

- [x] No.

- [x] No.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/R/mlflow/R/tracking-runs.R
+++ b/mlflow/R/mlflow/R/tracking-runs.R
@@ -258,13 +258,14 @@ mlflow_log_param <- function(key, value, run_id = NULL, client = NULL) {
   c(client, run_id) %<-% resolve_client_and_run_id(client, run_id)
 
   key <- cast_string(key)
-  value <- cast_string(value)
+  value <- cast_string(value, allow_na = TRUE)
+  value <- ifelse(is.na(value), "NA", value)
 
   data <- list(
     run_uuid = run_id,
     run_id = run_id,
     key = key,
-    value = cast_string(value)
+    value = value
   )
   mlflow_rest("runs", "log-parameter", client = client, verb = "POST", data = data)
   mlflow_register_tracking_event("log_param", data)
--- a/mlflow/R/mlflow/tests/testthat/test-tracking-runs.R
+++ b/mlflow/R/mlflow/tests/testthat/test-tracking-runs.R
@@ -136,6 +136,8 @@ test_that("logging functionality", {
 
   mlflow_set_tag("tag_key", "tag_value")
   mlflow_log_param("param_key", "param_value")
+  mlflow_log_param("na", NA)
+  mlflow_log_param("nan", NaN)
 
   run <- mlflow_get_run()
   metrics <- run$metrics[[1]]
@@ -148,8 +150,8 @@ test_that("logging functionality", {
   run_id <- run$run_uuid
   tags <- run$tags[[1]]
   expect_identical("tag_value", tags$value[tags$key == "tag_key"])
-  expect_identical(run$params[[1]]$key, "param_key")
-  expect_identical(run$params[[1]]$value, "param_value")
+  expect_setequal(run$params[[1]]$key, c("na", "nan", "param_key"))
+  expect_setequal(run$params[[1]]$value, c("NA", "NaN", "param_value"))
 
   mlflow_delete_tag("tag_key", run_id)
   run <- mlflow_get_run()


mlflow/R/mlflow/R/tracking-runs.R
```diff
@@ -258,7 +258,7 @@ mlflow_log_param <- function(key, value, run_id = NULL, client = NULL) {
   c(client, run_id) %<-% resolve_client_and_run_id(client, run_id)
 
   key <- cast_string(key)
-  value <- cast_string(value)
+  value <- cast_string(value, allow_na = TRUE)
```
harupy, line 261: @nathaneastwood Can you add a test for this?

dbczumar (approved):
LGTM! Thanks @nathaneastwood

mlflow/R/mlflow/R/tracking-runs.R
harupy: @nathaneastwood We can add a test here:

```diff
diff --git a/mlflow/R/mlflow/tests/testthat/test-tracking-runs.R b/mlflow/R/mlflow/tests/testthat/test-tracking-runs.R
index 920f14909..e826fcc1f 100644
--- a/mlflow/R/mlflow/tests/testthat/test-tracking-runs.R
+++ b/mlflow/R/mlflow/tests/testthat/test-tracking-runs.R
@@ -136,6 +136,8 @@ test_that("logging functionality", {
 
   mlflow_set_tag("tag_key", "tag_value")
   mlflow_log_param("param_key", "param_value")
+  mlflow_log_param("na", NA)
+  mlflow_log_param("nan", NaN)
 
   run <- mlflow_get_run()
   metrics <- run$metrics[[1]]
```

nathaneastwood:
Turns out there are still some issues here which I discovered when adding the tests.

https://github.com/mlflow/mlflow/blob/master/mlflow/R/mlflow/R/tracking-runs.R#L267

The above line also uses `cast_string(value)` when really it should just be `value` since we already cast it. The problem, however, comes when I modify the tests suggested by @harupy. We need to modify the test to include

```r
   mlflow_log_param("na", NA)
   mlflow_log_param("nan", NaN)
```

as suggested but we also need to modify lines https://github.com/mlflow/mlflow/blob/master/mlflow/R/mlflow/tests/testthat/test-tracking-runs.R#L151-L152 to be

```r
  expect_identical(run$params[[1]]$key, c("na", "nan", "param_key"))
  expect_identical(run$params[[1]]$value, c("NA", "NaN", "param_value"))
```

However this isn't true since what we actually get is

```
# A tibble: 3 x 2
  key       value
  <chr>     <chr>
1 na        ""
2 nan       "NaN"
3 param_key "param_value"
```

This I guess makes sense since we are uploading(?) a "missing" value, i.e. `NA_character_`. I'm not too sure what the solution is here, I guess you want to upload `"NA"`, i.e. a string and ensure that when you download(?) the value, it returns `NA_character_`, a missing value. FWIW, this isn't anything to do with `tibble` as `tibble::tibble(a = c(NA_character_, "hello"))` returns a character column with a valid `NA`.

harupy:
> upload "NA", i.e. a string

This sounds good to me.

> and ensure that when you download(?) the value, it returns NA_character_, a missing value.

I think we can just return a string `"NA"`.

nathaneastwood:
Apologies I didn't get around to fixing this yesterday. I am currently travelling and do not have access to my laptop until next Tuesday. I will update this PR and rebase on master then.

harupy:
@nathaneastwood No worry, can I push commits to this PR?

harupy (approved):
LGTM!

nathaneastwood:
I'm not quite sure what's happening here. Locally, these tests run fine and are returned in the order I specify in the tests I altered. However in the CICD it seems there is a mismatch. I'll do some investigating.

nathaneastwood:
> It looks like the order of param keys/values is equal to the order of log_param calls.

Indeed, this is true. But on my Mac, the order is as I have specified in the tests which is why I did it that way. But logically I would have thought it would be the order of the `log_param()` calls which seems to be what the CICD pipeline expects. Digging into the code, the function `mlflow_get_run()` calls `mlflow_rest()` which queries the mlflow REST API, of course, and when the data are returned, they are indeed returned in the order my Mac shows. I even tried adding `mlflow_log_param("a", 1)` after the fact and this gives the order `"na", "nan", "a", "param_key"`. Anyway, I am not too sure why this occurs but I think we can just use `expect_setequal()` in the test to ensure that all values are returned and not worry to much about the order - I don't think the order is important?

harupy:
> I don't think the order is important?

The order is not important. Let's use `expect_setequal` to avoid unexpected flakiness.

harupy:
Enabled the auto-merge :)

PR #6297, opened 2022-07-21
update to run profiles on the first 100 dataframe columns

sunishsheth2009:
Update to run profiles on the first 100 data frame columns

## How is this patch tested?
- [x] Verified running this notebook https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/584469380481482 to verify that it reduces the time before and after the change

Before:
image

After:
image

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/utils/step.py
+++ b/mlflow/pipelines/utils/step.py
@@ -124,7 +124,7 @@ def get_pandas_data_profile(data_frame, title: str):
     max_cols = min(data_frame.columns.size, _MAX_PROFILE_COL_SIZE)
     max_rows = min(max(max_cells // max_cols, 1), _MAX_PROFILE_ROW_SIZE)
     truncated_df = data_frame.drop(columns=data_frame.columns[max_cols:]).sample(
-        n=max_rows, ignore_index=True, random_state=42
+        n=max_rows, ignore_index=True, random_state=42, replace=True
     )
     if (
         data_frame.size != max_cells
--- a/mlflow/pipelines/utils/step.py
+++ b/mlflow/pipelines/utils/step.py
@@ -122,14 +122,14 @@ def get_pandas_data_profile(data_frame, title: str):
 
     max_cells = min(data_frame.size, _MAX_PROFILE_CELL_SIZE)
     max_cols = min(data_frame.columns.size, _MAX_PROFILE_COL_SIZE)
-    max_rows = min(max(max_cells // max_cols, 1), _MAX_PROFILE_ROW_SIZE)
+    max_rows = min(max(max_cells // max_cols, 1), len(data_frame), _MAX_PROFILE_ROW_SIZE)
     truncated_df = data_frame.drop(columns=data_frame.columns[max_cols:]).sample(
-        n=max_rows, ignore_index=True, random_state=42, replace=True
+        n=max_rows, ignore_index=True, random_state=42
     )
     if (
-        data_frame.size != max_cells
-        or data_frame.columns.size != max_cols
-        or max(max_cells // max_cols, 1) != max_rows
+        max_cells == _MAX_PROFILE_CELL_SIZE
+        or max_cols == _MAX_PROFILE_COL_SIZE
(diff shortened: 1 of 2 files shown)


dbczumar (approved):
LGTM! We print a message indicating that we're truncating to 100, right?

mlflow/pipelines/utils/step.py
```diff
@@ -116,8 +116,20 @@ def get_pandas_data_profile(data_frame, title: str):
     max_cols = min(data_frame.columns.size, _MAX_PROFILE_COL_SIZE)
     max_rows = min(max(max_cells // max_cols, 1), _MAX_PROFILE_ROW_SIZE)
     truncated_df = data_frame.drop(columns=data_frame.columns[max_cols:]).sample(
-        n=max_rows, ignore_index=True, random_state=42
+        n=max_rows, ignore_index=True, random_state=42, replace=True
```
dbczumar, line 119: I don't think we should use `replace=True`; this seems like it could sample the same row twice. See https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sample.html.

tests/pipelines/test_step_utils.py
```diff
@@ -75,7 +75,7 @@ def test_get_merged_eval_metrics_works(
         (DataFrame(np.arange(160).reshape(8, 20)), 1000, 20, 8, 20, 8),
         (DataFrame(np.arange(160).reshape(8, 20)), 1000, 21, 8, 20, 8),
         (DataFrame(np.arange(160).reshape(8, 20)), 1000, 20, 9, 20, 8),
-        (DataFrame(np.arange(160).reshape(8, 20)), 10, 5, 9, 5, 2),
+        (DataFrame(np.arange(160).reshape(8, 20)), 10, 5, 9, 5, 8),
```
dbczumar, line 78: In this case, the expected_rows * expected_columns  exceeds max_cells, so I don't think the adjustment to https://github.com/mlflow/mlflow/pull/6297/files#diff-e02b9988953b168d7b15748927274c8a44225b3caeb346ce655b90b79c4aa815R125 is correct

sunishsheth2009: Basically I was trying to solve for the case where

we have 2000 cells, 1000 columns and 2row.
Now the Max is 1000 cells, 100 columns.  Based on the old math, it would be 10rows in the sample. The number of available rows are 2. So sampling 10rows would be a lot and incorrect. 

That caused me to change the above math.


Now in the new way, I added:

```
max_rows = min(
        max(max_cells // max_cols, 1),
        (data_frame.size // data_frame.columns.size),
        _MAX_PROFILE_ROW_SIZE,
    )
    ```
    
    Which should hopefully catch the other usecase as well. Let me add a test for what I was saying above.

File: mlflow/pipelines/utils/step.py [hunk 3]
```diff
@@ -122,10 +122,26 @@ def get_pandas_data_profile(data_frame, title: str):
 
     max_cells = min(data_frame.size, _MAX_PROFILE_CELL_SIZE)
     max_cols = min(data_frame.columns.size, _MAX_PROFILE_COL_SIZE)
-    max_rows = min(max(max_cells // max_cols, 1), _MAX_PROFILE_ROW_SIZE)
+    max_rows = min(
+        max(max_cells // max_cols, 1),
+        (data_frame.size // data_frame.columns.size),
+        _MAX_PROFILE_ROW_SIZE,
+    )
```
dbczumar, line 129: As proposed offline, I think it may be simpler for us to skip sampling if `len(df) < max_rows`.

sunishsheth2009: Spoke offline that this might be a bit easier rather than adding a if check:
```
   max_rows = min(
        max(max_cells // max_cols, 1),
        len(data_frame),
        _MAX_PROFILE_ROW_SIZE,
    )
 ```

PR #6310, opened 2022-07-24
[Model Validation] Disable logging metrics and artifacts for baseline model evaluation

zhe-db:
Model Validation - ML-23673 and ML-23676

- Add baseline_model and validation_thresholds argument to MLflow.evaluate API.
- Add evaluation of baseline models during model validation.
- Add `is_baseline_model` flag to `evaluator_config` for disabling logging metrics and disable generating and logging artifacts for evaluation of baseline model during model validation.

## How is this patch tested?
Added unit tests for
- Parametrizing the existing unit tests for testing if evaluation has the same behavior for candidate model during model validation.
- If evaluation of candidate model with is_baseline_model flag set disables logging metrics and generating artifacts works for.
  - Linear Regressor
  - Binary classifier
  - Multiclass classifier
  - Spark Linear Regressor
  - SVM model
  - Pipeline model

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -477,6 +477,7 @@ class ModelEvaluator(metaclass=ABCMeta):
         run_id,
         evaluator_config,
         custom_metrics=None,
+        baseline_model=None,
         **kwargs,
     ):
         """
@@ -497,8 +498,12 @@ class ModelEvaluator(metaclass=ABCMeta):
         :param custom_metrics: A list of callable custom metric functions.
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
-        :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
-                 evaluation results.
+        :param baseline_model: (Optional) A string URI referring to a MLflow model as baseline model
+                                          to be compared with the candidate model
+                                          for model validation.
+                                          (pyfunc model instance is not allowed)
+        :return: A tuple of :py:class:`mlflow.models.EvaluationResult` instance containing
+                 evaluation results for candidate model and baseline model (if provided else None).
         """
         raise NotImplementedError()
 
@@ -643,6 +648,7 @@ def _evaluate(
     evaluator_name_list,
     evaluator_name_to_conf_map,
     custom_metrics,
+    baseline_model,
 ):
     """
     The public API "evaluate" will verify argument first, and then pass normalized arguments
@@ -657,10 +663,9 @@ def _evaluate(
     client = MlflowClient()
     model_uuid = model.metadata.model_uuid
 
-    if not _is_eval_for_baseline_model(evaluator_name_to_conf_map):
-        dataset._log_dataset_tag(client, run_id, model_uuid)
+    dataset._log_dataset_tag(client, run_id, model_uuid)
 
-    eval_results = []
+    candidate_model_eval_results, baseline_model_eval_results = [], []
     for evaluator_name in evaluator_name_list:
         config = evaluator_name_to_conf_map.get(evaluator_name) or {}
         try:
@@ -672,30 +677,49 @@ def _evaluate(
         _last_failed_evaluator = evaluator_name
         if evaluator.can_evaluate(model_type=model_type, evaluator_config=config):
             _logger.info(f"Evaluating the model with the {evaluator_name} evaluator.")
-            result = evaluator.evaluate(
+            candidate_model_eval_result, baseline_model_eval_result = evaluator.evaluate(
                 model=model,
                 model_type=model_type,
                 dataset=dataset,
                 run_id=run_id,
                 evaluator_config=config,
                 custom_metrics=custom_metrics,
+                baseline_model=baseline_model,
             )
-            eval_results.append(result)
+            candidate_model_eval_results.append(candidate_model_eval_result)
+            baseline_model_eval_results.append(baseline_model_eval_result)
 
     _last_failed_evaluator = None
 
-    if len(eval_results) == 0:
+    if len(candidate_model_eval_results) == 0:
         raise ValueError(
             "The model could not be evaluated by any of the registered evaluators, please "
             "verify that the model type and other configs are set correctly."
         )
 
-    merged_eval_result = EvaluationResult(dict(), dict())
-    for eval_result in eval_results:
-        merged_eval_result.metrics.update(eval_result.metrics)
-        merged_eval_result.artifacts.update(eval_result.artifacts)
+    if baseline_model and len(baseline_model_eval_results) == 0:
+        raise ValueError(
+            "The baseline model could not be evaluated by any of the registered evaluators, please "
+            "verify that the model type and other configs are set correctly."
+        )
 
-    return merged_eval_result
+    candidate_model_merged_eval_result = EvaluationResult(dict(), dict())
+    for eval_result in candidate_model_eval_results:
+        if not eval_result:
+            continue
+        candidate_model_merged_eval_result.metrics.update(eval_result.metrics)
+        candidate_model_merged_eval_result.artifacts.update(eval_result.artifacts)
+
+    if not baseline_model:
+        return candidate_model_merged_eval_result, None
+
+    baseline_model_merged_eval_result = EvaluationResult(dict(), dict())
+    for eval_result in baseline_model_eval_results:
+        if not eval_result:
+            continue
+        baseline_model_merged_eval_result.metrics.update(eval_result.metrics)
+        baseline_model_merged_eval_result.artifacts.update(eval_result.artifacts)
+    return candidate_model_merged_eval_result, baseline_model_merged_eval_result
 
 
 @experimental
@@ -977,7 +1001,7 @@ def evaluate(
     )
 
     with _start_run_or_reuse_active_run() as run_id:
-        candidate_model_eval_result = _evaluate(
+        candidate_model_eval_results, baseline_model_eval_results = _evaluate(
             model=model,
             model_type=model_type,
             dataset=dataset,
@@ -985,25 +1009,13 @@ def evaluate(
             evaluator_name_list=evaluator_name_list,
             evaluator_name_to_conf_map=evaluator_name_to_conf_map,
             custom_metrics=custom_metrics,
+            baseline_model=baseline_model,
         )
-
         if not baseline_model:
-            return candidate_model_eval_result
+            return candidate_model_eval_results
 
-        evaluator_name_to_conf_map_with_baseline_flag = _add_is_baseline_model_flag_to_eval_config(
-            evaluator_name_to_conf_map
-        )
-
-        baseline_model_eval_result = _evaluate(
-            model=baseline_model,
-            model_type=model_type,
-            dataset=dataset,
-            run_id=run_id,
-            evaluator_name_list=evaluator_name_list,
-            evaluator_name_to_conf_map=evaluator_name_to_conf_map_with_baseline_flag,
-            custom_metrics=custom_metrics,
-        )
         # TODO: Add Model Validation here
-        if validation_thresholds and baseline_model_eval_result:
+        if validation_thresholds and baseline_model_eval_results:
             pass
-        return candidate_model_eval_result
+
+        return candidate_model_eval_results
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -160,7 +160,7 @@ class EvaluationResult:
     @property
     def baseline_model_metrics(self) -> Dict[str, Any]:
         """
-        A dictionary mapping scalar metric names to scalar metric values for baseline model
+        A dictionary mapping scalar metric names to scalar metric values for the baseline model
         """
         return self._baseline_model_metrics
 
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -519,9 +519,9 @@ class ModelEvaluator(metaclass=ABCMeta):
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
         :param baseline_model: (Optional) A string URI referring to a MLflow model with the pyfunc
-                                          flavor as baseline model to be compared with the
-                                          candidate model for model validation.
-                                          (pyfunc model instance is not allowed)
+                                          flavor as a baseline model to be compared with the
+                                          candidate model (specified by the `model` param) for model
+                                          validation. (pyfunc model instance is not allowed)
         :return: A tuple of :py:class:`mlflow.models.EvaluationResult` instance containing
                  evaluation results for candidate model and baseline model (if provided else None).
         """
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -3,6 +3,7 @@ import mlflow
 import hashlib
 import json
 import os
+from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
 from mlflow.tracking.client import MlflowClient
 from contextlib import contextmanager
 from mlflow.exceptions import MlflowException
@@ -90,9 +91,10 @@ class EvaluationResult:
     both scalar metrics and output artifacts such as performance plots.
     """
 
-    def __init__(self, metrics, artifacts):
+    def __init__(self, metrics, artifacts, baseline_model_metrics=None):
         self._metrics = metrics
         self._artifacts = artifacts
+        self._baseline_model_metrics = baseline_model_metrics
 
     @classmethod
     def load(cls, path):
@@ -154,6 +156,13 @@ class EvaluationResult:
         artifact content and location information
         """
         return self._artifacts
+
+    @property
+    def baseline_model_metrics(self) -> Dict[str, Any]:
+        """
+        A dictionary mapping scalar metric names to scalar metric values for baseline model
+        """
+        return self._baseline_model_metrics
 
 
 _cached_mlflow_client = None
@@ -249,9 +258,15 @@ class EvaluationDataset:
         import pandas as pd
 
         if name is not None and '"' in name:
-            raise ValueError(f'Dataset name cannot include a double quote (") but got {name}')
+            raise MlflowException(
+                message=f'Dataset name cannot include a double quote (") but got {name}',
+                error_code=INVALID_PARAMETER_VALUE,
+            )
         if path is not None and '"' in path:
-            raise ValueError(f'Dataset path cannot include a double quote (") but got {path}')
+            raise MlflowException(
+                message=f'Dataset path cannot include a double quote (") but got {path}',
+                error_code=INVALID_PARAMETER_VALUE,
+            )
 
         self._user_specified_name = name
         self._path = path
@@ -271,33 +286,37 @@ class EvaluationDataset:
             pass
 
         if feature_names is not None and len(set(feature_names)) < len(list(feature_names)):
-            raise ValueError(
-                "`feature_names` argument must be a list containing unique feature names."
+            raise MlflowException(
+                message="`feature_names` argument must be a list containing unique feature names.",
+                error_code=INVALID_PARAMETER_VALUE,
             )
 
         if isinstance(data, (np.ndarray, list)):
             if not isinstance(targets, (np.ndarray, list)):
-                raise ValueError(
-                    "If data is a numpy array or list of evaluation features, "
-                    "`targets` argument must be a numpy array or list of evaluation labels."
+                raise MlflowException(
+                    message="If data is a numpy array or list of evaluation features, "
+                    "`targets` argument must be a numpy array or list of evaluation labels.",
+                    error_code=INVALID_PARAMETER_VALUE,
                 )
             if isinstance(data, list):
                 data = np.array(data)
 
             if len(data.shape) != 2:
-                raise ValueError(
-                    "If the `data` argument is a numpy array, it must be a 2 dimension array "
-                    "and second dimension represent the number of features. If the `data` "
+                raise MlflowException(
+                    message="If the `data` argument is a numpy array, it must be a 2 dimension"
+                    " array and second dimension represent the number of features. If the `data` "
                     "argument is a list, each of its element must be a feature array of "
-                    "numpy array or list and all element must has the same length."
+                    "numpy array or list and all element must has the same length.",
+                    error_code=INVALID_PARAMETER_VALUE,
                 )
 
             self._features_data = data
             self._labels_data = targets if isinstance(targets, np.ndarray) else np.array(targets)
 
             if len(self._features_data) != len(self._labels_data):
-                raise ValueError(
-                    "The input features example rows must be the same length with labels array."
+                raise MlflowException(
+                    message="The input features example rows must be the same length with labels array.",
+                    erorr_code=INVALID_PARAMETER_VALUE,
                 )
 
             num_features = data.shape[1]
@@ -305,7 +324,10 @@ class EvaluationDataset:
             if feature_names is not None:
                 feature_names = list(feature_names)
                 if num_features != len(feature_names):
-                    raise ValueError("feature name list must be the same length with feature data.")
+                    raise MlflowException(
+                        message="feature name list must be the same length with feature data.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
                 self._feature_names = feature_names
             else:
                 self._feature_names = [
@@ -314,10 +336,11 @@ class EvaluationDataset:
                 ]
         elif isinstance(data, self._supported_dataframe_types):
             if not isinstance(targets, str):
-                raise ValueError(
-                    "If data is a Pandas DataFrame or Spark DataFrame, `targets` argument must "
-                    "be the name of the column which contains evaluation labels in the `data` "
-                    "dataframe."
+                raise MlflowException(
+                    message="If data is a Pandas DataFrame or Spark DataFrame, `targets` argument "
+                    "must be the name of the column which contains evaluation labels in the `data`"
+                    " dataframe.",
+                    error_code=INVALID_PARAMETER_VALUE,
                 )
             if self._spark_df_type and isinstance(data, self._spark_df_type):
                 _logger.warning(
@@ -337,9 +360,10 @@ class EvaluationDataset:
                 self._features_data = data.drop(targets, axis=1, inplace=False)
                 self._feature_names = list(self._features_data.columns)
         else:
-            raise ValueError(
-                "The data argument must be a numpy array, a list or a Pandas DataFrame, or "
-                "spark DataFrame if pyspark package installed."
+            raise MlflowException(
+                message="The data argument must be a numpy array, a list or a Pandas DataFrame, or "
+                "spark DataFrame if pyspark package installed.",
+                error_code=INVALID_PARAMETER_VALUE,
             )
 
         # generate dataset hash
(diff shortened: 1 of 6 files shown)


tests/models/test_evaluation.py
```diff
@@ -713,21 +786,178 @@ def test_evaluator_interface(multiclass_logistic_regressor_model_uri, iris_datas
...
+            evaluator1_config_baseline.update({"is_baseline_model": True})
+        with mock.patch.object(
+            FakeEvauator1, "can_evaluate", return_value=True
+        ) as mock_can_evaluate, mock.patch.object(
+            FakeEvauator1, "evaluate", return_value=evaluator1_return_value
+        ) as mock_evaluate:
+            classifier_model = mlflow.pyfunc.load_model(multiclass_logistic_regressor_model_uri)
+            baseline_model = None
+            with mlflow.start_run() as run:
+                if valid_baseline_model_uri:
```
dbczumar, line 834: Can we split this if / elif / else cases into separate tests to make each test more modular and debuggable?

tests/models/test_evaluation.py
```diff
@@ -176,6 +176,10 @@ def convert_num_to_label(x):
 
 @pytest.fixture
 def pipeline_model_uri():
+    return get_pipeline_modlel_uri()
+
+
+def get_pipeline_modlel_uri():
```
dbczumar, line 182: Do we need to extract this into a separate method that can be called without a fixture? Because the method doesn't take any arguments, it doesn't seem necessary; test cases can just use the fixture.

tests/models/test_default_evaluator.py
```diff
@@ -65,20 +68,120 @@ def assert_dict_equal(d1, d2, rtol):
...
+        if not test_is_baseline_model_flag:
+            assert np.isclose(
+                expected_metrics[metric_key],
+                logged_metrics[metric_key + dataset_log_key],
+                rtol=1e-3,
+            )
+        assert np.isclose(expected_metrics[metric_key], result_metrics[metric_key], rtol=1e-3)
+
+
+def artifacts_check_helper(
```
dbczumar, line 131: Are there more descriptive names we can use for these methods that would make it obvious what we're checking? I also think it's acceptable to split these into multiple methods for readability.

mlflow/models/evaluation/base.py
```diff
@@ -481,6 +490,10 @@ def evaluate(
         :param run_id: The ID of the MLflow Run to which to log results.
         :param evaluator_config: A dictionary of additional configurations for
                                  the evaluator.
+                                 evaluator_config may contain
+                                 - is_baseline_model: A boolean indicating whether the evaulation
+                                   is for baseline model. For baseline model, no artifact will
+                                   be generated and no metric will be logged.
```
dbczumar, line 496: This seems to imply that we intend to have `mlflow.evaluate()` call `default_evaluator.evaluate()` twice; once with `is_baseline_model=False` and once with `is_baseline_model=True`. I don't think we should be doing this. Instead, we should make a single call to `default_evaluator.evaluate()` and pass in the baseline model as its own argument (e.g. `baseline_model`).

dbczumar (changes requested):
@zhe-db Following from https://github.com/mlflow/mlflow/pull/6310/files#r931485573, I think it would be better to introduce a `baseline_model` parameter to the evaluator abstract API, rather than `is_baseline_model`. Please let me know if you have questions.

tests/models/test_evaluation.py
zhe-db: This one I am kinda stuck because I put those into functions is so that the parameterized `baseline_model_uri` fixture need to call these models to get different models; pytest does not allow call fixture directly; hence I made those into functions to be callable

tests/models/test_default_evaluator.py
zhe-db: Renamed them and split them into two functions for better readability.

mlflow/models/evaluation/base.py
```diff
@@ -481,11 +491,19 @@ def evaluate(
...
         :param custom_metrics: A list of callable custom metric functions.
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
-        :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
-                 evaluation results.
+        :param baseline_model: (Optional) A string URI referring to a MLflow model as baseline model
+                                          to be compared with the candidate model
```
zhe-db, line 502: Only for testing purposes (testing if evaluate for baseline model disable logging metrics and artifacts)

mlflow/models/evaluation/base.py
zhe-db: Made changes as directed, `ModelEvaluator.evaluate` now takes optional param `baseline_model` and return tuple of `EvaluationResult`

mlflow/models/evaluation/base.py
```diff
@@ -481,11 +490,19 @@ def evaluate(
         :param run_id: The ID of the MLflow Run to which to log results.
         :param evaluator_config: A dictionary of additional configurations for
                                  the evaluator.
+                                 evaluator_config may contain
+                                 - is_baseline_model: A boolean indicating whether the evaulation
+                                   is for baseline model. For baseline model, no artifact will
+                                   be generated and no metric will be logged.
```
dbczumar, line 496: Can we remove this docstring update?

mlflow/models/evaluation/base.py
```diff
@@ -481,11 +490,19 @@ def evaluate(
...
         :param custom_metrics: A list of callable custom metric functions.
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
-        :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
-                 evaluation results.
+        :param baseline_model: (Optional) A string URI referring to a MLflow model as baseline model
```
dbczumar, line 500: Can we clarify that the `model_uri` must refer to an MLflow Model with the pyfunc flavor?

mlflow/models/evaluation/base.py
```diff
@@ -481,11 +490,19 @@ def evaluate(
...
+                                          to be compared with the candidate model
+                                          for model validation.
+                                          (pyfunc model instance is not allowed)
+        :return: A tuple of :py:class:`mlflow.models.EvaluationResult` instance containing
+                 evaluation results for candidate model and baseline model (if provided else None).
```
dbczumar, line 505: Could we extend `EvaluationResult` to include information about baseline model performance, rather than returning a tuple of multiple evaluation results where the second element is sometimes None?

mlflow/models/evaluation/base.py
```diff
@@ -909,6 +956,12 @@ def scatter_plot(eval_df, builtin_metrics, artifacts_dir):
             "The model argument must be a string URI referring to an MLflow model or "
             "an instance of `mlflow.pyfunc.PyFuncModel`."
         )
+    if isinstance(baseline_model, str):
+        baseline_model = mlflow.pyfunc.load_model(baseline_model)
+    elif baseline_model is not None:
+        raise ValueError(
```
dbczumar, line 962: Can we raise `MlflowException` with error code `INVALID_PARAMETER_VALUE`?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -944,7 +950,9 @@ def _evaluate_classifier(self):
                 lambda: plot_lift_curve(self.y, self.y_probs),
                 "lift_curve_plot",
             )
-
+        # Skip confusion matrix for baseline model evaluation
+        if self.is_baseline_model:
+            return self._log_and_return_evaluation_result()
```
dbczumar, line 955: Does `_log_and_return_evaluation_result()` still log an EvaluationResult instance to MLflow? If so, we probably don't want to log baseline results to MLflow.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -636,9 +640,10 @@ def _log_model_explainability(self):
             _logger.debug("", exc_info=True)
             return
         try:
-            mlflow.shap.log_explainer(
-                explainer, artifact_path=_gen_log_key("explainer", self.dataset_name)
-            )
+            if not self.is_baseline_model:
```
dbczumar, line 643: Our model validation only depends on scalar metrics, right? If so, I don't think we should call `_log_model_explainability()` at all if `is_baseline_model` is True.

In general, I think that adding all of these checks for `is_baseline_model` to determine whether or not we log to MLflow Tracking is somewhat brittle / error-prone. Instead, can we split `_evaluate()` into multiple methods, such as `_get_model_predictions()`, `_compute_metrics()` (with an argument for `log_to_mlflow_tracking=True/False`), and `_compute_explanations()`. For the candidate model, we'll call all 3 methods and pass `log_to_mlflow_tracking=True` to `_compute_metrics()`. For the baseline model, we'll only call `_get_model_predictions()` and `_compute_metrics()` with `log_to_mlflow_tracking=False`.

Let me know what you think here.

tests/models/test_evaluation.py
dbczumar: Got it - makes sense! 

```suggestion
def get_pipeline_model_uri():
```

dbczumar (commented):
@zhe-db Awesome progress! The developer-facing `evaluate()` API is much cleaner, and the tests are much more modular / readable. Left a few more comments. Let me know if you have questions :)

mlflow/models/evaluation/default_evaluator.py
zhe-db: Agree! I refactored `DefaultEvaluator.evalaute` to multiple methods: `_generate_model_predictions`, `_compute_metrics`, `_evaluate_custom_metrics_and_log_produced_artifacts` and `_log_metrics_and_artifacts`.

mlflow/models/evaluation/base.py
```diff
@@ -155,6 +157,13 @@ def artifacts(self) -> Dict[str, "mlflow.models.EvaluationArtifact"]:
         """
         return self._artifacts
 
+    @property
+    def baseline_model_metrics(self) -> Dict[str, Any]:
```
dbczumar, line 161:
```suggestion
    def baseline_model_metrics(self) -> Optional[Dict[str, Any]]:
```

mlflow/models/evaluation/base.py
```diff
@@ -484,8 +518,12 @@ def evaluate(
...
         :param kwargs: For forwards compatibility, a placeholder for additional arguments that
                        may be added to the evaluation interface in the future.
-        :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
-                 evaluation results.
+        :param baseline_model: (Optional) A string URI referring to a MLflow model with the pyfunc
+                                          flavor as baseline model to be compared with the
+                                          candidate model for model validation.
+                                          (pyfunc model instance is not allowed)
+        :return: A tuple of :py:class:`mlflow.models.EvaluationResult` instance containing
+                 evaluation results for candidate model and baseline model (if provided else None).
```
dbczumar, line 526: Can we update the docstring here? We're not returning a tuple anymore, right?

mlflow/models/evaluation/base.py
```diff
@@ -634,28 +678,38 @@ def _evaluate(
...
+            message="The model could not be evaluated by any of the registered evaluators, please "
+            "verify that the model type and other configs are set correctly.",
+            erorr_code=INVALID_PARAMETER_VALUE,
         )
 
-    merged_eval_result = EvaluationResult(dict(), dict())
+    if baseline_model:
+        merged_eval_result = EvaluationResult(dict(), dict(), dict())
+    else:
+        merged_eval_result = EvaluationResult(dict(), dict())
```
dbczumar, line 704: Would it be acceptable for the constructor of `EvaluationResult()` to represent `baseline_metrics` as an empty dict by default and then populate it? That way, consumers don't have to reason about `None` vs `dict` values for the property and we can simplify this code.

mlflow/models/evaluation/base.py
dbczumar: Instead of making this nullable, can we make it an empty dict by default? See https://github.com/mlflow/mlflow/pull/6310/files#r933521860

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -847,7 +857,7 @@ def _log_custom_metric_artifact(self, artifact_name, raw_artifact, custom_metric
         artifact._load(artifact_file_local_path)
         return artifact
 
-    def _evaluate_custom_metrics_and_log_produced_artifacts(self):
+    def _evaluate_custom_metrics_and_log_produced_artifacts(self, disable_logging=False):
```
dbczumar, line 860: Nit: Can we switch the boolean to `log_to_mlflow_tracking=True` by default? Makes it clearer and removes the double negative.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -971,25 +916,86 @@ def plot_confusion_matrix():
...
-        return self._log_and_return_evaluation_result()
+    def _log_metrics_and_artifacts(self):
+        """
+        Helper method for generating artifacts, logging metrics and artifacts.
+        """
+        if self.model_type == "classifier":
+            if self.is_binomial:
+                self._log_binary_classifier()
+            else:
+                self._log_multiclass_classifier()
```
jerrylian-db, line 990: Should these two function names be updated to be explicitly about logging artifacts? Otherwise, I find the functions a bit confusing.

jerrylian-db (commented):
Looks great! Some minor comments. (Haven't yet got to review the tests. I'll get to them later today)

tests/models/test_default_evaluator.py
```diff
@@ -65,15 +68,91 @@ def assert_dict_equal(d1, d2, rtol):
...
+    targets,
+    model_type: str,
+    dataset_name=None,
+    evaluators=None,
+    evaluator_config=None,
+    eval_baseline_model_only=False,
+):
+
+    """
+    Helper function for testing MLflow.evaluate
```
dbczumar, line 84: Nit: Can we make this docstring a bit more perspective (i.e. what does the helper function do?)

dbczumar (approved):
LGTM with a few small comments. Thanks @zhe-db !

tests/models/test_default_evaluator.py
```diff
@@ -65,15 +68,91 @@ def assert_dict_equal(d1, d2, rtol):
...
+        if not evaluator_config:
+            evaluator_config = {"is_baseline_model": True}
+        elif not evaluators or evaluators == "default":
+            evaluator_config.update({"is_baseline_model": True})
+        else:
+            for config in evaluator_config.values():
+                config.update({"is_baseline_model": True})
+
+    return evaluate(
+        model=baseline_model if eval_baseline_model_only else model,
```
jerrylian-db, line 98: Why would the baseline_model be passed into this parameter? I'm confused.

tests/models/test_default_evaluator.py
```diff
@@ -182,15 +318,74 @@ def test_multi_classifier_evaluation(multiclass_logistic_regressor_model_uri, ir
     }
 
 
-def test_bin_classifier_evaluation(binary_logistic_regressor_model_uri, breast_cancer_dataset):
+def test_multi_classifier_evaluation_disable_logging_metrics_and_artifacts(
```
jerrylian-db, line 321: Correct my understanding if I'm wrong. The default evaluator now evaluates both the candidate and the baseline model at the same time, not logging metrics and artifacts for the baseline model. In that case, doesn't it make sense to get rid of these `disable_logging` tests? Instead, for each model type, pass in a baseline model as well and verify that the baseline model did not have any side-effects. Maybe then, you won't have to have the `evaluate_model_helper` anymore.

jerrylian-db (commented):
More comments around testing

tests/models/test_default_evaluator.py
```diff
@@ -87,6 +166,11 @@ def test_regressor_evaluation(linear_regressor_model_uri, diabetes_dataset):
     expected_metrics["score"] = model._model_impl.score(
         diabetes_dataset.features_data, diabetes_dataset.labels_data
     )
+
+    assert json.loads(tags["mlflow.datasets"]) == [
+        {**diabetes_dataset._metadata, "model": model.metadata.model_uuid}
+    ]
+
```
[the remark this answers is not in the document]
zhe-db: For verifying [datasets tag is logged for candidate model](https://src.dev.databricks.com/mlflow/mlflow@87fcc536f120733f251911fa504c11e92618349b/-/blob/mlflow/models/evaluation/base.py?L623:13)

tests/models/test_default_evaluator.py
zhe-db: For testing if metrics and artifacts are not logged for evaluation of baseline model, it is hard to test that if we pass in both `model` and `baseline_model`, because it is hard to tell whether the `baseline_model` logs some metrics that overwritten the ones `model` generated, so I came up with this to test if we pass in a baseline model and only evaluate it without candidate model to verify that nothing is logged and no artifacts is generated so that we know the boolean flag`is_baseline_model` actually disable logging metrics and artifacts. To do that I have to a way to pass the boolean flag to `DefaultEvaluator` which is through the evaluator config.

tests/models/test_default_evaluator.py
zhe-db: For testing if metrics and artifacts are not logged for evaluation of baseline model, it is hard to test that if we pass in both model and baseline_model, because it is hard to tell whether the baseline_model logs some metrics that overwritten the ones model generated, so I came up with this to test if we pass in a baseline model and only evaluate it without candidate model to verify that nothing is logged and no artifacts is generated so that we know the boolean flagis_baseline_model actually disable logging metrics and artifacts. To do that I have to pass the boolean flag to DefaultEvaluator which is through the evaluator config and evaluate only baseline model without candidate model to see if there is nothing logged.

tests/models/test_default_evaluator.py
zhe-db: It is possible to test it that way; but It is hard to verify if a baseline model has no side effect, especially in the case that a baseline model may have same metrics that have same value as candidate model; hence a direct way is just evaluate a model only with `is_baseline_model` flag set to True and see if there is nothing logged and no artifact generated.

zhe-db: For example, the `example_count` should be the same for all models, so it is not possible to test if baseline model also logs `example_count` that overwrote candidate model's logged metrics, it is not likely that would happen; but to test it strictly I think we should do this way which is more deterministic as this method is agnostic to model.

mlflow/models/evaluation/base.py
```diff
@@ -894,6 +952,14 @@ def scatter_plot(eval_df, builtin_metrics, artifacts_dir):
...
+                                             model validation.
+                                             Metric name must be one of the builtin metric's name or
+                                             name of a custom metric in custom_metrics argument.
+    :param baseline_model: (Optional) A string URI referring to a MLflow model with the pyfunc
+                                      flavor as a baseline model to be compared with the
+                                      candidate model (specified by the `model` param) for model
+                                      validation. (pyfunc model instance is not allowed)
 
     :return: An :py:class:`mlflow.models.EvaluationResult` instance containing
              evaluation results.
```
jerrylian-db, line 965: Can you update this doc string as well?

tests/models/test_default_evaluator.py
```diff
@@ -65,15 +68,93 @@ def assert_dict_equal(d1, d2, rtol):
...
+    )
+
+
+def check_metrics_not_logged_for_baseline_model_evaluation(
+    logged_metrics, result_metrics, expected_metrics
+):
+    """
+    Helper function for checking metrics of evaluation of baseline_model
+     - Metrics should not be logged
+     - Meetrics should be returned as expected
```
jerrylian-db, line 117:
```suggestion
     - Metrics should be returned as expected
```

tests/resources/mlflow-test-plugin/mlflow_test_plugin/dummy_evaluator.py
```diff
@@ -23,7 +23,9 @@ def _load_content_from_file(self, local_artifact_path):
         return pdf.to_numpy()
 
 
+# pylint: disable=attribute-defined-outside-init
 class DummyEvaluator(ModelEvaluator):
+    is_baseline_model = False
```
jerrylian-db, line 28: Is this line needed?

tests/resources/mlflow-test-plugin/mlflow_test_plugin/dummy_evaluator.py
```diff
@@ -42,65 +44,94 @@ def _log_metrics(self, run_id, metrics, dataset_name):
...
+    ):
+        self.model_type = model_type
+        self.client = MlflowClient()
+        self.dataset = dataset
+        self.run_id = run_id
+        self.X = dataset.features_data
+        self.y = dataset.labels_data
+        y_pred = model.predict(self.X)
+        eval_result = self._evaluate(
+            y_pred, is_baseline_model=evaluator_config.get("is_baseline_model", False)
```
jerrylian-db, line 125: I don't see a test where you set `is_baseline_model` for the the `evaluator_config` for the Dummy Evaluator. Do you need this `evaluator_config.get("is_baseline_model", False)`?

tests/models/test_default_evaluator.py
```diff
@@ -104,13 +190,56 @@ def test_regressor_evaluation(linear_regressor_model_uri, diabetes_dataset):
...
+    y_pred = model.predict(diabetes_dataset.features_data)
+
+    expected_metrics = _get_regressor_metrics(y, y_pred)
+    expected_metrics["score"] = model._model_impl.score(
+        diabetes_dataset.features_data, diabetes_dataset.labels_data
+    )
+
+    check_metrics_not_logged_for_baseline_model_evaluation(
+        expected_metrics=expected_metrics,
+        result_metrics=result.metrics,
```
jerrylian-db, line 231: Should this be baseline metrics instead?

tests/models/test_default_evaluator.py
jerrylian-db: This makes a lot more sense to me now. Would it be possible to pass in some dummy model here if `eval_baseline_model_only`? Otherwise, we can add a comment to explain what this line is doing.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -1015,26 +1015,69 @@ def evaluate(
...
+        if inferred_model_type is not None and model_type != inferred_model_type:
+            _logger.warning(
+                f"According to the evaluation dataset label values, the model type looks like "
+                f"{inferred_model_type}, but you specified model type {model_type}. Please "
+                f"verify that you set the `model_type` and `dataset` arguments correctly."
+            )
+
+        evaluation_result = self._evaluate(
+            model, is_baseline_model=evaluator_config.get("is_baseline_model", False)
+        )
```
jerrylian-db, line 1069: An idea: can we change the evaluate config from `is_baseline_model` to `disable_candidate_model_eval`? Then, if the conf is on, we can short-circuit the `evaluation_result` to equal an empty dictionary if the conf is on. I think this would be a clearer way to implement the test path.

tests/models/test_default_evaluator.py
zhe-db: Cause different dataset need different type of model, I will add a comment to explain this line.

tests/resources/mlflow-test-plugin/mlflow_test_plugin/dummy_evaluator.py
zhe-db: Yes, there are tests in evaluation that tests passing in baseline model and candidate model and test if evaluation behavior is preserved.

tests/models/test_default_evaluator.py
zhe-db: no, we technically treated baseline model as "candidate_model`

jerrylian-db: In that case, we aren't testing the candidate model code path, which defeats the purpose of the test.

tests/models/test_default_evaluator.py
```diff
@@ -239,16 +447,74 @@ def test_bin_classifier_evaluation(binary_logistic_regressor_model_uri, breast_c
...
+            binary_logistic_regressor_model_uri,
+            breast_cancer_dataset._constructor_args["data"],
+            model_type="classifier",
+            targets=breast_cancer_dataset._constructor_args["targets"],
+            dataset_name=breast_cancer_dataset.name,
+            evaluators="default",
+            eval_baseline_model_only=True,
+        )
+
+    _, metrics, tags, artifacts = get_run_data(run.info.run_id)
```
jerrylian-db, line 466:
```suggestion
    _, logged_metrics, tags, artifacts = get_run_data(run.info.run_id)
```

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -1015,26 +1015,69 @@ def evaluate(
...
+        if inferred_model_type is not None and model_type != inferred_model_type:
+            _logger.warning(
+                f"According to the evaluation dataset label values, the model type looks like "
+                f"{inferred_model_type}, but you specified model type {model_type}. Please "
+                f"verify that you set the `model_type` and `dataset` arguments correctly."
+            )
+
+        evaluation_result = self._evaluate(
+            model, is_baseline_model=evaluator_config.get("disable_candidate_model", False)
+        )
```
jerrylian-db, line 1069:
```suggestion
          if evaluator_config.get("_disable_candidate_model", False):
           evaluation_result = EvaluationResult()
        else:
          evaluation_result = self._evaluate(
              model, is_baseline_model=False
          )
```

tests/resources/mlflow-test-plugin/mlflow_test_plugin/dummy_evaluator.py
jerrylian-db: Set to false

tests/resources/mlflow-test-plugin/mlflow_test_plugin/dummy_evaluator.py
jerrylian-db: Remove

tests/models/test_default_evaluator.py
jerrylian-db: Update to baseline metrics for all disable logging tests

tests/models/test_default_evaluator.py
jerrylian-db: Now that you are actually evaluating the `baseline_model` now as opposed to the `model`, can you update this conditional to pass None when `eval_baseline_model_only` is `True`?

tests/models/test_default_evaluator.py
```diff
@@ -65,15 +68,97 @@ def assert_dict_equal(d1, d2, rtol):
...
+def evaluate_model_helper(
+    model,
+    baseline_model,
+    data,
+    targets,
+    model_type: str,
+    dataset_name=None,
+    evaluators=None,
+    evaluator_config=None,
+    eval_baseline_model_only=False,
```
jerrylian-db, line 80: In fact, after you implement the disable candidate model eval function, maybe it is possible to pass `model=None` into the `evaluate()` call when the flag ON. In that case you can simplify this function to only take in one model and update the `eval_baseline_model_only` to `eval_as_baseline_model`. Does that make sense?

zhe-db: pass model=None will cause the MLflow.evaluate to throw an exception which is one level above the DefaultEvaluator.evaluate

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -1015,26 +1015,72 @@ def evaluate(
...
+                f"{inferred_model_type}, but you specified model type {model_type}. Please "
+                f"verify that you set the `model_type` and `dataset` arguments correctly."
+            )
+
+        if evaluator_config.get("_disable_candidate_model", False):
+            evaluation_result = EvaluationResult(metrics=dict(), artifacts=dict())
+        else:
+            evaluation_result = self._evaluate(
+                model, is_baseline_model=evaluator_config.get("_disable_candidate_model", False)
+            )
```
jerrylian-db, line 1072: Nit: I think the config get is no longer necessary here?
```suggestion
            evaluation_result = self._evaluate(model, is_baseline_model=False)
```

jerrylian-db (approved):
LGTM, with a nit. Amazing work on this!

Issue #6180, opened 2022-07-01
[FR] Rollback to Previous Model Version in Stage


MinuraPunchihewa:
### Willingness to contribute

Yes. I would be willing to contribute this feature with guidance from the MLflow community.

### Proposal Summary

It would be useful to be able to rollback to a previous model version in given stage, especially Production.

For example, say that I am working on a particular a model and after several iterations, I decide to move version 3 into Production. Now, after a few more rounds development and tuning, I have what I believe to be a better model in version 7 and I decide that I can move it into Production.

After a while, when scoring on the new model version, my analysts inform me that there is some kind of bug with the new version. In such a situation, I would like to be able to quickly move back to the previous version, which I know was working fine, so that no further predictions will be affected.

### Motivation

> #### What is the use case for this feature?
The use case is to be able to quickly rollback to a previous model version in a stage, particularly in Production, that was working fine, without having to remember or search through model versions.

> #### Why is this use case valuable to support for MLflow users in general?
The need to move back to a previous model version is a scenario that I believe anyone can come across.

> #### Why is this use case valuable to support for your project(s) or organization?
I have come across the need to rollback, as a result of bugs that exist in the newer model versions that are transitioned.

> #### Why is it currently difficult to achieve this use case?
At the moment, I am not sure if it is possible to achieve this, outside of searching through all of the models and deciding which one to bring back.

### What component(s) does this bug affect?

- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [X] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/pipelines`: Pipelines, Pipeline APIs, Pipeline configs, Pipeline Templates
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [X] `area/tracking`: Tracking Service, tracking client APIs, autologging

### What interface(s) does this bug affect?

- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
[...]

PR #6320, opened 2022-07-25
Tag based registered model search and model version search

WeichenXu123:
## Related Issues/PRs

https://github.com/mlflow/mlflow/issues/6095

Tag based registered model search and model version search.

Testing code:

```python
import mlflow
mlflow.tracking.set_tracking_uri("...")

result1 = client.search_registered_models(
    filter_string="name='sk-learn-random-forest-reg-model' and tag.axaxb iLike '123XXy'"
)

result2 = client.search_model_versions(filter_string="name = 'sk-learn-random-forest-reg-model' and tags.kk1 ILIKE '%abc%'")
```

## How is this patch tested?

Unit tests

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -364,7 +364,7 @@ class SqlAlchemyStore(AbstractStore):
         for f in parsed_filters:
             type_ = f["type"]
             key = f["key"]
-            comparator = f["comparator"]
+            comparator = f["comparator"].upper()
             value = f["value"]
             if type_ == "attribute":
                 if key not in ("name",):
@@ -383,8 +383,7 @@ class SqlAlchemyStore(AbstractStore):
             elif type_ == "tag":
                 if comparator not in ("=", "!=", "LIKE", "ILIKE"):
                     raise MlflowException.invalid_parameter_value(
-                        f"Invalid comparator for tag: {comparator}",
-                        error_code=INVALID_PARAMETER_VALUE,
+                        f"Invalid comparator for tag: {comparator}"
                     )
                 val_filter = SearchUtils.get_sql_filter_ops(
                     SqlRegisteredModelTag.value, comparator, dialect
@@ -410,7 +409,7 @@ class SqlAlchemyStore(AbstractStore):
         for f in parsed_filters:
             type_ = f["type"]
             key = f["key"]
-            comparator = f["comparator"]
+            comparator = f["comparator"].upper()
             value = f["value"]
             if type_ == "attribute":
                 if key not in ("name", "source_path", "run_id"):
@@ -435,7 +434,6 @@ class SqlAlchemyStore(AbstractStore):
                 if comparator not in ("=", "!=", "LIKE", "ILIKE"):
                     raise MlflowException.invalid_parameter_value(
                         f"Invalid comparator for tag: {comparator}",
-                        error_code=INVALID_PARAMETER_VALUE,
                     )
                 val_filter = SearchUtils.get_sql_filter_ops(
                     SqlModelVersionTag.value, comparator, dialect
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -427,7 +427,7 @@ class SqlAlchemyStore(AbstractStore):
                 if comparator == "IN":
                     # Note: Here the run_id values in databases contains only lower case letters,
                     # so we already filter out comparison values containing upper case letters
-                    # in `SearchModelUtils._get_values`. This address MYSQL IN clause case
+                    # in `SearchModelUtils._get_values`. This addresses MySQL IN clause case
                     # in-sensitive issue.
                     f = attr.in_(value)
                 else:
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -425,7 +425,10 @@ class SqlAlchemyStore(AbstractStore):
                     )
                 attr = getattr(SqlModelVersion, "source" if key == "source_path" else key)
                 if comparator == "IN":
-                    # TODO: Make it case sensitive in MYSQL
+                    # Note: Here the run_id values in databases contains only lower case letters,
+                    # so we already filter out comparison values containing upper case letters
+                    # in `SearchModelUtils._get_values`. This address MYSQL IN clause case
+                    # in-sensitive issue.
                     f = attr.in_(value)
                 else:
                     f = SearchUtils.get_sql_filter_ops(attr, comparator, dialect)(value)
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -2,6 +2,9 @@ import time
 
 import logging
 import sqlalchemy
+from functools import reduce
+
+from sqlalchemy.future import select
 
 from mlflow.entities.model_registry.model_version_stages import (
     get_canonical_stage,
@@ -62,8 +65,7 @@ def _get_search_model_versions_filter_clauses(parsed_filters, dialect):
         "\"source_path = '<source_path>'\" " \
         'or "run_id = \'<run_id>\' or "run_id IN (<run_ids>)" ' \
         'or \"tags.<tag key>\" comparison filter ' \
-        'or "AND" expression composed of the above filters.' \
-        f"Invalid filter string: {filter_string}"
+        'or "AND" expression composed of the above filters.'
     attribute_filters = []
     non_attribute_filters = []
     for f in parsed_filters:
@@ -78,7 +80,7 @@ def _get_search_model_versions_filter_clauses(parsed_filters, dialect):
                 raise MlflowException(
                     "Model Registry search filter only supports the equality(=) "
                     "comparator and the IN operator "
-                    "for the run_id parameter. Input filter string: %s" % filter_string,
+                    "for the run_id parameter.",
                     error_code=INVALID_PARAMETER_VALUE,
                 )
             attr = getattr(SqlModelVersion, key)
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -31,7 +31,7 @@ from mlflow.store.model_registry.dbmodels.models import (
     SqlRegisteredModelTag,
     SqlModelVersionTag,
 )
-from mlflow.utils.search_utils import SearchUtils
+from mlflow.utils.search_utils import SearchUtils, SearchModelVersionsUtils
 from mlflow.utils.uri import extract_db_type_from_uri
 from mlflow.utils.validation import (
     _validate_registered_model_tag,
@@ -54,6 +54,57 @@ sqlalchemy.orm.configure_mappers()
 
 def now():
     return int(time.time() * 1000)
+
+
+def _get_search_model_versions_filter_clauses(parsed_filters, dialect):
+    err_msg = \
+        "Model Registry expects filter to be one of \"name = '<model_name>'\" or " \
+        "\"source_path = '<source_path>'\" " \
+        'or "run_id = \'<run_id>\' or "run_id IN (<run_ids>)" ' \
+        'or \"tags.<tag key>\" comparison filter ' \
+        'or "AND" expression composed of the above filters.' \
+        f"Invalid filter string: {filter_string}"
+    attribute_filters = []
+    non_attribute_filters = []
+    for f in parsed_filters:
+        type_ = f["type"]
+        key = f["key"]
+        comparator = f["comparator"]
+        value = f["value"]
+        if type_ == "attribute":
+            if key not in ("name", "source_path", "run_id"):
+                raise MlflowException(err_msg, error_code=INVALID_PARAMETER_VALUE)
+            if comparator not in ("=", "IN") or (comparator == "IN" and key != "run_id"):
+                raise MlflowException(
+                    "Model Registry search filter only supports the equality(=) "
+                    "comparator and the IN operator "
+                    "for the run_id parameter. Input filter string: %s" % filter_string,
+                    error_code=INVALID_PARAMETER_VALUE,
+                )
+            attr = getattr(SqlModelVersion, key)
+            if comparator == "IN":
+                # TODO: Make it case sensitive in MYSQL
+                f = attr.in_(value)
+            else:
+                f = SearchUtils.get_sql_filter_ops(attr, comparator, dialect)(value)
+            attribute_filters.append(f)
+        elif type_ == "tag":
+            if comparator not in ("=", "!=", "LIKE", "ILIKE"):
+                raise MlflowException.invalid_parameter_value(
+                    f"Invalid comparator for tag: {comparator}, "
+                    "only '=', '!=', 'LIKE', 'ILIKE' comparators are supported."
+                )
+            val_filter = SearchUtils.get_sql_filter_ops(
+                SqlModelVersionTag.value, comparator, dialect
+            )(value)
+            key_filter = SqlModelVersionTag.key == key
+            non_attribute_filters.append(
+                select(SqlModelVersionTag).filter(key_filter, val_filter).subquery()
+            )
+        else:
+            raise MlflowException(err_msg, error_code=INVALID_PARAMETER_VALUE)
+
+    return attribute_filters, non_attribute_filters
 
 
 class SqlAlchemyStore(AbstractStore):
@@ -738,46 +789,24 @@ class SqlAlchemyStore(AbstractStore):
         :return: PagedList of :py:class:`mlflow.entities.model_registry.ModelVersion`
                  objects.
         """
-        parsed_filter = SearchUtils.parse_filter_for_model_versions(filter_string)
-        if len(parsed_filter) == 0:
-            conditions = []
-        elif len(parsed_filter) == 1:
-            filter_dict = parsed_filter[0]
-            if filter_dict["comparator"] not in SearchUtils.VALID_MODEL_VERSIONS_SEARCH_COMPARATORS:
-                raise MlflowException(
-                    "Model Registry search filter only supports the equality(=) "
-                    "comparator and the IN operator "
-                    "for the run_id parameter. Input filter string: %s" % filter_string,
-                    error_code=INVALID_PARAMETER_VALUE,
-                )
-            if filter_dict["key"] == "name":
-                conditions = [SqlModelVersion.name == filter_dict["value"]]
-            elif filter_dict["key"] == "source_path":
-                conditions = [SqlModelVersion.source == filter_dict["value"]]
-            elif filter_dict["key"] == "run_id":
-                if filter_dict["comparator"] == "IN":
-                    conditions = [SqlModelVersion.run_id.in_(filter_dict["value"])]
-                else:
-                    conditions = [SqlModelVersion.run_id == filter_dict["value"]]
-            else:
-                raise MlflowException(
-                    "Invalid filter string: %s" % filter_string, error_code=INVALID_PARAMETER_VALUE
-                )
-        else:
-            raise MlflowException(
-                "Model Registry expects filter to be one of "
-                "\"name = '<model_name>'\" or "
-                "\"source_path = '<source_path>'\" "
-                'or "run_id = \'<run_id>\' or "run_id IN (<run_ids>)".'
-                "Input filter string: %s. " % filter_string,
-                error_code=INVALID_PARAMETER_VALUE,
-            )
+        parsed_filters = SearchModelVersionsUtils.parse_search_filter(filter_string)
+
+        attribute_filters, non_attribute_filters = _get_search_model_versions_filter_clauses(
+            parsed_filters, self.engine.dialect.name
+        )
 
         with self.ManagedSessionMaker() as session:
-            conditions.append(SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL)
-            sql_model_version = session.query(SqlModelVersion).filter(*conditions).all()
-            model_versions = [mv.to_mlflow_entity() for mv in sql_model_version]
+            attribute_filters.append(SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL)
+
+            stmt = (
+                reduce(lambda s, f: s.join(f), non_attribute_filters, select(SqlModelVersion))
+                .options(*self._get_eager_model_version_query_options())
+                .filter(*attribute_filters)
+            )
+            sql_model_versions = session.execute(stmt).scalars(SqlModelVersion).all()
+            model_versions = [mv.to_mlflow_entity() for mv in sql_model_versions]
             return PagedList(model_versions, None)
+
 
     @classmethod
     def _get_model_version_tag(cls, session, name, version, key):
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -462,7 +462,7 @@ class SqlAlchemyStore(AbstractStore):
                 )
 
         mv_query = select(SqlModelVersion).filter(*attribute_filters)
-        if attribute_filters:
+        if tag_filters:
             tag_filter_query = (
                 select(SqlModelVersionTag.name, SqlModelVersionTag.version)
                 .filter(sqlalchemy.or_(*tag_filters))
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -2,8 +2,6 @@ import time
 
 import logging
 import sqlalchemy
-from collections import defaultdict
-
 
 from sqlalchemy.future import select
 
@@ -369,7 +367,7 @@ class SqlAlchemyStore(AbstractStore):
             comparator = f["comparator"]
             value = f["value"]
             if type_ == "attribute":
-                if key not in ("name",):
+                if key != "name":
                     raise MlflowException(
                         f"Invalid attribute name: {key}", error_code=INVALID_PARAMETER_VALUE
                     )
@@ -856,9 +854,11 @@ class SqlAlchemyStore(AbstractStore):
         )
 
         with self.ManagedSessionMaker() as session:
-            stmt = filter_query.options(*self._get_eager_model_version_query_options()).filter(
-                SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL
-            ).order_by(SqlModelVersion.last_updated_time.desc())
+            stmt = (
+                filter_query.options(*self._get_eager_model_version_query_options())
+                .filter(SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL)
+                .order_by(SqlModelVersion.last_updated_time.desc())
+            )
             sql_model_versions = session.execute(stmt).scalars(SqlModelVersion).all()
             model_versions = [mv.to_mlflow_entity() for mv in sql_model_versions]
             return PagedList(model_versions, None)
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -78,7 +78,7 @@ def _get_search_model_versions_filter_clauses(parsed_filters, dialect):
                     f"Invalid comparator for attribute: {comparator}",
                     error_code=INVALID_PARAMETER_VALUE,
                 )
-            attr = getattr(SqlModelVersion, key)
+            attr = getattr(SqlModelVersion, "source" if key == "source_path" else key)
             if comparator == "IN":
                 # TODO: Make it case sensitive in MYSQL
                 f = attr.in_(value)
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -328,7 +328,7 @@ class SqlAlchemyStore(AbstractStore):
 
         parsed_filters = SearchModelUtils.parse_search_filter(filter_string)
 
-        attribute_filters, non_attribute_filters = self._get_search_registered_model_filter_clauses(
+        attribute_filters, tag_filter_query = self._get_search_registered_model_filter_clauses(
             parsed_filters, self.engine.dialect.name
         )
 
@@ -347,8 +347,9 @@ class SqlAlchemyStore(AbstractStore):
 
         with self.ManagedSessionMaker() as session:
             query = (
-                reduce(lambda s, f: s.join(f), non_attribute_filters, select(SqlRegisteredModel))
-                .options(*self._get_eager_registered_model_query_options())
+                select(SqlRegisteredModel).join(
+                    tag_filter_query, SqlRegisteredModel.name == tag_filter_query.c.name
+                ).options(*self._get_eager_registered_model_query_options())
                 .filter(*attribute_filters)
                 .order_by(*parsed_orderby)
                 .limit(max_results_for_query)
@@ -363,7 +364,7 @@ class SqlAlchemyStore(AbstractStore):
     @classmethod
     def _get_search_registered_model_filter_clauses(cls, parsed_filters, dialect):
         attribute_filters = []
-        non_attribute_filters = []
+        tag_filters = []
         for f in parsed_filters:
             type_ = f["type"]
             key = f["key"]
@@ -394,21 +395,24 @@ class SqlAlchemyStore(AbstractStore):
                 key_filter = SearchUtils.get_sql_filter_ops(
                     SqlRegisteredModelTag.key, '=', dialect
                 )(key)
-                non_attribute_filters.append(
-                    select(SqlRegisteredModelTag).filter(key_filter, val_filter).subquery()
-                )
+                tag_filters.append(sqlalchemy.and_(key_filter, val_filter))
             else:
                 raise MlflowException(
                     f"Invalid token type: {type_}",
                     error_code=INVALID_PARAMETER_VALUE
                 )
 
-        return attribute_filters, non_attribute_filters
+        tag_filter_query = select(SqlRegisteredModelTag.name) \
+            .filter(sqlalchemy.or_(*tag_filters)) \
+            .group_by(SqlRegisteredModelTag.name) \
+            .having(sqlalchemy.func.count(sqlalchemy.literal(1)) == len(tag_filters)).subquery()
+
+        return attribute_filters, tag_filter_query
 
     @classmethod
     def _get_search_model_versions_filter_clauses(cls, parsed_filters, dialect):
         attribute_filters = []
-        non_attribute_filters = []
+        tag_filters = []
         for f in parsed_filters:
             type_ = f["type"]
             key = f["key"]
@@ -447,16 +451,19 @@ class SqlAlchemyStore(AbstractStore):
                 key_filter = SearchUtils.get_sql_filter_ops(
                     SqlModelVersionTag.key, '=', dialect
                 )(key)
-                non_attribute_filters.append(
-                    select(SqlModelVersionTag).filter(key_filter, val_filter).subquery()
-                )
+                tag_filters.append(sqlalchemy.and_(key_filter, val_filter))
             else:
                 raise MlflowException(
                     f"Invalid token type: {type_}",
                     error_code=INVALID_PARAMETER_VALUE
                 )
 
-        return attribute_filters, non_attribute_filters
+        tag_filter_query = select(SqlModelVersionTag.name, SqlModelVersionTag.version) \
+            .filter(sqlalchemy.or_(*tag_filters)) \
+            .group_by(SqlModelVersionTag.name, SqlModelVersionTag.version) \
+            .having(sqlalchemy.func.count(sqlalchemy.literal(1)) == len(tag_filters)).subquery()
+
+        return attribute_filters, tag_filter_query
 
     @classmethod
     def _parse_search_registered_models_order_by(cls, order_by_list):
@@ -820,14 +827,18 @@ class SqlAlchemyStore(AbstractStore):
         """
         parsed_filters = SearchModelUtils.parse_search_filter(filter_string)
 
-        attribute_filters, non_attribute_filters = self._get_search_model_versions_filter_clauses(
+        attribute_filters, tag_filter_query = self._get_search_model_versions_filter_clauses(
             parsed_filters, self.engine.dialect.name
         )
 
         with self.ManagedSessionMaker() as session:
             stmt = (
-                reduce(lambda s, f: s.join(f), non_attribute_filters, select(SqlModelVersion))
-                .options(*self._get_eager_model_version_query_options())
+                select(SqlModelVersion).join(
+                    tag_filter_query, sqlalchemy.and_(
+                        SqlModelVersion.name == tag_filter_query.c.name,
+                        SqlModelVersion.version == tag_filter_query.c.version,
+                    )
+                ).options(*self._get_eager_model_version_query_options())
                 .filter(*attribute_filters, SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL)
             )
             sql_model_versions = session.execute(stmt).scalars(SqlModelVersion).all()
--- a/mlflow/store/model_registry/sqlalchemy_store.py
+++ b/mlflow/store/model_registry/sqlalchemy_store.py
@@ -34,7 +34,7 @@ from mlflow.store.model_registry.dbmodels.models import (
     SqlRegisteredModelTag,
     SqlModelVersionTag,
 )
-from mlflow.utils.search_utils import SearchUtils, SearchModelVersionsUtils
+from mlflow.utils.search_utils import SearchUtils, SearchModelUtils
 from mlflow.utils.uri import extract_db_type_from_uri
 from mlflow.utils.validation import (
     _validate_registered_model_tag,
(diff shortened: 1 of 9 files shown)


harupy:
What we need to support is **_tag-based registered model search_**, right? The title and proposal of https://github.com/mlflow/mlflow/issues/6095 seem incorrect.

WeichenXu123:
@harupy 
uh ? I think it's for addressing this: https://github.com/mlflow/mlflow/issues/6180#issuecomment-1176552233
so it's model version searching.

WeichenXu123:
I can also support tag-based registered model search as well.

WeichenXu123:
Yea the proposal in https://github.com/mlflow/mlflow/issues/6095 is wrong..

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -1074,6 +1108,42 @@ def test_search_registered_models(self):
...
+        self.assertEqual(set(rms), {name2})
+
+        rms, _ = self._search_registered_models(f"name = '{name1}' and tag.t1='abc'")
+        self.assertEqual(set(rms), {name1})
+
+        rms, _ = self._search_registered_models(f"tag.t1 LIKE 'ab%'")
+        self.assertEqual(set(rms), {name1, name2})
+
+        rms, _ = self._search_registered_models(f"tag.t1 LIKE 'ab%' AND tag.t2 LIKE 'xy%'")
+        self.assertEqual(set(rms), {name1, name2})
```
harupy, line 1136: Do we use `set` in other tests?

harupy: Looks like we do.

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -334,50 +342,119 @@ def compute_next_token(current_size):
...
+                        or (comparator == "IN" and key != "run_id"):
+                    raise MlflowException(
+                        f"Invalid comparator for attribute: {comparator}",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+                attr = getattr(SqlModelVersion, "source" if key == "source_path" else key)
+                if comparator == "IN":
+                    # Note: Here the run_id values in databases contains only lower case letters,
+                    # so we already filter out comparison values containing upper case letters
+                    # in `SearchModelUtils._get_values`. This address MYSQL IN clause case
```
harupy, line 430:
```suggestion
                    # in `SearchModelUtils._get_values`. This address MySQL IN clause case
```

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -947,6 +945,41 @@ def search_versions(filter_string):
...
+        mv2 = self._mv_maker(
+            name=name, source="A/C", run_id=run_id_2,
+            tags=[ModelVersionTag("t1", "abc"), ModelVersionTag("t2", "x123")]
+        )
+        self.assertEqual(mv2.version, 2)
+
+        def search_versions(filter_string):
+            return [mvd.version for mvd in self.store.search_model_versions(filter_string)]
+
+        self.assertEqual(set(search_versions(f"name='{name}' and tag.t2='xyz'")), {1})
```
harupy, line 969:
```suggestion
        self.assertEqual(set(search_versions(f"name='{name}' and tag.t2 = 'xyz'")), {1})
```

Can we add spaces? `tag.t2='xyz'` is a bit difficult to read for me.

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -738,47 +815,25 @@ def search_model_versions(self, filter_string):
...
-                "\"source_path = '<source_path>'\" "
-                'or "run_id = \'<run_id>\' or "run_id IN (<run_ids>)".'
-                "Input filter string: %s. " % filter_string,
-                error_code=INVALID_PARAMETER_VALUE,
-            )
+        parsed_filters = SearchModelUtils.parse_search_filter(filter_string)
+
+        attribute_filters, non_attribute_filters = self._get_search_model_versions_filter_clauses(
+            parsed_filters, self.engine.dialect.name
+        )
```
harupy, line 822: We're going to revert changes for `search_model_versions` in this PR, right?

WeichenXu123: We need to support tag searching for both search_registered_model and search_model_versions as @dbczumar said.

harupy: So we'll update the Databricks backend later?

mlflow/utils/search_utils.py
```diff
@@ -102,7 +102,7 @@ def get_sql_filter_ops(cls, column, operator, dialect):
         import sqlalchemy as sa
 
         # Use case-sensitive collation for MSSQL
-        if dialect == MSSQL:
+        if dialect == MYSQL:
```
harupy, line 105: Why do we need this change?

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -1074,6 +1112,42 @@ def test_search_registered_models(self):
...
+        ]
+        tags2 = [
+            RegisteredModelTag("t1", "abcd"),
+            RegisteredModelTag("t2", "xyz123"),
+            RegisteredModelTag("t3", "XYZ"),
+        ]
+        self._rm_maker(name1, tags1)
+        self._rm_maker(name2, tags2)
+
+        rms, _ = self._search_registered_models(f"tag.t3='XYZ'")
```
harupy, line 1130:
```suggestion
        rms, _ = self._search_registered_models(f"tag.t3 = 'XYZ'")
```

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -947,6 +936,54 @@ def search_versions(filter_string):
...
+        mv2 = self._mv_maker(
+            name=name, source="A/C", run_id=run_id_2,
+            tags=[ModelVersionTag("t1", "abc"), ModelVersionTag("t2", "x123")]
+        )
+        self.assertEqual(mv2.version, 2)
+
+        def search_versions(filter_string):
+            return [mvd.version for mvd in self.store.search_model_versions(filter_string)]
+
+        self.assertEqual(set(search_versions(f"name='{name}' and tag.t2 = 'xyz'")), {1})
```
harupy, line 973:
```suggestion
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 = 'xyz'")), {1})
```

for consistency

mlflow/store/model_registry/sqlalchemy_store.py
WeichenXu123: Databricks backend already supports tab-based model version search,
but Databricks backend haven't support  tab-based search_registered_model which we need update later.

mlflow/utils/search_utils.py
WeichenXu123: oh, my fault. don't need.

mlflow/store/model_registry/sqlalchemy_store.py
harupy: Ok, I thought we were going to update `search_registered_models` and `search_model_versions` in the following order:

1. First, update `search_registered_models` (because Databricks backend support tag-based registered model search)
2. Then, update Databricks backend to support tab-based model version search.
3. Then, `search_model_versions` to support tags.

WeichenXu123: ok.
Or I can file databricks side PR and merge it first, then merge this OSS PR, this avoid to split the PR into 2 parts.

harupy:
I think we need to merge https://github.com/mlflow/mlflow/pull/6353 first to test case sensitivity properly.

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -876,9 +875,10 @@ def search_versions(filter_string):
             search_versions("run_id IN (")
         assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
 
-        with self.assertRaisesRegex(MlflowException, r"Invalid filter '.+'") as exception_context:
-            search_versions("run_id IN")
-        assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
+        # TODO: fix
+        # with self.assertRaisesRegex(MlflowException, r"Invalid filter '.+'") as exception_context:
+        #     search_versions("run_id IN")
+        # assert exception_context.exception.error_code == ErrorCode.Name(INVALID_PARAMETER_VALUE)
```
harupy, line 881: Reminder to fix this.

mlflow/store/model_registry/sqlalchemy_store.py
harupy: Splitting a PR sounds ok to me.

mlflow/utils/search_utils.py
```diff
@@ -987,3 +916,105 @@ def _apply_reversor(experiment, key, ascending):
...
+                    raise MlflowException(
+                        "Only run_id attribute support compare with a list of quoted string values.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+                run_id_list = cls._parse_list_from_sql_token(token)
+                # Because MYSQL IN clause is case in-sensitive, but all run_ids only contains lower case
+                # letters, so that we filter out run_ids containing upper case letters here.
+                run_id_list = [
+                    run_id for run_id in run_id_list if run_id.lower() == run_id
+                ]
```
harupy, line 988: Do we really need this? Do users really use a query like `run_id IN ('AbC', 'EfG')?`

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -738,47 +815,25 @@ def search_model_versions(self, filter_string):
...
-            conditions.append(SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL)
-            sql_model_version = session.query(SqlModelVersion).filter(*conditions).all()
-            model_versions = [mv.to_mlflow_entity() for mv in sql_model_version]
+            attribute_filters.append(SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL)
+
+            stmt = (
+                reduce(lambda s, f: s.join(f), non_attribute_filters, select(SqlModelVersion))
+                .options(*self._get_eager_model_version_query_options())
+                .filter(*attribute_filters)
+            )
```
harupy, line 831:
```suggestion
            stmt = (
                reduce(lambda s, f: s.join(f), non_attribute_filters, select(SqlModelVersion))
                .options(*self._get_eager_model_version_query_options())
                .filter(*attribute_filters, SqlModelVersion.current_stage != STAGE_DELETED_INTERNAL)
            )
```

mlflow/utils/search_utils.py
```diff
@@ -987,3 +916,105 @@ def _apply_reversor(experiment, key, ascending):
...
+            )
+
+    @classmethod
+    def _invalid_statement_token_search_model_registry(cls, token):
+        if isinstance(token, Comparison):
+            return False
+        elif isinstance(token, Identifier):
+            return False
+        elif isinstance(token, Parenthesis):
+            return False
```
harupy, line 1010:
```suggestion
        if isinstance(token, (Comparison, Identifier, Parenthesis)):
            return False
```

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -1074,6 +1112,42 @@ def test_search_registered_models(self):
...
+        tags2 = [
+            RegisteredModelTag("t1", "abcd"),
+            RegisteredModelTag("t2", "xyz123"),
+            RegisteredModelTag("t3", "XYZ"),
+        ]
+        self._rm_maker(name1, tags1)
+        self._rm_maker(name2, tags2)
+
+        rms, _ = self._search_registered_models(f"tag.t3 = 'XYZ'")
+        self.assertEqual(set(rms), {name2})
```
harupy, line 1131: I don't think we should use `set` because the order matters `search_registered_models`.

harupy: `test_search_registered_models` doesn't use `set`.

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -947,6 +936,54 @@ def search_versions(filter_string):
         assert mvds[0].source == "A/B"
         assert mvds[0].description == "Online prediction model!"
 
+    def test_search_model_versions_in_clause_mysql_case_sensitive(self):
```
harupy, line 939: I think we can just do:

```python
self.assertEqual(set(search_versions(f"run_id IN ('{run_id1.upper()}','{run_id2}')")), {...})
```

mlflow/utils/search_utils.py
WeichenXu123: I think this is more of a correctness issue.
e.g.
suppose run_id "135eeedb755f44e0a2ddde9ea999f573"
user type it as  `run_id IN ("135EEEdb755f44e0a2ddde9ea999f573", ...)` by accident, 
then the filter should be regarded as false.

@dbczumar What do you think ? If no need to handle this we can remove it.

mlflow/store/model_registry/sqlalchemy_store.py
dbczumar: Let's continue with reviewing / merging this PR now that we have a PR on the Databricks side :). Keeping the PRs together is fine.

mlflow/utils/search_utils.py
dbczumar: I think we should keep this here.

tests/store/model_registry/test_sqlalchemy_store.py
dbczumar: +1

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -947,6 +936,54 @@ def search_versions(filter_string):
...
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 = 'xyz'")), {1})
+        self.assertEqual(set(search_versions(f"name = 'wrong_name' and tag.t2 = 'xyz'")), set())
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.`t2` = 'xyz'")), {1})
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t3 = 'xyz'")), set())
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 != 'xy'")), {1, 2})
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 LIKE 'xy%'")), {1})
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 LIKE 'xY%'")), set())
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 ILIKE 'xY%'")), {1})
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 LIKE 'x%'")), {1, 2})
+        self.assertEqual(set(search_versions(f"name = '{name}' and tag.T2 = 'xyz'")), set())
```
WeichenXu123, line 982: Why this test passed ?

harupy: See https://github.com/mlflow/mlflow/runs/7535341371?check_suite_focus=true. `tests/store/model_registry/test_sqlalchemy_store.py` is not tested on MySQL. You need to merge master.

WeichenXu123: The similar test in databricks backend is also passed:
https://github.com/databricks/universe/pull/193846/files#r936134623

dbczumar: Sounds like this isn't an issue

harupy: It turns out this `tests/store/model_registry/test_sqlalchemy_store.py` is not executed on MySQL. I manually modified the code, ran this line on MySQL and got this error:

```
______________________________________________________________________________________________________________________________________ TestSqlAlchemyStoreSqlite.test_search_model_versions_by_tag _______________________________________________________________________________________________________________________________________

self = <test_sqlalchemy_store.TestSqlAlchemyStoreSqlite testMethod=test_search_model_versions_by_tag>

    def test_search_model_versions_by_tag(self):
        # create some model versions
        name = "test_for_search_MV_by_tag"
        self._rm_maker(name)
        run_id_1 = uuid.uuid4().hex
        run_id_2 = uuid.uuid4().hex
    
        mv1 = self._mv_maker(
            name=name,
            source="A/B",
            run_id=run_id_1,
            tags=[ModelVersionTag("t1", "abc"), ModelVersionTag("t2", "xyz")],
        )
        self.assertEqual(mv1.version, 1)
        mv2 = self._mv_maker(
            name=name,
            source="A/C",
            run_id=run_id_2,
            tags=[ModelVersionTag("t1", "abc"), ModelVersionTag("t2", "x123")],
        )
        self.assertEqual(mv2.version, 2)
    
        def search_versions(filter_string):
            return [mvd.version for mvd in self.store.search_model_versions(filter_string)]
    
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 = 'xyz'")), {1})
        self.assertEqual(set(search_versions(f"name = 'wrong_name' and tag.t2 = 'xyz'")), set())
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.`t2` = 'xyz'")), {1})
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t3 = 'xyz'")), set())
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 != 'xy'")), {1, 2})
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 LIKE 'xy%'")), {1})
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 LIKE 'xY%'")), set())
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 ILIKE 'xY%'")), {1})
        self.assertEqual(set(search_versions(f"name = '{name}' and tag.t2 LIKE 'x%'")), {1, 2})
>       self.assertEqual(set(search_versions(f"name = '{name}' and tag.T2 = 'xyz'")), set())
E       AssertionError: Items in the first set but not the second:
E       1
```

harupy: @dbczumar @WeichenXu123 I'm looking at `case_sensitive_mysql_eq`:

https://github.com/mlflow/mlflow/blob/aa9ad78d175839144d40237ef662a9f8a0e9e4e6/mlflow/utils/search_utils.py#L109

```
        def case_sensitive_mysql_eq(value):
            return sa.and_(column.__eq__(value), column.op("= BINARY", is_comparison=True)(value))
```

This is compiled to `<column> = 'foo' AND <column> = BINARY 'foo'`, but what we want is `<column> = 'foo' AND BINARY <column> = 'foo'`. I'll file a PR.

tests/store/model_registry/test_sqlalchemy_store.py
harupy: What if `rms` contains duplicates? The test would pass, but it should not.

mlflow/utils/search_utils.py
```diff
@@ -115,7 +115,7 @@ def case_sensitive_mysql_eq(value):
             )
 
         def case_sensitive_mysql_ne(value):
-            return sa.text(f"{col} != :value OR BINARY {col} != :value").bindparams(
+            return sa.text(f"({col} != :value OR BINARY {col} != :value)").bindparams(
```
harupy, line 118: Nice catch.

WeichenXu123: @harupy See this failure test: (corresponding commit: 1596ab9)
https://github.com/mlflow/mlflow/runs/7647679932?check_suite_focus=true#step:4:606

The searching result contains duplicated items. I am still thinking what causes this.

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -337,62 +345,142 @@ def compute_next_token(current_size):
...
+
+        rm_query = select(SqlRegisteredModel).filter(*attribute_filters)
+        if tag_filters:
+            tag_filter_query = (
+                select(SqlRegisteredModelTag.name)
+                .filter(sqlalchemy.or_(*tag_filters))
+                .group_by(SqlRegisteredModelTag.name)
+                .having(sqlalchemy.func.count(sqlalchemy.literal(1)) == len(tag_filters))
+                .subquery()
+            )
```
WeichenXu123, line 408: Avoid generating multiple table join on tag table when there are multiple tag filters.
This is the similar approach with databricks backend, I think it is a more efficient approach.

WeichenXu123, line 408: Avoid generating multiple table join on tag table when there are multiple tag filters.
This is the similar approach with databricks backend, I think it is a more efficient approach.

mlflow/utils/search_utils.py
WeichenXu123: oh I got it. The wrong `OR` operator causes tag table filter out multiple rows, and it joins with the model table so it generate duplicated results.

tests/store/model_registry/test_sqlalchemy_store.py
WeichenXu123: Fixed!

mlflow/store/model_registry/sqlalchemy_store.py
harupy: Cool! Any benchmark?

harupy: Does this approach really work?

Let's say we only have a registered model tag `{"key": "key", "value": "value"}` and perform a search with `filter_string="tag.key LIKE 'v%' AND tag.key LIKE '%e'"`. `len(tag_filters)` is 2, but `sqlalchemy.func.count(sqlalchemy.literal(1))` can never be larger than 1 since we only have a single registered model tag.

harupy: We might need to replace `len(tag_filters)` with the number of unique tag keys in `len(tag_filters)`.

WeichenXu123: yes. I updated.

tests/store/model_registry/test_sqlalchemy_store.py
WeichenXu123: test updated, now comparing result with list.

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -972,6 +969,55 @@ def search_versions(filter_string):
...
+            source="A/C",
+            run_id=run_id_2,
+            tags=[ModelVersionTag("t1", "abc"), ModelVersionTag("t2", "x123")],
+        )
+        self.assertEqual(mv2.version, 2)
+
+        def search_versions(filter_string):
+            return [mvd.version for mvd in self.store.search_model_versions(filter_string)]
+
+        self.assertEqual(search_versions(f"name = '{name}' and tag.t2 = 'xyz'"), [1])
```
harupy, line 997:
```suggestion
        assert search_versions(f"name = '{name}' and tag.t2 = 'xyz'" == [1]
```

Let's use `assert` to make use of pytest assert rewrite.

https://docs.pytest.org/en/7.1.x/_modules/_pytest/assertion.html

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -2,6 +2,10 @@
 
 import logging
 import sqlalchemy
+from collections import defaultdict
```
harupy, line 5: Looks like `defaultdict` is not used at all.

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -337,62 +346,149 @@ def compute_next_token(current_size):
...
+                raise MlflowException(
+                    f"Invalid token type: {type_}", error_code=INVALID_PARAMETER_VALUE
+                )
+
+        mv_query = select(SqlModelVersion).filter(*attribute_filters)
+        if tag_filters:
+            sql_tag_filters = map(lambda x: sqlalchemy.and_(*x), tag_filters.values())
+            tag_filter_query = (
+                select(SqlModelVersionTag.name, SqlModelVersionTag.version)
+                .filter(sqlalchemy.or_(*sql_tag_filters))
```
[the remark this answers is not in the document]
WeichenXu123: no. We need use `AND` to connect filter within each key group, then use `OR` to connect these groups.

e.g.

`tag.k1 like 'v%' and tag.k1 != 'v1' and tag.k2 like 'w%' and tag.k2 != 'w1'`, we should generate sql filter:

```
(key = 'k1' and value like 'v%' and value != 'v1')
or
(key = 'k2' and value like 'w%' and value != 'w1')
```

each AND group condition corresponding to one row in tag table.

mlflow/utils/search_utils.py
```diff
@@ -994,3 +923,106 @@ def _apply_reversor(experiment, key, ascending):
...
+                return cls._strip_quotes(token.value, expect_quoted_value=True)
+            elif isinstance(token, Parenthesis):
+                cls._check_valid_identifier_list(token)
+                if key != "run_id":
+                    raise MlflowException(
+                        "Only run_id attribute support compare with a list of quoted string values.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+                run_id_list = cls._parse_list_from_sql_token(token)
+                # Because MYSQL IN clause is case in-sensitive, but all run_ids only contains lower case
```
harupy, line 994:
```suggestion
                # Because MySQL IN clause is case insensitive, but all run_ids only contains lower case
```

mlflow/utils/search_utils.py
```diff
@@ -994,3 +923,106 @@ def _apply_reversor(experiment, key, ascending):
...
+                    "Expected a quoted string value or a list of quoted string values for "
+                    "attributes. Got value {value}".format(value=token.value),
+                    error_code=INVALID_PARAMETER_VALUE,
+                )
+        else:
+            # Expected to be either "param" or "metric".
+            raise MlflowException(
+                "Invalid identifier type. Expected one of "
+                "{}.".format([cls._ATTRIBUTE_IDENTIFIER, cls._TAG_IDENTIFIER])
+            )
```
harupy, line 1009:
```suggestion
            raise MlflowException(
                "Invalid identifier type. Expected one of "
                "{}.".format([cls._ATTRIBUTE_IDENTIFIER, cls._TAG_IDENTIFIER]),
                error_code=INVALID_PARAMETER_VALUE,
            )
```

mlflow/utils/search_utils.py
```diff
@@ -994,3 +923,106 @@ def _apply_reversor(experiment, key, ascending):
...
+                cls._check_valid_identifier_list(token)
+                if key != "run_id":
+                    raise MlflowException(
+                        "Only run_id attribute support compare with a list of quoted string values.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+                run_id_list = cls._parse_list_from_sql_token(token)
+                # Because MYSQL IN clause is case in-sensitive, but all run_ids only contains lower case
+                # letters, so that we filter out run_ids containing upper case letters here.
+                run_id_list = [run_id for run_id in run_id_list if run_id.lower() == run_id]
```
harupy, line 996:
```suggestion
                run_id_list = [run_id for run_id in run_id_list if run_id.islower()]
```

mlflow/store/model_registry/sqlalchemy_store.py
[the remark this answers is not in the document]
harupy: Makes sense. I thought `tag_filters` contains a list of key and value filter pairs but it only contains value filters.

mlflow/tracking/client.py
```diff
@@ -2682,9 +2694,27 @@ def search_model_versions(self, filter_string: str) -> PagedList[ModelVersion]:
...
+              - ``source_path``: model version source path.
+              - ``run_id``: The id of the mlflow run that generates the model version.
+              - ``tags.<tag_key>``: model version tag. If ``tag_key`` contains paces, it must be
+                wrapped with backticks (e.g., ``"tags.`extra key`"``).
+
+            Comparators
+              - ``=``: Equal to.
+              - ``!=``: Not equal to.
+              - ``LIKE``: Case-sensitive pattern match.
+              - ``ILIKE``: Case-insensitive sensitive pattern match.
```
harupy, line 2712:
```suggestion
              - ``ILIKE``: Case-insensitive pattern match.
```

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -972,6 +969,47 @@ def search_versions(filter_string):
...
+        assert search_versions(f"name = '{name}' and tag.t2 LIKE 'xy%'") == [1]
+        assert search_versions(f"name = '{name}' and tag.t2 LIKE 'xY%'") == []
+        assert search_versions(f"name = '{name}' and tag.t2 ILIKE 'xY%'") == [1]
+        assert search_versions(f"name = '{name}' and tag.t2 LIKE 'x%'") == [2, 1]
+        assert search_versions(f"name = '{name}' and tag.T2 = 'xyz'") == []
+        assert search_versions(f"name = '{name}' and tag.t1 = 'abc' and tag.t2 = 'xyz'") == [1]
+        assert search_versions(f"name = '{name}' and tag.t1 = 'abc' and tag.t2 LIKE 'x%'") == [2, 1]
+        assert search_versions(f"name = '{name}' and tag.t1 = 'abc' and tag.t2 LIKE 'y%'") == []
+        # test filter with duplicated keys
+        search_versions(f"name = '{name}' and tag.t2 like 'x%' and tag.t2 != 'xyz'") == [2]
```
harupy, line 1011:
```suggestion
        assert search_versions(f"name = '{name}' and tag.t2 like 'x%' and tag.t2 != 'xyz'") == [2]
```

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -870,6 +870,12 @@ def search_versions(filter_string):
             set([1, 2, 3]),
         )
 
+        # search IN operator is case sensitive
+        self.assertEqual(
+            set(search_versions(f"run_id IN ('{run_id_1.upper()}','{run_id_2}')")),
+            set([2, 3]),
+        )
+
```
harupy, line 878: Do we have tests for `run_id LIKE` and `run_id ILIKE`?

mlflow/utils/search_utils.py
```diff
@@ -994,3 +923,108 @@ def _apply_reversor(experiment, key, ascending):
...
+        elif token.is_whitespace:
+            return False
+        elif token.match(ttype=TokenType.Keyword, values=["AND", "IN"]):
+            return False
+        elif token.match(ttype=TokenType.Operator.Comparison, values=["IN"]):
+            # `IN` is a comparison token in sqlparse >= 0.4.0:
+            # https://github.com/andialbrecht/sqlparse/pull/567
+            if token.value == "IN":
+                # This case it represent the IN filter parsed failed.
+                # e.g. "run_id IN"
```
harupy, line 1026: What does this mean?

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -972,6 +969,47 @@ def search_versions(filter_string):
...
+            source="A/C",
+            run_id=run_id_2,
+            tags=[ModelVersionTag("t1", "abc"), ModelVersionTag("t2", "x123")],
+        )
+        assert mv2.version == 2
+
+        def search_versions(filter_string):
+            return [mvd.version for mvd in self.store.search_model_versions(filter_string)]
+
+        assert search_versions(f"name = '{name}' and tag.t2 = 'xyz'") == [1]
```
harupy, line 997:
```suggestion
        assert search_versions(f"tag.t2 = 'xyz'") == [1]
```

Do we need `name = '{name}' and `?

mlflow/utils/search_utils.py
```diff
@@ -994,3 +923,108 @@ def _apply_reversor(experiment, key, ascending):
...
+                if key != "run_id":
+                    raise MlflowException(
+                        "Only run_id attribute support compare with a list of quoted string "
+                        "values.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+                run_id_list = cls._parse_list_from_sql_token(token)
+                # Because MySQL IN clause is case-insensitive, but all run_ids only contains lower
+                # case letters, so that we filter out run_ids containing upper case letters here.
+                run_id_list = [run_id for run_id in run_id_list if run_id.islower()]
```
harupy, line 997:
```suggestion
                if dialect == MYSQL:
                    run_id_list = [run_id for run_id in run_id_list if run_id.islower()]
```

mlflow/utils/search_utils.py
```diff
@@ -994,3 +923,108 @@ def _apply_reversor(experiment, key, ascending):
...
+            return False
+        elif token.match(ttype=TokenType.Keyword, values=["AND", "IN"]):
+            return False
+        elif token.match(ttype=TokenType.Operator.Comparison, values=["IN"]):
+            # `IN` is a comparison token in sqlparse >= 0.4.0:
+            # https://github.com/andialbrecht/sqlparse/pull/567
+            if token.value == "IN":
+                # This case it represent the IN filter parsed failed.
+                # e.g. "run_id IN"
+                return True
```
harupy, line 1027: Does this happen for other operators? What happens if `filter_string` looks like `tag.a LIKE`?

WeichenXu123: Parsing `tag.a LIKE` works fine (the parser will raise correct exception),
because if filter is `tag.a LIKE`, then it will run into L1030
the L1021 branch is for backwards compatibility on lsqlparse<0.4 package, and the "run_id IN" filter accidently run into this branch, so the issue only happen on "IN" operator.

mlflow/utils/search_utils.py
WeichenXu123: Getting `dialect` in this method is a bit difficult, (it is in an overridden function, adding argument is not convenient)
but the filtering line does not hurt perf too much, I think we can keep it.

WeichenXu123: Or what about directly support case sensitive IN clause, so we don't this ugly patch code.

harupy: > Or what about directly support case sensitive IN clause, so we don't this ugly patch code.

+1

dbczumar: I don't think we need to focus too much on `IN`, if the cost is high.

mlflow/tracking/client.py
```diff
@@ -2682,9 +2694,27 @@ def search_model_versions(self, filter_string: str) -> PagedList[ModelVersion]:
...
-                              ``run_id = '...'``.
+        :param filter_string: Filter query string (e.g., ``"name = 'a_model_name'"``),
+            defaults to searching for all model versions. The following identifiers, comparators,
+            and logical operators are supported.
+
+            Identifiers
+              - ``name``: model name.
+              - ``source_path``: model version source path.
+              - ``run_id``: The id of the mlflow run that generates the model version.
+              - ``tags.<tag_key>``: model version tag. If ``tag_key`` contains paces, it must be
```
dbczumar, line 2705:
```suggestion
              - ``tags.<tag_key>``: model version tag. If ``tag_key`` contains spaces, it must be
```

mlflow/tracking/client.py
```diff
@@ -1990,12 +1990,24 @@ def search_registered_models(
...
-                to match a pattern in the registered model name.
-                For example, ``name LIKE 'Boston%'`` (case sensitive) or
-                ``name ILIKE '%boston%'`` (case insensitive).
+        :param filter_string: Filter query string (e.g., ``"name = 'a_model_name'"``),
+            defaults to searching for all registered models. The following identifiers, comparators,
+            and logical operators are supported.
+
+            Identifiers
+              - ``name``: registered model name.
+              - ``tags.<tag_key>``: registered model tag. If ``tag_key`` contains paces, it must be
```
dbczumar, line 1999:
```suggestion
              - ``tags.<tag_key>``: registered model tag. If ``tag_key`` contains spaces, it must be
```

tests/store/model_registry/test_sqlalchemy_store.py
dbczumar: +1

tests/store/model_registry/test_sqlalchemy_store.py
```diff
@@ -1099,6 +1139,46 @@ def test_search_registered_models(self):
             ([names[4]], None),
         )
 
+    def test_search_registered_models_by_tag(self):
```
dbczumar, line 1142: Can we add ILIKE test cases here too?

mlflow/utils/search_utils.py
dbczumar: Can we add a test case for this, if one does not exist?

dbczumar (approved):
LGTM once requested test coverage has been added and tests pass. Thanks @WeichenXu123 !

mlflow/tracking/client.py
```diff
@@ -1990,12 +1990,24 @@ def search_registered_models(
         """
         Search for registered models in backend that satisfy the filter criteria.
 
-        :param filter_string: Filter query string, defaults to searching all registered
-                models. Currently, it supports only a single filter condition as the name
-                of the model, for example, ``name = 'model_name'`` or a search expression
-                to match a pattern in the registered model name.
-                For example, ``name LIKE 'Boston%'`` (case sensitive) or
-                ``name ILIKE '%boston%'`` (case insensitive).
+        :param filter_string: Filter query string (e.g., ``"name = 'a_model_name'"``),
```
dbczumar, line 1993: Can we add an example of a filter_string with `and`?

mlflow/tracking/client.py
```diff
@@ -2682,9 +2694,27 @@ def search_model_versions(self, filter_string: str) -> PagedList[ModelVersion]:
         """
         Search for model versions in backend that satisfy the filter criteria.
 
-        :param filter_string: A filter string expression. Currently, it supports a single filter
-                              condition either a name of model like ``name = 'model_name'`` or
-                              ``run_id = '...'``.
+        :param filter_string: Filter query string (e.g., ``"name = 'a_model_name'"``),
```
dbczumar, line 2697: Can we add an example of a filter_string with `and`?

harupy (approved):
LGTM once the remaining comments are addressed.

mlflow/utils/search_utils.py
WeichenXu123: added.

mlflow/utils/search_utils.py
WeichenXu123: ok let's keep current code

tests/store/model_registry/test_sqlalchemy_store.py
WeichenXu123: removed.
but I keep the first one for testing both attribute filter and tag filter existing case.

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -751,47 +845,22 @@ def search_model_versions(self, filter_string):
...
-        else:
-            raise MlflowException(
-                "Model Registry expects filter to be one of "
-                "\"name = '<model_name>'\" or "
-                "\"source_path = '<source_path>'\" "
-                'or "run_id = \'<run_id>\' or "run_id IN (<run_ids>)".'
-                "Input filter string: %s. " % filter_string,
-                error_code=INVALID_PARAMETER_VALUE,
-            )
+                 objects. The order of returned records is descending order of last updated time.
```
harupy, line 848:
```suggestion
                 objects sorted by last updated time in descending order.
```

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -337,62 +344,149 @@ def compute_next_token(current_size):
...
+                if comparator not in ("=", "!=", "LIKE", "ILIKE", "IN") or (
+                    comparator == "IN" and key != "run_id"
+                ):
+                    raise MlflowException(
+                        f"Invalid comparator for attribute: {comparator}",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+                attr = getattr(SqlModelVersion, "source" if key == "source_path" else key)
+                if comparator == "IN":
+                    # Note: Here the run_id values in databases contains only lower case letters,
```
harupy, line 442:
```suggestion
                    # Note: Here the run_id values in databases contain only lower case letters,
```

mlflow/store/model_registry/sqlalchemy_store.py
```diff
@@ -337,62 +344,149 @@ def compute_next_token(current_size):
...
+                    # so we already filter out comparison values containing upper case letters
+                    # in `SearchModelUtils._get_values`. This addresses MySQL IN clause case
```
harupy, line 444:
```suggestion
                    # so we already filter out comparison values containing upper case letters
                    # in `SearchModelUtils._get_value`. This addresses MySQL IN clause case
```

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -512,6 +512,10 @@ def create_run(self, experiment_id, user_id, start_time, tags):
             experiment = self.get_experiment(experiment_id)
             self._check_experiment_is_active(experiment)
 
+            # Note: we need to ensure the generated "run_id" only contains digits and lower
+            # case letters, because some query filters contains "IN" clause, and in MYSQL the
```
harupy, line 516:
```suggestion
            # case letters, because some query filters contain "IN" clause, and in MYSQL the
```

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -512,6 +512,10 @@ def create_run(self, experiment_id, user_id, start_time, tags):
...
+            # "IN" clause is case in-sensitive, we use a trick that filter out comparison values
```
harupy, line 517:
```suggestion
            # "IN" clause is case in-sensitive, we use a trick that filters out comparison values
```

mlflow/utils/search_utils.py
```diff
@@ -994,3 +923,108 @@ def _apply_reversor(experiment, key, ascending):
...
+            elif isinstance(token, Parenthesis):
+                cls._check_valid_identifier_list(token)
+                if key != "run_id":
+                    raise MlflowException(
+                        "Only run_id attribute support compare with a list of quoted string "
+                        "values.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+                run_id_list = cls._parse_list_from_sql_token(token)
+                # Because MySQL IN clause is case-insensitive, but all run_ids only contains lower
```
harupy, line 995:
```suggestion
                # Because MySQL IN clause is case-insensitive, but all run_ids only contain lower
```

aravinds502:
I was trying with mlflow==2.5.0, but when I try the filter as follows 
`mlflow_client.search_registered_models(filter_string="tag.`{}` LIKE '{}'".format("project_id", project_id))`
or
`mlflow_client.search_registered_models(filter_string="tags.`{}` LIKE '{}'".format("project_id", project_id))`

it gives errors as

`{RestException}INVALID_PARAMETER_VALUE: Invalid attribute key 'tag.`project_id`' specified. Valid keys  are '{'name'}'`

May I know in which version this fix went in.

Thanks

WeichenXu123:
@aravinds502 Could you file a ticket, including the full repro code ? Including the code that logs the model, registers the model, and sets model tags.

aravinds502:
@WeichenXu123 My bad, the MLFlow tracking component I was using was 1.20.2, whereas the application using that tracking server had mlflow ==2.5.0 package. When I upgraded the MLFlow tracking server to 2.5.0, it started working.
Thank you!

PR #6376, opened 2022-08-01
CLI command suggestion after running the pipeline

hubertzub-db:
Signed-off-by: Hubert Zub <hubert.zub@databricks.com>

After running the pipeline with a given `profile` and `step` parameters, those parameters are saved in run tags. On the run details page, this data is being utilized to suggest a CLI command that recreates the same execution.
 
Screen Shot 2022-08-05 at 11 10 44

## How is this patch tested?
- done manual tests on both full runs (without defined `step`) and on runs with a named step (e.g. `-s train`).
- existing unit test suite was extended with tag case

   next step, otherwise fix it.
3. Click the `Artifacts` tab.

Display CLI command in the tracking UI that reproduces given ML Pipelines run.

Components

Interface

Language

Integrations


jinzhang21 (commented):
Can we actually guarantee reproducing the run by showing a command with recorded profile name and step name? What if users changed the content of the profile in between? If we just want to show a tip for running the pipeline, I'd prefer to not recording the step but simply showing how to run the entire pipeline via `mlp run --profile PROFILE`.

hubertzub-db:
> Can we actually guarantee reproducing the run by showing a command with recorded profile name and step name? What if users changed the content of the profile in between? If we just want to show a tip for running the pipeline, I'd prefer to not recording the step but simply showing how to run the entire pipeline via `mlp run --profile PROFILE`.

Makes sense, however I'm not sure if there was no other intention here. @sunishsheth2009 what do you think?

sunishsheth2009:
> Can we actually guarantee reproducing the run by showing a command with recorded profile name and step name? What if users changed the content of the profile in between? If we just want to show a tip for running the pipeline, I'd prefer to not recording the step but simply showing how to run the entire pipeline via `mlp run --profile PROFILE`.

Don't you think @jinzhang21 that just using the profile context would have the same problem of users changing the underlying profile? Or even the profile-name for that matter? Step names for now are static. For future it would be good to show step name as well once users can create their own steps which have a unique step name. What do you think?

dbczumar:
> > Can we actually guarantee reproducing the run by showing a command with recorded profile name and step name? What if users changed the content of the profile in between? If we just want to show a tip for running the pipeline, I'd prefer to not recording the step but simply showing how to run the entire pipeline via `mlp run --profile PROFILE`.
> 
> Don't you think @jinzhang21 that just using the profile context would have the same problem of users changing the underlying profile? Or even the profile-name for that matter? Step names for now are static. For future it would be good to show step name as well once users can create their own steps which have a unique step name. What do you think?

For reproducibility, it would suffice to include `git clone <pipeline_repo>` and `git checkout <commit>` commands before the `mlflow pipelines run` invocation.

jinzhang21:
> > > Can we actually guarantee reproducing the run by showing a command with recorded profile name and step name? What if users changed the content of the profile in between? If we just want to show a tip for running the pipeline, I'd prefer to not recording the step but simply showing how to run the entire pipeline via `mlp run --profile PROFILE`.
> > 
> > 
> > Don't you think @jinzhang21 that just using the profile context would have the same problem of users changing the underlying profile? Or even the profile-name for that matter? Step names for now are static. For future it would be good to show step name as well once users can create their own steps which have a unique step name. What do you think?
> 
> For reproducibility, it would suffice to include `git clone <pipeline_repo>` and `git checkout <commit>` commands before the `mlflow pipelines run` invocation. It would be more convenient syntactically if `mlflow pipelines run` supported passing in a repo URL and commit as commandline flags.

All great recommendations. Makes sense to me. One more thing on the step recorded here:
Instead of recording the step user asked to run, we might want to record the actual last step executed successfully. Imagine a partial execution `p.run("evaluate")` that succeeds in `train` but fails in `evaluate` (for whatever reason), do we want to show a command that guarantees to fail in `evaluate` again?

mlflow/pipelines/pipeline.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, pipeline_root_path: str, profile: str) -> None:
                         pipeline.yaml to generate the configuration to run the pipeline.
         """
         self._pipeline_root_path = pipeline_root_path
+        self._initial_config_step = ""
```
jinzhang21, line 44: `_initial_config_step` -> `_run_args`, and record all args provided to `Pipeline.run()`. This will come in handy in future iterations we plan to add more args to the API.

mlflow/pipelines/pipeline.py
```diff
@@ -166,6 +171,8 @@ def _resolve_pipeline_steps(self) -> List[BaseStep]:
         Constructs and returns all pipeline step objects from the pipeline configuration.
         """
         pipeline_config = get_pipeline_config(self._pipeline_root_path, self._profile)
+        pipeline_config["initial_config_profile"] = self.profile
```
jinzhang21, line 174: `initial_config_profile` -> `profile` to be internally consistent

mlflow/pipelines/steps/train.py
```diff
@@ -114,6 +119,8 @@ def _run(self, output_directory):
         tags = {
             MLFLOW_SOURCE_TYPE: SourceType.to_string(SourceType.PIPELINE),
             MLFLOW_PIPELINE_TEMPLATE_NAME: self.step_config["template_name"],
+            MLFLOW_PIPELINE_PROFILE_NAME: self.step_config["initial_config_profile"],
+            MLFLOW_PIPELINE_STEP_NAME: self.step_config["initial_config_step"]
```
jinzhang21, line 123: Here you should be able to get the step name via `self.step_config["run_args"]["step"]`

mlflow/pipelines/steps/train.py
```diff
@@ -119,8 +119,8 @@ def _run(self, output_directory):
         tags = {
             MLFLOW_SOURCE_TYPE: SourceType.to_string(SourceType.PIPELINE),
             MLFLOW_PIPELINE_TEMPLATE_NAME: self.step_config["template_name"],
-            MLFLOW_PIPELINE_PROFILE_NAME: self.step_config["initial_config_profile"],
-            MLFLOW_PIPELINE_STEP_NAME: self.step_config["initial_config_step"]
+            MLFLOW_PIPELINE_PROFILE_NAME: self.step_config["profile"],
+            MLFLOW_PIPELINE_STEP_NAME: self.step_config["run_args"]["step"]
```
jinzhang21, line 123: `self.step_config["run_args"].get("step") or ""` is probably better way to handle the `step == None` case.

mlflow/pipelines/pipeline.py
```diff
@@ -71,8 +71,10 @@ def run(self, step: str = None) -> None:
         :return: None
         """
 
-        # Save the step requested by the user so it can be saved as a tag
-        self._initial_config_step = step
+        # Save the run parameters for later
+        self._run_args = {
+            "step": step
```
jinzhang21, line 76: Is there a way to automatically capture all run_args from the function? Something from inspect package could help.
I have not investigated carefully, but `inspect.getargvalues(inspect.currentframe())` may work. If it's too complicated to implement, don't bother. The current approach is not great but OK.

jinzhang21 (approved):
Please fix the lint and address my comments. Otherwise, LGTM!

mlflow/pipelines/steps/train.py
```diff
@@ -114,6 +119,8 @@ def _run(self, output_directory):
         tags = {
             MLFLOW_SOURCE_TYPE: SourceType.to_string(SourceType.PIPELINE),
             MLFLOW_PIPELINE_TEMPLATE_NAME: self.step_config["template_name"],
+            MLFLOW_PIPELINE_PROFILE_NAME: self.step_config["profile"],
```
jinzhang21, line 122: Can we also add a unit test to verify that these fields are correctly logged? Somewhere here: https://github.com/mlflow/mlflow/blob/master/tests/pipelines/test_train_step.py#L113

mlflow/pipelines/pipeline.py
hubertzub-db: @jinzhang21 
I was looking for a solution for this before and the only thing I found is https://stackoverflow.com/a/10724602/18956398
It seems to be pretty hacky so I didn't use it - but maybe it's ok? 🤔 I don't feel super confident in python, not sure what's elegant and what's not.

jinzhang21: The solution looks fine (not as hacky to me). But we don't have to do it in this PR. You can submit the code as is. We could discuss what to do when it comes to the point of extending the `run()` API.

mlflow/pipelines/steps/train.py
hubertzub-db: I'll add the unit test in the next PR.

hubertzub-db: I've extended existing unit test.

jinzhang21 (approved):

mlflow/server/js/src/experiment-tracking/components/RunView.js
```diff
@@ -125,6 +125,26 @@ export class RunViewImpl extends Component {
...
     const entryPointName = Utils.getEntryPointName(tags);
     const backend = Utils.getBackend(tags);
+
+    if (Utils.getSourceType(tags) === 'PIPELINE') {
+      const profileName = Utils.getPipelineProfileName(tags);
+      const stepName = Utils.getPipelineStepName(tags);
+      runCommand = '';
+      if (sourceName) {
+        runCommand += `git clone ${sourceName}\n`;
+      }
```
sunishsheth2009, line 135: I am assuming the sourceName and the sourceVersion is automatically tracked while to create a run?

sunishsheth2009 (commented):
LGTM overall.

mlflow/server/js/src/experiment-tracking/components/RunView.js
hubertzub-db: I assume it's always here, at least it something that deducted from the code.

PR #6517, opened 2022-08-19
[ML-22889] Get artifact cli

prithvikannan:
## Related Issues/PRs

#xxx

Added `mlflow pipelines get-artifact` command to cli. This already existed via Python API but not in CLI. The Python API will return actual python objects, but the CLI behavior is to return the file path in the execution directory instead.

## How is this patch tested?

- [] Yes. Make sure the changed pages / sections render correctly by following the steps below.

Added `mlflow pipelines get-artifact` command to CLI. This already existed via Python API but not in CLI. The Python API will return actual python objects, but the CLI behavior is to return the file path in the execution directory instead.

Components

Interface

Language

Integrations


mlflow/pipelines/cli.py
```diff
@@ -93,3 +93,22 @@ def inspect(step, profile):
...
+
+
+@commands.command(short_help=("Get an artifact."))
+@click.option(
+    "--artifact",
+    "-a",
+    type=click.STRING,
+    default=None,
+    required=False,
+    help="The name artifact to retrieve.",
```
dbczumar, line 105:
```suggestion
    help="The name of the artifact to retrieve.",
```

mlflow/pipelines/pipeline.py
```diff
@@ -180,7 +180,7 @@ def _resolve_pipeline_steps(self) -> List[BaseStep]:
 
     @experimental
     @abc.abstractmethod
-    def get_artifact(self, artifact_name: str):
+    def get_artifact(self, artifact_name: str, from_cli: bool = False):
```
dbczumar, line 183:
```suggestion
    def get_artifact(self, artifact_name: str, get_path: bool = False):
```

dbczumar (approved):
LGTM! Thanks @prithvikannan

mlflow/pipelines/cli.py
```diff
@@ -93,3 +93,22 @@ def inspect(step, profile):
...
     Pipeline(profile=profile).inspect(step)
+
+
+@commands.command(short_help=("Get the location of an artifact."))
+@click.option(
+    "--artifact",
+    "-a",
+    type=click.STRING,
+    default=None,
+    required=False,
```
jinzhang21, line 104: `artifact` is not required? What's the expected behavior for None?

mlflow/pipelines/pipeline.py
```diff
@@ -180,7 +180,7 @@ def _resolve_pipeline_steps(self) -> List[BaseStep]:
 
     @experimental
     @abc.abstractmethod
-    def get_artifact(self, artifact_name: str):
+    def get_artifact(self, artifact_name: str, get_path: bool = False):
```
jinzhang21, line 183: I don't like changing this API to return either an object or a path depending on an input parameter.

Can we add a separate function like `get_artifact_path` to specifically address this needs?

tests/pipelines/test_cli.py
```diff
@@ -62,3 +62,13 @@ def test_pipelines_cli_works_with_non_default_profile():
     result = CliRunner().invoke(pipelines_cli.clean, env={_PIPELINE_PROFILE_ENV_VAR: "databricks"})
     assert "with profile: 'databricks'" in result.output
     assert result.exit_code == 0
+
+
+@pytest.mark.usefixtures("enter_pipeline_example_directory", "clean_up_pipeline")
+def test_pipelines_get_artifact_works():
```
jinzhang21, line 68: Add tests on non-existent artifact name and empty artifact name cases.

mlflow/pipelines/cli.py
prithvikannan: good catch. will fix this.

mlflow/pipelines/pipeline.py
prithvikannan: yea i explored this option originally. i believe that there will be a lot of duplicated code with a separate function `get_artifact_path`.

cc: @dbczumar

dbczumar: No strong preference here.

prithvikannan:
@jinzhang21 per offline convo, i have implemented `_get_artifact_path()`. re-requesting your review

mlflow/pipelines/regression/v1/pipeline.py
```diff
@@ -579,18 +542,69 @@ def read_dataframe(artifact_name, output_dir, file_name, step_name):
...
+        transform_output_dir = get_step_output_path(
+            self._pipeline_root_path, transform_step.name, ""
+        )
+        train_output_dir = get_step_output_path(self._pipeline_root_path, train_step.name, "")
+
+        ingest_scoring_output_dir = get_step_output_path(
+            self._pipeline_root_path, ingest_scoring_step.name, ""
+        )
+        predict_output_dir = get_step_output_path(self._pipeline_root_path, predict_step.name, "")
+        register_output_dir = get_step_output_path(self._pipeline_root_path, register_step.name, "")
```
jinzhang21, line 582: Shall we move these into the if-else clause so that only the directory needed by the artifact is computed? The cost of doing so is low in current implementation but it could be expensive if we separate the output storage to some remote location.

jinzhang21 (approved):
LGTM! Thanks, @prithvikannan !

mlflow/pipelines/regression/v1/pipeline.py
```diff
@@ -579,18 +542,80 @@ def read_dataframe(artifact_name, output_dir, file_name, step_name):
...
+
+        else:
+            raise MlflowException(
+                f"The artifact with name '{artifact_name}' is not supported.",
+                error_code=INVALID_PARAMETER_VALUE,
             )
 
+    def _get_artifact_path(self, artifact_name: str) -> Optional[str]:
+        """
+        Returns a path that may or may not exist
```
dbczumar, line 558:
```suggestion
        Returns a path to an artifact, which may or may not exist depending on whether or not the corresponding pipeline step has been run.
```

mlflow/pipelines/pipeline.py
```diff
@@ -196,6 +196,18 @@ def get_artifact(self, artifact_name: str):
         """
         pass
 
+    @experimental
+    @abc.abstractmethod
+    def _get_artifact_path(self, artifact_name: str):
+        """
+        Read an artifact from pipeline output. artifact names can be obtained from
+        `Pipeline.inspect()` or `Pipeline.run()` output.
```
dbczumar, line 204: This isn't quite accurate - this method returns the path to a pipeline artifact output. It does not read the artifact contents.

mlflow/pipelines/pipeline.py
```diff
@@ -196,6 +196,18 @@ def get_artifact(self, artifact_name: str):
...
+
+        Returns None if the specified artifact is not found.
```
dbczumar, line 206: Can we update the docstring to explain what happens when the artifact is found?

tests/pipelines/test_cli.py
```diff
@@ -62,3 +62,34 @@ def test_pipelines_cli_works_with_non_default_profile():
...
+    result = CliRunner().invoke(
+        pipelines_cli.get_artifact,
+        args="--artifact model",
+        env={_PIPELINE_PROFILE_ENV_VAR: "local"},
+    )
+    assert result.exit_code == 0
```
dbczumar, line 74: Can we assert that the artifact path is present in the output?

mlflow/pipelines/cli.py
```diff
@@ -93,3 +93,22 @@ def inspect(step, profile):
     nothing is displayed.
     """
     Pipeline(profile=profile).inspect(step)
+
+
+@commands.command(short_help=("Get the location of an artifact."))
```
dbczumar, line 98:
```suggestion
@commands.command(short_help=("Get the location of an artifact output from the pipeline."))
```

PR #6518, opened 2022-08-19
Add Tracking Uri to run cli

### LINKED ISSUE #4830: [BUG] RESOURCE_DOES_NOT_EXIST when mlflow call start_run()
### Willingness to contribute

- [X] No. I cannot contribute a bug fix at this time.

### System information
- **Have I written custom code (as opposed to using a stock example script provided in MLflow)**:
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Linux Ubuntu 18.04 - **AWS EC2**
- **MLflow installed from (source or binary)**: conda
- **MLflow version (run ``mlflow --version``)**: mlflow, version 1.20.2
- **Python version**: 3.6.9
- **Exact command to reproduce**: mlflow.start_run()

### Describe the problem
I have remote tracking server (the access policies for EC2 to server are setted correct, but I'm not sure at 100%).
I have a main run (parent), and under that parent I also have a few child runs. The issue is related to first start_run() (parent run). When the script calls `with mlflow.start_run()`, script crashes.

The resposne from server calls: `RESOURCE_DOES_NOT_EXIST` when looking for run_id

### Code to reproduce issue
```
remote_server_uri = "http://x.x.x.x:xxxx" # set to your server URI
    mlflow.set_tracking_uri(remote_server_uri)
    mlflow.set_experiment('/cargo_movement')
    # You can get the path at the root of the MLflow project with this:
    root_path = os.path.abspath('.')

    # Check which steps we need to execute
    if isinstance(config["main"]["execute_steps"], str):
        # This was passed on the command line as a comma-separated list of steps
        steps_to_execute = config["main"]["execute_steps"].split(",")
    else:

        steps_to_execute = list(config["main"]["execute_steps"])
    
    with mlflow.start_run() as parent_run:
        # Download step
        if "1_download" in steps_to_execute:

            _ = mlflow.run(
                os.path.join(root_path, "1_download"),
                "main",
                parameters={
                    "parent_run_id": parent_run.info.run_id,
                }
            )
        ...
```

### Other info / logs
```
$ mlflow run .
2021/09/18 13:30:47 INFO mlflow.projects.utils: === Created directory /tmp/tmpy661fhzb for downloading remote URIs passed to arguments of type 'path' ===
2021/09/18 13:30:47 INFO mlflow.projects.backend.local: === Running command 'source /home/ubuntu/anaconda3/bin/../etc/profile.d/conda.sh && conda activate mlflow-167823303a9c0913bc4240ea63b3cb92329b0538 1>&2 && python main.py' in run with ID 'f7b8bafb58404dcb8e27ae1b901b2524' === 
ENV VAR: f7b8bafb58404dcb8e27ae1b901b2524
  File "/home/ubuntu/fchardnet/main.py", line 109, in <module>
    go(config)
  File "/home/ubuntu/fchardnet/main.py", line 25, in go
    with mlflow.start_run() as parent_run:
  File "/home/ubuntu/anaconda3/envs/mlflow-167823303a9c0913bc4240ea63b3cb92329b0538/lib/python3.9/site-packages/mlflow/tracking/fluent.py", line 204, in start_run
    active_run_obj = client.get_run(existing_run_id)
  File "/home/ubuntu/anaconda3/envs/mlflow-167823303a9c0913bc4240ea63b3cb92329b0538/lib/python3.9/site-packages/mlflow/tracking/client.py", line 150, in get_run
    return self._tracking_client.get_run(run_id)
  File "/home/ubuntu/anaconda3/envs/mlflow-167823303a9c0913bc4240ea63b3cb92329b0538/lib/python3.9/site-packages/mlflow/tracking/_tracking_service/client.py", line 65, in get_run
    return self.store.get_run(run_id)
  File "/home/ubuntu/anaconda3/envs/mlflow-167823303a9c0913bc4240ea63b3cb92329b0538/lib/python3.9/site-packages/mlflow/store/tracking/rest_store.py", line 132, in get_run
    response_proto = self._call_endpoint(GetRun, req_body)
  File "/home/ubuntu/anaconda3/envs/mlflow-167823303a9c0913bc4240ea63b3cb92329b0538/lib/python3.9/site-packages/mlflow/store/tracking/rest_store.py", line 56, in _call_endpoint
  File "/home/ubuntu/anaconda3/envs/mlflow-167823303a9c0913bc4240ea63b3cb92329b0538/lib/python3.9/site-packages/mlflow/utils/rest_utils.py", line 217, in call_endpoint
  File "/home/ubuntu/anaconda3/envs/mlflow-167823303a9c0913bc4240ea63b3cb92329b0538/lib/python3.9/site-packages/mlflow/utils/rest_utils.py", line 169, in verify_rest_response
mlflow.exceptions.RestException: RESOURCE_DOES_NOT_EXIST: Run with id=f7b8bafb58404dcb8e27ae1b901b2524 not found
2021/09/18 13:30:48 ERROR mlflow.cli: === Run (ID 'f7b8bafb58404dcb8e27ae1b901b2524') failed ===
```

Components 
- [X] `area/examples`: Example code
- [X] `area/projects`: MLproject format, project running backends

Interface 

Language 

Integrations

### LINKED ISSUE #6519: [BUG] CLI Arguments do not match Python API environment and they have higher precedence  when they should not
### Willingness to contribute

Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.

### MLflow version

1.28.0

### System information

- Ubuntu 20.04
- Python 3.9.0

### Describe the problem

NOTE: This BUG references #4830 due to the similar behavior.

There seems to be a bug in the CLI vs API configuration.
Scenario 1:
You have `main.py` that has `set_experiment` function call. Example:
`mlflow.set_experiment("my-experiment")`
Running the file using `python main.py` will correctly run and set the experiment.

Scenario 2:
You have `main.py` that has `set_tracking_uri`. Example:
`mlflow.set_tracking_uri("http://localhost:5000")`
Running the file using `python main.py` will correctly run and set the tracking url.

Scenario 3:
Using the same file in Scenario 1:
Running the file using `mlflow run /path/to/project` will give an error
```
Cannot start run with ID {} because active run ID does not match environment run ID. Make sure --experiment-name or --experiment-id matches experiment set with set_experiment(), or just use command-line
```
Running the file using `mlflow run /path/to/project --experiment-name my-experiment` works.
We note from this Scenario that there is an override happening here where the CLI's default experiment name `Default` is overriding the python API call to `set_experiment`. Usually CLI commands should override the code, but in this scenario the CLI command was NOT passed, therefore the API should override.

Scenario 4:
Using the same file as Scenario 2:
Running the file using `mlflow run /path/to/project/` will give an error
```
mlflow.exceptions.RestException: RESOURCE_DOES_NOT_EXIST: Run with id={} not found
```
There is no CLI argument to pass the tracking_uri therefore this scenario will NEVER work, thus creating a huge bug in the API. Current workaround envolves the use of environment variables, but this behavior completely eliminates the API `set_tracking_uri` as it simply cannot be used.

Potential fixes include either giving the API higher priority over the CLI (with a warning that CLI and API override) or to add a parameter `tracking-uri` to the `run` command to allow user to pass tracking URI.

I have added tracking uri to the cli as a proposed fix for Scenario 4 (but not 3 as that requires major changes to the API). The fix is found as PR #6518

### Tracking information

_No response_

### Code to reproduce issue

See above

### Stack trace

See above

### Other info / logs

_No response_

### What component(s) does this bug affect?

- [X] `area/projects`: MLproject format, project running backends

### What interface(s) does this bug affect?

### What language(s) does this bug affect?

### What integration(s) does this bug affect?

thethiny:
## Related Issues/PRs

Close #6519
Close #4830

Adding --tracking-uri to `mlflow run`

## How is this patch tested?
Tested with respect to `mlflow run` and `python /path/to/pipeline.py`

`set_tracking_uri` API does not work when used with `mlflow run` as opposed to `python path/to/file.py`, therefore provided a CLI parameter that accepts it.

Documentation shall be changed in case accepted to add new cli param.

Check attached Issue above.

Components
- [X] `area/projects`: MLproject format, project running backends
- [X] `area/server-infra`: MLflow Tracking server backend

Interface

Language

Integrations


thethiny:
According to the documentation, the tracking-uri affects subsequent runs but not the current run. So I didn't know if I should add a check similar to this one
image

In case the merge is approved this can be discussed further to give an error to the user when the tracking-uri passed in CLI is different than the one passed in the API.

mlflow/projects/__init__.py
```diff
@@ -262,6 +282,9 @@ def run(
                         - virtualenv: use virtualenv (and pyenv for Python version management)
 
                         If unspecified, default to conda.
+    :param tracking_uri: The uri which the tracking is supposed to interact with.
+                     If ``None``, the tracking will be against local mlflow server.
+                     If environment variable ``MLFLOW_TRACKING_URI`` is set, it will be used by default.
```
dbczumar, line 287:
```suggestion
                     If ``None``, the currently configured tracking URI reflected by
                     :py:func:`mlflow.get_tracking_uri()` is used by default.
```

mlflow/projects/__init__.py
```diff
@@ -321,8 +344,12 @@ def run(
...
     if backend == "databricks":
         mlflow.projects.databricks.before_run_validations(mlflow.get_tracking_uri(), backend_config)
-    elif backend == "local" and run_id is not None:
-        backend_config_dict[MLFLOW_LOCAL_BACKEND_RUN_ID_CONFIG] = run_id
+    else:
+        if backend == "local" and run_id is not None:
+            backend_config_dict[MLFLOW_LOCAL_BACKEND_RUN_ID_CONFIG] = run_id
+        if tracking_uri is not None:
+            if _resolve_tracking_uri(tracking_uri) is not None:
+                mlflow.set_tracking_uri(_resolve_tracking_uri(tracking_uri))
```
dbczumar, line 352: Can we set the tracking URI for the project by propagating it to the project backends: `databricks` & `kubernetes` (https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/__init__.py#L141-L192) as well as pluggable backends inheriting from `AbstractBackend` (https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/backend/abstract_backend.py#L4)?

This way, `mlflow.projects.run()` doesn't have side effects, which is currently the case if we set the global tracking URI via `mlflow.set_tracking_uri()`.

Fortunately, the abstract pluggable backend's `run()` method already accepts a `tracking_uri` parameter, so we should just be able to thread this tracking URI straight through. For Databricks and Kubernetes, we can set the `MLFLOW_TRACKING_URI` environment variable in the resulting cluster environment specification and Kubernetes deployment specification, respectively.

mlflow/projects/__init__.py
```diff
@@ -321,8 +344,12 @@ def run(
 
     if backend == "databricks":
         mlflow.projects.databricks.before_run_validations(mlflow.get_tracking_uri(), backend_config)
-    elif backend == "local" and run_id is not None:
-        backend_config_dict[MLFLOW_LOCAL_BACKEND_RUN_ID_CONFIG] = run_id
+    else:
+        if backend == "local" and run_id is not None:
+            backend_config_dict[MLFLOW_LOCAL_BACKEND_RUN_ID_CONFIG] = run_id
+        if tracking_uri is not None:
+            if _resolve_tracking_uri(tracking_uri) is not None:
```
dbczumar, line 351: I don't think we need `_resolve_tracking_uri`. It should be sufficient to check if `tracking_uri` is not None`

mlflow/projects/__init__.py
thethiny: I didn't want to get near this because I found that there's a separate handler for databricks that goes through a different pipeline. If I do the change as you suggested the databricks api will be prone to errors.

mlflow/projects/__init__.py
dbczumar: On line 346, `mlflow.projects.databricks.before_run_validations()` is called on `mlflow.get_tracking_uri()`, rather than the specified `tracking_uri`. Can we address that as well?

mlflow/cli.py
```diff
@@ -135,6 +135,12 @@ def cli():
     help="The name to give the MLflow Run associated with the project execution. If not specified, "
     "the MLflow Run name is left unset.",
 )
+@click.option(
+    "--tracking-uri",
+    envvar=tracking._TRACKING_URI_ENV_VAR,
+    help="The URI to which to send tracking data. If not specified, "
+    "the tracking value is file:///<current_path>/mlruns"
```
dbczumar, line 142:
```suggestion
    help="The URI to which to send tracking data. If not specified, "
    "the tracking URI is resolved based on the current environment."
```

dbczumar (commented):
@thethiny Apologies for the delay here. Awesome PR! Just left a few comments before merge - let me know if you have any questions :)

mlflow/projects/__init__.py
thethiny: Will update accordingly

thethiny:
@dbczumar thanks for your comments. Regarding databricks I didn't do any changes to that, as I explained in the comment under your second suggestion. I'm open to your suggestions and proposals I just need to understand the incentive first. Would you like me to override the databricks uri with the passed one through the cli? As I noticed the databricks has its own override. What do you purpose the behavior to be?

dbczumar:
@thethiny Thanks for clarifying! If the user specifies a tracking URI via the CLI, it should be used regardless of which backend is specified, Databricks or otherwise. It should be safe to pass that specified URI along in place of the preexisting default, after verifying that the specified URI is not None.

mlflow/projects/__init__.py
thethiny: For your first comment, I will remove `_resolve_tracking_uri` completely. The only reason I added it is because I wanted to give the environment higher precedence than the CLI but after our conversation it's apparent to me that you want the CLI to have more power over the Env Vars. Please correct me if I'm wrong @dbczumar

mlflow/projects/__init__.py
thethiny: I'm going to address this in a commit on its own as I believe I would require further discussion with you (perhaps different PR?). For now I will handle the other comments.
From what I've seen in the code there are multiple functions that assume their own `tracking_uri` such as https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/databricks.py#L404.
And if I understand you correctly, you want me to update all the functions so that they **receive** and **pass** `tracking_uri` wherever it's applicable so that it can be passed in. Though what I don't understand is that if you want me to pass the `tracking_uri` variable on its own to each and every or if you want me to `mlflow.set_tracking_uri()` in the `run()` function and all functions then will automatically work as intended since they will rely on `mlflow.get_tracking_uri()`.

Example:
Either accept `tracking_uri` in the line shown above and pass it on as `databricks_profile_uri=tracking_uri or tracking.get_tracking_uri()`
or
do the `None` check on tracking uri in the [`run()`](https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/__init__.py#L344) function **only**.

thethiny:
I'm going to address [this](https://github.com/mlflow/mlflow/pull/6518#discussion_r956776806) in a commit on its own as I believe I would require further discussion with you (perhaps different PR?). For now I will handle the other comments.
From what I've seen in the code there are multiple functions that assume their own `tracking_uri` such as https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/databricks.py#L404.
And if I understand you correctly, you want me to update all the functions so that they **receive** and **pass** `tracking_uri` wherever it's applicable so that it can be passed in. Though what I don't understand is that if you want me to pass the `tracking_uri` variable on its own to each and every or if you want me to `mlflow.set_tracking_uri()` in the `run()` function and all functions then will automatically work as intended since they will rely on `mlflow.get_tracking_uri()`.

Example:
Either accept `tracking_uri` in the line shown above and pass it on as `databricks_profile_uri=tracking_uri or tracking.get_tracking_uri()`
or
do the `None` check on tracking uri in the [`run()`](https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/__init__.py#L344) function **only**.

@dbczumar

dbczumar:
> I'm going to address [this](https://github.com/mlflow/mlflow/pull/6518#discussion_r956776806) in a commit on its own as I believe I would require further discussion with you (perhaps different PR?). For now I will handle the other comments. From what I've seen in the code there are multiple functions that assume their own `tracking_uri` such as https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/databricks.py#L404. And if I understand you correctly, you want me to update all the functions so that they **receive** and **pass** `tracking_uri` wherever it's applicable so that it can be passed in. Though what I don't understand is that if you want me to pass the `tracking_uri` variable on its own to each and every or if you want me to `mlflow.set_tracking_uri()` in the `run()` function and all functions then will automatically work as intended since they will rely on `mlflow.get_tracking_uri()`.
> 
> Example: Either accept `tracking_uri` in the line shown above and pass it on as `databricks_profile_uri=tracking_uri or tracking.get_tracking_uri()` or do the `None` check on tracking uri in the [`run()`](https://github.com/thethiny/mlflow/blob/0acb133389f4c89c65ef24f274cde4bb04af1b79/mlflow/projects/__init__.py#L344) function **only**.
> 
> @dbczumar

@thethiny Apologies for the lack of clarity. We should do `tracking_uri = tracking_uri or mlflow.get_tracking_uri()` in `run()` and then pass `tracking_uri` to `run_databricks` by adding a parameter and do something similar for Kubernetes and the other MLflow Project backends.

thethiny:
> @thethiny Apologies for the lack of clarity. We should do `tracking_uri = tracking_uri or mlflow.get_tracking_uri()` in `run()` and then pass `tracking_uri` to `run_databricks` by adding a parameter and do something similar for Kubernetes and the other MLflow Project backends.

So what I will do is add a parameter `tracking_uri` to every function that uses it or propagates it. Will ping you after I commit.

harupy:
@mlflow-automation autoformat

harupy:
@thethiny We updated the command to trigger auto-formatting today.

thethiny:
> @thethiny We updated the command to trigger auto-formatting today. 
Thank you. I will most likely commit a push either today or tomorrow, is there a way to commit the push along with auto formatting at the same time?

harupy:
> is there a way to commit the push along with auto formatting at the same time?

No. I think you can fix lint errors, and then push a commit.

thethiny:
>  Note the auto-formatting bot doesn't fix pylint errors.

Yes I'm aware of this but I usually use black formatter that doesn't abide by the lint you're using (for example I don't have the 100 character limit). So I was planning on fixing the lint issues but was hoping the bot would autoformat the line count.

harupy:
> the bot would autoformat the line count.

The bot doesn't because `line-too-long` is a pylint error.

thethiny:
@mlflow-automation autoformat

thethiny:
@dbczumar I have pushed commit per each comment we have agreed on. Please let me know if there is anything I missed or if you have any other comments. Here are my draft notes about the merge for the [Propagation discussion](https://github.com/mlflow/mlflow/pull/6518#issuecomment-1230023241):

`run_databricks` in databricks and `DatabricksJobRunner` added `tracking_uri`
`get_run_env_vars` added *optional* `tracking_uri` field and organized variables
exported `tracking._resolve_tracking_uri`
`run` and `_run` added `tracking_uri`
`projects.databricks._get_tracking_uri_for_run` added *optional* `tracking_uri` to return correct environment
- **Decision to add it in the function instead of run_databricks is because the function handles URIs that start with** `databricks` so we don't wanna mess with that

`projects.backend.local _run_entry_point` accepts `tracking_uri`
`LocalBackend` now utilizes the `tracking_uri` param
`_invoke_mlflow_run_subprocess` accepts `tracking_uri`

added 2 **tests** for `get_tracking_uri_for_run` that utilize `tracking_uri`

**areas not touched**
- paddle._paddle_autolog.patched_fit
- pipelines:
    - steps.evaluate.EvaluateStep._build_profiles_and_card > _url = get_databricks_run_url
    - steps.register.RegisterStep._build_card > _url = get_databricks_run_url
    - steps.train.TrainStep._build_card > _url = get_databricks_run_url
- pytorch._pytorch_autolog.patched_fit
- store.artifact.databricks_artifact_repo.DatabricksArtifactRepository.__init__ > self.databricks_profile_uri
- store.artifact.mlflow_artifacts_repo.MlflowArtifactsRepository.__init__ > super()
- tests.helper_functions.pyfunc_serve_and_score_model > env
- 
*The tests should probably be replicated with testing against a provided tracking_uri*


I have a comment that is unrelated to this PR, perhaps even @harupy can shed some information about as well.
`projects.backend.local.LocalBackend.run` > `work_dir = fetch_and_validate`. `fetch_and_validate` already calls `load_project` so calling `load_project` in the next line is redundant, which happens multiple times in the codebase. Shouldn't the validate function return the `project` or at least have an `override` where `return_project = True` returns the `work_dir` and the `project`?

mlflow/projects/backend/local.py
```diff
@@ -180,8 +193,11 @@ def _invoke_mlflow_run_subprocess(
         run_id=run_id,
         parameters=parameters,
     )
-    env_vars = get_run_env_vars(run_id, experiment_id)
-    env_vars.update(get_databricks_env_vars(mlflow.get_tracking_uri()))
+
+    env_vars = get_run_env_vars(run_id, experiment_id, tracking_uri)
+    env_vars.update(
+        get_databricks_env_vars(tracking_uri=tracking._resolve_tracking_uri(tracking_uri))
+    )
```
thethiny, line 200: Suggestion to replace with with
```python
def get_databricks_run_env_vars(run_id, experiment_id, tracking_uri):
  env_vars = get_run_env_vars(run_id, experiment_id, tracking_uri)
  env_vars.update(get_databricks_env_vars(tracking_uri=tracking._resolve_tracking_uri(tracking_uri)))
  return env_vars
```

mlflow/projects/backend/local.py
```diff
@@ -226,17 +242,19 @@ def _run_mlflow_run_cmd(mlflow_run_arr, env_map):
...
     :param work_dir: Working directory in which to run the command
     :param run_id: MLflow run ID associated with the entry point execution.
+    :param tracking_id: The URI to which to send tracking data. If not specified,
+    the tracking URI is resolved based on the current environment.
     """
     env = os.environ.copy()
-    env.update(get_run_env_vars(run_id, experiment_id))
-    env.update(get_databricks_env_vars(tracking_uri=mlflow.get_tracking_uri()))
+    env.update(get_run_env_vars(run_id, experiment_id, tracking_uri))
+    env.update(get_databricks_env_vars(tracking_uri=tracking._resolve_tracking_uri(tracking_uri)))
```
thethiny, line 257: suggestion similar as before, `env.update(get_databricks_run_env_vars(...))`

thethiny:
Added #4830 to closelist

thethiny:
@mlflow-automation autoformat

mlflow/projects/__init__.py
```diff
@@ -319,10 +328,20 @@ def run(
...
-    elif backend == "local" and run_id is not None:
-        backend_config_dict[MLFLOW_LOCAL_BACKEND_RUN_ID_CONFIG] = run_id
+        _databricks_uri = tracking_uri if _has_valid_tracking_uri else mlflow.get_tracking_uri()
+        mlflow.projects.databricks.before_run_validations(_databricks_uri, backend_config)
+    else:
+        if backend == "local" and run_id is not None:
+            backend_config_dict[MLFLOW_LOCAL_BACKEND_RUN_ID_CONFIG] = run_id
+        if _has_valid_tracking_uri:
+            mlflow.set_tracking_uri(tracking_uri)
+            # prevent setting tracking_uri to the env MLFLOW_TRACKING_URI
```
dbczumar, line 344: Now that we're passing `tracking_uri` to `_run()`, I think we can safely remove `mlflow.set_tracking_uri()` and the associated `if _has_valid_tracking_uri` check

mlflow/projects/backend/local.py
dbczumar: Creating a helper function sounds great!

mlflow/projects/__init__.py
dbczumar: Can we also update the Project execution logic for Kubernetes to use the  specified tracking URI? https://github.com/mlflow/mlflow/blob/c20c664e037b895ff4bd49dda4a26604e55a9b93/mlflow/projects/__init__.py#L139-L179. Setting the `MLFLOW_TRACKING_URI` environment variable seems like a good approach, similar to your approach for Databricks support.

dbczumar (commented):
@thethiny Apologies for the delay. Awesome work here! I left a couple more small comments, after which we should be ready to merge. Thank you so much for all your effort!

mlflow/projects/__init__.py
thethiny: I will have a look 👍🏻

thethiny:
@dbczumar is we remove mlflow.set_tracking_uri() I believe the `local` backend will no longer use the passed tracking uri since I haven't seen a place where `local` utilises it, did I miss something?
And thanks for your comments!

thethiny:
@dbczumar According to your [previous comment](https://github.com/mlflow/mlflow/pull/6518#discussion_r963316373), I think you applied the review on a different commit as that is resolved in the current `HEAD`. As you can see in the [current state](https://github.com/thethiny/mlflow/blob/add-track-uri/mlflow/projects/__init__.py#L170-L174) Lines 170 to 174 reflect the expected behavior as they propagate the `tracking_uri` into the environment using [`get_run_env_vars`](https://github.com/thethiny/mlflow/blob/3ba76d5d2e1fd324a1afd31a8e748a2bfa38dd88/mlflow/projects/utils.py#L334-L346)

thethiny:
@harupy unsure if it's intended behavior or not but automation bot is allowed to trigger all runs, as seen above.

mlflow/projects/__init__.py
```diff
@@ -195,6 +200,7 @@ def run(
     run_id=None,
     run_name=None,
     env_manager=None,
+    tracking_uri=None,  # Inserted at the bottom for backwards compatibility
```
dbczumar, line 203: Fantastic! Thank you! 

```suggestion
    tracking_uri=None,
```

thethiny:
@dbczumar all set

dbczumar (commented):
@thethiny This looks great! I tried it out manually and encountered one small issue. I ran 

```
mlflow run --tracking-uri sqlite:///myruns.sqlite --env-manager local .
```

from within https://github.com/mlflow/mlflow/tree/master/examples/sklearn_elasticnet_wine.

It appears that `mlflow run` creates a run using the default tracking URI `./mlruns`, but then the project tries to fetch that run from the specified tracking URI `sqlite:///myruns.sqlite`. This produces the following error:

```
2022/09/12 21:48:51 INFO mlflow.projects.utils: === Created directory /var/folders/v3/fwmb62ds0bd7700x2vgm051r0000gp/T/tmpq0n7izqu for downloading remote URIs passed to arguments of type 'path' ===
2022/09/12 21:48:51 INFO mlflow.projects.backend.local: === Running command 'python train.py 0.5 0.1' in run with ID '4e449580c4fe4915976b50cf87b79675' ===
  File "train.py", line 60, in <module>
    with mlflow.start_run():
  File "/Users/corey.zumar/mlflow/mlflow/tracking/fluent.py", line 288, in start_run
    active_run_obj = client.get_run(existing_run_id)
  File "/Users/corey.zumar/mlflow/mlflow/tracking/client.py", line 150, in get_run
    return self._tracking_client.get_run(run_id)
  File "/Users/corey.zumar/mlflow/mlflow/tracking/_tracking_service/client.py", line 72, in get_run
    return self.store.get_run(run_id)
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/sqlalchemy_store.py", line 632, in get_run
    run = self._get_run(run_uuid=run_id, session=session, eager=True)
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/sqlalchemy_store.py", line 560, in _get_run
    raise MlflowException(
mlflow.exceptions.MlflowException: Run with id=4e449580c4fe4915976b50cf87b79675 not found
2022/09/12 21:48:55 ERROR mlflow.cli: === Run (ID '4e449580c4fe4915976b50cf87b79675') failed ===
```

Fortunately, it seems we could still address #4830 and #6519 without introducing a `--tracking-uri` argument. The important piece of the changes you've made is that we're now passing along the currently-configured MLflow tracking URI to the project. Assuming users can configure the URI using `mlflow.set_tracking_uri()` and that they don't need to use a different URI for project executions, we can remove the argument from `mlflow run` and `mlflow.run()` in the PR.

Alternatively, we can fix this by passing the tracking URI to https://github.com/mlflow/mlflow/blob/2b2fe58742ce44c6b88d21f6f555f897c5ad2745/mlflow/projects/__init__.py#L115-L117. We'll also want to pass it to the `MlflowClient()` constructors that are used to 1. set tags on the run and 2. periodically check the run's status throughout the files in https://github.com/mlflow/mlflow/tree/master/mlflow/projects. Though I'm leaning towards the former approach where we remove the `tracking_uri` argument from `mlflow.run()`.

Let me know what your thoughts are here :)

thethiny:
@dbczumar there's actually a 3rd approach but it would be inconsistent with our changes.
We can undo the Removal of https://github.com/mlflow/mlflow/pull/6518#discussion_r963311493 which will globally set the tracking uri so that every other request to get the uri will return the correct value.

The problem in both of the issues was that when you use mlflow run it would initialize the entire project using the default tracking uri, so that when you manually use `set_tracking_uri` it would be too late into the initialization process that it no longer has any effect as the default uri was already propagated throughout the entire project, hence my proposal to pass the tracking uri as a parameter.
However, I'm not opposing your proposal to remove it if it will work but I fail to understand the flow of the code as in how its removal will fix the issue. It appears to me that its removal would take us back to square 1 with `mlflow run` but would still work fine with `python file.py`.
On the other hand, passing the tracking URI to client and to the other function you've identified would open the door something bigger entirely which is passing the uri to the tracking handlers, not only project initializers which would require some work, but I think that sooner or later the parameter will have to be passed to those sections anyway.

Let me know what you feel about this, and if we're going to be adding the uri to the other places then I would like to ask you to point me to where I should add it, as this keeps getting bigger and I'm starting to lose track. My repo is also write enable for maintainers so if you had any commits to push you can do them directly if you want to.

Looking forward to your response.

PR #6525, opened 2022-08-20
[ML-24419] Updated artifact downloads to pass in http headers

kriscon-db:
## Related Issues/PRs

#xxx

We add the passthrough of encryption headers recieved from MLflow service into the download request for both artifacts and models.

## How is this patch tested?

I modified existing tests to verify that the headers are passed through properly.

Components

Interface

Language

Integrations


mlflow/store/artifact/databricks_models_artifact_repo.py
```diff
@@ -110,12 +110,19 @@ def _get_signed_download_uri(self, path=None):
...
                 " status code %s. Response body: %s" % (path, response.status_code, response.text)
             )
-        return json_response.get("signed_uri", None)
+        return json_response.get("signed_uri", None), json_response.get("headers", None)
+
+    @staticmethod
+    def _process_raw_headers(raw_headers):
+        if raw_headers == None or len(raw_headers) == 0:
+            return {}
+        return {cur_header.get("key"): cur_header.get("value") for cur_header in raw_headers}
```
kriscon-db, line 119: not sure how defensive we want to be here about malformed header responses/changes to the header structure.

dbczumar: `DatabricksArtifactRepo._convert_http_headers` seems more robust. Can we use that implementation? For example, `cur_header.get("key")` will return `None` if `key` isn't defined as opposed to skipping the entry.

dbczumar (approved):
LGTM! Can we merge this into a feature branch? Happy to stamp a separate PR :)

mlflow/store/artifact/databricks_models_artifact_repo.py
kriscon-db: I can do this but I'll need to patch the universe protobuf definition of HttpHeader to match the one that the databricks artifact repo is using.  I think there is a slight difference in field naming.

dbczumar: I see - no need in that case. Separate method is fine, but let's make sure it's similarly robust to issues

kriscon-db: I have a PR to fix the headers in universe.  I had originally made that method in DatabricksArtifactRepo static but I don't actually like how that looks and have now backed out and we just have convenience methods in both classes.  If we really want, we can set up a utils.py file and have both classes read from it, but we can do that in a follow on commit.

Issue #6554, opened 2022-08-23
[BUG] Truncation issues when using default_evaluator with explainability


dsgibbons:
### Willingness to contribute

Yes. I can contribute a fix for this bug independently.

### MLflow version

eb273d656fa2778cf6c98031fd33c9ee85bec304

### System information

- macOS
- Python 3.7.13

### Describe the problem

#6503 was recently fixed to allow `mlflow.evaluate` to accept non-string column names. Since the bug fix, if the code example from #6503 is changed to set `log_model_explainability=True`, the code will fail when the feature names are truncated as a result of the call to `truncate_str_from_middle`, as ints don't have a `len` method.

Upon coming up with a fix for this, I actually found that if a feature ends up being given the name: `f"{truncated_name}(f_{i+1})` due to truncation,  `_shap_predict_fn` ends up creating a useless prediction function that fills the DataFrame with NaNs. These NaNs don't prevent the associated test in `tests/test_default_evaluator.py` from passing.

TL;DR:
1. Some minor modifications are required to allow non-string column names during evaluation with explainability.
2. The logic used to form the DataFrame in  `_shap_predict_fn` needs tweaking to account for DataFrames with pre-existing column names. 

I will submit a PR that should demonstrate this issue and the intended fix more clearly.

### Code to reproduce issue

```python
import mlflow

import pandas as pd
from sklearn import datasets
from sklearn.linear_model import LogisticRegression
from sklearn.model_selection import train_test_split

# prepare train and test data
iris = datasets.load_iris()
X = iris.data
y = iris.target
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)

xy_test = pd.DataFrame(X_test)
xy_test["label"] = y_test

with mlflow.start_run():
    logistic = LogisticRegression(C=1.0)
    logistic.fit(X_train, y_train)
    mod = mlflow.sklearn.log_model(logistic, artifact_path="mymodel")

    results = mlflow.evaluate(
        model=mod.model_uri,
        model_type="classifier",
        data=xy_test,
        evaluator_config={
            "log_model_explainability": True,
        },
        evaluators="default",
        targets="label",
    )
```

### Stack trace

**Error 1 (non-string column names with explainability):**
From `mlflow.utils.string_utils.py`:

```python
Traceback (most recent call last):
  File "a.py", line 30, in <module>
    targets="label",
[...]

Issue #6503, opened 2022-08-18
[BUG] mlflow.evaluate() fails for dataframes containing non-string column names


dbczumar:
### Willingness to contribute

No. I cannot contribute a bug fix at this time.

### MLflow version

1.28.0

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: MacOS
- **Python version**: 3.8
- **yarn version, if running the dev UI**: NA

### Describe the problem

When passing a Pandas DataFrame dataset to ``mlflow.evaluate()`` that contains non-string column names, evaluation fails because it attempts to string concatenate the column names with the `,` separator without converting the column names to strings:

```
Traceback (most recent call last):
  File "eval.py", line 50, in <module>
    results = mlflow.evaluate(
  File "/Users/corey.zumar/mlflow/mlflow/models/evaluation/base.py", line 1164, in evaluate
    dataset = EvaluationDataset(
  File "/Users/corey.zumar/mlflow/mlflow/models/evaluation/base.py", line 379, in __init__
    md5_gen.update(",".join(self._feature_names).encode("UTF-8"))
TypeError: sequence item 0: expected str instance, int found
```

### Tracking information

MLflow version: 1.28.0
Tracking URI: file:///Users/corey.zumar/mlexamples/sklearn/mlruns
Artifact URI: file:///Users/corey.zumar/mlexamples/sklearn/mlruns/0/b51876eec22b45468af3c4bb8be84121/artifacts

### Code to reproduce issue

```
import mlflow
from mlflow.models.evaluation import EvaluationDataset
from mlflow.models import MetricThreshold

import pandas as pd
from sklearn import datasets
from sklearn.linear_model import LogisticRegression
from sklearn.model_selection import train_test_split

# prepare train and test data
iris = datasets.load_iris()
X = iris.data
y = iris.target
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)

xy_test = pd.DataFrame(X_test)
xy_test["label"] = y_test

with mlflow.start_run():
    logistic = LogisticRegression(C=1.0)
    logistic.fit(X_train, y_train)
    mod = mlflow.sklearn.log_model(logistic, artifact_path="mymodel")

    results = mlflow.evaluate(
        model=mod.model_uri,
        model_type="classifier",
        data=xy_test,
        evaluator_config={
            "log_model_explainability": False,
        },
        evaluators="default",
        targets="label",
    )
```

### Stack trace

[...]

PR #6555, opened 2022-08-23
Fix truncation issues when using explainable evaluator

dsgibbons:
## Related Issues/PRs

#6554

All changes relate to `default_evaluator.py`. 

1) Use `str(f)` to allow for non-string features to be truncated `truncated_feature_names = [truncate_str_from_middle(str(f), 20) for f in self.feature_names]`
2) Use `str`  to prevent truncation mapping being applied to short non-string names like 1, 2 etc. : `if truncated_name != str(self.feature_names[i])`
3) When creating `shap_predict_fn`, don't create a new DataFrame from `x` if `x` is already a DataFrame. If there is a mismatch between the column names of `x` and the column names passed to `DataFrame(x, columns=feature_names)`, then the resulting DataFrame is filled with NaNs.

## How is this patch tested?

Not yet, but I will write a test to make sure that evaluating with non-string, or very long feature names does not result in NaNs. Once I have written this test, I will remove the draft tag from this PR

Components

Interface

Language

Integrations


harupy:
@dsgibbons Thanks for the RP! Can you add tests?

dsgibbons:
Yes I was planning to add a test tomorrow and then I'll mark it ready to review.

harupy:
Got it, thanks!

dsgibbons:
Introduced a few more alterations:

- added `EvaluationDatasetWithSavedConstructor` to `test_evaluation.py` to remove repeated lines from the various `pytest.fixture` datasets.
- added a new test case using the `iris_pandas_df_unusual_cols_dataset` that specifically checks non-string columns and very long feature names.
- added a manual check for NaN errors to `default_evaluator.py`.

Ideally, I'd prefer that the `default_evaluator` halt execution rather than allowing exceptions to just be written to the logger. Maybe there is a better way we can handle exceptions in  `default_evaluator` - perhaps by allowing the user to decide whether or not the program should be halted if an exception is encountered.

One more thing to note is that we often fit a model with unlabelled columns, but then test with labelled ones. This leads to scikit-learn throwing lots of warnings during evaluation. Should this be raised as an issue?

tests/models/test_evaluation.py
```diff
@@ -101,43 +101,43 @@ def spark_session():
     session.stop()
 
 
+class EvaluationDatasetWithSavedConstructor(EvaluationDataset):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._constructor_args = kwargs
```
harupy, line 107: I think we can define:

```python
Dataset = namedtuple('Dataset', ['data', 'targets', 'name'])
```

and replace:

```python
xxx_yyy_dataset._constructor_args["data"]
xxx_yyy_dataset._constructor_args["targets"]
```

with:

```python
xxx_yyy_dataset.data
xxx_yyy_dataset.targets
```

harupy: We can address this later in a separate PR.

harupy: Can we revert changes related to `EvaluationDatasetWithSavedConstructor`?

dsgibbons: Yes. So just go back to using `EvaluationDataset` with manual setting of `_constructor_args`? And then use `namedtuple` later?

harupy:
> One more thing to note is that we often fit a model with unlabelled columns, but then test with labelled ones.

Is this really true? If we train a model with unlabelled columns, I think we should test with unlabelled columns.

tests/models/test_evaluation.py
harupy: Yes, or you can file a PR to use `namedtuple` first.

dsgibbons:
> > One more thing to note is that we often fit a model with unlabelled columns, but then test with labelled ones.
> 
> Is this really true? If we train a model with unlabelled columns, I think we should test with unlabelled columns.

Perhaps you're right. I was somewhat basing that on the code example from #6503.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -627,6 +632,11 @@ def _log_model_explainability(self):
             # Shap evaluation might fail on some edge cases, e.g., unsupported input data values
             # or unsupported model on specific shap explainer. Catch exception to prevent it
             # breaking the whole `evaluate` function.
+
+            # NaN errors should break evaluation
+            if str(e).find("NaN") != -1:
+                raise e
```
harupy, line 638: Why should NaN errors break evaluation?

dsgibbons:
> Is this really true? If we train a model with unlabelled columns, I think we should test with unlabelled columns.

There are also examples of training on unlabelled columns and testing on labelled columns in `test_evaluation.py`. For example `test_regressor_evaluate` and `test_classifier_evaluate`.

mlflow/models/evaluation/default_evaluator.py
dsgibbons: I was not sure how else to catch the NaNs created by the old truncation logic. The NaNs don't carry through to the results passed via the `evaluate` function. If the user can override the exception handler to raise exceptions then this would be easier to test against.

dsgibbons:
> Is this really true? If we train a model with unlabelled columns, I think we should test with unlabelled columns.

The `DefaultEvaluator` forces the test `pd.DataFrame` to have labelled columns, even if the columns were originally unlabelled. Should this logic be changed to be consistent with the above comment?

harupy:
@WeichenXu123 Could you also take a look? I think you're more familiar than me with this feature.

tests/models/test_evaluation.py
```diff
@@ -301,7 +304,29 @@ def iris_pandas_df_num_cols_dataset():
...
+    eval_X, eval_y = X[0::3], y[0::3]
+    data = pd.DataFrame(
+        {
+            "f1": eval_X[:, 0],
+            2: eval_X[:, 1],
+            "f3longnamelongnamelongname": eval_X[:, 2],
+            123456789012345678901234567890: eval_X[:, 3],
+            "y": eval_y,
+        }
+    )
```
harupy, line 325: Can we use a numpy array?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -540,9 +544,10 @@ def _log_model_explainability(self):
             "explainability_nsamples", _DEFAULT_SAMPLE_ROWS_FOR_SHAP
         )
 
-        truncated_feature_names = [truncate_str_from_middle(f, 20) for f in self.feature_names]
+        truncated_feature_names = [truncate_str_from_middle(str(f), 20) for f in self.feature_names]
```
harupy, line 547: https://github.com/mlflow/mlflow/blob/9c2010c1bf4466ab6f78532411d945d79069c7ba/mlflow/models/evaluation/base.py#L363-L368

Can we stringify feature names here?

cc @WeichenXu123

mlflow/models/evaluation/default_evaluator.py
harupy: Does the NaN error occur in the updated code?

tests/models/test_evaluation.py
dsgibbons: Using a numpy array would test the use of unlabelled columns with explainability, but would not test the truncation feature on long string names

WeichenXu123:
To simplify the code, I think we should stringify the feature names here:

https://github.com/mlflow/mlflow/blob/9c2010c1bf4466ab6f78532411d945d79069c7ba/mlflow/models/evaluation/base.py#L368

And I propose if "column_name" is a integer, we generate the dataset.feature_name to be `feature_{num}`

@dsgibbons @harupy

tests/models/test_evaluation.py
harupy: Ok, can we create two datasets, one for testing unlabelled columns (numpy array) with explainability, and one for testing truncation feature on long string names?

dsgibbons: Yes

dsgibbons:
> To simplify the code, I think we should stringify the feature names here:
> 
> https://github.com/mlflow/mlflow/blob/9c2010c1bf4466ab6f78532411d945d79069c7ba/mlflow/models/evaluation/base.py#L368
> 
> And I propose if "column_name" in pd.Dataframe is not a string, we generate the dataset.feature_name to be `feature_{str(raw_column_name)}`
> 
> @dsgibbons @harupy

That could work. However, sklearn will generate warnings if you do it this way as you are now evaluating a model that was trained on unlabelled columns with labelled column data. My unpushed fix from before avoids these warnings:

```
def _shap_predict_fn(x, predict_fn, feature_names):
    if isinstance(x, pd.DataFrame):
        df = x
        df.columns = feature_names
    else:
        df = pd.DataFrame(x, columns=feature_names)
    return predict_fn(df)
```

mlflow/models/evaluation/default_evaluator.py
dsgibbons: No, but there should probably be a way to catch NaN errors in the tests

dsgibbons: How about this instead? 

```
except Exception as e:
    # Shap evaluation might fail on some edge cases, e.g., unsupported input data values
    # or unsupported model on specific shap explainer. Catch exception to prevent it
    # breaking the whole `evaluate` function.

    if not self.evaluator_config.get("ignore_exceptions"):
        raise e
```

and then the test can just include `evaluator_config={"ignore_exceptions": False}` in the `evaluate` call.

dsgibbons:
@WeichenXu123 are you OK with sklearn models raising warnings about being tested on labelled columns after being trained on unlabelled columns?

WeichenXu123:
> @WeichenXu123 are you OK with sklearn models raising warnings about being tested on labelled columns after being trained on unlabelled columns?

We should address it. We can address this by:

Add a new bool flag `is_dataframe_column_unlabeled` in `EvalautionDataset`, and set it to be True in this case,
and in `shap_predict_fn` code, if this flag is True, then we don't attach auto-generated column name to the dataframe.

dsgibbons:
@WeichenXu123 In the tests, the `EvalautionDataset` is instantiated after the model has already been trained. The `is_dataset_column_unlabeled` flag can't tell us if the model was trained on unlabelled columns. It looks like we would need to check against `model.feature_names_in_` instead. This can't be done inside `EvalautionDataset`. This problem is not just in `_shap_predict_fn` - we get the same warning when not using explainability in `DefaultEvaluator._generate_model_predictions`. I'll make a push soon with my current changes. These changes won't fix the `sklearn` warnings.

dsgibbons:
I'm in a bit of a mess as I made a mistake when syncing with the latest changes from master. Is it OK if I open a new pull request with a clean branch?

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -415,6 +415,11 @@ def _compute_df_mode_or_mean(df):
 
 
 def _shap_predict_fn(x, predict_fn, feature_names):
+    if isinstance(x, pd.DataFrame):
+        df = x.copy(deep=True)
+        df.columns = feature_names
+        return predict_fn(df)
+
     return predict_fn(pd.DataFrame(x, columns=feature_names))
```
dsgibbons, line 424: If `x` is already a `DataFrame` and the column names are mismatched, the resulting `DataFrame` will contain NaNs.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -627,6 +632,10 @@ def _log_model_explainability(self):
             # Shap evaluation might fail on some edge cases, e.g., unsupported input data values
             # or unsupported model on specific shap explainer. Catch exception to prevent it
             # breaking the whole `evaluate` function.
+
+            if not self.evaluator_config.get("ignore_exceptions", True):
+                raise e
+
```
dsgibbons, line 638: added explicit config argument to not bury exceptions - this is useful when trying to detect NaN errors

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -1039,9 +1048,15 @@ def X(self) -> pd.DataFrame:
...
-        )
+        if isinstance(self.dataset.features_data, pd.DataFrame):
+            df = DefaultEvaluator._MutationGuardedData(
+                self.dataset.features_data
+            ).copy_to_avoid_mutation()
+            df.columns = self.dataset.feature_names
+        else:
+            df = pd.DataFrame(self.dataset.features_data, columns=self.dataset.feature_names)
+
+        return DefaultEvaluator._MutationGuardedData(df)
```
dsgibbons, line 1059: required to avoid same problem from `_shap_predict_fn`

tests/models/test_default_evaluator.py
```diff
@@ -329,6 +332,45 @@ def test_svm_classifier_evaluation(svm_model_uri, breast_cancer_dataset):
...
+            dataset._constructor_args["data"],
+            model_type="classifier",
+            targets=dataset._constructor_args["targets"],
+            dataset_name=dataset.name,
+            evaluators="default",
+            evaluator_config={
+                "ignore_exceptions": False,
+            },
+        )
+
```
dsgibbons, line 348: not sure if there is a better, more "pytest" way of doing this

mlflow/models/evaluation/base.py
```diff
@@ -336,7 +338,11 @@ def __init__(self, data, *, targets, name=None, path=None, feature_names=None):
                 self._feature_names = feature_names
             else:
                 self._features_data = data.drop(targets, axis=1, inplace=False)
-                self._feature_names = list(self._features_data.columns)
+                self._feature_names = [stringify_column(c) for c in self._features_data.columns]
+
+            self.is_dataset_column_unlabeled = all(
+                (i == c for i, c in enumerate(self._features_data.columns))
+            )
```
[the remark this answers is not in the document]
WeichenXu123: No need. :)

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: What about adding a member `original_dataframe_column_keys` in `DatasetEvaluator` instead of `is_dataset_column_unlabeled`,

then we modify `_shap_predict_fn` to be

```
def _shap_predict_fn(x, predict_fn, original_dataframe_column_keys):
  return predict_fn(pd.DataFrame(x, columns=original_dataframe_column_keys))
```

and when calling `_shap_predict_fn`, pass `eval_dataset.original_dataframe_column_keys` to `_shap_predict_fn`

WeichenXu123:
> In the tests, the EvaluationDataset is instantiated after the model has already been trained. The is_dataset_column_unlabeled flag can't tell us if the model was trained on unlabelled columns.

* A change proposal : https://github.com/mlflow/mlflow/pull/6555#issuecomment-1231101796
* Always generate the `original_dataframe_column_keys` member from Test dataframe User is responsible to ensure Test dataframe have the same column name / label with the Train dataframe.

> we get the same warning when not using explainability in DefaultEvaluator._generate_model_predictions

Let's fix it as well using similar approach.

dsgibbons:
@WeichenXu123 your proposed fix doesn't address the issue of why the NaNs are being created.

Take `_shap_predict_fn` for example, 

If the local variable `x` has truncated column names, then you get NaNs from `pd.DataFrame(x, columns=feature_names)`. The NaNs do **not** come from the `predict_fn(.)` afterwards (as the models in the tests are trained on unlabelled columns anyway - sklearn will simply raise a warning). In this case, `pd.DataFrame(x, columns=feature_names)` will already contain NaNs as `x.columns != feature_names`. This issue remains even if `feature_names` is replaced with `original_dataframe_column_keys`.

Instead, you need instantiate the `DataFrame` first, and then rename the columns (either using `DataFrame.rename` or directly overwriting the `DataFrame.columns` attribute:

`df = pd.DataFrame(x)`

then either
    
`df.columns = feature_names`

or 

`df = df.rename(columns={k: v for k, v in zip(df.columns, feature_names)})`

This is consistent with the fix I pushed yesterday. Including `original_dataframe_column_keys` as an attribute of `EvaluationDataset` doesn't help us, and ends up breaking downstream plotting functions if used instead of `self.dataset.feature_names`. 

I think that the only way to simplify the code I pushed yesterday is to remove the `isinstance` checks - which I will push soon. I don't think there is a way to get around creating a `DataFrame` first, and then renaming the columns as required.

dsgibbons:
To summarise, the NaN errors have been removed, so this pull request fixes the issue raised in #6554. My most recent push is basically equivalent to yesterday but contains some simplifications to remove redundant `isinstance` checks.

The only remaining issue is that sklearn will raise warnings when trained on unlabelled columns but tested on labelled columns. There is no way around this as far as I can tell unless the `evaluate` function in `base.py` directly accesses `model.feature_names_in_`. I believe however that this should be left as a future issue/pull request, as this issue is unrelated to #6554.

mlflow/models/evaluation/base.py
```diff
@@ -365,7 +366,7 @@ def __init__(self, data, *, targets, name=None, path=None, feature_names=None):
                 self._feature_names = feature_names
             else:
                 self._features_data = data.drop(targets, axis=1, inplace=False)
-                self._feature_names = list(self._features_data.columns)
+                self._feature_names = [stringify_column(c) for c in self._features_data.columns]
```
harupy, line 369:
```suggestion
                self._feature_names = [str(c) for c in self._features_data.columns]
```

Why not just apply `str`?

dsgibbons: I was just following this suggestion: https://github.com/mlflow/mlflow/pull/6555#issuecomment-1230084981

harupy: Got it. @WeichenXu123 Why should we use `feature_{str(raw_column_name)}`?

WeichenXu123: To clarify, I didn't mean rename column name.

My previous comment say, if the dataframe column has the column name, do not change it, but if the dataframe column is unlabeled, then auto generate the column name by `feature_{index}` to make it clear. And the column name is only used in shap plot display, we should not use the column name while running prediction.

harupy: Makes sense! Thanks for the explanation.

harupy: I think `stringify_column` is confusing. It sounds like it just applies `str`, but it doesn't. Can we use a name such as `generate_feature_name_if_not_string`?

mlflow/utils/string_utils.py
```diff
@@ -14,6 +14,13 @@ def is_string_type(item):
     return isinstance(item, str)
 
 
+def stringify_column(s):
+    if isinstance(s, str):
+        return s
+
+    return f"feature_{s}"
```
WeichenXu123, line 21: I suggest to use `return f"feature_{s}_{uuid.uuid4()}"` to avoid generate duplicate feature names.

WeichenXu123:
> If the local variable x has truncated column names, then you get NaNs from pd.DataFrame(x, columns=feature_names).

@dsgibbons Thanks for pointing out this !

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -1089,8 +1094,9 @@ def X(self) -> pd.DataFrame:
         """
         The features (`X`) portion of the dataset, guarded against accidental mutations.
         """
+        df = pd.DataFrame(self.dataset.features_data)
         return DefaultEvaluator._MutationGuardedData(
-            pd.DataFrame(self.dataset.features_data, columns=self.dataset.feature_names)
+            df.rename(columns=dict(zip(df.columns, self.feature_names)))
```
WeichenXu123, line 1099: Let's add comment explain why cannot use `pd.DataFrame(self.dataset.features_data, columns=self.dataset.feature_names)`, it is a pitfall.

mlflow/models/evaluation/default_evaluator.py
```diff
@@ -411,7 +411,8 @@ def _compute_df_mode_or_mean(df):
 
 
 def _shap_predict_fn(x, predict_fn, feature_names):
-    return predict_fn(pd.DataFrame(x, columns=feature_names))
+    df = pd.DataFrame(x)
+    return predict_fn(df.rename(columns=dict(zip(df.columns, feature_names))))
```
WeichenXu123, line 415: ditto. Can we make this (change column names) to be a helper method ?

WeichenXu123:
Can we fix the warning `X has feature names, but LogisticRegression was fitted without feature names` ?

We can do it in follow-up PR.

dsgibbons:
Sounds good - I'll sort address those comments tomorrow!

mlflow/utils/string_utils.py
harupy: A pandas dataframe normally contains duplicate column names. Do we reaaly need to add uuid?

harupy: I think `uuid.uuid4().hex[:6]` confuses users. If I saw `feature_0_af82jk`, I would wonder what `af82jk` is and where it comes from.

harupy: Let me know if I'm missing something.

dsgibbons: I agree that a pandas dataframe is unlikely to include duplicate column names. If the column names are uninitialized, they will automatically be named with integers anyway. I'll leave off the uuid for now unless you really want it @WeichenXu123.

dsgibbons:
Looks like an unrelated docker issue is failing the python check

mlflow/utils/string_utils.py
WeichenXu123: OK I agree. let's keep current code

WeichenXu123 (approved):
LGTM

mlflow/models/evaluation/default_evaluator.py
harupy: `ignore_exceptions` might sound like "ignore any exceptions during the evaluation process", but that's not correct. Can we rename it?

tests/models/test_default_evaluator.py
harupy: What do you want to this in this test?

WeichenXu123:
@dsgibbons 

Would you like to file a PR to fix the warning ?

> X has feature names, but LogisticRegression was fitted without feature names 

We can save the original column name in DatasetEvaluator and when calling predict / predict_proba, we restore the original dataset column name to address the issue. :)

tests/models/test_default_evaluator.py
harupy: To test an exception, you can use `pytest.raises`.


```
with pytest.raises(<ExceptionType>, match="<error message>"):
    ...
```

mlflow/models/evaluation/default_evaluator.py
WeichenXu123: Sorry for merge the PR too quickly.
We can rename it to be `explainability_ignore_exceptions" .

dsgibbons:
> @dsgibbons
> 
> Would you like to file a PR to fix the warning ?
> 
> > X has feature names, but LogisticRegression was fitted without feature names
> 
> We can save the original column name in DatasetEvaluator and when calling predict / predict_proba, we restore the original dataset column name to address the issue. :)

Yes I can address this later in the week.

tests/models/test_evaluation.py
```diff
@@ -333,7 +336,33 @@ def iris_pandas_df_num_cols_dataset():
...
+@pytest.fixture
+def iris_pandas_df_longname_cols_dataset():
+    X, y = get_iris()
+    eval_X, eval_y = X[0::3], y[0::3]
+    data = pd.DataFrame(
+        {
+            "f1": eval_X[:, 0],
+            "f2": eval_X[:, 1],
+            "f3longnamelongnamelongname": eval_X[:, 2],
+            "f4longnamelongnamelongname": eval_X[:, 3],
```
harupy, line 354:
```suggestion
            # Column names longer than the truncation threshold
            "f3_" + "x" * 20: eval_X[:, 2],
            "f4_" + "x" * 20: eval_X[:, 3],
```

I'd use `+ "x" * 20` here. It's a bit hard to tell if `f3longnamelongnamelongname` has more than 20 characters.

tests/models/test_default_evaluator.py
WeichenXu123: This code is a helper method, I think the code means we run the test without swallowing exceptions.
And all tests should pass and do not raise exception, if raise exception, the test will fail and we can see the exception stack trace.

harupy: I think I understand what you want to do.

WeichenXu123:
@dsgibbons Also remember to update the config https://github.com/mlflow/mlflow/pull/6555#discussion_r962603389 :)

tests/models/test_default_evaluator.py
dsgibbons: `_evaluate_explainer_with_exceptions` is a helper function that runs evaluation with explainability. It is called for various dataframes by:

- df with string columns: `test_default_explainer_pandas_df_str_cols`
- df with integer columns: `test_default_explainer_pandas_df_num_cols`
- df with long-name columns:`test_default_explainer_pandas_df_longname_cols`

If NaNs are generated, an error will be raised by:

```
if not self.evaluator_config.get("ignore_exceptions", True):
    raise e
```

and so, if the above test runs without raising an exception, that is considered a success and the test passes. We don't need `pytest.raises`, as we aren't trying to purposely raise an exception.

I was more asking whether there is a way to parameterize `_evaluate_explainer_with_exceptions` in a more "pytest" way that doesn't require calling the function three times from three separate places.

mlflow/models/evaluation/default_evaluator.py
harupy: If we add this config, we should update `mlflow.evalute` docstring.

dsgibbons: Possibly, but I don't know what you would call it. Maybe `ignore_shap_explainer_exceptions`?

WeichenXu123: We have other configs like `explainability_nsamples", we'd better keep the same format and make them have the same prefix

dsgibbons: ah sorry I just saw the replies

dsgibbons:
> @dsgibbons Also remember to update the config [#6555 (comment)](https://github.com/mlflow/mlflow/pull/6555#discussion_r962603389) :)

Yes makes sense, I'll do that. I'm guessing I can just open up a fresh PR for these extra changes?

tests/models/test_default_evaluator.py
harupy: Ok, `_evaluate_explainer_with_exceptions` sounds like `mlflow.evaluate succeeds even when exceptions occur during evaluation`.

tests/models/test_default_evaluator.py
```diff
@@ -642,6 +645,45 @@ def test_svm_classifier_evaluation(svm_model_uri, breast_cancer_dataset, baselin
...
+        multiclass_logistic_regressor_model_uri, iris_pandas_df_num_cols_dataset
+    )
+
+
+def test_default_explainer_pandas_df_longname_cols(
+    multiclass_logistic_regressor_model_uri, iris_pandas_df_longname_cols_dataset
+):
+    _evaluate_explainer_with_exceptions(
+        multiclass_logistic_regressor_model_uri, iris_pandas_df_longname_cols_dataset
+    )
```
harupy, line 684: image

image

image

The SHAP plots logged in this test look like this. Is this the expect result?

harupy: Is `(f_3)` expected?

WeichenXu123: Do we made some change on shap plot rendering recently ? I remember previously I adjust the plot layout well.

harupy: This line says `For duplicated truncated name, attach "(f_{feature_index})" at the end`, but there is no duplicates in the truncated names.

https://github.com/mlflow/mlflow/blob/5fa68ff6addd439ba0f267b30cd73279379b823d/mlflow/models/evaluation/default_evaluator.py#L566

harupy: > Do we made some change on shap plot rendering recently ? I remember previously I adjusted the plot layout well.

I don't know.

WeichenXu123: > Is (f_3) expected?

Yes. After name being truncated, conflicts might happen, e.g., 

`longlonglonglongabcdefghijklmnlonglonglonglong` and `longlonglonglongxyzxyzxyzxyzxlonglonglonglong`
after truncation, they both becomes `longlonglonglong...longlonglonglong`

> Looks like the layout has some issue when feature name is long

I will try to fix it.

harupy: @WeichenXu123 Why do we need to add the suffix when there is no duplicate?

WeichenXu123: @harupy 
Truncation fixing PR : https://github.com/mlflow/mlflow/pull/6705

WeichenXu123: shap_beeswarm_plot_on_data_diabetes_dataset
shap_feature_importance_plot_on_data_diabetes_dataset
shap_summary_plot_on_data_diabetes_dataset

In my test plot rendering works well.
@harupy Could you provide your testing code ?

harupy: @WeichenXu123 You can run `test_default_explainer_pandas_df_longname_cols`.

PR #6567, opened 2022-08-24
Feature/csv data preview

### LINKED ISSUE #6439: [FR] Format CSV output as a table in ArtifactViewer
### Willingness to contribute

Yes. I can contribute this feature independently.

### Proposal Summary

If a logged artifact is a CSV file, we try parsing the file and laying out a slice of the data in an embedded table in the artifact viewer.

### Motivation

> #### What is the use case for this feature?
To have a more aesthetically pleasing experience viewing tabular data in MLflow

> #### Why is this use case valuable to support for MLflow users in general?
As users rely more on the artifact viewer as a part of their mlops experience, this could create less time switching to another app to show a slice of training data or testing results.

> #### Why is this use case valuable to support for your project(s) or organization?
We don't shy away from showing our business partners our experiment runs. Being able to preview CSV files (especially training data) would be a big help when we present our work.

> #### Why is it currently difficult to achieve this use case?
* You would need to utilize a CSV parser dependency to be able to lay out the data preview

### Details

* If a CSV file cannot be parsed, fallback to the plaintext preview of the file with a banner saying it was unable to be parsed
* To limit performance issues in the browser, maybe we limit the CSV preview to the first `N` rows
* We can reuse the CSS in the experiments table in the experiments page
* Utilize an alternating cell background color to ease the eyes
* Important that we don't turn this feature into a tabular data viewer and gently encourage the user to download their CSV file to view in their spreadsheet editor of choice by:
1. Not allowing a sort feature on the column header, and 
2. Not allowing the user to re-order columns

See below for current way CSVs are visualized in MLflow
Screen Shot 2022-08-10 at 8 02 30 PM

See below on implementations from similar tools

Snowflake

snowflake_table_example

Dataiku

dataiku_table_example

GitHub

github

### What component(s) does this bug affect?

### What interface(s) does this bug affect?

- [X] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server

### What language(s) does this bug affect?

### What integration(s) does this bug affect?

nnethery:
## Related Issues/PRs

Close #6439

Add a table visualization to CSV artifacts in the artifacts viewer in a specific experiment's run page.

## How is this patch tested?

A script to log data for testing:

```python
import tempfile
import requests
import zipfile
import pandas as pd
import mlflow
from pathlib import Path


def unzip(path, dst):
    with zipfile.ZipFile(path, "r") as zip_ref:
        zip_ref.extractall(dst)


def download_url(url, save_path, chunk_size=128):
    r = requests.get(url, stream=True)
    with open(save_path, "wb") as fd:
        for chunk in r.iter_content(chunk_size=chunk_size):
            fd.write(chunk)


with tempfile.TemporaryDirectory() as tmpdir:
    tmpdir = Path(tmpdir)
    url = "https://github.com/mlflow/mlflow/files/9454188/Airbnb_Open_Data.csv.zip"
    name = url.split("/")[-1]
    zip_path = tmpdir.joinpath(name)
    download_url(url, zip_path)
    dst = tmpdir.joinpath("unzipped")
    unzip(zip_path, dst)
    csv_file = next(dst.glob("*.csv"))
    print(pd.read_csv(csv_file).head(1000))

    with mlflow.start_run():
        mlflow.log_artifact(csv_file)
```

CSV file artifacts that are logged in experiments are now visualized with a table in the artifact viewer. This table visualizes a slice of the data up to the first 500 rows and the slice is sortable via the column headers.

Components

Interface

Language

Integrations


nnethery:
Hello maintainers, I would really like some help on this pull request on the following aspects:

- The header row and its column separators jitter when the data is loaded into the grid
- I would really like the header row to be sticky as is the normal convention of `ag-grid` but I can't figure it out. I suspect when I set `overflow: scroll` on the grid CSS overrides it is messing something up.
- When the number of columns is less than the viewport, it does not fill the entire width of the viewport (test this with a 3 or 4 column CSV with only integers as the row values)

nnethery:
I see that tab separated value files are also still common so I will add support for this in the next commit

dbczumar:
> Hello maintainers, I would really like some help on this pull request on the following aspects:
> 
> * The header row and its column separators jitter when the data is loaded into the grid
> * I would really like the header row to be sticky as is the normal convention of `ag-grid` but I can't figure it out. I suspect when I set `overflow: scroll` on the grid CSS overrides it is messing something up.
> * When the number of columns is less than the viewport, it does not fill the entire width of the viewport (test this with a 3 or 4 column CSV with only integers as the row values)
> * The redrawing of data in the grid on a window resize is very slow

@harupy @sunishsheth2009 Can you review this PR and help to address these points from @nnethery?

harupy:
@nnethery Is there a reason we need to use `AgGrid`? Can we use the `Table` component in `@databricks/design-system`? Here's a quick demo using the `Table` component:


https://user-images.githubusercontent.com/17039389/187144219-b28ac13d-0526-4543-94a7-7571631e74ef.mov

nnethery:
@harupy I tried a bigger CSV that has long strings as values and column headers where the names are longer than the values they contain. I attached it for your own testing and you can pull the latest commit to try. I apologize that I don't know the style overrides available in the custom `Table` component. Any suggestions?

Screen Shot 2022-08-30 at 3 54 01 PM
Screen Shot 2022-08-30 at 4 04 01 PM
[Airbnb_Open_Data.csv.zip](https://github.com/mlflow/mlflow/files/9454188/Airbnb_Open_Data.csv.zip)

harupy:
@nnethery Setting the column props might help: https://ant.design/components/table/#Column.

nnethery:
> @nnethery Setting the column props might help: https://ant.design/components/table/#Column.
> 
> `databricks/design-system` was created based on `antd` so (not all) props available in `antd` should be available in `databricks/design-system` as well.

@harupy Ok I updated the branch. The best solution I could figure out with `antd` is just hard-coding a column width and adding ellipsis if that isn't enough room for the header. It's a little too much space if the header is long but column values are short, but I think it looks good enough to ship.

harupy:
@nnethery Thanks for the udpate! Can you take a screenshot and share it?

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,136 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { getSrc } from './ShowArtifactPage';
+import { getArtifactContent } from '../../../common/utils/ArtifactUtils';
+import { Table } from '@databricks/design-system';
+import Papa from 'papaparse';
+
+class ShowArtifactTableView extends Component {
```
harupy, line 8: Can we convert this to a functional component?

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /** Fetches artifacts and updates component state with the result */
+  fetchArtifacts() {
+    const artifactLocation = getSrc(this.props.path, this.props.runUuid);
+    this.props
+      .getArtifact(artifactLocation)
+      .then((text) => {
+        try {
+          const result = Papa.parse(text, { header: true });
```
harupy, line 107: Is it possible to parse only the first `MAX_ROW_LENGTH` rows?

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,136 @@
...
+      return (
+        <div className='artifact-text-view-error'>
+          Oops we couldn't load your file because of an error.
+        </div>
+      );
+    }
+
+    if (this.state.data) {
+      // eslint-disable-next-line max-len
+      const rowPreviewMessage = `Previewing the first ${this.state.data.length} rows out of ${this.state.originalRowLength}`;
```
harupy, line 60: Can we use `FormattedMessage` for internationalization?

Example:

https://github.com/mlflow/mlflow/blob/4c0f21025322f4681e5c4cf01d2ad57e0eb93115/mlflow/server/js/src/experiment-tracking/components/MetricView.js#L28-L35

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,136 @@
...
+        },
+        width: 200,
+        ellipsis: {
+          showTitle: true,
+        },
+      }));
+
+      return (
+        <div css={{ overscrollBehaviorX: 'contain', overflowX: 'scroll', margin: 10 }}>
+          <span style={{ display: 'flex', justifyContent: 'center' }}>{rowPreviewMessage}</span>
```
harupy, line 81:
```suggestion
          <span css={{ display: 'flex', justifyContent: 'center' }}>{rowPreviewMessage}</span>
```

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
nnethery: This has been resolved by using a `preview` config option in Papaparse which streams each row and aborts the serialization of the rows once the row limit is met. You should notice a performance boost. 

Perhaps for a second iteration of this feature, we can avoid downloading the whole CSV and stream the file’s bytes and parse the rows as we go.

nnethery:
> @nnethery Thanks for the udpate! Can you take a screenshot and share it?

Screen Shot 2022-09-06 at 9 51 22 AM

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,132 @@
...
+import Papa from 'papaparse';
+
+const ShowArtifactTableView = ({ runUuid, path, getArtifact }) => {
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState();
+  const [data, setData] = useState();
+  const [headers, setHeaders] = useState();
+  const [text, setText] = useState();
+
+  const MAX_ROW_LENGTH = 500;
```
harupy, line 16:
```suggestion
  const MAX_NUM_ROWS = 500;
```

`MAX_ROW_LENGTH` sounds like the maximum number of elements in a row. What do you think?

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.test.js
```diff
@@ -0,0 +1,23 @@
...
+      runUuid: 'fakeUuid',
+    };
+    commonProps = { ...minimalProps };
+    wrapper = shallow(<ShowArtifactTableView {...commonProps} />);
+  });
+
+  test('should render with minimal props without exploding', () => {
+    wrapper = shallow(<ShowArtifactTableView {...minimalProps} />);
+    expect(wrapper.length).toBe(1);
+  });
```
harupy, line 22: Can we test the following behaviors?

1. Render CSV /TSV file fine.
2. When the number of rows is > 500, only show the first 500 rows.
3. Render the raw text if it's not valid CSV.

harupy:
https://user-images.githubusercontent.com/17039389/188789900-748f49b7-5767-4151-b176-400c42307f7d.mov

(for other reviewers)

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactPage.js
```diff
@@ -15,6 +16,7 @@ import ShowArtifactTextView from './ShowArtifactTextView';
 import ShowArtifactMapView from './ShowArtifactMapView';
 import ShowArtifactHtmlView from './ShowArtifactHtmlView';
 import ShowArtifactPdfView from './ShowArtifactPdfView';
+import ShowArtifactTableView from './ShowArtifactTableView';
```
hubertzub-db, line 19: Is it possible to lazily import this component here (using `React.lazy`)? This should end up having `papaparse` dependency in the separate chunk, keeping the main bundle clean and thin ✨

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,132 @@
...
+    );
+  }
+
+  if (data) {
+    const columns = headers.map((f) => ({
+      title: f,
+      dataIndex: f,
+      key: f,
+      sorter: (a, b) => {
+        if (isNaN(a)) {
```
harupy, line 81: Why are we checking whether `a` is NaN here?

nnethery: I add a check if a cell value is not a number and run a string compare function for the sorter. `papaparse` casts parsed data to their most likely data type so I'm just accounting for numbers and strings in the sorter.

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.test.js
nnethery: I'm having some issues handling test case 2, can you please review the last test in `ShowArtifactTableView.test.js` in the latest commit and help me understand why `wrapper` is an empty array?

harupy: `mount` needs to be replaced with `mountWithIntl` since `ShowArtifactTableView` contains `FormattedMessage`.

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
nnethery: After double-checking, you are correct. Everything comes in as a string, and `null`/`undefined` becomes an empty string. I updated the sorter accordingly.

harupy (approved):
LGTM!

harupy:
@hubertzub-db Could you take another look?

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,124 @@
...
+import { getSrc } from './ShowArtifactPage';
+import { getArtifactContent } from '../../../common/utils/ArtifactUtils';
+import { Table } from '@databricks/design-system';
+import { FormattedMessage } from 'react-intl';
+import Papa from 'papaparse';
+
+const ShowArtifactTableView = ({ runUuid, path, getArtifact }) => {
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState();
+  const [data, setData] = useState();
```
sunishsheth2009, line 12: I am curious if we could use the [RequestStateWrapper](https://github.com/mlflow/mlflow/blob/4c0f21025322f4681e5c4cf01d2ad57e0eb93115/mlflow/server/js/src/common/components/RequestStateWrapper.js) here rather than dealing with loading and error states. I think we can let this component handle that for us. What do you think?

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
```diff
@@ -0,0 +1,124 @@
...
+  const [headers, setHeaders] = useState();
+  const [text, setText] = useState();
+
+  const MAX_NUM_ROWS = 500;
+
+  useEffect(() => {
+    resetState();
+    fetchArtifacts({ path, runUuid, getArtifact });
+  }, [runUuid, path]);
```
hubertzub-db, line 21: there's a missing dependency here - please add it or add a eslint ignore directive

hubertzub-db (approved):
Looks good! Just left a comment re eslint warning and +1 to @sunishsheth2009 's comment: https://github.com/mlflow/mlflow/pull/6567#discussion_r966639800

mlflow/server/js/src/experiment-tracking/components/artifact-view-components/ShowArtifactTableView.js
nnethery: I will take a crack at this before merging!

nnethery: @sunishsheth2009 Is this just for network requests? There are two main errors that could occur with this component: 
1. A call to download the artifact from the artifact server which has its own function wrapped around the API call
2. An error parsing the CSV file in the browser which does not call to the server

Is this wrapper still suitable for these errors and loading?

sunishsheth2009: I think if we want to use [RequestStateWrapper](https://github.com/mlflow/mlflow/blob/4c0f21025322f4681e5c4cf01d2ad57e0eb93115/mlflow/server/js/src/common/components/RequestStateWrapper.js) we need to use redux pattern to use request_ids to work correctly. If we feel that we don't want to use redux here then I think we are fine leaving this approach as it is.
cc @harupy @hubertzub-db thoughts?

hubertzub-db: I think it's fine to leave it as it is for now - looking at the forthcoming plans for improving the API connectivity code and restructuring the data store, there will be an opportunity to unify it anyway.

nnethery: Sounds good, all. That’s all the changes left on my end!

harupy:
@nnethery Thanks for the contribution!

@sunishsheth2009 @hubertzub-db Thanks for the review :)

PR #6575, opened 2022-08-24
Update scoring api.

tomasatdatabricks:
This PR makes the following changes to the scoring API:

headers: No longer recognize format flags used in mlflow 1.0 to specify pandas encoding (split or records or numpy split). 
Input data: 
   Input is a dictionary with exactly one of dataframe_split, dataframe_records, instances, inputs fields. 
Output data:
  Output is a dictionary with predictions field that contains jsonized model output. The output itself remains unchanged. 

Implementation wise, we can no longer use pandas.read_json function since the json is nested. Instead we parse the json using the built-in json library and turn the input into pandas DataFrame. This means that the type hints are not applied until the data is parsed and we have to explicitly cast the types. 

Signed-off-by: tomasatdatabricks <tomas.nykodym@databricks.com>


mlflow/pyfunc/scoring_server/client.py
```diff
@@ -1,9 +1,32 @@
...
 from mlflow.pyfunc import scoring_server
-from mlflow.utils.proto_json_utils import _DateTimeEncoder
+from mlflow.utils.proto_json_utils import (
+    dataframe_from_parsed_json,
+    parse_tf_serving_input,
+)
+from mlflow.exceptions import MlflowException
+
+
+def infer_and_parse_json_output(decoded_input):
```
tomasatdatabricks, line 15: NOTE: I am not sure what the client is for and what is the intended API. I tried to preserve the existing behavior as closely as possible, but maybe we should change it?

mlflow/pyfunc/__init__.py
```diff
@@ -1200,7 +1209,7 @@ def udf(
             pyfunc_backend = _get_flavor_backend(
                 local_model_path_on_executor,
                 workers=1,
-                install_mlflow=False,
+                install_mlflow=(mlflow_home is None),
```
dbczumar, line 1212: Previously, we would never install MLflow. Now, it seems like we'll almost always install MLflow, since users don't typically set `mlflow_home`. Is this necessary? cc @WeichenXu123

mlflow/pyfunc/__init__.py
```diff
@@ -1024,6 +1030,9 @@ def spark_udf(spark, model_uri, result_type="double", env_manager=_EnvManager.VI
     from pyspark.sql.types import DoubleType, IntegerType, FloatType, LongType, StringType
     from mlflow.models.cli import _get_flavor_backend
 
+    # Used in test to force install lcoal version of mlflow when starting a model server
+    mlflow_home = os.environ.get("MLFLOW_HOME")
```
dbczumar, line 1034: The impact of this appears to extend beyond tests: https://github.com/mlflow/mlflow/pull/6575/files#r959175274.

```suggestion
    # Used in test to force install local version of mlflow when starting a model server
    mlflow_home = os.environ.get("MLFLOW_HOME")
```

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -76,60 +74,45 @@ def infer_and_parse_json_input(json_input, schema: Schema = None):
...
+                message=(
+                    "Failed to parse input from JSON. Ensure that input is a valid JSON"
+                    f" formatted string. Error: '{ex}'"
+                ),
+                error_code=BAD_REQUEST,
+            )
+    if isinstance(decoded_input, dict):
+        format_keys = set(decoded_input.keys()).intersection(SUPPORTED_FORMATS)
+        if len(format_keys) != 1:
+            message = f"Received dictionary with input fields: {format_keys}"
```
dbczumar, line 93: The error message doesn't make it obvious that the format is unsupported or that multiple formats are specified. Can we separate these two conditions into separate error messages that more clearly state what the issue is?

dbczumar: Nevermind, missed `REQUIRED_INPUT_FORMAT` :)

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -76,60 +74,45 @@ def infer_and_parse_json_input(json_input, schema: Schema = None):
...
+            raise MlflowException(
+                message=f"{REQUIRED_INPUT_FORMAT}. {message}.", error_code=BAD_REQUEST
```
dbczumar, line 95: Nit: `REQUIRED_INPUT_FORMAT` already contains a period. Messages containing `{REQUIRED_INPUT_FORMAT}.` will contain two periods - `..`

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -258,66 +213,49 @@ def transformation():  # pylint: disable=unused-variable
...
+        unexpected_content_parameters = set(parameter_values.keys()).difference(set(["charset"]))
+        if unexpected_content_parameters:
+            return flask.Response(
+                response=(
+                    f"Unrecognized content type parameters: "
+                    f"{', '.join(unexpected_content_parameters)}."
+                ),
+                status=415,
+                mimetype="text/plain",
+            )
```
dbczumar, line 225: Is the motivation for throwing on unrecognized type parameters to make it clear to users of 1.x workflows that the `format=pandas-records`, etc parameters are no longer supported?

Just making sure this is safe for existing use cases, i.e. is it possible that some users / orgs might want to pass custom content type parameters so that some upstream layer can do request logging / auditing or something?

mlflow/sagemaker/__init__.py
```diff
@@ -2597,14 +2598,17 @@ def predict(self, deployment_name=None, df=None, endpoint=None):
...
             response = sage_client.invoke_endpoint(
                 EndpointName=deployment_name,
-                Body=json.dumps(_get_jsonable_obj(df, pandas_orient="split")),
+                Body=body,
                 ContentType="application/json",
             )
-
             response_body = response["Body"].read().decode("utf-8")
-            return infer_and_parse_json_input(response_body)
+            return ScoringServerResponse.from_raw_json(response_body)
```
dbczumar, line 2611: This breaks the API contract of `BaseDeploymenntClient.predict()`: https://mlflow.org/docs/latest/python_api/mlflow.deployments.html#mlflow.deployments.BaseDeploymentClient.predict, which states that the API returns a pandas DataFrame, series, or numpy array (i.e pyfunc output, though list should also be added to the docs)

The goal of the original implementation was to convert the response JSON into the corresponding output object. Can we preserve that?

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -1,9 +1,35 @@
...
+            raise MlflowException(
+                "Invalid predictions data. Prediction object must be a dictionary "
+                "with 'predictions' field."
+            )
+        return ScoringServerResponse(parsed_response)
+
+    def get_predictions_dataframe(self):
+        """Get the predictions returned from the server as pandas.DataFrame. this method will fail
+        if the returned predictions is not a valid"""
+        return pd.DataFrame(data=self["predictions"])
```
dbczumar, line 32: Following from https://github.com/mlflow/mlflow/pull/6575/files#r959181856, is this class necessary? It doesn't support the full set of pyfunc outputs

mlflow/utils/proto_json_utils.py
```diff
@@ -193,50 +196,125 @@ def default(self, o):  # pylint: disable=E0202
...
 
 
-def _dataframe_from_json(
-    path_or_str, schema=None, pandas_orient: str = "split", precise_float=False
-):
+class MlflowFailedTypeConversion(MlflowException):
+    def __init__(self, col_name, col_type, ex):
+        super().__init__(
+            message=f"Data is not compatible with model signature. "
+            f"Failed to convert columns {col_name} to type '{col_type}'. Error: '{ex}'",
```
dbczumar, line 203: Nit:

```suggestion
            f"Failed to convert column {col_name} to type '{col_type}'. Error: '{ex}'",
```

mlflow/utils/proto_json_utils.py
```diff
@@ -193,50 +196,125 @@ def default(self, o):  # pylint: disable=E0202
...
-        return df
+                typemessage = f"type {type(decoded_input)}"
+            raise MlflowBadScoringInputException(
+                f"Dataframe records format must be a list of records. Got {typemessage}."
+            )
+        try:
+            pdf = pd.DataFrame(data=decoded_input)
+        except Exception as ex:
+            raise MlflowBadScoringInputException(
+                f"Provided dataframe_records field is not a valid dataframe representation. "
```
dbczumar, line 262:
```suggestion
                f"Provided dataframe_records field is not a valid Pandas dataframe representation in 'records' format. "
```

mlflow/utils/proto_json_utils.py
```diff
@@ -193,50 +196,125 @@ def default(self, o):  # pylint: disable=E0202
...
+                f"Error: '{ex}'"
+            )
+    elif pandas_orient == "split":
+        if not isinstance(decoded_input, dict):
+            if isinstance(decoded_input, list):
+                typemessage = "list"
+            else:
+                typemessage = f"type {type(decoded_input)}"
+            raise MlflowBadScoringInputException(
+                f"Dataframe split format must be a dictionary with at least 'data' field. "
```
dbczumar, line 272:
```suggestion
                f"Dataframe split format must be a dictionary containing a 'data' field. "
```

mlflow/utils/proto_json_utils.py
```diff
@@ -193,50 +196,125 @@ def default(self, o):  # pylint: disable=E0202
...
+            )
+        try:
+            pdf = pd.DataFrame(
+                index=decoded_input.get("index"),
+                columns=decoded_input.get("columns"),
+                data=decoded_input["data"],
+            )
+        except Exception as ex:
+            raise MlflowBadScoringInputException(
+                f"Provided dataframe_split field is not a valid dataframe representation. "
```
dbczumar, line 291:
```suggestion
                f"Provided dataframe_split field is not a valid dataframe representation in 'split' format. "
```

mlflow/pyfunc/scoring_server/client.py
dbczumar: when `mlflow.pyfunc.spark_udf()` is called with `env_manager="virtualenv"` or `env_manager="conda"`, MLflow starts a model server and uses this client to send requests to it in order to perform inference; this is the long-awaited "restore model environments in spark UDF" feature that we talked about ~ 3 years ago :). It isn't a public API, so changes are okay as long as Spark UDF continues to work.

cc @WeichenXu123

dbczumar (commented):
@tomasatdatabricks Awesome stuff! Left some small comments and questions. Do we have time allocated to update our docs for this, such as https://mlflow.org/docs/latest/models.html#deploy-mlflow-models? Looks like a lot of this content would be out of date after merging this PR.

Can we add docs updates to the current PR or file a separate PR that we can merge at the same time as this one?

mlflow/pyfunc/__init__.py
tomasatdatabricks: Oh, nice catch, it's inverted. I want to install mlflow only when MLFLOW_HOME is set. MLFLOW_HOME (not very descriptive name in retrospect) is meant for development - so that you can test your local changes when running stuff in a conda environment. For example, in this case, the tests were failing because the conda environment was installing mllfow from pip and hence was expecting the old input.

mlflow/pyfunc/__init__.py
tomasatdatabricks: commented above. hopefully it explains it. we can also rename MLFLOW_HOME to something like `MLFLOW_DEV_PROJECT_HOME` in a followup

mlflow/pyfunc/scoring_server/__init__.py
tomasatdatabricks: nice catch.

mlflow/pyfunc/scoring_server/__init__.py
tomasatdatabricks: Yes that's exactly the intent - so we don't silently ignore stuff. Good question about the upstream - I would say we can start with being restrictive and loosen it in the future if we have to?

mlflow/sagemaker/__init__.py
tomasatdatabricks: Ah, this is a hard one. We could certainly preserve the behavior but I am not sure if we should because:
1. we wrap the response in a dictionary so we can return additional stuff. I think we should enable access to any additional output via the client interface.
2. It's actually impossible to know whether the response should be pandas dataframe or series or numpy array. I saw most clients will always return pandas dataframe. infer_and_parse_json returns a dataframe unless the input is explicitly a tensor input - i.e. a dict with `instances` or `inputs` field. Unfortunately, our current server does not make that distinction and just jsonizes whatever the model returned. It is a bit late to change that now unfortunately. 

Therefore, I would propose we update the interface to return pyfunc reponse object that can access the prediction as dataframe or numpy array per user's request and will have access to all response fields. Wdyt? Is that too big a change too late? We could also preserve the original behavior by default and add a flag to use the new behavior.

mlflow/pyfunc/scoring_server/client.py
tomasatdatabricks: Responded above.

tomasatdatabricks:
Thanks for the review @dbczumar ! I am planning to update the docs, but I am still fixing tests :/ The PR grew unexpectedly in scope as I am working on it cause I see it impacts more things. So I was hoping to get all tests green before I bring stuff up, get agreement and update the docs in the end. I will be working on this full time next sprint - it's my last sprint before I go on parental leave so I gotta finish it.

mlflow/pyfunc/scoring_server/client.py
WeichenXu123: Yes. The `scoring_server.client` module is used in `spark_udf` with env_manager="virtualenv" or env_manager="conda".

mlflow/pyfunc/__init__.py
WeichenXu123: yes I think only in testing case we need set `install_mlflow` to be true

examples/spark_udf/spark_udf_datetime.py
```diff
@@ -63,4 +63,10 @@ def main():
 
 
 if __name__ == "__main__":
+    import os
+
+    assert os.environ.get("MLFLOW_HOME") is not None
+    from mlflow.version import VERSION
+
+    assert "dev" in VERSION
```
dbczumar, line 71: Reminder to remove this test code (I'll go ahead and do this :) )

mlflow/pyfunc/scoring_server/__init__.py
dbczumar: @dbczumar Reminder to address this nit of a comment :)

dbczumar: Nevermind, already seems to be addressed

mlflow/utils/environment.py
```diff
@@ -526,3 +529,55 @@ def _get_pip_install_mlflow():
...
+                env=command_env,
+                capture_output=capture_output,
+                synchronous=synchronous,
+                preexec_fn=preexec_fn,
+                close_fds=True,
+                stdout=stdout,
+                stderr=stderr,
+            )
+        except ShellCommandException as ex:
+            raise ex
```
dbczumar, line 583: Nit: Unnecessary catch and reraise

mlflow/pyfunc/scoring_server/client.py
```diff
@@ -1,11 +1,43 @@
...
+
+    def get_predictions_dataframe(self):
+        """Get the predictions returned from the server as pandas.DataFrame. this method will fail
+        if the returned predictions is not a valid DataFrame representation."""
+        return pd.DataFrame(data=self["predictions"])
+
+    def get_predictions_nparray(self, dtype=None):
+        """Get the predictions returned from the server as a numpy array. this method will fail
+        if the returned predictions is not a valid numpy array"""
+        return np.array(self["predictions"], dtype)
```
dbczumar, line 38: Let's consolidate this into a single method:

```
def get_predictions(format="dataframe", dtype=None)
```

dbczumar (commented):
@dbczumar As part of this PR, we'll need to update `BaseDeploymentClient`'s predict method to return a `MlflowModelServerOutput` object (which we may want to rename slightly - will brainstorm)

mlflow/utils/virtualenv.py
```diff
@@ -221,7 +221,7 @@ def _get_virtualenv_name(python_env, work_dir_path, env_id=None):
 
 
 def _create_virtualenv(
-    local_model_path, python_bin_path, env_dir, python_env, extra_env=None, capture_output=False
+    local_model_path, python_bin_path, env_dir, python_env, extra_env=None, capture_output=True
```
dbczumar, line 224: Unclear why the `capture_output` default value needs to be changed to `True`. @dbczumar to investigate

tomasatdatabricks: oh I probably  used it to debug stuff and forgot to switch it back!

examples/spark_udf/spark_udf_datetime.py
dbczumar: Fixed!

mlflow/pyfunc/scoring_server/client.py
dbczumar: Done!

dbczumar (commented):
Addressed some comments. Working on deployment client incorporation.

dbczumar:
@mlflow-automation autoformat

dbczumar (approved):
LGTM! Thanks @tomasatdatabricks!

dbczumar:
Test failures are caused by sqlparse upgrade and unrelated to this PR. Merging!

PR #6586, opened 2022-08-25
Wheeled Model

arjundc-db:

## Related Issues/PRs

#xxx

Add functionality to create a `wheeled` model where the dependencies are stored as wheels along with the model. Credit goes to [anirudhachal-db](https://github.com/anirudhachal-db) who started this https://github.com/mlflow/mlflow/pull/6416
How to use:

- With the helper function: 
```
import mlflow.models.utils
mlflow.models.utils.add_libraries_to_model(<model_uri>)
```
- Standalone
```
from mlflow.models.wheeled_model import WheeledModel
WheeledModel.log_model(<model_uri>, <optional_registered_model_name>)
```

## How is this patch tested?
Unit tests have been added.
I also manually tested this functionality.

- [ x] I have written tests (not required for typo or doc fix) and confirmed the proposed feature/bug-fix/change works.

- [x ] No. You can skip the rest of this section.

Add functionality to create a `wheeled` model where the dependencies are stored as wheels along with the model.

Components
- [ x] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry

Interface

Language

Integrations


mlflow/models/wheeled_model.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        Downloads all the wheels of the dependencies specified in the requirements.txt file
+        :param dst_path: Path to the directory where the wheels are to be downloaded
+        :param pip_requirements_path: Path to requirements.txt in the model directory
+        """
+        if not os.path.exists(dst_path):
+            os.makedirs(dst_path)
+
+        download_command = (
+            f"python -m pip wheel --only-binary=:all: --wheel-dir={dst_path} -r"
```
arjundc-db, line 138: I wonder if there is a better mechanism?

mlflow/models/wheeled_model.py
```diff
@@ -0,0 +1,158 @@
...
+            raise Exception("Can not add wheels for model with no 'requirements.txt'.")
+
+        self._download_wheels(dst_path=wheels_dir, pip_requirements_path=pip_requirements_path)
+
+        # Update requirements.txt with wheels
+        pip_wheels = self._overwrite_pip_requirements_with_wheels(
+            wheels_dir=wheels_dir, pip_requirements_path=pip_requirements_path
+        )
+
+        _mlflow_conda_env(path=conda_env_path, additional_pip_deps=pip_wheels, install_mlflow=False)
```
tomasatdatabricks, line 97: Would this overwrite the whole conda file? I think want to change only the pip requirements section - i.e. we want to preserve channels and python version and any conda packages.

mlflow/models/wheeled_model.py
```diff
@@ -0,0 +1,158 @@
...
+
+        model_metadata = Model.load(model_metadata_path)
+
+        # Check if the model file has `wheels` set to True
+        if model_metadata.__dict__.get(_WHEELS_FOLDER_NAME, None):
+            raise MlflowException("Cannot add wheels to a wheeled model", BAD_REQUEST)
+
+        conda_env = model_metadata.flavors.get(FLAVOR_NAME, {}).get(ENV, None)
+        conda_env_path = os.path.join(local_model_path, conda_env)
+        if conda_env is None:
```
tomasatdatabricks, line 83: I think we can let the model to have only one of the conda / requirements. Do we need both?

mlflow/models/utils.py
```diff
@@ -264,3 +263,17 @@ def plot_lines(data_series, xlabel, ylabel, legend_loc=None, line_kwargs=None):
     ax.set(xlabel=xlabel, ylabel=ylabel)
 
     return fig, ax
+
+
+def create_wheeled_model(artifact_path, model_uri):
```
tomasatdatabricks, line 268: Do we need this? Looks like it's just an alias to `WheeledModel.log_model`, I'd rather have only one top level API.

mlflow/models/wheeled_model.py
tomasatdatabricks, line 138: why the --only-binary flag?

mlflow/models/utils.py
dbczumar: I'd rather have this API and keep the other one internal. Adding a wheel to a model is a utility; I don't think we need to expose a new class with its own API surface to accomplish it

dbczumar: Can we call this `add_libraries_to_model`? Referring to "wheeled models" creates a whole new "wheeled model" concept that users shouldn't need to reason about. Libraries instead of wheels gives us flexibility to add other input types in the future (eggs, etc)

arjundc-db: Yea, I added this as a utility since we agreed to avoid having the user learn a new concept. The alias makes it easier to just add wheels. 
We also give the user the ability to use the internal API if required.

For now I will only change this method name to `add_libraries_to_model`. The rest will remain as is.
@tomasatdatabricks  and @dbczumar do you have thoughts on any other name like:

- include_model_dependencies 
- include_model_libraries
- store_model_dependencies
- store_model_libraries
- download_model_dependencies
- download_model_libraries

mlflow/models/wheeled_model.py
arjundc-db: I thought every mlflow model will contain both by default.
Let's focus on the requirements.txt.

mlflow/models/utils.py
arjundc-db: Or `install_x`

mlflow/models/wheeled_model.py
arjundc-db: This was to prevent downloading source packages. I guess that wont be as big of a problem as I am imagine it, since sdist can be harder to work with?

mlflow/models/wheeled_model.py
```diff
@@ -0,0 +1,158 @@
...
+            raise MlflowException(
+                "Can not add wheels for model with no conda environment.", BAD_REQUEST
+            )
+        if not os.path.isfile(pip_requirements_path):
+            raise Exception("Can not add wheels for model with no 'requirements.txt'.")
```
dbczumar, line 88: All models have both of these specifications. If conda is missing, we should still allow the user to add models, since we're discouraging use of conda in MLflow 2.0. Similarly, pip requirements are missing from old models logged before we introduced requirements.txt support. Accordingly, can we throw iff their is neither a requirements.txt file *nor* a conda.yaml file?

Can we also add the wheels to conda.yaml and requirements.txt, if both files are present?

tests/models/test_wheeled_model.py
```diff
@@ -0,0 +1,246 @@
...
+    validate_updated_model_file(original_model_config, wheeled_model_config)
+    # Assert correct run_id
+    assert wheeled_model_config["run_id"] == wheeled_model_run_id
+
+    validate_updated_conda_dependencies(model_path, wheeled_model_path)
+
+    validate_wheeled_dependencies(wheeled_model_path)
+
+
+def test_model_save_load(tmp_path, sklearn_knn_model):
```
dbczumar, line 162: Testing inference with a wheel model would also be very useful :D

mlflow/models/utils.py
```diff
@@ -264,3 +263,17 @@ def plot_lines(data_series, xlabel, ylabel, legend_loc=None, line_kwargs=None):
...
+    """
+    Given a registered model, this helper will re-log the model along with the wheels of all the
+    the model dependencies (stored along with the model artifacts).
+
+    :param artifact_path: Run-relative artifact path.
+    :param model_uri: registered model uri of the form
+                        models:/<model_name>/<model_version/stage/latest>
```
dbczumar, line 275: Can we include a Python code example demonstrating how this API can be used to:

1. Add libraries to an existing model stored in MLflow Tracking
2. Re-log a registered model version with extra libraries

?

mlflow/models/utils.py
```diff
@@ -264,3 +263,17 @@ def plot_lines(data_series, xlabel, ylabel, legend_loc=None, line_kwargs=None):
     ax.set(xlabel=xlabel, ylabel=ylabel)
 
     return fig, ax
+
+
+def create_wheeled_model(artifact_path, model_uri):
+    """
+    Given a registered model, this helper will re-log the model along with the wheels of all the
+    the model dependencies (stored along with the model artifacts).
```
dbczumar, line 271: This function accepts arbitrary model URIs, not just registered model URIs. I.e. it can be used to add library dependencies to a model in MLflow Tracking.

The current method description doesn't really make it clear what this method does precisely. Can we describe its behavior precisely and then explain the use cases it satisfies?

mlflow/models/utils.py
dbczumar, line 268: Nit: I think we should swap these arguments, since `model_uri` is an input location and `artifact_path` is an output location

mlflow/models/utils.py
```diff
@@ -264,3 +263,17 @@ def plot_lines(data_series, xlabel, ylabel, legend_loc=None, line_kwargs=None):
...
+
+    :param artifact_path: Run-relative artifact path.
```
dbczumar, line 273: Without additional context, this parameter docstring is unclear: "Why do I have to reason about a run-relative artifact path? I thought I was adding libraries to a model version in the MLflow model registry?" 

As I understand it, this API could be used to add dependencies to an existing model in MLflow Tracking without re-logging, right? In that case, would we be able to make `artifact_path` optional and infer it from the URI if the URI is ` runs:/` URI?

mlflow/models/utils.py
dbczumar, line 268: Currently, `mlflow.models.utils` doesn't show up in the MLflow docs. Can we import it / add it here in the `mlflow.models` module, which is documented? https://github.com/mlflow/mlflow/blob/8098785831a3414bed62ec3239b60a6a500f4dbd/mlflow/models/__init__.py#L43

mlflow/models/wheeled_model.py
```diff
@@ -0,0 +1,158 @@
...
+        Downloads all the wheels of the dependencies specified in the requirements.txt file
+        :param dst_path: Path to the directory where the wheels are to be downloaded
+        :param pip_requirements_path: Path to requirements.txt in the model directory
+        """
+        if not os.path.exists(dst_path):
+            os.makedirs(dst_path)
+
+        download_command = (
+            f"python -m pip wheel --only-binary=:all: --wheel-dir={dst_path} -r"
+            f"{pip_requirements_path} --no-cache-dir"
```
harupy, line 139: If we run this method on windows, wheels for windows (e.g. [scikit_learn-1.1.2-cp39-cp39-win_amd64.whl](https://files.pythonhosted.org/packages/45/d1/9dfb055ce2893c309936d01efab048ac8ee89ec7ac006fa1f65ff67edaad/scikit_learn-1.1.2-cp39-cp39-win_amd64.whl) for scikit-learn) will be downloaded and the logged model cannot be served on Mac or Linux. Is this a problem?

mlflow/models/utils.py
dbczumar, line 268: Can we add the `@experimental` decorator to this API?

mlflow/models/wheeled_model.py
arjundc-db: Hey Haru, great question. We discussed this and decided it is not a problem. We are also storing the platform on which the wheels are built.

mlflow/models/utils.py
```diff
@@ -589,3 +591,94 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
         mlflow.models.validate_schema(input_data, model_signature.inputs)
     """
     _enforce_schema(data, expected_schema)
+
+
+@experimental
+def add_libraries_to_model(model_uri, run_id=None, registered_model_name=None):
+    """
+    This utility function only operates on a model which has been registered to the model registry.
```
dbczumar, line 599: Can we move this to an RST note below the description?

mlflow/models/utils.py
```diff
@@ -589,3 +591,94 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
...
+
+    Given the registered model_uri (models:/<model_name>/<model_version/stage/latest>), it will
```
dbczumar, line 601: Can we drop the use of the word `will` and instead phrase in the present tense (talk about what the API does, rather than what it will do)

```suggestion
    Given a registered model_uri (models:/<model_name>/<model_version/stage/latest>), it
```

mlflow/models/wheeled_model.py
dbczumar: @arjundc-db Can we document this limitation?

mlflow/models/utils.py
```diff
@@ -589,3 +591,97 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
...
+def add_libraries_to_model(model_uri, run_id=None, registered_model_name=None):
+    """
+    Given a registered model_uri (models:/<model_name>/<model_version/stage/latest>), this utility
+    re-logs the model along with all the required model libraries back to the model registry.
+    The required model libraries are stored along with the model as model artifacts. In
+    addition, supporting files to the model (e.g. conda.yaml, requirements.txt) are modified
+    to use the added libraries.
+
+    The default working of this util creates a new model version under the same registered model
+    name, this behavior can be overridden by passing in a registered_model_name to the utility.
```
dbczumar, line 606:
```suggestion
    By default, this utility creates a new model version under the same registered model specified by
    ``model_uri``. This behavior can be overridden by specifying the ``registered_model_name`` argument.
```

mlflow/models/utils.py
```diff
@@ -589,3 +591,97 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
         mlflow.models.validate_schema(input_data, model_signature.inputs)
     """
     _enforce_schema(data, expected_schema)
+
+
+@experimental
+def add_libraries_to_model(model_uri, run_id=None, registered_model_name=None):
+    """
+    Given a registered model_uri (models:/<model_name>/<model_version/stage/latest>), this utility
```
dbczumar, line 599:
```suggestion
    Given a registered model_uri (e.g. models:/<model_name>/<model_version>), this utility
```

mlflow/models/utils.py
```diff
@@ -589,3 +591,97 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
...
+    to use the added libraries.
+
+    The default working of this util creates a new model version under the same registered model
+    name, this behavior can be overridden by passing in a registered_model_name to the utility.
+
+    :param model_uri: A registered model uri in the model registry of the form
+                        models:/<model_name>/<model_version/stage/latest>
+    :param run_id: The run_id to which the model artifacts will be stored. If None, the model
+                    artifacts are stored in the original run_id where the inputted model version
+                    was created. Can be overridden by explicitly passing a run_id.
```
dbczumar, line 612:
```suggestion
    :param run_id: The run_id to which the model artifacts will be stored. If None, the model
                    artifacts are stored in the source run corresponding to model version specified
                    by ``model_uri``; if the model version does not have a source run, a new run is
                    created.
```

mlflow/models/utils.py
```diff
@@ -589,3 +591,97 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
...
+    re-logs the model along with all the required model libraries back to the model registry.
+    The required model libraries are stored along with the model as model artifacts. In
+    addition, supporting files to the model (e.g. conda.yaml, requirements.txt) are modified
+    to use the added libraries.
+
+    The default working of this util creates a new model version under the same registered model
+    name, this behavior can be overridden by passing in a registered_model_name to the utility.
+
+    :param model_uri: A registered model uri in the model registry of the form
+                        models:/<model_name>/<model_version/stage/latest>
```
dbczumar, line 609: Nittiest of nits: Can we align the param docstrings?

```suggestion
    :param model_uri: A registered model uri in the model registry of the form
                      models:/<model_name>/<model_version/stage/latest>
```

mlflow/models/utils.py
```diff
@@ -589,3 +591,97 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
...
+    :param run_id: The run_id to which the model artifacts will be stored. If None, the model
+                    artifacts are stored in the original run_id where the inputted model version
+                    was created. Can be overridden by explicitly passing a run_id.
+    :param registered_model_name: The new model version (model with its libraries) is
+                        registered under the inputted registered_model_name. If None, a new version
+                        is logged to the existing model in the model registry.
+
+
+    .. note::
+    This utility only operates on a model that has been registered to the model registry.
```
dbczumar, line 619: I think this line needs to be indented (?)

mlflow/models/utils.py
```diff
@@ -589,3 +591,97 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
...
+    name, this behavior can be overridden by passing in a registered_model_name to the utility.
+
+    :param model_uri: A registered model uri in the model registry of the form
+                        models:/<model_name>/<model_version/stage/latest>
+    :param run_id: The run_id to which the model artifacts will be stored. If None, the model
+                    artifacts are stored in the original run_id where the inputted model version
+                    was created. Can be overridden by explicitly passing a run_id.
+    :param registered_model_name: The new model version (model with its libraries) is
+                        registered under the inputted registered_model_name. If None, a new version
+                        is logged to the existing model in the model registry.
```
dbczumar, line 615: Can we capitalize Model Registry throughout the docstring?

tests/models/test_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    knn_model = knn.KNeighborsClassifier()
+    knn_model.fit(X, y)
+    return ModelWithData(model=knn_model, inference_data=X)
+
+
+def random_int(lo=1, hi=1000000000):
+    return random.randint(lo, hi)
+
+
+def test_adding_libraries_to_model_default(sklearn_knn_model):
```
dbczumar, line 29: Can we test the behavior when `model_uri` refers to a model version that does not have a source run? (A new run should be created)

mlflow/models/utils.py
```diff
@@ -589,3 +591,101 @@ def validate_schema(data: PyFuncInput, expected_schema: Schema) -> None:
...
+        model_uri = "models:/model-with-libs/1"
+
+        # Import utility
+        from mlflow.models.utils import add_libraries_to_model
+
+        # Log libraries to the original run of the model
+        add_libraries_to_model(model_uri)
+
+        # Log libraries to some run_id
+        add_libraries_to_model(model_uri, run_id=run_id)
```
harupy, line 655: `run_id` here is undefined.

dbczumar (approved):
LGTM once https://github.com/mlflow/mlflow/pull/6586/files#r965454194 is addressed. Thanks @arjundc-db !

mlflow/models/wheeled_model.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        Downloads all the wheels of the dependencies specified in the requirements.txt file.
+        :param pip_requirements_path: Path to requirements.txt in the model directory
+        :param dst_path: Path to the directory where the wheels are to be downloaded
+        """
+        if not os.path.exists(dst_path):
+            os.makedirs(dst_path)
+
+        download_command = (
+            f"python -m pip wheel --only-binary=:all: --wheel-dir={dst_path} -r"
```
harupy, line 200:
```suggestion
            f"{sys.executable} -m pip wheel --only-binary=:all: --wheel-dir={dst_path} -r"
```

It might be safe to use [`sys.executable`](https://docs.python.org/3/library/sys.html#sys.executable) here because `python` might point to python 2.

harupy:
I got the following warning when I served the model logged by `add_libraries_to_model`:

```
[2022-09-09 16:44:13 +0900] [68271] [INFO] Starting gunicorn 20.1.0
[2022-09-09 16:44:13 +0900] [68271] [INFO] Listening at: http://127.0.0.1:5000 (68271)
[2022-09-09 16:44:13 +0900] [68271] [INFO] Using worker: sync
[2022-09-09 16:44:13 +0900] [68277] [INFO] Booting worker with pid: 68277
2022/09/09 16:44:14 WARNING mlflow.pyfunc: Encountered an unexpected error (InvalidRequirement('Parse error at "\'/websock\'": Expected string_end')) while detecting model dependency mismatches. Set logging level to DEBUG to see the full traceback.
```

harupy (approved):

PR #6600, opened 2022-08-26
Add support for Decimal object to Double cast in ML Flow 

shitaoli-db:

## How is this patch tested?
Added unit test for Decimal to float object in enforce schema cast.
Also added end to end test with databricks automl model inference, which has double schema in model, the inference was successful

image
image

Model inference now allow user to use Decimal object as input. Note decimal input will only be allowed when schema is double and the data will be casted to double and thus may lost precision.

Components

Interface

Language

Integrations


mlflow/pyfunc/__init__.py
```diff
@@ -397,6 +402,15 @@ def _enforce_mlflow_datatype(name, values: pandas.Series, t: DataType):
             raise MlflowException(
                 "Failed to convert column {0} from type {1} to {2}.".format(name, values.dtype, t)
             )
+    if t == DataType.double and values.dtype == decimal.Decimal:
```
yxiong, line 405: I thought in our test the `values.dtype` is `object`, while `type(values[0])` is `decimal.Decimal`. Can you confirm?

yxiong (changes requested):

mlflow/pyfunc/__init__.py
shitaoli-db: yeah, verified
image

shitaoli-db: Also our unit test and e2e notebook test all passed when object is Decimal

yxiong: Ah, I didn't know `dtype == object` and `dtype == decimal.Decimal` can both be true. LGTM!

yxiong (approved):

mlflow/pyfunc/__init__.py
```diff
@@ -397,6 +402,15 @@ def _enforce_mlflow_datatype(name, values: pandas.Series, t: DataType):
...
+        # NB: Pyspark Decimal columne get converted to decimal.Decimal when converted to pandas
```
tomasatdatabricks, line 406: nit: columne -> column.
in order to support ...  - is part of the comment missing?

tomasatdatabricks (commented):
Looks good.

Can you also update docs please?
Can we make this work only in spark_udf environment or protect it with an argument? E.g. we could use an env variable or extra argument to pyfunc.load_model. Would that be feasible?

mlflow/pyfunc/__init__.py
shitaoli-db: Thank you. Done.

shitaoli-db:
> 

Let me try with an extra argument first. Do you know how did we generate this message?

image
If we use the extra argument, then we have to customize this message in the UI to let customer know that if they want their model with decimal training data work in inference, they have to pass an extra argument.


> Looks good.
> 
> Can you also update docs please? Can we make this work only in spark_udf environment or protect it with an argument? E.g. we could use an env variable or extra argument to pyfunc.load_model. Would that be feasible?

mlflow/pyfunc/__init__.py
```diff
@@ -709,7 +729,10 @@ def _warn_dependency_requirement_mismatches(model_path):
 
 
 def load_model(
-    model_uri: str, suppress_warnings: bool = False, dst_path: str = None
+    model_uri: str,
+    suppress_warnings: bool = False,
+    dst_path: str = None,
+    support_decimal: bool = False,
```
dbczumar, line 735: Is there a way to accomplish the goal of providing conditional decimal enforcement without adding this flag to the `mlflow.pyfunc.load_model()` API? I'd rather not extend this API for a niche use case.

yxiong: I personally don't think we need this flag at all. Can we just always enable this conversion? I don't see any real use case of setting that to false (or only add that flag later when we confirm it's a real use case).

@dbczumar @tomasatdatabricks what do you think?

shitaoli-db: I think tom provided 2 approach to me 
image

The other option I could think on top of my head is use environment variable

https://github.com/mlflow/mlflow/blob/master/mlflow/environment_variables.py

I am not sure which approach is more preferrable, let's have a quick agreement.

dbczumar: I agree with @yxiong here

shitaoli-db: Reverted back the change, now we always enable the cast.

yxiong (approved):

dbczumar (approved):
LGTM!

shitaoli-db:
autoformat

shitaoli-db:
automerge

Issue #6485, opened 2022-08-16
[FR] Add Model Registry support to the FileStore


dbczumar:
### Willingness to contribute

No. I cannot contribute this feature at this time.

### Proposal Summary

Add Model Registry support to the MLflow File Store so that users can use the Model Registry with better scale than sqlite without setting up SQL databases.

### Motivation

> #### What is the use case for this feature?

Enable easier / more scalable Model Registry use than what we currently provide with sqlite.

> #### Why is this use case valuable to support for MLflow users in general?

^

> #### Why is this use case valuable to support for your project(s) or organization?

^

> #### Why is it currently difficult to achieve this use case?

FileStore does not currently support Model Registry

### What component(s) does this bug affect?

- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [X] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/pipelines`: Pipelines, Pipeline APIs, Pipeline configs, Pipeline Templates
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

### What interface(s) does this bug affect?

- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

### What language(s) does this bug affect?

- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

### What integration(s) does this bug affect?

- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations
- [ ] `integrations/databricks`: Databricks integrations

dbczumar:
@serena-ruan Would you be interested in working on this item?

[...]

Issue #6339, opened 2022-07-27
[BUG] Internal Server Error on delete


jaelynlitz:
### Willingness to contribute

Yes. I would be willing to contribute a fix for this bug with guidance from the MLflow community.

### MLflow version

1.19.0

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: CentOS 7
- **Python version**: 3.9
- **yarn version, if running the dev UI**:

### Describe the problem

After using mlflow for ~1 year, we've run into this difficult to reproduce bug several times where an experiment tab will show an INTERNAL_SERVER_ERROR and stops displaying the runs that were stored in that experiment.

This issue seems to come up when deleting one or multiple runs from an experiment, but it doesn't happen every time a delete is attempted. But it seems to usually happen on the first delete attempt in an experiment if it's going to fail... The delete command hangs longer than usual then prompts the INTERNAL_SERVER_ERROR.

<img width="1624" alt="Screen Shot 2022-07-26 at 4 30 40 PM" src="https://user-images.githubusercontent.com/93339979/181131068-8ea76d64-f1a8-473d-8c4c-63b7e4d0fad0.png">

Besides a full fix, I would also be interested to learn if it's possible to retrieve the runs that were stored in the broken experiment.

### Steps to reproduce the bug

This is inconsistently reproducible (and I apologize, I know that's not very helpful), but I logged 50-100 runs in a new experiment. Each run had 5 metrics with ~200 values. I then tried deleting a random amount of runs. Sometimes one or the whole page. Sometimes the experiment broke, but most times the delete executed no problem.

### Code to generate data required to reproduce the bug

```python
import mlflow
import boto3
import random

MLFLOW_EXPERIMENT_NAME = "break server"
if not mlflow.get_experiment_by_name(MLFLOW_EXPERIMENT_NAME):
      mlflow.create_experiment(MLFLOW_EXPERIMENT_NAME)
mlflow.set_experiment(MLFLOW_EXPERIMENT_NAME)

mlflow.start_run()

mlflow.log_param("number", 1)

randomlist = random.sample(range(10, 500), 200)
for x in randomlist:
  mlflow.log_metric("rand1", x)
  mlflow.log_metric("rand2", x)
  mlflow.log_metric("rand3", x)
  mlflow.log_metric("rand4", x)
  mlflow.log_metric("rand5", x)

mlflow.end_run()
```
I then ran this 50-100 times to fill the experiment with runs to delete.

[...]

PR #6605, opened 2022-08-26
Support FileStore for Model Registry

serena-ruan:
Signed-off-by: Xinyue Ruan <serena.rxy@gmail.com>

## Related Issues/PRs

#6485

Add FileStore support for model registry

## How is this patch tested?

Will add tests soon.

Components

Interface

Language

Integrations


serena-ruan:
Hi @dbczumar, this is still a draft PR without unit tests, but I'll add them in following commits. But there're some details I might need some guidance/clarification 😃 
1. What's the default behavior of saving model versions under a registered model? My current implementation is like creating subfolder named "version-{version_numer}" for different versions under the registered model folder, is this expected?
2. Method `transition_model_version_stage` contains a param named `stage` but the docstring shows as `new_stage` so we might want consistency here. Though I prefer new_stage, seems base and other storages are all using stage.
3. When we update a model_version stage, do we also update the last_update_timestamp of the parent registered_model? And a general question would be that should we update the parent registered_model whenever there's an action happening for model_versions within it?
4. Is it possible if the same tag (key) for registered_model/model_version set multiple times? If so should we raise an error or we override previous value?
5. Do we support restore of registered models? For now seems not, but any plans in the future?

Pls let me know if you have other concerns or suggestions. Thanks!

WeichenXu123:
> What's the default behavior of saving model versions under a registered model? My current implementation is like creating subfolder named "version-{version_numer}" for different versions under the registered model folder, is this expected?

Sounds good

> Method transition_model_version_stage contains a param named stage but the docstring shows as new_stage so we might want consistency here. Though I prefer new_stage, seems base and other storages are all using stage.

Doc error. We should modify doc param name to be stage

> When we update a model_version stage, do we also update the last_update_timestamp of the parent registered_model? And a general question would be that should we update the parent registered_model whenever there's an action happening for model_versions within it?

In sqlalchemy backend, some operations update registered_model.last_update_timestamp (e.g. transition_model_version_stage),  but some others not (e.g. update_model_version).
@dbczumar What's the target behavior do you expect ? All related operations update registered_model.last_update_timestamp  ? 

> Is it possible if the same tag (key) for registered_model/model_version set multiple times? If so should we raise an error or we override previous value?

We should override previous value, You can try sqlalchemy backend to check its behavior.

> Do we support restore of registered models? For now seems not, but any plans in the future?

Do you mean restore registered model after it being deleted ? Is it an important in some scenarios ?

serena-ruan:
> > What's the default behavior of saving model versions under a registered model? My current implementation is like creating subfolder named "version-{version_numer}" for different versions under the registered model folder, is this expected?
> 
> Sounds good
> 
> > Method transition_model_version_stage contains a param named stage but the docstring shows as new_stage so we might want consistency here. Though I prefer new_stage, seems base and other storages are all using stage.
> 
> Doc error. We should modify doc param name to be stage
> 
> > When we update a model_version stage, do we also update the last_update_timestamp of the parent registered_model? And a general question would be that should we update the parent registered_model whenever there's an action happening for model_versions within it?
> 
> In sqlalchemy backend, some operations update registered_model.last_update_timestamp (e.g. transition_model_version_stage), but some others not (e.g. update_model_version). @dbczumar What's the target behavior do you expect ? All related operations update registered_model.last_update_timestamp ?
> 
> > Is it possible if the same tag (key) for registered_model/model_version set multiple times? If so should we raise an error or we override previous value?
> 
> We should override previous value, You can try sqlalchemy backend to check its behavior.
> 
> > Do we support restore of registered models? For now seems not, but any plans in the future?
> 
> Do you mean restore registered model after it being deleted ? Is it an important in some scenarios ?

Thanks! I guess for the last one it's just out of curiosity. Because we support restore of deleting experiments :) No special scenarios

dbczumar:
> In sqlalchemy backend, some operations update registered_model.last_update_timestamp (e.g. transition_model_version_stage), but some others not (e.g. update_model_version). @dbczumar What's the target behavior do you expect ? All related operations update registered_model.last_update_timestamp ?

I think that all operations related to a registered model should update the timestamp.

WeichenXu123:
> I think that all operations related to a registered model should update the timestamp.

Then sqlalchemy backend need some update to satisfy this demand.

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+                f"Registered Model (name={name}) already exists.",
+                RESOURCE_ALREADY_EXISTS,
+            )
+
+    def _update_registered_model_last_updated_time(self, name, updated_time):
+        registered_model = self._get_registered_model(name)
+        registered_model.last_updated_timestamp = updated_time
+        registered_model_dict = dict(registered_model)
+        del registered_model_dict["tags"]
+        write_yaml(
```
dbczumar, line 107: We noticed in https://github.com/mlflow/mlflow/issues/6339 that, if a failure occurs during YAML write, file contents may be lost. As a result, we added https://github.com/mlflow/mlflow/blob/08205116b0844d43fdeef5062ff1d81304cce17a/mlflow/store/tracking/file_store.py#L998-L1026 to `FileStore` to write updates to temporary files and then use `shutil.move` to replace the preexisting file with the copied file. 

Can we turn that into a reusable utility, e.g. by placing it in `file_utils` and use it throughout this Model Registry FileStore?

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+            name=name,
+            creation_timestamp=creation_time,
+            last_updated_timestamp=creation_time,
+            description=description,
+            latest_versions=latest_versions,
+            tags=tags,
+        )
+        registered_model_dict = dict(registered_model)
+        # tags are stored under TAGS_FOLDER_NAME so remove them in meta file.
+        del registered_model_dict["tags"]
```
dbczumar, line 144: Nit, instead of setting the tags on Line 140 and then deleting them here, can we pass `tags=None` on line 140? Can we apply similar changes for `create_model_version`?

dbczumar (commented):
@serena-ruan This is fantastic! Apologies for the delayed review - I'll try it out manually tomorrow and will provide more detailed feedback. Left a couple of very minor comments to start. Looking forward to getting this merged!

mlflow/store/model_registry/file_store.py
serena-ruan: Sounds good!

mlflow/tracking/_model_registry/utils.py
```diff
@@ -161,6 +162,9 @@ def _get_store_registry():
     for scheme in DATABASE_ENGINES:
         _model_registry_store_registry.register(scheme, _get_sqlalchemy_store)
 
+    for scheme in ["", "file"]:
+        _model_registry_store_registry.register(scheme, _get_file_store)
```
dbczumar, line 166: `_get_file_store()` returns an instance of `mlflow.store.tracking.file_store.FileStore`, but we want to return an instance of `mlflow.store.model_registry.FileStore`

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+        """
+        Run checks before running directory operations.
+        """
+        if not exists(self.root_directory):
+            raise Exception("'%s' does not exist." % self.root_directory)
+        if not is_directory(self.root_directory):
+            raise Exception("'%s' is not a directory." % self.root_directory)
+
+    def _validate_registered_model_does_not_exist(self, name):
+        model = self.get_registered_model(name)
```
dbczumar, line 95: `self.get_registered_model()` throws if the model doesn't exist, but this method asserts that the model doesn't exist. As a result, this method always throws.

mlflow/store/model_registry/file_store.py
dbczumar: Saw this was resolved, but I don't think changes were pushed. Unresolving for now :)

mlflow/store/model_registry/file_store.py
dbczumar: Unresolving until changes are pushed :)

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+        """
+        Get registered model instance by name.
+
+        :param name: Registered model name.
+        :return: A single :py:class:`mlflow.entities.model_registry.RegisteredModel` object.
+        """
+        try:
+            return self._get_registered_model(name)
+        except MlflowException as e:
+            if e.error_code == RESOURCE_DOES_NOT_EXIST:
```
dbczumar, line 333: This check doesn't quite work as-is because `e.error_code` is a string while `RESOURCE_DOES_NOT_EXIST` is an `int`.

I think we need 

```suggestion
            if e.error_code == ErrorCode.Name(RESOURCE_DOES_NOT_EXIST):
```

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+                warnings.warn(e.message)
```
dbczumar, line 334: Because `get_registered_model()` is called by `_validate_registered_model_does_not_exist()`, I don't think we should emit a warning here, since the absence of the model is expected in `_validate_registered_model_does_not_exist()`. I think it's up to the caller of this method to handle the `None` result.

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+                logging.warning(
+                    "Model Version creation error (name=%s) Retrying %s more time%s.",
+                    name,
+                    str(more_retries),
+                    "s" if more_retries > 1 else "",
+                )
+        raise MlflowException(
+            "Model Version creation error (name={}). Giving up after "
+            "{} attempts.".format(name, self.CREATE_MODEL_VERSION_RETRIES)
+        )
```
dbczumar, line 550: Can we move this inside the `except Exception` and throw it if we're out of retries? That way we can make this a raise `MlflowException from exc` (where `exc` is the caught `Exception`) so that it's easy to see what the failure was. Right now, there isn't any exception content to debug with.

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+                return 1
+
+        _validate_model_name(name)
+        for tag in tags or []:
+            _validate_model_version_tag(tag.key, tag.value)
+        for attempt in range(self.CREATE_MODEL_VERSION_RETRIES):
+            try:
+                creation_time = now()
+                registered_model = self._get_registered_model(name)
+                registered_model.last_updated_timestamp = creation_time
```
dbczumar, line 514: This appears to throw with `AttributeError: can't set attribute`

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+                    last_updated_timestamp=creation_time,
+                    description=description,
+                    current_stage=None,
+                    source=source,
+                    run_id=run_id,
+                    run_link=run_link,
+                )
+                model_version.tags = [
+                    ModelVersionTag(key=tag.key, value=tag.value) for tag in tags or []
+                ]
```
dbczumar, line 529: This similarly throws with `AttributeError: can't set attribute`

mlflow/store/model_registry/file_store.py
dbczumar: @serena-ruan I see. I don't think we can apply my suggestion because the returned `RegisteredModel` needs to contain tags.

dbczumar (commented):
@serena-ruan Awesome progress! Left a few comments based on manual testing. To identify / patch more issues, I think adding a basic test case that uses the file store to run the following workflow might be useful:

```
import mlflow

class Mod(mlflow.pyfunc.PythonModel):
    def predict(self, ctx, inp):
        return 7

for i in range(5):
    for _ in range(3):
        mlflow.pyfunc.log_model(
            python_model=Mod(),
            artifact_path="pyfunc_mod",
            registered_model_name=f"mod{i}"
        )
```

Once this logging code succeeds, you can add `MlflowClient().get_registered_model() / get_model_version() / search_registered_models() / ...` calls to the workflow to ensure that they also succeed.

serena-ruan:
> 

Thanks for the review!! I'll take a look at those comments while adding unit tests for all functions, and will ping you once it's ready for review 😃

mlflow/store/model_registry/file_store.py
WeichenXu123: The implementation in https://github.com/mlflow/mlflow/pull/6605#discussion_r957813112 is not thread-safe. Shall we add file lock for it ? (don't need do it in this PR)

WeichenXu123:
@serena-ruan Could you add unit tests ? :)

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+    def _get_registered_model_path(self, name):
+        self._check_root_dir()
+        _validate_model_name(name)
+        model_path = join(self.root_directory, name)
+        if exists(model_path):
+            return model_path
+        else:
+            return None
+
+    def __get_registered_model_from_path(self, model_path):
```
WeichenXu123, line 160: why the method starts with double `_` ?

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,767 @@
...
+                mv.last_updated_timestamp = updated_time
+                new_model_version_dir = join(new_meta_dir, f"version-{mv.version}")
+                mkdir(new_model_version_dir)
+                model_version_dict = dict(mv)
+                del model_version_dict["tags"]
+                write_yaml(new_model_version_dir, FileStore.META_DATA_FILE_NAME, model_version_dict)
+                if mv.tags is not None:
+                    for tag in mv.tags:
+                        self.set_model_version_tag(new_name, mv.version, tag)
+            shutil.rmtree(model_path)
```
WeichenXu123, line 232: The loop performance is bad. We can remove the loop by:
1. Avoid saving "mv.name" in model version meta file
we can check the parent diretory name of the model version meta file, to get the register-model name.
2. Avoid update "last_updated_timestamp" field in mv meta file when register model name is updated. Instead, when we want to read `mv.last_updated_timestamp`, we read last_updated_timestamp of both "registered model" and this "model version", and return the bigger timestamp

mlflow/store/model_registry/file_store.py
serena-ruan: I found that sql store is designed to raise Mlflow exception when the model doesn't exist when calling `get_registered_model()`, so I mimic that logic in the latest version :)

mlflow/store/model_registry/file_store.py
serena-ruan: I think changing to this logic will make other processes more complicated: for example, whenever you do updates to the model version, and it returns a `ModelVersion` object, we need to go over this logic to fetch the upper folder name and timestamp; Besides, if we save extra field in registered_model meta file, we need to exclude it when constructing the object from meta file; 
I do agree that saving `mv.name` in the registered model only makes sense, but this `rename_registered_model` looks like a relative less frequent move, WDYT?

serena-ruan:
@dbczumar Hey I updated some unit tests, could you have another look when you got time? Thanks :D

serena-ruan:
@WeichenXu123 @dbczumar Hey could you help review this again? 😃

mlflow/utils/search_utils.py
```diff
@@ -957,3 +1052,140 @@ def _invalid_statement_token_search_model_registry(cls, token):
             return False
         else:
             return True
+
+
+class SearchModelVersionUtils(SearchUtils):
+    NUMERIC_ATTRIBUTES = {"creation_timestamp", "last_updated_timestamp"}
+    # what should we support here?
```
dbczumar, line 1059: This looks good to me, and it's consistent with `sqlalchemy_store.py` :)

mlflow/utils/search_utils.py
```diff
@@ -957,3 +1052,140 @@ def _invalid_statement_token_search_model_registry(cls, token):
...
+    VALID_SEARCH_ATTRIBUTE_KEYS = {
+        "name",
+        "run_id",
+        "source_path",  # this is for sql store only
```
dbczumar, line 1063: Can we support it for file store as well?

mlflow/store/model_registry/file_store.py
```diff
@@ -0,0 +1,769 @@
...
+        if len(source_dirs) == 0:
+            return root_dir, []
+        file_names = []
+        for root, _, files in os.walk(source_dirs[0]):
+            for name in files:
+                abspath = join(root, name)
+                file_names.append(os.path.relpath(abspath, source_dirs[0]))
+        if sys.platform == "win32":
+            # Turn metric relative path into metric name.
+            # Metrics can have '/' in the name. On windows, '/' is interpreted as a separator.
```
dbczumar, line 395: Nit: Can registered models / model versions contain the `/` name? Perhaps we want to update the comment here to refer to registered models & model versions instead of metrics

dbczumar (commented):
Hi @serena-ruan , awesome progress! The code looks great!

I tried out the following workflow and ran into a search error:

```
import mlflow

class MyMod(mlflow.pyfunc.PythonModel):
    def predict(self, ctx, inp):
        return 7


mlflow.pyfunc.log_model(python_model=MyMod(), artifact_path="foo", registered_model_name="model1")
mlflow.pyfunc.log_model(python_model=MyMod(), artifact_path="foo", registered_model_name="model2")

print(mlflow.MlflowClient().search_registered_models())
print(mlflow.MlflowClient().search_model_versions("name = 'model1'"))
print(mlflow.MlflowClient().search_model_versions("name = 'model2'"))
```

```
Successfully registered model 'model1'.
2022/10/20 17:37:09 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.                     Model name: model1, version 1
Created version '1' of model 'model1'.
Successfully registered model 'model2'.
2022/10/20 17:37:11 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.                     Model name: model2, version 1
Created version '1' of model 'model2'.
  File "log.py", line 11, in <module>
    print(mlflow.MlflowClient().search_registered_models())
  File "/Users/corey.zumar/mlflow/mlflow/tracking/client.py", line 2084, in search_registered_models
    return self._get_registry_client().search_registered_models(
  File "/Users/corey.zumar/mlflow/mlflow/tracking/_model_registry/client.py", line 130, in search_registered_models
    return self.store.search_registered_models(filter_string, max_results, order_by, page_token)
  File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 309, in search_registered_models
    registered_models = self._list_all_registered_models()
  File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 274, in _list_all_registered_models
    registered_models.append(self._get_registered_model_from_path(path))
  File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 175, in _get_registered_model_from_path
    meta = FileStore._read_yaml(model_path, FileStore.META_DATA_FILE_NAME)
  File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 769, in _read_yaml
    return _read_helper(root, file_name, attempts_remaining=retries)
  File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 762, in _read_helper
    result = read_yaml(root, file_name)
  File "/Users/corey.zumar/mlflow/mlflow/utils/file_utils.py", line 213, in read_yaml
    raise MissingConfigException("Yaml file '%s' does not exist." % file_path)
mlflow.exceptions.MissingConfigException: Yaml file '/private/tmp/mreg/mlruns/.trash/meta.yaml' does not exist.
```

I'll continue trying out other workflows in the meantime. Can we address this error and add test coverage for it? Thanks so much!

mlflow/store/model_registry/file_store.py
serena-ruan: I don't see any checks preventing '/' in the model name so we can keep this. I'll update the comment, thanks for catching it!

serena-ruan:
> Hi @serena-ruan , awesome progress! The code looks great!
> 
> I tried out the following workflow and ran into a search error:
> 
> ```
> import mlflow
> 
> class MyMod(mlflow.pyfunc.PythonModel):
>     def predict(self, ctx, inp):
>         return 7
> 
> 
> mlflow.pyfunc.log_model(python_model=MyMod(), artifact_path="foo", registered_model_name="model1")
> mlflow.pyfunc.log_model(python_model=MyMod(), artifact_path="foo", registered_model_name="model2")
> 
> print(mlflow.MlflowClient().search_registered_models())
> print(mlflow.MlflowClient().search_model_versions("name = 'model1'"))
> print(mlflow.MlflowClient().search_model_versions("name = 'model2'"))
> ```
> 
> ```
> Successfully registered model 'model1'.
> 2022/10/20 17:37:09 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.                     Model name: model1, version 1
> Created version '1' of model 'model1'.
> Successfully registered model 'model2'.
> 2022/10/20 17:37:11 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.                     Model name: model2, version 1
> Created version '1' of model 'model2'.
> Traceback (most recent call last):
>   File "log.py", line 11, in <module>
>     print(mlflow.MlflowClient().search_registered_models())
>   File "/Users/corey.zumar/mlflow/mlflow/tracking/client.py", line 2084, in search_registered_models
>     return self._get_registry_client().search_registered_models(
>   File "/Users/corey.zumar/mlflow/mlflow/tracking/_model_registry/client.py", line 130, in search_registered_models
>     return self.store.search_registered_models(filter_string, max_results, order_by, page_token)
>   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 309, in search_registered_models
>     registered_models = self._list_all_registered_models()
>   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 274, in _list_all_registered_models
>     registered_models.append(self._get_registered_model_from_path(path))
>   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 175, in _get_registered_model_from_path
>     meta = FileStore._read_yaml(model_path, FileStore.META_DATA_FILE_NAME)
>   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 769, in _read_yaml
>     return _read_helper(root, file_name, attempts_remaining=retries)
>   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 762, in _read_helper
>     result = read_yaml(root, file_name)
>   File "/Users/corey.zumar/mlflow/mlflow/utils/file_utils.py", line 213, in read_yaml
>     raise MissingConfigException("Yaml file '%s' does not exist." % file_path)
> mlflow.exceptions.MissingConfigException: Yaml file '/private/tmp/mreg/mlruns/.trash/meta.yaml' does not exist.
> ```
> 
> I'll continue trying out other workflows in the meantime. Can we address this error and add test coverage for it? Thanks so much!

I just added this testcase in the test file, and it works fine on my side 🤔 But I'll see the pipeline results just for double confirmation.

mlflow/utils/search_utils.py
serena-ruan: Supporting search by source has some conflicts with existing methods, as `source` is identified as a TokenType.Keyword and the value is identified as TokenType.Literal.String.Single. I'll investigate more on this :)

dbczumar:
> > Hi @serena-ruan , awesome progress! The code looks great!
> > I tried out the following workflow and ran into a search error:
> > ```
> > import mlflow
> > 
> > class MyMod(mlflow.pyfunc.PythonModel):
> >     def predict(self, ctx, inp):
> >         return 7
> > 
> > 
> > mlflow.pyfunc.log_model(python_model=MyMod(), artifact_path="foo", registered_model_name="model1")
> > mlflow.pyfunc.log_model(python_model=MyMod(), artifact_path="foo", registered_model_name="model2")
> > 
> > print(mlflow.MlflowClient().search_registered_models())
> > print(mlflow.MlflowClient().search_model_versions("name = 'model1'"))
> > print(mlflow.MlflowClient().search_model_versions("name = 'model2'"))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > Successfully registered model 'model1'.
> > 2022/10/20 17:37:09 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.                     Model name: model1, version 1
> > Created version '1' of model 'model1'.
> > Successfully registered model 'model2'.
> > 2022/10/20 17:37:11 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.                     Model name: model2, version 1
> > Created version '1' of model 'model2'.
> > Traceback (most recent call last):
> >   File "log.py", line 11, in <module>
> >     print(mlflow.MlflowClient().search_registered_models())
> >   File "/Users/corey.zumar/mlflow/mlflow/tracking/client.py", line 2084, in search_registered_models
> >     return self._get_registry_client().search_registered_models(
> >   File "/Users/corey.zumar/mlflow/mlflow/tracking/_model_registry/client.py", line 130, in search_registered_models
> >     return self.store.search_registered_models(filter_string, max_results, order_by, page_token)
> >   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 309, in search_registered_models
> >     registered_models = self._list_all_registered_models()
> >   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 274, in _list_all_registered_models
> >     registered_models.append(self._get_registered_model_from_path(path))
> >   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 175, in _get_registered_model_from_path
> >     meta = FileStore._read_yaml(model_path, FileStore.META_DATA_FILE_NAME)
> >   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 769, in _read_yaml
> >     return _read_helper(root, file_name, attempts_remaining=retries)
> >   File "/Users/corey.zumar/mlflow/mlflow/store/model_registry/file_store.py", line 762, in _read_helper
> >     result = read_yaml(root, file_name)
> >   File "/Users/corey.zumar/mlflow/mlflow/utils/file_utils.py", line 213, in read_yaml
> >     raise MissingConfigException("Yaml file '%s' does not exist." % file_path)
> > mlflow.exceptions.MissingConfigException: Yaml file '/private/tmp/mreg/mlruns/.trash/meta.yaml' does not exist.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll continue trying out other workflows in the meantime. Can we address this error and add test coverage for it? Thanks so much!
> 
> I just added this testcase in the test file, and it works fine on my side 🤔 But I'll see the pipeline results just for double confirmation.

Thanks @serena-ruan ! The issue I'm running into is that `_get_all_registered_models()` is including the `.trash` folder in the list of subdirectories. I think the broader problem is that we're storing models in the top-level `mlruns` directory at the same level as experiments and the trash. For example, if I create an experiment and then try to list registered models, I see the experiment in the list of registered model paths.

To test, I ran the following script:

```
import mlflow
from mlflow import MlflowClient
from mlflow.pyfunc import PythonModel

class MyModel(PythonModel):
    def predict(self, context, model_input):
        return 7

with mlflow.start_run():
    mlflow.pyfunc.log_model(
        python_model=MyModel(), artifact_path="foo", registered_model_name="model1"
    )

with mlflow.start_run():
    mlflow.log_param("A", "B")

from mlflow.store.model_registry.file_store import FileStore

fs = FileStore("mlruns")
print(fs._get_all_registered_model_paths())
``` 

This produces the following output:

```
Registered model 'model1' already exists. Creating a new version of this model...
2022/10/25 22:00:56 INFO mlflow.tracking._model_registry.client: Waiting up to 300 seconds for model version to finish creation.                     Model name: model1, version 3
Created version '3' of model 'model1'.
['mlruns/0', 'mlruns/model1', 'mlruns/.trash', 'mlruns/model2']
```

`mlruns/0` corresponds to the default experiment with ID 0.

To solve this problem, I think we need to make a reserved folder called `models` inside `mlruns`. This should be safe since existing experiments in the directory use integer-like string IDs, so there shouldn't be any existing top-level folders called `models`.

mlflow/store/model_registry/__init__.py
```diff
@@ -1,2 +1,5 @@
+# Path to default location for backend when using local FileStore.
+DEFAULT_LOCAL_FILE_AND_ARTIFACT_PATH = "./mlruns/models"
```
dbczumar, line 2: Instead of setting the default path to `./mlruns/models` here, I think we should update the logic of the `FileStore` class to always write to / read from a `models` subdirectory of the path that the user specifies for their file store location. For example, if the user calls `mlflow.set_tracking_uri("/tmp/dir")`, then the current approach will still write experiments and models to the same directory.

serena-ruan:
@dbczumar Hey Corey, hope I fixed the issue this time 😅

dbczumar (commented):
Hi @serena-ruan , apologies for the delay (was OOO). I think this is very close. One small thing I noticed is that, after logging models, calls to `mlflow.search_runs()` and `mlflow.search_experiments()` print a warning because they think that the `models` directory is an experiment directory:

```
WARNING:root:Malformed experiment 'models'. Detailed error Yaml file '/private/tmp/mlruns/models/meta.yaml' does not exist.
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/file_store.py", line 271, in list_experiments
    experiment = self._get_experiment(exp_id, view_type)
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/file_store.py", line 403, in _get_experiment
    meta = FileStore._read_yaml(experiment_dir, FileStore.META_DATA_FILE_NAME)
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/file_store.py", line 1060, in _read_yaml
    return _read_helper(root, file_name, attempts_remaining=retries)
  File "/Users/corey.zumar/mlflow/mlflow/store/tracking/file_store.py", line 1053, in _read_helper
    result = read_yaml(root, file_name)
  File "/Users/corey.zumar/mlflow/mlflow/utils/file_utils.py", line 213, in read_yaml
    raise MissingConfigException("Yaml file '%s' does not exist." % file_path)
mlflow.exceptions.MissingConfigException: Yaml file '/private/tmp/mlruns/models/meta.yaml' does not exist.
```

Can we update file store to exclude the `models` directory when reading experiments?

dbczumar (approved):
LGTM! Thanks so much, @serena-ruan ! Tested this out, and it works like a charm!

Issue #694, opened 2018-11-08
Proposal:Add the function returning the result of predict_proba() to Model API


masa26hiro:
### System information
- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**:Windows10(Docker : Debian GNU/Linux 9)
- **MLflow installed from (source or binary)**: source
- **MLflow version (run ``mlflow --version``)**:0.7.0
- **Python version**: 3.7.0

### Describe the problem
 Currently, the serve_model method of mlflow/sklearn.py returns only "model.predict(records)".
I think that there are times when we want to know the classification probability of each class when classifying.
So I suggest adding a function to return "model.predict_proba(records)" results to this method.

Thank you.

MichaelMarien:
Hi,

a quick fix would be to overwrite the "predict" function of your sklearn model with "predict_proba" before saving it. This is a bit hacky though. Not sure what the view of the core dev team is, but if they want to adress this mroe nicely, I'd be happy to pick it up.

masa26hiro:
Hi,

Thanks for your reply.
I used to rewrite predict () into predict_proba () by myself before too. I believe there is a demand for data scientists who wish to use predict_proba ().

yassineAlouini:
I have a similar use case but with another custom function `predict_distribution`. 
For now, I will override the `predict` method during the training but it would be better to offer a way to do it natively. Any help or suggestions are welcome. Thanks in advance. :)

smurching:
Hi all, sorry for the delayed response - this should be possible in MLflow >= 1.0 by using custom Python models ([see the docs](https://www.mlflow.org/docs/latest/models.html#model-customization)), in particular [this example](https://www.mlflow.org/docs/latest/models.html#example-saving-an-xgboost-model-in-mlflow-format) - you can define a wrapper model class that loads the persisted sklearn model & returns the results of `predict_proba`. Please feel free to reopen with any questions!

ssimontacchi:
Hi, can you explain what you mean here in a little more detail? Thanks!

skylarbpayne:
@smurching Coming back to this. This seems like something a lot of folks would find useful. Seems a tad silly to have all users write their own custom wrapper just to change the predict function.

How would you feel about me making a PR to add an optional argument to the sklearn model flavor to specify the "predict" function that should be used? This would be backwards compatible and would avoid the need for everyone to create their own wrapper.

E.g.

```python
[...]

PR #6631, opened 2022-08-29
Improve flexibility of PyFuncModel prediction

skylarbpayne:
## Related Issues/PRs
https://github.com/mlflow/mlflow/issues/694

# Improve flexibility of PyFuncModel prediction

Prior to this change, `PyFuncModel` assumes the implemented class has a `predict` function. This is a common interface, notably used in libraries such as `scikit-learn`. However, there are many use cases where we may want to use the `pyfunc` model flavor (as it is flexible!) while customizing the actual predict function used.

A common use case is illustrated in [issue 694](https://github.com/mlflow/mlflow/issues/694); folks want to use the `predict_proba` function instead of `predict`. However, the resolution in that ticket is effectively to have users write their own custom wrappers.

Given that (as far as I can tell), the pyfunc flavor is intended to "unify" underlying base implementations so you can have a similar interface for potentially dissimilar model formats, it seems prudent to allow folks to customize the predict function used without having to write a custom wrapper. Such functionality is demonstrated in this PR for the `sklearn` model logging, but can possibly be leveraged by other model flavors as well.

## How is this patch tested?

A new test is added which failed before implementation.
Existing tests all pass ensuring backwards compatibility.

## Release Notes

Improve flexibility of `PyFuncModel` by allowing predict function override. For example, this can be used to log a `pyfunc` flavor which uses the `scikit-learn` `predict_proba` function instead of `predict`.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -624,13 +624,15 @@ class PyFuncModel:
     ``model_meta`` contains model metadata loaded from the MLmodel file.
     """
 
-    def __init__(self, model_meta: Model, model_impl: Any):
+    def __init__(self, model_meta: Model, model_impl: Any, pyfunc_predict_func: str = "predict"):
         if not hasattr(model_impl, "predict"):
             raise MlflowException("Model implementation is missing required predict method.")
         if not model_meta:
             raise MlflowException("Model is missing metadata.")
         self._model_meta = model_meta
         self._model_impl = model_impl
+        assert hasattr(model_impl, pyfunc_predict_func)
+        self._predict_fn = getattr(model_impl, pyfunc_predict_func)
 
     def predict(self, data: PyFuncInput) -> PyFuncOutput:
         """
@@ -649,7 +651,7 @@ class PyFuncModel:
         input_schema = self.metadata.get_input_schema()
         if input_schema is not None:
             data = _enforce_schema(data, input_schema)
-        return self._model_impl.predict(data)
+        return self._predict_fn(data)
 
     def __eq__(self, other):
         if not isinstance(other, PyFuncModel):
@@ -714,7 +716,9 @@ def _warn_dependency_requirement_mismatches(model_path):
 
 
 def load_model(
-    model_uri: str, suppress_warnings: bool = False, dst_path: str = None
+    model_uri: str,
+    suppress_warnings: bool = False,
+    dst_path: str = None,
 ) -> PyFuncModel:
     """
     Load a model stored in Python function format.
@@ -759,7 +763,10 @@ def load_model(
     _add_code_from_conf_to_system_path(local_path, conf, code_key=CODE)
     data_path = os.path.join(local_path, conf[DATA]) if (DATA in conf) else local_path
     model_impl = importlib.import_module(conf[MAIN])._load_pyfunc(data_path)
-    return PyFuncModel(model_meta=model_meta, model_impl=model_impl)
+    pyfunc_predict_func = conf.get("pyfunc_predict_func", "predict")
+    return PyFuncModel(
+        model_meta=model_meta, model_impl=model_impl, pyfunc_predict_func=pyfunc_predict_func
+    )
 
 
 def _download_model_conda_env(model_uri):
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -624,7 +624,7 @@ class PyFuncModel:
     ``model_meta`` contains model metadata loaded from the MLmodel file.
     """
 
-    def __init__(self, model_meta: Model, model_impl: Any, pyfunc_predict_func: str = "predict"):
+    def __init__(self, model_meta: Model, model_impl: Any, predict_fn: str = "predict"):
         if not hasattr(model_impl, "predict"):
             raise MlflowException("Model implementation is missing required predict method.")
         if not model_meta:
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -625,14 +625,13 @@ class PyFuncModel:
     """
 
     def __init__(self, model_meta: Model, model_impl: Any, predict_fn: str = "predict"):
-        if not hasattr(model_impl, "predict"):
-            raise MlflowException("Model implementation is missing required predict method.")
+        if not hasattr(model_impl, predict_fn):
+            raise MlflowException(f"Model implementation is missing required {predict_fn} method.")
         if not model_meta:
             raise MlflowException("Model is missing metadata.")
         self._model_meta = model_meta
         self._model_impl = model_impl
-        assert hasattr(model_impl, pyfunc_predict_func)
-        self._predict_fn = getattr(model_impl, pyfunc_predict_func)
+        self._predict_fn = getattr(model_impl, predict_fn)
 
     def predict(self, data: PyFuncInput) -> PyFuncOutput:
         """
@@ -763,10 +762,8 @@ def load_model(
     _add_code_from_conf_to_system_path(local_path, conf, code_key=CODE)
     data_path = os.path.join(local_path, conf[DATA]) if (DATA in conf) else local_path
     model_impl = importlib.import_module(conf[MAIN])._load_pyfunc(data_path)
-    pyfunc_predict_func = conf.get("pyfunc_predict_func", "predict")
-    return PyFuncModel(
-        model_meta=model_meta, model_impl=model_impl, pyfunc_predict_func=pyfunc_predict_func
-    )
+    predict_fn = conf.get("predict_fn", "predict")
+    return PyFuncModel(model_meta=model_meta, model_impl=model_impl, predict_fn=predict_fn)
 
 
 def _download_model_conda_env(model_uri):
--- a/mlflow/sklearn/__init__.py
+++ b/mlflow/sklearn/__init__.py
@@ -152,6 +152,7 @@ def save_model(
     input_example: ModelInputExample = None,
     pip_requirements=None,
     extra_pip_requirements=None,
+    pyfunc_predict_func="predict",
 ):
     """
     Save a scikit-learn model to a path on the local file system. Produces an MLflow Model
@@ -195,6 +196,7 @@ def save_model(
                           base64-encoded.
     :param pip_requirements: {{ pip_requirements }}
     :param extra_pip_requirements: {{ extra_pip_requirements }}
+    :param pyfunc_predict_fun: the name of the prediction function to use; e.g. 'predict_proba'.
 
     .. code-block:: python
         :caption: Example
@@ -254,14 +256,15 @@ def save_model(
         serialization_format=serialization_format,
     )
 
-    # `PyFuncModel` only works for sklearn models that define `predict()`.
-    if hasattr(sk_model, "predict"):
+    # `PyFuncModel` only works for sklearn models that define a predict function
+    if hasattr(sk_model, pyfunc_predict_func):
         pyfunc.add_to_model(
             mlflow_model,
             loader_module="mlflow.sklearn",
             model_path=model_data_subpath,
             env=_CONDA_ENV_FILE_NAME,
             code=code_path_subdir,
+            pyfunc_predict_func=pyfunc_predict_func,
         )
     mlflow_model.add_flavor(
         FLAVOR_NAME,
@@ -320,6 +323,7 @@ def log_model(
     await_registration_for=DEFAULT_AWAIT_MAX_SLEEP_SECONDS,
     pip_requirements=None,
     extra_pip_requirements=None,
+    pyfunc_predict_func="predict",
 ):
     """
     Log a scikit-learn model as an MLflow artifact for the current run. Produces an MLflow Model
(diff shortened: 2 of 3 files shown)


mlflow/sklearn/__init__.py
```diff
@@ -152,6 +152,7 @@ def save_model(
     input_example: ModelInputExample = None,
     pip_requirements=None,
     extra_pip_requirements=None,
+    pyfunc_predict_func="predict",
```
dbczumar, line 155: Nit: Can we use `fn` instead of `func`?

```suggestion
    pyfunc_predict_fn="predict",
```

mlflow/sklearn/__init__.py
```diff
@@ -195,6 +196,7 @@ def save_model(
                           base64-encoded.
     :param pip_requirements: {{ pip_requirements }}
     :param extra_pip_requirements: {{ extra_pip_requirements }}
+    :param pyfunc_predict_fun: the name of the prediction function to use; e.g. 'predict_proba'.
```
dbczumar, line 199: Nit:

```suggestion
    :param pyfunc_predict_fn: The name of the prediction function to use for inference with the pyfunc representation of the resulting MLflow Model; e.g. ``"predict_proba"``.
```

mlflow/sklearn/__init__.py
```diff
@@ -368,6 +372,7 @@ def log_model(
                             waits for five minutes. Specify 0 or None to skip waiting.
     :param pip_requirements: {{ pip_requirements }}
     :param extra_pip_requirements: {{ extra_pip_requirements }}
+    :param pyfunc_predict_func: the prediction function to use for the pyfunc flavor
```
dbczumar, line 375: Nit: can we make this docstring consistent with the one in `mlflow.sklearn.save_model()`?

mlflow/pyfunc/__init__.py
```diff
@@ -624,13 +624,15 @@ class PyFuncModel:
     ``model_meta`` contains model metadata loaded from the MLmodel file.
     """
 
-    def __init__(self, model_meta: Model, model_impl: Any):
+    def __init__(self, model_meta: Model, model_impl: Any, pyfunc_predict_func: str = "predict"):
```
dbczumar, line 627:
```suggestion
    def __init__(self, model_meta: Model, model_impl: Any, predict_fn: str = "predict"):
```

mlflow/pyfunc/__init__.py
```diff
@@ -624,13 +624,15 @@ class PyFuncModel:
...
         if not hasattr(model_impl, "predict"):
             raise MlflowException("Model implementation is missing required predict method.")
         if not model_meta:
             raise MlflowException("Model is missing metadata.")
         self._model_meta = model_meta
         self._model_impl = model_impl
+        assert hasattr(model_impl, pyfunc_predict_func)
```
dbczumar, line 634: Can we `raise MlflowException` if the predict function isn't available instead of `assert`?

tests/sklearn/test_sklearn_model_export.py
```diff
@@ -617,3 +617,15 @@ def test_log_model_with_code_paths(sklearn_knn_model):
...
+def test_log_predict_proba(sklearn_logreg_model):
+    model, inference_dataframe = sklearn_logreg_model
+    artifact_path = "model"
+    with mlflow.start_run():
+        mlflow.sklearn.log_model(model, artifact_path, pyfunc_predict_func="predict_proba")
+        model_uri = mlflow.get_artifact_uri(artifact_path)
+
+    loaded_model = pyfunc.load_model(model_uri)
+    scores = loaded_model.predict(inference_dataframe)
+    assert scores.shape[1] == 3
```
dbczumar, line 631: Can we assert that the predictions are equivalent to those produced by `sklearn_logreg_model.predict_proba()`?

mlflow/sklearn/__init__.py
```diff
@@ -254,14 +256,15 @@ def save_model(
...
-    if hasattr(sk_model, "predict"):
+    # `PyFuncModel` only works for sklearn models that define a predict function
+    if hasattr(sk_model, pyfunc_predict_func):
         pyfunc.add_to_model(
             mlflow_model,
             loader_module="mlflow.sklearn",
             model_path=model_data_subpath,
             env=_CONDA_ENV_FILE_NAME,
             code=code_path_subdir,
+            pyfunc_predict_func=pyfunc_predict_func,
```
dbczumar, line 267:
```suggestion
            predict_fn=pyfunc_predict_func,
```

dbczumar (commented):
@skylarbpayne This is fantastic and long overdue! I just left a few small nits around docs & naming. Should be all set to merge once they're addressed. Thank you for your contribution!

WeichenXu123:
Thanks for the PR !

From my side, I think the ideal solution should be , we add more methods to the PyFuncModel base class, like:

```
class PyFuncModel:

  def predict(..)

  # only available for probability based classifier, otherwise raise unsupported error
  def predict_proba(...)

  # only available for classifier, otherwise raise unsupported error
  def num_classes(...)

```

WDYT ? @dbczumar

skylarbpayne:
> Thanks for the PR !
> 
> From my side, I think the ideal solution should be , we add more methods to the PyFuncModel base class, like:
> 
> ```
> class PyFuncModel:
> 
>   def predict(..)
> 
>   # only available for probability based classifier, otherwise raise unsupported error
>   def predict_proba(...)
> 
>   # only available for classifier, otherwise raise unsupported error
>   def num_classes(...)
> ```
> 
> WDYT ? @dbczumar
> 
> And NOTE that in mlflow `DefaultEvaluator`, it calls `predict` function to compute metrics / generating plots, and assuming "predict" function returning the prediction result, NOT the prob result. If user custom PyFuncModel override "predict" but implement it as "predict_proba", it might cause issues in `DefaultEvaluator`.

Thanks for raising this! How many model flavors have a corresponding `predict_proba`? I think `keras` and `scikit-learn`, but I'm not aware of others. E.g. not `pytorch` to my knowledge. If corresponding `predict_proba` is not common, then it seems already the case that `DefaultEvaluator` might not be broadly usable across flavors anyway, and I wonder whether there's a fix for that or a different `Evaluator` we need?

Regardless, I like the idea of including multiple predict functions (even if I don't like the idea of more work 😅). IMHO, I think such a solution probably warrants a bit more design. In particular, I'm curious about:

- What is the supported set of prediction functions? (e.g. some scikit-learn models also have `decision_function`)
- What are the interfaces of these methods? What further constraints should be expected by users?
- How should users make these available, and in which instances are they available?
- How can we ensure that the runtime loading the model can determine which methods are available? (e.g. with regression models you probably won't have `predict_proba`)

I'll wait for @dbczumar 's thoughts to address any comments as it seems like this thread may change the direction of the solution. Thanks!

mlflow/sklearn/__init__.py
```diff
@@ -254,14 +257,15 @@ def save_model(
         serialization_format=serialization_format,
     )
 
-    # `PyFuncModel` only works for sklearn models that define `predict()`.
-    if hasattr(sk_model, "predict"):
+    # `PyFuncModel` only works for sklearn models that define a predict function
+    if hasattr(sk_model, pyfunc_predict_fn):
```
dbczumar, line 261: Can we log a warning if the specified predict function isn't present on the model indicating that the pyfunc model flavor won't be added to the resulting model?

dbczumar (approved):
LGTM once a small `_logger.warning()` call is introduced (https://github.com/mlflow/mlflow/pull/6631/files#r958772687). Thanks so much, @skylarbpayne !

mlflow/sklearn/__init__.py
skylarbpayne: Great suggestion! Done!

But maybe another wrench here... 😅. The straightforward way to add a warning log, like I did, will mean that warnings go off for what is possibly legitimate code today (though I'm not familiar with scikit-learn models without a predict function). This code structure makes me think that perhaps what we want is something like:

```
def save_model(..., pyfunc_predict_fn=None):
  ...
  # When
  if not pyfunc_predict_fn and hasattr(sk_model, 'predict'):
    pyfunc_predict_fn = 'predict'

  if pyfunc_predict_fn:
    if not hasattr(sk_model, pyfunc_predict_fn):
      _logger.warning(f'Model was missing {pyfunct_predict_fn} function!')
    else:
      
```

It looks uglier to me (perhaps better way to write it?) But would avoid warnings going off for what is today legitimate code. I'm not familiar enough with scikit-learn models w/o `predict` methods to know if this is worth doing -- so feel free to merge if not. But let me know if it is and I'm happy to comply :)

skylarbpayne:
> LGTM once a small `_logger.warning()` call is introduced (https://github.com/mlflow/mlflow/pull/6631/files#r958772687). Thanks so much, @skylarbpayne !

done, thanks for the swift reviews!

mlflow/sklearn/__init__.py
dbczumar: I think that adding warning for the existing case(s) where users are logging models that don't define `predict()` is just fine :D. We can always revisit this if necessary. Thanks for considering this!

pbower:
Hi all, 

Could I please confirm how we use this change to use predict_proba but still have it integrate with the serving tools? 

The API reference mentions predict_proba, but for e.g. if we are using predict proba and subsequently thresholding this to produce class predictions, can this integrate with model serving ?

I feel the documentation on this change is potentially unclear, which led to google and then to here.

Please see https://www.mlflow.org/docs/latest/python_api/mlflow.sklearn.html

Thanks

pbower:
Hi, I note this change had 'documentation update required: no" - however should this page be updated? I found it unclear for a use case where I am looking to threshold the probabilities to produce the final decision, in the serving tool.

https://www.mlflow.org/docs/latest/python_api/mlflow.sklearn.html

RooieRakkert:
I've stumbled upon this thread after facing issues using the predict_proba of an (already logged to MLflow) model. 

I've noticed that the  source code for `load_model` [here](https://mlflow.org/docs/latest/_modules/mlflow/pyfunc.html#load_model) tries to find a `predict_fn` within its `conf` (which I assume is the model metadata configuration for the specified FLAVOR_NAME). 

However, I did not find any documentation on where I can actually **_set_** the `predict_fn` during the saving of the model to MLflow. Neither the `mlflow.pyfunc.log_model()` or `mlflow.pyfunc.save_model()` has any argument that seems to hold this configuration.

So my question comes down to; how do I store a custom model to MLflow, whilst also passing the appropriate `predict_fn`, in order to use any other function than the default `.predict()` after loading the model?

skylarbpayne:
Hey @RooieRakkert , thanks for the question.

When I put together this PR, it appears I neglected to provide a straight-forward path for your use case. I think we would similarly need to add `predict_fn` as an argument to `mlflow.pyfunc.log_model`. It is available when you call `mlflow.pyfunc.add_to_model`. So I think this would need to be added to `mlflow.pyfunc.log_model`

I'd be a bit curious to learn a little more about your use case though. In the case that you are using the direct form of `mlflow.pyfunc.log_model`, I would think you have a custom class, where you control which method is "predict" and perhaps have less of a problem, so I may be missing something.

Could you share a little more detail about the object you are trying to log?

RooieRakkert:
Yes you're right. 
Basically our usecase is as follows:

We've created a custom class:


```
class CustomModelWrapper(mlflow.pyfunc.PythonModel):
    """ Wrapper class to store arbitrary model to mlflow""""
    def __init__ (self, model):
        self.model = model
        
    def predict(self, context, model_input: ArrayLike) -> ArrayLike:
        self.model.predict(model_input)
        
    def predict_proba(self, context, model_input:ArrayLike) -> ArrayLike:
      self.model.predict_proba(model_input)
```

We then (after some HyperOpt GridSearch) store our model pipeline (which is an SKlearn pipeline object with an estimator as its final step) with:
```
mlflow.pyfunc.log_model('model_name', python_model=our_model_instance, conda_env=... signature=...)
```

Loading the model is done via:
```
...
model_uri = f"models:/{model_name}/{version}"
model = mlflow.pyfunc.load_model(model_uri=model_uri, dst_path=dst_path)
````


So yes, our current work-around is to overwrite the `.predict()` method of the `CustomModelWrapper` to return `self.model.predict_proba(model_input)`, but I think it would be nicer if we can still have our custom class, which will then have both support for a `.predict()` and `.predict_proba()` method, and then have our MLflow Model specification determine which of the predict- methods we will use by default, after loading up our model.

skylarbpayne:
Ah yes, that makes sense! I am not a maintainer of mlflow, so might make sense to open another issue to assess whether the maintainers are comfortable with the change; I'd be surprised if they weren't. The change isn't particularly "hard", but unfortunately there's multiple places to touch...

In case you or someone else wants to implement it, I think what effectively what needs to happen:

- Add the `predict_fn` argument to [pyfunc.log_model](https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/__init__.py#L1492)
- Pass `predict_fn` [through to `Model.log`](https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/__init__.py#L1615)

From here, `Model.log` will eventually call back into `pyfunc.save_model`and you'll want to thread the `predict_fn` arg down:

- Add `predict_fn` to [pyfunc.save_model](https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/__init__.py#L1312); pay attention to the kwargs here, the maintainers may have particular opinions about how to represent new args (e.g. might want you to just parse out of kwargs instead of making explicit arg?)
- Thread `predict_fn` through the [two](https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/__init__.py#L1468) [paths](https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/__init__.py#L1479) to save the model, all the way through to the calls to `pyfunc.add_to_model` ... [for example here](https://github.com/mlflow/mlflow/blob/93ee3525205ab217db7261552d3190d4bf02f371/mlflow/pyfunc/model.py#L203)

Hope that's helpful!

PR #6662, opened 2022-08-31
Creating a data profiler to profile data frames.

sunishsheth2009:
Creating a data profiler to profile data frames.

## How is this patch tested?
- [x] Running 
```
from mlflow.pipelines.cards import pandas_renderer
pandas_renderer.render_html(df)
``` 
with different df.

Components

Interface

Language

Integrations


mlflow/pipelines/cards/histogram_generator.py
```diff
@@ -0,0 +1,124 @@
...
+    be ordered and corresponding to equal distant percentiles.
+
+    :param quantiles: The quantiles that capture the frequency distribution.
+    :param num_buckets: The number of buckets in the generated equal height histogram.
+    :return: An equal height histogram or None if inputs are invalid.
+    """
+    if (
+        (len(quantiles) == 0)
+        or (len(quantiles) < 3)
+        or (len(quantiles) % 2 == 0)
```
dbczumar, line 19: Why does the length of quantiles need to be even? E.g. is 3 buckets demarcated by 4 quantiles invalid?

mlflow/pipelines/cards/histogram_generator.py
```diff
@@ -0,0 +1,124 @@
...
+    """Generates the equal height histogram from the input quantiles. The quantiles are assumed to
+    be ordered and corresponding to equal distant percentiles.
+
+    :param quantiles: The quantiles that capture the frequency distribution.
+    :param num_buckets: The number of buckets in the generated equal height histogram.
+    :return: An equal height histogram or None if inputs are invalid.
+    """
+    if (
+        (len(quantiles) == 0)
+        or (len(quantiles) < 3)
```
dbczumar, line 18: Nit: `len(quantiles) == 0` is redundant here

mlflow/pipelines/cards/histogram_generator.py
dbczumar: (I realize we're pulling this in from Databricks, but I'd still like to understand better if possible :D)

tests/pipelines/cards/test_pandas_renderer.py
```diff
@@ -0,0 +1,30 @@
...
+Test to convert data from tabular format to DatasetFeatureStatisticsList proto.
+"""
+import numpy as np
+
+from facets_overview import feature_statistics_pb2
+from mlflow.pipelines.cards import pandas_renderer
+
+# _NUMERIC_TYPES = ("tinyint", "smallint", "int", "bigint", "float", "double", "decimal(10,0)")
+# _TIME_TYPES = ("timestamp", "date")
+# _STRUCT_TYPES = ("array<bigint>", "map<string,bigint>", "struct<col1:int,col2:int>")
```
dbczumar, line 11: Can we remove commented code?

mlflow/pipelines/cards/histogram_generator.py
sunishsheth2009: Spoke to Shupeng Sun. It seems like we can remove them and ease the restrictions. 

We still want to keep `(len(quantiles) < 3)` since we ideally don't want to just show 1 bucket. So still keeping that condition.

tests/pipelines/cards/test_pandas_renderer.py
sunishsheth2009: Yup still working on the test

mlflow/protos/feature_statistics.proto
```diff
@@ -0,0 +1,332 @@
...
+// ==============================================================================
+
+// Adapted from https://github.com/PAIR-code/facets/blob/master/facets_overview/proto/feature_statistics.proto
+// - Ported from proto3 to proto2 by adding "optional" modifiers to relevant fields
+// - unsigned int64 types changed to int64 types
+
+// Definitions for aggregated feature statistics for datasets.
+package featureStatistics;
+
+option java_package = "com.databricks.api.proto.featureStatistics";
```
dbczumar, line 25: Can we change the package name here? Maybe reuse the facets package name?

mlflow/pipelines/cards/pandas_renderer.py
```diff
@@ -0,0 +1,244 @@
...
+            feat_stats = feat.string_stats
+            strs = []
+            compute_unique_str = current_column_value.dropna()
+            for item in compute_unique_str:
+                strs.append(
+                    item
+                    if hasattr(item, "__len__")
+                    else item.encode("utf-8")
+                    if hasattr(item, "encode")
+                    else str(item)
```
jinzhang21, line 142: Will these 2 lines ever be called?

mlflow/pipelines/cards/pandas_renderer.py
```diff
@@ -0,0 +1,244 @@
...
+    :param column: A column from a DataFrame.
+    :return: A CommonStatistics proto.
+    """
+    common_stats = facet_feature_statistics_pb2.CommonStatistics()
+    common_stats.num_missing = column.isnull().sum()
+    common_stats.num_non_missing = len(column) - common_stats.num_missing
+    # Default set using: https://src.dev.databricks.com/databricks/universe/-/blob/model-monitoring/python/databricks/model_monitoring/rendering/html_renderer.py?L33-L35&subtree=true
+    common_stats.min_num_values = 1
+    common_stats.max_num_values = 1
+    common_stats.avg_num_values = 1.0
```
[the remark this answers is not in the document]
sunishsheth2009: I went through this code here: 
https://github.com/PAIR-code/facets/blob/4742b8b93c2dacf22fc8ace2cee42dd06382c48e/facets_overview/facets_overview/base_generic_feature_statistics_generator.py#L290-L304

And understanding this code I realized that these are the values is what it was computed to. 

And then I realized that this is also done exactly the same way here: https://src.dev.databricks.com/databricks/universe/-/blob/model-monitoring/python/databricks/model_monitoring/rendering/html_renderer.py?L33-L35&subtree=true

Do you think we should replicate what facet-overview is doing or it is not required? It seems to be working for our databricks profiling as well.

mlflow/pipelines/cards/pandas_renderer.py
sunishsheth2009: Got this code from here: https://github.com/PAIR-code/facets/blob/4742b8b93c2dacf22fc8ace2cee42dd06382c48e/facets_overview/facets_overview/base_generic_feature_statistics_generator.py#L263-L265
I assumed that this might be required for an edge case when the strings are not encoded. So left this code path here. 

Do you think we need to support this usecase? I am happy to remove it

tests/pipelines/cards/test_pandas_renderer.py
```diff
@@ -0,0 +1,250 @@
...
+                  high_rank: 2
+                  label: "NFLX"
+                  sample_count: 1.0
+                }
+              }
+            }
+          }
+          features {
+            name: "Shares"
+            type: FLOAT
```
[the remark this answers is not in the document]
sunishsheth2009: Here is what I found:
``` 
0    100
1    170
2    150
3    200
4    123
Name: Shares, dtype: int64
```

And 
```
0    100.0
1    170.0
2    150.0
3    200.0
4      NaN
Name: Shares, dtype: float64
```

So the None value in the column makes it a float.

mlflow/pipelines/cards/pandas_renderer.py
```diff
@@ -0,0 +1,244 @@
...
+
+from typing import Union, Iterable, Tuple
+from mlflow.protos import facet_feature_statistics_pb2
+from mlflow.pipelines.cards import histogram_generator
+
+
+def DtypeToType(dtype):
+    """Converts a Numpy dtype to the FeatureNameStatistics.Type proto enum."""
+    fs_proto = facet_feature_statistics_pb2.FeatureNameStatistics
+    if dtype.char in np.typecodes["AllFloat"]:
```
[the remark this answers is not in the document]
sunishsheth2009: Good catch.
Checked how this is handled by databricks profiling and facet_oveview as well. 
They both throw a TypeError that the `Found type complex, but expected one of: int, long, float`
We could replicate that. What do you think?

tests/pipelines/cards/test_pandas_renderer.py
```diff
@@ -0,0 +1,250 @@
...
+from mlflow.pipelines.cards import pandas_renderer
+
+
+def test_convert_to_html():
+    proto = facet_feature_statistics_pb2.DatasetFeatureStatisticsList()
+    html = pandas_renderer.construct_facets_html(proto)
+    assert len(html) != 0
+
+
+def testDTypeToType():
```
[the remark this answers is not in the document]
sunishsheth2009: Done

mlflow/pipelines/cards/pandas_renderer.py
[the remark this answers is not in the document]
jinzhang21: I think we should match the Facet implementation. Setting them to 1 is not the correct approach.

mlflow/pipelines/cards/pandas_renderer.py
jinzhang21: No I'm just questioning the usage of 2 consecutive if-else clauses. I doubt the 2nd one will ever get called. In what circumstances will it happen? It'd be great to demonstrate with an example.

mlflow/pipelines/cards/pandas_renderer.py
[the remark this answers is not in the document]
jinzhang21: SGTM. Stats really is all about real numbers anyway.

tests/pipelines/cards/test_pandas_renderer.py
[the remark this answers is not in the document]
jinzhang21: Got it. That's probably fine.

mlflow/protos/facet_feature_statistics.proto
```diff
@@ -0,0 +1,329 @@
+syntax = "proto2";
```
jinzhang21, line 1: Can we add a comment to link explicitly to the GH commit to this file we copy from Facet? It'd be helpful to debug in the future if Facet introduces some backward in compatible changes.

mlflow/pipelines/cards/pandas_renderer.py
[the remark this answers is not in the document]
sunishsheth2009: Spoke offline to add a TODO to support multi-dimensional columns later on.

jinzhang21 (approved):

PR #6705, opened 2022-09-06
Fix shap plot feature name de-duplication

WeichenXu123:

## Related Issues/PRs

#xxx

Fix shap plot feature name de-duplication.

Before:
It always append suffix `(f_{feature_index})` on truncated names.

After:
Only append suffix `({dup_index}` on duplicated names. 

## How is this patch tested?
Unit tests.

Before:
It always append suffix (f_{feature_index}) on truncated names.

After:
Only append suffix ({dup_index} on duplicated names.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/utils/string_utils.py
+++ b/mlflow/utils/string_utils.py
@@ -29,3 +29,20 @@ def truncate_str_from_middle(s, max_length):
         left_part_len = (max_length - 3) // 2
         right_part_len = max_length - 3 - left_part_len
         return f"{s[:left_part_len]}...{s[-right_part_len:]}"
+
+
+def dedup_string_list(str_list):
+    """
+    De-duplicate a list of strings. For duplicated strings, add suffix such as (1), (2), etc.
+    """
+    count_dict = {}
+    dedup_list = []
+    for s in str_list:
+        if s not in count_dict:
+            count_dict[s] = 1
+            new_s = s
+        else:
+            count_dict[s] += 1
+            new_s = f"{s}({count_dict[s]})"
+        dedup_list.append(new_s)
+    return dedup_list
--- a/mlflow/models/evaluation/default_evaluator.py
+++ b/mlflow/models/evaluation/default_evaluator.py
@@ -9,7 +9,7 @@ from mlflow.models.evaluation.base import (
 from mlflow.entities.metric import Metric
 from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
 from mlflow.utils.file_utils import TempDir
-from mlflow.utils.string_utils import truncate_str_from_middle
+from mlflow.utils.string_utils import dedup_string_list, truncate_str_from_middle
 from mlflow.models.utils import plot_lines
 from mlflow.models.evaluation.artifacts import (
     ImageEvaluationArtifact,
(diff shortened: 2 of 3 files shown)


mlflow/utils/string_utils.py
```diff
@@ -29,3 +29,20 @@ def truncate_str_from_middle(s, max_length):
         left_part_len = (max_length - 3) // 2
         right_part_len = max_length - 3 - left_part_len
         return f"{s[:left_part_len]}...{s[-right_part_len:]}"
+
+
+def dedup_string_list(str_list):
+    """
+    De-duplicate a list of strings. For duplicated strings, add suffix such as (1), (2), etc.
```
harupy, line 36: This docstring indicates this function converts `["xb", "xb"]` to `["xb(1)", "xb(2)"]`, but it doesn't. Can you fix it?

harupy: I'd prefer `["xb(1)", "xb(2)"]`

WeichenXu123:
Discussion:
I am rethinking whether the de-dup logic in this PR is good. e.g., let's see the following example (applying previous de-dup logic) :
shap_beeswarm_plot_on_data_diabetes_dataset

Note the feature display order is changed on the plot. If using the de-dup logic in this PR, seemingly we cannot distinguish the original feature well. But if we append suffix with the exact "feature_index", it is easier for user to distinguish them.

mlflow/utils/string_utils.py
WeichenXu123: Considering https://github.com/mlflow/mlflow/pull/6705#issuecomment-1237618055

I propose in this PR to change the suffix to be `({f_{feature_index})` instead of `({dup_index})`

harupy:
> But if we append suffix with the exact "feature_index", it is easier for user to distinguish them.

If users understand what `f_1` means.

harupy:
btw do we really need to truncate the feature names?

WeichenXu123:
> btw do we really need to truncate the feature names?

If we don't truncate feature names, too long feature name might cause the plot layout broken.

harupy:
> If we don't truncate feature names, too long feature name might cause the plot layout broken.

Can we create a function that creates the plot so that we can control the layout?

WeichenXu123:
> > If we don't truncate feature names, too long feature name might cause the plot layout broken.
> 
> Can we create a function that creates the plot so that we can control the layout?

Then we might need to copy and modify shap plot generation code, which makes maintenance harder.
And for long feature names, without truncation we cannot show it in plot well, we have to either truncate it or make font size smaller.

harupy:
> And for long feature names, without truncation we cannot show it in plot well

```python
import xgboost
import shap
from textwrap import wrap

# train an XGBoost model
X, y = shap.datasets.boston()
X.columns = ["_".join((c,) * 10) for c in X.columns]
model = xgboost.XGBRegressor().fit(X, y)

explainer = shap.Explainer(model)
shap_values = explainer(X)

shap.plots.beeswarm(shap_values, show=False)
plt.xticks(fontsize=10)
plt.yticks(fontsize=10)
plt.savefig("beeswarm.png", bbox_inches="tight")
```

This code generates the plot below:

image

harupy:
More realistic example:

image

dsgibbons:
> @WeichenXu123
> 
> > And for long feature names, without truncation we cannot show it in plot well
> 
> ```python
> import xgboost
> import shap
> import matplotlib.pyplot as plt
> 
> # train an XGBoost model
> X, y = shap.datasets.boston()
> X.columns = ["_".join((c,) * 10) for c in X.columns]
> model = xgboost.XGBRegressor().fit(X, y)
> 
> explainer = shap.Explainer(model)
> shap_values = explainer(X)
> 
> shap.plots.beeswarm(shap_values, show=False)
> plt.title(f"Maximum feature name length: {len(max(X.columns, key=len))}")
> plt.xticks(fontsize=10)
> plt.yticks(fontsize=10)
> plt.savefig("beeswarm.png", bbox_inches="tight")
> ```
> 
> This code generates the plot below:
> 
> image
> 
> The layout doesn't seem broken. The image might be a bit too wide, but I think it's much better than truncated feature names. What do you think?
> 
> @dsgibbons Any thoughts?

LGTM!

PR #6680, opened 2022-09-01
hyperparameter tuning scaffolding and search space

prithvikannan:
## Related Issues/PRs

https://github.com/mlflow/mlflow/pull/6649

## What changes are proposed in this pull request?

Begin to handle the HP tuning confs during the train step. In this PR, we only construct the `hyperopt` search space from the confs. We will fill in the `hyperopt` objective function in a future PR. 

This HP tuning code will only be executed if the user has specified that `tuning enabled: True` and `hyperopt` is installed.

## How is this patch tested?

- [x] I have written tests (not required for typo or doc fix) and confirmed the proposed feature/bug-fix/change works.

## Does this PR change the documentation?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure the changed pages / sections render correctly by following the steps below.

1. Click the `Details` link on the `Preview docs` check.
2. Find the changed pages / sections and make sure they render correctly.

### Is this a user-facing change?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [x] `area/pipelines`: Pipelines, Pipeline APIs, Pipeline configs, Pipeline Templates
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface
- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language
- [ ] `language/r`: R APIs and clients
[...]

PR #6724, opened 2022-09-07
hyperparameter tuning 

prithvikannan:
## Related Issues/PRs

building on https://github.com/mlflow/mlflow/pull/6680, blocked by https://github.com/mlflow/mlp-regression-template/pull/75

Filling in the `objective()` function in HP tuning and handle hardcoded parameters. Some small refactors to reduce code repetition

## How is this patch tested?

Manual testing using mlp-regression-template repo and a source mlflow installation.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/steps/train.py
+++ b/mlflow/pipelines/steps/train.py
@@ -144,14 +144,13 @@ class TrainStep(BaseStep):
                     with mlflow.start_run(tags=tags, nested=True):
                         # create unfitted estimator from yaml
                         estimator = estimator_fn(args)
-                        # TODO: sample training data
-                        sample_fraction = (  # pylint: disable=unused-variable
-                            tuning_params["sample_fraction"]
+                        sample_fraction = (
+                            float(tuning_params["sample_fraction"])
                             if "sample_fraction" in tuning_params
-                            else 1
+                            else 1.0
                         )
-                        X_train_sampled = X_train
-                        y_train_sampled = y_train
+                        X_train_sampled = X_train.sample(frac=sample_fraction, random_state=42)
+                        y_train_sampled = y_train.sample(frac=sample_fraction, random_state=42)
                         # fit estimator to training
                         estimator.fit(X_train_sampled, y_train_sampled)
                         if hasattr(estimator, "best_score_P"):
--- a/mlflow/pipelines/steps/train.py
+++ b/mlflow/pipelines/steps/train.py
@@ -488,7 +488,7 @@ class TrainStep(BaseStep):
 
                 if "sample_fraction" in step_config["tuning"]:
                     sample_fraction = float(step_config["tuning"]["sample_fraction"])
-                    if sample_fraction >= 0 and sample_fraction <= 1.0:
+                    if sample_fraction > 0 and sample_fraction <= 1.0:
                         step_config["sample_fraction"] = sample_fraction
                     else:
                         raise MlflowException(
@@ -501,6 +501,18 @@ class TrainStep(BaseStep):
 
                 if "algorithm" not in step_config["tuning"]:
                     step_config["tuning"]["algorithm"] = "hyperopt.rand.suggest"
+
+                if "max_trials" not in step_config["tuning"]:
+                    raise MlflowException(
+                        "The 'max_trials' configuration in the train step must be provided.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
+
+                if "parameters" not in step_config["tuning"]:
+                    raise MlflowException(
+                        "The 'parameters' configuration in the train step must be provided.",
+                        error_code=INVALID_PARAMETER_VALUE,
+                    )
 
             else:
                 step_config["tuning_enabled"] = False
--- a/mlflow/pipelines/steps/train.py
+++ b/mlflow/pipelines/steps/train.py
@@ -138,76 +138,9 @@ class TrainStep(BaseStep):
         with mlflow.start_run(tags=tags) as run:
             estimator_hardcoded_params = self.step_config["estimator_params"]
             if self.step_config["tuning_enabled"]:
-                tuning_params = self.step_config["tuning"]
-                try:
-                    from hyperopt import fmin, Trials
-                except ModuleNotFoundError:
-                    raise MlflowException(
-                        "Hyperopt not installed and is required if tuning is enabled",
-                        error_code=BAD_REQUEST,
-                    )
-
-                # wrap training in objective fn
-                def objective(hyperparameter_args):
-                    with mlflow.start_run(tags=tags, nested=True):
-                        estimator = estimator_fn(
-                            dict(estimator_hardcoded_params, **hyperparameter_args)
-                        )
-
-                        sample_fraction = self.step_config["sample_fraction"]
-                        X_train_sampled = X_train.sample(frac=sample_fraction, random_state=42)
-                        y_train_sampled = y_train.sample(frac=sample_fraction, random_state=42)
-
-                        estimator.fit(X_train_sampled, y_train_sampled)
-
-                        logged_estimator = self._log_estimator_to_mlflow(estimator, X_train_sampled)
-
-                        eval_result = mlflow.evaluate(
-                            model=logged_estimator.model_uri,
-                            data=validation_df,
-                            targets=self.target_col,
-                            model_type="regressor",
-                            evaluators="default",
-                            dataset_name="validation",
-                            custom_metrics=_load_custom_metric_functions(
-                                self.pipeline_root,
-                                self.evaluation_metrics.values(),
-                            ),
-                            evaluator_config={
-                                "log_model_explainability": False,
-                            },
-                        )
-
-                        # return +/- metric
-                        sign = (
-                            -1
-                            if self.evaluation_metrics_greater_is_better[self.primary_metric]
-                            else 1
-                        )
-                        return sign * eval_result.metrics[self.primary_metric]
-
-                search_space = self._construct_search_space_from_yaml(tuning_params["parameters"])
-                algo_type, algo_name = tuning_params["algorithm"].rsplit(".", 1)
-                tuning_algo = getattr(importlib.import_module(algo_type, "hyperopt"), algo_name)
-                max_trials = tuning_params["max_trials"]
-                hp_trials = Trials()
-
-                best_hp_params = fmin(
-                    objective,
-                    search_space,
-                    algo=tuning_algo,
-                    max_evals=max_trials,
-                    trials=hp_trials,
+                estimator = self._tune_and_get_best_estimator(
+                    estimator_hardcoded_params, estimator_fn, X_train, y_train, validation_df
                 )
-                best_hp_estimator_loss = hp_trials.best_trial["result"]["loss"]
-                hardcoded_estimator_loss = objective({})
-
-                if best_hp_estimator_loss < hardcoded_estimator_loss:
-                    best_combined_params = dict(estimator_hardcoded_params, **best_hp_params)
-                else:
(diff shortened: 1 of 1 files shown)


mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
+                def objective(hyperparameter_args):
+                    # log as a child run
+                    with mlflow.start_run(tags=tags, nested=True):
+                        # create unfitted estimator from yaml
+                        estimator = estimator_fn(
+                            dict(estimator_hardcoded_params, **hyperparameter_args)
+                        )
+                        sample_fraction = (
+                            float(tuning_params["sample_fraction"])
+                            if "sample_fraction" in tuning_params
```
dbczumar, line 152: Do we validate that sample fraction is > 0 and <= 1?

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
+                            else 1.0
+                        )
+                        X_train_sampled = X_train.sample(frac=sample_fraction, random_state=42)
+                        y_train_sampled = y_train.sample(frac=sample_fraction, random_state=42)
+                        # fit estimator to training
+                        estimator.fit(X_train_sampled, y_train_sampled)
+                        if hasattr(estimator, "best_score_P"):
```
dbczumar, line 159:
```suggestion
                        if hasattr(estimator, "best_score_"):
```

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
+                            mlflow.log_metric("best_cv_score", estimator.best_score_)
+                        if hasattr(estimator, "best_params_"):
+                            mlflow.log_params(estimator.best_params_)
```
dbczumar, line 162: Is it ever possible for these best params / best score to be invalid (e.g. `None`)? Shall we validate them?

dbczumar: Ah, I see. This is duplicated from below. Can we factor it out?

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
+                                self.pipeline_root,
+                                self.evaluation_metrics.values(),
+                            ),
+                            evaluator_config={
+                                "log_model_explainability": False,
+                            },
+                        )
+
+                        # return +/- metric
+                        return eval_result.metrics[self.primary_metric]
```
dbczumar, line 192: Hyperopt needs to know whether the objective is to minimize or maximize this metric based on the "greater_is_better" value associated with the metric. Can we incorporate that into this PR? This needs to be supported for predefined metrics and custom metrics.

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
-                    param_details_to_pass.pop("distribution")
-                    search_space[param_name] = hp_tuning_fn(param_name, **param_details_to_pass)
-                else:
+        mlflow.autolog(log_models=False)
+        with mlflow.start_run(tags=tags) as run:
+            estimator_hardcoded_params = self.step_config["estimator_params"]
+            if self.step_config["tuning_enabled"]:
+                # gate all HP tuning code within this condition
+                tuning_params = self.step_config["tuning"]
+                # import hyperopt or throw error
```
dbczumar, line 133: Nit: can we remove this comment? Self-explanatory from the following lines

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
-                    hp_tuning_fn = getattr(hp, param_details["distribution"])
-                    param_details_to_pass = param_details.copy()
-                    param_details_to_pass.pop("distribution")
-                    search_space[param_name] = hp_tuning_fn(param_name, **param_details_to_pass)
-                else:
+        mlflow.autolog(log_models=False)
+        with mlflow.start_run(tags=tags) as run:
+            estimator_hardcoded_params = self.step_config["estimator_params"]
+            if self.step_config["tuning_enabled"]:
+                # gate all HP tuning code within this condition
```
dbczumar, line 131: Nit: can we remove this comment? Not particularly necessary

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
-                logged_estimator = mlflow.sklearn.log_model(
-                    estimator,
-                    f"{self.name}/estimator",
-                    signature=estimator_schema,
-                    code_paths=code_paths,
+                # wrap training in objective fn
+                def objective(hyperparameter_args):
+                    # log as a child run
+                    with mlflow.start_run(tags=tags, nested=True):
+                        # create unfitted estimator from yaml
```
dbczumar, line 146: Can we remove these comments unless the code that follows them is particularly hard to read / unintuitive?

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
+                        estimator = estimator_fn(
+                            dict(estimator_hardcoded_params, **hyperparameter_args)
+                        )
+                        sample_fraction = (
+                            float(tuning_params["sample_fraction"])
+                            if "sample_fraction" in tuning_params
+                            else 1.0
+                        )
+                        X_train_sampled = X_train.sample(frac=sample_fraction, random_state=42)
+                        y_train_sampled = y_train.sample(frac=sample_fraction, random_state=42)
```
dbczumar, line 156: Is this sampling stable such that it will always pick the same records given the same dataset? If not, we should investigate using the sampling logic from the MLP split step: https://github.com/mlflow/mlflow/blob/eb693368c00595b6d41f42759c6d2243430534f7/mlflow/pipelines/steps/split.py#L194-L195

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +124,181 @@ def _run(self, output_directory):
...
-                elif "distribution" in param_details:
-                    hp_tuning_fn = getattr(hp, param_details["distribution"])
-                    param_details_to_pass = param_details.copy()
-                    param_details_to_pass.pop("distribution")
-                    search_space[param_name] = hp_tuning_fn(param_name, **param_details_to_pass)
-                else:
+        mlflow.autolog(log_models=False)
+        with mlflow.start_run(tags=tags) as run:
+            estimator_hardcoded_params = self.step_config["estimator_params"]
+            if self.step_config["tuning_enabled"]:
```
dbczumar, line 130: As discussed offline, can we factor out the tuning logic into multiple helper methods that are invoked from within this branch and deduplicate some of the code that might be shared with the non-tuning branch?

mlflow/pipelines/steps/train.py
prithvikannan: @dbczumar looks like `pandas.DataFrame.sample` is stable
> If you pass it a RandomState it will use this (already initialized/seeded) RandomState to generate pseudo random numbers. This also allows you to get reproducible results by setting a fixed seed when initializing the RandomState and then passing this RandomState around.

source: https://stackoverflow.com/a/45213353

mlflow/pipelines/steps/train.py
prithvikannan: @dbczumar i'm not quite sure how to solve this. the only documentation i see about `greater_is_better` is in [sklearn.metrics.make_scorer](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.make_scorer.html), which i don't think is useful here.

what i'm looking for is something like `is_metric_greater_is_better(metric_name)`, as in [catboost](https://github.com/catboost/catboost/issues/1890), but i don't think this is supported in sklearn.

mlflow/pipelines/steps/train.py
dbczumar: Yay!

mlflow/pipelines/steps/train.py
dbczumar: Ah, the user specifies `greater_is_better` for custom metrics here: https://github.com/mlflow/mlp-regression-template/blob/74284ee13d907b0b427fd76d144986ac333d520f/pipeline.yaml#L85. We just need to read this value.

For builtin metrics, `greater_is_better` is defined here: https://github.com/mlflow/mlflow/blob/7219771141c7140fd68f6ed3a5b649e5ad1dab13/mlflow/pipelines/utils/metrics.py#L40-L44

mlflow/pipelines/steps/train.py
```diff
@@ -124,135 +125,149 @@ def _run(self, output_directory):
...
+
+                best_hp_params = fmin(
+                    objective,
+                    search_space,
+                    algo=tuning_algo,
+                    max_evals=max_trials,
+                    trials=hp_trials,
                 )
+                best_hp_estimator_loss = hp_trials.best_trial["result"]["loss"]
+                hardcoded_estimator_loss = objective({})
```
dbczumar, line 189: This won't work if the user-defined model returned by `estimator_fn` initializes model parameters using the args passed to `estimator_fn`. We should instead call `objective()` with the hardcoded parameter dict, rather than an empty dict.

mlflow/pipelines/steps/train.py
dbczumar: @prithvikannan Can we move the tuning logic into a separate method so that we can just call

```
if self.step_config["tuning_enabled"]:
    estimator = tune_and_get_best_estimator(...)
else:
    estimator = estimator_fn(estimator_hardcoded_params)
```

?

mlflow/pipelines/steps/train.py
```diff
@@ -548,3 +530,110 @@ def environment(self):
...
+                )
+
+                # return +/- metric
+                sign = -1 if self.evaluation_metrics_greater_is_better[self.primary_metric] else 1
+                return sign * eval_result.metrics[self.primary_metric]
+
+        search_space = self._construct_search_space_from_yaml(tuning_params["parameters"])
+        algo_type, algo_name = tuning_params["algorithm"].rsplit(".", 1)
+        tuning_algo = getattr(importlib.import_module(algo_type, "hyperopt"), algo_name)
+        max_trials = tuning_params["max_trials"]
```
dbczumar, line 582: Can we add validation in `from_pipeline_config` to ensure that these `parameters`, `algorithm`, and `max_trials` fields are defined? For algorithm, I think we can default to hyperopt random search, so that field can be optional.

mlflow/pipelines/steps/train.py
```diff
@@ -523,9 +485,29 @@ def from_pipeline_config(cls, pipeline_config, pipeline_root):
                         "'enabled' key whose value is either true or false.",
                         error_code=INVALID_PARAMETER_VALUE,
                     )
+
+                if "sample_fraction" in step_config["tuning"]:
+                    sample_fraction = float(step_config["tuning"]["sample_fraction"])
+                    if sample_fraction >= 0 and sample_fraction <= 1.0:
```
dbczumar, line 491:
```suggestion
                    if sample_fraction > 0 and sample_fraction <= 1.0:
```

dbczumar: Sample fraction of zero shouldn't be allowed

dbczumar (approved):
LGTM! Thanks @prithvikannan !

mlflow/pipelines/steps/train.py
prithvikannan: updated. already had `algorithm` set to hp random as default, but added validation for `parameters` and `max_trials`.

Issue #6093, opened 2022-06-17
[FR] [Roadmap] Publish official Docker image for MLflow Tracking server


BenWilson2:
## MLflow Roadmap Item

This is an MLflow Roadmap item that has been prioritized by the MLflow maintainers. We’ve identified this feature as a highly requested addition to the MLflow package based on community feedback. 

## Contribution Note

As with other roadmap items, there may be a desire for multiple contributors to work on an issue. While we don’t discourage collaboration, we strongly encourage that a primary contributor is assigned to roadmap issues to simplify the merging process. 
The items on the roadmap are of a high priority. Due to the wide-spread demand of roadmap features, we encourage potential contributors to only agree to take on the work of creating a PR, making changes, and ensuring that test coverage is adequately created for the feature if they are willing and able to see the implementation through to a merged state. 

## Feature scope

This roadmap feature’s complexity is classified as:

- [ ] `good-first-issue`: This feature is limited in complexity and effort required to implement.
- [ ] `simple`: This feature does not require a large amount of effort to implement and / or is clear enough to not need a design discussion with maintainers.
- [X] `involved`: This feature will require a substantial amount of development effort but does not require an agreed-upon design from the maintainers. The feedback given during the PR phase may be involved and necessitate multiple iterations before approval. (Please bear with us as we collaborate with you to make a great contribution)
- [ ] `design-recommended`: This is a substantial feature that should have a design document approved prior to working on an implementation (to save your time, not ours). After agreeing to work on this feature, a maintainer will be assigned to support you throughout the development process.

## Proposal Summary

Build and configure automated versioned release docker images for the MLflow tracking server.

Note: the triggered execution of generating these images and pushing them to a public container repository will need to be handled by MLflow maintainers. If you wish to work on this FR, there will be heavy involvement with us.

## Motivation
> #### What is the use case for this feature? 
  To greatly simplify the process of starting and configuring MLflow for users.

> #### Why is this use case valuable to support for MLflow users in general? ^

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [X] `area/build`: Build and test infrastructure for MLflow
- [X] `area/docs`: MLflow documentation pages
- [X] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
[...]

PR #6731, opened 2022-09-08
Tracking server image

oojo12:
## Related Issues/PRs
#6093  

This PR introduces a minimal Dockerfile for the MlFLow tracking server as well as a docker compose file that can be invoked for building and pushing all MlFlow docker images.

## How is this patch tested?
Requires maintainer testing to ensure the associated workflow PR that goes along with this properly automatically tags and pushes the image to GitHub container registry.

- [X ] Yes. Give a description of this change to be included in the release notes for MLflow users.

Introduction of Docker image for the MlFlow tracking server.

Components
- [X] `area/build`: Build and test infrastructure for MLflow

Interface
- [X] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,4 @@
 # example build command docker build . --build-arg TAG=v1.28.0
 FROM python:3.10.6-slim-bullseye
-ARG TAG
-RUN pip install --no-cache mlflow==$TAG
-ENTRYPOINT ["mlflow", "server"]
+ARG VERSION
+RUN pip install --no-cache mlflow==$VERSION
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,3 @@
-# example build command docker build . --build-arg TAG=v1.28.0
 FROM python:3.10.6-slim-bullseye
 ARG VERSION
 RUN pip install --no-cache mlflow==$VERSION
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,7 @@
+version: "3.9"  # optional since v1.27.0
+services:
+  tracking:
+    build: .
+    args:
+        VERSION: ${GITHUB_REF_NAME}
+    entrypoint: ["mlflow", "server"]
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,7 +1,0 @@
-version: "3.9"  # optional since v1.27.0
-services:
-  tracking:
-    build: .
-    args:
-        VERSION: ${GITHUB_REF_NAME}
-    entrypoint: ["mlflow", "server"]
--- a/mlflow/docker-compose.yml
+++ b/mlflow/docker-compose.yml
@@ -1,0 +1,6 @@
+version: "3.9"  # optional since v1.27.0
+services:
+  tracking:
+    build: ./tracking
+    args:
+        TAG: ${GITHUB_REF}
--- a/mlflow/docker-compose.yml
+++ b/mlflow/docker-compose.yml
@@ -1,6 +1,0 @@
-version: "3.9"  # optional since v1.27.0
-services:
-  tracking:
-    build: ./tracking
-    args:
-        TAG: ${GITHUB_REF}
--- a/mlflow/tracking/Dockerfile
+++ b/mlflow/tracking/Dockerfile
@@ -1,0 +1,7 @@
+# example build command docker build . --build-arg TAG=v1.28.0
+FROM python:3.10.6-slim-bullseye
+ARG TAG
+RUN echo $TAG 
+RUN apt-get update && apt-get install git -y && \
+	pip install git+https://github.com/mlflow/mlflow.git@$TAG
+ENTRYPOINT ["mlflow", "server"]
--- a/mlflow/tracking/Dockerfile
+++ b/mlflow/tracking/Dockerfile
@@ -1,7 +1,5 @@
 # example build command docker build . --build-arg TAG=v1.28.0
 FROM python:3.10.6-slim-bullseye
 ARG TAG
-RUN echo $TAG 
-RUN apt-get update && apt-get install git -y && \
-	pip install git+https://github.com/mlflow/mlflow.git@$TAG
+RUN pip install --no-cache mlflow==$TAG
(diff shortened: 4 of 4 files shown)


mlflow/tracking/Dockerfile
```diff
@@ -0,0 +1,7 @@
+# example build command docker build . --build-arg TAG=v1.28.0
+FROM python:3.10.6-slim-bullseye
+ARG TAG
+RUN echo $TAG 
+RUN apt-get update && apt-get install git -y && \
+	pip install git+https://github.com/mlflow/mlflow.git@$TAG
```
harupy, line 6: MLflow installed from GitHub doesn't contain UI assets. Can we install it from PyPI?

harupy, line 6:
```suggestion
	pip install --no-cache git+https://github.com/mlflow/mlflow.git@$TAG
```

to reduce the image size.

mlflow/tracking/Dockerfile
```diff
@@ -0,0 +1,7 @@
+# example build command docker build . --build-arg TAG=v1.28.0
```
harupy, line 1: Can we create a new directory in the repository root and move this file there?

mlflow/docker-compose.yml
```diff
@@ -0,0 +1,6 @@
+version: "3.9"  # optional since v1.27.0
+services:
+  tracking:
+    build: ./tracking
+    args:
+        TAG: ${GITHUB_REF}
```
harupy, line 6: Do we need this file? Can we just run `build . --build-arg TAG=${GITHUB_REF}`?

mlflow/tracking/Dockerfile
oojo12: yes I can make that change.

oojo12: addressed with Pypi commit

mlflow/tracking/Dockerfile
oojo12: Is there an organization benefit to this as opposed to leaving the Dockerfiles under the related components? I am not opposed to making this change I am just curious what the expected benefit is.

harupy: Does this Dockerfile interact with files in `mlflow/tracking`?

mlflow/docker-compose.yml
oojo12: No, the file is not necessary and was meant to simplify builds so we could use a single docker-compose build command to build all mlflow component images with the appropriate entrypoints preconfigured. However, I suppose MLFlow is a monolithic application so a single docker image could be made with the entrypoints and commands documented instead?

mlflow/tracking/Dockerfile
harupy: I'd put `Dockerfile` in `mlflow/tracking` if it was used in `mlflow/tracking`.

harupy: I think it's cleaner if we organize files like this:

```
- docker
  - Dockerfile
  - README.md
  - compose.yml
```

oojo12: we actually don't use any of the files in mlflow tracking. I will make this change.

mlflow/tracking/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+FROM python:3.10.6-slim-bullseye
+ARG TAG
+RUN pip install --no-cache mlflow==$TAG
```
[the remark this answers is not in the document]
oojo12: noted. Taking your strip code from [here](https://github.com/harupy/docker-image-test/blob/main/.github/workflows/push-images.yml)

mlflow/docker-compose.yml
oojo12: I am thinking with the adjustment to the workflow, your proposed directory structure, and to easily allow other MlFlow Docker images to be made that this makes sense.

your directory structure
```
- <repo root>
  - docker
    - BaseDockerfile
    - TrackingDockerfile
    - README.md
```

BaseDockerfile
```
ARG VERSION
FROM python:3.10.6-slim-bullseye
RUN pip install --no-cache mlflow==$VERSION
```
here version comes from the build args of the workflow

TrackingDockerfile
```
ARG VERSION
FROM ghcr.io/[USERNAME]/mlflow:$VERSION
ENTRYPOINT ["mlflow", "server"]
```

harupy: I think we can just use a single image for now.

oojo12: sounds good there is now a single dockerfile and composefile with a mini test [here](https://github.com/oojo12/mlflow/runs/8285887241?check_suite_focus=true)

going to go through and validate the other workflow today. If this gets merged before I finish that I will pull from Master to remove the compose file and update the workflow file.

docker/docker-compose.yml
```diff
@@ -0,0 +1,7 @@
+version: "3.9"  # optional since v1.27.0
+services:
+  tracking:
+    build: .
+    args:
+        VERSION: ${GITHUB_REF_NAME}
+    entrypoint: ["mlflow", "server"]
```
harupy, line 7: Is this file used in the workflow added in https://github.com/mlflow/mlflow/pull/6735?

oojo12: you mean #6732? No, it isn't and has been deleted.

harupy: yes, #6732

harupy:
@oojo12 Thanks for the update! Can you fix the DCO check failure?

harupy (approved):
LGTM!

docker/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM python:3.10.6-slim-bullseye
+ARG VERSION
+RUN pip install --no-cache mlflow==$VERSION
```
Rusteam, line 3: It complains about not having `psycopg2` installed when using with **postgres**

oojo12: @Rusteam can you create an issue with a reproducible example, since this PR has already been merged? With the reproducible example I can identify and rectify any missing OS level dependencies.

@harupy do you think it is better to add this to the requirements.txt for mlflow?

Issue #3123, opened 2020-07-17
[FR] Allow users to specify a cache directory for loading models


Focus:
## Willingness to contribute
The MLflow Community encourages new feature contributions. Would you or another member of your organization be willing to contribute an implementation of this feature (either as an MLflow Plugin or an enhancement to the MLflow code base)?

- [x] Yes. I can contribute this feature independently.
- [ ] Yes. I would be willing to contribute this feature with guidance from the MLflow community.
- [ ] No. I cannot contribute this feature at this time.

## Proposal Summary

When using the `load_model` functionality I would like to have a persistent cache directory which is checked before the model is downloaded. For example:

```python
# download the model to /my_cache and open it
model = mlflow.pytorch.load_model('models:/my_model/4', cache='/my_cache') 
# loads the model from the local directory
model2 = mlflow.pytorch.load_model('models:/my_model/4', cache='/my_cache')
```

## Motivation

We have large models stored on GCP which are referenced by tagging names in MLFlow. These take quite a while to download and experiment with each time they are used and are a major inconvenience. We have written some code to manually look up the artifact location, download and load models to get around this.

### What component(s), interfaces, languages, and integrations does this feature affect?
Components 
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for
Model Registry
- [x] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: Local serving, model deployment tools, spark UDFs
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interfaces
- [ ] `area/uiux`: Front-end, user experience, JavaScript, plotting
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Languages 
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients

Integrations
- [ ] `integrations/azure`: Azure and Azure ML integrations
- [ ] `integrations/sagemaker`: SageMaker integrations

Kuzat:
[...]

PR #6735, opened 2022-09-08
Only download the file if it does not already exist

michaguenther:
## Related Issues/PRs

#3123

Before downloading a file form the s3 bucket it is useful to check if it already exists. This saves e.g a lot of time if the same model is loaded twice with load_model.

## How is this patch tested?

I have tested the code manually. Since the change is minimalistic and the reliance depends on os.path.exists I think it is not necessary to write a pytest function for this change.

Components

Interface

Language

Integrations


mlflow/store/artifact/s3_artifact_repo.py
```diff
@@ -190,10 +190,11 @@ def _verify_listed_object_contains_artifact_path_prefix(listed_object_path, arti
             )
 
     def _download_file(self, remote_file_path, local_path):
-        (bucket, s3_root_path) = data.parse_s3_uri(self.artifact_uri)
-        s3_full_path = posixpath.join(s3_root_path, remote_file_path)
-        s3_client = self._get_s3_client()
-        s3_client.download_file(bucket, s3_full_path, local_path)
```
harupy, line 196: @michaguenther Does `s3_client.download_file` overwrite the existing file?

dbczumar: @michaguenther How can we verify that the local file is up-to-date?

michaguenther: @dbczumar @harupy I have further developed the check. In case a file already exists locally I get the etag, hash of the file in the s3 bucket (without downloading the file in the first place). Then I compare it with the local hash and skip the download if the hashes match. The hash comparison was adopted from [stackoverflow](https://stackoverflow.com/questions/1775816/how-to-get-the-md5sum-of-a-file-on-amazons-s3).

dbczumar: @michaguenther Thanks! Is there an S3 client API for computing this ETag? Alternatively, does AWS provide official code for calculating it? Reason I ask is that the etag doesn't always seem to be an md5 hash: https://stackoverflow.com/a/41735844/11952869.

mlflow/store/artifact/s3_artifact_repo.py
```diff
@@ -188,13 +189,48 @@ def _verify_listed_object_contains_artifact_path_prefix(listed_object_path, arti
...
+        return m.hexdigest()
+
+    @staticmethod
+    def _etag_checksum(filename, chunk_size=8 * 1024 * 1024):
+        md5s = []
+        with open(filename, 'rb') as f:
+            for data in iter(lambda: f.read(chunk_size), b''):
+                md5s.append(hashlib.md5(data).digest())
+        m = hashlib.md5(b"".join(md5s))
+        return '{}-{}'.format(m.hexdigest(), len(md5s))
```
dbczumar, line 223: `len(md5s)` seems to depend on `chunk_size`, but we can't know the actual chunk size that was used to upload the file. Accordingly, it seems like this checksum won't be valid for multi-part uploads.

mlflow/store/artifact/s3_artifact_repo.py
michaguenther: @dbczumar There seems to be a [new api](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/), but this has to be activated in the aws settings. 
The code be then further optimised with the following [method](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_attributes).

I am not sure if your described issue can be resolved at the moment. 
Overall the worst thing that happens with the current code is that the file gets downloaded although it would not have been necessary but for many case we manage to skip the download.

ArVar:
Is anyone still working on that? This would be a great feature!

michaguenther:
> Is anyone still working on that? This would be a great feature!

@ArVar The suggested feature worked for me like a charm, but was unfortunately not approved. I guess not recommended but you can always copy this change into your local mlflow source files.

PR #6736, opened 2022-09-08
Add generated run name to create run backend stores if not supplied

BenWilson2:

## Related Issues/PRs

#xxx

Runs that are started and do not have a defined `run_name` associated with them will now have a generated name created and set as a tag to the run. 
The format of the generated name is: {predicate}-{noun}-{random integer}

## How is this patch tested?

Modifications to existing test suites to validate the generated name and the persistence of overridden names supplied by users.

New runs will now have a `run_name` generated for them if not supplied at run creation time. 

Components

Interface

Language

Integrations


mlflow/store/tracking/rest_store.py
```diff
@@ -183,6 +185,9 @@ def create_run(self, experiment_id, user_id, start_time, tags):
 
         :return: The created Run object
         """
+        if MLFLOW_RUN_NAME not in [tag.key for tag in tags]:
+            tags.append(RunTag(MLFLOW_RUN_NAME, _generate_random_name()))
+
```
dbczumar, line 190: Rather than adding this in rest_store, which is still a client-side abstraction, can we add the run name generation to the `mlflow server` handler here: https://github.com/mlflow/mlflow/blob/7219771141c7140fd68f6ed3a5b649e5ad1dab13/mlflow/server/handlers.py#L666-L683 ?

mlflow/store/tracking/file_store.py
```diff
@@ -588,6 +589,13 @@ def create_run(self, experiment_id, user_id, start_time, tags):
         mkdir(run_dir, FileStore.METRICS_FOLDER_NAME)
         mkdir(run_dir, FileStore.PARAMS_FOLDER_NAME)
         mkdir(run_dir, FileStore.ARTIFACTS_FOLDER_NAME)
+        if isinstance(tags, list):
+            if MLFLOW_RUN_NAME not in [tag.key for tag in tags]:
+                tags.append(RunTag(MLFLOW_RUN_NAME, _generate_random_name()))
+        # cli sends tags to the backend store as a dict
+        elif isinstance(tags, dict) and not tags:
+            if MLFLOW_RUN_NAME not in tags.keys():
+                tags = [RunTag(MLFLOW_RUN_NAME, _generate_random_name())]
```
dbczumar, line 598: Are you sure? It's not clear to me how line 592 would work if that were the case, unless the dict keys are `RunTag` entities, in which case we should fix the CLI behavior rather than expanding the type handling for FileStore.

mlflow/store/tracking/sqlalchemy_store.py
```diff
@@ -539,7 +540,14 @@ def create_run(self, experiment_id, user_id, start_time, tags):
                 lifecycle_stage=LifecycleStage.ACTIVE,
             )
 
-            run.tags = [SqlTag(key=tag.key, value=tag.value) for tag in tags] if tags else []
+            if MLFLOW_RUN_NAME not in [tag.key for tag in tags]:
+                run_name_tag = RunTag(MLFLOW_RUN_NAME, _generate_random_name())
+                if tags:
```
dbczumar, line 545: if `tags` were `None`, then line 543 would fail. I think we can get rid of the if / else and always call `append()`. If there's any possibility of tags being `None`, we should validate it prior to line 543.

dbczumar (commented):
@BenWilson2 Truly spectacular! Just left a few small comments. Can't wait to ship this in once they're addressed.

cc @sunishsheth2009 @jinzhang21

mlflow/utils/name_utils.py
```diff
@@ -0,0 +1,305 @@
...
+
+    :param sep: String seperator for word spacing
+    :param integer_scale: dictates the maximum scale range for random integer sampling (power of 10)
+
+    :return: A random string phrase comprised of a predicate, noun, and random integer
+    """
+
+    predicate = random.choice(_GENERATOR_PREDICATES).lower()
+    noun = random.choice(_GENERATOR_NOUNS).lower()
+    num = random.randint(0, 10**integer_scale)
```
dbczumar, line 15: I like these words. There aren't too too many of them, so we don't risk adding an excessive amount of text content to the MLflow library, and it's pretty easy to extend them later. We can generate ~ 20k unique pairings of predicate / animal and 20 million unique run names, which seems sufficiently distinct :D

mlflow/store/tracking/file_store.py
BenWilson2: 100% due to trying to address a test failure (the test config in the unit test was incorrect). Fixed.

mlflow/store/tracking/rest_store.py
BenWilson2: added (and removed from rest_store)!

mlflow/store/tracking/sqlalchemy_store.py
BenWilson2: changed!

mlflow/server/handlers.py
```diff
@@ -669,6 +671,8 @@ def _create_run():
     )
 
     tags = [RunTag(tag.key, tag.value) for tag in request_message.tags]
+    if MLFLOW_RUN_NAME not in [tag.key for tag in tags]:
+        tags.append(RunTag(MLFLOW_RUN_NAME, _generate_random_name()))
```
harupy, line 675: Can we document MLflow automatically generates a random run name somewhere?

jinzhang21 (commented):
Awesome PR, @BenWilson2 ! Is there a limit on total number of chars for generated run names? It'd make display easier if we know the max length of autogen run names.

harupy:
@jinzhang21 

>  Is there a limit on total number of chars for generated run names?

Wrote a script to check the longest name:

```python
def choose_max(strings):
    return max(strings, key=len)


longest_name = "-".join(
    [choose_max(_GENERATOR_PREDICATES), choose_max(_GENERATOR_NOUNS), str(10**3)]
)
print(longest_name)
print(len(longest_name))
```

```
knowledgeable-crocodile-1000
28
```

(assuming we always use the default `sep` and `integer_scale`).

harupy:
How `knowledgeable-crocodile-1000` is displayed on UI:

image

harupy (approved):
Looks good to me once https://github.com/mlflow/mlflow/pull/6736#discussion_r966547302 is addressed!

dbczumar (approved):
LGTM! Before merging, can we also display the run name column by default in the UI now? It's currently hidden by default, so users can't see our awesome new run names unless they use the column selector to show the column.

Thanks @BenWilson2 !

sunishsheth2009:
> LGTM! Before merging, can we also display the run name column by default in the UI now? It's currently hidden by default, so users can't see our awesome new run names unless they use the column selector to show the column.

I think @hubertzub-db is going to make that change on the FE in the new version of experiment management. Is it okay to wait for his change to go in or do we want to make that change in the old UI for the time being as well?

harupy:
Is the run name column hidden by default?

image

On my machine, it's shown by default (after removing the local storage in the browser)

mlflow/server/handlers.py
sunishsheth2009: Just for knowledge purpose. I know that we use all the diff types of stores where we need to set the run_name. Why are we setting it in the `server/handler` as well?

sunishsheth2009 (approved):
Awesome work @BenWilson2 :) 
Thank you for taking this on.  Love it.

harupy:
@dbczumar

> LGTM! Before merging, can we also display the run name column by default in the UI now? It's currently hidden by default, so users can't see our awesome new run names unless they use the column selector to show the column.

The run name column should show up after clearing the local storage of the browser :)

mlflow/server/handlers.py
BenWilson2: yep! Added additional information in docs and added a note in pydoc for `fluent.py`

BenWilson2:
> Just for knowledge purpose. I know that we use all the diff types of stores where we need to set the run_name. Why are we setting it in the server/handler as well?

For the REST interface :) (FYI, I originally put the change in the rest_store.py client-side logic before being educated by @dbczumar about how setting it in the backend server handler is more robust with fewer places to manage the logic)

jinzhang21:
> How `knowledgeable-crocodile-1000` is displayed on UI:
> 
> image

Thanks, @harupy . 28 chars seem a bit long. Can we limit it to 20 or even 16? Essentially automatically reject any names longer than the max length and regenerate a new one in the logic. @BenWilson2 WDYT?

BenWilson2:
> > How `knowledgeable-crocodile-1000` is displayed on UI:
> > image
> 
> Thanks, @harupy . 28 chars seem a bit long. Can we limit it to 20 or even 16? Essentially automatically reject any names longer than the max length and regenerate a new one in the logic. @BenWilson2 WDYT?

@jinzhang21 I'm going to remove the longer noun names (replace them with shorter names) and provide a retry for elements max length of 20. I'll push the changes soon.

jinzhang21 (approved):

docs/source/tracking.rst
```diff
@@ -746,7 +749,7 @@ The :py:func:`MlflowClient.set_tag() <mlflow.client.MlflowClient.set_tag>` funct
 
   client.set_tag(run.info.run_id, "tag_key", "tag_value")
 
-.. important:: Do not use the prefix ``mlflow`` for a tag.  This prefix is reserved for use by MLflow.
+.. important:: Do not use the prefix ``mlflow.`` (i.e. ``mlflow.note``) for a tag.  This prefix is reserved for use by MLflow. See :ref:`system_tags` for a list of reserved tag keys.
```
dbczumar, line 752:
```suggestion
.. important:: Do not use the prefix ``mlflow.`` (e.g. ``mlflow.note``) for a tag.  This prefix is reserved for use by MLflow. See :ref:`system_tags` for a list of reserved tag keys.
```

dbczumar (approved):
LGTM with the tiniest of doc nits!

mlflow/utils/name_utils.py
```diff
@@ -0,0 +1,320 @@
...
+    max_iter = 10
+    i = 0
+    while True:
+        name = _generate_string(sep, integer_scale)
+        if len(name) <= max_length:
+            return name
+        elif i == max_iter:
+            return name[:max_length]
+        else:
+            i += 1
```
harupy, line 30:
```suggestion
    for _ in range(10):
        name = _generate_string(sep, integer_scale)
        if len(name) <= max_length:
            return name
    return name[:max_length]
```

nit: just for simplifying the code.

BenWilson2: much better :)

mlflow/utils/name_utils.py
thefifthhead: @BenWilson2 @dbczumar - I am not able to understand how we are guarantying no name collisions by just relying on random. Is that handled in some other code flow that I have missed or there is there something that I am missing here.

BenWilson2: Run name isn't a primary key. Why would a collision-free solution be required here?

thefifthhead: Thanks @BenWilson2 . Yes you are correct, i think i got confused that name is a primary key as duplicate experiment name was not allowed in mlflow. It is cleared now.

PR #6752, opened 2022-09-09
writing params to yaml and show in step card

prithvikannan:
Signed-off-by: Prithvi Kannan <prithvi.kannan@databricks.com>

## Related Issues/PRs

#xxx

1. Writes the output of HP tuning to `best_parameters.yaml`. Any hardcoded parameters that are overriden by the tuned hyperparameters are not outputted.

```
# tuned hyperparameters
alpha: 0.003672067524975855

# hardcoded parameters
penalty: l1
```
Users can now copy the `best_parameters.yaml` outputs into the `estimator_params` section of the pipeline.yaml and reproduce the best result from HP tuning.

2. Adds a new step card to show best parameters.
image

3. Adds a step card to show tuning trials.
image

4. Also, I have created a custom `_write_yaml()` function which handles issues with dumping numpy floats (result of hyperopt tuning) into yaml. 

## How is this patch tested?

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/steps/train.py
+++ b/mlflow/pipelines/steps/train.py
@@ -34,6 +34,7 @@ from mlflow.projects.utils import get_databricks_env_vars
 from mlflow.tracking import MlflowClient
 from mlflow.tracking.fluent import _get_experiment_id
 from mlflow.utils.databricks_utils import get_databricks_run_url
+from mlflow.utils.file_utils import write_yaml
 from mlflow.utils.mlflow_tags import (
     MLFLOW_SOURCE_TYPE,
     MLFLOW_PIPELINE_TEMPLATE_NAME,
@@ -142,7 +143,12 @@ class TrainStep(BaseStep):
             estimator_hardcoded_params = self.step_config["estimator_params"]
             if self.step_config["tuning_enabled"]:
                 estimator = self._tune_and_get_best_estimator(
-                    estimator_hardcoded_params, estimator_fn, X_train, y_train, validation_df
+                    estimator_hardcoded_params,
+                    estimator_fn,
+                    X_train,
+                    y_train,
+                    validation_df,
+                    output_directory,
                 )
             else:
                 estimator = estimator_fn(estimator_hardcoded_params)
@@ -547,7 +553,13 @@ class TrainStep(BaseStep):
         return environ
 
     def _tune_and_get_best_estimator(
-        self, estimator_hardcoded_params, estimator_fn, X_train, y_train, validation_df
+        self,
+        estimator_hardcoded_params,
+        estimator_fn,
+        X_train,
+        y_train,
+        validation_df,
+        output_directory,
     ):
         tuning_params = self.step_config["tuning"]
         try:
@@ -611,6 +623,14 @@ class TrainStep(BaseStep):
             best_combined_params = dict(estimator_hardcoded_params, **best_hp_params)
         else:
             best_combined_params = estimator_hardcoded_params
+
+        # write outputs
+        write_yaml(
+            root=output_directory,
+            file_name="best_hardcoded_params.yaml",
+            data=estimator_hardcoded_params,
+        )
+        write_yaml(root=output_directory, file_name="best_tuned_params.yaml", data=best_hp_params)
 
         return estimator_fn(best_combined_params)
 
--- a/mlflow/pipelines/steps/train.py
+++ b/mlflow/pipelines/steps/train.py
@@ -353,14 +353,19 @@ class TrainStep(BaseStep):
 
     def _get_tuning_df(self, run, params):
         exp_id = _get_experiment_id()
+        primary_metric_tag = f"metrics.{self.primary_metric}_on_data_validation"
+        order_str = (
+            "DESC" if self.evaluation_metrics_greater_is_better[self.primary_metric] else "ASC"
+        )
         tuning_runs = mlflow.search_runs(
             [exp_id],
             filter_string=f"tags.mlflow.parentRunId like '{run.info.run_id}'",
+            order_by=[f"{primary_metric_tag} {order_str}"],
         )
         params = [f"params.{param}" for param in params]
-        tuning_runs = tuning_runs.filter(
-            [f"metrics.{self.primary_metric}_on_data_validation", *params]
-        )
+        tuning_runs = tuning_runs.filter([primary_metric_tag, *params])
+        # tuning_runs.index.name = "Model Rank"
+        tuning_runs = tuning_runs.reset_index().rename(columns={"index": "Model Rank"})
         return tuning_runs
 
     def _build_step_card(
@@ -509,15 +514,25 @@ class TrainStep(BaseStep):
         # Tab 9: HP trials
         if tuning_df is not None:
             tuning_trials_card_tab = card.add_tab(
-                "Tuning Trials", "{{ SEARCH_SPACE }}" + "{{ TUNING_TABLE }}"
+                "Tuning Trials",
+                "{{ SEARCH_SPACE }}" + "{{ TUNING_TABLE_TITLE }}" + "{{ TUNING_TABLE }}",
             )
             tuning_params = yaml.dump(self.step_config["tuning"]["parameters"])
             tuning_trials_card_tab.add_html(
                 "SEARCH_SPACE",
                 f"<b>Tuning search space:</b> <br><pre>{tuning_params}</pre><br><br>",
             )
+            tuning_trials_card_tab.add_html("TUNING_TABLE_TITLE", "<b>Tuning results:</b><br>")
             tuning_trials_card_tab.add_html(
-                "TUNING_TABLE", BaseCard.render_table(tuning_df, hide_index=False)
+                "TUNING_TABLE",
+                BaseCard.render_table(
+                    tuning_df.style.apply(
+                        lambda row: pd.Series("font-weight: bold", row.index),
+                        axis=1,
+                        subset=tuning_df.index[0],
+                    ),
+                    hide_index=True,
+                ),
             )
 
         return card
@@ -660,10 +675,12 @@ class TrainStep(BaseStep):
                 manual_log_params = {}
                 for param_name, param_value in estimator_args.items():
                     if param_name in autologged_params:
-                        if not self._is_equal(param_value, autologged_params[param_name]):
+                        if not self._is_tuning_param_equal(
+                            param_value, autologged_params[param_name]
+                        ):
                             _logger.warning(
-                                f"Failed to log parameter due to conflict. Attempting to log "
-                                f"search space parameter {param_name} as {param_value}, "
+                                f"Failed to log search space parameter due to conflict. Attempted "
+                                f"to log search space parameter {param_name} as {param_value}, "
                                 f"but {param_name} is already logged as "
                                 f"{autologged_params[param_name]} during training. "
                                 f"Are you passing `estimator_params` properly to the estimator?"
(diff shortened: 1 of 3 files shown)


mlflow/pipelines/steps/train.py
```diff
@@ -608,9 +659,25 @@ def objective(hyperparameter_args):
...
+        best_combined_params = dict(estimator_hardcoded_params, **best_hp_params)
+
+        self._write_yaml(
+            root=output_directory, file_name="best_hyperparameters.yaml", data=best_hp_params
+        )
+        self._write_yaml(
+            root=output_directory,
+            file_name="best_hardcoded_parameters.yaml",
+            data=best_hardcoded_params,
+        )
```
dbczumar, line 680: Apologies, when we discussed displaying the best parameters on the step card in separate cells / divs of the same tab last week, I didn't think about the YAML artifact output behavior. IMO, we should still write parameters to a single file where tuned parameters are listed first, followed by hardcoded parameters. We should separate these sections by a comment, perhaps by making one `yaml.safe_dump()` call, writing a comment at the end of the file, and then making a second `yaml.safe_dump()` call where the second calls *appends* to the file

dbczumar: Can we also make `best_parameters.yaml` accessible via `Pipeline().get_artifact("best_parameters")`?

mlflow/utils/file_utils.py
```diff
@@ -186,6 +186,17 @@ def read_yaml(root, file_name):
...
 
+def dump_yaml(data):
+    """
+    Dump data from a dictionary into a formatted yaml
+
+    :param contents: Dictionary containing yaml contents
+
+    :return: Formatted yaml containing data
+    """
+    return yaml.dump(data)
```
dbczumar, line 197: We already have a `write_yaml` util in this file. Can we use that instead (after augmenting for append capabilities as suggested in https://github.com/mlflow/mlflow/pull/6752/files#r968847408)?

mlflow/pipelines/steps/train.py
```diff
@@ -457,6 +465,43 @@ def render_schema(inputs, title):
         else:
             run_card_tab.add_markdown("MODEL_URI", f"**MLflow Model URI:** `{model_uri}`")
 
+        # Tab 8: Best Parameters
+        if self.step_config["tuning"]["enabled"]:
```
dbczumar, line 469: Instead of check the step conf here, can we have the `_build_step_card` method accept a `best_parameters` object and populate tabs depending on whether this object is present? Ditto for `hp_trials_df` below

mlflow/utils/file_utils.py
prithvikannan: here i am using `dump_yaml` for outputting data as yaml (rather than as a python dict) for the step card. i can't use `write_yaml` because that writes to a file instead.

mlflow/pipelines/steps/train.py
```diff
@@ -457,6 +489,34 @@ def render_schema(inputs, title):
...
+            if os.path.exists(best_parameters_yaml):
+                best_hardcoded_parameters = open(best_parameters_yaml).read()
+                tuning_params_card_tab.add_html(
+                    "BEST_PARAMETERS",
+                    f"<b>Best parameters:</b><br>"
+                    f"<pre>{best_hardcoded_parameters}</pre><br><br>",
+                )
+
+        # Tab 9: HP trials
+        if tuning_df is not None:
```
prithvikannan, line 513: @dbczumar i think you had a comment on a previous version that this tab should be conditioned on `best_estimator_params`. the creation of `tuning_df` on line 232 only happens if `best_estimator_params` exists, so i believe i am reflecting that behavior

mlflow/pipelines/steps/train.py
```diff
@@ -652,3 +725,27 @@ def _construct_search_space_from_yaml(self, params):
...
+            file.write("# hardcoded parameters\n")
+            self._process_and_safe_dump(best_hardcoded_params, file, default_flow_style=False)
+        mlflow.log_artifact(best_parameters_path, artifact_path="best_parameters.yaml")
+
+    def _process_and_safe_dump(self, data, file, **kwargs):
+        import numpy as np
+
+        processed_data = {}
+        for key, value in data.items():
+            if isinstance(value, np.floating):
```
prithvikannan, line 745: @dbczumar `np.floating` covers various precisions of numpy floats 

https://stackoverflow.com/questions/28292542/how-to-check-if-a-number-is-a-np-float64-or-np-float32-or-np-float16

mlflow/pipelines/steps/train.py
dbczumar: Agreed! Nice!

mlflow/pipelines/steps/train.py
```diff
@@ -586,6 +653,17 @@ def objective(hyperparameter_args):
...
                     },
                 )
+                autologged_params = mlflow.get_run(run_id=tuning_run.info.run_id).data.params
+                for param_name, param_value in estimator_args.items():
+                    if param_name in autologged_params:
+                        if not self._is_equal(param_value, autologged_params[param_name]):
+                            _logger.warning(
+                                f"Failed to log parameter {param_name} due to "
+                                f"conflict. old_value: {autologged_params[param_name]} "
+                                f"new_value: {param_value} type: {type(param_value)}"
```
dbczumar, line 663: Can we be a bit more specific about the source of the conflict and what parameter we're attempting to log? I.e. it should be clear that we're trying to log a parameter from the hyperparameter tuning search space but we can't because the model trained at that point within the search space had a different value for the parameter. We should encourage the user to make sure that they're passing `estimator_params` properly to their estimator.

mlflow/pipelines/steps/train.py
```diff
@@ -586,6 +653,17 @@ def objective(hyperparameter_args):
...
+                            )
+                    else:
+                        mlflow.log_param(param_name, param_value)
```
dbczumar, line 666: Can we batch up the parameters that we want to log (the ones without conflicts) and make a single `mlflow.log_params()` call? This is more efficient, especially with API rate limits enforced on Databricks

mlflow/pipelines/steps/train.py
dbczumar: wonderful! let's make sure to include a test case for this kind of behavior in our tests PR

mlflow/pipelines/steps/train.py
```diff
@@ -457,6 +489,34 @@ def render_schema(inputs, title):
...
+        if best_estimator_params:
+            tuning_params_card_tab = card.add_tab(
+                "Best Parameters",
+                "{{ SEARCH_SPACE }} " + "{{ BEST_PARAMETERS }} ",
+            )
+            tuning_params = yaml.dump(self.step_config["tuning"]["parameters"])
+            tuning_params_card_tab.add_html(
+                "SEARCH_SPACE",
+                f"<b>Tuning search space:</b> <br><pre>{tuning_params}</pre><br><br>",
+            )
```
dbczumar, line 502: Can we move the tuning search space information to the HP trials section? The best parameters section is concerned with displaying the best parameters. The tuning section is concerned with displaying the search space and results at each point.

mlflow/pipelines/steps/train.py
```diff
@@ -218,6 +226,12 @@ def _run(self, output_directory):
             leaderboard_df = self._get_leaderboard_df(run, eval_metrics)
         except Exception as e:
             _logger.warning("Failed to build model leaderboard due to unexpected failure: %s", e)
+        tuning_df = None
+        if best_estimator_params:
+            try:
+                tuning_df = self._get_tuning_df(run, best_estimator_params.keys())
+            except Exception as e:
+                _logger.warning("Failed to build tuning table due to unexpected failure: %s", e)
```
dbczumar, line 234:
```suggestion
                _logger.warning("Failed to build tuning results table due to unexpected failure: %s", e)
```

mlflow/pipelines/steps/train.py
```diff
@@ -652,3 +737,37 @@ def _construct_search_space_from_yaml(self, params):
...
         return search_space
+
+    def _write_yaml_output(self, best_hp_params, best_hardcoded_params, output_directory):
+        best_parameters_path = os.path.join(output_directory, "best_parameters.yaml")
+        if os.path.exists(best_parameters_path):
+            os.remove(best_parameters_path)
+        with open(best_parameters_path, "a") as file:
+            file.write("# tuned hyperparameters\n")
+            self._process_and_safe_dump(best_hp_params, file, default_flow_style=False)
+            file.write("# hardcoded parameters\n")
```
dbczumar, line 748: Nit: Can we add a newline between the two sections of hyperparameters?

mlflow/pipelines/steps/train.py
```diff
@@ -457,6 +491,35 @@ def render_schema(inputs, title):
...
+        if tuning_df is not None:
+            tuning_trials_card_tab = card.add_tab(
+                "Tuning Trials", "{{ SEARCH_SPACE }}" + "{{ TUNING_TABLE }}"
+            )
+            tuning_params = yaml.dump(self.step_config["tuning"]["parameters"])
+            tuning_trials_card_tab.add_html(
+                "SEARCH_SPACE",
+                f"<b>Tuning search space:</b> <br><pre>{tuning_params}</pre><br><br>",
+            )
+            tuning_trials_card_tab.add_html(
```
dbczumar, line 519: Can we add some text before this table to indicate that these are the tuning results? "Tuning Results" seems good. Can we sort the results by the primary metric such that the first result is the best and also bold the first result? Looks like our sort order is incorrect (worst model first) in the screenshot you've provided in the description.

mlflow/pipelines/steps/train.py
```diff
@@ -586,6 +656,23 @@ def objective(hyperparameter_args):
                         "log_model_explainability": False,
                     },
                 )
+                autologged_params = mlflow.get_run(run_id=tuning_run.info.run_id).data.params
+                manual_log_params = {}
+                for param_name, param_value in estimator_args.items():
+                    if param_name in autologged_params:
+                        if not self._is_equal(param_value, autologged_params[param_name]):
+                            _logger.warning(
+                                f"Failed to log parameter due to conflict. Attempting to log "
```
dbczumar, line 665:
```suggestion
                                f"Failed to log search space parameter due to conflict. Attempted to log "
```

mlflow/pipelines/steps/train.py
```diff
@@ -652,3 +746,37 @@ def _construct_search_space_from_yaml(self, params):
...
+        for key, value in data.items():
+            if isinstance(value, np.floating):
+                processed_data[key] = float(value)
+            elif isinstance(value, np.integer):
+                processed_data[key] = int(value)
+            else:
+                processed_data[key] = value
+        return yaml.safe_dump(processed_data, file, **kwargs)
+
+    def _is_equal(self, new_param, logged_param):
```
dbczumar, line 774:
```suggestion
    def _is_tuning_param_equal(self, new_param, logged_param):
```

mlflow/pipelines/steps/train.py
```diff
@@ -652,3 +746,37 @@ def _construct_search_space_from_yaml(self, params):
...
+        if os.path.exists(best_parameters_path):
+            os.remove(best_parameters_path)
+        with open(best_parameters_path, "a") as file:
+            file.write("# tuned hyperparameters\n")
+            self._process_and_safe_dump(best_hp_params, file, default_flow_style=False)
+            file.write("\n# hardcoded parameters\n")
+            self._process_and_safe_dump(best_hardcoded_params, file, default_flow_style=False)
+        mlflow.log_artifact(best_parameters_path)
+
+    def _process_and_safe_dump(self, data, file, **kwargs):
```
dbczumar, line 761:
```suggestion
    def _safe_dump_with_numeric_values(self, data, file, **kwargs):
```

mlflow/pipelines/steps/train.py
```diff
@@ -652,3 +746,37 @@ def _construct_search_space_from_yaml(self, params):
                     error_code=INVALID_PARAMETER_VALUE,
                 )
         return search_space
+
+    def _write_yaml_output(self, best_hp_params, best_hardcoded_params, output_directory):
```
dbczumar, line 750:
```suggestion
    def _write_tuning_yaml_outputs(self, best_hp_params, best_hardcoded_params, output_directory):
```

dbczumar (approved):
LGTM once remaining comments have been addressed! Thanks @prithvikannan !

PR #6721, opened 2022-09-07
Remove DefaultEvaluator shap plot de-duplicating feature names logic and fix color bar

WeichenXu123:
Signed-off-by: Weichen Xu <weichen.xu@databricks.com>

## Related Issues/PRs

#xxx

## What changes are proposed in this pull request?

Remove DefaultEvaluator shap plot de-duplicating feature names logic,
According to suggestion https://github.com/mlflow/mlflow/pull/6705#issuecomment-1237829665
and fix color bar.

Closes #6705

## How is this patch tested?

manually.
shap_beeswarm_plot_on_data_diabetes_dataset
shap_feature_importance_plot_on_data_diabetes_dataset
shap_summary_plot_on_data_diabetes_dataset

## Does this PR change the documentation?

- [x] No. You can skip the rest of this section.
- [ ] Yes. Make sure the changed pages / sections render correctly by following the steps below.

1. Click the `Details` link on the `Preview docs` check.
2. Find the changed pages / sections and make sure they render correctly.

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [x] Yes. Give a description of this change to be included in the release notes for MLflow users.

Remove DefaultEvaluator shap plot de-duplicating feature names logic

### What component(s), interfaces, languages, and integrations does this PR affect?
Components
- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/pipelines`: Pipelines, Pipeline APIs, Pipeline configs, Pipeline Templates
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [x] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface
- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language
[...]

PR #6754, opened 2022-09-11
Fix sklearn warnings during explainer

dsgibbons:
## Related Issues/PRs

#6555

1. Store and use `original_column_names` when running inference in `default_evaluator` to avoid sklearn warning: `X has feature names, but <MODEL> was fitted without feature names`
2. Rename and add `explainability_ignore_exceptions` to docstring for `evaluator_config` in `evaluate` 

## How is this patch tested?

No, uses existing tests. Some tests still show sklearn warning, but I believe this is because these tests specifically add in column names before evaluation. In this case, the sklearn warning is probably quite useful and should remain.

3. Find the changed pages / sections and make sure they render correctly.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -12,7 +12,6 @@ from mlflow.entities import RunTag
 from mlflow.tracking.artifact_utils import _download_artifact_from_uri
 from mlflow.utils import _get_fully_qualified_class_name
 from mlflow.utils.class_utils import _get_class_from_string
-from mlflow.utils.string_utils import generate_feature_name_if_not_string
 from mlflow.utils.annotations import experimental
 from mlflow.utils.proto_json_utils import NumpyEncoder
 from mlflow.models.evaluation.validation import (
@@ -341,7 +340,6 @@ class EvaluationDataset:
                     f"feature_{str(i + 1).zfill(math.ceil((math.log10(num_features + 1))))}"
                     for i in range(num_features)
                 ]
-            self._original_column_names = list(range(num_features))
         elif isinstance(data, self._supported_dataframe_types):
             if not isinstance(targets, str):
                 raise MlflowException(
@@ -365,13 +363,9 @@ class EvaluationDataset:
             if feature_names is not None:
                 self._features_data = data[list(feature_names)]
                 self._feature_names = feature_names
-                self._original_column_names = feature_names
             else:
                 self._features_data = data.drop(targets, axis=1, inplace=False)
-                self._feature_names = [
-                    generate_feature_name_if_not_string(c) for c in self._features_data.columns
-                ]
-                self._original_column_names = self._features_data.columns
+                self._feature_names = self._features_data.columns
         else:
             raise MlflowException(
                 message="The data argument must be a numpy array, a list or a Pandas DataFrame, or "
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -422,13 +422,6 @@ class EvaluationDataset:
         return self._hash
 
     @property
-    def original_column_names(self):
-        """
-        returns original column names of features data as a list.
-        """
-        return self._original_column_names
-
-    @property
     def _metadata(self):
         """
         Return dataset metadata containing name, hash, and optional path.
--- a/mlflow/models/evaluation/base.py
+++ b/mlflow/models/evaluation/base.py
@@ -341,6 +341,7 @@ class EvaluationDataset:
                     f"feature_{str(i + 1).zfill(math.ceil((math.log10(num_features + 1))))}"
                     for i in range(num_features)
                 ]
+            self._original_column_names = list(range(num_features))
         elif isinstance(data, self._supported_dataframe_types):
             if not isinstance(targets, str):
                 raise MlflowException(
@@ -364,11 +365,13 @@ class EvaluationDataset:
             if feature_names is not None:
                 self._features_data = data[list(feature_names)]
                 self._feature_names = feature_names
+                self._original_column_names = feature_names
             else:
                 self._features_data = data.drop(targets, axis=1, inplace=False)
                 self._feature_names = [
                     generate_feature_name_if_not_string(c) for c in self._features_data.columns
                 ]
+                self._original_column_names = self._features_data.columns
         else:
             raise MlflowException(
                 message="The data argument must be a numpy array, a list or a Pandas DataFrame, or "
@@ -423,6 +426,13 @@ class EvaluationDataset:
         Dataset hash, includes hash on first 20 rows and last 20 rows.
         """
         return self._hash
+
+    @property
+    def original_column_names(self):
+        """
+        returns original column names of features data as a list.
+        """
+        return self._original_column_names
 
     @property
     def _metadata(self):
@@ -921,6 +931,8 @@ def evaluate(
           explainability insights. Default value is 2000.
         - **explainability_kernel_link**: The kernel link function used by shap kernal explainer.
           Available values are "identity" and "logit". Default value is "identity".
+        - **explainability_ignore_exceptions**: A boolean value specifying whether to ignore 
+          exceptions encountered when using ``shap.Explainer``. Default value is True.
         - **max_classes_for_multiclass_roc_pr**:
           For multiclass classification tasks, the maximum number of classes for which to log
           the per-class ROC curve and Precision-Recall curve. If the number of classes is
--- a/mlflow/models/evaluation/default_evaluator.py
+++ b/mlflow/models/evaluation/default_evaluator.py
@@ -222,7 +222,6 @@ def _get_dataframe_with_renamed_columns(x, new_column_names):
     NaNs for every column in new_column_names that does not exist in x.columns. This function
     instead creates a new pd.DataFrame object from x, and then explicitly renames the columns
     to avoid NaNs.
-
     :param x: :param data: A data object, such as a Pandas DataFrame, numPy array, or list
     :param new_column_names: Column names for the output Pandas DataFrame
     :return: A pd.DataFrame with x as data, with columns new_column_names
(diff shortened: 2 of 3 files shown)


BenWilson2:
@WeichenXu123 any concerns with this?

mlflow/models/evaluation/base.py
```diff
@@ -341,6 +341,7 @@ def __init__(self, data, *, targets, name=None, path=None, feature_names=None):
                     f"feature_{str(i + 1).zfill(math.ceil((math.log10(num_features + 1))))}"
                     for i in range(num_features)
                 ]
+            self._original_column_names = list(range(num_features))
```
harupy, line 344: Do we need this change? We no longer truncate feature names (see https://github.com/mlflow/mlflow/pull/6721). I don't think we need to memorize `_original_column_names`.

@WeichenXu123 Thoughts?

dsgibbons: If the test data is a NumPy array, this fix will prevent the warnings from sklearn

dsgibbons: We can't just use self.feature_names as they are strings, which introduces the named columns producing the sklearn warnings when using explainability

harupy: @WeichenXu123 Can we directly pass a numpy array to shap explainer?

harupy: @WeichenXu123

WeichenXu123: > Can we pass a numpy array to shap explainer?

I think no. Because some Sklearn model require inference data with column names (e.g. some transformers in a pipeline need to read column name)

WeichenXu123: if original data is a dataframe with unlabeled features,
when calling shap evalaution, we should restore the original dataframe when call model.predict.

harupy: Can we just pass the data supplied by the user without auto-generating feature names?

WeichenXu123: This case `pd.DataFrame(np_array)` it generates unlabeled dataframe. This is valid case.
Or we can raise error when constructing EvaluationDataset if input dataset is unlabeled Dataframe

harupy: When do users do `pd.DataFrame(numpy.array)`?

WeichenXu123: It comes from @dsgibbons 's case.
If it is a rare case, we can forbid this case when constructing EvaluationDataset

harupy: I don't think we need to forbid. We can just pass the dataframe to shap without altering the columns.

dsgibbons: I remember trying this @harupy but I believe this led to issues with the categorical data test. I can have another look.

dsgibbons: Yep, I can confirm that if you just try to pass `pd.DataFrame` directly to shap without adding in column information, the test `test_pipeline_model_kernel_explainer_on_categorical_features` fails due to:

```*** KeyError: "None of [Index(['f1', 'f2'], dtype='object')] are in the [columns]"``` 

I'm not sure why, but it seems as though this test needs explicit column naming when using shap. 

The only other obvious fix (other than this PR) I can see for this problem:

1. Unstringify the feature names, i.e., replace
```
self._feature_names = [
    generate_feature_name_if_not_string(c) for c in self._features_data.columns
]
```
with 
```
self._feature_names = self._features_data.columns
```

2. Revert to `master`'s use of `feature_names` instead of `original_column_names` e.g.,
```
def _shap_predict_fn(x, predict_fn, feature_names):
    return predict_fn(_get_dataframe_with_renamed_columns(x, feature_names))
```

3. Stringify feature names after evaluation and just before the shap plots get created.
```
shap_values.feature_names = [generate_feature_name_if_not_string(n) for n in self.feature_names]
```

I can make another commit that implements these changes. If you don't like any of these changes and think that fixing this edge case warning makes the code too complicated then I can revert all changes relating to the sklearn warnings and just leave the naming fix for `explainability_ignore_exceptions`

**Comment by dsgibbons** [comment]
Failure case just needs new hash generating. Doesn't change rest of PR so feel free to review.

File: mlflow/models/evaluation/base.py (hunk 1 above)
Comment (WeichenXu123) [reply mlflow/models/evaluation/base.py]: To simplify code and avoid sklearn warning / errors, what about:

When constructing EvaluationDataset, if it is pandas dataframe input, disallow unlabeled dataframe input.

When doing shap evaluation, keep calling `model.predict` with original input data (original numpy array or original pandas dataframe)

Comment (WeichenXu123) [reply mlflow/models/evaluation/base.py]:
```
the test test_pipeline_model_kernel_explainer_on_categorical_features fails due to:

*** KeyError: "None of [Index(['f1', 'f2'], dtype='object')] are in the [columns]"
```

This is because the pipeline trained with labeled dataframe, and it contains some transformers which need to read the column name, so in prediction it also need column name.

File: mlflow/models/evaluation/default_evaluator.py [hunk 2]
```diff
@@ -647,12 +649,19 @@ def _log_model_explainability(self):
                 )
             else:
                 shap_values = explainer(sampled_X)
+            # We want shap plots to use self.feature_names, but we can't use always use
+            # self.feature_names when computing shap_values, as they may differ from the
+            # names used during training (e.g. when using np.ndarrays). Instead, we rename
+            # shap_values.feature_names here before creating the plots.
+            shap_values.feature_names = [
+                generate_feature_name_if_not_string(n) for n in self.feature_names
+            ]
```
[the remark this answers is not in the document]
dsgibbons: I successfully ran every test from `test_default_evaluator.py` and don't seem to have any issues creating the explainability plots. So this at least tests the default explainer and the kernel explainer. Do you happen to remember what issues you were getting by changing `shap_values.feature_names` here? It would be good to have a test case that highlights this problem.

mlflow/models/evaluation/base.py
dsgibbons: I'm not sure how to call `model.predict` with original input data in a way that still passes  `test_pipeline_model_kernel_explainer_on_categorical_features`? Do you have any suggestions?

dsgibbons: I personally preferred the previous solution of saving the original column names, as that also silenced the sklearn warning when evaluating on np.ndarrays (not just unlabeled dataframes). For example,  `test_regressor_evaluation[linear_regressor_model_uri]`

mlflow/models/evaluation/default_evaluator.py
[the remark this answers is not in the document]
WeichenXu123: > Do you happen to remember what issues you were getting by changing shap_values.feature_names here?

Previously I didn't test against approach of setting `shap_values.feature_names` ,
but I hit issue `feature_name set for explainer is not used, but feature_name in dataframe is used`. Shap code is not maintained well and some edge cases might not work well, so I suggest testing feature names with following cases:

Test on Sklearn logistic model ,use default explainer (i.e. linear explainer)
Test on xgboost classifier, use default explainer (i.e. tree explainer)
Test on any model, use specific explainer (via setting evaluator config `explainability_algorithm`): 'exact', 'permutation', 'partition', 'kernel'

If they all work fine, then it's good.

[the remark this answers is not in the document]
dsgibbons: @WeichenXu123 `Explanation.feature_names` comes with a proper setter function to alter the underlying slicer array `Explanation._s`. This setter function is used twice in `shap/_explanation.py`. For example,

```
def _numpy_func(self, fname, **kwargs):
        """ Apply a numpy-style function to this Explanation.
        """
        new_self = copy.copy(self)
        ...
        ...
        ...
            new_self.feature_names = np.array(list(new_values.keys()))
```

The other use in their source code can be found in `Explanation.__getitem__`.

I couldn't find any official API information other than: https://shap.readthedocs.io/en/latest/generated/shap.Explanation.html?highlight=feature_names#shap.Explanation.feature_names - but this doesn't show `feature_names` being set explicitly by the user. There are other elements in this slicer array that don't get setter methods though.

Do you think it is safe to use the setter for `shap_values.feature_names`? If so, then I can test those various explainability_algorithms. I can't foresee there being any issues as this is a postprocessing step that executes after the algorithm has been run - but I can double-check and add explicit test cases to `test_default_evaluator.py`.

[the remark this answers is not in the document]
WeichenXu123: OK. Seems shap doc does not explicitly said we can set the `shap_values.feature_names`, it seems unsafe to do so.

So I suggest we forbid EvaluationDataset input with "unlabeled pandas dataframe", WDYT ?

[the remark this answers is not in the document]
dsgibbons: @WeichenXu123 I disagree about forbidding unlabeled dataframes, as we get the exact same warning when using NumPy arrays. For example, the following block evaluates on NumPy arrays but generates the same warnings as if `X_test` and `y_test` were combined into a single unlabeled dataframe.

```
import mlflow

import pandas as pd
from sklearn import datasets

# prepare train and test data
iris = datasets.load_iris()
X = iris.data
y = iris.target

with mlflow.start_run():
    logistic = LogisticRegression(C=1.0)
    logistic.fit(X_train, y_train)
    mod = mlflow.sklearn.log_model(logistic, artifact_path="mymodel")

    results = mlflow.evaluate(
        model=mod.model_uri,
        model_type="classifier",
        data=X_test,
        targets=y_test,
        evaluator_config={
            "log_model_explainability": True,
        },
        evaluators="default",
    )
```

If we cannot change `shap_values.feature_names` using the built-in setter, then we probably can't avoid this warning for now (well, at least I don't know of a way that doesn't end up breaking `test_pipeline_model_kernel_explainer_on_categorical_features`). 

If we cannot resolve this issue, then I can just strip out the sklearn warning fix, and just leave in the renaming/docstring for `explainability_ignore_exceptions`.

PR #6769, opened 2022-09-13
Update RunInfo, CreateRun and UpdateRun infos to accept run names

apurva-koti:

## Related Issues/PRs

#xxx

Update the proto definitions for the named messages to acept run name as a field or argument.

## How is this patch tested?

Added run names to the RunInfo, CreateRun and UpdateRun proto messages.

Components

Interface

Language

Integrations


mlflow/protos/service.proto
```diff
@@ -835,6 +838,9 @@ message CreateRun {
   // MLflow release. Use 'mlflow.user' tag instead.
   optional string user_id = 2;
 
+  // Name of the run.
+  optional string run_name = 3;
```
dbczumar, line 842: For safety, can we make this `optional string run_name = 10`? I don't know if we accidentally deleted items from our proto instead of using the `reserve` keyword (oof), but I suspect that `3` may have previously been used for something. Can you check git history to make sure that 10 is safe?

mlflow/protos/service.proto
```diff
@@ -625,6 +625,9 @@ message RunInfo {
   // be removed in a future MLflow version.
   optional string run_uuid = 1;
 
+  // The name of the run.
+  optional string run_name = 3;
```
dbczumar, line 629: Ditto with https://github.com/mlflow/mlflow/pull/6769/files#r969066361. Can we use 15 and check git history to make sure it's safe? We really shouldn't have ever deleted fields from our protos...

mlflow/protos/service.proto
apurva-koti: Just checked blame. 3 was indeed used...for run name 😂 
I think we should be safe to use 3 given that any uses of that byte will be for name. Ditto for CreateRun.
For UpdateRun, neither 5 nor name were ever used, so we're fine there.

mlflow/protos/service.proto
apurva-koti: https://github.com/mlflow/mlflow/pull/6769#discussion_r969068536

dbczumar:
@mlflow-automation autoformat

dbczumar (approved):
LGTM!

PR #6782, opened 2022-09-13
Add run name field to MLflow create/get/update API implementations for all stores

apurva-koti:

## Related Issues/PRs

#6798

Add a `run_name` parameter to the implementations of the `CreateRun` `UpdateRun` and `RunInfo` proto messages. This affects all stores (file, sql and REST). Accordingly, add parameters up the chain to the fluent API at all call sites of these functions.
Accordingly, we _stop_ populating the run name in the tags and instead use the new parameter.
The implementation of `UpdateRun` in the file and SQL stores will _not_ update the run name if the passed argument is `None`. (This is to ensure run termination can occur safely without overwriting run name). This is not the case with the REST store, where it is up to the backend to determine whether to overwrite the name or not.
The majority of the diff is fixing tests that were breaking due to missing parameters.

## How is this patch tested?
Ensure tests work, and write tests for new behavior (setting and getting run name)

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/java/client/src/test/java/org/mlflow/tracking/MlflowClientTest.java
+++ b/mlflow/java/client/src/test/java/org/mlflow/tracking/MlflowClientTest.java
@@ -544,7 +544,7 @@ public class MlflowClientTest {
       Arrays.asList(1L, -5L, 4L, 4L));
 
     List<RunTag> tags = run.getData().getTagsList();
-    Assert.assertEquals(tags.size(), 2);
+    Assert.assertEquals(tags.size(), 1);
     assertTag(tags, "user_email", USER_EMAIL);
   }
 
--- a/mlflow/store/tracking/abstract_store.py
+++ b/mlflow/store/tracking/abstract_store.py
@@ -176,7 +176,7 @@ class AbstractStore:
         pass
 
     @abstractmethod
-    def update_run_info(self, run_id, run_status, end_time, name):
+    def update_run_info(self, run_id, run_status, end_time, run_name):
         """
         Update the metadata of the specified run.
 
@@ -185,7 +185,7 @@ class AbstractStore:
         pass
 
     @abstractmethod
-    def create_run(self, experiment_id, user_id, start_time, tags, name):
+    def create_run(self, experiment_id, user_id, start_time, tags, run_name):
         """
         Create a run under the specified experiment ID, setting the run's status to "RUNNING"
         and the start time to the current time.
--- a/mlflow/store/tracking/abstract_store.py
+++ b/mlflow/store/tracking/abstract_store.py
@@ -176,7 +176,7 @@ class AbstractStore:
         pass
 
     @abstractmethod
-    def update_run_info(self, run_id, run_status, end_time):
+    def update_run_info(self, run_id, run_status, end_time, name):
         """
         Update the metadata of the specified run.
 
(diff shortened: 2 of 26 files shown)


mlflow/java/client/src/test/java/org/mlflow/tracking/MlflowClientTest.java
```diff
@@ -544,7 +544,7 @@ public void checkParamsAndMetrics() {
       Arrays.asList(1L, -5L, 4L, 4L));
 
     List<RunTag> tags = run.getData().getTagsList();
-    Assert.assertEquals(tags.size(), 2);
+    Assert.assertEquals(tags.size(), 1);
```
apurva-koti, line 547: This was recently changed to 2 because we now expect an autogenerated name tag. Back to 1 because that name stays out of the tags :)

mlflow/store/tracking/abstract_store.py
```diff
@@ -176,7 +176,7 @@ def get_run(self, run_id):
         pass
 
     @abstractmethod
-    def update_run_info(self, run_id, run_status, end_time):
+    def update_run_info(self, run_id, run_status, end_time, name):
```
BenWilson2, line 179: Should the naming stay consistent throughout? `run_name`

mlflow/store/tracking/abstract_store.py
```diff
@@ -185,7 +185,7 @@ def update_run_info(self, run_id, run_status, end_time):
         pass
 
     @abstractmethod
-    def create_run(self, experiment_id, user_id, start_time, tags):
+    def create_run(self, experiment_id, user_id, start_time, tags, name):
```
BenWilson2, line 188: Ditto here https://github.com/mlflow/mlflow/pull/6782/files#r974607012

mlflow/store/tracking/dbmodels/models.py
```diff
@@ -198,6 +198,7 @@ def to_mlflow_entity(self):
         run_info = RunInfo(
             run_uuid=self.run_uuid,
             run_id=self.run_uuid,
+            run_name=self.name,
```
BenWilson2, line 201: And here https://github.com/mlflow/mlflow/pull/6782/files#r974607012 for `run_name` instead of `name`?

BenWilson2:
Is there a plan to reconcile the pre-existing reserved run_name tag entry to return run info attribute `run_name` that resolves to the older methodology or are we just considering this to be a functional 2.0 change that has no backwards compatibility with legacy runs?

apurva-koti:
@BenWilson2 old run names saved as tags will be returned in the RunInfo attribute in upcoming PRs.

mlflow/store/tracking/dbmodels/models.py
apurva-koti: Will update throughout to be as consistent as possible. In this case `name` was a pre-existing field so I tried to minimize diff, but can update that one too if you like.

BenWilson2: It seems like it makes for the least amount of confusion (since experiments have names too) :)

dbczumar (approved):
LGTM! Can we make sure to track introducing these run name capabilities to the R and Java clients? These can be done via separate PRs in the future.

mlflow/store/tracking/dbmodels/models.py
apurva-koti: Reverting this one change because it required a schema change which feels excessive for this PR. Everything else is standardized to `run_name`

BenWilson2 (approved):
LGTM! awesome stuff @apurva-koti !

PR #6730, opened 2022-09-08
Added ADLS Gen2 support to mlflow

kriscon-db:
## Related Issues/PRs

#xxx

## What changes are proposed in this pull request?

(Please fill in changes proposed in this fix)

## How is this patch tested?

I have manually tested the low level changes in this notebook:

https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/2387091059360226/command/2600307357263946

- [ ] I have written tests (not required for typo or doc fix) and confirmed the proposed feature/bug-fix/change works.

## Does this PR change the documentation?

- [ ] No. You can skip the rest of this section.
- [ ] Yes. Make sure the changed pages / sections render correctly by following the steps below.

1. Click the `Details` link on the `Preview docs` check.
2. Find the changed pages / sections and make sure they render correctly.

### Is this a user-facing change?

- [ ] No. You can skip the rest of this section.
- [ ] Yes. Give a description of this change to be included in the release notes for MLflow users.

(Details in 1-2 sentences. You can just refer to another PR with a description if this PR is part of a larger change.)

### What component(s), interfaces, languages, and integrations does this PR affect?
Components
- [x] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/pipelines`: Pipelines, Pipeline APIs, Pipeline configs, Pipeline Templates
- [ ] `area/projects`: MLproject format, project running backends
- [ ] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

Interface
- [ ] `area/uiux`: Front-end, user experience, plotting, JavaScript, JavaScript dev server
- [ ] `area/docker`: Docker use across MLflow's components, such as MLflow Projects and MLflow Models
- [ ] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry
- [ ] `area/windows`: Windows support

Language
- [ ] `language/r`: R APIs and clients
- [ ] `language/java`: Java APIs and clients
- [ ] `language/new`: Proposals for new client languages

[...]

PR #6845, opened 2022-09-21
Add unit test for DatabricksArtifactRepository uploading artifacts through ADL Gen2

mingyu89:
Signed-off-by: Mingyu Li <mingyu.li@databricks.com>
## Changes 
**Added unit test**
Screen Shot 2022-09-20 at 5 59 19 PM

**Minor improvement to avoid fetching credentials more times than needed when initial credentials expires**

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/store/artifact/databricks_artifact_repo.py
+++ b/mlflow/store/artifact/databricks_artifact_repo.py
@@ -211,10 +211,10 @@ class DatabricksArtifactRepository(ArtifactRepository):
                             "Failed to authorize request, possibly due to credential expiration."
                             " Refreshing credentials and trying again..."
                         )
-                        credential_info = self._get_write_credential_infos(
+                        credentials = self._get_write_credential_infos(
                             run_id=self.run_id, paths=[artifact_path]
                         )[0]
-                        put_block(credential_info.signed_uri, block_id, chunk, headers=headers)
+                        put_block(credentials.signed_uri, block_id, chunk, headers=headers)
                     else:
                         raise e
                 uploading_block_list.append(block_id)
@@ -226,12 +226,10 @@ class DatabricksArtifactRepository(ArtifactRepository):
                         "Failed to authorize request, possibly due to credential expiration."
                         " Refreshing credentials and trying again..."
                     )
-                    credential_info = self._get_write_credential_infos(
+                    credentials = self._get_write_credential_infos(
                         run_id=self.run_id, paths=[artifact_path]
                     )[0]
-                    put_block_list(
-                        credential_info.signed_uri, uploading_block_list, headers=headers
-                    )
+                    put_block_list(credentials.signed_uri, uploading_block_list, headers=headers)
                 else:
                     raise e
         except Exception as err:
@@ -253,10 +251,10 @@ class DatabricksArtifactRepository(ArtifactRepository):
                         "Failed to authorize ADLS file creation request, possibly due "
                         "to credential expiration. Refreshing credentials and trying again..."
                     )
-                    credential_info = self._get_write_credential_infos(
+                    credentials = self._get_write_credential_infos(
                         run_id=self.run_id, paths=[artifact_path]
                     )[0]
-                    put_adls_file_creation(credential_info.signed_uri, headers=headers)
+                    put_adls_file_creation(credentials.signed_uri, headers=headers)
                 else:
                     raise e
 
@@ -272,11 +270,11 @@ class DatabricksArtifactRepository(ArtifactRepository):
                             "Failed to authorize ADLS patch request, possibly due to "
                             "credential expiration. Refreshing credentials and trying again..."
                         )
-                        credential_info = self._get_write_credential_infos(
+                        credentials = self._get_write_credential_infos(
                             run_id=self.run_id, paths=[artifact_path]
                         )[0]
                         patch_adls_file_upload(
-                            credential_info.signed_uri, chunk, offset, headers=headers
+                            credentials.signed_uri, chunk, offset, headers=headers
                         )
                     else:
                         raise e
@@ -290,10 +288,10 @@ class DatabricksArtifactRepository(ArtifactRepository):
                         "Failed to authorize flush request, possibly due to credential expiration."
                         " Refreshing credentials and trying again..."
                     )
-                    credential_info = self._get_write_credential_infos(
+                    credentials = self._get_write_credential_infos(
                         run_id=self.run_id, paths=[artifact_path]
                     )[0]
-                    patch_adls_flush(credential_info.signed_uri, offset, headers=headers)
+                    patch_adls_flush(credentials.signed_uri, offset, headers=headers)
                 else:
                     raise e
         except Exception as err:
--- a/mlflow/store/artifact/databricks_artifact_repo.py
+++ b/mlflow/store/artifact/databricks_artifact_repo.py
@@ -211,10 +211,10 @@ class DatabricksArtifactRepository(ArtifactRepository):
                             "Failed to authorize request, possibly due to credential expiration."
                             " Refreshing credentials and trying again..."
                         )
-                        credentials = self._get_write_credential_infos(
+                        credential_info = self._get_write_credential_infos(
                             run_id=self.run_id, paths=[artifact_path]
                         )[0]
-                        put_block(credentials.signed_uri, block_id, chunk, headers=headers)
+                        put_block(credential_info.signed_uri, block_id, chunk, headers=headers)
                     else:
                         raise e
                 uploading_block_list.append(block_id)
@@ -226,10 +226,12 @@ class DatabricksArtifactRepository(ArtifactRepository):
                         "Failed to authorize request, possibly due to credential expiration."
                         " Refreshing credentials and trying again..."
                     )
-                    credentials = self._get_write_credential_infos(
+                    credential_info = self._get_write_credential_infos(
                         run_id=self.run_id, paths=[artifact_path]
                     )[0]
-                    put_block_list(credentials.signed_uri, uploading_block_list, headers=headers)
+                    put_block_list(
+                        credential_info.signed_uri, uploading_block_list, headers=headers
+                    )
                 else:
                     raise e
         except Exception as err:
@@ -251,10 +253,10 @@ class DatabricksArtifactRepository(ArtifactRepository):
                         "Failed to authorize ADLS file creation request, possibly due "
                         "to credential expiration. Refreshing credentials and trying again..."
                     )
-                    credentials = self._get_write_credential_infos(
+                    credential_info = self._get_write_credential_infos(
                         run_id=self.run_id, paths=[artifact_path]
                     )[0]
-                    put_adls_file_creation(credentials.signed_uri, headers=headers)
+                    put_adls_file_creation(credential_info.signed_uri, headers=headers)
                 else:
                     raise e
 
@@ -270,11 +272,11 @@ class DatabricksArtifactRepository(ArtifactRepository):
                             "Failed to authorize ADLS patch request, possibly due to "
                             "credential expiration. Refreshing credentials and trying again..."
                         )
-                        credentials = self._get_write_credential_infos(
+                        credential_info = self._get_write_credential_infos(
                             run_id=self.run_id, paths=[artifact_path]
                         )[0]
                         patch_adls_file_upload(
-                            credentials.signed_uri, chunk, offset, headers=headers
+                            credential_info.signed_uri, chunk, offset, headers=headers
                         )
                     else:
                         raise e
@@ -288,10 +290,10 @@ class DatabricksArtifactRepository(ArtifactRepository):
                         "Failed to authorize flush request, possibly due to credential expiration."
                         " Refreshing credentials and trying again..."
                     )
-                    credentials = self._get_write_credential_infos(
+                    credential_info = self._get_write_credential_infos(
                         run_id=self.run_id, paths=[artifact_path]
                     )[0]
-                    patch_adls_flush(credentials.signed_uri, offset, headers=headers)
+                    patch_adls_flush(credential_info.signed_uri, offset, headers=headers)
                 else:
                     raise e
         except Exception as err:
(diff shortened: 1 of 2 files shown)


tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,89 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+    ):
+        mock_azure_headers = {
+            "x-ms-content-type": "test-type",
+            "x-ms-owner": "some-owner",
+            "x-ms-blob-type": "some-type",
+        }
+        filtered_azure_headers = {
+            "x-ms-content-type": "test-type",
+            "x-ms-owner": "some-owner",
+        }
```
liangz1, line 316: This seems to be testing the existing header removal logic for Gen 1 URLs. Shall we add tests for Gen 2 headers handling? cc @kriscon-db

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,89 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+            write_credential_infos_mock.assert_called_with(
+                run_id=MOCK_RUN_ID, paths=[expected_location]
+            )
+            request_mock.assert_called_with(
+                "patch",
+                MOCK_ADL_GEN2_SIGNED_URI + "?action=flush&position=14",
+                headers=filtered_azure_headers,
+            )
+
+    def test_log_artifact_adl_gen2_blob_client_sas_error(self, databricks_artifact_repo, test_file):
```
liangz1, line 345: It seems that we don't need this because Gen 1 uses blob endpoint and Gen 2 uses dfs endpoint. Do we need a counterpart `test_log_artifact_adl_gen2_dfs_client_sas_error`?

tests/store/artifact/test_databricks_artifact_repo.py
liangz1: It seems that we don't have special handling logic for Gen 2 headers. The current test should be sufficient.

mingyu89: This is for the v2 header. It's different from v1.

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,89 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
                 databricks_artifact_repo.log_artifact(test_file.strpath)
             write_credential_infos_mock.assert_called_with(run_id=MOCK_RUN_ID, paths=ANY)
 
+    @pytest.mark.parametrize(
+        ("artifact_path", "expected_location"),
+        [(None, "test.txt"), ("output", "output/test.txt"), ("", "test.txt")],
+    )
+    def test_log_artifact_adl_gen2(
```
kriscon-db, line 282: We should also test that the file chunking is behaving correctly and that the proper number of PATCH's are called when we have to chunk.

tests/store/artifact/test_databricks_artifact_repo.py
kriscon-db, line 345: So the idea here is to test that we raise if we send blob store url/headers to the gen2 code path?

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,89 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+            DATABRICKS_ARTIFACT_REPOSITORY + "._get_write_credential_infos"
+        ) as write_credential_infos_mock, mock.patch(
+            "azure.storage.blob.BlobClient.from_blob_url"
+        ) as mock_create_blob_client:
+            mock_credential_info = ArtifactCredentialInfo(
+                signed_uri=MOCK_ADL_GEN2_SIGNED_URI,
+                type=ArtifactCredentialType.AZURE_ADLS_GEN2_SAS_URI,
+            )
+            write_credential_infos_mock.return_value = [mock_credential_info]
+            mock_create_blob_client.side_effect = MlflowException("MOCK ERROR")
```
liangz1, line 356: To confirm with test author, but we probably don't need this test case.

jinzhang21 (commented):
I'll review after @liangz1 and @kriscon-db 's initial comments are addressed. Thanks!

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -211,10 +211,10 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
                             "Failed to authorize request, possibly due to credential expiration."
                             " Refreshing credentials and trying again..."
                         )
-                        credential_info = self._get_write_credential_infos(
+                        credentials = self._get_write_credential_infos(
```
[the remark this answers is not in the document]
mingyu89: It's to reuse it on line 222 instead of getting it again. 
If the credential expires, there's no need to get it in both put_block nad put_block_list steps.

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: Modified test

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: Modified test

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: Updated

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: Modified test to return success for put and append and return error for flush.

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,130 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+        ) as write_credential_infos_mock, mock.patch(
+            "mlflow.utils.rest_utils.cloud_storage_http_request",
+            side_effect=[mock_successful_response, mock_successful_response, mock_error_response],
+        ) as request_mock:
+            mock_credential_info = ArtifactCredentialInfo(
+                signed_uri=MOCK_ADLS_GEN2_SIGNED_URI,
+                type=ArtifactCredentialType.AZURE_ADLS_GEN2_SAS_URI,
+            )
+            write_credential_infos_mock.return_value = [mock_credential_info]
+            with pytest.raises(MlflowException, match=r".+"):
```
jinzhang21, line 387: Can we match on a more precise error message here?

jinzhang21 (approved):

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: We could. It's basically our mocked exception message "MOCK ERROR"

mingyu89: I see. The place I copied from is not "MOCK ERROR" and must use ".+".

jinzhang21: Right. If we don't match the message, it could be raised from other places, which is not what you want to test.

tests/store/artifact/test_databricks_artifact_repo.py
liangz1: We don't use header filtering logic for Gen 2 URLs.  The `x-ms-blob-type` header is only added for Gen 1 URLs. See https://src.dev.databricks.com/databricks/universe@229ded1437202b91079a80fb7867b5e8e7fee983/-/blob/common/cloudstorage/MultiCloudPresignedUrlGenerator.scala?L108. This header is explicitly added by `MultiCloudPresignedUrlGenerator`.

I think you can just use:
```python
        mock_azure_headers = {
            "x-ms-encryption-key": "test-key",
            "x-ms-encryption-algorithm": "AES256",
        }
```
and remove `filtered_azure_headers`.

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,130 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+            databricks_artifact_repo.log_artifact(test_file.strpath, artifact_path)
+            write_credential_infos_mock.assert_called_with(
+                run_id=MOCK_RUN_ID, paths=[expected_location]
+            )
+            azure_upload_mock.assert_called_with(
+                mock_credential_info, test_file.strpath, expected_location
+            )
+
+    @pytest.mark.parametrize(("artifact_path", "expected_location"), [(None, "test.txt")])
+    def test_log_artifact_adls_gen2_with_headers(
```
liangz1, line 305: Can you rename `test_log_artifact_adls_gen2_with_headers` -> `test_log_artifact_adls_gen2_put_patch_flush`?

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,130 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+                data=ANY,
+                headers=filtered_azure_headers,
+            )
+            request_mock.assert_called_with(
+                "patch",
+                MOCK_ADLS_GEN2_SIGNED_URI + "?action=flush&position=14",
+                headers=filtered_azure_headers,
+            )
+
+    def test_log_artifact_adls_gen2_flush_error(self, databricks_artifact_repo, test_file):
```
liangz1, line 371: Do you mean `test_log_artifact_adls_gen2_patch_error`? The mocked error occurs at the second `"patch"` call while the `"flush"` is the fourth `"patch"` call.

liangz1 (changes requested):
Thanks for working on this! Let me know if my comments make sense.

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: I'm explicitly testing these lines of code 
https://github.com/mlflow/mlflow/pull/6730/files#diff-efd79f313ceede37a7b002e5bfcb679e7eecab3451cc5a16eae5c635f9334336R30-R33
It's client side header filtering logic and not in the context of production put/patch requests.

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: This test is different from the previous one.
The previous test has this mock and the current test doesn't have it.
```
request_mock, mock.patch(
            "mlflow.store.artifact.databricks_artifact_repo._AZURE_MAX_BLOCK_CHUNK_SIZE",
            5,
        )
```

The total size is 14.
The previous test will send 5 requests because the max chunk size is 5
- put
- patch(append-position0)
- patch(append-position5)
- patch(append-position10)
- patch(flush-position14)

The current test will send 3 requests because the max chunk size is 100000000
- put, 
- patch(append-position0)
- patch(flush-position14)

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: For the "with headers" questions, I replied here https://github.com/mlflow/mlflow/pull/6845#discussion_r978051812.
github didn't make it appear in all places it should be.

mingyu89:
Reverted production change and only leave test change in the PR so that we know it's not the cause of the example failures.

tests/store/artifact/test_databricks_artifact_repo.py
liangz1: I see. So actually the issue is that the header `"x-ms-blob-type"` is specially handled for Gen1 URLs (see details below). To avoid further confusion, I would suggest not mentioning this header in the mock date and using:
```
mock_adl2_headers = { 
    "x-ms-encryption-key": "test-key", 
    "x-ms-encryption-algorithm": "AES256", 
    "non-adl2-header": "some-value"
}

filtered_adl2_headers = {
    "x-ms-encryption-key": "test-key", 
    "x-ms-encryption-algorithm": "AES256",
 }
```
**Details:**

1. CP [adds](https://src.dev.databricks.com/databricks/universe@70d7bb8052a714be0512369ede75a7c1c9ba56e7/-/blob/common/cloudstorage/MultiCloudPresignedUrlGenerator.scala?L108:14) `"x-ms-blob-type"` header explicitly in the presigned URL generator.
2. MLflow Backend immediately [removes this header](https://src.dev.databricks.com/databricks/universe@0a13390b0deb4a0f05fbe29f899dd9f09faf4c3e/-/blob/mlflow/artifacts/main/MlflowArtifactManager.scala?L281) due to MLflow client incompatibility issue.
3. MLflow Client (current) [again removes this header](https://src.dev.databricks.com/mlflow/mlflow@786ecc41704e9a7dbd07e9801c2af14e3e89a782/-/blob/mlflow/azure/client.py?L36) (which shouldn't exist) and [adds this header back](https://src.dev.databricks.com/mlflow/mlflow@786ecc41704e9a7dbd07e9801c2af14e3e89a782/-/blob/mlflow/azure/client.py?L31) before sending the `put_block` request.

The removal of `"x-ms-blob-type"` header already happens in the [backend](https://src.dev.databricks.com/databricks/universe@0a13390b0deb4a0f05fbe29f899dd9f09faf4c3e/-/blob/mlflow/artifacts/main/MlflowArtifactManager.scala?L281). So in the real world, this header shouldn't appear in the return value of `_get_write_credential_infos`.

tests/store/artifact/test_databricks_artifact_repo.py
liangz1: Got it. I didn't notice `_AZURE_MAX_BLOCK_CHUNK_SIZE` is using it's default value here. Would you mind adding a comment before L395 to make it explicit & increase readability?
```
Using the default _AZURE_MAX_BLOCK_CHUNK_SIZE, the file is uploaded in a single patch request.
```

liangz1 (commented):
Thanks for clarifying! The tests look good to me. I left some comments for better readability.

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: Sure. I was testing sometime supported by V1 but not by V2.

tests/store/artifact/test_databricks_artifact_repo.py
mingyu89: Sure

liangz1 (approved):
LGTM, thanks!

Issue #6369, opened 2022-07-29
[BUG] MLflow Pipelines cannot get active spark session in Jupyter notebook


jerrylian-db:
### Willingness to contribute

No. I cannot contribute a bug fix at this time.

### MLflow version

1.27.1.dev0

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 16.04)**: Mac OS
- **Python version**: 3.7.13
- **yarn version, if running the dev UI**:

### Describe the problem

I tried running the mlp regression pipeline locally using a Jupyter notebook and ingesting data from a source that requires spark. I started a spark session within the notebook. However, the ingest step continues to fail due to not being able to get an active spark session.

### Code to reproduce issue

Download and setup the example MLP regression repo.

Use spark to read the sample parquet file and write it to a single delta file at `"./data/sample_table.delta"`. Then,
Set the local profile to
```
experiment:
  name: "sklearn_regression_experiment"
  tracking_uri: "sqlite:///metadata/mlflow/mlruns.db"
  artifact_location: "./metadata/mlflow/mlartifacts"
INGEST_DATA_LOCATION: "./data/sample_table.delta"
INGEST_DATA_FORMAT: delta
SPLIT_RATIOS: [0.80, 0.10, 0.10]
```
In the provided Jupyter notebook, run the following to start a spark session
```
from pyspark.sql import SparkSession

session = (
    SparkSession.builder.master("local[*]")
    .config("spark.jars.packages", "io.delta:delta-core_2.12:1.2.1")
    .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
    .config(
        "spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog"
    )
    .getOrCreate()
)
```
Finally, run `p.run("ingest")` in the notebook.

### What component(s) does this bug affect?

- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [X] `area/pipelines`: Pipelines, Pipeline APIs, Pipeline configs, Pipeline Templates
- [ ] `area/projects`: MLproject format, project running backends
[...]

PR #6846, opened 2022-09-21
Spark session jupyter

prithvikannan:
## Related Issues/PRs

https://github.com/mlflow/mlflow/issues/6369

Creates a new spark session when spark is not available (such as when run on a subprocess via make).

## How is this patch tested?

Fixing a bug where MLflow pipeline commands run via subprocess were not able to access a spark session by creating a new spark session in the subprocess.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/utils/_spark_utils.py
+++ b/mlflow/utils/_spark_utils.py
@@ -17,6 +17,23 @@ def _get_active_spark_session():
         return SparkSession._instantiatedSession
 
 
+def _create_spark_session():
+    try:
+        from pyspark.sql import SparkSession
+    except ImportError:
+        # Return None if user doesn't have PySpark installed
+        return None
+    return (
+        SparkSession.builder.master("local[*]")
+        .config("spark.jars.packages", "io.delta:delta-core_2.12:1.2.1")
+        .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
+        .config(
+            "spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog"
+        )
+        .getOrCreate()
+    )
+
+
 class _SparkDirectoryDistributor:
     """Distribute spark directory from driver to executors."""
 
--- a/mlflow/utils/_spark_utils.py
+++ b/mlflow/utils/_spark_utils.py
@@ -17,7 +17,9 @@ def _get_active_spark_session():
         return SparkSession._instantiatedSession
 
 
-def _create_spark_session():
+def _create_local_spark_session_for_pipelines():
+    """Create a sparksession to be used within an pipeline step run in a subprocess locally."""
+
     try:
         from pyspark.sql import SparkSession
     except ImportError:
--- a/mlflow/pipelines/steps/ingest/datasets.py
+++ b/mlflow/pipelines/steps/ingest/datasets.py
@@ -508,7 +508,7 @@ class _SparkDatasetMixin:
             else:
                 spark_session = _create_spark_session()
                 _logger.info("Creating new spark session")
-                return spark_session
+            return spark_session
         except Exception as e:
             raise MlflowException(
                 message=(
--- a/mlflow/pipelines/steps/ingest/datasets.py
+++ b/mlflow/pipelines/steps/ingest/datasets.py
@@ -26,7 +26,7 @@ from mlflow.utils.file_utils import (
     read_parquet_as_pandas_df,
     download_file_using_http_uri,
 )
-from mlflow.utils._spark_utils import _get_active_spark_session
+from mlflow.utils._spark_utils import _get_active_spark_session, _create_spark_session
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


jerrylian-db:
A few question about this approach:
* Will a new spark session be created for each run of the ingest step?
* Will the spark session terminate or stop when the subprocess for the ingest step exits? Will this approach leave a lot of dangling spark sessions on the computer?
* Can you add this logic to the predict step? The predict step also uses a spark session.

mlflow/utils/_spark_utils.py
```diff
@@ -17,6 +17,23 @@ def _get_active_spark_session():
         return SparkSession._instantiatedSession
 
 
+def _create_spark_session():
```
BenWilson2, line 20: This should probably be renamed to be explicit in its purpose. `_create_local_spark_session_for_pipelines()` or something like that. And a solid docstring to explain its purpose since it implements a very restrictive functionality of an otherwise very common task (getting or creating a spark session).

jinzhang21: +1

jinzhang21 (commented):
LGTM overall. Thanks for the quick resolution, @prithvikannan !

@WeichenXu123 is there a way to create global SparkSession on a local machine so that it can be shared / passed around among processes? i.e. a reference-counted SparkSession object

mlflow/utils/_spark_utils.py
prithvikannan: Thanks @BenWilson2 for the feedback. Updated.

prithvikannan:
@jerrylian-db 
> * Will a new spark session be created for each run of the ingest step?

Yes. But this doesn't seem to have too much overhead, so this may be okay. 

> * Will the spark session terminate or stop when the subprocess for the ingest step exits? Will this approach leave a lot of dangling spark sessions on the computer?

I believe these spark sessions will be cleaned up when the subprocess finishes. So this won't leave spark sessions. 

> * Can you add this logic to the predict step? The predict step also uses a spark session.

Done! ✅

jerrylian-db (commented):
Logic looks great to me! Wondering if you could write some tests to verify the functionality of running steps that need spark locally. One idea I have is to run the predict step via the CLI here: https://github.com/mlflow/mlflow/blob/db802333561817f08db6a5f584e4a57ab11abfa4/tests/pipelines/test_cli.py#L14 or run the step via the `pipeline.run` here
https://github.com/mlflow/mlflow/blob/master/tests/pipelines/test_pipeline.py

@jinzhang21 can confirm if this is the right approach? Note that if you hit some bugs with example directory set-up or the predict step, maybe this should be done in a follow-up task instead.

jinzhang21:
> Logic looks great to me! Wondering if you could write some tests to verify the functionality of running steps that need spark locally. One idea I have is to run the predict step via the CLI here:
> 
> https://github.com/mlflow/mlflow/blob/db802333561817f08db6a5f584e4a57ab11abfa4/tests/pipelines/test_cli.py#L14
> 
> or run the step via the `pipeline.run` here
> https://github.com/mlflow/mlflow/blob/master/tests/pipelines/test_pipeline.py
> @jinzhang21 can confirm if this is the right approach? Note that if you hit some bugs with example directory set-up or the predict step, maybe this should be done in a follow-up task instead.

We have more than 1 steps using this functionality. Adding a test in `https://github.com/mlflow/mlflow/blob/master/tests/pipelines/test_utils.py` might make more sense. We can simply assert SparkSession is obtained correctly (either already exists or newly created).

@WeichenXu123 let us know if you have better ideas to persist SparkSession across Python processes. If not, we can merge the change as is.

WeichenXu123 (approved):

WeichenXu123:
> @WeichenXu123 let us know if you have better ideas to persist SparkSession across Python processes. If not, we can merge the change as is.

@jinzhang21 This is not possible. We can make subprocess connect to the same spark cluster used in parent process though.

jinzhang21:
> > @WeichenXu123 let us know if you have better ideas to persist SparkSession across Python processes. If not, we can merge the change as is.
> 
> @jinzhang21 This is not possible. We can make subprocess connect to the same spark cluster used in parent process though.

Thanks, @WeichenXu123 ! Sharing the same cluster does sound useful (e.g. multiple steps don't have to launch Spark cluster in local mode on their own, which may take time). How to make it happen?

WeichenXu123:
> > > @WeichenXu123 let us know if you have better ideas to persist SparkSession across Python processes. If not, we can merge the change as is.
> > 
> > 
> > @jinzhang21 This is not possible. We can make subprocess connect to the same spark cluster used in parent process though.
> 
> Thanks, @WeichenXu123 ! Sharing the same cluster does sound useful (e.g. multiple steps don't have to launch Spark cluster in local mode on their own, which may take time). How to make it happen?

1. first, it requires the spark is cluster mode.
2. get the `spark.master` config in parent process, and connect it in child process. `SparkSession.builder.master("spark://spark_driver_host:port").getOrCreate()

jinzhang21:
> > > > @WeichenXu123 let us know if you have better ideas to persist SparkSession across Python processes. If not, we can merge the change as is.
> > > 
> > > 
> > > @jinzhang21 This is not possible. We can make subprocess connect to the same spark cluster used in parent process though.
> > 
> > 
> > Thanks, @WeichenXu123 ! Sharing the same cluster does sound useful (e.g. multiple steps don't have to launch Spark cluster in local mode on their own, which may take time). How to make it happen?
> 
> 1. first, it requires the spark is cluster mode.
> 2. get the `spark.master` config in parent process, and connect it in child process. `SparkSession.builder.master("spark://spark_driver_host:port").getOrCreate()

Got it. In this case, we can merge this PR as is. The objective is to support Spark in local mode when not running on Databricks. We can think about how to pass the Spark config to individual steps when we solve the large data access universally later.

jinzhang21 (approved):

PR #6860, opened 2022-09-22
Merge train/validation/test profiles into one card in the split step

sunishsheth2009:
Merge train/validation/test profiles into one card in the split step

## How is this patch tested?
- [x] Created a notebook to make sure this works
image

Merging split step profiles into 1 card for easy comparison of profiles

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/steps/ingest/__init__.py
+++ b/mlflow/pipelines/steps/ingest/__init__.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from mlflow.exceptions import MlflowException
 from mlflow.pipelines.cards import BaseCard
 from mlflow.pipelines.step import BaseStep
-from mlflow.pipelines.utils.step import get_pandas_data_profile
+from mlflow.pipelines.utils.step import get_pandas_data_profiles
 from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
 from mlflow.utils.file_utils import read_parquet_as_pandas_df
 from mlflow.pipelines.steps.ingest.datasets import (
@@ -76,8 +76,8 @@ class BaseIngestStep(BaseStep, metaclass=abc.ABCMeta):
         ingested_dataset_profile = None
         if not self.skip_data_profiling:
             _logger.info("Profiling ingested dataset")
-            ingested_dataset_profile = get_pandas_data_profile(
-                ingested_df, "Profile of Ingested Dataset"
+            ingested_dataset_profile = get_pandas_data_profiles(
+                [["Profile of Ingested Dataset", ingested_df]]
             )
             dataset_profile_path = Path(
                 str(os.path.join(output_directory, BaseIngestStep._DATASET_PROFILE_OUTPUT_NAME))
--- a/mlflow/pipelines/steps/split.py
+++ b/mlflow/pipelines/steps/split.py
@@ -8,7 +8,7 @@ from typing import Dict, Any
 from mlflow.pipelines.cards import BaseCard
 from mlflow.pipelines.step import BaseStep
 from mlflow.pipelines.utils.execution import get_step_output_path
-from mlflow.pipelines.utils.step import get_pandas_data_profile
+from mlflow.pipelines.utils.step import get_pandas_data_profiles
 from mlflow.exceptions import MlflowException, INVALID_PARAMETER_VALUE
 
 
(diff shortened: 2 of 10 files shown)


mlflow/pipelines/steps/ingest/__init__.py
```diff
@@ -76,8 +76,8 @@ def _run(self, output_directory: str) -> BaseCard:
         ingested_dataset_profile = None
         if not self.skip_data_profiling:
             _logger.info("Profiling ingested dataset")
-            ingested_dataset_profile = get_pandas_data_profile(
-                ingested_df, "Profile of Ingested Dataset"
+            ingested_dataset_profile = get_pandas_data_profiling_comparison_view(
+                [["Profile of Ingested Dataset", ingested_df]]
```
jinzhang21, line 80: There's only 1 dataset in `ingest`, right? Why do we need to use the comparison view? consistency?

mlflow/pipelines/steps/split.py
```diff
@@ -117,28 +117,17 @@ def _build_profiles_and_card(self, train_df, validation_df, test_df) -> BaseCard
...
             # Tab #1 - #3: data profiles for train/validation and test.
-            card.add_tab("Data Profile (Train)", "{{PROFILE}}").add_pandas_profile(
-                "PROFILE", train_profile
-            )
-            card.add_tab("Data Profile (Validation)", "{{PROFILE}}").add_pandas_profile(
-                "PROFILE", validation_profile
-            )
-            card.add_tab("Data Profile (Test)", "{{PROFILE}}").add_pandas_profile(
-                "PROFILE", test_profile
+            card.add_tab("Data Profile Comparison", "{{PROFILE}}").add_pandas_profile(
```
jinzhang21, line 129: `Data Profile Comparison` -> `Data Profiles (Train/Validation/Test)`

mlflow/pipelines/steps/transform.py
```diff
@@ -127,13 +127,11 @@ def _build_profiles_and_card(
...
-                "Profile of Train Transformed Dataset",
+            train_transformed_profile = get_pandas_data_profiling_comparison_view(
+                [["Profile of Train Transformed Dataset", train_transformed]]
             )
-            validation_transformed_profile = get_pandas_data_profile(
-                validation_transformed,
-                "Profile of Validation Transformed Dataset",
+            validation_transformed_profile = get_pandas_data_profiling_comparison_view(
+                [["Profile of Validation Transformed Dataset", validation_transformed]]
             )
```
jinzhang21, line 135: Shouldn't these 2 profiles be compared in a single tab?

mlflow/pipelines/utils/step.py
```diff
@@ -105,15 +106,28 @@ def _get_pool_size():
     return 1 if "PYTEST_CURRENT_TEST" in os.environ and os.name == "nt" else 0
 
 
-def get_pandas_data_profile(data_frame, title: str):
-    """Returns a data profiling object over input data frame.
+def get_pandas_data_profiling_comparison_view(inputs: Iterable[Tuple[str, pd.DataFrame]]) -> str:
```
jinzhang21, line 109: Let's not call it `_comparison_view` since there's only 1 function exists (as compared to having a separate function `get_pandas_data_profile`). It is confusing when only 1 dataset is provided. `get_pandas_data_profiles` probably is sufficient.

jinzhang21 (commented):
LGTM mostly. See comments below. Thanks, @sunishsheth2009 !

mlflow/pipelines/steps/ingest/__init__.py
sunishsheth2009: We have logic in pandas_renderer that splits the view between compare view v/s single view. So to keep the API clean and have the split logic in one place, we can just pass in an array and let pandas_renderer handle that.

mlflow/pipelines/steps/transform.py
sunishsheth2009: I was waiting for @mengxr comment to confirm if we like to do the same for transform step as well.

mlflow/pipelines/utils/step.py
sunishsheth2009: Sounds good :) 
Let me change it back

jinzhang21 (approved):
LGTM! Please ping XM for the transform step output. IMO we don't even need data profiling for transform step at all. The most useful data profiles are generated in the split step with train/validation/test split compared against each other.

mengxr:
* Shall we rename the step card to "Compare Splits"?
* Shall we rename the label from "Profile of Train Dataset" to "Train Data" or simply "Train"? 
* I think it is useful to keep the profile of transformed train data. Sometimes transformers might create many nulls / zeros in an unexpected way. But we don't need the profile for transformed validation data. It should be used mainly for tuning and evaluation instead. We don't need to do it in this PR.
* In the "worst predictions" step card, it would be nice to include data profile comparison between the worst group and the overall train dataset so it is easy to see any feature stands out. We don't need to do it in this PR either.

sunishsheth2009:
> I think it is useful to keep the profile of transformed train data. Sometimes transformers might create many nulls / zeros in an unexpected way. But we don't need the profile for transformed validation data. It should be used mainly for tuning and evaluation instead. We don't need to do it in this PR.
In the "worst predictions" step card, it would be nice to include data profile comparison between the worst group and the overall train dataset so it is easy to see any feature stands out. We don't need to do it in this PR either.

Changed based on the recommendation from split step. Will create separate PRs/ticket for each of the other 2 suggestions above. Thank you :)

PR #6921, opened 2022-09-28
Add conf to specify spark write / save mode in MLP predict step

jerrylian-db:
Signed-off-by: Jerry Liang <jerry.liang@databricks.com>

## Related Issues/PRs

#xxx

In this PR, we expose the spark write mode as an option in the predict step configuration. When the write mode is default, we add logic to detect early if an overwrite will occur and throw an exception.

## How is this patch tested?

Components

Interface

Language

Integrations


mlflow/pipelines/steps/predict.py
```diff
@@ -107,6 +107,27 @@ def _run(self, output_directory):
...
             )
 
+        # check if output location is already populated as we allow no overwrites
+        output_format = self.step_config["output_format"]
+        output_location = self.step_config["output_location"]
+        if output_format == "parquet" or output_format == "delta":
+            output_populated = os.path.exists(output_location)
+        else:
+            try:
+                output_populated = spark._jsparkSession.catalog().tableExists(output_location)
```
jerrylian-db, line 117: There is a `spark.catalog.tableExists` method, but this method is only available in pyspark 3.3.0. Using this method instead, which is available in spark 2.1.x or later.

kriscon-db (approved):
LGTM.

mlflow/pipelines/steps/predict.py
```diff
@@ -107,6 +107,27 @@ def _run(self, output_directory):
                 error_code=BAD_REQUEST,
             )
 
+        # check if output location is already populated as we allow no overwrites
+        output_format = self.step_config["output_format"]
+        output_location = self.step_config["output_location"]
+        if output_format == "parquet" or output_format == "delta":
+            output_populated = os.path.exists(output_location)
+        else:
```
jinzhang21, line 115: `else if output_format == "spark_sql"` to be more specific on when this code block runs. We could introduce more data format in the future.

mlflow/pipelines/steps/predict.py
```diff
@@ -107,6 +107,27 @@ def _run(self, output_directory):
                 error_code=BAD_REQUEST,
             )
 
+        # check if output location is already populated as we allow no overwrites
+        output_format = self.step_config["output_format"]
+        output_location = self.step_config["output_location"]
+        if output_format == "parquet" or output_format == "delta":
```
jinzhang21, line 113: `if output_format in ["parquet", "delta"]:` for future extensibility.

mlflow/pipelines/steps/predict.py
```diff
@@ -107,6 +107,27 @@ def _run(self, output_directory):
...
+                output_populated = spark._jsparkSession.catalog().tableExists(output_location)
+            except Exception:
+                # swallow spark failures
+                output_populated = False
+        if output_populated:
+            raise MlflowException(
+                message=(
+                    f"Output location `{output_location}` of format `{output_format}` is already "
+                    "populated and overwrites are not allowed. Please clear or modify the output "
+                    "location and try again."
```
jinzhang21, line 126: Is this behavior really desired? I can clear see for Delta table I just want to keep updating the table with new versions. Not sure about Parquet. Do we have customer feedback on whether to overwrite it?

jerrylian-db: We don't have any customer feedback on this. But I remember preventing overwrites being an earlier decision we made in the Batch Scoring development process. @kriscon-db do you remember the rationale?

jerrylian-db: I could see allowing overwrites for the "table" format only, as Delta Tables keep track of versions. (Not possible for parquet files)

kriscon-db: I seem to remember someone, at some point saying they preferred not to allow overwrites from one of our standups awhile ago.  Is there someone we should ask to get a final decision on this behavior?  Bilal? XM?

jinzhang21: I'm inclined to add an option allowing overwrite but default to not allowing it. For Delta tables we should just write with new versions.
Feel free to reach out to XM.

mlflow/pipelines/steps/predict.py
```diff
@@ -129,21 +129,30 @@ def _run(self, output_directory):
...
+        output_location = self.step_config["output_location"]
+        if (
+            not self.force_overwrite
+            and output_format not in ["delta", "table"]
+            and os.path.exists(output_location)
+        ):
             raise MlflowException(
                 message=(
-                    "No active SparkSession detected to score dataset with spark UDF. "
-                    "Please create a Spark session and try again."
```
jerrylian-db, line 146: Removed this outdated exception message.

mlflow/pipelines/steps/predict.py
jerrylian-db: Resolving this thread as we've reached a decision on this.

mlflow/pipelines/steps/predict.py
```diff
@@ -71,6 +70,7 @@ def _validate_and_apply_step_config(self):
         self.tracking_config = TrackingConfig.from_dict(self.step_config)
         self.registry_uri = self.step_config.get("registry_uri", None)
         self.skip_data_profiling = self.step_config.get("skip_data_profiling", False)
+        self.force_overwrite = self.step_config.get("force_overwrite", False)
```
jinzhang21, line 73: `force_overwrite` sounds aggressive. Maybe just call it `overwrite_output`?

Related, sorry for the late suggestion, does it make sense to just adopt the Spark SaveMode option here? Since we always use Spark to save the output, instead of a `True/False` option, we allow user to set all available SaveMode, defaulting to "default".
https://spark.apache.org/docs/latest/sql-data-sources-load-save-functions.html#save-modes

In this case we don't have to think about when to overwrite and when not. Just add a link to Spark in the doc solves the problem.

mlflow/pipelines/steps/predict.py
```diff
@@ -129,21 +129,30 @@ def _run(self, output_directory):
...
+            and output_format not in ["delta", "table"]
+            and os.path.exists(output_location)
+        ):
             raise MlflowException(
                 message=(
-                    "No active SparkSession detected to score dataset with spark UDF. "
-                    "Please create a Spark session and try again."
+                    f"Output location `{output_location}` is already populated. To enable "
+                    "overwrites, please specify a Delta-based output format or set "
+                    "`force_overwrite` to true in the predict step configuration."
```
jinzhang21, line 155: update the comment `force_overwrite`

mlflow/pipelines/steps/predict.py
jerrylian-db: Sounds good. I will add a conf for the spark write mode.

mlflow/pipelines/steps/predict.py
```diff
@@ -129,21 +129,36 @@ def _run(self, output_directory):
...
                 ),
                 error_code=BAD_REQUEST,
             ) from e
-        if not spark:
+
+        # check if output location is already populated for non-delta output formats
+        output_format = self.step_config["output_format"]
+        output_location = self.step_config["output_location"]
+        output_populated = False
+        if self.write_mode == "default":
```
jinzhang21, line 147: There's a synonym for `default`: 
[ErrorIfExists](https://spark.apache.org/docs/1.4.0/api/java/org/apache/spark/sql/SaveMode.html#ErrorIfExists)

Also what do we do with [Ignore](https://spark.apache.org/docs/1.4.0/api/java/org/apache/spark/sql/SaveMode.html#Ignore)?
I think we should print a warning and exist with status 0 (success) in this case for Parquet.

mlflow/pipelines/steps/predict.py
```diff
@@ -129,21 +129,36 @@ def _run(self, output_directory):
...
+            if output_format == "parquet" or output_format == "delta":
```
jinzhang21, line 148: What's the difference between `delta` and `table`? Why is `delta` treated differently here?

jinzhang21 (commented):
LGTM overall. Some comments, see below.

mlflow/pipelines/steps/predict.py
jerrylian-db: `delta` is also a delta table, but you specify the storage location / path for that delta table. With `table`, you don't specify the storage location (I believe it just gets written to the spark warehouse location) and it creates a delta table and adds it to the hive metastore under the specified table name. I'm open to removing combining these two options, now that we have a better understanding of how writing in delta works.

mlflow/pipelines/steps/predict.py
jerrylian-db: I just implemented `default` because I didn't want to re-implement how spark handles save mode options. But it seems like this write mode options enum is small and has been around for a while.

I can add "default" and its synonyms in this PR. I can also rename "write_mode" to "save_mode" to make it more consistent with the Spark API. For the `ignore` case, maybe I can add it in a follow up PR? I'm guessing that I would need to create a new custom result card for this case, all of scoring would be skipped.

jerrylian-db: Actually, would you be comfortable not short circuiting "ignore"? To use this option, a user would need to open the spark SaveMode docs, read that doc, and deliberately set the SaveMode to "ignore" - so they should be aware of the consequences. We can just document that the MLP batch artifacts would be updated but the output location would not. We short-circuit "default" because we know an error will happen during the spark writing, which we can flag early. For "ignore", the spark writing will not have an error.

jinzhang21: SGTM!

jinzhang21 (approved):

PR #6927, opened 2022-09-28
SSE-CPK support added to MLflow

kriscon-db:
## Related Issues/PRs

#xxx

In this PR we:

- Add protobuf definitions to support the new presigned URL type
- Add azure client support for ADLS Gen2 HTTP create, append and flush operations
- Update the artifact repositories that use the current Azure blob store upload/download APIs to be able to handle the new PresignedUrl type and handle ADLS Gen2 URLs.
- Single part and multipart upload for ADLS Gen2 in the databricks artifact repository.
- Threaded encryption headers into all upload and download calls
- Unit tests

## How is this patch tested?

We have added new unit tests to unit test the repository code paths.  We have also manually tested this change on a CPK test shard and manually validated that this feature branch works.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/azure/client.py
+++ b/mlflow/azure/client.py
@@ -38,7 +38,7 @@ def put_adls_file_creation(sas_url, headers):
         rest_utils.augmented_raise_for_status(response)
 
 
-def patch_adls_file_upload(sas_url, data, position, headers):
+def patch_adls_file_upload(sas_url, data, position, headers, is_single):
     """
     Performs an ADLS Azure file create `Patch` operation
     (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/update)
@@ -49,7 +49,10 @@ def patch_adls_file_upload(sas_url, data, position, headers):
     :param position: Positional offset of the data in the Patch request
     :param headers: Additional headers to include in the Patch request body
     """
-    request_url = _append_query_parameters(sas_url, {"action": "append", "position": str(position)})
+    new_params = {"action": "append", "position": str(position)}
+    if is_single:
+        new_params["flush"] = "true"
+    request_url = _append_query_parameters(sas_url, new_params)
 
     request_headers = {}
     for name, value in headers.items():
--- a/mlflow/azure/client.py
+++ b/mlflow/azure/client.py
@@ -40,11 +40,11 @@ def put_adls_file_creation(sas_url, headers):
 
 def patch_adls_file_upload(sas_url, data, position, headers):
     """
-    Performs an ADLS Azure file create `Put` operation
-    (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create)
+    Performs an ADLS Azure file create `Patch` operation
+    (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/update)
 
     :param sas_url: A shared access signature URL referring to the Azure ADLS server
-                    to which the file creation command should be issued.
+                    to which the file update command should be issued.
     :param data: Data to include in the Patch request body.
     :param position: Positional offset of the data in the Patch request
     :param headers: Additional headers to include in the Patch request body
@@ -67,12 +67,12 @@ def patch_adls_file_upload(sas_url, data, position, headers):
 def patch_adls_flush(sas_url, position, headers):
     """
     Performs an ADLS Azure file flush `Patch` operation
-    (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create)
+    (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/update)
 
     :param sas_url: A shared access signature URL referring to the Azure ADLS server
-                    to which the file creation command should be issued.
+                    to which the file update command should be issued.
     :param position: The final size of the file to flush.
-    :param headers: Additional headers to include in the Put request body.
+    :param headers: Additional headers to include in the Patch request body.
     """
     request_url = _append_query_parameters(sas_url, {"action": "flush", "position": str(position)})
 
--- a/mlflow/azure/client.py
+++ b/mlflow/azure/client.py
@@ -14,6 +14,81 @@ _PUT_BLOCK_HEADERS = {
 }
 
 
+def put_adls_file_creation(sas_url, headers):
+    """
+    Performs an ADLS Azure file create `Put` operation
+    (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create)
+
+    :param sas_url: A shared access signature URL referring to the Azure ADLS server
+                    to which the file creation command should be issued.
+    :param headers: Additional headers to include in the Put request body
+    """
+    request_url = _append_query_parameters(sas_url, {"resource": "file"})
+
+    request_headers = {}
+    for name, value in headers.items():
+        if _is_valid_adls_put_header(name):
+            request_headers[name] = value
+        else:
+            _logger.debug("Removed unsupported '%s' header for ADLS Gen2 Put operation", name)
+
+    with rest_utils.cloud_storage_http_request(
+        "put", request_url, headers=request_headers
+    ) as response:
+        rest_utils.augmented_raise_for_status(response)
+
+
+def patch_adls_file_upload(sas_url, data, position, headers):
+    """
+    Performs an ADLS Azure file create `Put` operation
+    (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create)
+
+    :param sas_url: A shared access signature URL referring to the Azure ADLS server
+                    to which the file creation command should be issued.
+    :param data: Data to include in the Patch request body.
+    :param position: Positional offset of the data in the Patch request
+    :param headers: Additional headers to include in the Patch request body
+    """
+    request_url = _append_query_parameters(sas_url, {"action": "append", "position": str(position)})
+
+    request_headers = {}
+    for name, value in headers.items():
+        if _is_valid_adls_patch_header(name):
+            request_headers[name] = value
+        else:
+            _logger.debug("Removed unsupported '%s' header for ADLS Gen2 Patch operation", name)
+
+    with rest_utils.cloud_storage_http_request(
+        "patch", request_url, data=data, headers=request_headers
+    ) as response:
+        rest_utils.augmented_raise_for_status(response)
+
+
+def patch_adls_flush(sas_url, position, headers):
+    """
+    Performs an ADLS Azure file flush `Patch` operation
+    (https://docs.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create)
+
+    :param sas_url: A shared access signature URL referring to the Azure ADLS server
+                    to which the file creation command should be issued.
+    :param position: The final size of the file to flush.
+    :param headers: Additional headers to include in the Put request body.
+    """
+    request_url = _append_query_parameters(sas_url, {"action": "flush", "position": str(position)})
+
+    request_headers = {}
+    for name, value in headers.items():
+        if _is_valid_adls_put_header(name):
+            request_headers[name] = value
+        else:
+            _logger.debug("Removed unsupported '%s' header for ADLS Gen2 Patch operation", name)
+
+    with rest_utils.cloud_storage_http_request(
+        "patch", request_url, headers=request_headers
+    ) as response:
+        rest_utils.augmented_raise_for_status(response)
+
+
 def put_block(sas_url, block_id, data, headers):
     """
     Performs an Azure `Put Block` operation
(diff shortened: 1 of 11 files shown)


jinzhang21 (commented):
Replaced @dbczumar with @harupy to lessen his burden on reviews. Feel free to still take a look if you wish, @dbczumar !

mlflow/spark.py
```diff
@@ -794,6 +792,8 @@ def load_model(model_uri, dfs_tmpdir=None, dst_path=None):
         ):
             return PipelineModel.load(mlflowdbfs_path)
 
+    local_model_path = _download_artifact_from_uri(artifact_uri=model_uri, output_path=dst_path)
+    _add_code_from_conf_to_system_path(local_model_path, flavor_conf)
```
liangz1, line 796: This change is not related to supporting ADLS Gen 2.
I moved these two lines after checking mlflowdbfs because mlflowdbfs doesn't need `_download_artifact_from_uri` to download the model to the driver's local filesystem. The model will be downloaded by mlflowdbfs. So without this code change, the model will be downloaded twice, which I think is unnecessarily harming the performance.

However, it looks like the `local_model_path` is needed by `_add_code_from_conf_to_system_path`, which should be called before loading the model. Could anyone help me understand in which scenario would Spark model need to use the custom code path? I looked at the test [here](https://src.dev.databricks.com/mlflow/mlflow/-/blob/tests/spark/test_spark_model_export.py?L829)  and I still don't get why we need to add the `__file__` to the system path.

If we do need to support this, we should REVERT the code change in this file.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -231,6 +240,85 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
...
+        Uploads a file to a given Azure storage location using the ADLS gen2 API.
+        """
+        try:
+            headers = self._extract_headers_from_credentials(credentials.headers)
+
+            # try to create the file
+            try:
+                put_adls_file_creation(credentials.signed_uri, headers=headers)
+            except requests.HTTPError as e:
+                if e.response.status_code in [401, 403]:
```
jinzhang21, line 254: Please make sure these are the only codes will be returned on authorization failure scenarios.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -231,6 +240,85 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
...
+                    _logger.info(
+                        "Failed to authorize ADLS file creation request, possibly due "
+                        "to credential expiration. Refreshing credentials and trying again..."
+                    )
+                    credential_info = self._get_write_credential_infos(
+                        run_id=self.run_id, paths=[artifact_path]
+                    )[0]
+                    put_adls_file_creation(credential_info.signed_uri, headers=headers)
```
jinzhang21, line 262: Under what circumstances does this special handling of refreshing credentials will actually work? or why does it fail in the first place that we only need to retry exactly once to resolve?

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -231,6 +240,85 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
...
+                    if e.response.status_code in [401, 403]:
+                        _logger.info(
+                            "Failed to authorize ADLS patch request, possibly due to "
+                            "credential expiration. Refreshing credentials and trying again..."
+                        )
+                        credential_info = self._get_write_credential_infos(
+                            run_id=self.run_id, paths=[artifact_path]
+                        )[0]
+                        patch_adls_file_upload(
+                            credential_info.signed_uri,
```
jinzhang21, line 294: Shouldn't `credentials` be reset as `credentials_infos` here before retrying? What happens for the next chunk?

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -231,6 +240,85 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
...
+                else:
+                    raise e
+
+            # next try to patch the file
+            offset = 0
+            is_first_chunk = True
+            use_single_part_upload = False
+            for chunk in yield_file_in_chunks(local_file, _AZURE_MAX_BLOCK_CHUNK_SIZE):
+                try:
+                    cur_offset = len(chunk)
```
jinzhang21, line 272: Calling the length of a chunk `offset` is really confusing. Maybe `chunk_size`?

mlflow/store/artifact/databricks_artifact_repo.py
liangz1: In my understanding, presigned urls will expire after 300s or so (maybe 900s, I don't remember the exact number), if there are many chunks and the uploading takes longer than that, the `put` will fail. In this case, getting a new presigned url for write will resolve this issue.

mlflow/store/artifact/databricks_artifact_repo.py
liangz1: It seems that all the occurrences of `credentials_infos` in this function should be replaced with `credentials`.

mlflow/store/artifact/databricks_artifact_repo.py
jinzhang21: OK. This makes sense to me.

jinzhang21: @kriscon-db Can we try to abstract away this try-except block with an explicit aim to retry with refreshed credentials? It's very repetitive w/o clear indication of its functionality. Something like `def retry_with_credential_refresh(fn, credentials):`, or make it a decorator if you want.

mlflow/store/artifact/databricks_artifact_repo.py
kriscon-db: It actually looks like there are no 401's returned?  https://learn.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/create#response

Reading through these codes, the one that would trigger on a potential token expiration is one of the 403's.  I'll change to limit to just 403.

mlflow/store/artifact/databricks_artifact_repo.py
kriscon-db: Yep.  Let me push a fix.

mlflow/store/artifact/databricks_artifact_repo.py
kriscon-db: Will rename to `cursor_position` and rename `cur_offset` to `cur_chunk_size`

mlflow/store/artifact/databricks_artifact_repo.py
kriscon-db: Will fix.

mlflow/store/artifact/databricks_artifact_repo.py
kriscon-db: Please re-review this section.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -231,6 +240,72 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
         except Exception as err:
             raise MlflowException(err)
 
+    def _retryable_adls_function(self, func, artifact_path, **kwargs):
```
kriscon-db, line 243: To avoid the potential repeated calls to `_get_write_credential_infos`, we could have this method return the final credentials used and the calling points would overwrite their credentials with what is returned.  I am not a fan of this because it is tricky/clever code and tricky/clever code takes awhile to understand when other engineers have to read it.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -231,6 +240,71 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
...
+                self._retryable_adls_function(
+                    func=patch_adls_file_upload,
+                    artifact_path=artifact_path,
+                    sas_url=credentials.signed_uri,
+                    data=chunk,
+                    position=cursor_position,
+                    headers=headers,
+                    is_single=use_single_part_upload,
+                )
+                is_first_chunk = False
```
harupy, line 293:
```suggestion
            use_single_part_upload = False
            for idx, chunk in enumerate(yield_file_in_chunks(local_file, _AZURE_MAX_BLOCK_CHUNK_SIZE):
                cur_chunk_size = len(chunk)
                if idx == 0 and cur_chunk_size < _AZURE_MAX_BLOCK_CHUNK_SIZE:
                    use_single_part_upload = True
                self._retryable_adls_function(
                    func=patch_adls_file_upload,
                    artifact_path=artifact_path,
                    sas_url=credentials.signed_uri,
                    data=chunk,
                    position=cursor_position,
                    headers=headers,
                    is_single=use_single_part_upload,
                )
```

Can we use `enumerate` and remove `is_first_chunk`?

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,132 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+                headers=filtered_azure_headers,
+            )
+
+    def test_log_artifact_adls_gen2_flush_error(self, databricks_artifact_repo, test_file):
+        mock_successful_response = Response()
+        mock_successful_response.status_code = 200
+        mock_successful_response.close = lambda: None
+        mock_error_response = MlflowException("MOCK ERROR")
+        with mock.patch(
+            DATABRICKS_ARTIFACT_REPOSITORY + "._get_write_credential_infos"
```
harupy, line 378:
```suggestion
            DATABRICKS_ARTIFACT_REPOSITORY + "._get_write_credential_infos",
            return_value=[...]
```

Nit: we can specify `return_value` when calling `mock.patch`

tests/store/artifact/test_databricks_artifact_repo.py
```diff
@@ -274,6 +275,132 @@ def test_log_artifact_azure_blob_client_sas_error(self, databricks_artifact_repo
...
+                MOCK_ADLS_GEN2_SIGNED_URI + "?resource=file",
+                headers={},
+            )
+            # test with azure max block size
+            request_mock.assert_any_call(
+                "patch",
+                MOCK_ADLS_GEN2_SIGNED_URI + "?action=append&position=0&flush=true",
+                data=ANY,
+                headers={},
+            )
```
harupy, line 402:
```suggestion
            assert request_mock.mock_calls == [
                mock.call(
                    "put",
                    MOCK_ADLS_GEN2_SIGNED_URI + "?resource=file",
                    headers={},
                ),
                # test with azure max block size
                mock.call(
                    "patch",
                    MOCK_ADLS_GEN2_SIGNED_URI + "?action=append&position=0&flush=true",
                    data=ANY,
                    headers={},
                ),
            ]
```

to ensure the number of requests and order.

mlflow/azure/client.py
```diff
@@ -14,6 +14,84 @@
...
+
+    :param sas_url: A shared access signature URL referring to the Azure ADLS server
+                    to which the file update command should be issued.
+    :param data: Data to include in the Patch request body.
+    :param position: Positional offset of the data in the Patch request
+    :param headers: Additional headers to include in the Patch request body
+    """
+    new_params = {"action": "append", "position": str(position)}
+    if is_single:
+        new_params["flush"] = "true"
```
harupy, line 54: Why do we set `flush` to `true` only when we have a single chunk? What happens if we have multiple chunks and set `flush` to `true` in the final append?


```
append
append
append with flush
```

kriscon-db: I believe this is a shortcut that microsoft added to reduce the number of calls in a single part upload.  My understanding is this only works when the position is 0 for a single part upload.

mlflow/store/artifact/databricks_artifact_repo.py
```diff
@@ -231,6 +240,71 @@ def _azure_upload_file(self, credentials, local_file, artifact_path):
...
+            if e.response.status_code in [403]:
+                _logger.info(
+                    "Failed to authorize ADLS operation, possibly due "
+                    "to credential expiration. Refreshing credentials and trying again..."
+                )
+                new_credentials = self._get_write_credential_infos(
+                    run_id=self.run_id, paths=[artifact_path]
+                )[0]
+                kwargs['sas_url'] = new_credentials.signed_uri
+                put_adls_file_creation(**kwargs)
```
jinzhang21, line 257: `put_adls_file_creation` -> `func`?

jinzhang21 (approved):
LGTM on a high level. I didn't go through every piece of the code though. Please wait for @harupy 's LGTM before merging.

mlflow/store/artifact/databricks_artifact_repo.py
kriscon-db: Will change.

tests/store/artifact/test_databricks_artifact_repo.py
kriscon-db: Will change

tests/store/artifact/test_databricks_artifact_repo.py
kriscon-db: Fixing

harupy (approved):
LGTM!

mlflow/spark.py
kriscon-db: I believe this is the only open question.  cc @jinzhang21 who may have context on this?

kriscon-db:
After talking with @liangz1, I'm reverting the spark.py change.  We can revisit this in a followup commit.

PR #6965, opened 2022-10-03
Make each MLP step declare its own output

bbarnes52:

This PR makes each pipeline step declare its own output and introduces a generic implementation of `get_artifact` that iterates through the steps for a given pipeline to identify and return the request artifact. This will facilitate better code reuse as we introduce additional templates. See also: [the open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle).

## How is this patch tested?

[MLP regression template](https://github.com/mlflow/mlp-regression-template) executed locally.

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/artifacts.py
+++ b/mlflow/pipelines/artifacts.py
@@ -1,0 +1,200 @@
+from abc import ABC, abstractmethod
+import json
+import logging
+import os
+
+import mlflow
+from mlflow.pipelines.utils.execution import get_step_output_path
+from mlflow.tracking import MlflowClient
+from mlflow.tracking._tracking_service.utils import _use_tracking_uri
+
+_logger = logging.getLogger(__name__)
+
+
+class Artifact(ABC):
+    @abstractmethod
+    def name(self):
+        pass
+
+    @abstractmethod
+    def path(self):
+        pass
+
+    @abstractmethod
+    def load(self):
+        pass
+
+
+class DataframeArtifact(Artifact):
+    def __init__(self, name, pipeline_root, step_name, rel_path=""):
+        self._name = name
+        self._path = get_step_output_path(pipeline_root, step_name, rel_path)
+        self._step_name = step_name
+
+    def name(self):
+        return self._name
+
+    def path(self):
+        return self._path
+
+    def load(self):
+        import pandas as pd
+
+        if os.path.exists(self._path):
+            return pd.read_parquet(self._path)
+        log_artifact_not_found_warning(self._name, self._step_name)
+        return None
+
+
+class ModelArtifact(Artifact):
+    def __init__(self, name, pipeline_root, step_name, tracking_uri):
+        self._name = name
+        self._path = get_step_output_path(pipeline_root, step_name, "model/model.pkl")
+        self._pipeline_root = pipeline_root
+        self._step_name = step_name
+        self._tracking_uri = tracking_uri
+
+    def name(self):
+        return self._name
+
+    def path(self):
+        return self._path
+
+    def load(self):
+        run_id = read_run_id(self._pipeline_root)
+        if run_id:
+            with _use_tracking_uri(self._tracking_uri, self._pipeline_root):
+                return mlflow.pyfunc.load_model(f"runs:/{run_id}/{self._step_name}/model")
+        log_artifact_not_found_warning(self._name, self._step_name)
+        return None
+
+
+class TransformerArtifact(Artifact):
+    def __init__(self, name, pipeline_root, step_name, tracking_uri):
+        self._name = name
+        self._path = get_step_output_path(pipeline_root, step_name, "transformer.pkl")
+        self._pipeline_root = pipeline_root
+        self._step_name = step_name
+        self._tracking_uri = tracking_uri
+
+    def name(self):
+        return self._name
+
+    def path(self):
+        return self._path
+
+    def load(self):
+        run_id = read_run_id(self._pipeline_root)
+        if run_id:
+            with _use_tracking_uri(self._tracking_uri, self._pipeline_root):
+                return mlflow.sklearn.load_model(f"runs:/{run_id}/{self._step_name}/transformer")
+        log_artifact_not_found_warning(self._name, self._step_name)
+        return None
+
+
+class RunArtifact(Artifact):
+    def __init__(self, name, pipeline_root, step_name, tracking_uri):
+        self._name = name
+        self._path = get_step_output_path(pipeline_root, step_name, "run_id")
+        self._pipeline_root = pipeline_root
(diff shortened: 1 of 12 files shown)


mlflow/pipelines/artifacts.py
```diff
@@ -0,0 +1,149 @@
+import json
```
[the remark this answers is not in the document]
bbarnes52: If this is your preference, I'm more than happy to oblige. A few comments/concerns:
1. There is not much opportunity for code reuse here, as far as I can see. 
2. My preference is to prefer composition over inheritance unless there is good reason to the contrary, and I'm not sure that's the case here. What I'd like to do is define a Protocol (https://peps.python.org/pep-0544/) for an artifact and have each artifact implement the protocol, I originally did so in this PR but our CI environment seems to run a python version < 3.8 (which is the minimum required version for protocols).

mlflow/pipelines/artifacts.py
```diff
@@ -0,0 +1,149 @@
...
+
+import mlflow
+from mlflow.pipelines.utils.execution import get_step_output_path
+from mlflow.tracking import MlflowClient
+from mlflow.tracking._tracking_service.utils import _use_tracking_uri
+
+_logger = logging.getLogger(__name__)
+
+
+class DataframeArtifact:
```
prithvikannan, line 13: Should we make an abstract `Artifact` class object? This way we can enforce that `load()` is implemented.

tests/pipelines/test_pipeline.py
```diff
@@ -266,27 +266,22 @@ def test_pipeline_get_artifacts():
...
-    assert isinstance(pipeline.get_artifact("training_data"), pd.DataFrame)
-    assert isinstance(pipeline.get_artifact("validation_data"), pd.DataFrame)
-    assert isinstance(pipeline.get_artifact("test_data"), pd.DataFrame)
-    assert isinstance(pipeline.get_artifact("transformed_training_data"), pd.DataFrame)
-    assert isinstance(pipeline.get_artifact("transformed_validation_data"), pd.DataFrame)
-    assert hasattr(pipeline.get_artifact("transformer"), "transform")
-    assert isinstance(pipeline.get_artifact("model"), mlflow.pyfunc.PyFuncModel)
-    assert isinstance(pipeline.get_artifact("run"), Run)
-    assert isinstance(pipeline.get_artifact("registered_model_version"), ModelVersion)
+    assert isinstance(pipeline.get_artifact("ingested_data").load(), pd.DataFrame)
```
prithvikannan, line 269: Since we are changing the behavior of `get_artifact` I think we should document this. Can we add a release note, and maybe change the wording of the `get_artifact` docstring?

prithvikannan (approved):
Awesome change @bbarnes52! Left a couple small comments.

mlflow/pipelines/artifacts.py
bbarnes52: Sure thing - @jinzhang21 this overlaps a tad with your feedback, is this a reasonable first step here? This avoids increasing the scope of the refactor, and we can unify the various artifact classes in the mlflow codebase separately, if desired?

jinzhang21: We can merge your change as is w/o doing the extra step. But I want to get us aligned on the overall directions and the "big picture" here, namely, do we want Artifact class to exist in MLflow in the near future, what's the pros and cons of having it etc. Maybe a small design doc clarifying this decision is a better place for discussion. No need to do everything in this PR.

jinzhang21 (approved):

tests/pipelines/test_pipeline.py
sunishsheth2009: Is there a reason we are making this backwards compact change?
We might need to update the notebook examples in template and example repo for regression template.

mlflow/pipelines/pipeline.py
```diff
@@ -213,19 +212,14 @@ def get_artifact(self, artifact_name: str):
...
-        `Pipeline.inspect()` or `Pipeline.run()` output.
-
-        Returns None if the specified artifact is not found.
-        Raise an error if the artifact is not supported.
-        """
-        pass
+        for step in self._steps:
+            for artifact in step.get_artifacts():
+                if artifact.name == artifact_name:
+                    return artifact
```
sunishsheth2009, line 218: Can we do return artifact.load()

sunishsheth2009 (approved):

tests/pipelines/test_pipeline.py
bbarnes52: Good point, yes the notebooks would need to be updated. My understanding is that the cost of breaking changes is small given lack of MLP usage, so the fact that it's a breaking change didn't really affect the direction of this PR. Do you disagree with that sentiment and/or do you prefer the previous interface?

mlflow/pipelines/pipeline.py
prithvikannan: @sunishsheth2009 i believe that if we return the artifact content here, then we would not be able to re-use this function to get artifact path via CLI.

tests/pipelines/test_pipeline.py
sunishsheth2009: I would surely prefer the previous interface since we don't need to call load additional from users perspective. 

Users just need to call `pipeline.get_artifact("ingested_data")` for example. Why is load important? 

Unless we can plan to add more function like
`pipeline.get_artifact("ingested_data").clean()`, `pipeline.get_artifact("ingested_data").update()`, etc.
For me adding the load doesn't make a lot of sense for the added complexity for the user. Just my thoughts.


What do you guys think?

mlflow/pipelines/pipeline.py
bbarnes52: If folks prefer, we can rename this method to `_get_artifact` and then implement `get_artifact` as follows:
```
get_artifact(self, name):
    return self._get_artifact(name).load()
```
My 2c would be to do this only if this interface is preferable (quite possible!) rather than to avoid a breaking change. IMO it depends on whether users would ever want other information about an artifact, such as its path.

sunishsheth2009: Ah yes! That makes sense. But we could go back to similar pattern we had previously of `get_artifact` and `_get_artifact_path`  or something that where we have similar logic which could be abstracted out as well. 
But compromising the User experience for that wouldn't be ideal IMO.

tests/pipelines/test_pipeline.py
sunishsheth2009: Thank you for making this change. Love the refactor part :)

sunishsheth2009 (approved):

Issue #6970, opened 2022-10-04
[FR] Allowing to configure timeout with mlflow.pyfunc.spark_udf interface


alonisser:
### Willingness to contribute

Yes. I can contribute this feature independently.

### Proposal Summary

Following on adding this ability with this [issue](https://github.com/mlflow/mlflow/issues/5381) adding this configuration on model "serving"  I propose adding am ability to configure the default (60 seconds) timeout also on the spark_udf interface. 

### Motivation

> #### What is the use case for this feature?
Allowing the run with spark_udf actions that might take more the 60 seconds - such as "narrow transformations" (clustering for example) where reducing the batch size isn't possible. 
> #### Why is this use case valuable to support for MLflow users in general?
spark_udf is probably a major interface to mlflow usage, and having the ability to control timeouts, allows it to be used in scenarios that need more control
> #### Why is this use case valuable to support for your project(s) or organization?
We are are hitting a problem where a "clustering" algo takes for bit batches more the 60 seconds, and getting killed by the network timeout. Our interface is via spark and spark_udf so the current configuration option doesn't help us

> #### Why is it currently difficult to achieve this use case?
There is no way to configure this, and there isn't a "fool proof" work around

### Details

Adding a config dict to spark_udf interface, defaults to {} which would be passed inside, eventually into pyfunc_backend.serve (as the other issue [commit](https://github.com/mlflow/mlflow/commit/e40493a03088e2f939a9d5f0e409207871e983a1) does for serving) 
Another option might be to allow "environment variable" configuration" . I'm open to suggestions

### What component(s) does this bug affect?

- [ ] `area/artifacts`: Artifact stores and artifact logging
- [ ] `area/build`: Build and test infrastructure for MLflow
- [ ] `area/docs`: MLflow documentation pages
- [ ] `area/examples`: Example code
- [ ] `area/model-registry`: Model Registry service, APIs, and the fluent client calls for Model Registry
- [ ] `area/models`: MLmodel format, model serialization/deserialization, flavors
- [ ] `area/pipelines`: Pipelines, Pipeline APIs, Pipeline configs, Pipeline Templates
- [ ] `area/projects`: MLproject format, project running backends
- [X] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs
- [ ] `area/server-infra`: MLflow Tracking server backend
- [ ] `area/tracking`: Tracking Service, tracking client APIs, autologging

### What interface(s) does this bug affect?

[...]

PR #6988, opened 2022-10-06
environment variables support for timeout

alonisser:
Signed-off-by: alonisser <alonisser@gmail.com>

## Related Issues/PRs

#6970 

Add environment variable under mlflow.environment_variables and reference it in the scoring server

## How is this patch tested?

Not sure on how to test this, I've looked at the current tests and I didn't see a relevant example. 

- [ X] No. You can skip the rest of this section.

Allow user to configure via environment variable in the deployed mlflow model the scoring server timeout

Components
- [X] `area/scoring`: MLflow Model server, model deployment tools, Spark UDFs

Interface

Language

Integrations

### DIFF
--- a/mlflow/environment_variables.py
+++ b/mlflow/environment_variables.py
@@ -176,7 +176,7 @@ MLFLOW_REQUIREMENTS_INFERENCE_TIMEOUT = _EnvironmentVariable(
     "MLFLOW_REQUIREMENTS_INFERENCE_TIMEOUT", int, 120
 )
 
-#: Configuring the scoring server request timeout via environment variable
+#: Specifies the MLflow Model Scoring server request timeout in seconds
 #: (default: ``60``)
 MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT = _EnvironmentVariable(
     "MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT", int, 60
--- a/mlflow/environment_variables.py
+++ b/mlflow/environment_variables.py
@@ -175,3 +175,9 @@ MLFLOW_SQLALCHEMYSTORE_POOLCLASS = _EnvironmentVariable(
 MLFLOW_REQUIREMENTS_INFERENCE_TIMEOUT = _EnvironmentVariable(
     "MLFLOW_REQUIREMENTS_INFERENCE_TIMEOUT", int, 120
 )
+
+#: Configuring the scoring server request timeout via environment variable
+#: (default: ``60``)
+MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT = _EnvironmentVariable(
+    "MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT", int, 60
+)
--- a/mlflow/pyfunc/scoring_server/__init__.py
+++ b/mlflow/pyfunc/scoring_server/__init__.py
@@ -46,7 +46,7 @@ except ImportError:
     from mlflow.pyfunc import load_pyfunc as load_model
 from mlflow.protos.databricks_pb2 import BAD_REQUEST
 from mlflow.server.handlers import catch_mlflow_exception
-
+from mlflow.environment_variables import MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT
 from io import StringIO
 
 _SERVER_MODEL_PATH = "__pyfunc_model_path__"
@@ -365,7 +365,7 @@ def get_cmd(
     model_uri: str, port: int = None, host: int = None, timeout: int = None, nworkers: int = None
 ) -> Tuple[str, Dict[str, str]]:
     local_uri = path_to_local_file_uri(model_uri)
-    timeout = timeout or 60
+    timeout = timeout or MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT.get()
     # NB: Absolute windows paths do not work with mlflow apis, use file uri to ensure
     # platform compatibility.
     if os.name != "nt":
--- a/mlflow/pyfunc/scoring_server/__init__.py
+++ b/mlflow/pyfunc/scoring_server/__init__.py
@@ -23,6 +23,8 @@ import pandas as pd
 import sys
 import traceback
 
+from mlflow.environment_variables import MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT
+
 # NB: We need to be careful what we import form mlflow here. Scoring server is used from within
 # model's conda environment. The version of mlflow doing the serving (outside) and the version of
 # mlflow in the model's conda environment (inside) can differ. We should therefore keep mlflow
@@ -364,6 +366,7 @@ def get_cmd(
     model_uri: str, port: int = None, host: int = None, timeout: int = None, nworkers: int = None
 ) -> Tuple[str, Dict[str, str]]:
     local_uri = path_to_local_file_uri(model_uri)
+    timeout = timeout or MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT.get()
     # NB: Absolute windows paths do not work with mlflow apis, use file uri to ensure
     # platform compatibility.
     if os.name != "nt":
--- a/mlflow/pyfunc/scoring_server/__init__.py
+++ b/mlflow/pyfunc/scoring_server/__init__.py
@@ -46,7 +46,6 @@ except ImportError:
     from mlflow.pyfunc import load_pyfunc as load_model
 from mlflow.protos.databricks_pb2 import BAD_REQUEST
 from mlflow.server.handlers import catch_mlflow_exception
-from mlflow.environment_variables import MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT
 from io import StringIO
 
 _SERVER_MODEL_PATH = "__pyfunc_model_path__"
@@ -365,7 +364,6 @@ def get_cmd(
     model_uri: str, port: int = None, host: int = None, timeout: int = None, nworkers: int = None
 ) -> Tuple[str, Dict[str, str]]:
     local_uri = path_to_local_file_uri(model_uri)
-    timeout = timeout or MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT.get()
     # NB: Absolute windows paths do not work with mlflow apis, use file uri to ensure
     # platform compatibility.
     if os.name != "nt":
--- a/mlflow/utils/cli_args.py
+++ b/mlflow/utils/cli_args.py
@@ -156,7 +156,11 @@ PORT = click.option(
 )
 
 TIMEOUT = click.option(
-    "--timeout", "-t", help="Timeout in seconds to serve a request (default: 60)."
+    "--timeout",
+    "-t",
+    envvar="MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT",
+    default=60,
+    help="Timeout in seconds to serve a request (default: 60).",
 )
 
 # We use None to disambiguate manually selecting "4"
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -573,7 +573,7 @@ def _load_model_or_server(model_uri: str, env_manager: str):
         model_uri=local_path,
         port=server_port,
         host="127.0.0.1",
-        timeout=60,
+        timeout=MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT.get(),
         enable_mlserver=False,
         synchronous=False,
         stdout=subprocess.PIPE,
--- a/mlflow/pyfunc/__init__.py
+++ b/mlflow/pyfunc/__init__.py
@@ -225,6 +225,7 @@ import yaml
 
 import mlflow
 import mlflow.pyfunc.model
+from mlflow.environment_variables import MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT
 from mlflow.exceptions import MlflowException
 from mlflow.models import Model, ModelSignature, ModelInputExample
 from mlflow.models.model import MLMODEL_FILE_NAME
@@ -1076,7 +1077,7 @@ def spark_udf(spark, model_uri, result_type="double", env_manager="local"):
                 model_uri=local_model_path_on_executor,
                 port=server_port,
                 host="127.0.0.1",
-                timeout=60,
+                timeout=MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT.get(),
                 enable_mlserver=False,
                 synchronous=False,
                 stdout=subprocess.PIPE,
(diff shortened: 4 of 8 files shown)


mlflow/environment_variables.py
```diff
@@ -170,3 +170,9 @@ def get(self):
 MLFLOW_REQUIREMENTS_INFERENCE_TIMEOUT = _EnvironmentVariable(
     "MLFLOW_REQUIREMENTS_INFERENCE_TIMEOUT", int, 120
 )
+
+#: Configuring the scoring server request timeout via environment variable
```
dbczumar, line 174:
```suggestion
#: Specifies the MLflow Model Scoring server request timeout in seconds
```

dbczumar (commented):
@alonisser This is fantastic! A few small questions:

1. Can we replace https://github.com/mlflow/mlflow/blob/96f861b9f32a8e6ad2473db3b3272a50f5f8e71a/mlflow/models/cli.py#L31 with 

```
click.option(
    "--timeout", "-t", help="Timeout in seconds to serve a model scoring request (default: 60).",
    envvar="MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT",
)
```
? This way, the MLflow CLI will clearly document that it respects the timeout environment variable.

2. Do we need to set this environment variable from within the Pandas UDF created by `mlflow.pyfunc.spark_udf`? (https://github.com/mlflow/mlflow/blob/96f861b9f32a8e6ad2473db3b3272a50f5f8e71a/mlflow/pyfunc/__init__.py#L1030-L1032)

As far as I'm aware, these environment variables from the driver aren't automatically propagated to the UDF.

3. Should we define a similar environment variable for the scoring server readiness check timeout? https://github.com/mlflow/mlflow/blob/b08aa03ed35f76ee143022ac2d535576fecfaeb2/mlflow/pyfunc/__init__.py#L584

mlflow/environment_variables.py
alonisser: Thanks . would do

alonisser:
1. I'll check if that propagates to the same var, not sure
2. I'm not aware of that, although I thought anyway we'd need to setup environment variables in all nodes and not just in the driver, is there a place where the udf environment is built? can you please point to where?
3. I'm not aware of this need, is there a need? Should I add it for completeness' sake?

alonisser:
About 1, I see it uses 
```python
        server_implementation = mlserver if enable_mlserver else scoring_server

```
What is mlserver and should I enable this change there also? or is it a stand alone package?

alonisser:
Pushing a fix for 1 , I understand now as I fixed that I need to change the place the timeout is checked vs envvars to support mlserver also, so changed it . still investigating 2, I'd be happy for pointers if possible

mlflow/environment_variables.py
alonisser: fixed

alonisser:
Following 2, As far as a I see the code, it's either handled magically by spark (probably not) or we need to setup environment variables in the executors too for that to work (What Is probably what we'll need to do anyway), it can't be handled from inside mlflow creating the udf afaik

Also following 2. I see my changes about scoring server didn't influence the actual scoring server "inside" the udf, so I've implemented the change there also 

Appreciate the help!

alonisser:
About 2, I see the code takes care of copying environment variables inside the subprocess process

https://github.com/mlflow/mlflow/blob/96f861b9f32a8e6ad2473db3b3272a50f5f8e71a/mlflow/pyfunc/scoring_server/__init__.py#L398

dbczumar:
> About 2, I see the code takes care of copying environment variables inside the subprocess process
> 
> https://github.com/mlflow/mlflow/blob/96f861b9f32a8e6ad2473db3b3272a50f5f8e71a/mlflow/pyfunc/scoring_server/__init__.py#L398

Awesome! Great find!

mlflow/utils/cli_args.py
```diff
@@ -156,7 +156,11 @@ def _resolve_env_manager(ctx, _, env_manager):
 )
 
 TIMEOUT = click.option(
-    "--timeout", "-t", help="Timeout in seconds to serve a request (default: 60)."
+    "--timeout",
+    "-t",
+    envvar="MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT",
```
dbczumar, line 161: This timeout is used in other CLIs such as `mlflow server`, which are not related to Model serving. Can we instead make the change in https://github.com/mlflow/mlflow/blob/96f861b9f32a8e6ad2473db3b3272a50f5f8e71a/mlflow/models/cli.py#L31?

dbczumar (approved):
LGTM once https://github.com/mlflow/mlflow/pull/6988/files#r991708001 is addressed. I think we can hold off on the scoring server setup timeout for now. Thanks @alonisser!

mlflow/utils/cli_args.py
alonisser: It's the same one I believe :) . `cli_args.TIMEOUT` references the other one. and semantically it's all the same "timeout for the scoring server"

But note, since this isn't passed into the spark_udf implementation, I've added a direct reference to the env var there also

mlflow/pyfunc/scoring_server/__init__.py
```diff
@@ -365,7 +364,6 @@ def get_cmd(
     model_uri: str, port: int = None, host: int = None, timeout: int = None, nworkers: int = None
 ) -> Tuple[str, Dict[str, str]]:
     local_uri = path_to_local_file_uri(model_uri)
-    timeout = timeout or 60
```
dbczumar, line 368: I think the following edit is necessary since `timeout` is `None` by default, which would break line 370:

```
timeout = timeout or MLFLOW_SCORING_SERVER_REQUEST_TIMEOUT.get()
```

mlflow/utils/cli_args.py
dbczumar: My mistake - this is, indeed the same arg. Thanks @alonisser !

dbczumar (approved):
LGTM once https://github.com/mlflow/mlflow/pull/6988/files#r993997278 is addressed. Thanks so much, @alonisser !

mlflow/pyfunc/scoring_server/__init__.py
alonisser: I've removed that because the code calling it is either from 
evaluate for model
https://github.com/zencity/mlflow/blob/6f2231b3cd4be7ad8558d4a61e4220668e2529c6/mlflow/pyfunc/__init__.py#L576 (yes I've found another entry point to the scoring server)

Or the inner pyspark udf 

https://github.com/zencity/mlflow/blob/6f2231b3cd4be7ad8558d4a61e4220668e2529c6/mlflow/pyfunc/__init__.py#L1080

Or from

https://github.com/zencity/mlflow/blob/6f2231b3cd4be7ad8558d4a61e4220668e2529c6/mlflow/pyfunc/backend.py#L162

I'd rather keep it outside, so we'll support future additions to other scoring backends, mlserver supporting timeout (currently doesn't) etc. 

if not, let's put None, in all of those places, and add the line you suggested, but I don't think it's future proof

alonisser: @dbczumar

dbczumar: Hi @alonisser , I'd rather put None in other places and put the line I suggested here. The reason is that, otherwise, if this method is called with the default `timeout=None`, it will break. It should not be the responsibility of each call site to supply a valid value of this argument else face an error, since this argument is optional.

alonisser:
Got it. Will do

בתאריך יום ו׳, 14 באוק׳ 2022, 7:39, מאת Corey Zumar ‏<
***@***.***>:

mlflow/pyfunc/scoring_server/__init__.py
alonisser: Done, I've kept two places also in the calling sites (removed one) as the original was hard coded to 60 and reverting to that is problematic. So we're both explicit in the calling site and ensured inside the scoring server logic @dbczumar

alonisser: @dbczumar

dbczumar (approved):
LGTM! Thanks so much, @alonisser !

alonisser:
So @dbczumar I think the added tests exposed an issue with mlserver with my changes, I'll explore it
But some tests failing seems to be unrelated to my changes, can you please verify if this is a recurring issue and a test problem or related to this PR?

https://github.com/mlflow/mlflow/actions/runs/3242324978/jobs/5374541707#step:7:916

alonisser:
Well I've looked into it

While the second ("models") failure is harder to understand (with gunicorn wrapped in docker) and still trying to recreate. The first one is because of the mlserver "scoring" backend not supporting timeout. and throwing an exception (in mlflow code, not mlserver) about not supporting it. Now that I've added the default value it gets the timeout and throws

There are two options
1. Remove the default value (while keeping it "documented" as it was before) and add it in another place inside the scoring server as a last line of defence default (as it can't get None) 

https://github.com/zencity/mlflow/blob/6f2231b3cd4be7ad8558d4a61e4220668e2529c6/mlflow/utils/cli_args.py#L162

Which in my view isn't great, as documentation becomes misleading, and add 60 in multiple places hard coded

2. remove the exception and just log and error and delete the "timeout" argument if passed

WDYT @dbczumar

dbczumar:
> Well I've looked into it
> 
> While the second ("models") failure is harder to understand (with gunicorn wrapped in docker) and still trying to recreate. The first one is because of the mlserver "scoring" backend not supporting timeout. and throwing an exception (in mlflow code, not mlserver) about not supporting it. Now that I've added the default value it gets the timeout and throws
> 
> There are two options
> 
> 1. Remove the default value (while keeping it "documented" as it was before) and add it in another place inside the scoring server as a last line of defence default (as it can't get None)
> 
> https://github.com/zencity/mlflow/blob/6f2231b3cd4be7ad8558d4a61e4220668e2529c6/mlflow/utils/cli_args.py#L162
> 
> Which in my view isn't great, as documentation becomes misleading, and add 60 in multiple places hard coded
> 
> 2. remove the exception and just log and error and delete the "timeout" argument if passed
> 
> WDYT @dbczumar

Hi @alonisser, thank you for investigating! I agree that (2) is the best option :)

alonisser:
@dbczumar I've pushed the discussed fix, I believe It would fix "MLflow tests / python (pull_request)" so I'd appreciate if you approve a rerun to validate that , what I'm wondering is about "MLflow tests / models (pull_request) " Which I couldn't recreate locally, if they would still fail, I'd need help to setup this locally, as those tests currently fail locally on completely different errors then in the CI

alonisser:
Also I've intergrated the changes merged into master since I've forked

alonisser:
@dbczumar I've managed to run the previously failed tests locally and they pass now. The problem is I don't think they would work in the CI, due to this commit being merged yesterday

https://github.com/mlflow/mlflow/commit/f76de49b49142a2f41eb478b3f30807fc1cbbff7#diff-535d7991b2be4fed9e50b08954351ac640d5be083e8ca0bae614a78f44e66e9bR5

the docker build image command (being tested there) tries to do a pip install of mlflow with 2.0.0dev version. but this version doesn't exist in PYPI yet. Not something I can solve from "my side". Can't figure out if any new PR would actually pass this until it's fixed

dbczumar:
> @dbczumar I've managed to run the previously failed tests locally and they pass now. The problem is I don't think they would work in the CI, due to this commit being merged yesterday
> 
> [f76de49#diff-535d7991b2be4fed9e50b08954351ac640d5be083e8ca0bae614a78f44e66e9bR5](https://github.com/mlflow/mlflow/commit/f76de49b49142a2f41eb478b3f30807fc1cbbff7#diff-535d7991b2be4fed9e50b08954351ac640d5be083e8ca0bae614a78f44e66e9bR5)
> 
> the docker build image command (being tested there) tries to do a pip install of mlflow with 2.0.0dev version. but this version doesn't exist in PYPI yet. Not something I can solve from "my side". Can't figure out if any new PR would actually pass this until it's fixed

Hi @alonisser, I've re-triggered tests against this PR and plan to merge unless there's a clear connection between any failures and the PR (I'd imagine any failures will be completely unrelated, as you've pointed out). Thanks for bearing with us and following through with this excellent contribution!

dbczumar:
> Thanks @dbczumar ! but I see there is a test checking the exception I've removed
> 
> FAILED tests/pyfunc/test_mlserver.py::test_raise_error_mlserver_timeout[params0-Timeout is not yet supported in MLServer.]
> 
> So please let me first remove this test and then can be merged - would it be preferred to keep the test and skip it or just remove it?

I think removing the test is perfect! :)

alonisser:
> > Thanks @dbczumar ! but I see there is a test checking the exception I've removed
> > FAILED tests/pyfunc/test_mlserver.py::test_raise_error_mlserver_timeout[params0-Timeout is not yet supported in MLServer.]
> > So please let me first remove this test and then can be merged - would it be preferred to keep the test and skip it or just remove it?
> 
> I think removing the test is perfect! :)

Done. thanks

dbczumar:
> > > Thanks @dbczumar ! but I see there is a test checking the exception I've removed
> > > FAILED tests/pyfunc/test_mlserver.py::test_raise_error_mlserver_timeout[params0-Timeout is not yet supported in MLServer.]
> > > So please let me first remove this test and then can be merged - would it be preferred to keep the test and skip it or just remove it?
> > 
> > 
> > I think removing the test is perfect! :)
> 
> Done. thanks

Awesome! Let me know once you've pushed the changes and I can re-trigger tests :)

alonisser:
Sorry @dbczumar forgot to push before. pushed now.

alonisser:
Rebased with latest changes from repo, frankly I don't know thy the R test is failing, and don't know how to check that and how is the R server actually related to my changes

dbczumar:
> Rebased with latest changes from repo, frankly I don't know thy the R test is failing, and don't know how to check that and how is the R server actually related to my changes

At a glance, I'm not sure either. I'll dig in and get this fixed very shortly :)

dbczumar:
Hi @alonisser , I've filed https://github.com/zencity/mlflow/pull/1 against your PR branch, which should resolve the issue. If it looks okay to you, can you merge it? Thanks so much!

dbczumar:
> > Hi @alonisser , I've filed [zencity#1](https://github.com/zencity/mlflow/pull/1) against your PR branch, which should resolve the issue. If it looks okay to you, can you merge it? Thanks so much!
> 
> Wow thanks! , I've merged

Thanks so much, @alonisser ! Awesome PR! Merging! :)

alonisser:
Hey, How do I know when this would be released? What's the lifecycle of releases on the mlflow project?

PR #6992, opened 2022-10-07
Change experiment_id from monotonically increasing to random fixed length integer

BenWilson2:

## Related Issues/PRs

#xxx

Convert experiment_id creation logic from incremental index id to a fixed-length (18 digits) int to prevent threading issues in concurrent experiment creation. 

## How is this patch tested?

Components

Interface

Language

Integrations


mlflow/store/tracking/dbmodels/initial_models.py
```diff
@@ -43,14 +43,14 @@
...
-    DB model for :py:class:`mlflow.entities.Experiment`. These are recorded in ``experiment`` table.
+    DB model for :py:class:`mlflow.entities.Experiment`. These are recorded in ``experiments`` table.
     """
 
     __tablename__ = "experiments"
 
     experiment_id = Column(Integer, autoincrement=True)
     """
-    Experiment ID: `Integer`. *Primary Key* for ``experiment`` table.
+    Experiment ID: `Integer`. *Primary Key* for ``experiments`` table.
```
BenWilson2, line 53: typo fix

tests/tracking/test_rest_tracking.py
```diff
@@ -604,8 +606,8 @@ def assert_bad_request(payload, expected_error_message):
 def test_log_model(mlflow_client):
     experiment_id = mlflow_client.create_experiment("Log models")
     with TempDir(chdr=True):
-        mlflow.set_experiment("Log models")
         model_paths = ["model/path/{}".format(i) for i in range(3)]
+        mlflow.set_tracking_uri(mlflow_client.tracking_uri)
```
BenWilson2, line 610: Needs to be set so that the run context has reference to the created experiment in the patched mlflow_client obj

mlflow/utils/name_utils.py
```diff
@@ -1,4 +1,31 @@
...
+
+def _generate_unique_integer_id():
+    """
+    Utility function for generating a random fixed-length integer
+    :param id_length: The target length of the string representation of the integer without
+                      leading zeros
+    :return: a fixed-width integer
+    """
+
+    random_int = int.from_bytes(uuid.uuid4().bytes, byteorder=sys.byteorder, signed=True)
```
harupy, line 17: Can we use `uuid.uuid4().int` or is the current approach more random?

tests/store/tracking/test_file_store.py
```diff
@@ -157,6 +158,31 @@ def test_valid_root(self):
...
+    def test_removing_default_experiment_recreates_default_id(self):
+        def _is_default_in_experiments(view_type):
+            search_result = file_store.search_experiments(view_type=view_type)
+            ids = [experiment.experiment_id for experiment in search_result]
+            return FileStore.DEFAULT_EXPERIMENT_ID in ids
+
+        file_store = FileStore(self.test_root)
+        self.assertTrue(_is_default_in_experiments(ViewType.ACTIVE_ONLY))
+        # Soft-delete the default experiment
+        file_store.delete_experiment(FileStore.DEFAULT_EXPERIMENT_ID)
```
harupy, line 170: Didn't know the default experiment can be deleted.

mlflow/utils/name_utils.py
BenWilson2: I was doing some tests for signed vs. unsigned to see what the distribution of unique values might be. Based on what I'm seeing it really doesn't matter. Having a signed int doesn't give us any benefit on collision avoidance (at least for the first 100M - I'm not testing more than that due to it taking ~20 minutes to test on my laptop). I'm going to use the standard unsigned UUID().int to cast to the 128bit int and position extract the last n places.

tests/store/tracking/test_file_store.py
BenWilson2: I don't think it's recommended to. Do you think it's pointless to test this?

harupy: I think we should not allow users to delete the default experiment.

mlflow/utils/name_utils.py
```diff
@@ -1,4 +1,30 @@
...
+    """
+
+    random_int = uuid.uuid4().int
+    # Cast to string to get a fixed length
+    random_str = str(random_int)[-_EXPERIMENT_ID_FIXED_WIDTH:]
+    # append a random int as string to the end of the generated string for as many
+    # leading zeros exist in the generated string in order to preserve the total length
+    # once cast back to int
+    for i in range(len(random_str)):
+        if random_str[i] == "0":
```
harupy, line 23:
```suggestion
    for s in random_str:
        if s == "0":
```

mlflow/store/tracking/dbmodels/models.py
```diff
@@ -72,11 +72,11 @@ class SqlExperiment(Base):
     Lifecycle Stage of experiment: `String` (limit 32 characters).
                                     Can be either ``active`` (default) or ``deleted``.
     """
-    creation_time = Column(BigInteger(), default=int(time.time() * 1000))
+    creation_time = Column(BigInteger, default=int(time.time() * 1000))
```
harupy, line 75: Is `BigInteger()` equivalent to `BigInteger`?

BenWilson2: yep!

mlflow/store/tracking/dbmodels/models.py
```diff
@@ -72,11 +72,11 @@ class SqlExperiment(Base):
...
     """
     Creation time of experiment: `BigInteger`.
     """
-    last_update_time = Column(BigInteger(), default=int(time.time() * 1000))
+    last_update_time = Column(BigInteger, default=int(time.time() * 1000))
```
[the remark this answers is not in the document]
harupy: If we do `default=int(time.time() * 1000)`, that would set the default value to the time when this class is defined, which is incorrect.

mlflow/utils/name_utils.py
BenWilson2: good idea. changed!

harupy:
@BenWilson2 Can I push a commit to update the mssql schema?

BenWilson2:
> @BenWilson2 Should I push a commit to update the mssql schema?

I just pushed the relative path requirements.txt change. We're good to push the updated schema for mssql :) Thanks @harupy !

tests/store/tracking/test_file_store.py
BenWilson2: I made some adjustments to the logic to avoid the identity insert logic that was there before. Instead of regenerating the default experiment_id and using that for the current `create_experiment()` call, I'm issuing a warning that recreates the default experiment and then continuing generating a new experiment with a two-row commit within `create_experiment()` so that the tracking server doesn't end up in a non-functional state if someone dropped a row entry or deleted the local fs directory for experiment_id 0. This will be largely silent (with just a warning emitted) but no change to the expected behavior of create_experiment(). 
I added some validation steps to the unit test for this.

BenWilson2:
Due to the extensive amount of changes needed to make SqlalchemyStore function with this change, I'm limiting the scope of the initial PR to just FileStore: https://github.com/mlflow/mlflow/pull/7070. Closing this draft PR.

PR #6999, opened 2022-10-07
Speedup split step

prithvikannan:
## Related Issues/PRs

#xxx

Parallelize hashing in the split step. Implemented with native python multiprocessing and pool. This function is still deterministic (See: [Pool.map ordering](https://stackoverflow.com/questions/41273960/python-3-does-pool-keep-the-original-order-of-data-passed-to-map))

Profiling
* Baseline (no parallel): `93.78527522087097 second(s).`
* Parallel with 2 workers: `59.39586281776428 second(s).`
* Parallel with 4 workers `28.890960931777954 second(s).`
 

## How is this patch tested?

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/steps/split.py
+++ b/mlflow/pipelines/steps/split.py
@@ -76,7 +76,7 @@ def _create_hash_buckets(input_df):
     pandarallel_start_time = time.time()
     pandarallel.initialize(progress_bar=False)
     hash_start_time = time.time()
-    hash_buckets = _hash_pandas_dataframe(input_df).parallel_apply(
+    hash_buckets = _hash_pandas_dataframe(input_df).parallel_map(
         lambda x: (x % _SPLIT_HASH_BUCKET_NUM) / _SPLIT_HASH_BUCKET_NUM
     )
     end_time = time.time()
--- a/mlflow/pipelines/steps/split.py
+++ b/mlflow/pipelines/steps/split.py
@@ -71,15 +71,22 @@ def _create_hash_buckets(input_df):
     # Create hash bucket used for splitting dataset
     # Note: use `hash_pandas_object` instead of python builtin hash because it is stable
     # across different process runs / different python versions
-    start_time = time.time()
-    hash_buckets = _hash_pandas_dataframe(input_df).map(
+    from pandarallel import pandarallel
+
+    pandarallel_start_time = time.time()
+    pandarallel.initialize(progress_bar=False)
+    hash_start_time = time.time()
+    hash_buckets = _hash_pandas_dataframe(input_df).parallel_apply(
         lambda x: (x % _SPLIT_HASH_BUCKET_NUM) / _SPLIT_HASH_BUCKET_NUM
     )
-    execution_duration = time.time() - start_time
+    end_time = time.time()
+    execution_duration = end_time - hash_start_time
+    pandarallel_overhead = hash_start_time - pandarallel_start_time
     _logger.info(
         f"Creating hash buckets on input dataset containing {len(input_df)} "
         f"rows consumes {execution_duration} seconds."
     )
+    _logger.info(f"Pandarallel overhead: {pandarallel_overhead} second(s).")
     return hash_buckets
 
 
--- a/mlflow/pipelines/steps/split.py
+++ b/mlflow/pipelines/steps/split.py
@@ -76,17 +76,21 @@ def _create_hash_buckets(input_df):
     pandarallel_start_time = time.time()
     pandarallel.initialize(progress_bar=False)
     hash_start_time = time.time()
-    hash_buckets = _hash_pandas_dataframe(input_df).parallel_map(
+    hashed_df = _hash_pandas_dataframe(input_df)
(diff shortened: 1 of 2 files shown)


mlflow/pipelines/steps/split.py
```diff
@@ -63,8 +63,11 @@ def _get_split_df(input_df, hash_buckets, split_ratios):
 
 def _hash_pandas_dataframe(input_df):
     from pandas.util import hash_pandas_object
+    from pandarallel import pandarallel
 
-    return hash_pandas_object(input_df.applymap(_make_elem_hashable))
+    pandarallel.initialize(progress_bar=False)
```
jinzhang21, line 68: I'm not sure we want to introduce this package but don't have a strong feeling about it. We might move to Spark Dataframe at some point but it could take quite a while.

Do you mind do more thorough measurements on the performance gains by varying the data along these 2 dimensions?
- data size: 10 MB / 1GB
- num of columns: 10 / 1000
- machine type: local laptop / Databricks
- number of cores (can you control this?): 2 / 4 / 8 / 16

If we get good speedup (2x or more) on most testing combinations, it makes sense to me to use the package.
Thanks a lot, @prithvikannan !

dbczumar: I don't think we should introduce new dependencies unless it's necessary / really hard to do without the new dependency. Are there any alternatives? What about a multithreaded approach that iterates over different portions of the dataset?

dbczumar: In any case, we want MLP to run on Databricks by default in MLflow 2.0. To do that, we would need this dependency to be present in MLR. Is it present in MLR?

prithvikannan: > What about a multithreaded approach that iterates over different portions of the dataset?

I believe that is what pandarallel does under the hood. [This stackoverflow post](https://stackoverflow.com/questions/26784164/pandas-multiprocessing-apply#:~:text=A%20more%20generic%20version%20based%20on%20the%20author%20solution%2C%20that%20allows%20to%20run%20it%20on%20every%20function%20and%20dataframe%3A) outlines a way to do this with just python native multiprocessing.

sunishsheth2009 (approved):

mlflow/pipelines/steps/split.py
```diff
@@ -61,10 +61,34 @@ def _get_split_df(input_df, hash_buckets, split_ratios):
     return train_df, validation_df, test_df
 
 
+def _parallelize(data, func, num_of_processes=8):
```
jinzhang21, line 64: Can we introduce a constant at the top of the file? e.g.
`_MULTI_PROCESS_POOL_SIZE = 8` for readability and patchability purpose?

mlflow/pipelines/steps/split.py
```diff
@@ -61,10 +61,34 @@ def _get_split_df(input_df, hash_buckets, split_ratios):
...
+    import numpy as np
+    import pandas as pd
+    from multiprocessing import Pool
+
+    data_split = np.array_split(data, num_of_processes)
```
jinzhang21, line 69: Ideally the num_of_processes should be data dependent. e.g. I don't see a reason to split it if it's only 10 MB or smaller. The actual optimal value is probably a piece-wise linear function to data size. 

Not sure where to make the cut. It can probably be determined experimentally by varying the data size vs num_of_processes, if you'd like to give it a try:) Otherwise, apply some simply binary logic, if smaller than 10 MB, use 1 process, otherwise, use the max pool size.

mlflow/pipelines/steps/split.py
```diff
@@ -61,10 +61,34 @@ def _get_split_df(input_df, hash_buckets, split_ratios):
...
+    pool = Pool(num_of_processes)
+    data = pd.concat(pool.map(func, data_split))
+    pool.close()
+    pool.join()
+    return data
+
+
+def _run_on_subset(func, data_subset):
+    return data_subset.applymap(func)
```
jinzhang21, line 78: Can probably move this into `_parallelize_on_rows` as local function

jinzhang21 (approved):

mlflow/pipelines/steps/split.py
prithvikannan: @jinzhang21  sorry had automerge on :/ -- will create a followup PR for this

mlflow/pipelines/steps/split.py
prithvikannan: @jinzhang21 agreed, there's probably some heuristic we can follow for the optimal `n`.

my thought process with this design is for small datasets the total duration of the split step is already low, so introducing some additional latency will likely not be noticeable. but for large data, this will provide significant speedup.

mlflow/pipelines/steps/split.py
prithvikannan: This creates some issues with pickle -- see this [stackoverflow thread](https://stackoverflow.com/questions/52265120/python-multiprocessing-pool-attributeerror).

PR #7010, opened 2022-10-10
Adding support for UC tables in ingest step

sunishsheth2009:
Adding support for UC tables in ingest step

## How is this patch tested?
- [x] notebook

image

image

image

Adding support to ingest data using spark table

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/steps/ingest/datasets.py
+++ b/mlflow/pipelines/steps/ingest/datasets.py
@@ -583,7 +583,7 @@ class SparkSqlDataset(_SparkDatasetMixin, _Dataset):
     (e.g. `SELECT * FROM my_spark_table`).
     """
 
-    def __init__(self, sql: str, dataset_format: str):
+    def __init__(self, sql: str, location: str, dataset_format: str):
         """
         :param sql: The Spark SQL query string that defines the dataset
                     (e.g. 'SELECT * FROM my_spark_table').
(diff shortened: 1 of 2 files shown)


mlflow/pipelines/steps/ingest/datasets.py
```diff
@@ -583,25 +583,36 @@ class SparkSqlDataset(_SparkDatasetMixin, _Dataset):
...
+                "Either location or sql configuration key must be specified for "
+                "dataset with format spark_sql"
+            ) from None
         spark_session = self._get_or_create_spark_session()
-        spark_df = spark_session.sql(self.sql)
+        spark_df = None
+        if self.location is not None:
+            spark_df = spark_session.table(self.location)
+        if self.sql is not None:
+            spark_df = spark_session.sql(self.sql)
```
jinzhang21, line 607: What happens when when both `self.location` and `self.sql` are specified? It could result in subtle bugs when the location table is different from the SQL query table. 

In addition, we should use `if-else if-else` clause to avoid reading the table twice.

mlflow/pipelines/steps/ingest/datasets.py
```diff
@@ -583,25 +583,36 @@ class SparkSqlDataset(_SparkDatasetMixin, _Dataset):
...
-    def __init__(self, sql: str, dataset_format: str):
+    def __init__(self, sql: str, location: str, dataset_format: str):
         """
         :param sql: The Spark SQL query string that defines the dataset
                     (e.g. 'SELECT * FROM my_spark_table').
         :param dataset_format: The format of the dataset (e.g. 'csv', 'parquet', ...).
         """
         super().__init__(dataset_format=dataset_format)
         self.sql = sql
+        self.location = location
```
jinzhang21, line 594: `location` only supports `catalog.schema.table` or `schema.table` format, correct? Can we call it `table_path` or `table_location` to be more specific?

mlflow/pipelines/steps/ingest/datasets.py
```diff
@@ -583,25 +583,36 @@ class SparkSqlDataset(_SparkDatasetMixin, _Dataset):
     (e.g. `SELECT * FROM my_spark_table`).
     """
 
-    def __init__(self, sql: str, dataset_format: str):
+    def __init__(self, sql: str, location: str, dataset_format: str):
         """
         :param sql: The Spark SQL query string that defines the dataset
                     (e.g. 'SELECT * FROM my_spark_table').
```
jinzhang21, line 589: Add a comment for `location`.

mlflow/pipelines/steps/ingest/datasets.py
sunishsheth2009: Yes! i forgot to add if..else. Good catch.

I think if `sql` is provided, we will choose that over the `location`. Location would be used to read the entire table if the sql is not provided as a fallback. 
What do you think?

mlflow/pipelines/steps/ingest/datasets.py
sunishsheth2009: I think it could read any table. it doesn't need to be of the format mentioned above. Example in the test. We save the table as `spark_df.write.mode("overwrite").saveAsTable("test_table")` and using `spark_session.table(self.location)` we can read the table. So we can support any table. I don't think we need to be restrictive about it unless I am missing something. Thoughts?

I can switch it to `table_location` but I was just trying to be consistent with other variable name associated with `INGEST_DATA_LOCATION`.

jinzhang21 (approved):

PR #7030, opened 2022-10-12
Move subgraph logic into base pipeline

bbarnes52:

This PR updates the `_get_subgraph_for_target` step implementation to decouple it from the regression pipeline and moves it into the base pipeline such that it is no longer abstract. This logic will be reused by the classification pipeline. To accomplish this change, each step is also updated to declare whether it runs at training time or inference time. 

## How is this patch tested?

Ran MLP regression template locally

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/pipeline.py
+++ b/mlflow/pipelines/pipeline.py
@@ -3,7 +3,7 @@ import logging
 
 from mlflow.exceptions import MlflowException
 from mlflow.pipelines.artifacts import Artifact
-from mlflow.pipelines.step import BaseStep, StepStatus
+from mlflow.pipelines.step import BaseStep, StepStatus, StepClass
 from mlflow.pipelines.utils import (
     get_pipeline_config,
     get_pipeline_name,
@@ -153,15 +153,18 @@ class _BasePipeline:
         return self._steps[step_names.index(step_name)]
 
     @experimental
-    @abc.abstractmethod
     def _get_subgraph_for_target_step(self, target_step: BaseStep) -> List[BaseStep]:
         """
         Return a list of step objects representing a connected DAG containing the target_step.
         The returned list should be a sublist of self._steps.
-
-        Concrete pipeline class should implement this method.
         """
-        pass
+        subgraph = []
+        if target_step.step_class == StepClass.UNKNOWN:
+            return subgraph
+        for step in self._steps:
+            if target_step.step_class() == step.step_class():
+                subgraph.append(step)
+        return subgraph
 
     @experimental
     @abc.abstractmethod
--- a/mlflow/pipelines/step.py
+++ b/mlflow/pipelines/step.py
@@ -35,6 +35,19 @@ class StepStatus(Enum):
     FAILED = "FAILED"
 
 
+class StepClass(Enum):
+    """
+    Represents the class of a step.
+    """
+
+    # Indicates that the step class is unknown.
+    UNKNOWN = "UNKNOWN"
+    # Indicates that the step runs at training time.
+    TRAINING = "TRAINING"
+    # Indicates that the step runs at inference time.
+    PREDICTION = "PREDICTION"
+
+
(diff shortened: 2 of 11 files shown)


mlflow/pipelines/pipeline.py
```diff
@@ -153,15 +153,16 @@ def _get_step(self, step_name) -> BaseStep:
...
         Return a list of step objects representing a connected DAG containing the target_step.
         The returned list should be a sublist of self._steps.
-
-        Concrete pipeline class should implement this method.
         """
-        pass
+        subgraph = []
+        for step in self._steps:
+            if target_step.is_predict_step() == step.is_predict_step():
+                subgraph.append(step)
```
jinzhang21, line 164: This would mean that in the all future pipelines, we'll have 2 graphs at least: training graph and inference graph. It seems fine to me but can you add a comment to state this explicitly? It'd be helpful for future readers and new pipeline implementers.

mlflow/pipelines/step.py
```diff
@@ -233,6 +233,14 @@ def get_artifacts(self) -> List[Any]:
         """
         return {}
 
+    @experimental
+    @abc.abstractmethod
+    def is_predict_step(self) -> bool:
```
jinzhang21, line 238: I really don't like this approach. Why is `predict` so special? What if we want to add another subgraph, e.g. `deploy` with only one step `deploy`? Can we think about a more general approach to solve this issue, e.g. through some enum or predefined subgraphs?
A very rough thought, if we explicitly define subgraphs in the base class, e.g.
```
_TRAINING_STEPS = {"ingest", "split", "train", "evaluate", "register"}
_INFERENCE_STEPS = {"ingest_scoring", "predict"}
# (in the future) _DEPLOY_STEPS = {"deploy"}
```
then you may be able to tell which subgraph the step is in based on the predefined name sets.

mlflow/pipelines/pipeline.py
dbczumar: Not sure that making this assumption is worthwhile, as it'll likely break in the future. Is including the subgraph logic in each pipeline implementation too cumbersome?

mlflow/pipelines/step.py
bbarnes52: I contemplated using an enum. It seems like a premature optimization to me, but happy to do it! 

I don't think we should define subgraphs in the base class, that violates open-closed. Too easy to forget to update those constants when adding/removing steps.

mlflow/pipelines/pipeline.py
bbarnes52: I can update this to an enum, though I don't see any reason to include the subgraph logic in each pipeline implementation - not that it's terribly cumbersome, just unnecessarily not DRY. Do you see things differently?

"This would mean that in the all future pipelines, we'll have 2 graphs at least: training graph and inference graph." - if a pipeline consists only of training steps, it would have a single subgraph.

mlflow/pipelines/step.py
dbczumar: Agreed with both points here. I'd suggest we leave the structure as-is for now :)

bbarnes52: Say this is updated to:
```
def step_class(self) -> StepClass
```
The benefit of this approach is you don't have to duplicate code across pipelines implementations. Can you help me understand your opposition to this approach?

jinzhang21: `StepClass` reads much better. @dbczumar I have trouble understand your comment. Which "points" do you agree, and what does it mean by "leave the structure as-is"? What "structure" were you referring to?

dbczumar: I was saying that I prefer leaving master as-is. Wasn't thrilled with special casing prediction and also wasn't thrilled with the idea of defining the pipeline subgraphs in the base class.

StepClass sounds good to me!

bbarnes52: Great, I agree this is in a better state now, so thank you for the feedback!

jinzhang21 (approved):

PR #7047, opened 2022-10-13
Adding validation for ingest and split step

sunishsheth2009:
Adding validation for ingest and split step
Note: Others steps would be added in follow up PRs

## How is this patch tested?
- [x] Tested in notebook example

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/steps/ingest/datasets.py
+++ b/mlflow/pipelines/steps/ingest/datasets.py
@@ -436,6 +436,20 @@ class CustomDataset(_PandasConvertibleDataset):
             self.custom_loader_method_name,
         ) = custom_loader_method.rsplit(".", 1)
 
+    def _validate_user_code_output(self, func, *args):
+        import pandas as pd
+
+        ingested_df = func(*args)
+        if not isinstance(ingested_df, pd.DataFrame):
+            raise MlflowException(
+                message=(
+                    "The `ingested_data` is not a DataFrame, please make sure "
+                    f"'{self.custom_loader_method_name}' returns a Pandas DataFrame object."
+                ),
+                error_code=INVALID_PARAMETER_VALUE,
+            ) from None
+        return ingested_df
+
     def _load_file_as_pandas_dataframe(self, local_data_file_path: str):
         try:
             sys.path.append(self.pipeline_root)
@@ -454,7 +468,11 @@ class CustomDataset(_PandasConvertibleDataset):
             ) from e
 
         try:
-            return custom_loader_method(local_data_file_path, self.dataset_format)
+            return self._validate_user_code_output(
+                custom_loader_method, local_data_file_path, self.dataset_format
+            )
+        except MlflowException as e:
+            raise e
         except NotImplementedError:
             raise MlflowException(
                 message=(
--- a/mlflow/pipelines/steps/split.py
+++ b/mlflow/pipelines/steps/split.py
@@ -109,6 +109,45 @@ def _create_hash_buckets(input_df):
     return hash_buckets
 
 
+def _validate_user_code_output(post_split, train_df, validation_df, test_df):
+    try:
+        (
+            post_filter_train_df,
+            post_filter_validation_df,
+            post_filter_test_df,
+        ) = post_split(train_df, validation_df, test_df)
+    except Exception:
+        raise MlflowException(
+            message="Error in cleaning up the data frame post split step."
+            " Expected output is a tuple with (train_df, validation_df, test_df)"
+        ) from None
+
+    import pandas as pd
+
+    for (post_split_df, pre_split_df, split_type) in [
+        [post_filter_train_df, train_df, "train"],
+        [post_filter_validation_df, validation_df, "validation"],
+        [post_filter_test_df, test_df, "test"],
+    ]:
+        if not isinstance(post_split_df, pd.DataFrame):
+            raise MlflowException(
+                message="The split data is not a DataFrame, please return the correct data."
+            ) from None
+        if list(pre_split_df.columns) != list(post_split_df.columns):
+            raise MlflowException(
+                message="The number of columns post split step are different."
+                f" Column list for {split_type} dataset pre-slit is {list(pre_split_df.columns)}"
+                f" and post split is {list(post_split_df.columns)}. "
+                "Split filter function should be used to filter rows rather than filtering columns."
+            ) from None
+
+    return (
+        post_filter_train_df,
+        post_filter_validation_df,
+        post_filter_test_df,
+    )
+
+
 class SplitStep(BaseStep):
     def _validate_and_apply_step_config(self):
         self.run_end_time = None
@@ -228,7 +267,12 @@ class SplitStep(BaseStep):
                 importlib.import_module(post_split_module_name), post_split_fn_name
             )
             _logger.debug(f"Running {post_split_fn_name} on train, validation and test datasets.")
-            (train_df, validation_df, test_df) = post_split(train_df, validation_df, test_df)
+            (
+                train_df,
+                validation_df,
+                test_df,
+            ) = _validate_user_code_output(post_split, train_df, validation_df, test_df)
+
         elif post_split_filter_config is not None:
             (
                 post_split_filter_module_name,
--- a/tests/pipelines/test_ingest_step.py
+++ b/tests/pipelines/test_ingest_step.py
@@ -242,6 +242,30 @@ def test_ingest_throws_for_custom_dataset_when_custom_loader_function_not_implem
         ).run(output_directory=tmp_path)
 
 
+def custom_load_file_as_array(local_data_file_path, dataset_format):
+    return [local_data_file_path, dataset_format]
+
+
+@pytest.mark.usefixtures("enter_test_pipeline_directory")
+def test_ingest_throws_for_custom_dataset_when_custom_method_returns_array(pandas_df, tmp_path):
+    dataset_path = tmp_path / "df.fooformat"
+    pandas_df.to_csv(dataset_path, sep="#")
+
+    with pytest.raises(MlflowException, match="The `ingested_data` is not a DataFrame"):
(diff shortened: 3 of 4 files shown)


sunishsheth2009 (commented):
Would love to get feedback on fixing the error messages.

mlflow/pipelines/steps/split.py
```diff
@@ -238,6 +244,19 @@ def _run(self, output_directory):
...
+
+        if not isinstance(post_split_df, pd.DataFrame):
+            raise MlflowException(
+                message="The split data is not a DataFrame, please return the correct data"
+            )
+        if len(pre_split_df.columns) != len(post_split_df.columns):
+            raise MlflowException(
+                message="The number of columns post split step are different."
+                "Split filter step is used to filter rows rather than filtering out columns."
+            )
```
prithvikannan, line 258: I don't think this case is possible given the constraint on the user defined filtering fn.

mlflow/pipelines/steps/split.py
```diff
@@ -238,6 +244,19 @@ def _run(self, output_directory):
         self.execution_duration = self.run_end_time - run_start_time
         return self._build_profiles_and_card(train_df, validation_df, test_df)
 
+    def _validate_user_code_output(self, pre_split_df, post_split_df):
+        import pandas as pd
+
+        if not isinstance(post_split_df, pd.DataFrame):
+            raise MlflowException(
+                message="The split data is not a DataFrame, please return the correct data"
+            )
```
prithvikannan, line 253: Same thing here - I think this will always be a df

prithvikannan (commented):
Thanks for working on this @sunishsheth2009. In the case of the split logic, I am not sure that either of those cases that you check for are possible given the constraints on the user defined split function. Do you have an example that hits these?

mlflow/pipelines/steps/split.py
sunishsheth2009: Yeah! I added this and the other function to do this validation check. So I think we can still validate the new function as well. What do you think?

mlflow/pipelines/steps/split.py
sunishsheth2009: Same as below

mlflow/pipelines/steps/ingest/__init__.py
```diff
@@ -100,6 +101,12 @@ def _run(self, output_directory: str) -> BaseCard:
         )
         return step_card
 
+    def _validate_user_code_output(self, ingested_df):
+        if not isinstance(ingested_df, pd.DataFrame):
+            raise MlflowException(
+                message="The `ingested_data` is not a DataFrame, please ingest the correct data format"
```
jinzhang21, line 107: This message is not helpful enough. As a user I'd have a hard time to figure out what went wrong. It should be directly connected to the user defined function name and its required output.

mlflow/pipelines/steps/ingest/__init__.py
```diff
@@ -76,6 +76,7 @@ def _run(self, output_directory: str) -> BaseCard:
         _logger.debug("Successfully stored data in parquet format at '%s'", dataset_dst_path)
 
         ingested_df = read_parquet_as_pandas_df(data_parquet_path=dataset_dst_path)
+        self._validate_user_code_output(ingested_df)
```
jinzhang21, line 79: Instead of validating here, can we validate at the point of `load_file_as_dataframe` being called? It'd be much easier to debug in terms of stack traces and error message.

mlflow/pipelines/steps/split.py
```diff
@@ -213,7 +213,19 @@ def _run(self, output_directory):
                 importlib.import_module(post_split_module_name), post_split_fn_name
             )
             _logger.debug(f"Running {post_split_fn_name} on train, validation and test datasets.")
-            (train_df, validation_df, test_df) = post_split(train_df, validation_df, test_df)
+            (post_filter_train_df, post_filter_validation_df, post_filter_test_df) = post_split(
+                train_df, validation_df, test_df
+            )
+            self._validate_user_code_output(train_df, post_filter_train_df)
+            self._validate_user_code_output(validation_df, post_filter_validation_df)
+            self._validate_user_code_output(test_df, post_filter_test_df)
```
jinzhang21, line 221: Can we move these into a single function, something like `_validate_post_split_output`?

mlflow/pipelines/steps/split.py
```diff
@@ -226,9 +238,15 @@ def _run(self, output_directory):
             _logger.debug(
                 f"Running {post_split_filter_fn_name} on train, validation and test datasets."
             )
-            train_df = train_df[post_split_filter(train_df)]
```
jinzhang21, line 229: This is a great place to use decorators. You can keep the code intact but still perform output validation with a decorated `post_split_filter` method.

mlflow/pipelines/steps/split.py
```diff
@@ -253,6 +258,36 @@ def _run(self, output_directory):
         self.execution_duration = self.run_end_time - run_start_time
         return self._build_profiles_and_card(train_df, validation_df, test_df)
 
+    def _validate_user_code_output(post_split, train_df, validation_df, test_df):
+        (
+            post_filter_train_df,
+            post_filter_validation_df,
+            post_filter_test_df,
```
jinzhang21, line 265: Should try-catch here too since user may return something not decomposable to a 3-tuple.

mlflow/pipelines/steps/split.py
```diff
@@ -253,6 +258,36 @@ def _run(self, output_directory):
...
+            [post_filter_test_df, test_df],
+        ]:
+            if not isinstance(post_split_df, pd.DataFrame):
+                raise MlflowException(
+                    message="The split data is not a DataFrame, please return the correct data"
+                )
+            if len(pre_split_df.columns) != len(post_split_df.columns):
+                raise MlflowException(
+                    message="The number of columns post split step are different."
+                    "Split filter step is used to filter rows rather than filtering out columns."
```
jinzhang21, line 282: Can we print out the column diff here so it's easier to debug?

mlflow/pipelines/steps/ingest/datasets.py
```diff
@@ -436,6 +436,14 @@ def __init__(
             self.custom_loader_method_name,
         ) = custom_loader_method.rsplit(".", 1)
 
+    def _validate_user_code_output(self, func, *args):
+        ingested_df = func(*args)
+        if not isinstance(ingested_df, pd.DataFrame):
+            raise MlflowException(
+                message="The `ingested_data` is not a DataFrame, please ingest the correct data format"
```
jinzhang21, line 443: "...please ingest the correct data format" -> "please make sure `self.custom_loader_method_name` returns a Pandas DataFrame object".

jinzhang21 (commented):
Can we add a unit test to validate the validation works as expected?

prithvikannan (approved):
Thanks @sunishsheth2009, this looks good to me. Let's merge this as-is, and then we can remove the validation on split step once we remove `post_split_filter_config`

mlflow/pipelines/steps/split.py
```diff
@@ -109,6 +109,45 @@ def _create_hash_buckets(input_df):
...
+    try:
+        (
+            post_filter_train_df,
+            post_filter_validation_df,
+            post_filter_test_df,
+        ) = post_split(train_df, validation_df, test_df)
+    except Exception:
+        raise MlflowException(
+            message="Error in cleaning up the data frame post split step."
+            " Expected output is a list with (train_df, validation_df, test_df)"
```
jinzhang21, line 122: `list` -> `tuple`

mlflow/pipelines/steps/split.py
```diff
@@ -109,6 +109,45 @@ def _create_hash_buckets(input_df):
...
+        [post_filter_test_df, test_df, "test"],
+    ]:
+        if not isinstance(post_split_df, pd.DataFrame):
+            raise MlflowException(
+                message="The split data is not a DataFrame, please return the correct data."
+            ) from None
+        if len(pre_split_df.columns) != len(post_split_df.columns):
+            raise MlflowException(
+                message="The number of columns post split step are different."
+                f" Column length for {split_type} dataset pre slit is {len(pre_split_df.columns)}"
```
jinzhang21, line 139: `pre slit` -> `pre-split`

mlflow/pipelines/steps/split.py
```diff
@@ -109,6 +109,45 @@ def _create_hash_buckets(input_df):
...
+    for (post_split_df, pre_split_df, split_type) in [
+        [post_filter_train_df, train_df, "train"],
+        [post_filter_validation_df, validation_df, "validation"],
+        [post_filter_test_df, test_df, "test"],
+    ]:
+        if not isinstance(post_split_df, pd.DataFrame):
+            raise MlflowException(
+                message="The split data is not a DataFrame, please return the correct data."
+            ) from None
+        if len(pre_split_df.columns) != len(post_split_df.columns):
```
jinzhang21, line 136: Instead of matching the size of columns, can we match on the column name instead?
`if list(pre_split_df.columns) != list(post_split_df.columns):`

mlflow/pipelines/steps/split.py
jinzhang21, line 139: Update the error message if compare columns directly.

mlflow/pipelines/steps/split.py
```diff
@@ -109,6 +109,45 @@ def _create_hash_buckets(input_df):
...
+            raise MlflowException(
+                message="The number of columns post split step are different."
+                f" Column length for {split_type} dataset pre slit is {len(pre_split_df.columns)}"
+                f" and post split is {len(post_split_df.columns)}."
+                " Split filter step is used to filter rows rather than filtering out columns."
```
jinzhang21, line 141: `Split filter step` -> `Split filter function`
`is used to` -> `should be used to`

PR #7067, opened 2022-10-16
Throw exception if not all required step files are present

mingyu89:
Signed-off-by: Mingyu Li <mingyu.li@databricks.com>

As title. Please not that the exception will be thrown from the first step command, which is ingest.
This should be fine because it's checking all files required by the pipeline. User will have to add the required file sooner or later.

## Test
Tested by unit test test_create_required_step_files

## Related Issues/PRs

#xxx

## How is this patch tested?

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/utils/execution.py
+++ b/mlflow/pipelines/utils/execution.py
@@ -412,7 +412,10 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) -> 
         ]:
             required_file_path = os.path.join(steps_folder_path, required_file)
             if not os.path.exists(required_file_path):
-                with open(required_file_path, "w"):
+                try:
+                    with open(required_file_path, "w"):
+                        pass
+                except OSError:
                     pass
             if not os.path.exists(required_file_path):
                 raise ValueError(
--- a/mlflow/pipelines/utils/execution.py
+++ b/mlflow/pipelines/utils/execution.py
@@ -400,6 +400,9 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) -> 
 
     if template == "regression/v1":
         makefile_to_use = _REGRESSION_MAKEFILE_FORMAT_STRING
+        steps_folder_path = os.path.join(pipeline_root_path, "steps")
+        if not os.path.exists(steps_folder_path):
+            os.mkdir(steps_folder_path)
         for required_file in [
             "ingest.py",
             "split.py",
@@ -407,7 +410,10 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) -> 
             "transform.py",
             "custom_metrics.py",
         ]:
-            required_file_path = os.path.join(pipeline_root_path, "steps", required_file)
+            required_file_path = os.path.join(steps_folder_path, required_file)
+            if not os.path.exists(required_file_path):
+                with open(required_file_path, "w") as _:
+                    pass
             if not os.path.exists(required_file_path):
                 raise ValueError(
                     f"Can not find required file {required_file_path} from steps folder. "
--- a/mlflow/pipelines/utils/execution.py
+++ b/mlflow/pipelines/utils/execution.py
@@ -400,6 +400,19 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) -> 
 
     if template == "regression/v1":
         makefile_to_use = _REGRESSION_MAKEFILE_FORMAT_STRING
+        for required_file in [
+            "ingest.py",
+            "split.py",
+            "train.py",
+            "transform.py",
+            "custom_metrics.py",
+        ]:
+            required_file_path = os.path.join(pipeline_root_path, "steps", required_file)
+            if not os.path.exists(required_file_path):
+                raise ValueError(
+                    f"Can not find required file {required_file_path} from steps folder. "
+                    "Please create empty python file if the step is not used."
+                )
     else:
         raise ValueError(f"Invalid template: {template}")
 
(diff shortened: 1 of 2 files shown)


jinzhang21 (approved):

jinzhang21 (approved):
Please fix the unit test failures before merging, and add a new unit test verifying the behavior change.

mlflow/pipelines/utils/execution.py
```diff
@@ -400,6 +400,19 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) ->
 
     if template == "regression/v1":
         makefile_to_use = _REGRESSION_MAKEFILE_FORMAT_STRING
+        for required_file in [
+            "ingest.py",
+            "split.py",
+            "train.py",
+            "transform.py",
+            "custom_metrics.py",
+        ]:
```
sunishsheth2009, line 409: I am not sure if all of these files are required if the pipeline config doesn't have them mentioned in the yaml config. These could be optional. But if they are specified, I agree we should have them. So can check for that here? 
Let me know if I am missing something. 
cc @mengxr

dbczumar (approved):
LGTM! Thanks @mingyu89 !

mlflow/pipelines/utils/execution.py
bbarnes52: They are required by the current implementation even if unspecified, though I agree it would be desirable for them only be required if specified. Perhaps we could make that change separately, at any rate I think this PR is an improvement on what exists today. Nice work @mingyu89 !

bbarnes52 (approved):

mingyu89:
@sunishsheth2009  @bbarnes52 @jinzhang21 @dbczumar @prithvikannan Updated the PR to create the files for user instead of directly throwing exception. Added unit test.

mlflow/pipelines/utils/execution.py
```diff
@@ -400,6 +400,25 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) ->
...
+            required_file_path = os.path.join(steps_folder_path, required_file)
+            if not os.path.exists(required_file_path):
+                with open(required_file_path, "w") as _:
```
harupy, line 415:
```suggestion
                with open(required_file_path, "w"):
```

mlflow/pipelines/utils/execution.py
```diff
@@ -400,6 +400,25 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) ->
...
+                    pass
+            if not os.path.exists(required_file_path):
+                raise ValueError(
+                    f"Can not find required file {required_file_path} from steps folder. "
+                    "Please create empty python file if the step is not used."
+                )
```
harupy, line 421: Do we need this block? Looks like `not os.path.exists(required_file_path)` always returns False as we create an empty file at `required_file_path` in line 415.

tests/pipelines/test_execution_utils.py
```diff
@@ -104,6 +104,34 @@ def clean_test_pipeline(enter_test_pipeline_directory):  # pylint: disable=unuse
...
+    def check_required_files(check_exist):
+        for required_file in [
+            "steps",
+            "steps/ingest.py",
+            "steps/split.py",
+            "steps/train.py",
+            "steps/transform.py",
+            "steps/custom_metrics.py",
+        ]:
+            assert not check_exist ^ (tmp_path / required_file).exists()
```
harupy, line 125:
```suggestion
            assert (tmp_path / required_file).exists() is check_exist
```

This might be easier to understand.

harupy:
@mingyu89 Left a few minor comments, not a blocker.

mlflow/pipelines/utils/execution.py
mingyu89: It's probably not required. Any chance the `open(required_file_path, "w")` fails to create the file?

bbarnes52: There can be an IO error, consider:
```
try:
    open(required_file_path, 'w').close()
except OSError e:
    # Do something with e
```
If you want to add context to the error.

mlflow/pipelines/utils/execution.py
```diff
@@ -400,6 +400,25 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) ->
```
bbarnes52, line 400: Apologies for not asking earlier, but is `_create_makefile` definitely the right place to implement this logic? And more broadly, `_get_or_create_execution_dir`? This is creating files within the source repository and not the execution directory, correct?

mingyu89: `_create_makefile ` will run before running a step. Same for `_get_or_create_execution_directory`.
I put it in `_create_makefile ` because there's already a template check inside the method and it only makes sense to check step files for `regression/v1`. The step files are required by the specific make file - that another reason we are creating them together.

Maybe rename the method to "_create_makefile_and_steps"?

A better approach will be create a layer for project integrity check during `new Pipeline(x)`.

mlflow/pipelines/utils/execution.py
```diff
@@ -400,6 +400,28 @@ def _create_makefile(pipeline_root_path, execution_directory_path, template) ->
...
+            "split.py",
+            "train.py",
+            "transform.py",
+            "custom_metrics.py",
+        ]:
+            required_file_path = os.path.join(steps_folder_path, required_file)
+            if not os.path.exists(required_file_path):
+                try:
+                    with open(required_file_path, "w"):
+                        pass
```
harupy, line 417:
```suggestion
                    with open(required_file_path, "w") as f:
                        f.write("# Created by MLflow Pipeliens\n")
```

How about adding a comment in the file to clarify what created this file?

PR #7082, opened 2022-10-17
Implement classification pipeline

bbarnes52:

This PR introduces a pipeline implementation for building classification models and updates associated utility methods that perform logic conditioned on the template type. Note that the classification pipeline implementation is currently identical to the regression pipeline implementation, this is expected to change as we add additional facilities such as a rebalancing step. A few other notes:
- This PR also adds support for declaring the `positive_class` and threads this value through the mlflow evaluation service.
- This PR updates builtin metrics name which are inconsistent with the metrics supported by the mlflow evaluation service. I _think_ the [docs](https://www.mlflow.org/docs/latest/python_api/mlflow.html) may need to be updated.
- The worst_examples function doesn't work that well, the current error function returns True if the prediction is correct and false otherwise (rather than using the probability of the prediction to compute the error). I think this is OK for the first version in the interest of getting customer feedback ASAP, but please challenge me on this!

Importantly, I have not disabled this behind a feature flag, users will be able to build classification pipelines once this is released. If reviewers would prefer to disable this until we have a template + example repo, I'd be happy to do so.

## How is this patch tested?

Locally + Databricks using the Iris dataset.

Components

Interface

Language

Integrations

- [] `rn/none` - No description will be included. The PR will be mentioned only by the PR number in the "Small Bugfixes and Documentation Updates" section

### DIFF
--- a/docs/source/python_api/mlflow.pipelines.rst
+++ b/docs/source/python_api/mlflow.pipelines.rst
@@ -18,3 +18,12 @@ Regression Pipeline
 .. autoclass:: mlflow.pipelines.regression.v1.pipeline.RegressionPipeline()
     :members:
     :undoc-members:
+
+Classification Pipeline
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. automodule:: mlflow.pipelines.classification.v1.pipeline
+
+.. autoclass:: mlflow.pipelines.classification.v1.pipeline.ClassificationPipeline()
+    :members:
+    :undoc-members:
--- a/mlflow/pipelines/classification/v1/pipeline.py
+++ b/mlflow/pipelines/classification/v1/pipeline.py
@@ -1,0 +1,395 @@
+"""
+.. _mlflow-classification-pipeline:
+
+The MLflow Classification Pipeline is an MLflow Pipeline for developing binary classification
+models. Multiclass classifiers are currently not supported.
+The classification pipeline is designed for developing models using scikit-learn and
+frameworks that integrate with scikit-learn, such as the ``XGBClassifier`` API from XGBoost.
+The :py:class:`ClassificationPipeline API Documentation <ClassificationPipeline>` provides
+instructions for executing the pipeline and inspecting its results.
+
+The training pipeline contains the following sequential steps:
+
+**ingest** -> **split** -> **transform** -> **train** -> **evaluate** -> **register**
+
+The batch scoring pipeline contains the following sequential steps:
+
+**ingest_scoring** -> **predict**
+
+The pipeline steps are defined as follows:
+
+   - **ingest**
+      - The **ingest** step resolves the dataset specified by the |'data' section in pipeline.yaml|
+        and converts it to parquet format, leveraging the custom dataset parsing code defined in
+        |steps/ingest.py| if necessary. Subsequent steps convert this dataset into training,
+        validation, & test sets and use them to develop a model.
+
+        .. note::
+            If you make changes to the dataset referenced by the **ingest** step (e.g. by adding
+            new records or columns), you must manually re-run the **ingest** step in order to
+            use the updated dataset in the pipeline. The **ingest** step does *not* automatically
+            detect changes in the dataset.
+
+        .. note::
+            `target_col` must have a cardinality of two and `positive_class` must be specified.
+
+   .. _mlflow-classification-pipeline-split-step:
+
+   - **split**
+      - The **split** step splits the ingested dataset produced by the **ingest** step into
+        a training dataset for model training, a validation dataset for model performance
+        evaluation  & tuning, and a test dataset for model performance evaluation. The fraction
+        of records allocated to each dataset is defined by the ``split_ratios`` attribute of the
+        |'split' step definition in pipeline.yaml|. The **split** step also preprocesses the
+        datasets using logic defined in |steps/split.py|. Subsequent steps use these datasets
+        to develop a model and measure its performance.
+
+   - **transform**
+      - The **transform** step uses the training dataset created by **split** to fit
+        a transformer that performs the transformations defined in |steps/transform.py|. The
+        transformer is then applied to the training dataset and the validation dataset, creating
+        transformed datasets that are used by subsequent steps for estimator training and model
+        performance evaluation.
+
+   .. _mlflow-classification-pipeline-train-step:
+
+   - **train**
+      - The **train** step uses the transformed training dataset output from the **transform**
+        step to fit an estimator with the type and parameters defined in |steps/train.py|. The
+        estimator is then joined with the fitted transformer output from the **transform** step
+        to create a model pipeline. Finally, this model pipeline is evaluated against the
+        transformed training and validation datasets to compute performance metrics; custom
+        metrics are computed according to definitions in |steps/custom_metrics.py| and the
+        |'metrics' section of pipeline.yaml|. The model pipeline and its associated parameters,
+        performance metrics, and lineage information are logged to MLflow Tracking, producing
+        an MLflow Run.
+
+            .. note::
+                The **train** step supports hyperparameter tuning with hyperopt by adding 
+                configurations in the 
+                |'tuning' section of the 'train' step definition in pipeline.yaml|. 
+
+   - **evaluate**
+      - The **evaluate** step evaluates the model pipeline created by the **train** step on
+        the test dataset output from the **split** step, computing performance metrics and
+        model explanations. Performance metrics are compared against configured thresholds to
+        compute a ``model_validation_status``, which indicates whether or not a model is good
+        enough to be registered to the MLflow Model Registry by the subsequent **register**
+        step. Custom performance metrics are computed according to definitions in
+        |steps/custom_metrics.py| and the |'metrics' section of pipeline.yaml|. Model
+        performance thresholds are defined in the
+        |'validation_criteria' section of the 'evaluate' step definition in pipeline.yaml|. Model
+        performance metrics and explanations are logged to the same MLflow Tracking Run used by
+        the **train** step.
+
+   - **register**
+      - The **register** step checks the ``model_validation_status`` output of the preceding
+        **evaluate** step and, if model validation was successful
+        (as indicated by the ``'VALIDATED'`` status), registers the model pipeline created by
+        the **train** step to the MLflow Model Registry. If the ``model_validation_status`` does
+        not indicate that the model passed validation checks (i.e. its value is ``'REJECTED'``),
+        the model pipeline is not registered to the MLflow Model Registry.
+        If the model pipeline is registered to the MLflow Model Registry, a
+        ``registered_model_version`` is produced containing the model name and the model version.
+
+            .. note::
+                The model validation status check can be disabled by specifying
+                ``allow_non_validated_model: true`` in the
+                |'register' step definition of pipeline.yaml|, in which case the model pipeline is
+                always registered with the MLflow Model Registry when the **register** step is
+                executed.
+
+   - **ingest_scoring**
+      - The **ingest_scoring** step resolves the dataset specified by the 
+        |'data/scoring' section in pipeline.yaml| and converts it to parquet format, leveraging 
+        the custom dataset parsing code defined in |steps/ingest.py| if necessary. 
+    
+            .. note::
+                If you make changes to the dataset referenced by the **ingest_scoring** step 
+                (e.g. by adding new records or columns), you must manually re-run the 
+                **ingest_scoring** step in order to use the updated dataset in the pipeline. 
+                The **ingest_scoring** step does *not* automatically detect changes in the dataset.
+    
+   - **predict**
+      - The **predict** step uses the ingested dataset for scoring created by the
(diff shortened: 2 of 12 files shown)


mlflow/pipelines/utils/metrics.py
```diff
@@ -41,13 +41,10 @@ def from_custom_metric_dict(cls, custom_metric_dict):
...
     PipelineMetric(name="true_positives", greater_is_better=True),
-    PipelineMetric(name="recall", greater_is_better=True),
-    PipelineMetric(name="precision", greater_is_better=True),
+    PipelineMetric(name="recall_score", greater_is_better=True),
+    PipelineMetric(name="precision_score", greater_is_better=True),
     PipelineMetric(name="f1_score", greater_is_better=True),
     PipelineMetric(name="accuracy_score", greater_is_better=True),
-    PipelineMetric(name="log_loss", greater_is_better=False),
-    PipelineMetric(name="roc_auc", greater_is_better=True),
-    PipelineMetric(name="precision_recall_auc", greater_is_better=True),
```
prithvikannan, line 50: why do we need to remove this? it doesn't look like we are adding these back either. i'm not following.

prithvikannan (commented):
This looks awesome @bbarnes52! Very neat solution. I've left a small comment about the metrics changes.

prithvikannan:
the PR description says 
> This PR also adds support for declaring the positive_class in the eval step and threads this value through the mlflow evaluation service.

but from what i can tell from reading the code and testing, `positive_class` needs to be declared at the top level. i think the way it currently is makes the most sense, since the positive_class is critical to defining the problem. can you update the PR description to match @bbarnes52?

mlflow/pipelines/utils/metrics.py
bbarnes52: That's a great question, and one that we should probably follow up on.

When I attempted to use `log_loss` as my primary metric, the train step failed and returned an error that included the list of allowed metric names. The list did not include the three metrics removed here.

mlflow/pipelines/steps/evaluate.py
```diff
@@ -60,6 +60,12 @@ def _validate_and_apply_step_config(self):
                 "Missing template_name config in pipeline config.",
                 error_code=INVALID_PARAMETER_VALUE,
             )
+        self.positive_class = self.step_config.get("positive_class")
```
jinzhang21, line 63: Shall we verify that positive_class actually exists (i.e. a valid label) in the dataset[target_col] in the ingest step?

jinzhang21 (approved):
LGTM. Do we have a task to follow up with creating an E2E test running through the classification pipeline example repo, once we have it release?

mlflow/pipelines/steps/evaluate.py
bbarnes52: Good question, I think it depends. As things stand, the `positive_class` does not need to exist in the raw data, it could be created during the transform step. Do we want to require that `positive_class` exists in the raw data? I don't immediately see a reason to do so, but please challenge me on this!

We could also alternatively verify that `positive_class` exists after the transform step.

Let me know what you think!

jinzhang21: Let's don't add anything yet given we don't have any user feedback. We can always revisit this decision later on.

dbczumar: As I understand it, the transform step is only used for feature transformations. Transformations to target columns should be performed upstream (e.g. in ETL). The regression pipeline doesn't pass targets to the transform step.

mlflow/pipelines/utils/metrics.py
dbczumar: Were there other builtin metric names included in the list that aren't captured here? Where was the list of builtins printed from?

mlflow/pipelines/steps/train.py
```diff
@@ -135,6 +135,12 @@ def _validate_and_apply_step_config(self):
                 "Missing template_name config in pipeline config.",
                 error_code=INVALID_PARAMETER_VALUE,
             )
+        self.positive_class = self.step_config.get("positive_class")
+        if not self.positive_class and self.template == "classification/v1":
```
dbczumar, line 139: What if `self.positive_class` is `0`, `False`, `None`, or `""`? Perhaps we should check if `positive_class in not in self.step_config` instead.

dbczumar:
> The worst_examples function doesn't work that well, the current error function returns True if the prediction is correct and false otherwise (rather than using the probability of the prediction to compute the error). I think this is OK for the first version in the interest of getting customer feedback ASAP, but please challenge me on this!

Seems okay to me for starters :)

dbczumar:
Re: docs updates, once we have a template repo, we should add it / link it here: https://output.circle-artifacts.com/output/job/eb41f865-1ada-416f-af1d-d7c39639852a/artifacts/0/docs/build/html/pipelines.html#pipeline-templates

docs/source/python_api/mlflow.pipelines.rst
```diff
@@ -18,3 +18,12 @@ Regression Pipeline
 .. autoclass:: mlflow.pipelines.regression.v1.pipeline.RegressionPipeline()
     :members:
     :undoc-members:
+
+Classification Pipeline
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. automodule:: mlflow.pipelines.classification.v1.pipeline
+
+.. autoclass:: mlflow.pipelines.classification.v1.pipeline.ClassificationPipeline()
```
dbczumar, line 27: Classification pipeline class members don't seem to render in the generated docs (https://output.circle-artifacts.com/output/job/eb41f865-1ada-416f-af1d-d7c39639852a/artifacts/0/docs/build/html/python_api/mlflow.pipelines.html#mlflow.pipelines.classification.v1.pipeline.ClassificationPipeline)

Screen Shot 2022-10-21 at 12 01 43 PM

mlflow/pipelines/classification/v1/pipeline.py
```diff
@@ -0,0 +1,188 @@
+"""
+.. _mlflow-classification-pipeline:
+
+The MLflow Classification Pipeline is an MLflow Pipeline for developing classification models.
```
dbczumar, line 4: Do our docs mention that only binary classification is supported initially? 

Also, what's the behavior if a user passes in a dataset with > 2 classes? Do we throw or build a one-vs-rest classifier? We should probably make sure that we do one of these.

dbczumar (commented):
In the design for classification, I see that the following remain listed as goals:

- Split should do stratified sampling.
- Users may specify whether the trained model should output predicted labels or probabilities.
- A declarative API for handling imbalanced data. Class weights should be applied automatically for models which output probabilities. Validation and test metrics should be based on the original distribution. 
- Train step cards should show validation confusion matrices, ROC curve and precision-recall curves.


Are we incorporating all of these?

dbczumar:
Ideally, I'd prefer if we introduced the classification template repo (a less visible "test" location is fine for now), introduced test coverage, and then merged this PR once the coverage is in place. If we need to get this PR merged into `master` right now, I won't strongly object.

mlflow/pipelines/utils/metrics.py
bbarnes52: This is the error thrown by the evaluate step if `log_loss` is included in this list and specified in `pipeline.yaml`:
```
mlflow.exceptions.MlflowException: The metric log_loss is defined in the pipeline's validation criteria but was not returned from mlflow evaluation.
```

Here's a warning from the train step:
```
2022/10/21 12:25:41 WARNING mlflow.pipelines.utils.step: Input metric names {'log_loss'} not found in eval metrics: {'score', 'precision_score', 'false_negatives', 'true_positives', 'accuracy_score', 'recall_score', 'f1_score', 'example_count', 'true_negatives', 'false_positives'}
```

`example_count` and `score` are not captured here, let me know if you'd like to see them added. And also if you'd like me to create a bug report here (unless I've missed something obvious, quite possible!)

mlflow/pipelines/steps/evaluate.py
bbarnes52: `The regression pipeline doesn't pass targets to the transform step, for example.`
-> Right you are, my apologies. Would you like to see this validation performed during ingest (or split?)

docs/source/python_api/mlflow.pipelines.rst
bbarnes52: Ah nice catch, I didn't override methods that just deferred to their super class. Will update.

mlflow/pipelines/classification/v1/pipeline.py
bbarnes52: Good question. We would built a one-vs-rest classifier. What do you think desired behavior should be here?

bbarnes52:
> In the design for classification, I see that the following remain listed as goals:
> 
> * Split should do stratified sampling.
> * Users may specify whether the trained model should output predicted labels or probabilities.
> * A declarative API for handling imbalanced data. Class weights should be applied automatically for models which output probabilities. Validation and test metrics should be based on the original distribution.
> * Train step cards should show validation confusion matrices, ROC curve and precision-recall curves.
> 
> Are we incorporating all of these?

The current plan is to incorporate these after this PR is landed, I don't feel strongly though. @prithvikannan has a PR for the visualizations, the rest would be handled in the next sprint.

> Ideally, I'd prefer if we introduced the classification template repo (a less visible "test" location is fine for now), introduced > test coverage, and then merged this PR once the coverage is in place. If we need to get this PR merged into master sooner > (after comments are addressed), I won't strongly object.

I don't have a strong preference myself. The classification template repo will be nearly identical to the regression template repo, it would not take long to produce (the example repo will diverge a bit).

@jinzhang21 would you like to weigh in on these points?

mlflow/pipelines/steps/train.py
bbarnes52: Good call, will update.

jinzhang21:
> 
Since all existing tests are passing. I don't object to merging the current PR as is.

On classification template, agreed it shouldn't be too different from the regression one (in fact there are customers adapting the regression template to classification problems already).

However on classification example, I think we should have a small doc to discuss it before implementation. For example,
on the ML problem selection, we should consider
- Ease of understanding the nature of the problem
- Data source (need to vet its license e.g.)
- Extensibility (to future multi-class / imbalanced class problems).
- Should incorporate batch inference

mlflow/pipelines/classification/v1/pipeline.py
dbczumar: One-vs-rest sounds great! Can we make sure to document this behavior?

mlflow/pipelines/steps/evaluate.py
dbczumar: No worries! I think ingest is a good place for this.

mlflow/pipelines/steps/ingest/__init__.py
```diff
@@ -77,6 +93,21 @@ def _run(self, output_directory: str) -> BaseCard:
...
+                raise MlflowException(
+                    f"Target column '{self.target_col}' not found in ingested dataset.",
+                    error_code=INVALID_PARAMETER_VALUE,
+                )
+            if self.positive_class is not None:
+                cardinality = ingested_df.apply(pd.Series.nunique)[self.target_col]
+                if cardinality != 2:
+                    raise MlflowException(
+                        f"Target column '{self.target_col}' must have a cardinality of 2,"
+                        "found '{cardinality}'.",
```
harupy, line 107:
```suggestion
                        f"Target column '{self.target_col}' must have a cardinality of 2, "
                        f"found '{cardinality}'.",
```

Nit

mlflow/pipelines/steps/ingest/__init__.py
```diff
@@ -77,6 +93,21 @@ def _run(self, output_directory: str) -> BaseCard:
...
 
         ingested_df = read_parquet_as_pandas_df(data_parquet_path=dataset_dst_path)
+        if self.step_class() == StepClass.TRAINING:
+            if self.target_col not in ingested_df.columns:
+                raise MlflowException(
+                    f"Target column '{self.target_col}' not found in ingested dataset.",
+                    error_code=INVALID_PARAMETER_VALUE,
+                )
+            if self.positive_class is not None:
+                cardinality = ingested_df.apply(pd.Series.nunique)[self.target_col]
```
harupy, line 103:
```suggestion
                cardinality = ingested_df[self.target_col].nunique()
```

Can we only apply `nunique` to the target column?


### Benchmark:

```python
import numpy as np
import pandas as pd
import timeit

df = pd.DataFrame(np.random.randint(0, 10, size=(10000, 5)), columns=list("ABCDE"))


def subscript_then_nunique():
    return df["A"].nunique()


def nunique_then_subscript():
    return df.apply(pd.Series.nunique)["A"]


assert subscript_then_unique() == unique_then_subscript()

print(timeit.timeit(subscript_then_nunique, number=10000), "seconds")
print(timeit.timeit(nunique_then_subscript, number=10000), "seconds")
```

Output

```
0.542079824999746 seconds
4.465884477000145 seconds
```

PR #7086, opened 2022-10-18
Add tests for MLP Batch Scoring DAG

jerrylian-db:

As title

## How is this patch tested?

Components

Interface

Language

Integrations


tests/pipelines/helper_functions.py
```diff
@@ -80,6 +83,17 @@ def train_log_and_register_model(model_name, is_dummy=False):
...
 
 
+def get_execution_directory_location(pipeline_root_path, custom_execution_directory=None):
+    if custom_execution_directory:
+        return pathlib.Path(custom_execution_directory)
+    return (
+        pathlib.Path.home()
+        / ".mlflow"
+        / "pipelines"
+        / _get_execution_directory_basename(pipeline_root_path)
```
jinzhang21, line 93: Can we use `get_or_create_base_execution_directory` instead? 
https://src.dev.databricks.com/mlflow/mlflow/-/blob/mlflow/pipelines/utils/execution.py?L224&subtree=true

tests/pipelines/regression/v1/test_batch_scoring_dag.py
```diff
@@ -0,0 +1,64 @@
...
+    enter_pipeline_example_directory,
+    get_execution_directory_location,
+    PIPELINE_EXAMPLE_PATH_FROM_MLFLOW_ROOT,
+    chdir,
+)  # pylint: enable=unused-import
+
+_STEP_NAMES = ["ingest_scoring", "predict"]
+
+
+# Because the Batch Scoring DAG takes a lot of time to run (around 5 minutes), we run the entire
```
jinzhang21, line 20: What's the reason it takes so long to run? Data size? Spark start-up? Can we speed it up to make it finish within 1 min?

jerrylian-db: I think it is the environment restoration in the Spark UDF. When the predict step is run with the `p.run` API, I have no been able to find a way to patch the env_manager to be `"local"` to speed up the process. Thus it is quite slow.

Other reasons might be that it runs the entire training DAG + starts up spark.

jerrylian-db: I see two ways to fix this: one is to expose the env_manager in the step config. Two is to speed up environment restoration by detecting if the inference environment is the same as the training environment, and not materialize a new environment if that is the case.

jerrylian-db: I've also started a slack thread on this in the OSS channel. I'm currently not leaning towards exposing the env_manager in the step config, especially since we plan to deprecate the `local` env_manager in MLflow 2.0.

tests/pipelines/regression/v1/test_batch_scoring_dag.py
```diff
@@ -0,0 +1,64 @@
+import pandas as pd
```
jinzhang21, line 1: Mind adding these tests to `tests/pipelines/regression/v1/test_regression_pipeline.py` since they are from the same template?

tests/pipelines/regression/v1/test_batch_scoring_dag.py
jinzhang21: SG. Let's see how this could be better solved.

jinzhang21 (approved):

tests/pipelines/regression/v1/test_batch_scoring_dag.py
jerrylian-db: Given the special setup of this test suite (having a setup steps that creates the conditions that multiple steps check). I'm not sure that would work, unless we compile all the checks of this test file into one single test in that file. Would that be preferable?

jinzhang21: Can we at least change the file name to `test_regression_pipeline_batch_scoring.py` then?

PR #7105, opened 2022-10-19
Add fixed length random experiment_id support to SqlAlchemyStore

BenWilson2:

## Related Issues/PRs

 #xxx

Create a random fixed-length int64 for experiment_id (other than the default experiment_id) for consistency with thread-safe changes in FileStore. This removes the auto-increment behavior in the underlying databases for the primary key column in the `experiments` table (and removes the identity property in SQL SERVER from this field).  

## How is this patch tested?

experiment_id is no longer a monotonically increasing id. Instead, it is a fixed-length int64 that is generated randomly. 

Components
- [X] `area/server-infra`: MLflow Tracking server backend

Interface
- [X] `area/sqlalchemy`: Use of SQLAlchemy in the Tracking Service or Model Registry

Language

Integrations

- [X] `rn/breaking-change` - The PR will be mentioned in the "Breaking Changes" section

### DIFF
--- a/mlflow/models/docker_utils.py
+++ b/mlflow/models/docker_utils.py
@@ -96,6 +96,9 @@ RUN apt-get install -y --no-install-recommends \
          maven \
     && rm -rf /var/lib/apt/lists/*
 
+# Increase default timeout for API throttling
+ENV PIP_DEFAULT_TIMEOUT=100
+
 {setup_miniconda}
 {setup_pyenv_and_virtualenv}
 
--- a/mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
+++ b/mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
@@ -1,9 +1,12 @@
-"""convert experiment id to fixed length random
+"""convert experiment id to fixed length random integer
 
 Revision ID: 2c7e13772396
 Revises: 97727af70f4d
 Create Date: 2022-10-18 16:39:31.183846
 
+This migration converts the primary key `experiment_id` within the Experiments table from an
+auto-incrementing primary key to a non-nullable unique-constrained Integer column. This is to
+support concurrent experiment creation and avoid collisions.
 """
 from alembic import op
 import sqlalchemy as sa
--- a/mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
+++ b/mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
@@ -1,0 +1,205 @@
+"""convert experiment id to fixed length random
+
+Revision ID: 2c7e13772396
+Revises: 97727af70f4d
+Create Date: 2022-10-18 16:39:31.183846
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy import PrimaryKeyConstraint, ForeignKeyConstraint, Sequence
+from sqlalchemy.inspection import inspect
+import logging
+
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
+
+# revision identifiers, used by Alembic.
+revision = "2c7e13772396"
+down_revision = "97727af70f4d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # As part of MLflow 2.0 upgrade, the Experiment table's primary key `experiment_id`
+    # has changed from an auto-incrementing column to a non-nullable unique-constrained Integer
+    # column to support the uuid-based random id generation change.
+
+    engine = op.get_bind()
+    engine_name = engine.engine.name
+
+    # NB: sqlite doesn't support foreign keys even if they are defined. Altering a constraint
+    # in sqlite outside of batch operations doesn't work.
+    if engine_name != "sqlite":
+
+        foreign_keys_in_experiment_tags = inspect(engine).get_foreign_keys("experiment_tags")
+        fk = foreign_keys_in_experiment_tags[0]
+        op.drop_constraint(fk["name"], table_name="experiment_tags", type_="foreignkey")
+
+        # NB: MSSQL and MySQL have special restrictions on batch updates that foreign keys.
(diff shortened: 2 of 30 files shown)


mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
```diff
@@ -0,0 +1,205 @@
+"""convert experiment id to fixed length random
```
harupy, line 1:
```suggestion
"""convert experiment id to fixed length random integer
```

Is `integer` missing?

tests/store/tracking/test_file_store.py
```diff
@@ -182,11 +182,11 @@ def test_search_experiments_filter_by_attribute(self):
         experiments = self.store.search_experiments(filter_string="attribute.`name` = 'a'")
         assert [e.name for e in experiments] == ["a"]
         experiments = self.store.search_experiments(filter_string="attribute.`name` != 'a'")
-        assert [e.name for e in experiments] == ["Abc", "ab", "Default"]
+        assert {e.name for e in experiments} == {"Abc", "ab", "Default"}
```
harupy, line 185: Is the experiment order not detereministic?

harupy: If it's not deterministic, we can also use `sorted`:

```suggestion
        assert sorted(e.name for e in experiments) == ["Abc", "ab", "Default"]
```

tests/projects/test_projects.py
```diff
@@ -64,16 +64,19 @@ def clean_mlruns_dir():
...
+        ("add another", None, "generated"),
     ],
 )
 def test_resolve_experiment_id(experiment_name, experiment_id, expected):
-    assert expected == _resolve_experiment_id(
-        experiment_name=experiment_name, experiment_id=experiment_id
-    )
+    actual_id = _resolve_experiment_id(experiment_name=experiment_name, experiment_id=experiment_id)
+    if expected == "generated":
+        assert _EXPERIMENT_ID_FIXED_WIDTH == len(actual_id)
```
harupy, line 77:
```suggestion
        assert len(actual_id) == _EXPERIMENT_ID_FIXED_WIDTH
```

mlflow/models/docker_utils.py
```diff
@@ -96,6 +96,9 @@ def _get_maven_proxy():
          maven \
     && rm -rf /var/lib/apt/lists/*
 
+# Increase default timeout for API throttling
+ENV PIP_DEFAULT_TIMEOUT=100
```
harupy, line 100: Why do we need this change?

BenWilson2: While I was debugging failures in that draft branch (and some PR test failures over the last few days, I noticed a few pip install failures with timeout exceptions from the package repository. These intermittent failures generally resolve themselves but the default timeout doesn't allow for enough total time with the 5x retry policy within `pip install`. I checked a few discussion boards where people were seeing these intermittent failures in VMs that are building multiple Docker containers for CI and the general consensus was to use a container-level global config for pip (not in the main CI) of overriding https://pip.pypa.io/en/stable/topics/configuration/?highlight=timeout#environment-variables . If we don't want to do this, I'll remove it :)

tests/store/tracking/test_file_store.py
BenWilson2: It was showing different ordering between linux and windows with the default ordering change from `experiment_id` to either `last_update_time` or `creation_time`. I think this might be resolved by merging branch-2.0 though, so I'll revert these changes and only apply a `sorted()` comparison to the lists iif the platform differences exist. 
I'll let you know when I do the branch merge and conflict resolution :)

tests/projects/test_projects.py
BenWilson2: definitely!

mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
BenWilson2: good catch. I'll update

mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
```diff
@@ -0,0 +1,208 @@
...
+_logger.setLevel(logging.INFO)
+
+# revision identifiers, used by Alembic.
+revision = "2c7e13772396"
+down_revision = "97727af70f4d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
dbczumar, line 27: Given the complexity of this migration, do we have test coverage that ensures a migration from 1.30 -> 2.0 succeeds on all three databases with preexisting data in the `experiments` and `runs` table?

If such a test doesn't already exist, I'm admittedly a bit concerned that we might miss something in this migration and block folks from upgrading to MLflow 2.0. Given that there doesn't seem to be a big motivation for this change beyond creating a consistent experience with FileStore (SQLAlchemy store already has nice concurrency properties for experiment operations), I might suggest we hold off unless the test coverage is already present. Let me know what you think here.

tests/store/tracking/test_file_store.py
harupy: The difference comes from the clock precision. As described in https://stackoverflow.com/questions/1938048/high-precision-clock-in-python, Python on Windows uses +- 16 milliseconds precision. 16 milliseconds is long enough to create multiple experiments.

harupy: Just sleeping should fix the issue.

mlflow/store/db_migrations/versions/2c7e13772396_convert_experiment_id_to_fixed_length_.py
BenWilson2: This is a very valid point. We could add a new test suite that validates migration backwards compatibility. I feel as though the time commitment to building that out is significantly higher than the effort involved in this PR. 
How about we hold off on any further work here until when/if it becomes an actual necessity (unless we think that pursuing a migration validation test suites is worthwhile for 2.0?).

PR #7108, opened 2022-10-19
Introduce pipeline helper objects to abstract template-specific logic from step implementations

bbarnes52:

This PR introduces an interface for template-specific logic required by step implementations. This interface is expected to grow in the immediate future, as steps currently implement a nontrival amount of logic that is specific to regression (step cards in particular). The idea here is that steps are agnostic to the template in which they run (evidenced by removing the `template` field from TrainStep and EvaluateStep) and the pipeline helper factory method is the only place in which we switch on the template type. 

## How is this patch tested?

Locally using mlp-regression-example

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/pipelines/helpers.py
+++ b/mlflow/pipelines/helpers.py
@@ -1,0 +1,111 @@
+from abc import ABC, abstractmethod
+
+from mlflow.exceptions import MlflowException
+from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
+
+from typing import Optional
+
+
+class PipelineHelper(ABC):
+    @abstractmethod
+    def error_fn(self):
+        """
+        :return: The error function for the template type.
+        """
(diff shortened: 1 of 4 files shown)


mlflow/pipelines/helpers.py
```diff
@@ -0,0 +1,111 @@
+from abc import ABC, abstractmethod
+
+from mlflow.exceptions import MlflowException
+from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
+
+from typing import Optional
+
+
+class PipelineHelper(ABC):
```
prithvikannan, line 9: I think this approach looks great currently. One thing that I am worried is that as we build out more templates and more complex templates this `PipelineHelper` class may contain arbitrary functions without a cohesive story.

Is this a problem? And if so, how would we solve it?

PR #7124, opened 2022-10-20
Handle read-only model mounts during virtualenv creation

dbczumar:

Handle read-only model mounts during virtualenv creation

## How is this patch tested?

Added unit tests. Also manually tested the case where symlinking fails.

- [ ] Yes. Make sure the changed pages / sections render correctly in the documentation preview.

## Release Notes

Handle read-only model mounts during virtualenv creation

Components

Interface

Language

Integrations

### DIFF
--- a/mlflow/utils/virtualenv.py
+++ b/mlflow/utils/virtualenv.py
@@ -243,13 +243,13 @@ def _create_virtualenv(
             # in it correctly. To do this, we must first symlink or copy the model directory's
             # contents to a temporary location for compatibility with deployment tools that store
             # models in a read-only mount
-            model_dir = t.path("model")
+            tmp_model_dir = t.path("model")
+            os.makedirs(tmp_model_dir)
             try:
-                os.makedirs(model_dir)
                 for model_item in os.listdir(local_model_path):
                     os.symlink(
                         src=os.path.join(local_model_path, model_item),
-                        dst=os.path.join(model_dir, model_item),
+                        dst=os.path.join(tmp_model_dir, model_item),
                     )
             except Exception as e:
                 _logger.warning(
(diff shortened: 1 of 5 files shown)


mlflow/utils/virtualenv.py
```diff
@@ -237,15 +238,32 @@ def _create_virtualenv(
...
+        with TempDir() as t:
+            # Create a temporary requirements file in the model directory to resolve the references
+            # in it correctly. To do this, we must first symlink or copy the model directory's
+            # contents to a temporary location for compatibility with deployment tools that store
+            # models in a read-only mount
+            model_dir = t.path("model")
+            try:
+                os.makedirs(model_dir)
+                for model_item in os.listdir(local_model_path):
+                    os.symlink(
```
BenWilson2, line 250: Do we have to specify `target_is_directory` arg for Windows since symbolic links are handled autonomously only in developer mode (I think there are 2 distinct implementations based on whether the target (src) is a file or a directory, but I may be wrong here...)?

dbczumar: Looks like we don't based on the following: 

```
On Windows, a symlink represents either a file or a directory, and does not morph to the target dynamically. If the target is present, the type of the symlink will be created to match. 
```

tests/pyfunc/test_virtualenv.py
```diff
@@ -65,11 +67,16 @@ def test_serve_and_score(sklearn_model):
 
 
 @use_temp_mlflow_env_root
-def test_restore_environment_with_virtualenv(sklearn_model):
-    with mlflow.start_run():
-        model_info = mlflow.sklearn.log_model(sklearn_model.model, artifact_path="model")
-
-    scores = serve_and_score(model_info.model_uri, sklearn_model.X_pred)
+def test_serve_and_score_read_only_model_directory(sklearn_model, tmp_path):
```
dbczumar, line 70: This test is WIP

mlflow/utils/virtualenv.py
```diff
@@ -237,15 +238,32 @@ def _create_virtualenv(
...
+                shutil.rmtree(model_dir)
+                shutil.copytree(local_model_path, model_dir)
+
+            tmp_req_file = f"requirements.{uuid.uuid4().hex}.txt"
+            Path(model_dir).joinpath(tmp_req_file).write_text("\n".join(deps))
             cmd = _join_commands(activate_cmd, f"python -m pip install --quiet -r {tmp_req_file}")
-            _exec_cmd(cmd, capture_output=capture_output, cwd=local_model_path, extra_env=extra_env)
-        finally:
-            local_model_path.joinpath(tmp_req_file).unlink()
+            _exec_cmd(cmd, capture_output=capture_output, cwd=model_dir, extra_env=extra_env)
```
BenWilson2, line 266: does it matter if there are hanging symlinks? Would they need to have `os.unlink()` called on the destination path or is that irrelevant?

tests/pyfunc/test_virtualenv.py
dbczumar: Works

mlflow/utils/virtualenv.py
dbczumar: `os.unlink()` is the same as `os.remove()`. The key point here is that the `TempDir` context manager removes the temp directory upon context manager exit, and remove / unlink deletes symlinks too.

harupy (approved):
LGTM!

tests/resources/example_virtualenv_project/requirements.txt
```diff
@@ -1,2 +1,2 @@
-../../..
+mlflow
```
dbczumar, line 1: @harupy CI caught a change in this PR where we can no longer support requirements files that make relative path references to locations outside the directory where the requirement file is located. This seems okay to me, since relying on this kind of relative pathing is brittle / probably a bad practice. I've instead swapped in `mlflow` from PyPI here since the example project doesn't require any special MLflow code from a dev branch. Let me know what you think.

harupy: This actually fails when the local mlflow contains a new migration script that the latest version of MLflow on PyPI doesn't.

dbczumar: Got it! Added some logic to install MLflow from dev

tests/projects/test_virtualenv_projects.py
```diff
@@ -18,6 +20,33 @@
...
+    mlflow_root = os.path.dirname(os.path.dirname(mlflow.__file__))
+
+    conda_env = read_yaml(TEST_VIRTUALENV_CONDA_PROJECT_DIR, "conda.yaml")
+    conda_pip_dependencies = [
+        item for item in conda_env["dependencies"] if isinstance(item, dict)
+        and "pip" in item
+    ][0]["pip"]
+    if "mlflow" in conda_pip_dependencies:
+        conda_pip_dependencies.remove("mlflow")
+        conda_pip_dependencies.append(mlflow_root)
```
BenWilson2, line 34: clever :)

BenWilson2 (approved):
LGTM! nice fix!

dbczumar:
@mlflow-automation autoformat

